Issue #169, opened 2021-06-03
Allow entry-point for `rmc` other than `main`


vecchiot-aws:
CBMC allows us to specify an entry-point other than `main`, which would be useful for analyzing a library crate, for example. However, if we try to do this, we get an error from `rmc-rustc`:

File (test.rs):
```
fn main2() {
    assert!(true);
}
```

Command: `rmc --verbose test.rs -- --function main2`

Error:
```
[RMC] stage: compile (FAIL)
[RMC] cmd: rmc-rustc -Z codegen-backend=gotoc --cfg=rmc -o test.json test.rs
error[E0601]: `main` function not found in crate `test`
 --> test.rs:1:1
  |
1 | / fn main2() {
2 | |     assert!(true);
3 | | }
  | |_^ consider adding a `main` function to `test.rs`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0601`.
```

vecchiot-aws:
Note: I believe this only applies to `rmc`, not `cargo-rmc`. If we go to `src/test/cargo-rmc/simple-lib` and run `cargo-rmc --function test_sum .`, verification works properly.

celinval:
I believe the problem here is that we are trying to compile the crate as an executable. We should be able to combine `--function` with  `--crate-type=lib` construction.

PR #239, opened 2021-06-21
Added --function flag to cargo-rmc

vecchiot-aws:
### Description of changes: 

Currently, we hard-code the entry point for both `rmc` and `cargo-rmc` to be `main` (see #169). However, this issue only applies to `rmc`, not `cargo-rmc`, so we can add the `--function` flag in for `cargo-rmc`.

### Testing:

### Checklist

### DIFF
--- a/src/tools/compiletest/src/runtest.rs
+++ b/src/tools/compiletest/src/runtest.rs
@@ -2430,11 +2430,10 @@ impl<'test> TestCx<'test> {
         let function_name = self.testpaths.file.file_stem().unwrap().to_str().unwrap();
         cargo
             .arg("rmc")
+            .args(["--function", function_name])
             .arg("--target")
             .arg(self.output_base_dir().join("target"))
-            .arg(&parent_dir)
-            .arg("--")
-            .args(["--function", function_name]);
+            .arg(&parent_dir);
         self.add_rmc_dir_to_path(&mut cargo);
         let proc_res = self.compose_and_run_compiler(cargo, None);
         let expected = fs::read_to_string(self.testpaths.file.clone()).unwrap();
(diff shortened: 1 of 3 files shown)


avanhatt:
Just to confirm: this isn't yet plumbed all the way through, right? I think it'll need some mucking with the readable names for this to work within RMC.

vecchiot-aws:
This is specifically addressing the test at `src/test/cargo-rmc/simple-lib/`. This test requires you to run `cargo-rmc` with the `--function` CBMC flag, so adding this as an RMC flag simplifies the command, and seems to be working fine.

adpaco-aws:
This change is not being tested so we do not know if it's working.

The current `cargo-rmc` tests invoke `cargo rmc` with arguments that depend on `*.expected` files present in each crate under test. This is done by function `run_cargo_rmc_test` in `src/tools/compiletest/src/runtest.rs`. Note that `--function` is always passed with `function_name` there.

You can extend the code in compiletest to consider a new test with no `*.expected` files. Talk to @bdalrhm if you have trouble understanding the code in there.

src/tools/compiletest/src/runtest.rs
```diff
@@ -2430,11 +2430,11 @@ impl<'test> TestCx<'test> {
         let function_name = self.testpaths.file.file_stem().unwrap().to_str().unwrap();
         cargo
             .arg("rmc")
+            .args(["--function", function_name])
             .arg("--target")
             .arg(self.output_base_dir().join("target"))
             .arg(&parent_dir)
-            .arg("--")
-            .args(["--function", function_name]);
+            .arg("--");
```
adpaco-aws, line 2437: Can you remove .arg("--")? I do not think we are passing CBMC options anymore.

adpaco-aws (approved):

Issue #189, opened 2021-06-09
Can we remove "assertion failed" from successful assertions?


avanhatt:
When viewing a CBMC trace from RMC, a user might see something like this:
```
[main.assertion.1] line 28 assertion failed: std_string == \"1\": FAILURE
[main.assertion.2] line 30 assertion failed: weird_string == \"w\": SUCCESS
```

The top assertion is failing and the bottom is provably true, but both have `assertion failed`. I think this is just an artifact of how Rust proper prints assertions; it would be nice to remove this for CBMC usability.

adpaco-aws:
Yes, I think these come specifically from `compiler/rustc_middle/src/mir/mod.rs`.

It is true that it may be confusing at times but RMC has nothing to do with this: the assertion text comes from MIR, the verification output comes from CBMC.

PR #263, opened 2021-06-25
Use default CBMC flags in RMC

### LINKED ISSUE #156: RMC unwinding-assertions should be on by default
RMC should use `--unwinding-assertions` by default and have an option to disable the checks. The tool should suggest increasing loop bounds or disabling unwinding assertions if these assertions fail.

adpaco-aws:
### Description of changes: 

This PR adds a couple of features to the RMC script:
  1. The requested feature to have a set of default CBMC flags used by RMC, which as of now consists of `--unwinding-assertions`.
  2. The scanner class which is used for matching text within CBMC's output an edit it. At present, there is a single scanner instance that looks for failed unwinding assertions in CBMC's output and prints a tip if it founds any.

I wrote the scanner class while also thinking about issue #189 , we would only need to write another instance that replaces text based on the failing assertion pattern.

### Resolved issues:

Resolves #156 

### Call-outs:

This PR is quite large because the goal was to solve #156. We can split it into two smaller PRs if that will make reviewing easier.

### Testing:

* How is this change tested? Existing regression, plus a couple of new (positive) `cbmc` tests for testing `--no-default-flags` and another `expected` that checks the unwind failure recommendation is printed.

### Checklist


scripts/rmc.py
```diff
@@ -31,6 +48,17 @@ def delete_file(filename):
     except OSError:
         pass
 
+# Add a set of default CBMC arguments
+def add_default_cbmc_flags(args):
+    if not args.no_default_flags:
```
vecchiot-aws, line 53: Is there a reason we block on `no_default_flags` both here and in `rmc`/`cargo-rmc`?

nchong-at-aws:
Looks good Adrian! A request: a way for the user to enumerate the default flags so they can get the information without reading source.

danielsn:
I think we'd be better off breaking the default flags into groups:
`unwinding-assertions`
`memory-safety`
`overflow-safety`
Then, we could have a the following flags
`no-memory-safety-checks`
`no-overflow-safety-checks`
`no-unwinding-checks`
and a 
`no-default-checks` which implies all of them

scripts/rmc.py
adpaco-aws: It's a mistake, thanks for pointing this out. The condition is always checked before calling this function.

adpaco-aws:
> I think we'd be better off breaking the default flags into groups:
> `unwinding-assertions`
> `memory-safety`
> `overflow-safety`

I have changed the proposal to include the mentioned groups, but we have to discuss the specific flags for each group. In particular, what memory safety checks we want to enable?

> Looks good Adrian! A request: a way for the user to enumerate the default flags so they can get the information without reading source.

Thanks! I have added a flag `--print-cbmc-flags` that prints what CBMC flags will be used in an RMC invocation. Unfortunately, this does not work if the user does not specify an input file (as in a regular invocation) since the parser does not allow it. We could revisit this in the future.

scripts/cargo-rmc
```diff
@@ -31,6 +31,11 @@ def main():
     parser.add_argument("--wkdir", default=".")
     parser.add_argument("--function", default="main")
     parser.add_argument("crate", help="crate to verify")
+    parser.add_argument("--no-default-checks", action="store_true", help="Disable all default checks")
+    parser.add_argument("--no-memory-safety-checks", action="store_true", help="Disable default memory safety checks")
+    parser.add_argument("--no-overflow-safety-checks", action="store_true", help="Disable default overflow checks")
+    parser.add_argument("--no-unwinding-checks", action="store_true", help="Disable default unwinding checks")
+    parser.add_argument("--print-cbmc-flags", action="store_true", help="Print CBMC flags associated to an RMC invocation")
```
danielsn, line 38: Maybe make this a `--dry-run` flag?

danielsn:
We might also want to consider
- [ ]  `--div-by-zero-check          enable division by zero checks`
- [ ] ` --conversion-check           check whether values can be represented after type cast`
- [ ] ` --undefined-shift-check      check shift greater than bit-width`
- [ ] `-float-overflow-check       check floating-point for +/-Inf`
- [ ] `--nan-check                  check floating-point for NaN`
- [ ] `--enum-range-check           checks that all enum type expressions have values in the enum range`
- [ ] `--pointer-primitive-check    checks that all pointers in pointer primitives are valid or null`
- [ ] ` --malloc-may-fail            allow malloc calls to return a null pointer`
- [ ] ` --malloc-fail-null           set malloc failure mode to return null`

src/test/firecracker/virtio-block-parse/main.rs
```diff
@@ -1,13 +1,10 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-// --unwinding-assertions
-//
+
+// rmc-flags: --no-overflow-checks
```
danielsn, line 4: Whenever we remove a check, lets have a comment explaining why, and maybe a tracking issue if its worth it

src/test/expected/dry-run/expected
```diff
@@ -0,0 +1,3 @@
+rmc-rustc -Z codegen-backend=gotoc -Z symbol-mangling-version=v0 --cfg=rmc
+symtab2gb
+cbmc --pointer-check --pointer-primitive-check --bounds-check --conversion-check --div-by-zero-check --float-overflow-check --nan-check --pointer-overflow-check --signed-overflow-check --undefined-shift-check --unsigned-overflow-check --unwinding-assertions --function main
```
danielsn, line 3: newline

src/test/cbmc/Whitespace/main.rs
```diff
@@ -1,7 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-// cbmc-flags: --unwind 2 --unwinding-assertions
+// rmc-flags: --no-overflow-checks
```
danielsn, line 4: why?

src/test/cbmc/DynTrait/boxed_closure.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+//rmc-flags: --no-memory-safety-checks
```
danielsn, line 4: why?

scripts/rmc
```diff
@@ -41,6 +46,10 @@ def main():
 
     if not rmc.dependencies_in_path():
         return 1
+    
+    # Add some CBMC flags by default unless `--no-default-checks` is being used
+    if not args.no_default_checks:
+        rmc.add_default_cbmc_flags(args)
```
danielsn, line 52: it wasn't clear to me from the name that this only adds some default flags (controlled by args)

scripts/rmc
```diff
@@ -27,6 +27,11 @@ def main():
     parser.add_argument("--visualize", action="store_true")
     parser.add_argument("--srcdir", default=".")
     parser.add_argument("--wkdir", default=".")
+    parser.add_argument("--no-default-checks", action="store_true", help="Disable all default checks")
+    parser.add_argument("--no-memory-safety-checks", action="store_true", help="Disable default memory safety checks")
```
danielsn, line 31: Do we want a way to selectively add these back in?

danielsn (changes requested):

danielsn:
For the future, the `scanner` is a big enough change it probably deserves its own PR. But for this one its OK

scripts/rmc
adpaco-aws: The user can always add them back in with `rmc --no-memory-safety-checks file.rs -- <cbmc-flags>`

scripts/rmc
adpaco-aws: Not sure I follow. What changes to this are you suggesting?

adpaco-aws:
Updated with explanations on each test that removes default checks. CI is failing due to `rustfmt` but I think @avanhatt has is taking care of it in #304 - In the meantime, let me know if we need other changes.

> For the future, the `scanner` is a big enough change it probably deserves its own PR. But for this one its OK

Yes, I agree.

scripts/rmc
danielsn: naming it something like `add_selected_default_cbmc_flags`

scripts/rmc.py
```diff
@@ -5,12 +5,40 @@
...
+import re
 
 
 RMC_CFG = "rmc"
 RMC_RUSTC_EXE = "rmc-rustc"
 EXIT_CODE_SUCCESS = 0
 
+MEMORY_SAFETY_CHECKS = ["--pointer-check",
+                        "--pointer-primitive-check",
+                        "--bounds-check"]
```
danielsn, line 17: put these in alphabetical order

scripts/rmc.py
```diff
@@ -31,11 +59,29 @@ def delete_file(filename):
...
         pass
 
+# Add a set of CBMC flags to the CBMC arguments
+def add_set_cbmc_flags(args, flags):
+    # We print a warning if the user has passed the flag via `cbmc_args`
+    # Otherwise we append it to the CBMC arguments
+    for arg in flags:
+        # This behavior must be reviewed if the set of flags is extended
+        if arg in args.cbmc_args:
+            print("WARNING: Default CBMC argument `{}` not added (already specified)".format(arg))
```
danielsn, line 69: might be useful to also print the `cbmc_args` here

scripts/rmc.py
```diff
@@ -117,27 +186,27 @@ def run_visualize(cbmc_filename, cbmc_args, verbose=False, quiet=False, keep_tem
...
-    run_goto_cc(cbmc_filename, temp_goto_filename, verbose, quiet, function=function)
+    run_goto_cc(cbmc_filename, temp_goto_filename, verbose, quiet, function=function, dry_run=dry_run)
     
-    def run_cbmc_local(cbmc_args, output_to):
+    def run_cbmc_local(cbmc_args, output_to, dry_run=False):
         cbmc_cmd = ["cbmc"] + cbmc_args + [temp_goto_filename]
-        return run_cmd(cbmc_cmd, label="cbmc", output_to=output_to, verbose=verbose, quiet=quiet)
+        return run_cmd(cbmc_cmd, label="cbmc", output_to=output_to, verbose=verbose, quiet=quiet, dry_run=dry_run)
 
     cbmc_xml_args = cbmc_args + ["--xml-ui"]
```
danielsn, line 195: putting all the property checks into the coverage checks is going to be super expensive. Lets split these out.

danielsn (changes requested):

scripts/rmc.py
adpaco-aws: Have to disagree with this. I expect the user to make a dry run if he really wants to see them. Plus, at this point `cbmc_args` is not complete (there may be other flags to add) so it could be misleading.

scripts/rmc.py
adpaco-aws: Added another argument here for coverage checks even though I pass an empty list. The starter kit does not include any by default except for `--malloc-may-fail` and `--malloc-fail-null`, which we did not consider as default flags.

scripts/rmc
adpaco-aws: Done.

nchong-at-aws (approved):

PR #264, opened 2021-06-25
Regression test for codegen'ing Firecracker

### LINKED ISSUE #262: Create regression workflow for codegen of virtio module

nchong-at-aws:
### Description of changes: 

No changes to RMC. Add codegen test for Firecracker virtio module to regression.

### Resolved issues:

Resolves #262

### Checklist

### DIFF
--- a/scripts/codegen-firecracker.sh
+++ b/scripts/codegen-firecracker.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+# Test for platform
+platform=`uname -sp`
+if [[ $platform != "Linux x86_64" ]]; then
+  echo "Codegen script only works on Linux x86 platform"
+  exit 0
+fi
+
+# At the moment, we only test codegen for the virtio module
+cd /tmp
+git clone https://github.com/firecracker-microvm/firecracker.git
+cd firecracker/src/devices/src/virtio/
+RUST_BACKTRACE=1 RUSTFLAGS="-Z trim-diagnostic-paths=no -Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo build --target x86_64-unknown-linux-gnu -j1
--- a/scripts/codegen-firecracker.sh
+++ b/scripts/codegen-firecracker.sh
@@ -13,4 +13,4 @@ fi
 cd /tmp
 git clone https://github.com/firecracker-microvm/firecracker.git
 cd firecracker/src/devices/src/virtio/
-RUST_BACKTRACE=1 RUSTFLAGS="-Z trim-diagnostic-paths=no -Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo build --target x86_64-unknown-linux-gnu -j1
+RUST_BACKTRACE=1 RUSTFLAGS="-Z trim-diagnostic-paths=no -Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo build --target x86_64-unknown-linux-gnu
--- a/scripts/codegen-firecracker.sh
+++ b/scripts/codegen-firecracker.sh
@@ -9,8 +9,11 @@ if [[ $platform != "Linux x86_64" ]]; then
   exit 0
 fi
 
+# Get RMC root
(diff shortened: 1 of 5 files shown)


scripts/codegen-firecracker.sh
```diff
@@ -0,0 +1,16 @@
...
+# Test for platform
+platform=`uname -sp`
+if [[ $platform != "Linux x86_64" ]]; then
+  echo "Codegen script only works on Linux x86 platform"
+  exit 0
+fi
+
+# At the moment, we only test codegen for the virtio module
+cd /tmp
+git clone https://github.com/firecracker-microvm/firecracker.git
```
danielsn, line 14: We might want to pin this to a particular commit?

scripts/rmc-regression.sh
```diff
@@ -22,3 +22,6 @@ check-cbmc-viewer-version.py --major 2 --minor 5
 # Standalone rmc tests, expected tests, and cargo tests
 ./x.py build -i --stage 1 library/std ${EXTRA_X_PY_BUILD_ARGS}
 ./x.py test -i --stage 1 cbmc firecracker prusti smack expected cargo-rmc
+
+# Check codegen of firecracker
+$SCRIPT_DIR/codegen-firecracker.sh
```
adpaco-aws, line 27: What about adding this as a separate workflow?

adpaco-aws (commented):
Seems we have regressed already :(

scripts/rmc-regression.sh
nchong-at-aws: Yes, can do. I was putting it into the regression to avoid rebuilding RMC in two different flows. Is there a way we can share artifacts across?

adpaco-aws: We have been looking into this but we are not sure if it's good for caching RMC builds: https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts

But since it's not clear, any solution would be fine with me. I would like to revisit in the future to have 1 workflow that builds and multiple workflows which depend on that one to execute the regression.

scripts/codegen-firecracker.sh
robin-aws: Could this be a submodule instead? That's a standard way to pin a specific commit, plus tooling support like Dependabot. :)

avanhatt:
It looks like the CI failure might be fixed in https://github.com/model-checking/rmc/issues/259, merging to rerun tests.

avanhatt:
Filed to track new crash: https://github.com/model-checking/rmc/issues/272

avanhatt:
Added a commit that appears to fix the regression; if we want to keep this PR to no changes to RMC I can break it off into a separate PR.

scripts/codegen-firecracker.sh
```diff
@@ -0,0 +1,16 @@
...
+cd firecracker/src/devices/src/virtio/
+RUST_BACKTRACE=1 RUSTFLAGS="-Z trim-diagnostic-paths=no -Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo build --target x86_64-unknown-linux-gnu -j1
```
adpaco-aws, line 16: Do we need `-j1` here?

adpaco-aws (commented):
I have changed my mind: Adding to normal regression is fine with me.

scripts/codegen-firecracker.sh
nchong-at-aws: Nope, we can remove. This does make errors easier to see when we get a failure though (since we parse each package serially)

nchong-at-aws:
Ah, we need the copyright check to ignore Firecracker

adpaco-aws:
I'm fine with overriding the copyright check for this PR. The regression failure is a concern though, I think it is due to `rustfmt` so please try to exclude it as we do for other submodules in `rustfmt.toml`.

nchong-at-aws:
Rustfmt now ignores firecracker submodule. Will deal with the copyright check next.

avanhatt (approved):
Looks good to me!

adpaco-aws (approved):

PR #260, opened 2021-06-24
C lib flag

vecchiot-aws:
### Description of changes: 

Currently, to pass external C files to RMC, you have to pass them as CBMC flags. Additionally, this caused them to not be added to the symbol table, and so linking did not work properly in visualizer reports.

This patch adds a `--c-lib` flag to both `rmc` and `cargo-rmc`, additionally fixing the missing links to external C files in the visualizer.

Example usage:

```
rmc --c-lib helper.c -- main.rs
rmc --c-lib helper.c --end-rmc-flags main.rs
rmc --c-lib helper.c --visualize main.rs
rmc --c-lib helper.c --end-rmc-flags main.rs -- --show-symbol-table
rmc --c-lib helper.c --visualize main.rs -- --show-symbol-table

cargo-rmc --function test_sum --c-lib src/helper.c -- .
cargo-rmc --function test_sum --c-lib src/helper.c --end-rmc-flags .
cargo-rmc --function test_sum --c-lib src/helper.c --visualize .
cargo-rmc --function test_sum --c-lib src/helper.c --end-rmc-flags . -- --show-symbol-table
cargo-rmc --function test_sum --c-lib src/helper.c --visualize . -- --show-symbol-table
```

### Resolved issues:

Resolves #246 

### Call-outs:

There is not test for the `--c-lib` flag for `cargo-rmc` since there is no way of writing such a test.

### Testing:

* How is this change tested? Existing regression suite with new `rmc` test.

* Is this a refactor change? No.

### Checklist
- [x] Each commit message has a non-empty body, explaining why the change was made
- [x] Methods or procedures are documented
- [x] Regression or unit tests are included, or existing tests cover the modified code
- [x] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #273, opened 2021-06-29
Support passing options to cargo-rmc.

bdalrhm:
### Description of changes: 

This PR supports passing options to `cargo-rmc` in `src/test/cargo-rmc` testing suite. It does so by requireing each tested function to have a `<fun-name>.config` file along with `<fun-name>.expected`. Just like for `cbmc` tests, this is a temporary solution until RMC supports a proper API with macros to pass options.

### Resolved issues:

None. However, this PR should help @vecchiot-aws test the behavior of `cargo-rmc` with external C files in #260.

### Testing:

* How is this change tested?
  * `src/simple-main` was slightly modified to test the feature
* Is this a refactor change?
  * No.

### Checklist

### DIFF
--- a/src/test/cargo-rmc/simple-lib/test_one_plus_two.config
+++ b/src/test/cargo-rmc/simple-lib/test_one_plus_two.config
@@ -1,0 +1,2 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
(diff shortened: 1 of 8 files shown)


adpaco-aws:
I talked to @vecchiot-aws about this and he mentioned that the work he's doing on the `--c-lib` flag should not need these config files because the compilation of C libraries should work out of the box with `cargo-rmc`.

@vecchiot-aws we would appreciate your input.

vecchiot-aws:
Agreed. The change will make it so that no flags need to be passed into `cargo-rmc` from command line.

bdalrhm:
OK. In that case, there is no need for the options anymore. Closing the PR.

PR #278, opened 2021-06-30
Handle boxed closures as arguments (fix RMC std lib crash)

### LINKED ISSUE #220: Standard library function `"backtrace_rs::backtrace::trace_unsynchronized"` does not codegen (closure crash)
```
error: internal compiler error: compiler/rustc_middle/src/ty/fold.rs:829:17: Trying to collect bound vars with a bound region: DebruijnIndex(0) BoundRegion { var: 0, kind: BrAnon(0) }
```
related #109

avanhatt:
### Description of changes: 

In order to codegen the std library, we need to handle a relatively niche Rust feature: boxed closure arguments such as ` Box<dyn FnOnce()>`. 

Previously, RMC would crash on this:
```rust
    // Create a boxed once-callable closure
    let f: Box<dyn FnOnce()> = Box::new(|| {});
    f();
```

We previously "skipped" operands that had a `Dynamic` kind, this patch now has a special case to pass along the fat pointer in that case. Not the most elegant solution, but works for all regression tests and lets us get significantly further into the standard library. 

### Resolved issues:

Resolves #220.

### Call-outs:

- Added more assertions to the path for codegening operands/projections.
- Refactored funtion call code to be more explicit. 
- There is still something weird happening with codegening closures; the resulting GotoC code has some bad pointer dereferences that can lead to unsound verification results. Tracking that in https://github.com/model-checking/rmc/issues/240. 

### Testing:

* How is this change tested?

A new regression test for the minimal boxed closure example from above. 

* Is this a refactor change?

Somewhat; I refactor function call code to be more explicit about cases.

### Checklist


src/test/cbmc/DynTrait/boxed_closure.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
```
avanhatt, line 1: Add copyright

avanhatt (commented):
DSN live review

avanhatt:
TODO: add `_fail_fixme`

compiler/rustc_codegen_llvm/src/gotoc/intrinsic.rs
```diff
@@ -453,7 +453,6 @@ impl<'tcx> GotocCtx<'tcx> {
             "prefetch_read_instruction" => unimplemented!(),
             "prefetch_write_data" => unimplemented!(),
             "prefetch_write_instruction" => unimplemented!(),
-            "try" => unimplemented!(),
```
avanhatt, line 456: See above:
```rust
"try" => {
                codegen_unimplemented_intrinsic!("https://github.com/model-checking/rmc/issues/267")
            }
```

adpaco-aws:
> TODO: add `_fail_fixme`

Resolved. PR includes `src/test/cbmc/DynTrait/boxed_closure_fail_fixme.rs`.

adpaco-aws (approved):

PR #280, opened 2021-07-01
Add support for multiple expected failures.

### LINKED ISSUE #249: Allow multiple failures in a test with _VERIFIER_expect_fail

bdalrhm:
### Description of changes: 

This PR adds support for multiple expected failures. For each expected failure, we ensure that there is a corresponding failure in RMC's output.

### Resolved issues:

Resolves #249

### Call-outs:

The changes in this PR only ensure that assertions that are expected to fail, do indeed fail. It does not ensure that those assertions are the only ones that fail. For example, if a test contains both normal assertions AND expected failures, only expected failures are checked and that test might pass even if normal assertions fail.

### Testing:

* How is this change tested?
  * Tests that use `__VERIFIER_expect_fail` now work as expected
* Is this a refactor change?
  * No

### Checklist

### DIFF
--- a/src/test/cbmc/DynTrait/vtable_size_align_drop_fail.rs
+++ b/src/test/cbmc/DynTrait/vtable_size_align_drop_fail.rs
@@ -13,6 +13,7 @@
 #![allow(deprecated)]
 
 use std::intrinsics::size_of;
+use std::ptr::drop_in_place;
 
 include!("../Helpers/vtable_utils_ignore.rs");
 include!("../../rmc-prelude.rs");
@@ -68,7 +69,7 @@ fn main() {
 
     // Check layout/values for Sheep
     unsafe {
-        let animal_sheep = random_animal(1);
+        let animal_sheep = &*random_animal(1);
 
         // Check that the struct's data is what we expect
         let data_ptr = data!(animal_sheep);
@@ -90,15 +91,15 @@ fn main() {
     }
     // Check layout/values for Cow
     unsafe {
-        let animal_sheep = random_animal(1);
+        let animal_cow = &*random_animal(6);
 
         // Check that the struct's data is what we expect
-        let data_ptr = data!(animal_sheep);
+        let data_ptr = data!(animal_cow);
 
         // Note: i8 ptr cast
         __VERIFIER_expect_fail(*(data_ptr as *mut i8) != 9, "Wrong data"); // From Cow
 
-        let vtable_ptr = vtable!(animal_sheep);
+        let vtable_ptr = vtable!(animal_cow);
 
         // Drop pointer
         __VERIFIER_expect_fail(
--- a/src/tools/compiletest/src/runtest.rs
+++ b/src/tools/compiletest/src/runtest.rs
@@ -2406,16 +2406,60 @@ impl<'test> TestCx<'test> {
             .args(&self.props.cbmc_flags);
         self.add_rmc_dir_to_path(&mut rmc);
         let proc_res = self.compose_and_run_compiler(rmc, None);
-        // Print an error if the verification result is not expected.
-        if self.should_compile_successfully(self.pass_mode()) {
-            if !proc_res.status.success() {
-                self.fatal_proc_rec("test failed: expected success, got failure", &proc_res);
+        // If the test file contains expected failures in some locations, ensure
+        // that verification does indeed fail in those locations
+        let fail_lines = TestCx::find_expect_fail_lines(&self.testpaths.file);
+        if !fail_lines.is_empty() {
+            if let Some(num) = TestCx::verify_expect_fail_in_lines(&proc_res.stdout, fail_lines) {
+                self.fatal_proc_rec(
+                    &format!("test failed: expected failure in line {}, got success", num),
+                    &proc_res,
+                )
             }
         } else {
-            if proc_res.status.success() {
-                self.fatal_proc_rec("test failed: expected failure, got success", &proc_res);
+            // Print an error if the verification result is not expected.
+            if self.should_compile_successfully(self.pass_mode()) {
+                if !proc_res.status.success() {
+                    self.fatal_proc_rec("test failed: expected success, got failure", &proc_res);
+                }
+            } else {
+                if proc_res.status.success() {
+                    self.fatal_proc_rec("test failed: expected failure, got success", &proc_res);
+                }
             }
         }
+    }
+
+    /// Verify that `str` contains expected failures corresponding to each line in `lines`.
+    fn verify_expect_fail_in_lines(str: &str, line_nums: Vec<usize>) -> Option<usize> {
+        let mut lines = str.lines();
+        let first_fail = &format!("line {} EXPECTED FAIL", line_nums[0]);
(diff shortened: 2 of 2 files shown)


src/test/cbmc/DynTrait/dyn_fn_param_closure_capture_fail.rs
```diff
@@ -13,7 +13,7 @@ fn takes_dyn_fun(fun: &dyn Fn() -> i32) {
     let x = fun();
     __VERIFIER_expect_fail(x != 5, "Wrong return");
     /* The closure captures `a` and thus is sized */
-    __VERIFIER_expect_fail(size_from_vtable(vtable!(fun)) == 8, "Wrong size");
+    __VERIFIER_expect_fail(size_from_vtable(vtable!(fun)) != 8, "Wrong size");
```
avanhatt, line 16: Yes, thank you for fixing this typo! Also changed this in one of my in progress PRs.

src/test/cbmc/DynTrait/vtable_size_align_drop_fail.rs
```diff
@@ -90,15 +91,15 @@ fn main() {
     }
     // Check layout/values for Cow
     unsafe {
-        let animal_sheep = random_animal(1);
```
avanhatt, line 93: Oops, thanks!

src/tools/compiletest/src/runtest.rs
```diff
@@ -2406,16 +2406,60 @@ impl<'test> TestCx<'test> {
...
+    fn verify_expect_fail_in_lines(str: &str, line_nums: Vec<usize>) -> Option<usize> {
+        let mut lines = str.lines();
+        let first_fail = &format!("line {} EXPECTED FAIL", line_nums[0]);
+        let mut line = lines.find(|&line| line.contains(first_fail));
+
+        for line_num in line_nums {
+            // It is safe to call unwrap here because RMC always prints an
+            // "EXPECTED FAIL" line for each call to `__VERIFIER_expect_fail`.
+            if line.unwrap().contains("SUCCESS") {
+                return Some(line_num);
```
avanhatt, line 2443: Should this keep going instead of returning, to check all lines?

src/tools/compiletest/src/runtest.rs
```diff
@@ -2406,16 +2406,60 @@ impl<'test> TestCx<'test> {
...
+            }
+            // "EXPECTED FAIL" lines are printed in sequence, so we can
+            // immediately check the next line.
+            line = lines.next()
```
adpaco-aws, line 2447: Worried about this. Say you have expected-fail assertions in two functions from different files - they will not appear next to each other, right?

src/tools/compiletest/src/runtest.rs
adpaco-aws, line 2443: As @avanhatt said, you may accumulate `line_num` here and return all of them at the end.

src/test/cbmc/DynTrait/vtable_size_align_drop_fail.rs
```diff
@@ -13,6 +13,7 @@
 #![allow(deprecated)]
 
 use std::intrinsics::size_of;
+use std::ptr::drop_in_place;
```
adpaco-aws, line 16: Why do you need this?

adpaco-aws (changes requested):
As discussed earlier, ensuring that assertions expected to fail are failing and not saying anything about the rest is totally fine.

src/tools/compiletest/src/runtest.rs
bdalrhm: Yes. You're right. I did not consider that. What should be the plan here? Should I look for expected failures in other files too? That complicates things quite a bit... I could also check for the file name first...

adpaco-aws: Why not use the the `lines.find` call that you use for finding the first element?

bdalrhm: It does not solve this problem. You can have two `__VERIFIER_expect_fail` on different files, but the same line.

adpaco-aws: One option is to add more lines to be checked from other files after getting the output. But you can open an issue and deal with that later. Adding a check for the file name first should be fine for this PR.

adpaco-aws: There's another way to go about this: You can reverse the procedure to inspect files looking for expected-fail assertions within the files based on the output. I would check anyway that the main file contains expected-fail assertions (in case the text changes in the future), but then the problem is reduced to building the files-lines relation.

src/tools/compiletest/src/runtest.rs
bdalrhm: Done! To some degree...

bdalrhm: Done! To some degree...

src/test/cbmc/DynTrait/vtable_size_align_drop_fail.rs
bdalrhm: `drop_in_place` is used in the test and the test will not compile without it. I copied this use statement from the nonfailing version of the test.

src/tools/compiletest/src/runtest.rs
bdalrhm: Done! I added a naïve check to ensure that at least few tests fail if the output changes in the future. I opened an issue to come up with a better check #286. Building a file-line relation complicates the check. I did not implement it because it can be manually extracted from RMC's output and there is no immediate use. I opened an issue to track this as well #285.

src/test/cbmc/DynTrait/dyn_fn_param_closure_capture_fail.rs
adpaco-aws: Undo this since the change is already merged.

src/tools/compiletest/src/runtest.rs
```diff
@@ -2406,16 +2406,52 @@ impl<'test> TestCx<'test> {
...
+    }
+
+    /// If the test file contains expected failures in some locations, ensure
+    /// that verification does not succeed in those locations.
+    fn verify_expect_fail(str: &str) -> Vec<usize> {
+        let re = Regex::new(r"line [0-9]+ EXPECTED FAIL: SUCCESS").unwrap();
+        let mut lines = vec![];
+        for m in re.find_iter(str) {
+            lines
+                .push(m.as_str().split_ascii_whitespace().skip(1).next().unwrap().parse().unwrap());
```
adpaco-aws, line 2452: I would use intermediate values here.
```rust
let number_str = m.as_str().split_ascii_whitespace().skip(1).next().unwrap();
let number = number_str.parse().unwrap();
lines.push(number);
```

adpaco-aws (changes requested):
Has to be rebased against current branch.

src/tools/compiletest/src/runtest.rs
bdalrhm: Done!

adpaco-aws (approved):

PR #304, opened 2021-07-07
Custom panic hook for RMC including current item

### LINKED ISSUE #287: RMC should have its own panic hook
Currently, when RMC panics, we panic with the Rustc hook that includes something like this:
```rust
error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports
```

We can use `panic::set_hook` to set our own custom hook that links to this repo instead.

avanhatt:
### Description of changes: 

When RMC panics currently, you get the standard Rustc panic message, including a link to their bug reporting. Add a custom panic hook with our bug template link and include the current codegen item (usually function) causing the problem. 

Example panic now:

```rust
   Compiling libc v0.2.93
thread 'rustc' panicked at 'oops', compiler/rustc_codegen_llvm/src/gotoc/mod.rs:208:9
stack backtrace:
   0: std::panicking::begin_panic
             at /home/ubuntu/rmc/library/std/src/panicking.rs:541:12
…

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/issues/new?labels=C-bug%2C+I-ICE%2C+T-compiler&template=ice.md

note: rustc 1.55.0-dev running on x86_64-unknown-linux-gnu

note: compiler flags: -Z force-unstable-if-unmarked -Z trim-diagnostic-paths=no -Z codegen-backend=gotoc -C embed-bitcode=no -C debuginfo=2 --crate-type lib

note: some of the compiler flags provided by cargo are hidden

query stack during panic:
end of query stack

[RMC] current codegen item: codegen_static: DefId(0:8806 ~ core[e967]::fmt::num::DEC_DIGITS_LUT)

RMC unexpectedly panicked during code generation.

If you are seeing this message, please file an issue here instead of on the Rust compiler: https://github.com/model-checking/rmc/issues/new?labels=bug&template=bug_report.md
error: could not compile `core`

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

### Resolved issues:

Resolves #287

### Call-outs:

Unfortunately, we can't use the same approach Cranelift uses to remove rustc's bug report link, since that would introduce a circular dependency on `rustc_driver`. Adding a comment explaining that.

### Testing:

* How is this change tested?

Manually introducing panics. 

* Is this a refactor change?

No.

### Checklist

### DIFF
--- a/compiler/rustc_codegen_llvm/src/gotoc/mod.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/mod.rs
@@ -44,18 +44,6 @@ mod utils;
 // codegen item in the panic trace.
 thread_local!(static CURRENT_CODEGEN_ITEM: RefCell<Option<String>> = RefCell::new(None));
 
-// Updates the tracking global variable for panic debugging.
-fn set_panic_debug_codegen_item(s: String) {
-    CURRENT_CODEGEN_ITEM.with(|odb_cell| {
-        odb_cell.replace(Some(s));
-    });
-}
-fn clear_panic_debug_codegen_item() {
-    CURRENT_CODEGEN_ITEM.with(|odb_cell| {
-        odb_cell.replace(None);
-    });
-}
-
 // Include RMC's bug reporting URL in our panics.
 const BUG_REPORT_URL: &str =
     "https://github.com/model-checking/rmc/issues/new?labels=bug&template=bug_report.md";
@@ -102,6 +90,19 @@ static DEFAULT_HOOK: SyncLazy<Box<dyn Fn(&panic::PanicInfo<'_>) + Sync + Send + 
 pub struct GotocCodegenBackend();
 
 impl<'tcx> GotocCtx<'tcx> {
+    // Calls the closure while updating the tracked global variable marking the
+    // codegen item for panic debugging.
+    pub fn call_with_panic_debug_info<F: FnOnce(&mut GotocCtx<'tcx>) -> ()>(
+        &mut self,
+        call: F,
+        panic_debug: String,
+    ) {
+        CURRENT_CODEGEN_ITEM.with(|odb_cell| {
+            odb_cell.replace(Some(panic_debug));
+            call(self);
+            odb_cell.replace(None);
+        });
+    }
     pub fn codegen_block(&mut self, bb: BasicBlock, bbd: &BasicBlockData<'tcx>) {
         self.current_fn_mut().set_current_bb(bb);
         let label: String = self.current_fn().find_label(&bb);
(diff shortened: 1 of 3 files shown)


compiler/rustc_codegen_llvm/src/gotoc/mod.rs
```diff
@@ -36,6 +39,60 @@ pub mod stubs;
 mod typ;
 mod utils;
 
+// Use a thread-local global variable to track the current codegen item for debugging.
+// If RMC panics during codegen, we can grab this item to include the problematic
+// codegen item in the panic trace.
+thread_local!(static CURRENT_CODEGEN_ITEM: RefCell<Option<String>> = RefCell::new(None));
+
+// Updates the tracking global variable for panic debugging.
+fn set_debug_item(s: String) {
```
danielsn, line 48: Is there a more descriptive name to use here?

compiler/rustc_codegen_llvm/src/gotoc/mod.rs
```diff
@@ -294,8 +354,17 @@ impl CodegenBackend for GotocCodegenBackend {
...
-                    MonoItem::Static(def_id) => c.declare_static(def_id, item),
+                    MonoItem::Fn(instance) => {
+                        set_debug_item(format!(
+                            "declare_function: {}",
+                            c.readable_instance_name(instance)
+                        ));
+                        c.declare_function(instance)
+                    }
+                    MonoItem::Static(def_id) => {
+                        set_debug_item(format!("declare_static: {:?}", def_id));
```
danielsn, line 365: We should reset the item at the end, to avoid spuriously having a set item if there is none.  We could use a RAII pattern, tho that might be overkill

avanhatt: Done, reset after both loops.

avanhatt: Not sure what exactly RAII would look like here, feel free to follow up if this change is not sufficient?

compiler/rustc_codegen_llvm/src/gotoc/mod.rs
avanhatt: Changed, thanks!

compiler/rustc_codegen_llvm/src/gotoc/mod.rs
```diff
@@ -305,18 +379,29 @@ impl CodegenBackend for GotocCodegenBackend {
                 }
             }
         }
+        clear_panic_debug_codegen_item();
 
         // then we move on to codegen
         for cgu in codegen_units {
             let items = cgu.items_in_deterministic_order(tcx);
             for (item, _) in items {
                 match item {
```
danielsn, line 388: Moving/adding the clear as the first item in the loop would make it clear that we're doing clear when needed :)

danielsn (changes requested):

compiler/rustc_codegen_llvm/src/gotoc/mod.rs
avanhatt: Changed to implement as a closure that sets/clears the value as appropriate!

danielsn (approved):

PR #317, opened 2021-07-08
Runnable gen-c output

### LINKED ISSUE #237: Output of gen-c is not valid C code
Running RMC with the `--gen-c` flag produces a C-like program output, but the outputted file is not valid C code.

Invalid parts:
- `src/test/cbmc/FunctionCall_Ret-Param/main.rs`
  - `main` can't return `struct Unit`, must return `int`
  - Tuple name (e.g. `struct (u32, bool)`) invalid
  - For tuples, `.0`, `.1`, etc. are invalid field names; replace with `._1`?
  - `overflow` is not defined for addition, etc.
  - `memcpy` and `memmove` are macros, and so their headers cause compile errors
  - `nondet_<k>()` is not defined
- `src/test/cbmc/FunctionCall_Ret-Param/main.rs` + print at end
  - Array name (e.g. `struct [&str]` or `struct [&str; 2]`) invalid
  - Qualified type name (e.g. `std::fmt::ArgumentV1`) invalid
  - Dynamic types (e.g. `dyn std::fmt::Write::vtable`) invalid
  - Parameterized types (e.g. `std::option::Option<usize>`) invalid
  - Union types (e.g. `Option-union`) invalid
  - `byte_extract_little_endian` is not defined
- `src/test/cbmc/FatPointers/boxtrait.c`
  - `_RINvNtCsj5Zaw82J3PU_5alloc5alloc8box_freeDNtCs21hi0yVfW1J_8boxtrait5TraitEL_NtB2_6GlobalEBG_(var_1._0, var_1._1);`
    - Use `->` instead of `.`
  - `OBJECT_SIZE` is not defined
  - `dyn_Trait__vtable` has wrong field names `.__Trait__get` -> `.get`, `.__Trait__increment` -> `.increment`

vecchiot-aws:
### Description of changes: 

A PR introducing a `--gen-c-runnable` flag which produces a runnable C program from the compiled goto program. The existing `--gen-c` flag acts as a starting point, but there are a few issues listed in #237 that prevent the generated C file from compiling.

The change introduces a generic visitor pattern for transforming `Stmt`s, `Expr`s, and `Type`s, which is then used to create a transformer over symbol tables. We then create an implementation for transforming identifiers in the symbol table to valid C identifiers.

To signify whether we want the original symbol table or the transformed names one, we pass a different backend (`"gotoc" vs "genc"`) to the `rmc-rustc` command. We then do some post-processing to add in some CBMC-specific special functions/macros.

### Resolved issues:

Partially resolves #237 (possibly fully after more changes.)

### Call-outs:

Todo:

- [ ] Generate `nondet_<k>()` functions from a sample trace.
- [ ] Add definition for `overflow`
- [ ] Add definition for `byte_extract_little_endian`
- [ ] Fix field access in `src/test/cbmc/FatPointers/boxtrait.c`
- [x] Set up distinct flags for readable and runnable gen-c
- [ ] Figure out testing strategy
- [ ] Documentation
- [ ] Clean up commits

### Testing:

### Checklist


compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/gen_c_transformer.rs
```diff
@@ -0,0 +1,182 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use super::{
+    CIntType, DatatypeComponent, Expr, Location, Parameter, Stmt, StmtBody, Symbol, SymbolTable,
+    SymbolValues, Transformer, Type,
+};
+use std::collections::{BTreeMap, HashMap, HashSet};
+
```
vecchiot-aws, line 9: Add a check that injective

vecchiot-aws:
Closing in favor of #334

PR #320, opened 2021-07-09
Fix codegen for dynamic `Fn::call` and `FnMut::call_mut`

avanhatt:
### Description of changes: 

This is the first component of addressing https://github.com/model-checking/rmc/issues/240. 

In Rust's MIR generation, shim functions are generated for calls to `Fn::call` and `FnMut::call_mut`that include calls to the wrapped function, for example, conceptually:

```rust
// self is a dyn Fn
fn call(&self) {
    call_actual_function();
}
```

Previously, the Gotoc code we were generating was call the first (self) argument instead of inserting an actual call to the wrapped function. In the case of `Fn::call`, self is a zero-sized type, that is, a pointer to an empty allocation (see [this zulip post](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/Empty.20alloc.20as.20data.20of.20dyn.20function.20fat.20pointer.3F) for more). Trying to call that as a pointer is an invalid function call, and produces Gotoc that is both incorrect and fails when passing the `--pointer-check` flag.

```c
unsigned int call(unsigned int (*var_1)(void))
{

bb0:
  ;
  unsigned int var_0=var_1();

bb1:
  ;
  return var_0;
}
```

Instead, we should generate this:

```c
unsigned int call(unsigned int (*var_1)(void))
{

bb0:
  ;
  unsigned int var_0=call_actual_function();

bb1:
  ;
  return var_0;
}
```

The fix here is to edit a special case we already had for `FnDef` locals to also include pointers to function definitions. 

### Resolved issues:

Resolves https://github.com/model-checking/rmc/issues/319

### Call-outs:

- We still fail on some `Box` cases, keeping https://github.com/model-checking/rmc/issues/240 open for that.

### Testing:

* How is this change tested?

4 new tests, edits to existing tests that were `fixme`. 

* Is this a refactor change?

No.

### Checklist


compiler/rustc_codegen_llvm/src/gotoc/place.rs
```diff
@@ -199,22 +199,40 @@ impl<'tcx> GotocCtx<'tcx> {
...
     }
 
-    /// codegen for a local
-    pub fn codegen_local(&mut self, l: Local) -> Expr {
+    // This is nasty. A local might be a function definition or a pointer to
+    // one. For example, the auto-generated code for Fn::call_once
+    // uses a local FnDef to call the wrapped function, while the auto-generated
+    // code for Fn::call and Fn::call_mut both use pointers to a FnDef.
+    // In these cases, we need to generate an expression that references the
+    // existing fndef rather than a named variable.
```
adpaco-aws, line 207: Nice!

src/test/cbmc/DynTrait/dyn_fn_once.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Check that we can pass a dyn FnOnce pointer to a stand alone
+// function definition
```
adpaco-aws, line 5: Newline.

src/test/cbmc/DynTrait/dyn_fn_once_fail.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Check that we can pass a dyn FnOnce pointer to a stand alone
+// function definition. Fail case checks wrong int.
```
adpaco-aws, line 5:
```suggestion
// function definition. Fails when comparing with a non-expected value.
```
Or something similar.

compiler/rustc_codegen_llvm/src/gotoc/place.rs
```diff
@@ -199,22 +199,40 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    pub fn codegen_local_fndef(&mut self, l: Local) -> Option<Expr> {
```
adpaco-aws, line 208: Can you add a `///` comment here?

src/test/cbmc/DynTrait/dyn_fn_mut_fail.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Check that we can pass a dyn FnMut pointer to a stand alone
+// function definition. Fail checks the wrong int.
```
adpaco-aws, line 5:
```suggestion
// function definition. Fails when comparing with a non-expected value.
```
Or something similar.

adpaco-aws (commented):
LGTM - Just a few comments.

adpaco-aws (approved):

Issue #324, opened 2021-07-13
Pre-process extracted examples before running RMC on them.


bdalrhm:
Before we run RMC on the extracted examples from the Rust books, we need to do the following pre-processing steps:

- [x] Extract compiler annotations in code blocks and translate them to appropriate RMC flags.
- [x] #407
- [x] #408

adpaco-aws:
Closing this issue because the tools to do this are already there.

We only need to study the examples and include assertions on a case-by-case basis. Opened #517 for that.

PR #325, opened 2021-07-13
Extract examples from reference and display their results in a dashboard.

### LINKED ISSUE #293: Automatically extract examples from the Rust Reference.

bdalrhm:
### Description of changes: 

This PR introduces a new tool, `dashboard`, that extracts examples from The Rust Reference, run them through RMC, and displays their results in a terminal dashboard.

### Resolved issues:

Resolves #293 and #294

### Call-outs:

This PR omits code for two tasks, which will be handled in different PRs. The first task is to pre-process extracted examples from The Rust Reference, before running them through RMC (#323). The second task is to integrate the step for building the dashboard tool with RMC's build steps (#324).

### Testing:

* How is this change tested?
  * Running `scripts/display-dashboard` displays the terminal dashboard.
* Is this a refactor change?
  * No.

### Checklist


src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,287 @@
...
+//! and display their results.
+
+use pulldown_cmark::{Parser, Tag};
+use std::io::{BufRead, BufReader};
+use std::process::Stdio;
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    process::Command,
+};
```
adpaco-aws, line 13: Why are you doing this for some `std` imports and not others?

src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,287 @@
...
+use crate::dashboard;
+
+/// Parses the chapter/section hierarchy in the markdown file specified by
+/// `summary_path` and returns a mapping from markdown files containing rust
+/// code to corresponding directories where the extracted rust code should
+/// reside.
+fn parse_hierarchy(summary_path: &Path) -> HashMap<PathBuf, PathBuf> {
+    let summary = fs::read_to_string(summary_path).unwrap();
+    // Skip the title and introduction.
+    let parser = Parser::new(&summary).skip(8);
```
adpaco-aws, line 25: Can we make give a name to `8` and add a check here? I'm thinking of a scenario where they update the reference and this hard-coded number consumes too little/much.

src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,287 @@
...
+                .unwrap(),
+            "--persist-doctests",
+            gen_dir.to_str().unwrap(),
+            "--no-run",
+        ]);
+        cmd.stdout(Stdio::null());
+        cmd.spawn()
+            .expect(&format!("Error: failed to spawn process:\n{:?}", cmd))
+            .wait()
+            .expect(&format!("Error: spawned process did not run:\n{:?}", cmd));
```
adpaco-aws, line 80: When can this happen?

src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,287 @@
...
+
+/// Copies the extracted rust code in `from_dir` to `src/test` following the
+/// hierarchy specified by `rel_map`.
+fn organize_examples(rel_map: &HashMap<PathBuf, PathBuf>, book_dir: &Path, from_dir: &Path) {
+    let to_dir: PathBuf = ["src", "test"].iter().collect();
+    // The names of the extracted examples generated by `rustdoc` have the
+    // format `<path>_<line-num>_<test-num>` where occurrences of '/', '-', and
+    // '.' in <path> are replaced by '_'. This translation is not injective, so
+    // we cannot map those names to the original markdown file path. Instead, we
+    // create a new map that does the same translation.
```
adpaco-aws, line 92: 🤩

src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,287 @@
...
+        &to
+    ));
+    fs::copy(&from, &to).expect(&format!("Error: failed to copy {:?} to {:?}", &from, &to));
+}
+
+/// Pre-processes the tests in the specified `paths` before running them with
+/// `compiletest`.
+fn preprocess_examples(_paths: Vec<&PathBuf>) {
+    // For now, we will only pre-process the tests that cause infinite loops.
+    // TODO: properly implement this step.
```
adpaco-aws, line 136: Reference the issue here.

src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,287 @@
...
+        .expect(&format!("Error: failed to spawn process:\n{:?}", cmd))
+        .wait()
+        .expect(&format!("Error: spawned process did not run:\n{:?}", cmd));
+}
+
+/// Creates a new [`Tree`] from a path, `ns`, and a test result, `p`.
+fn tree_from_path(mut ns: Vec<String>, p: bool) -> dashboard::Tree {
+    assert!(ns.len() > 0, "Error: `ns` must contain at least 1 element.");
+    let mut tree = dashboard::Tree::new(
+        dashboard::Node::new(ns.pop().unwrap(), if p { 1 } else { 0 }, if p { 0 } else { 1 }),
```
adpaco-aws, line 206: What is `ns`?

adpaco-aws (changes requested):
This looks nice! I'm quite convinced that `tools` is a good place to put this code.
Can we have some sort of test to ensure this keeps working as intended?

bdalrhm:
Output of running the dashboard program:
```
~/rmc$ scripts/display-dashboard.sh
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `src/tools/dashboard/target/debug/dashboard`
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
    Finished release [optimized + debuginfo] target(s) in 0.12s
warning: unused import: `CodegenResults`
 --> compiler/rustc_codegen_ssa/src/traits/backend.rs:4:13
  |
4 | use crate::{CodegenResults, ModuleCodegen};
  |             ^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: 1 warning emitted

warning: function is never used: `time_trace_profiler_finish`
   --> compiler/rustc_codegen_llvm/src/llvm_util.rs:151:8
    |
151 | pub fn time_trace_profiler_finish(file_name: &str) {
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |

warning: 1 warning emitted

warning: associated function is never used: `process_rlink`
   --> compiler/rustc_driver/src/lib.rs:596:8
    |
596 |     fn process_rlink(sess: &Session, compiler: &interface::Compiler) -> Result<(), ErrorReported> {
    |        ^^^^^^^^^^^^^
    |

warning: 1 warning emitted

    Finished release [optimized + debuginfo] target(s) in 0.18s
    Finished release [optimized + debuginfo] target(s) in 0.12s
    Finished release [optimized + debuginfo] target(s) in 0.12s
Some tests failed in compiletest suite=ref mode=rmc host=x86_64-unknown-linux-gnu target=x86_64-unknown-linux-gnu
# of tests: 322 ✔️ 264  ❌ 58
ref ✔️ 264 ❌ 58
  Appendices ❌ 2
    Macro Follow-Set Ambiguity Formal Specification ❌ 1
    Glossary ❌ 1
  Attributes ✔️ 22 ❌ 7
    Diagnostics ✔️ 10 ❌ 3
    Code generation ✔️ 5
    Derive ✔️ 2
    Limits ❌ 3
    Testing ✔️ 3
    Type System ✔️ 1
  Application Binary Interface ✔️ 3
  Crates and source files ✔️ 1 ❌ 1
  Conditional compilation ✔️ 2
  Items ✔️ 73 ❌ 24
    Associated Items ✔️ 11
    Enumerations ✔️ 5 ❌ 3
    Constant items ✔️ 4
    External blocks ✔️ 3
    Functions ✔️ 12 ❌ 4
    Generic parameters ✔️ 3 ❌ 6
    Implementations ✔️ 5 ❌ 2
    Modules ✔️ 1
    Static items ✔️ 2
    Traits ✔️ 10 ❌ 5
    Structs ✔️ 4
    Type aliases ✔️ 1 ❌ 1
    Unions ✔️ 7 ❌ 1
    Use declarations ✔️ 5 ❌ 2
  Introduction ✔️ 1
  Lexical structure ✔️ 7 ❌ 3
    Keywords ❌ 1
    Comments ✔️ 1
    Tokens ✔️ 6 ❌ 2
  Linkage ✔️ 1 ❌ 1
  Macros ✔️ 8 ❌ 3
    Macros By Example ✔️ 8 ❌ 2
  Memory model ✔️ 1
    Variables ✔️ 1
  Names ✔️ 19
    Namespaces ✔️ 1
    Paths ✔️ 13
    Visibility and privacy ✔️ 4
    Preludes ✔️ 1
  Patterns ✔️ 20 ❌ 2
  Statements and expressions ✔️ 57 ❌ 8
    Expressions ✔️ 54 ❌ 8
      Block expressions ✔️ 3 ❌ 2
      Array and index expressions ✔️ 1 ❌ 1
      Call expressions ✔️ 2
      Field access expressions ✔️ 2
      Grouped expressions ✔️ 1 ❌ 1
      Literal expressions ✔️ 1
      If and if let expressions ✔️ 4
      Loop expressions ✔️ 7 ❌ 1
      Match expressions ✔️ 4
      Method call expressions ✔️ 1 ❌ 1
      Operator expressions ✔️ 13 ❌ 2
      Range expressions ✔️ 3
      Path expressions ✔️ 1
      Return expressions ✔️ 1
      Struct expressions ✔️ 6
      Tuple and index expressions ✔️ 1
      Closure expressions ✔️ 1
    Statements ✔️ 3
  The Rust runtime ✔️ 1
  Type system ✔️ 48 ❌ 7
    Destructors ✔️ 7 ❌ 1
    Lifetime elision ✔️ 5 ❌ 3
    Subtyping and Variance ✔️ 3
    Trait and lifetime bounds ✔️ 5
    Type coercions ✔️ 6
    Type layout ✔️ 7
    Types ✔️ 15 ❌ 3
      Boolean type ✔️ 1
      Closure types ✔️ 3 ❌ 1
      Function item types ✔️ 1 ❌ 1
      Impl trait type ✔️ 3
      Function pointer types ✔️ 1
      Type parameters ✔️ 1
      Array types ✔️ 1
      Slice types ✔️ 1
      Trait object types ❌ 1
      Inferred type ✔️ 1
```

src/tools/dashboard/src/dashboard.rs
```diff
@@ -0,0 +1,91 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! Data-structures representing the dashboard and and their utilities.
+
+use std::fmt::{Display, Write};
+
+/// This data-structure holds test results and info for each directory and test
```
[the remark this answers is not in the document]
nchong-at-aws: Why can't we use a standard data structure for this task?

src/tools/dashboard/src/reference.rs
nchong-at-aws: Agreed, this is fine for now, but brittle with hardcoding

src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,287 @@
...
+        _ => false,
+    };
+    let mut ns: Vec<_> = splits[1].split(&['/', '.'][..]).map(String::from).collect();
+    // Remove unnecessary `.rs` suffix.
+    ns.pop();
+    (ns, l)
+}
+
+/// Display the dashboard in the terminal.
+fn display_dashboard(dashboard: dashboard::Tree) {
```
nchong-at-aws, line 255: If our preprocessing allows us to inject assertions into the tests then is there a way we can distinguish between:
- codegen failure
- codegen success + verification fail
- codegen success + verification success
?

nchong-at-aws (commented):
This looks great, Abdalrhman! So my understanding of the code is that we parse the TOC from the Rust reference to form a map from md-file to a destination directory; populates `src/test`; preprocesses tests to add RMC command-line options etc.; runs compiletest and builds a dashboard.

src/tools/dashboard/src/reference.rs
vecchiot-aws: I also like to write the actual expected string and use `.len()` to get the value rather than writing the number itself.

src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,287 @@
...
+
+/// Copy the file specified by `from` to the corresponding location specified by
+/// `map`.
+fn copy(from: &Path, map: &HashMap<String, PathBuf>) {
+    // First, get the name of the directory where `rustdoc` saved the extracted
+    // code.
+    let file_name = from.parent().unwrap().file_name().unwrap().to_str().unwrap();
+    // Remove `_<line-num>_<test-num>` suffix from the name to lookup the name
+    // in the map.
+    let i = file_name.find("_md").unwrap() + 3;
```
vecchiot-aws, line 119: Is this `3` the length of `"_md"`?

src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,287 @@
...
+        if let Some(example) = dir.read_dir().unwrap().next() {
+            let example = example.unwrap().path();
+            copy(&example, &abs_map);
+        }
+    }
+}
+
+/// Copy the file specified by `from` to the corresponding location specified by
+/// `map`.
+fn copy(from: &Path, map: &HashMap<String, PathBuf>) {
```
vecchiot-aws, line 113: Could you include a comment with the full pattern of what `from` looks like? I'm having trouble piecing together these `find`s.

src/tools/dashboard/src/reference.rs
vecchiot-aws, line 119: The comment says "Remove ... suffix", but then this is used as the starting bound of the range; wouldn't `j` denote the suffix?

src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,287 @@
...
+fn parse_log(path: &Path) -> dashboard::Tree {
+    let file = fs::File::open(path).expect(&format!("Error: failed to open log file: {:?}", path));
+    let reader = BufReader::new(file);
+    let mut tests = dashboard::Tree {
+        data: dashboard::Node { name: String::from("ref"), num_pass: 0, num_fail: 0 },
+        children: vec![],
+    };
+    for line in reader.lines() {
+        let (ns, l) = parse_log_line(&line.unwrap());
+        tests = dashboard::Tree::merge(tests, tree_from_path(ns, l)).unwrap();
```
vecchiot-aws, line 233: Addressed above (when do you ever return `None`), but if you are internally using options, you should probably be handling them with better errors here.

src/tools/dashboard/src/dashboard.rs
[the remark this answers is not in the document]
bdalrhm: The standard library does not expose an API for a tree data structure (as far as I know) and other data structures do not allow us to display test results in an organized manner. Added this explanation as a comment.

src/tools/dashboard/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+[package]
+name = "dashboard"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+pulldown-cmark = { version = "0.8.0", default-features = false }
+either = "1.6.1"
```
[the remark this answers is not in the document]
bdalrhm: Not anymore. Removed!

src/tools/dashboard/src/dashboard.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub children: Vec<Tree>,
+}
+
+impl Tree {
+    /// Creates a new [`Tree`] representing a dashboard or a part of it.
+    pub fn new(data: Node, children: Vec<Tree>) -> Tree {
+        Tree { data, children }
+    }
+
+    /// Merges two trees, if they have equal roots, and returns the merged tree.
```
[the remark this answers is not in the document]
bdalrhm: Right. I refactored this function and I forgot to re-add the checks.

src/tools/dashboard/src/dashboard.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn merge(mut l: Tree, r: Tree) -> Option<Tree> {
```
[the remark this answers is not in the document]
bdalrhm: Yes... done!

src/tools/dashboard/src/dashboard.rs
```diff
@@ -0,0 +1,91 @@
...
+        // For each subtree of `r`...
+        for cnr in r.children {
+            // Look for a subtree of `l` with an equal root.
+            let index = l.children.iter().position(|cnl| cnl.data.name == cnr.data.name);
+            if let Some(index) = index {
+                // If you find one, merge it with `r`'s subtree.
+                let cnl = l.children.remove(index);
+                l.children.insert(index, Tree::merge(cnl, cnr).unwrap());
```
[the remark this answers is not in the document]
bdalrhm: Done!

src/tools/dashboard/src/dashboard.rs
```diff
@@ -0,0 +1,91 @@
...
+        f.write_fmt(format_args!("{:1$}", "", p))?;
+        f.write_str(&self.data.name)?;
+        let num = self.data.num_pass;
+        if num > 0 {
+            f.write_fmt(format_args!(" ✔️ {}", num))?;
+        }
+        let num = self.data.num_fail;
+        if num > 0 {
+            f.write_fmt(format_args!(" ❌ {}", num))?;
+        }
```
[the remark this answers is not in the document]
bdalrhm: Remnants of old code. Removed!

src/tools/dashboard/src/reference.rs
bdalrhm: I trusted `rustfmt` and let it take care of that. Cleaned up.

src/tools/dashboard/src/reference.rs
bdalrhm: @vecchiot-aws great idea!

src/tools/dashboard/src/reference.rs
bdalrhm: I don't know. I just followed the examples. But the error is too verbose and probably printed by default. Removed.

src/tools/dashboard/src/reference.rs
bdalrhm: > Reference the issue here.

Done!

> Agreed, this is fine for now, but brittle with hardcoding

Don't worry. We will come back to it... maybe... 🙃

src/tools/dashboard/src/reference.rs
bdalrhm: Renamed `ns` to `path`

src/tools/dashboard/src/reference.rs
bdalrhm: Yes. We can add a new step that runs `rmc-rustc` on the extracted code (after pre-processing?) and save the result in a new field in the `Node` struct. I wonder if I should do that through `compiletest`...

src/tools/dashboard/src/reference.rs
bdalrhm: Yes. This code is deleted now.

src/tools/dashboard/src/reference.rs
bdalrhm: Done! I rewrote the whole function too. Take another look and let me know if it makes sense to you now.

src/tools/dashboard/src/reference.rs
bdalrhm: Please take a look at the new code.

src/tools/dashboard/src/reference.rs
bdalrhm: That's not necessary. All generated test trees have `ref` (the new test suite) as their root.

src/tools/dashboard/src/reference.rs
```diff
@@ -0,0 +1,280 @@
...
+    // `src/tools/dashboard/target/ref/<key>_<line-num>_<test-num>/rust_out`
+    // We copy the file in this path to a new path of the form:
+    // `src/test/<val>/<line-num>.rs
+    // where `map[<key>] == <val>`. We omit <test-num> because all tests have
+    // the same number, 0.
+    // Extract `<key>_<line-num>_<test-num>`.
+    let key_line_test = from.parent().unwrap().file_name().unwrap().to_str().unwrap();
+    // Extract <key> and <line-num> from `key_line_test` to get <val> and
+    // construct destination path.
+    let _2 = key_line_test.rfind("_").unwrap();
```
nchong-at-aws, line 123: Could you use https://doc.rust-lang.org/std/string/struct.String.html#method.split ?

bdalrhm: Good idea! `<key>` contains 5+ underscores. So it's tricky to use this version. But I think, `rsplitn` will do the trick.

nchong-at-aws (approved):
Looks good to go

adpaco-aws (approved):

Issue #240, opened 2021-06-21
Incorrect verification results for function/closure dyn traits


avanhatt:
The following fails to verify:

```rust
fn foo(fun: &mut dyn FnMut(i32) -> i32) -> i32 {
    fun(1)
}

fn main() {
    let r = foo(&mut move |z : i32| { z } );
    assert!(r == 1); // should succeed
}
```

In the generated CBMC-C code, in some places we pass the closure argument as a tuple, but in others we expect the flattened int.

avanhatt:
This is blocking https://github.com/model-checking/rmc/issues/220

avanhatt:
A similar example (without an explicit `FnOnce`) that demonstrates that this is a soundness issue:

```rust
fn takes_dyn_fun(fun: &dyn Fn() -> u32) {
    let x = fun();
}

pub fn unit_to_u32() -> u32 {
    assert!(false); 
    0
}

fn main() {
    takes_dyn_fun(&unit_to_u32)
}
```

Should fail from this explicit assert false, but:
```
VERIFICATION SUCCESSFUL
```

avanhatt:
This does fail as expected with `--pointer-check`:

```
$ rmc --gen-c src/test/cbmc/DynTrait/dyn_fn_minimal.rs --pointer-check | grep FAIL
[pointer_dereference.7] invalid function pointer: FAILURE
VERIFICATION FAILED
```

avanhatt:
Weird thing to maybe investigate separately:

The following hangs `RMC` at the `gotoc-cc` call (?!?)
```rust
    // Create a nested boxed once-callable closure
    let f: Box<Box<dyn FnOnce(i32)>> = Box::new(Box::new(|x| assert!(x == 1)));
    f(1);

    // Create a pointer to a closure
    let g = |y| assert!(y == 2);
    let p: &dyn Fn(i32) = &g;
    p(2);

    // Additional level of pointer nesting
    let q: &dyn Fn(i32) = &p;
    q(2);
```

Issue #307, opened 2021-07-07
Tracking issue: Tests with default checks disabled


adpaco-aws:
In #263 we had to disable default checks in some tests to keep them working the same way. The affected tests are:
- [ ] `src/test/cbmc/CodegenConstValue/main.rs`

adpaco-aws:
#335 removed the `--no-memory-safety-checks` flag in `DynTrait/boxed_closure.rs`

avanhatt:
https://github.com/model-checking/rmc/pull/373 resolves `DynTrait/dyn_fn_param.rs`

celinval:
#647 fixed the following tests:

- `src/test/firecracker/virtio-block-parse/main.rs`
- `src/test/rmc/PointerOffset/Stable/main.rs`
- `src/test/rmc/PointerOffset/Unstable/main.rs`
- `src/test/rmc/Strings/main.rs`
- `src/test/rmc/SwitchInt/main.rs`
- `src/test/rmc/Whitespace/main.rs`

PR #335, opened 2021-07-19
Correct function type for vtable shim (fix some boxed closures)

### LINKED ISSUE #367: Correct function type for vtable shim
Breaking off from https://github.com/model-checking/rmc/issues/240 to have distinct issues.

The problem is that we fail verification for calls through `FnOnce` vtable closures:
```
let f: Box<dyn FnOnce()> = Box::new(|x| {
    assert(x == 2); // should fail
});
f(1);
```

avanhatt:
### Description of changes: 

RMC currently fails to verify boxed dynamic `FnOnce` closures, like this:

```rust
let f: Box<dyn FnOnce()> = Box::new(|x| {
    assert(x == 2); // should fail
});
f(1);
```

The issue is somewhat specific to `FnOnce::call_once`, which takes a `self` pointer instead of a `self *` pointer at the source language level. Vtable calls expect a pointer as the first argument, so Rust itself repairs vtable shims of this type to take an argument:

```rust
    /// `<T as Trait>::method` where `method` receives unsizeable `self: Self` (part of the
    /// `unsized_locals` feature).
    ///
    /// The generated shim will take `Self` via `*mut Self` - conceptually this is `&owned Self` -
    /// and dereference the argument to call the original function.
    VtableShim(DefId),

...
if let ty::InstanceDef::VtableShim(..) = self.def {
    // Modify `fn(self, ...)` to `fn(self: *mut Self, ...)`.
    sig = sig.map_bound(|mut sig| {
        let mut inputs_and_output = sig.inputs_and_output.to_vec();
        inputs_and_output[0] = tcx.mk_mut_ptr(inputs_and_output[0]);
        sig.inputs_and_output = tcx.intern_type_list(&inputs_and_output);
        sig
    });
}
```

We need to do the same fix.

### Resolved issues: 

Resolves #367

### Call-outs:

Known remaining issues: https://github.com/model-checking/rmc/issues/240

Broader concerns are tracked in: https://github.com/model-checking/rmc/issues/299.

### Testing:

* How is this change tested?

- Existing unit test (fixes 2 tests that previously required disabling `--pointer-check`).
- New `_fixme` test to be fixed in #240 

* Is this a refactor change?

No.

### Checklist


compiler/rustc_codegen_llvm/src/gotoc/place.rs
```diff
@@ -208,26 +208,26 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                _ => None,
-            },
+            // A local can be pointer to a FnDef, like Fn::call and Fn::call_mut
+            ty::RawPtr(inner) => self.codegen_local_fndef(inner.ty).map(|f| f.address_of()),
+            // A local can be a boxed function pointer
+            ty::Adt(def, _) if def.is_box() => {
+                let boxed_ty = self.codegen_ty(ty);
+                self.codegen_local_fndef(ty.boxed_ty())
+                    .map(|f| self.box_value(f.address_of(), boxed_ty))
+            }
```
adpaco-aws, line 222: Can you update the function-level comment regarding this case?

src/test/cbmc/DynTrait/boxed_closure.rs
```diff
@@ -1,13 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//rmc-flags: --no-memory-safety-checks
-
-// We use `--no-memory-safety-checks` in this test to avoid getting
-// a verification failure:
-// [pointer_dereference.7] invalid function pointer: FAILURE
-// Tracking issue: https://github.com/model-checking/rmc/issues/307
```
adpaco-aws, line 9: Let's not forget to mark this one as "done" in #307 once this PR is merged.

avanhatt:
Found a new test that still fails, converting to draft while I work on the fix.

compiler/rustc_codegen_llvm/src/gotoc/place.rs
```diff
@@ -208,26 +208,26 @@ impl<'tcx> GotocCtx<'tcx> {
     /// code for Fn::call and Fn::call_mut both use pointers to a FnDef.
     /// In these cases, we need to generate an expression that references the
     /// existing fndef rather than a named variable.
-    pub fn codegen_local_fndef(&mut self, l: Local) -> Option<Expr> {
-        let t = self.local_ty(l);
-        match t.kind() {
+    pub fn codegen_local_fndef(&mut self, ty: &'tcx ty::TyS<'tcx>) -> Option<Expr> {
```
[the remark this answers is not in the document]
avanhatt: Added `_fixme` test; but actual fix to support it is involved and will be split off into a different PR.

compiler/rustc_codegen_llvm/src/gotoc/place.rs
avanhatt: Done!

danielsn (approved):

Issue #66, opened 2021-04-14
Incorrect call to`drop_in_place` with dynamic argument


adpaco-aws:
Expected: assertion failure due to not-implemented

PR #336, opened 2021-07-20
Codegen drop for non-dynamic objects

### LINKED ISSUE #184: Codegen failure with vector extend example
The piece of code:
```
fn main() {
    let mut v: Vec<u32> = Vec::new();
    for (start, len) in vec![(0,1), (1,2)] {
        v.extend(start..=(start + len - 1));
    }
```
is failing to codegen. This is part of the balloon code and was working before.

chinmaydd:
### Description of changes: 

RMC currently does not codegen a call to object destructors (`drop_in_place`). This creates a soundness issue such as the one described in #184. This change codegens a call to `drop_in_place` for statically typed objects.

### Resolved issues:

Resolves #184 

### Call-outs:

This PR still does not codegen drop for dynamic objects, which is related to #66. 

### Testing:

* How is this change tested?
TBD

* Is this a refactor change?
No

### Checklist

### DIFF
--- a/compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs
@@ -8,6 +8,7 @@ use super::{DatatypeComponent, Location, Parameter, Stmt, SwitchCase, SymbolTabl
 use num::bigint::BigInt;
 use std::collections::BTreeMap;
 use std::fmt::Debug;
+use tracing::debug;
 
 ///////////////////////////////////////////////////////////////////////////////////////////////
 /// Datatypes
@@ -635,12 +636,13 @@ impl Expr {
             components
         );
 
+        // NOTE: Disabling this check for now.
         // Check that each formal field has an value
-        for field in non_padding_fields {
-            let field_typ = field.field_typ().unwrap();
-            let value = components.get(field.name()).unwrap();
-            assert_eq!(value.typ(), field_typ);
-        }
+        // for field in non_padding_fields {
+        //     let field_typ = field.field_typ().unwrap();
+        //     let value = components.get(field.name()).unwrap();
+        //     assert_eq!(value.typ(), field_typ);
+        // }
 
         let values = fields
             .iter()
@@ -648,7 +650,14 @@ impl Expr {
                 if field.is_padding() {
                     field.typ().nondet()
                 } else {
-                    components.remove(field.name()).unwrap()
+                    // components.remove(field.name()).unwrap()
+                    let comp = components.remove(field.name()).unwrap();
+                    if comp.typ() != &field.typ() {
+                        debug!("struct_expr: Force casting {:?} to {:?}", comp.typ(), field.typ());
+                        comp.cast_to(field.typ())
+                    } else {
+                        comp
+                    }
                 }
             })
             .collect();
--- a/compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/rvalue.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/rvalue.rs
@@ -134,7 +134,32 @@ impl<'tcx> GotocCtx<'tcx> {
         let metadata = if self.use_slice_fat_pointer(place_mir_type) {
             intermediate_fat_pointer.member("len", &self.symbol_table)
         } else if self.use_vtable_fat_pointer(place_mir_type) {
-            intermediate_fat_pointer.member("vtable", &self.symbol_table)
+            if self.is_unsized(place_mir_type) {
+                let vtable_name = match place_mir_type.kind() {
+                    ty::Dynamic(..) => self.vtable_name(place_mir_type),
+                    _ => {
+                        let (_, trait_type) = self
+                            .nested_pair_of_concrete_and_trait_types(place_mir_type, place_mir_type)
+                            .unwrap();
+                        let trait_vtable_type = self.codegen_trait_vtable_type(trait_type);
+                        let mir_codegen_ty = self.codegen_ty(place_mir_type);
+                        let place_mir_mangled_name = self.ty_mangled_name(place_mir_type);
+                        let mir_ref_type_name = format!("&{}", &place_mir_mangled_name);
+                        self.ensure_struct(&mir_ref_type_name, |_, _| {
+                            vec![
+                                Type::datatype_component("data", mir_codegen_ty.to_pointer()),
+                                Type::datatype_component("vtable", trait_vtable_type.to_pointer()),
+                            ]
+                        });
+                        format!("{}", &self.vtable_name(trait_type))
+                    }
+                };
+                intermediate_fat_pointer
+                    .member("vtable", &self.symbol_table)
+                    .cast_to(Type::struct_tag(&vtable_name).to_pointer())
+            } else {
+                intermediate_fat_pointer.member("vtable", &self.symbol_table)
+            }
         } else {
             unreachable!()
         };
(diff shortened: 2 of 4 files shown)


compiler/rustc_codegen_llvm/src/gotoc/statement.rs
```diff
@@ -106,22 +106,35 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                InstanceDef::DropGlue(..) => Stmt::skip(Location::none()),
-                _ => {
-                    let func = self.codegen_func_expr(instance, None);
-                    func.call(vec![self.codegen_place(location).goto_expr.address_of()])
-                        .as_stmt(Location::none())
+            // Extract nested object if type is pointed to by a fat pointer
+            let nested_ty = match loc_ty.kind() {
+                ty::Ref(_, pointee_type, ..) => pointee_type,
+                ty::Adt(adt_def, adt_subst) => {
+                    if adt_def.is_box() || self.vtable_field_name(adt_def.did).contains("cell") {
```
danielsn, line 113: Is there a helper function somewhere that already does this? What does the llvm code do here?

compiler/rustc_codegen_llvm/src/gotoc/statement.rs
```diff
@@ -106,22 +106,35 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        adt_subst.first().unwrap().expect_ty()
+                    } else {
+                        loc_ty
+                    }
                 }
+                _ => loc_ty,
             };
+            let drop_implementation = match instance.def {
```
danielsn, line 121: I don't say this often, but this might be easier to read as a set of `if let ... else if let ... else`

compiler/rustc_codegen_llvm/src/gotoc/statement.rs
chinmaydd: The LLVM code does not handle boxes and cells specially. I am rethinking this approach.

compiler/rustc_codegen_llvm/src/gotoc/statement.rs
chinmaydd: Agreed!

chinmaydd:
New regression failures due to codegenning `drop_in_place` for all allocations in the program.

```
failures:
    [rmc] cbmc/FatPointers/boxmuttrait.rs
    [rmc] cbmc/FatPointers/boxmuttrait_fail.rs
    [rmc] cbmc/SizeAndAlignOfDst/main.rs
    [rmc] cbmc/SizeAndAlignOfDst/main_fail.rs
    [rmc] cbmc/Slice/pathbuf.rs
```

Working on fixing them. They are a combination of failures due to codegen typecheck issues and verification failures due to memory soundness issues such as double free.

compiler/rustc_codegen_llvm/src/gotoc/statement.rs
```diff
@@ -106,22 +106,18 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                    func.call(vec![self.codegen_place(location).goto_expr.address_of()])
-                        .as_stmt(Location::none())
-                }
-            };
+            let drop_implementation;
+            // For empty DropGlue function invocation, we ignore it
+            if let InstanceDef::DropGlue(_, None) = instance.def {
+                drop_implementation = Stmt::skip(Location::none());
+            } else if let ty::Dynamic(..) = loc_ty.kind() {
+                drop_implementation = Stmt::skip(Location::none());
```
danielsn, line 114: Why is this none

chinmaydd: This PR only aims to fix codegen for non-dynamic objects. Issue to track codegen for dynamic object is #66 .

compiler/rustc_codegen_llvm/src/gotoc/typ.rs
```diff
@@ -1038,9 +1051,14 @@ pub fn pointee_type(pointer_type: Ty<'tcx>) -> Option<Ty<'tcx>> {
 impl<'tcx> GotocCtx<'tcx> {
     /// A pointer to the mir type should be a thin pointer.
     pub fn use_thin_pointer(&self, mir_type: Ty<'tcx>) -> bool {
-        // ptr_metadata_ty is not defined on all types, the projection of an associated type
-        return !self.is_unsized(mir_type)
-            || mir_type.ptr_metadata_ty(self.tcx) == self.tcx.types.unit;
+        // If type is an UnsafeCell treat it similar to a thin pointer
```
danielsn, line 1054: Do we think we have all the cases here?

compiler/rustc_codegen_llvm/src/gotoc/typ.rs
```diff
@@ -113,6 +113,19 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
+    /// Check if a type is an UnsafeCell
+    pub fn is_ty_unsafecell(&self, t: &'tcx TyS<'_>) -> bool {
+        // Refer: https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/.E2.9C.94.20Identifying.20whether.20type.20is.20UnsafeCell
+        // Also refer: visit_aggregate in compiler/rustc_mir/src/interpret/intern.rs
+        if let Some(def) = t.ty_adt_def() {
+            if Some(def.did) == self.tcx.lang_items().unsafe_cell_type() {
```
danielsn, line 121: could just return this instead of `if this return true`

compiler/rustc_codegen_llvm/src/gotoc/statement.rs
danielsn: Should this be `codegen_unimplemented`

compiler/rustc_codegen_llvm/src/gotoc/statement.rs
```diff
@@ -106,22 +106,18 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                    )
-                    .as_stmt(Location::none()),
-                InstanceDef::DropGlue(..) => Stmt::skip(Location::none()),
-                _ => {
-                    let func = self.codegen_func_expr(instance, None);
-                    func.call(vec![self.codegen_place(location).goto_expr.address_of()])
-                        .as_stmt(Location::none())
-                }
-            };
+            let drop_implementation;
```
danielsn, line 109: just assign this to the result of the next section

compiler/rustc_codegen_llvm/src/gotoc/statement.rs
```diff
@@ -106,22 +106,18 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            // For empty DropGlue function invocation, we ignore it
+            if let InstanceDef::DropGlue(_, None) = instance.def {
+                drop_implementation = Stmt::skip(Location::none());
+            } else if let ty::Dynamic(..) = loc_ty.kind() {
+                drop_implementation = Stmt::skip(Location::none());
+            } else {
```
danielsn, line 115: get rid of explicit catch all, be explicit about what types we handle, and then panic on the rest

chinmaydd:
Closing this PR with some comments which could help address the issue:

For this PR, we currently enable the call which fails to typecheck in codegen_drop. The major bug here is in `rvalue_ref` generation for types which are unsized due to having unsized types as members. The ref for these types should be a fat pointer which points to the vtable of the underlying unsized type. The data field contains a pointer to the mir data for which the rvalue ref is being generated.

We are currently disabling the typecheck when generating a struct_expr to aggressively push through the codegen. For that, we force cast the expressions to match the required types.

This code currently fails at:
```
func: Expr { value: Symbol { identifier: "_RINvNtCs75dudyTCRYy_4core3ptr13drop_in_placeNtNtNtCsedW8enyxb1b_3std10sys_common5mutex12MovableMutexECshatX5H9cBZr_10main_fixme" }, typ: Code { parameters: [Parameter { typ: Pointer { typ: StructTag("tag-std::sys_common::mutex::MovableMutex::4176847340024427407") }, identifier: None, base_name: None }], return_type: StructTag("tag-Unit") }, location: None }
args: [Expr { value: Symbol { identifier: "_RINvNtCs75dudyTCRYy_4core3ptr13drop_in_placeINtNtNtCsedW8enyxb1b_3std4sync5mutex5MutexDNtCshatX5H9cBZr_10main_fixme10SubscriberEL_EEB1p_::1::var_1" }, typ: StructTag("tag-dyn Subscriber::10926333765724947806"), location: None }]', compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs:552:9
```

PR #353, opened 2021-07-21
Differentiate between check, codegen, and verification errors.

### LINKED ISSUE #338: `_fail` tests should fail if compilation fails
The test at `src/test/cbmc/DynTrait/main_fail.rs` uses `_VERIFIER_expect_fail` instead of `__VERIFIER_expect_fail`, causing it to have a rust compile error. 

I expect it went by unnoticed because we (accidentally?) left in the `_fail` suffix, when really it shouldn't fail, and we don't distinguish between compilation failure and verification failure.

bdalrhm:
### Description of changes: 

Currently, RMC can panic for many reasons including syntax errors, name resolution errors, type checking errors, codegen errors, and verification errors. RMC developers, however, are only interested in codegen and verification errors. If RMC panics for other reasons, it's usually because of oversight. The testing framework should notify the developer of those errors in the regression suite. Otherwise, regressions that are expected to panic may silently pass because of them. This PR achieves that by dividing errors into three categories: check errors (all errors that are caught by `rustc` before the codgen step), codegen errors, and verification errors (all errors that occur after the codgen step). The test procedure then ensures that RMC panics at the stage specified by the developer.

### Resolved issues:

Resolves #338 

### Testing:

* How is this change tested?
  * Each error category is used by at least 1 test.
* Is this a refactor change?
  * Yes. compiletest does not look for `fail` in the test name anymore. So, tests that fail to check/codegen/verify now include a `// rmc-*-fail` comment to tell compiletest that they should fail. The majority of those tests were also renamed (removing the `_fail` suffix).

### Checklist


src/tools/compiletest/src/common.rs
```diff
@@ -118,6 +118,14 @@ impl fmt::Display for PassMode {
     }
 }
 
+/// Stage where an RMC test should fail.
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd)]
+pub enum RMCFailMode {
+    Check,
+    Codegen,
+    Verify,
+}
```
adpaco-aws, line 127: Can you add a comment explaining what is tested for each one of the? I had trouble understanding the `Check` one so I think we should be clear about that.

src/tools/compiletest/src/header.rs
```diff
@@ -121,6 +123,8 @@ pub struct TestProps {
     pass_mode: Option<PassMode>,
     // Ignore `--pass` overrides from the command line for this test.
     ignore_pass: bool,
+    // How far this RMC test should proceed to start failing.
```
adpaco-aws, line 126:
```suggestion
    // At which stage is the RMC test expected to fail.
```

src/tools/compiletest/src/header.rs
```diff
@@ -383,6 +389,33 @@ impl TestProps {
...
+            Some(RMCFailMode::Check)
+        } else if config.parse_name_directive(ln, "rmc-codegen-fail") {
+            check_rmc("codegen");
+            Some(RMCFailMode::Codegen)
+        } else if config.parse_name_directive(ln, "rmc-verify-fail") {
+            check_rmc("verify");
+            Some(RMCFailMode::Verify)
+        } else {
+            None
+        };
```
adpaco-aws, line 411: Refactor into a function.

src/tools/compiletest/src/runtest.rs
```diff
@@ -2392,20 +2393,69 @@ impl<'test> TestCx<'test> {
...
             // Otherwise, insert PATH as a new enviornment variable and set its value to the RMC scripts directory.
-            rmc.env(String::from("PATH"), String::from(self.config.rmc_dir_path.to_str().unwrap()));
+            command.env(
+                String::from("PATH"),
+                String::from(self.config.rmc_dir_path.to_str().unwrap()),
+            );
+        }
+    }
+
+    /// Runs `rustc` on the test file specified by `self.testpaths.file`. An
```
adpaco-aws, line 2410:
```suggestion
    /// Runs `rmc-rustc` on the test file specified by `self.testpaths.file`. An
```

src/tools/compiletest/src/runtest.rs
```diff
@@ -2392,20 +2393,69 @@ impl<'test> TestCx<'test> {
...
+                self.fatal_proc_rec("test failed: expected check failure, got success", &proc_res);
+            }
+        } else {
+            if !proc_res.status.success() {
+                self.fatal_proc_rec("test failed: expected check success, got failure", &proc_res);
+            }
+        }
+    }
+
+    /// Runs `rustc` on the test file specified by `self.testpaths.file`. An
```
adpaco-aws, line 2428: Same. Let's distinguish `rustc` and `rmc-rustc`.

adpaco-aws (changes requested):

src/tools/compiletest/src/header.rs
bdalrhm: Done!

src/tools/dashboard/src/main.rs
```diff
@@ -4,6 +4,19 @@
 mod dashboard;
```
adpaco-aws, line 4: Is this supposed to be here?

src/tools/compiletest/src/common.rs
```diff
@@ -118,6 +118,21 @@ impl fmt::Display for PassMode {
     }
 }
 
+/// Step at which RMC test should fail.
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd)]
+pub enum RMCFailStep {
+    /// RMC panics before the codegen step. This step runs the same checks on
```
adpaco-aws, line 124: Maybe this will make it clearer?
```suggestion
    /// RMC panics before the codegen step (up to MIR generation). [...]
```

src/tools/compiletest/src/common.rs
```diff
@@ -118,6 +118,21 @@ impl fmt::Display for PassMode {
...
+    /// the test code as `cargo check` including syntax, type, name resolution,
+    /// and barrow checks.
```
adpaco-aws, line 126: Chuckled at "barrow" 😃

adpaco-aws (changes requested):

src/tools/dashboard/src/main.rs
bdalrhm: No. I was just testing the intrinsic functions out and wanted help from the rust analyzer. Removed.

adpaco-aws (approved):

PR #355, opened 2021-07-23
Add Cargo.toml `cargo rmc` flag parsing.

### LINKED ISSUE #349: Use yaml to specify `cargo rmc` settings.
Rather than manually passing flags into `cargo rmc`, we want to allow users to create a yaml that contains all of the desired settings. This can also act as a place to specify ignore paths and such.

vecchiot-aws:
### Description of changes: 

We introduce the ability to configure `cargo rmc` with your `Cargo.toml` file. 

Example `Cargo.toml`:
```
[package]
name = "test-cargo-rmc"
version = "0.1.0"
edition = "2018"

[dependencies]

[rmc.flags.loudness]
debug = false
verbose = false
quiet = false

[rmc.flags.harness]
c-lib = []
function = "main"

[rmc.flags.artifacts]
target = "."
keep-temps = false
gen-c = false
gen-symbols = false

# ...
```

Nested tags under `rmc.flags` are flattened, and have no effect, but can be used for organizational purposes. 

How to use:
- `flag = "string"` translates to `--flag "string"`
- `flag = true` translates to `--flag`
- `flag = false` does nothing
- `flag = ["str1", "str2"]` translates to `--flag "str1" "str2"`

### Resolved issues:

Resolves #349 

### Testing:

### Checklist


scripts/cargo-rmc
```diff
@@ -19,6 +20,13 @@ def main():
     if len(sys.argv) >= 2 and sys.argv[1] == "rmc":
         del sys.argv[1]
 
+    # When no flags provided, look up in Cargo.toml
+    if len(sys.argv) <= 2:
+        dir = sys.argv[1] if len(sys.argv) == 2 else "."
+        no_failure = try_load_toml(dir, sys.argv)
+        if not no_failure:
+            return 1
```
vecchiot-aws, line 28: Switch to default Cargo.toml, manual flags override.

scripts/cargo-rmc
```diff
@@ -19,6 +20,13 @@ def main():
...
+
     parser = argparse.ArgumentParser(description="Verify a Rust crate")
```
vecchiot-aws, line 30: Add ability to specify toml file.

scripts/cargo-rmc
```diff
@@ -94,6 +63,134 @@ def main():
...
+            rmc.ensure(pathlib.Path(config_toml).is_file(), f"Invalid config file: {config_toml}")
+        else:
+            crate_cargo_toml = pathlib.Path(crate).joinpath("Cargo.toml")
+            if crate_cargo_toml.is_file():
+                config_toml = crate_cargo_toml
+            else:
+                return cl_args
+        
+        # Extract flags from config toml
+        toml_args = load_toml(config_toml)
```
danielsn, line 120: Does this just load the toml? Or does it also extract the RMC specific args?

scripts/cargo-rmc
```diff
@@ -94,6 +63,134 @@ def main():
...
+        # --quiet overrides --verbose
+        if args.quiet:
+            args.verbose = False
+
+        # Add some CBMC flags by default unless `--no-default-checks` is being used
+        if args.default_checks:
+            rmc.add_selected_default_cbmc_flags(args)
+
+    # Combine args set by config toml and provided on command line
+    def get_combined_args(config_toml, crate, cl_args):
```
danielsn, line 108: The logic of this function was hard to follow

scripts/cargo-rmc
```diff
@@ -94,6 +63,134 @@ def main():
...
+    # Parse command line args to find crate and check for config flags
+    parser = create_parser()
+    cl_args = parser.parse_args()
+    validate(cl_args)
+
+    # Load flags from config toml
+    if cl_args.no_config_toml:
+        final_args = cl_args
+    else:
+        crate = cl_args.crate or cl_args.crate_flag or "."
```
danielsn, line 143: don't you have this same bit of code somewhere else?

scripts/rmc.py
```diff
@@ -43,16 +44,19 @@ def edit_output(self, text):
...
-
-def dependencies_in_path():
+    
+def ensure_dependencies_in_path():
     for program in [RMC_RUSTC_EXE, "symtab2gb", "cbmc", "cbmc-viewer", "goto-instrument", "goto-cc"]:
-        if not is_exe(program):
-            print("ERROR: Could not find {} in PATH".format(program))
-            return False
-    return True
+        ensure(is_exe(program), f"Could not find {program} in PATH")
```
danielsn, line 50: Do we want to check minimum versions

scripts/rmc.py
```diff
@@ -72,11 +76,11 @@ def add_set_cbmc_flags(args, flags):
 
 # Add sets of selected default CBMC flags
 def add_selected_default_cbmc_flags(args):
-    if not args.no_memory_safety_checks:
+    if args.memory_safety_checks:
```
danielsn, line 79: nice

danielsn (changes requested):

scripts/cargo-rmc
```diff
@@ -94,6 +63,134 @@ def main():
...
+                find_flags(map[key])
+            else:
+                flags[key] = map[key]
+    find_flags(flag_data)
+
+    # Add flags to args
+    success = True
+    args = []
+    def add_flag(flag, value):
+        if type(value) == bool:
```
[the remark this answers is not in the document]
vecchiot-aws: What do you mean by this? The `type(value) == bool` part?

scripts/rmc.py
vecchiot-aws: #362

scripts/cargo-rmc
vecchiot-aws: I've tried to rearrange the comments a bit, but I don't know how to decompose the function much more. I could break it into helpers but I'm worried that will just create a flood of helper functions.

scripts/cargo-rmc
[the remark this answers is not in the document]
danielsn: I mean the `--no` flag handling. Is there a standard python way of handling that (e.g. an argument to the parser), or do you need to do that manually?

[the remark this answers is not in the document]
vecchiot-aws: The `--no` flag handling is done by a particular class of `argparser`, and it's pretty rudimentary: https://github.com/python/cpython/blob/cc602a71dfe3b3af615e78b2ca4f213d98448d9e/Lib/argparse.py#L858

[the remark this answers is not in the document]
vecchiot-aws: It essentially recognizes when the flag passed in starts with `--no`, and sets the value to false instead of true in that case. Since I'm constructing the arg list (rather than directly working with the parser object), I can't think of a way of avoiding manual string manipulation here.

scripts/cargo-rmc
danielsn: Proposed logic:
```python
if no_config_toml:
    return cl_args
elif config_toml is not None:
    # If config_toml is explicitly given, ensure it exists
    rmc.ensure(pathlib.Path(config_toml).is_file(), f"Invalid path to config file: {config_toml}")
else:
    crate_cargo_toml = pathlib.Path(crate).joinpath("Cargo.toml")
        rmc.ensure(pathlib.Path(config_toml).is_file(), f"Cannot find configuration toml at expected location: {config_toml}")
```

scripts/cargo-rmc
```diff
@@ -83,6 +63,143 @@ def main():
...
+        # Add some CBMC flags by default unless `--no-default-checks` is being used
+        if args.default_checks:
+            rmc.add_selected_default_cbmc_flags(args)
+
+    # Combine args set by config toml and provided on command line
+    def get_combined_args(config_toml, crate, cl_args):
+        # Set config_toml to valid toml file
+        if config_toml is not None:
+            # If config_toml is explicitly given, ensure it exists
+            rmc.ensure(pathlib.Path(config_toml).is_file(), f"Invalid config file: {config_toml}")
```
danielsn, line 116: its not the file that's invalid, its the path to it

scripts/cargo-rmc
danielsn: One other thought: make this function just take a toml, and have it do the combining. then have a seperate logic to determine what cargo toml to use.

scripts/cargo-rmc
```diff
@@ -89,6 +64,152 @@ def main():
...
+        config_group.add_argument("--config-toml", help="Where to read configuration from; defaults to crate's Cargo.toml")
+        config_group.add_argument("--no-config-toml", action="store_true", help="Do not use any configuration toml")
+
+        exclude_flags = []
+        rmc_flags.add_flags(parser, {"default-target": "target"}, exclude_flags=exclude_flags)
+        
+        return parser
+
+    # Determine what crate to analyze from flags;
+    # args should be validated before calling this function to ensure
```
danielsn, line 86: Or this function itself could assert that

scripts/cargo-rmc
```diff
@@ -89,6 +64,152 @@ def main():
...
+            # If config_toml is explicitly given, ensure it exists
+            rmc.ensure(pathlib.Path(config_toml).is_file(), f"Invalid path to config file: {config_toml}")
+            return config_toml
+        else:
+            # If no config flags are set, look for config toml in default location (<crate>/Cargo.toml)
+            cargo_toml = pathlib.Path(crate).joinpath("Cargo.toml")
+            if cargo_toml.is_file():
+                return cargo_toml
+            else:
+                print(f"WARNING: Cannot find configuration toml at expected location: {cargo_toml}")
```
danielsn, line 107: do we want to warn or error here?

scripts/cargo-rmc
```diff
@@ -89,6 +64,152 @@ def main():
...
+    flags = dict()
+    def find_flags(map):
+        for key in map:
+            if type(map[key]) == dict:
+                find_flags(map[key])
+            else:
+                flags[key] = map[key]
+    find_flags(flag_data)
+
+    # Add flags to args
```
danielsn, line 186: so flags are what `cargo rmc` takes, and args are what we're passing to`rmc`?

scripts/rmc.py
```diff
@@ -51,9 +51,10 @@ def ensure_dependencies_in_path():
         ensure(is_exe(program), f"Could not find {program} in PATH")
 
 # Assert a condition holds, or produce a user error message.
-def ensure(condition, message, retcode=1):
+def ensure(condition, message=None, retcode=1):
```
danielsn, line 54: when do we not want to have a message

scripts/setup/macos-10.15/install_deps.sh
```diff
@@ -7,3 +7,10 @@ set -eux
 # Update tools in macOS 10.15 via `brew`
 brew update
 brew install ctags
+
+# Add Python package dependencies
+PYTHON_DEPS=(
+  toml # Used for parsing `cargo-rmc` config toml
+)
+
+python3 -m pip install "${PYTHON_DEPS[@]}"
```
danielsn, line 16: newling

scripts/setup/ubuntu-20.04/install_deps.sh
```diff
@@ -32,3 +32,10 @@ set -x
 
 sudo apt-get --yes update
 sudo DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends --yes "${DEPS[@]}"
+
+# Add Python package dependencies
+PYTHON_DEPS=(
+  toml # Used for parsing `cargo-rmc` config toml
+)
+
+python3 -m pip install "${PYTHON_DEPS[@]}"
```
danielsn, line 41: newline

scripts/rmc.py
vecchiot-aws: In `cargo-rmc`, we manually print out possibly multiple error messages, so we don't need the `ensure` itself to print a message.

vecchiot-aws: https://github.com/model-checking/rmc/pull/355/files#diff-876efdc1d5780b01eb535f6e409bbfe28bea1b5876d28456f3f7b2e03b7e6decR211

scripts/cargo-rmc
vecchiot-aws: I've generally had in my head that `flags` was the command line input, while `args` is the result of parsing this, though sometimes the distinction gets mixed up. I've fixed up a few places where this was backwards.

scripts/rmc_flags.py
```diff
@@ -4,15 +4,57 @@
 
 import argparse
 
+# Taken from https://github.com/python/cpython/blob/3.9/Lib/argparse.py#L858
+# Cannot use `BooleanOptionalAction` with Python 3.8
```
adpaco-aws, line 8: Is there an issue with upgrading to Python 3.9? I know there has been some chat about Python versions lately.

adpaco-aws (commented):
Can we a `cargo-rmc` test for this feature?

scripts/rmc_flags.py
vecchiot-aws: Only to reduce the burden on RMC users.

src/test/cargo-rmc/simple-extern/test_sum.ignore
```diff
@@ -1 +0,0 @@
-[call_into_rust.assertion.1] line 13 assertion takes_int(x) == x + 1: SUCCESS
```
adpaco-aws, line 1: Why do we need this change?

src/test/cargo-rmc/simple-extern/test_sum.expected
```diff
@@ -0,0 +1 @@
+[external_c_assertion.assertion.1] line 13 assertion rust_add1(x) == x + 1: SUCCESS
```
adpaco-aws, line 1: Why do we need this change?

adpaco-aws (commented):
Need some help to understand why you changes the `simple-extern` expected/ignore files.

src/test/cargo-rmc/simple-extern/test_sum.expected
vecchiot-aws: Now that we can configure `cargo rmc` with the toml file, we are able to set the `c-lib` flag and actually make this test work, so I moved it from `.ignore` to `.expected`.

adpaco-aws (approved):

scripts/cargo-rmc
vecchiot-aws: Done.

scripts/cargo-rmc
vecchiot-aws: Done.

scripts/cargo-rmc
vecchiot-aws: The logic is now to use `get_config_toml` to find the config toml file (or `None`), and then when the result is present, we use `get_combined_args` to actually parse the config toml and combine it with the command line args.

scripts/cargo-rmc
vecchiot-aws: Extracted to a helper function.

PR #373, opened 2021-08-02
Fix closure untupling/retupling

### LINKED ISSUE #240: Incorrect verification results for function/closure dyn traits

avanhatt:
### Description of changes: 

RMC currently fails to verify some nested boxed closures, such as this:
```rust
    let f: Box<Box<dyn FnOnce(f32, i32)>> = Box::new(Box::new(|x, y| { assert!(x == 1.0); assert!(y == 2) }));
    f(1.0, 2);
```

Here is an example of the code we now generate (with renaming for clarity):
```
struct Unit call_once_vtable_shim(struct [boxed_closure.rs:8:49: 11:6] *var_1, float spread_2, int spread_3)
{
  struct (f32, i32) var_2={ .0=spread_2, .1=spread_3 }; // new: retuple since the body requires it
  struct Unit var_0;

bb0:
  ;
  boxed_closure_main(*var_1, var_2.0, var_2.1);

bb1:
  ;
  return var_0;
}
```

The problem is that we were checking whether to untuple closure arguments based on whether the function's type kind was a `Closure`, but this misses some cases (where the trait points to a defined function, or with vtable shims) where the type is a `FnDef`. The right way to check for untupling is via a check to the function's ABI. However, in that case, we also need to _retuple_ arguments based on the MIR's `spread_arg` flag. See [this Zulip](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/Determine.20untupled.20closure.20args.20from.20Instance.3F) for more discussion.

The trickiness here is that we need to insert both the tuple and its individual components into the symbol table, and then write the components out to the tuple at the start of a function. To do this, I added a new naming convention for the untupled args, avoiding naming conflicts based on the local index. Local index math determines what gets retupled when necessary.

### Resolved issues:

Resolves #240

### Call-outs:

`ignore_var_ty` is mystifying, and there are some intrinsic cases where skipping `FnDef` there elides the environment argument of a call through a tuple. I don't like this, and we should probably eventually remove `ignore_var_ty` in favor of elides arguments only for specific, documented reasons (like the being Zero Sized Types that rust elides). 

### Testing:

* How is this change tested?

Existing tests, including a `fixme` added in my last PR.

* Is this a refactor change?

Somewhat.

### Checklist

### DIFF
--- a/compiler/rustc_codegen_llvm/src/gotoc/metadata.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/metadata.rs
@@ -255,13 +255,37 @@ impl<'tcx> GotocCtx<'tcx> {
         self.monomorphize(p.ty(self.current_fn().mir().local_decls(), self.tcx).ty)
     }
 
-    pub fn closure_params(&self, substs: ty::subst::SubstsRef<'tcx>) -> Vec<Ty<'tcx>> {
-        let sig = self.monomorphize(substs.as_closure().sig());
-        let args = match sig.skip_binder().inputs()[0].kind() {
-            ty::Tuple(substs) => substs.iter().map(|s| s.expect_ty()),
-            _ => unreachable!("this argument of a closure must be a tuple"),
-        };
-        args.collect()
+    /// Closures expect their last arg untupled at call site, see comment at
+    /// ty_needs_closure_untupled.
+    fn sig_with_closure_untupled(&self, sig: ty::PolyFnSig<'tcx>) -> ty::PolyFnSig<'tcx> {
+        debug!("sig_with_closure_untupled sig: {:?}", sig);
+        let fn_sig = sig.skip_binder();
+        if let Some((tupe, prev_args)) = fn_sig.inputs().split_last() {
+            let args: Vec<Ty<'tcx>> = match tupe.kind() {
+                ty::Tuple(substs) => substs.iter().map(|s| s.expect_ty()),
+                _ => unreachable!("the final argument of a closure must be a tuple"),
+            }
+            .collect();
+
+            // The leading argument should be exactly the environment
+            assert!(prev_args.len() == 1);
+            let env = prev_args[0].clone();
+
+            // Recombine arguments: environment first, then the flattened tuple elements
+            let recombined_args = iter::once(env).chain(args);
+
+            return ty::Binder::bind_with_vars(
+                self.tcx.mk_fn_sig(
+                    recombined_args,
+                    fn_sig.output(),
+                    fn_sig.c_variadic,
+                    fn_sig.unsafety,
+                    fn_sig.abi,
+                ),
+                sig.bound_vars(),
+            );
+        }
+        sig
     }
 
     fn closure_sig(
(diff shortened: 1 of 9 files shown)


adpaco-aws:
Hi @avanhatt , can you please add the example to the description?

compiler/rustc_codegen_llvm/src/gotoc/mod.rs
```diff
@@ -156,6 +161,84 @@ impl<'tcx> GotocCtx<'tcx> {
...
 
+    /// MIR functions have a `spread_arg` field that specifies whether the
+    /// final argument to the function is "spread" at the LLVM/codegen level
+    /// from a tuple into its individual components. (Used for the "rust-
+    /// call" ABI, necessary because dynamic trait closure cannot have an
+    /// argument list in MIR that is both generic and variadic, so Rust
+    /// allows a generic tuple).
+    ///
+    /// If `spread_arg` is Some, then the wrapped value is the local that is
+    /// to be "spead"/untupled. However, the MIR function body itself expects
```
adpaco-aws, line 172:
```suggestion
    /// to be "spread"/untupled. However, the MIR function body itself expects
```

src/test/cbmc/DynTrait/dyn_fn_param.rs
```diff
@@ -1,13 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//rmc-flags: --no-memory-safety-checks
-
-// We use `--no-memory-safety-checks` in this test to avoid getting
-// a verification failure:
-// [pointer_dereference.14] invalid function pointer: FAILURE
-// Tracking issue: https://github.com/model-checking/rmc/issues/307
-
```
adpaco-aws, line 10: Nice!

compiler/rustc_codegen_llvm/src/gotoc/metadata.rs
```diff
@@ -283,30 +307,41 @@ impl<'tcx> GotocCtx<'tcx> {
...
 
     pub fn fn_sig_of_instance(&self, instance: Instance<'tcx>) -> ty::PolyFnSig<'tcx> {
         let fntyp = instance.ty(self.tcx, ty::ParamEnv::reveal_all());
         self.monomorphize(match fntyp.kind() {
             ty::Closure(def_id, subst) => self.closure_sig(*def_id, subst),
-            _ => fntyp.fn_sig(self.tcx),
+            ty::FnPtr(..) | ty::FnDef(..) => {
+                let sig = fntyp.fn_sig(self.tcx);
+                // Some virtual calls through a vtable may actually be closures
+                // that also need the arguments uptupled, even though the kind of
```
adpaco-aws, line 337: Does "uptupled" have a typo?

compiler/rustc_codegen_llvm/src/gotoc/mod.rs
```diff
@@ -146,6 +146,11 @@ impl<'tcx> GotocCtx<'tcx> {
             let loc = self.codegen_span2(&ldata.source_info.span);
             let sym =
                 Symbol::variable(name, base_name, t, self.codegen_span2(&ldata.source_info.span));
+            if self.symbol_table.contains(&sym.name) {
+                debug!("Skipping already-added local: {:?}", sym.name);
```
adpaco-aws, line 150: Why do we need this check?

compiler/rustc_codegen_llvm/src/gotoc/statement.rs
```diff
@@ -195,31 +195,50 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        //     0. a struct representing the environment
+        //     1. a tuple containing the parameters
+        //
+        // However, for some reason, Rust decides to generate a function which still
+        // takes the first argument as the environment struct, but the tuple of parameters
+        // are flattened as subsequent parameters.
+        // Therefore, we have to project out the corresponding fields when we detect
+        // an invocation of a closure.
+        //
+        // Note: in some cases, the enviroment struct has type FnDef, so we skip it in
```
adpaco-aws, line 219:
```suggestion
        // Note: In some cases, the enviroment struct has type FnDef, so we skip it in
```

compiler/rustc_codegen_llvm/src/gotoc/utils.rs
```diff
@@ -25,6 +25,14 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
+    // Special naming conventions for parameters that are spread from a tuple
+    // into its indivdual components at the LLVM level, see comment at
```
adpaco-aws, line 29:
```suggestion
    // into its individual components at the LLVM level, see comment at
```

compiler/rustc_codegen_llvm/src/gotoc/utils.rs
```diff
@@ -25,6 +25,14 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    // compiler/rustc_codegen_llvm/src/gotoc/mod.rs:codegen_function_prelude
+    pub fn codegen_spread_arg_name(&self, l: &Local) -> String {
+        let fname = self.current_fn().name();
+        format!("{}::1::spread{:?}", fname, l)
```
adpaco-aws, line 33: Do we need to hard-code this?

adpaco-aws (commented):
Great work! A few minor comments and questions.

compiler/rustc_codegen_llvm/src/gotoc/mod.rs
avanhatt: The spread arg is added in the prelude, so we shouldn't re-add it to the table here. But I added a different, more explicit check and comment instead, thanks!

compiler/rustc_codegen_llvm/src/gotoc/utils.rs
avanhatt: I think so, the name `spread` is not special, we just need to name it something different from `var` and I chose this to try and help convey the purpose.

compiler/rustc_codegen_llvm/src/gotoc/mod.rs
```diff
@@ -135,10 +135,12 @@ impl<'tcx> GotocCtx<'tcx> {
         let mir = self.current_fn().mir();
         let ldecls = mir.local_decls();
         ldecls.indices().for_each(|lc| {
-            let base_name = match self.find_debug_info(&lc) {
-                None => format!("var_{}", lc.index()),
-                Some(info) => format!("{}", info.name),
-            };
+            if Some(lc) == mir.spread_arg {
+                // We have already added this local in the functin prelude, so
```
adpaco-aws, line 139:
```suggestion
                // We have already added this local in the function prelude, so
```

adpaco-aws (approved):
LGTM! Do you mind updating manually and fixing the typo below?

Issue #372, opened 2021-07-30
Vector Module Stubbing


chinmaydd:
This issue tracks the overarching project of implementing a stub for the Vector library. The goal here is to implement a stub which supports all methods soundly. This implies that we may choose to not add an implementation for some of them which may be experimental or cannot be implemented soundly.

Reference: https://doc.rust-lang.org/std/vec/struct.Vec.html

Methods implemented:

- [x] `append`
- [x] `as_mut_ptr`
- [x] `as_mut_slice`
- [x]  `as_ptr`
- [x] `as_slice`
- [x] `capacity`
- [x] `clear`
- [x] `extend_from_slice`
- [x] `insert`
- [x] `is_empty`
- [x] `len`
- [x] `new`
- [x] `new_in`
- [x] `pop`
- [x] `push`
- [x] `remove`
- [x] `reserve`
- [x] `reserve_exact`
- [x] `resize`
- [x] `resize_with`
- [x] `set_len`
- [x] `shrink_to`
- [x] `shrink_to_fit`
- [x] `split_off`
- [x] `swap_remove`
- [x] `truncate`
- [x] `with_capacity`
- [x]  `from_raw_parts`

Traits implemented:

- [x] `AsMut<[T]>`
- [x] `AsMut<Vec<T, A>>`
- [x] `AsRef<Vec<T, A>>`
- [x] `AsRef<[T]>`
- [x] `AsMut<[T]>`
- [x] `Clone`
- [x] `Debug`
- [x] `Default`
- [x] `Deref` 
- [x] `DerefMut`
- [x] `PartialEq` 
- [x] `Drop`
- [x] `Eq`
- [x] `Extend`
- [x] `Extend<&'a T>`
- [x] `From<&[T]>`
- [x] `From<&mut [T]>`
- [x] `Index` 
- [x] `IndexMut`
- [x] `Ord` 
- [x] `PartialOrd`
- [x] `Iterator`

carolynzech:
The description of https://github.com/model-checking/kani/pull/377 indicates that this issue should have been resolved once that PR was merged.

PR #377, opened 2021-08-03
Vector stubbing

chinmaydd:
### Description of changes: 

This PR implements an initial implementation of a stub for the Vector module for RMC. We implement 3 abstraction types:

1. `rmc`- This aims to be the most complete version of the stub and closest in terms of functionality to the standard library implementation. Here, the idea is to implement all functionality in Rust using simple modeling of the Vector data structure.

2. `c-ffi` - This implements most of the memory handling of the Vector operations in C while exporting those functions as FFI interfaces to Rust. Due to failure to communicate generics, the current implementation only supports a `Vec<u32>`. The idea is to write a script which allows us to construct the abstraction implementation on demand when the size of the value type is known.

3. `no-back` - This abstraction implements a Vector without a backing store, primarily for tests which dont care about the contents of the vector.

Usage:
```bash
$ rmc --use-abs --abs-type rmc v1.rs # Uses rmc abstraction
<...>
$ rmc v1.rs # Defaults to std
```

### Resolved issues:
#372 

### Call-outs:

This PR also contains some experimental code in files such as `hashset.rs`, `rmc_stub.c`.

### Testing:

* How is this change tested?
#465 has testing support for this change.

* Is this a refactor change?
No

### Checklist


scripts/rmc
```diff
@@ -27,14 +28,17 @@ def main():
...
+    if EXIT_CODE_SUCCESS != rmc.compile_single_rust_file(args.input, json_filename, args.verbose, args.debug, args.keep_temps, args.mangler, args.dry_run, args.abs_type, []):
         return 1
 
     if EXIT_CODE_SUCCESS != rmc.symbol_table_to_gotoc(json_filename, goto_filename, args.verbose, args.keep_temps, args.dry_run):
         return 1
 
     args.c_lib.append(str(RMC_C_LIB))
 
+    if args.abs_type == "c-ffi":
+        args.c_lib.append(str(RMC_C_STUB))
```
danielsn, line 40: Can we think of names that reflect the precision/speed tradeoff

scripts/rmc_flags.py
```diff
@@ -127,6 +127,8 @@ def add_developer_flags(make_group, add_flag, config):
              help="Pass through directly to CBMC; must be the last flag")
     add_flag(group, "--mangler", default="v0", choices=["v0", "legacy"],
              help="Change what mangler is used by the Rust compiler")
+    add_flag(group, "--abs-type", default="std", choices=["std", "rmc", "c-ffi", "no-back"],
```
danielsn, line 130: See above. We can have better names here

scripts/rmc_flags.py
```diff
@@ -127,6 +127,8 @@ def add_developer_flags(make_group, add_flag, config):
...
+             help="Choose abstraction for modules of standard library if available")
```
danielsn, line 131: Right now, we're only supporting a single choice, even if they have multiple data-structures that need to be abstracted.

src/test/rmc-prelude.rs
```diff
@@ -1,6 +1,23 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+mod stub;
+
+#[cfg(abs_type = "std")]
```
danielsn, line 6: Add comments explaining what is going on here

danielsn: And have an issue tracking merging this into the cargo flow

src/test/stub/c_vec.rs
```diff
@@ -0,0 +1,98 @@
...
+extern "C" {
+    fn ffi_new() -> *mut c_vec;
+    fn ffi_push(ptr: *mut c_vec, elem: c_uint);
+    fn ffi_cap(ptr: *mut c_vec) -> c_uint;
+    fn ffi_len(ptr: *mut c_vec) -> c_uint;
+    fn ffi_with_capacity(cap: size_t) -> *mut c_vec;
+    fn ffi_pop(ptr: *mut c_vec) -> c_uint;
+    fn ffi_append(ptr1: *mut c_vec, ptr2: *mut c_vec);
+    fn ffi_insert(ptr: *mut c_vec, index: size_t, elem: c_uint);
+}
```
danielsn, line 22: We should make sure this doesn't get broken by Alexa's fix to type names

src/test/stub/hashset.c
```diff
@@ -0,0 +1,125 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#include <stdint.h>
+#include <assert.h>
+#include <stdlib.h>
+
+/*
```
danielsn, line 8: Link to issue tracking why we can't do this with unbounded arrays. There should be a diffblue issue too

src/test/stub/hashset.rs
```diff
@@ -0,0 +1,57 @@
...
+
+// WIP: HashSet abstraction. Currently only implements the idea for a HashSet<u32> 
+// as an experiment. This function uses uninterpreted functions to model a set().
+
+extern "C" {
+    fn ffi_new() -> *mut c_hashset;
+    fn ffi_insert(ptr: *mut c_hashset, value: c_uint) -> c_uint;
+    fn ffi_contains(ptr: *mut c_hashset, value: c_uint) -> c_uint;
+    fn ffi_remove(ptr: *mut c_hashset, value: c_uint) -> c_uint;
+}
```
danielsn, line 15: You're going to want unique names here

danielsn: `ffi_new` for hashset and vector will conflict at link time

src/test/stub/hashset.rs
```diff
@@ -0,0 +1,57 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+extern crate libc;
+use self::libc::{c_int, c_uint};
+
+// WIP: HashSet abstraction. Currently only implements the idea for a HashSet<u32> 
+// as an experiment. This function uses uninterpreted functions to model a set().
+
+extern "C" {
+    fn ffi_new() -> *mut c_hashset;
```
danielsn, line 11: Do you not need nomangle?

src/test/stub/len_vec.rs
```diff
@@ -0,0 +1,127 @@
...
+
+extern crate libc;
+
+use std::mem;
+use std::ptr::{write, read};
+
+// Abstraction which tracks only the length and the last element of the vector.
+pub struct Vec<T> {
+    len: usize,
+    last: Option<*mut T>,
```
danielsn, line 12: Why do we track the last element?

danielsn: Do we have an example of a proof where this helps

danielsn: What if we track the `ith` element? And the first time we try to get that element, we assume i was the requested index?

danielsn:
```
for i in 1..10 {
    v[i] = i;
}

assert!(v[3] == 3);
```

src/test/stub/len_vec.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+impl<T> Vec<T> {
+    fn drain(&mut self) {
```
danielsn, line 16: Do we have to worry that the destructors aren't called here

src/test/stub/len_vec.rs
```diff
@@ -0,0 +1,127 @@
...
+        self.len = 0;
+        self.last = None;
+    }
+
+    fn update_last(&mut self, elem: T) {
```
danielsn, line 21: This is a helper method. Put it in its own impl block

src/test/stub/len_vec.rs
danielsn: Could we check that element types have only trivial drop?

src/test/stub/len_vec.rs
```diff
@@ -0,0 +1,127 @@
...
+        Vec {
+            len: 0,
+            last: None
+        }
+    }
+
+    pub fn push(&mut self, elem: T) {
+        if self.len == 0 {
+            let elem_size = mem::size_of::<T>();
+            let ptr = unsafe { libc::malloc(elem_size) as *mut T };
```
danielsn, line 50: This doesn't ensure that the malloced thing has basic validity constraints. We need a way to talk about that. Its sound, but will almost certainly cause problems during verification.

danielsn: Proposal: have a `trait NondetItem<T>`

danielsn: Or, for now take advantage of assumptions.rs, and keep `Option<Box<T>>`, and then say `Box::new(__nondet::<T>())`

src/test/stub/len_vec.rs
```diff
@@ -0,0 +1,127 @@
...
+            self.last = Some(ptr);
+        }
+        self.len += 1;
+        unsafe {
```
danielsn, line 54: Using Option<Box> would get rid of this unsafe, I think

src/test/stub/len_vec.rs
```diff
@@ -0,0 +1,127 @@
...
+        if other.len() != 0 {
+            self.update_last(other.pop().unwrap());
+            self.len += other.len();
+        }
+    }
+
+    pub fn insert(&mut self, index: usize, elem: T) {
+        assert!(index <= self.len);
+
+        if index == self.len {
```
danielsn, line 80: Should this be index out of bounds?
I see, it does a shift.  Maybe comment that?

src/test/stub/len_vec.rs
```diff
@@ -0,0 +1,127 @@
...
+        unsafe {
+            write(self.last.unwrap().offset(0), elem);
+        }
+    }
+
+    pub fn pop(&mut self) -> Option<T> {
+        if self.len == 0 {
+            None
+        } else {
+            self.len -= 1;
```
danielsn, line 63: Pop should set the `last` to `None`

src/test/stub/rmc_vec.rs
```diff
@@ -0,0 +1,820 @@
...
+use std::slice;
+
+const DEFAULT_CAPACITY: usize = 1024;
+
+pub struct RmcUnique<T> {
+    ptr: *const T, // *const for variance
+    _marker: PhantomData<T>,
+}
+
+// impl<T> Drop for RmcUnique<T> {
```
danielsn, line 27: This is needed

src/test/stub/rmc_vec.rs
```diff
@@ -0,0 +1,820 @@
...
+
+// sized_realloc implements resizing of memory which stores the elements contained
+// in the vector.
+//
+// TODO: This method is the fundamental bottleneck when resizing the array due to
+// the expensive malloc + memcpy procedures. Its remains to be seen if this can be
+// optimized using custom CBMC primitives.
+fn sized_realloc(orig_ptr: *mut u8, orig_size: usize, new_size: usize) -> *mut u8 {
+    unsafe {
+        let result = libc::malloc(new_size) as *mut u8;
```
danielsn, line 58: We can at least use `realloc`, so if CBMC does better, so would we

src/test/stub/rmc_vec.rs
```diff
@@ -0,0 +1,820 @@
...
+    ptr: RmcUnique<T>,
+    cap: usize,
+}
+
+impl<T> RmcRawVec<T> {
+    fn new() -> Self {
+        let elem_size = mem::size_of::<T>();
+        // NOTE: (Mark. B) This default allocation size is important.
+        // A default 0 size leads to complex solver queries for smaller vec operations
+        let cap = DEFAULT_CAPACITY;
```
danielsn, line 80: We could make this huge, with no penalty in CBMC, AFAIK

danielsn: Lets see if we can get a `__CBMC_MAX_MALLOC` intrinsic.

src/test/stub/rmc_vec.rs
```diff
@@ -0,0 +1,820 @@
...
+        let elem_size = mem::size_of::<T>();
+        let ptr = unsafe { RmcUnique::new(libc::malloc(cap * elem_size) as *mut T) };
+        RmcRawVec { ptr, cap }
+    }
+
+    // grow() calls sized_realloc. By default, the new memory has twice the capacity
+    // of the old one.
+    fn grow(&mut self) {
+        let elem_size = mem::size_of::<T>();
+        let new_cap = 2 * self.cap;
```
danielsn, line 95: If we had a max sized array, grow is a no-op

src/test/stub/rmc_vec.rs
```diff
@@ -0,0 +1,820 @@
...
+
+    pub fn insert(&mut self, index: usize, elem: T) {
+        assert!(index <= self.len);
+        if self.cap() == self.len {
+            self.buf.grow();
+        }
+
+        unsafe {
+            if index < self.len {
+                copy(
```
danielsn, line 248: This does the shift

src/test/stub/rmc_vec.rs
```diff
@@ -0,0 +1,820 @@
...
+            }
+            write(self.ptr().offset(index as isize), elem);
+            self.len += 1;
+        }
+    }
+
+    pub fn remove(&mut self, index: usize) -> T {
+        assert!(index < self.len);
+        unsafe {
+            self.len -= 1;
```
danielsn, line 262: This does the shift as well

src/test/stub/rmc_vec.rs
```diff
@@ -0,0 +1,820 @@
...
+            );
+            result
+        }
+    }
+
+    pub fn extend<I: Iterator>(&mut self, iter: I) where I: Iterator<Item = T> {
+        iter.for_each(move |element| {
+            self.push(element);
+        });
+    }
```
danielsn, line 277: Is there a way to coalesce these operations

src/test/stub/rmc_vec.rs
```diff
@@ -0,0 +1,820 @@
...
+    pub fn as_mut_ptr(&mut self) -> *mut T {
+        self.ptr()
+    }
+
+    pub fn as_slice(&self) -> &[T] {
+        self
+    }
+
+    pub fn as_mut_slice(&mut self) -> &mut [T] {
+        self
```
danielsn, line 306: This is some Rust magic here

src/test/stub/rmc_vec.rs
```diff
@@ -0,0 +1,820 @@
...
+    }
+
+    pub fn as_ptr(&self) -> *const T {
+        self.buf.ptr.as_const_ptr()
+    }
+
+    pub fn truncate(&mut self, len: usize) {
+        // TODO: For soundness wrt standard library, call drop_in_place() for pointer
```
danielsn, line 314: precondition: `len <= self.len`

src/test/stub/rmc_vec.rs
```diff
@@ -0,0 +1,820 @@
...
+
+        unsafe {
+            let last = read(self.as_ptr().add(len - 1));
+            let hole = self.as_mut_ptr().add(index);
+            self.set_len(len - 1);
+            replace(hole, last)
+        }
+    }
+    pub fn capacity(&self) -> usize {
+        self.buf.capacity()
```
danielsn, line 374: Technically, this is not an over-approximation of what the standard library can do

danielsn (changes requested):

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+// RMCRawVec consists of a pointer to allocated memory and another variable tracking
+// the capacity of the allocation.
+struct RmcRawVec<T> {
+    ptr: *const T,
+    cap: usize,
+}
+
+impl<T> RmcRawVec<T> {
+    fn new() -> Self {
+        let elem_size = mem::size_of::<T>();
```
danielsn, line 88: Maybe warn or assert that this is not zero, if you're worried about soundness in that case?

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+    // equal to self.len() + additional.
+    // Does nothing if capacity is already sufficient."
+    //
+    // The important point to note here is that it is expected to reserve space
+    // for "atleast" additional more elements. Because of which, there cannot be
+    // any guarantees made about how much the exact capacity would be after the
+    // grow()/reserve() operation is performed.
+    //
+    // For the purpose of this implementation, we follow the specifics implemented
+    // in raw_vec.rs -> grow_amortized(). We choose:
```
danielsn, line 134: link

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+    pub fn new() -> Self {
+        Vec { buf: RmcRawVec::new(), len: 0, allocator: RmcAllocator::new() }
+    }
+
+    pub fn with_capacity(cap: usize) -> Self {
+        Self::with_capacity_in(cap, RmcAllocator::new())
+    }
+
+    // A lot of invariants here are not checked:
+    // * If the pointer was not allocated via a String/Vec, it is highly likely to be
```
danielsn, line 282: How hard would it be to just assert some of these

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+        }
+    }
+
+    // Clears the vector, removing all values.
+    // This method has no effect on the allocated capacity of the vector
+    pub fn clear(&mut self) {
+        self.truncate(0);
+    }
+
+    pub fn swap_remove(&mut self, index: usize) -> T {
```
danielsn, line 444: What does this do

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+        while i < olen {
+            drain.push(other.pop().unwrap());
+            i += 1;
+        }
+        // Empty other
+        i = 0;
+        while i < olen {
+            self.push(drain.pop().unwrap());
+            i += 1;
+        }
```
danielsn, line 514: Why can't this whole thing be a `memmove` followed by reducing the length of the second array?

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+    {
+        let len = self.len();
+
+        if new_len > len {
+            let additional = new_len - len;
+            self.reserve(additional);
+            let mut closure = f;
+            for _ in 0..additional {
+                // This push is cheap as we have already reserved enough space.
+                self.push(closure());
```
danielsn, line 536: do we want to have a `push_unsafe` that avoids the checks?

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+    // Clones and appends all elements in a slice to the Vec.
+    //
+    // Iterates over the slice other, clones each element, and then appends it to this Vec. The
+    // other vector is traversed in-order.
+    pub fn extend_from_slice(&mut self, other: &[T]) {
+        let other_len = other.len();
+        self.reserve(other_len);
+        for i in 0..other_len {
+            // This push is cheap as we have already reserved enough space.
+            self.push(other[i].clone());
```
danielsn, line 610: `push_unsafe`

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+        }
+    }
+}
+
+// Drop is codegen for most types, no need to perform any action here.
+// TODO: Check here.
```
danielsn, line 616: ticket?

danielsn: Or at least some tests covering possible cases

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+        }
+    }
+}
+
+__impl_slice_eq1! { [A] Vec<T, A>, &[U] }
+__impl_slice_eq1! { [A] Vec<T, A>, &mut [U] }
+__impl_slice_eq1! { [A] &[T], Vec<U, A> }
+__impl_slice_eq1! { [A] &mut [T], Vec<U, A> }
+__impl_slice_eq1! { [A, const N: usize] Vec<T, A>, [U; N] }
+__impl_slice_eq1! { [A, const N: usize] Vec<T, A>, &[U; N] }
```
danielsn, line 668: nice

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+        unsafe { ::std::slice::from_raw_parts_mut(self.ptr() as *mut T, self.len) }
+    }
+}
+
+// Clone
+impl<T: Clone, A: Allocator + Clone> Clone for Vec<T, A> {
+    fn clone(&self) -> Self {
+        let mut v = Self::with_capacity_in(self.len(), self.allocator.clone());
+        for idx in 0..self.len() {
+            v.push(self[idx].clone());
```
danielsn, line 703: push_unsafe

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+// as a series of values. The Extend trait bridges this gap, allowing you to
+// extend a collection by including the contents of that iterator. When extending
+// a collection with an already existing key, that entry is updated or, in the
+// case of collections that permit multiple entries with equal keys, that
+// entry is inserted."
+//
+// We cannot reserve space for the elements which are added as we dont know
+// the size of the iterator. In this case, we perform sequential push operations.
+// However because our underlying Vector grows exponential in size, we can be
+// sure that we wont perform too many resizing operations.
```
danielsn, line 749: won't

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+}
+
+// AsMut to a slice is possible because we implement the Deref coercion
+impl<T, A: Allocator> AsMut<[T]> for Vec<T, A> {
+    fn as_mut(&mut self) -> &mut [T] {
+        self
+    }
+}
+
+// Debug
```
danielsn, line 806: do we even need/want to implement this

library/rmc/stubs/Rust/vec/rmc_vec.rs
```diff
@@ -0,0 +1,1085 @@
...
+            mem::forget(self);
+
+            RmcIntoIter { iter, _buf: buf }
+        }
+    }
+}
+
+// FromIterator defines how a Vec will be created from an Iterator.
+impl<T> FromIterator<T> for Vec<T> {
+    fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Vec<T> {
```
danielsn, line 1018: didn't we already see something like this?

library/rmc/stubs/Rust/vec/rmc_vec.rs
danielsn: Chinmay asserted the easy ones

library/rmc/stubs/Rust/vec/rmc_vec.rs
danielsn: The testcases are in the next PR

danielsn: #465

danielsn (approved):

PR #387, opened 2021-08-04
Fix extern statics to be codegenned as extern

### LINKED ISSUE #369: Generated symbols for statics are sometimes not labelled correctly.
Two cases here:

1) `SizeAndAlignOfDst/main_fail.rs`

Running `rmc --keep-temps main_fail.rs` and then opening `main_fail.json`, we find the following symbol:
```
    "_ZN3std9panicking11panic_count18GLOBAL_PANIC_COUNT17hbb0e4a14d9617703E": {
      "baseName": "_ZN3std9panicking11panic_count18GLOBAL_PANIC_COUNT17hbb0e4a14d9617703E",
      "isAuxiliary": false,
      "isExported": false,
      "isExtern": false,
      "isFileLocal": false,
      "isInput": false,
      "isLvalue": true,
      "isMacro": false,
      "isOutput": false,
      "isParameter": false,
      "isProperty": false,
      "isStateVar": true,
      "isStaticLifetime": false,
      "isThreadLocal": true,
      "isType": false,
      "isVolatile": false,
      "isWeak": false,
      ...
```

which comes from the Rust library: [Rust Internal Docs](https://stdrs.dev/nightly/x86_64-pc-windows-gnu/std/panicking/panic_count/static.GLOBAL_PANIC_COUNT.html)

I believe this should have, instead:
```
      "isStaticLifetime": true,
      "isThreadLocal": false,
```

2) The following code (`main.rs`):

```
static A: u32 = 1;
static mut B: u32 = 2;

extern "C" {
    static C: u32;
    static mut D: u32;
}

fn main() {
    static E: u32 = 5;
    static mut F: u32 = 6;

    unsafe {
        assert!(A == 1);
        assert!(B == 2);
        assert!(C == 3);
        assert!(D == 4);
        assert!(E == 5);
        assert!(F == 6);
    }
}
```

Running `rmc --keep-temps main.rs` and then opening `main.json`, we find the a symbol for each static. All of the fields are the same as above except `"baseName"`, and possibly `"isStaticLifetime"` and `"isThreadLocal"`.

The ones with `isStaticLifetime` as `true` and `isThreadLocal` as `false` are `A`, `B`, `E`,  and `F`. 
The ones with `isStaticLifetime` as `false` and `isThreadLocal` as `true` are `C` and `D`.

---

It looks from the above examples that statics within `extern` blocks and that are in the Rust library are given incorrect tags.

Note that if we don't include the assertions in the second example, then `A`, `B`, `E`, and `F` symbols are all still generated, but `C` and `D` are not.

vecchiot-aws:
### Description of changes: 

Currently, given code like the following, the symbol for `A` is not labelled as extern. This PR fixes that, and also cleans up how static variable symbols are created.

```Rust
extern "C" {
    static mut A: u32;
}
```

### Resolved issues:

Resolves #369 

### Testing:

* Is this a refactor change? None.

### Checklist

### DIFF
--- a/compiler/rustc_codegen_llvm/src/gotoc/operand.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/operand.rs
@@ -329,11 +329,11 @@ impl<'tcx> GotocCtx<'tcx> {
 
                 let sym = self.ensure(&self.symbol_name(instance), |ctx, name| {
                     // check if this static is extern
-                    // TODO: copying what cranelift did here
                     let rlinkage = ctx.tcx.codegen_fn_attrs(def_id).linkage;
 
-                    // we believe rlinkage some => not extern
-                    assert!(rlinkage.is_none()); // leave assert, add comment to make ticket
+                    // we believe rlinkage being `Some` means the static not extern
+                    // see https://github.com/model-checking/rmc/issues/388
+                    assert!(rlinkage.is_none());
 
                     let span = ctx.tcx.def_span(def_id);
                     Symbol::static_variable(
--- a/compiler/rustc_codegen_llvm/src/gotoc/operand.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/operand.rs
@@ -332,6 +332,7 @@ impl<'tcx> GotocCtx<'tcx> {
                     let rlinkage = ctx.tcx.codegen_fn_attrs(def_id).linkage;
 
                     // we believe rlinkage being `Some` means the static not extern
+                    // based on compiler/rustc_codegen_cranelift/src/linkage.rs#L21
                     // see https://github.com/model-checking/rmc/issues/388
                     assert!(rlinkage.is_none());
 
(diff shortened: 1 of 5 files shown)


compiler/rustc_codegen_llvm/src/gotoc/operand.rs
```diff
@@ -323,20 +323,28 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                    Symbol::variable(
-                        name.clone(),
-                        name.clone(),
-                        tcx.codegen_ty(instance.ty(tcx.tcx, ty::ParamEnv::reveal_all())),
-                        tcx.codegen_span2(&span),
+                let instance = Instance::mono(self.tcx, def_id);
+
+                let sym = self.ensure(&self.symbol_name(instance), |ctx, name| {
+                    // check if this static is extern
+                    // TODO: copying what cranelift did here
```
avanhatt, line 332: Nit: remove `TODO`, reference Cranelift file

compiler/rustc_codegen_llvm/src/gotoc/operand.rs
```diff
@@ -323,20 +323,28 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    let rlinkage = ctx.tcx.codegen_fn_attrs(def_id).linkage;
+
+                    // we believe rlinkage some => not extern
+                    assert!(rlinkage.is_none()); // leave assert, add comment to make ticket
```
avanhatt, line 336: Add issue

avanhatt (changes requested):

avanhatt (approved):

danielsn (approved):

PR #389, opened 2021-08-05
Add initial support for Litani build and dashboard.

### LINKED ISSUE #384: Use `Litani` to run the reference examples and display the results.

bdalrhm:
### Description of changes:

This PR adds initial support for Litani build. We use Litani to run RMC on the examples and generate an HTML dashboard with the results.

### Resolved issues:

Resolves #384 & #392

### Call-outs:

- `src/tools/dashboard/src/util.rs` contains a lot of duplicated code from `compiletest`. Consider reconciling with `compiletest` or porting all of our test suites to `Litani`.
- We run the examples in the reference twice to generate the terminal dashboard and Litani's dashboard. Consider using `run.json` to generate the terminal dashboard.
- The default names stages Litani uses (`build`, `test`, and `report`) does not match with our stages (`check`, `codegen`, `verification`). Switch to custom stage names when Litani adds support for that.

### Testing:

* How is this change tested?
  * The generated HTML dashboard closely matches the terminal version.
* Is this a refactor change?
  * No.

### Checklist

### DIFF
--- a/src/tools/dashboard/src/litani.rs
+++ b/src/tools/dashboard/src/litani.rs
@@ -50,6 +50,7 @@ impl Litani {
         pipeline: &str,
         stage: &str,
         exit_status: i32,
+        timeout: u32,
     ) {
         let mut job = Command::new("litani");
         // The given command may contain additional env vars. Prepend those vars
@@ -80,7 +81,7 @@ impl Litani {
             "--ok-returns",
             &exit_status.to_string(),
             "--timeout",
-            "10",
+            &timeout.to_string(),
         ]);
         if !inputs.is_empty() {
             job.arg("--inputs").args(inputs);
--- a/src/tools/dashboard/src/litani.rs
+++ b/src/tools/dashboard/src/litani.rs
@@ -1,0 +1,96 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! Utilities to interact with the `Litani` build accumulator.
+
+use pulldown_cmark::escape::StrWrite;
+use std::collections::HashMap;
+use std::path::Path;
+use std::process::{Child, Command};
+
+/// Data structure representing a `Litani` build.
+pub struct Litani {
+    /// A buffer of the `spawn`ed Litani jobs so far. `Litani` takes some time
+    /// to execute each `add-job` command and executing thousands of them
+    /// sequentially takes a considerable amount of time. To speed up the
+    /// execution of those commands, we spawn those commands sequentially (as
+    /// normal). However, instead of `wait`ing for each process to terminate,
+    /// we add its handle to a buffer of the `spawn`ed processes and continue
+    /// with our program. Once we are done adding jobs, we wait for all of them
+    /// to terminate before we run the `run-build` command.
+    spawned_commands: Vec<Child>,
+}
+
+impl Litani {
+    /// Sets up a new [`Litani`] run.
+    pub fn init(project_name: &str, output_prefix: &Path, output_symlink: &Path) -> Self {
+        Command::new("litani")
+            .args([
+                "init",
+                "--project-name",
+                project_name,
+                "--output-prefix",
+                output_prefix.to_str().unwrap(),
+                "--output-symlink",
+                output_symlink.to_str().unwrap(),
+            ])
+            .spawn()
+            .unwrap()
+            .wait()
+            .unwrap();
+        Self { spawned_commands: Vec::new() }
+    }
+
+    /// Adds a single command with its dependencies.
+    pub fn add_job(
+        &mut self,
+        command: &Command,
+        description: &str,
+        pipeline: &str,
+        stage: &str,
+        exit_status: i32,
+    ) {
+        let mut job = Command::new("litani");
+        // The given command may contain additional env vars. Prepend those vars
+        // to the command before passing it to Litani.
+        let job_envs: HashMap<_, _> = job.get_envs().collect();
+        let mut new_envs = String::new();
+        command.get_envs().filter(|(k, _)| !job_envs.contains_key(k)).fold(
+            &mut new_envs,
+            |fmt, (k, v)| {
+                fmt.write_fmt(format_args!(
+                    "{}=\"{}\" ",
+                    k.to_str().unwrap(),
+                    v.unwrap().to_str().unwrap()
+                ))
+                .unwrap();
+                fmt
+            },
+        );
+        job.args([
+            "add-job",
+            "--command",
+            &format!("{}{:?}", new_envs, command),
+            "--description",
+            description,
+            "--pipeline-name",
+            pipeline,
+            "--ci-stage",
+            stage,
+            "--ok-returns",
+            &exit_status.to_string(),
+        ]);
+        // Start executing the command, but do not wait for it to terminate.
+        self.spawned_commands.push(job.spawn().unwrap());
+    }
+
+    /// Starts a [`Litani`] run.
+    pub fn run_build(&mut self) {
+        // Wait for all spawned processes to terminate.
+        for command in self.spawned_commands.iter_mut() {
+            command.wait().unwrap();
+        }
+        self.spawned_commands.clear();
+        // Run `run-build` command and wait for it to finish.
+        Command::new("litani").arg("run-build").spawn().unwrap().wait().unwrap();
+    }
+}
--- a/src/tools/dashboard/src/litani.rs
+++ b/src/tools/dashboard/src/litani.rs
@@ -54,18 +54,17 @@ impl Litani {
         // to the command before passing it to Litani.
         let job_envs: HashMap<_, _> = job.get_envs().collect();
         let mut new_envs = String::new();
-        command.get_envs().filter(|(k, _)| !job_envs.contains_key(k)).fold(
-            &mut new_envs,
-            |fmt, (k, v)| {
+        command.get_envs().fold(&mut new_envs, |fmt, (k, v)| {
+            if !job_envs.contains_key(k) {
                 fmt.write_fmt(format_args!(
                     "{}=\"{}\" ",
                     k.to_str().unwrap(),
                     v.unwrap().to_str().unwrap()
                 ))
                 .unwrap();
-                fmt
-            },
-        );
+            }
+            fmt
+        });
         job.args([
             "add-job",
             "--command",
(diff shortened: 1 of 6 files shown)


src/tools/dashboard/src/litani.rs
```diff
@@ -0,0 +1,96 @@
...
+        // Start executing the command, but do not wait for it to terminate.
+        self.spawned_commands.push(job.spawn().unwrap());
+    }
+
+    /// Starts a [`Litani`] run.
+    pub fn run_build(&mut self) {
+        // Wait for all spawned processes to terminate.
+        for command in self.spawned_commands.iter_mut() {
+            command.wait().unwrap();
+        }
```
adpaco-aws, line 91: What if they do not terminate? Are we using timeouts when invoking litani?

src/tools/dashboard/src/litani.rs
```diff
@@ -0,0 +1,96 @@
...
+            |fmt, (k, v)| {
+                fmt.write_fmt(format_args!(
+                    "{}=\"{}\" ",
+                    k.to_str().unwrap(),
+                    v.unwrap().to_str().unwrap()
+                ))
+                .unwrap();
+                fmt
+            },
+        );
```
adpaco-aws, line 68: Is this block of code just pretty-printing a HashMap struct? There must be a more idiomatic way of doing that.

src/tools/dashboard/src/util.rs
```diff
@@ -0,0 +1,198 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! The types and procedures in this modules are similar to the ones in
+//! `compiletest`. Consider using `Litani` to run the test suites.
+
```
adpaco-aws, line 8: We should import these types and procedures instead. Or add them here and import in `compiletest`.

src/tools/dashboard/src/litani.rs
```diff
@@ -0,0 +1,96 @@
...
+/// Data structure representing a `Litani` build.
+pub struct Litani {
+    /// A buffer of the `spawn`ed Litani jobs so far. `Litani` takes some time
+    /// to execute each `add-job` command and executing thousands of them
+    /// sequentially takes a considerable amount of time. To speed up the
+    /// execution of those commands, we spawn those commands sequentially (as
+    /// normal). However, instead of `wait`ing for each process to terminate,
+    /// we add its handle to a buffer of the `spawn`ed processes and continue
+    /// with our program. Once we are done adding jobs, we wait for all of them
+    /// to terminate before we run the `run-build` command.
```
adpaco-aws, line 19: We can just say that we spawn all litani asynchronously and wait for them to terminate.

adpaco-aws (changes requested):

src/tools/dashboard/src/litani.rs
bdalrhm: We're not running the tests here. We're only adding them as jobs for Litani to run later. This command takes less around 0.1 second.  However, the `run-build` command after this may not terminate, but that's a good hint that we're missing a `--unwind` argument somewhere.

src/tools/dashboard/src/util.rs
bdalrhm: Opened an issue for this.

src/tools/dashboard/src/litani.rs
bdalrhm: hmm... I haven't found anything online. I could use `flat_map` instead if you prefer that.

src/tools/dashboard/src/litani.rs
adpaco-aws: Can we add a timeout to the run-build then? Litani should have that feature.

karkhaz: Can do. Currently, individual jobs can take a `--timeout` flag which is often more useful, but added an [issue for this](https://github.com/awslabs/aws-build-accumulator/issues/34).

src/tools/dashboard/src/litani.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /// Adds a single command with its dependencies.
+    pub fn add_job(
+        &mut self,
+        command: &Command,
+        description: &str,
+        pipeline: &str,
+        stage: &str,
+        exit_status: i32,
```
karkhaz, line 50: Suggest also taking an `inputs` and an `outputs` flag.

src/tools/dashboard/src/litani.rs
```diff
@@ -0,0 +1,97 @@
...
+                    "{}=\"{}\" ",
+                    k.to_str().unwrap(),
+                    v.unwrap().to_str().unwrap()
+                ))
+                .unwrap();
+            }
+            fmt
+        });
+        job.args([
+            "add-job",
```
karkhaz, line 69: If you're using inputs and outputs as I suggest below, I suggest also passing in `--phony-outputs`; this is available in the develop branch of Litani, we'll release it probably next week

src/tools/dashboard/src/util.rs
```diff
@@ -0,0 +1,203 @@
...
+    let mut rmc_rustc = Command::new("rmc-rustc");
+    rmc_rustc.args(&test_props.rustc_args).args(["-Z", "no-codegen"]).arg(&test_props.path);
+    // TODO: replace `build` with `check` when Litani adds support for custom
+    // stages (see issue #391).
+    litani.add_job(
+        &rmc_rustc,
+        "Is this valid Rust code?",
+        test_props.path.to_str().unwrap(),
+        "build",
+        exit_status,
```
karkhaz, line 141: Suggest output should be `test_props.path.to_str() + "build"`

src/tools/dashboard/src/util.rs
```diff
@@ -0,0 +1,203 @@
...
+        .args(["-Z", "codegen-backend=gotoc", "--cfg=rmc", "--out-dir", "build/tmp"])
+        .arg(&test_props.path);
+    // TODO: replace `test` with `codegen` when Litani adds support for custom
+    // stages (see issue #391).
+    litani.add_job(
+        &rmc_rustc,
+        "Does RMC support all the Rust features used in it?",
+        test_props.path.to_str().unwrap(),
+        "test",
+        exit_status,
```
karkhaz, line 163: Input should be `test_props.path.to_str() + "build"`, output should be `test_props.path.to_str() + "test"`

src/tools/dashboard/src/util.rs
```diff
@@ -0,0 +1,203 @@
...
+        rmc.env("RUSTFLAGS", test_props.rustc_args.join(" "));
+    }
+    // TODO: replace `report` with `verification` when Litani adds support for
+    // custom stages (see issue #391).
+    litani.add_job(
+        &rmc,
+        "Can RMC reason about it?",
+        test_props.path.to_str().unwrap(),
+        "report",
+        exit_status,
```
karkhaz, line 182: Input should be test_props.path.to_str() + "test"

src/tools/dashboard/src/litani.rs
```diff
@@ -0,0 +1,96 @@
...
+pub struct Litani {
+    /// A buffer of the `spawn`ed Litani jobs so far. `Litani` takes some time
+    /// to execute each `add-job` command and executing thousands of them
+    /// sequentially takes a considerable amount of time. To speed up the
+    /// execution of those commands, we spawn those commands sequentially (as
+    /// normal). However, instead of `wait`ing for each process to terminate,
+    /// we add its handle to a buffer of the `spawn`ed processes and continue
+    /// with our program. Once we are done adding jobs, we wait for all of them
+    /// to terminate before we run the `run-build` command.
+    spawned_commands: Vec<Child>,
```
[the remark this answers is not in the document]
bdalrhm: That's what `spawned_commands` is for 🙂. We _spawn_ the `litani add-job` processes sequentially. But the processes themselves _run_ concurrently. I experimented with _spawning_ those processes concurrently, but did not see any performance improvements as it takes much longer to execute and terminate `litani add-job` processes concurrently than it takes spawn them and parse all tests sequentially (it's Rust after all 😉).

src/tools/dashboard/src/litani.rs
bdalrhm: That's SO COOOL!!!

src/tools/dashboard/src/util.rs
```diff
@@ -0,0 +1,217 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! Utilities types and procedures to parse and run tests.
+//!
+//! TODO: The types and procedures in this modules are similar to the ones in
+//! `compiletest`. Consider using `Litani` to run the test suites (see issue
+//! #390).
```
adpaco-aws, line 7: Put the link to the issue here.

src/tools/dashboard/src/litani.rs
```diff
@@ -0,0 +1,105 @@
...
+            "--description",
+            description,
+            "--pipeline-name",
+            pipeline,
+            "--ci-stage",
+            stage,
+            "--ok-returns",
+            &exit_status.to_string(),
+            "--timeout",
+            "10",
```
adpaco-aws, line 83: We should avoid hard-coding here by making timeout a parameter of `add_job`.

src/tools/dashboard/src/util.rs
```diff
@@ -0,0 +1,217 @@
...
+        "build",
+        exit_status,
+    );
+}
+
+/// Is RMC expected to codegen all the Rust features in the test?
+pub fn add_codegen_job(litani: &mut Litani, test_props: &TestProps) {
+    // TODO: again, add logic to ensure that the next step runs only if this
+    // step succeeds (and RMC is not expected to panic in this step) (see issue
+    // #392).
```
adpaco-aws, line 153: Link instead of issue number.

src/tools/dashboard/src/util.rs
```diff
@@ -0,0 +1,217 @@
...
+    let exit_status = if test_props.fail_step == Some(FailStep::Codegen) { 1 } else { 0 };
+    let mut rmc_rustc = Command::new("rmc-rustc");
+    rmc_rustc
+        .args(&test_props.rustc_args)
+        .args(["-Z", "codegen-backend=gotoc", "--cfg=rmc", "--out-dir", "build/tmp"])
+        .arg(&test_props.path);
+    // TODO: replace `test` with `codegen` when Litani adds support for custom
+    // stages (see issue #391).
```
adpaco-aws, line 161: Same here.

adpaco-aws (changes requested):
I think it is better to insert links instead of referencing issue numbers. This way developers can just click on them.

src/tools/dashboard/src/litani.rs
bdalrhm: Done! Changed the timeout value for `check` and `codegen` steps to 1 sec.

adpaco-aws (approved):

PR #396, opened 2021-08-06
Fix failing codegen in due to incorrect boolean types

### LINKED ISSUE #394: Panic with a bitwise xor operator
I tried this code:

```rust
fn main() {
    let a: bool = __nondet();
    let b: bool = __nondet();
    let c = a ^ b;
    assert!((a == b && !c) || (a != b && c));
}
```

and it crashed with the following message:
```
thread 'rustc' panicked at 'BinaryOperation Expression does not typecheck Xor Expr { value: Symbol { identifier: "main::1::var_5" }, typ: CInteger(Bool), location: None } Expr { value: Symbol { identifier: "main::1::var_6" }, typ: CInteger(Bool), location: None }', compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs:863:9
```

chinmaydd:
### Description of changes: 

RMC fails in codegen when generating expression with boolean XOR if the type of operands is `CInteger(Bool)` as CBMC requires them to be `Bool` (https://github.com/diffblue/cbmc/blob/develop/src/solvers/prop/prop_conv_solver.cpp#L194)

### Resolved issues:

Resolves #394 

### Call-outs:

None

### Testing:

* How is this change tested?
Created a new test case out of #394 

* Is this a refactor change?
No

### Checklist

### DIFF
--- a/compiler/rustc_codegen_llvm/src/gotoc/rvalue.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/rvalue.rs
@@ -52,21 +52,21 @@ impl<'tcx> GotocCtx<'tcx> {
             BinOp::Rem => ce1.rem(ce2),
             BinOp::BitXor => {
                 if self.operand_ty(e1).is_bool() {
-                    ce1.xor(ce2)
+                    ce1.cast_to(Type::bool()).xor(ce2.cast_to(Type::bool()))
                 } else {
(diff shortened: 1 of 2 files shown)


chinmaydd:
Oh, I did not see #395. I guess this PR can be closed in favor of that.

PR #399, opened 2021-08-06
RMC log file

### LINKED ISSUE #382: Create `rmc_log` and `rmc_warn` macros
These macros will write both to stdout and to a log file in a structured format so we can extract information from them in a procedural way in post-processing. See #350, for example.

vecchiot-aws:
### Description of changes: 

Introduces a structured log file for RMC so that we don't have to do regexp parsing of debug output. User can use the `--save-logs` flag to keep a log file after execution terminates, and optionally can provide a particular filename for it (logs will go on end of provided file if it already exists).

If the flag is not provided, we still generate a log file as a temporary for use within the Python scripts (currently uses `rmc.log`).

### Resolved issues:

Resolves #382 

### Call-outs:

Currently, we use environment variables to pass the log file path to the Rust code. Unsure if this is the best approach.

### Testing:

* How is this change tested? Existing regression suite; manual testing for particular feature.

### Checklist

### DIFF
--- a/compiler/rustc_codegen_llvm/src/gotoc/logging.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/logging.rs
@@ -61,8 +61,10 @@ impl ToJson for LogType {
 #[derive(Debug)]
 pub enum WarningType {
     Concurrency,
+    GlobalAsm,
+    MissingSymbol,
+    TypeMismatch,
     Unsupported,
-    Other,
 }
 
 impl ToJson for WarningType {
--- a/compiler/rustc_codegen_llvm/src/gotoc/logging.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/logging.rs
@@ -1,0 +1,127 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! Logging for RMC to write structured logs to a log file.
+use crate::btree_string_map;
+use rustc_serialize::json::*;
+use std::env;
+use std::fs::OpenOptions;
+use std::io::prelude::*;
+use tracing::{debug, warn};
+
+pub use crate::{rmc_debug, rmc_log, rmc_warn};
+
+/// A line in the log file.
+struct LogLine {
+    log_type: LogType,
+    message: String,
+}
+
+impl ToJson for LogLine {
+    fn to_json(&self) -> Json {
+        let output = btree_string_map![
+            ("log_type", self.log_type.to_json()),
+            ("message", self.message.to_json())
+        ];
+
+        Json::Object(output)
+    }
+}
+
+/// The kind of log message.
+#[derive(Debug)]
+enum LogType {
+    Debug,
+    Log,
+    Warning(WarningType),
+}
+
+impl ToJson for LogType {
+    fn to_json(&self) -> Json {
+        match self {
+            LogType::Debug => {
+                let output = btree_string_map![("type", "DEBUG".to_json())];
+                Json::Object(output)
+            }
+            LogType::Log => {
+                let output = btree_string_map![("type", "LOG".to_json())];
+                Json::Object(output)
+            }
+            LogType::Warning(warning_type) => {
+                let output = btree_string_map![
+                    ("type", "WARNING".to_json()),
+                    ("warning_kind", warning_type.to_json())
+                ];
+                Json::Object(output)
+            }
+        }
+    }
+}
+
+/// The kind of warning message.
+#[derive(Debug)]
+pub enum WarningType {
+    Concurrency,
+    Unsupported,
+    Other,
+}
+
+impl ToJson for WarningType {
+    fn to_json(&self) -> Json {
+        format!("{:?}", self).to_json()
+    }
+}
+
+/// Writes a given log line to the log file specified through
+/// the the environment variable RMC_LOG_FILE.
+/// If this is not set, continues without failing.
+fn write_to_log_file(log_line: LogLine) {
+    let line = format!("{}\n", log_line.to_json().to_string());
+    match env::var("RMC_LOG_FILE") {
+        Err(_) => (),
+        Ok(path) => {
+            OpenOptions::new().append(true).create(true).open(&path)
+                .expect(&format!("Internal error: Unable to open log file at location {}; try specifying a different location for the log file using the `--save-logs` flag. If this issue persists, file a ticket at https://github.com/model-checking/rmc/issues/new?assignees=&labels=bug&template=bug_report.md", &path))
+                .write_all(line.as_bytes())
+                .expect(&format!("Internal error: Error writing to log file."))
+        }
+    }
+}
+
+#[macro_export]
+macro_rules! rmc_debug {
+    ($( $parts:expr ),*) => {
+        let message = rustc_middle::ty::print::with_no_trimmed_paths(|| format!($( $parts, )*));
+        crate::gotoc::logging::write_debug(message);
+    }
+}
+
+pub fn write_debug(message: String) {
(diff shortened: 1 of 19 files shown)


scripts/cargo-rmc
```diff
@@ -144,6 +145,14 @@ def parse_args():
         if args.quiet:
             args.verbose = False
 
+        # If --save-logs not present, use default and delete after done
```
celinval, line 148: Should we print to the console instead?

scripts/rmc_flags.py
```diff
@@ -82,6 +82,9 @@ def add_artifact_flags(make_group, add_flag, config):
              help="Generate a goto symbol table")
     add_flag(group, "--keep-temps", default=False, action=BooleanOptionalAction,
              help="Keep temporary files generated throughout RMC process")
+    # False => flag not present; None => flag present, no argument provided
+    add_flag(group, "--save-logs", nargs="?", type=pl.Path, default=False,
+             help="Save RMC logs to <target_dir>/rmc.log, or provide a file to save logs to.")
```
celinval, line 87: Can you please state the default behavior?

scripts/rmc_flags.py
```diff
@@ -82,6 +82,9 @@ def add_artifact_flags(make_group, add_flag, config):
              help="Generate a goto symbol table")
     add_flag(group, "--keep-temps", default=False, action=BooleanOptionalAction,
              help="Keep temporary files generated throughout RMC process")
+    # False => flag not present; None => flag present, no argument provided
+    add_flag(group, "--save-logs", nargs="?", type=pl.Path, default=False,
```
celinval, line 86: const=RMC_LOG_FILE

compiler/rustc_codegen_llvm/src/gotoc/logging.rs
```diff
@@ -0,0 +1,96 @@
...
+#[macro_export]
+macro_rules! rmc_log {
+    ($( $parts:expr ),*) => {
+        use crate::gotoc::logging::write_log;
+        let message = format!($( $parts, )*);
+        write_log(message).unwrap();
+    }
+}
+
+pub fn write_log(message: String) -> io::Result<()> {
```
celinval, line 81: Should this be private?

compiler/rustc_codegen_llvm/src/gotoc/logging.rs
```diff
@@ -0,0 +1,96 @@
...
+        format!("{:?}", self).to_json()
+    }
+}
+
+fn write_to_log_file(log_line: LogLine) -> io::Result<()> {
+    let line = format!("{}\n", log_line.to_json().to_string());
+    match env::var("RMC_LOG_FILE") {
+        Err(_) => Ok(()),
+        Ok(path) => {
+            OpenOptions::new().append(true).create(true).open(path)?.write_all(line.as_bytes())
```
celinval, line 67: What happens if this path is invalid or we don't have permission to write to it?

celinval (commented):
I really like the idea of being able to easily separate RMC logs from others. I do have a question though. Will the old behavior still be possible? Can we still log RMC statements together with the rest of the application logs?

Thanks!

scripts/rmc_flags.py
vecchiot-aws: Oooh, thanks!

scripts/cargo-rmc
vecchiot-aws: I'm currently implementing this, yes!

compiler/rustc_codegen_llvm/src/gotoc/logging.rs
vecchiot-aws: I'm going to add a clear error message for this case which will also mention setting the `--save-logs` flag. If you think we should try to recover instead, I'm happy to hear your case!

compiler/rustc_codegen_llvm/src/gotoc/logging.rs
vecchiot-aws: Since it's exposed in the macro, it can't be private (I don't think?).

compiler/rustc_codegen_llvm/src/gotoc/logging.rs
celinval: I am okay with a clear error message.

I would suggest to also check if we have permission to write to the file while parsing --save-logs so we can fail fast.

scripts/rmc_flags.py
vecchiot-aws: Done!

compiler/rustc_codegen_llvm/src/gotoc/mod.rs
```diff
@@ -458,7 +462,8 @@ impl CodegenBackend for GotocCodegenBackend {
                         );
                     }
                     MonoItem::GlobalAsm(_) => {
-                        warn!(
+                        rmc_warn!(
+                            WarningType::Other,
```
danielsn, line 466: Lets not leave this as Other.  We should have a good name that can be searched for. Maybe even just `GlobalAsm`

compiler/rustc_codegen_llvm/src/gotoc/logging.rs
```diff
@@ -0,0 +1,126 @@
...
+        format!("{:?}", self).to_json()
+    }
+}
+
+/// Writes a given log line to the log file specified through 
+/// the the environment variable RMC_LOG_FILE. 
+/// If this is not set, continues without failing.
+fn write_to_log_file(log_line: LogLine) {
+    // Still print debug message to user.
+    debug!("RMC: [{:?}]: {}", log_line.log_type, log_line.message);
```
danielsn, line 78: What about warning messages? Don't we want to print those to the user even if debug is off

danielsn (changes requested):

danielsn:
Make sure we set "flush" on the writes to the file

compiler/rustc_codegen_llvm/src/gotoc/logging.rs
```diff
@@ -0,0 +1,126 @@
...
+
+/// A line in the log file.
+struct LogLine {
+    log_type: LogType,
+    message: String,
+}
+
+impl ToJson for LogLine {
+    fn to_json(&self) -> Json {
+        let output = btree_map![
```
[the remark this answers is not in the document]
vecchiot-aws: Done.

compiler/rustc_codegen_llvm/src/gotoc/logging.rs
```diff
@@ -0,0 +1,126 @@
...
+    Debug,
+    Log,
+    Warning(WarningType),
+}
+
+impl ToJson for LogType {
+    fn to_json(&self) -> Json {
+        match self {
+            LogType::Debug => {
+                let output = btree_map![("type".to_string(), "DEBUG".to_json())];
```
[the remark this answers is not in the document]
vecchiot-aws: Done.

compiler/rustc_codegen_llvm/src/gotoc/logging.rs
vecchiot-aws: Fixed.

compiler/rustc_codegen_llvm/src/gotoc/logging.rs
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+    }
+}
+
+/// The kind of warning message.
+#[derive(Debug)]
+pub enum WarningType {
+    Unsupported,
+    Other,
```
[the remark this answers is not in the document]
vecchiot-aws: Done.

compiler/rustc_codegen_llvm/src/gotoc/mod.rs
vecchiot-aws: Done.

PR #402, opened 2021-08-06
Drop for dynamic objects

### LINKED ISSUE #66: Incorrect call to`drop_in_place` with dynamic argument

### LINKED ISSUE #435: RMC incorrectly reports an assertion failure on a two-dimensional vector
I tried this code:

```rust
fn main() {
    let a: Vec<Vec<i32>> = vec![vec![0; 2]; 1];
    assert!(a.len() == 1);
}
```

```
rmc test.rs
```

with RMC version: `main-154-2021-08-06`

I expected the assertion to pass, but it didn't:
```
[main.assertion.1] line 3 assertion failed: a.len() == 1: FAILURE
```
If I compile the test and run it, the assertion is not hit:
```
$ rustc test.rs 
$ ./test 
$ 
```

Also, if I modify the first line in the function to
```rust
    let a: Vec<Vec<i32>> = vec![vec![0; 2]];
```
i.e. leave out the length for the outer vector, the assertion passes.

avanhatt:
### Description of changes: 

RMC does not current codegen calls to drop for dynamic objects. This PR does that, inserting vtable calls as necessary.

### Resolved issues:

Resolves #66 
Should also resolve #435

### Call-outs:

- Known [issue](https://github.com/model-checking/rmc/issues/426) with the `SizeAndAlignOfDst` tests. This patch has a workaround to insert a `codegen_unimplemented` call in the case that drop fails to typecheck, which maintains soundness (the current implementation is not sound because it does not call drop at all).

### Testing:

* How is this change tested?

Existing plus 7 new regressions.

* Is this a refactor change?

Not really.

### Checklist


compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/place.rs
```diff
@@ -65,10 +65,18 @@ impl<'tcx> ProjectedPlace<'tcx> {
...
+    ) -> Option<(Type, Type)> {
         match typ {
-            TypeOrVariant::Type(t) => &ctx.codegen_ty(t) == expr.typ(),
-            TypeOrVariant::Variant(_) => true, //TODO, what to do here?
+            TypeOrVariant::Type(t) => {
+                let expr_ty = expr.typ().clone();
+                let type_from_mir = ctx.codegen_ty(t);
+                if expr_ty != type_from_mir { Some((expr_ty, type_from_mir)) } else { None }
+            }
+            TypeOrVariant::Variant(_) => None, //TODO, what to do here?
```
chinmaydd, line 79: Do we want to create an issue for this ? Not sure if this is tracked somewhere..

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/typ.rs
```diff
@@ -365,6 +367,7 @@ impl<'tcx> GotocCtx<'tcx> {
         use rustc_middle::ty::print::Printer;
         let mut name = String::new();
         let printer = FmtPrinter::new(self.tcx, &mut name, Namespace::TypeNS);
+        let t = self.monomorphize(t);
```
chinmaydd, line 370: I would like to add a comment here with the example:
```
StructTag("tag-std::boxed::Box<dyn std::error::Error + std::marker::Send + std::marker::Sync>") }
StructTag("tag-std::boxed::Box<(dyn std::error::Error + std::marker::Send + std::marker::Sync)>") }
```

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/place.rs
```diff
@@ -65,10 +65,18 @@ impl<'tcx> ProjectedPlace<'tcx> {
 
 /// Constructor
 impl<'tcx> ProjectedPlace<'tcx> {
-    fn check_expr_typ(expr: &Expr, typ: &TypeOrVariant<'tcx>, ctx: &mut GotocCtx<'tcx>) -> bool {
+    fn check_expr_typ_mismatch(
+        expr: &Expr,
+        typ: &TypeOrVariant<'tcx>,
+        ctx: &mut GotocCtx<'tcx>,
+    ) -> Option<(Type, Type)> {
```
chinmaydd, line 72: I like this QOL change !

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/typ.rs
avanhatt: Added, thanks!

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/rvalue.rs
```diff
@@ -716,26 +716,24 @@ impl<'tcx> GotocCtx<'tcx> {
...
                 .address_of()
                 .cast_to(trait_fn_ty)
         } else {
-            // TODO: check why in https://github.com/model-checking/rmc/issues/66
-            debug!(
-                "WARNING: drop_in_place not found for {:?}",
-                self.readable_instance_name(drop_instance),
-            );
-            Type::void_pointer().null().cast_to(trait_fn_ty)
+            self.codegen_unimplemented(
```
danielsn, line 731: Does this typecheck? We want a pointer to an unimplemented function, and instead we're getting an unimplemented function that returns a fn pointer.

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/statement.rs
```diff
@@ -99,27 +99,74 @@ impl<'tcx> GotocCtx<'tcx> {
     // TODO: this function doesn't handle unwinding which begins if the destructor panics
     // https://github.com/model-checking/rmc/issues/221
     fn codegen_drop(&mut self, location: &Place<'tcx>, target: &BasicBlock) -> Stmt {
-        // instance for drop function
         let loc_ty = self.place_ty(location);
-        let instance = Instance::resolve_drop_in_place(self.tcx, loc_ty);
-        if let Some(hk) = self.hooks.hook_applies(self.tcx, instance) {
+        let drop_instance = Instance::resolve_drop_in_place(self.tcx, loc_ty);
+        if let Some(hk) = self.hooks.hook_applies(self.tcx, drop_instance) {
```
danielsn, line 104: Do we still need the hook here? Or could we just move its logic inside the regular codegen?

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/statement.rs
```diff
@@ -99,27 +99,74 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                                trait_fat_ptr.to_owned().member("data", &self.symbol_table);
+                            let self_ref =
+                                self_ref.cast_to(trait_fat_ptr.typ().clone().to_pointer());
+
+                            let func_exp: Expr = fn_ptr.dereference();
+                            func_exp.call(vec![self_ref]).as_stmt(Location::none())
+                        }
+                        _ => {
+                            // Non-virtual, direct drop call
+                            assert!(!matches!(drop_instance.def, InstanceDef::Virtual(_, _)));
```
danielsn, line 137: Can we assert this above? Would that be a better check than matching on `loc_ty.kind()`

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/statement.rs
```diff
@@ -99,27 +99,74 @@ impl<'tcx> GotocCtx<'tcx> {
...
+
+                            let func = self.codegen_func_expr(drop_instance, None);
+                            let place = self.codegen_place(location);
+                            let arg = if let Some(fat_ptr) = place.fat_ptr_goto_expr {
+                                // Drop takes the fat pointer if it exists
```
danielsn, line 142: This would be in the case of dropping a slice?

src/test/cbmc/DynTrait/drop_nested_boxed_dyn_fail.rs
```diff
@@ -0,0 +1,41 @@
...
+        CELL = 0;
+    }
+
+    // Check nested box, still only incremented once
+    {
+        let x: Box<dyn Send> = Box::new(Concrete {});
+        let _nested: Box<dyn Send> = Box::new(x);
+    }
+    unsafe {
+        assert!(CELL == 2); // Should fail
```
danielsn, line 39: use the `expect_fail` macro

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/place.rs
avanhatt: Added

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/statement.rs
avanhatt: The conditional doesn't work as a check, because asserting the inverse in the dynamic case fails whenever the instance def is `InstanceDef::DropGlue` instead of `::Virtual`. This structure also matches how the Cranelift backend handles this case.

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/rvalue.rs
danielsn: The issue is that this code is called in a constructor, so it will always fail at CBMC time (which is sound, but not helpful).  I think what we want is to make a function whose body is `unimplemented`, and put a pointer to that function into the vtable.

danielsn (approved):

PR #409, opened 2021-08-09
Add support for manually passing options to examples.

### LINKED ISSUE #407: Look for custom configs in cases where the extracted annotations are not enough (e.g., `unwind` flag and `codegen` fail step).

bdalrhm:
### Description of changes: 

This PR adds support for manually passing options to examples in the Rust books. The dashboard now looks for configuration files under `src/tools/dashboard/configs` and prepends the specified options in those files to the top of the extracted examples. With this PR, the number of failing tests drops from 19 to 16.

### Resolved issues:

Resolves #407

### Call-outs:

This PR also fixes two encountered bugs where tests were not failing for the appropriate reasons:
  - 1 test was failing because of a timeout in the codegen step. Fix: increase timeout from 1 to 5 secs for the check and codegen steps.
  - 2 tests were failing because some command-line arguments were passed in the wrong order. Fix: reorder the arguments.

### Testing:

* How is this change tested?
  * The number of failing tests drop.
* Is this a refactor change?
  * No.

### Checklist

### DIFF
--- a/src/tools/dashboard/src/reference.rs
+++ b/src/tools/dashboard/src/reference.rs
@@ -20,6 +20,7 @@ use std::{
     path::{Path, PathBuf},
     process::{Command, Stdio},
 };
+use walkdir::WalkDir;
 
 /// Parses the chapter/section hierarchy in the markdown file specified by
 /// `summary_path` and returns a mapping from markdown files containing rust
@@ -159,16 +160,10 @@ fn extract(par_from: &Path, par_to: &Path) -> Vec<(Example, PathBuf)> {
 fn get_config_paths() -> HashSet<PathBuf> {
     let config_dir: PathBuf = ["src", "tools", "dashboard", "configs"].iter().collect();
     let mut config_paths = HashSet::new();
-    let mut stack = vec![config_dir];
-    while !stack.is_empty() {
-        let curr = stack.pop().unwrap();
-        for child in curr.read_dir().unwrap() {
-            let child = child.unwrap().path();
-            if child.is_dir() {
-                stack.push(child);
-            } else {
-                config_paths.insert(child);
-            }
+    for entry in WalkDir::new(config_dir) {
+        let entry = entry.unwrap().into_path();
+        if entry.is_file() {
+            config_paths.insert(entry);
         }
     }
     config_paths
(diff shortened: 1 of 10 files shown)


src/tools/dashboard/src/reference.rs
```diff
@@ -151,72 +155,104 @@ fn extract(par_from: &Path, par_to: &Path) -> Vec<(Example, PathBuf)> {
...
-fn prepend_text(path: &Path, text: &str) {
-    let code = fs::read_to_string(&path).unwrap();
-    let code = format!("{}\n{}", text, code);
-    fs::write(&path, code).unwrap();
+/// Returns a set of paths to the config files for examples in the Rust books.
+fn get_config_paths() -> HashSet<PathBuf> {
+    let config_dir: PathBuf = ["src", "tools", "dashboard", "configs"].iter().collect();
+    let mut config_paths = HashSet::new();
+    let mut stack = vec![config_dir];
+    while !stack.is_empty() {
```
tedinski, line 163: There's a walkdir crate and https://doc.rust-lang.org/1.2.0/std/fs/fn.walk_dir.html which seems to be marked unstable... are we able to use these? (Maybe the answer is no)

src/tools/dashboard/src/reference.rs
```diff
@@ -151,72 +155,104 @@ fn extract(par_from: &Path, par_to: &Path) -> Vec<(Example, PathBuf)> {
...
-        let code = format!("// rmc-flags: --cbmc-args --unwind 1\n{}", code);
-        fs::write(&test, code).unwrap();
+    if !config_paths.is_empty() {
+        panic!(
+            "Error: The examples corresponding to the following config files \
+             were not encountered in the pre-processing step:\n{}This is most \
+             likely because the line numbers of the config files are not in \
+             sync with the line numbers of the corresponding code blocks in \
+             the latest versions of the Rust books. Please update the line \
+             numbers of the config files and rerun the program.",
```
tedinski, line 251: Have you chatted with the group about this? Sounds like this approach might be quite flaky?

I don't have any alternative ideas off the top of my head, though. Other than trying to find some git-fu that could automatically update the line numbers somehow.

tedinski (approved):

src/tools/dashboard/src/reference.rs
bdalrhm: It looks like the `fs::walk_dir` got deprecated in favor of the crate. Replaced the code with the crate. Thanks!

src/tools/dashboard/src/reference.rs
bdalrhm: Yes. We talked about this in a previous Deep Dive. We considered automatically updating the line number, but that can introduce silent success/failure if the code-blocks change too. Nathan approved of this as a temporary solution for now, though it adds to the list of problems Adrian has to deal with in his weekly rebases.

PR #417, opened 2021-08-11
Introduce simd_shuffle intrinsic

### LINKED ISSUE #309: Codegen failure: Unsupported intrinisic simd_shuffle8
One version of RMC has shown this panic while codegen'ing the rand crate (#111):
```
thread 'rustc' panicked at 'not implemented: unsupported intrinsic: simd_shuffle8
```

With `main-153-2021-07-02` I got a different failure (#308) but I think we will eventually hit this one too.

tedinski:
### Description of changes: 

1. Add CBMC's `vector_exprt` to our gotoc representation.
2. Add support for the `simd_shuffle` instrinsics.
3. Add a test of the above.

### Resolved issues:

Resolves #309 

### Call-outs:

1. TODO: I need to create some new issues about future follow-ups (e.g. getting CBMC to expose `shuffle_vector_exprt` for us?) and possibly link to them from the code.
2. I still have a TODO in the code about the type of the array index expressions we generate. Currently, I use a signed index. Unsigned causes CBMC to object:

```
Invariant check failed
File: ../src/solvers/flattening/boolbv_mult.cpp:73 function: convert_mult
Condition: it->type() == expr.type()
Reason: multiplication operands should have same type as expression
```

I'm going to investigate this before merging. It's possible I should be picking the type in some other way, but this is difficult to diagnose because it seems to be internal to CBMC's transformations (e.g. I think we put "array index" in the symbol table, but them CBMC is transforming that into "pointer + index*size" internally, and that's where the "mult" is coming from. A guess, for now. Perplexing because it would seem to imply that the "sizeof" a type has signed type internally?)

### Testing:

* How is this change tested? Added tests.

### Checklist

### DIFF
--- a/compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs
@@ -387,7 +387,7 @@ impl Expr {
                 elems
             );
         } else {
-            unreachable!("Can't make an vector_val with non-vector target type {:?}", typ);
+            unreachable!("Can't make a vector_val with non-vector target type {:?}", typ);
         }
         expr!(Vector { elems }, typ)
     }
--- a/compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs
@@ -146,6 +146,10 @@ pub enum ExprValue {
         op: UnaryOperand,
         e: Expr,
     },
+    /// `vec_typ x = >>> {elems0, elems1 ...} <<<`
+    Vector {
+        elems: Vec<Expr>,
+    },
 }
 
 /// Binary operators. The names are the same as in the Irep representation.
@@ -371,6 +375,21 @@ impl Expr {
             unreachable!("Can't make an array_val with non-array target type {:?}", typ);
         }
         expr!(Array { elems }, typ)
+    }
+
+    pub fn vector_expr(typ: Type, elems: Vec<Expr>) -> Self {
+        if let Type::Vector { size, typ: value_typ } = typ.clone() {
+            assert_eq!(size as usize, elems.len());
+            assert!(
+                elems.iter().all(|x| x.typ == *value_typ),
+                "Vector type and value types don't match: \n{:?}\n{:?}",
+                typ,
+                elems
+            );
+        } else {
+            unreachable!("Can't make an vector_val with non-vector target type {:?}", typ);
+        }
+        expr!(Vector { elems }, typ)
     }
 
     /// `(__CPROVER_bool) >>> true/false <<<`. True/False as a single bit boolean.
--- a/compiler/rustc_codegen_llvm/src/gotoc/cbmc/irep/to_irep.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/cbmc/irep/to_irep.rs
@@ -175,7 +175,6 @@ impl ToIrep for ExprValue {
                 sub: vec![e.to_irep(mm), Expr::int_constant(*offset, Type::ssize_t()).to_irep(mm)],
                 named_sub: btree_map![],
             },
-
             ExprValue::CBoolConstant(i) => Irep {
                 id: IrepId::Constant,
                 sub: vec![],
@@ -292,6 +291,11 @@ impl ToIrep for ExprValue {
             ExprValue::UnOp { op, e } => {
                 Irep { id: op.to_irep_id(), sub: vec![e.to_irep(mm)], named_sub: btree_map![] }
             }
+            ExprValue::Vector { elems } => Irep {
+                id: IrepId::Vector,
+                sub: elems.iter().map(|x| x.to_irep(mm)).collect(),
+                named_sub: btree_map![],
+            },
         }
     }
 }
--- a/compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/intrinsic.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/intrinsic.rs
@@ -844,6 +844,7 @@ impl<'tcx> GotocCtx<'tcx> {
     ///
     /// We can't use shuffle_vector_exprt because it's not understood by the CBMC backend,
     /// it's immediately lowered by the C frontend.
+    /// Issue: https://github.com/diffblue/cbmc/issues/6297
     pub fn codegen_intrinsic_simd_shuffle(
         &mut self,
         mut fargs: Vec<Expr>,
@@ -857,9 +858,9 @@ impl<'tcx> GotocCtx<'tcx> {
         // [u32; n]: translated wrapped in a struct
         let indexes = fargs.remove(0);
 
-        // TODO: Why is this signed? Unsigned causes an invariant violation in CBMC in my tests.
-        // But it doesn't make sense why an array index type should be signed?
-        let st_rep = Type::CInteger(CIntType::SSizeT);
+        // An unsigned type here causes an invariant violation in CBMC.
+        // Issue: https://github.com/diffblue/cbmc/issues/6298
+        let st_rep = Type::ssize_t();
         let n_rep = Expr::int_constant(n, st_rep.clone());
 
         // P = indexes.expanded_map(v -> if v < N then vec1[v] else vec2[v-N])
--- a/compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/intrinsic.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/intrinsic.rs
@@ -217,6 +217,11 @@ impl<'tcx> GotocCtx<'tcx> {
             }};
         }
 
+        if let Some(stripped) = intrinsic.strip_prefix("simd_shuffle") {
+            let n: u64 = stripped.parse().unwrap();
+            return self.codegen_intrinsic_simd_shuffle(fargs, p, cbmc_ret_ty, n);
+        }
+
         match intrinsic {
             "abort" => Stmt::assert_false("abort intrinsic", loc),
             "add_with_overflow" => codegen_op_with_overflow!(add_overflow),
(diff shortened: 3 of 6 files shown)


compiler/rustc_codegen_llvm/src/gotoc/intrinsic.rs
```diff
@@ -217,6 +217,11 @@ impl<'tcx> GotocCtx<'tcx> {
             }};
         }
 
+        if let Some(stripped) = intrinsic.strip_prefix("simd_shuffle") {
```
avanhatt, line 220: Could this be integrated into the match below instead of its own code block?

```rust
i if i.strip_prefix("simd_shuffle") => ...
```
```

Comment (tedinski) [reply compiler/rustc_codegen_llvm/src/gotoc/intrinsic.rs:220]: Apparently this is  an "if let guard" and is marked as experimental currently: https://github.com/rust-lang/rfcs/blob/master/text/2294-if-let-guard.md

I'm going to leave it as-is for now.

File: compiler/rustc_codegen_llvm/src/gotoc/intrinsic.rs [hunk 2]
```diff
@@ -829,4 +835,46 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        p: &Place<'tcx>,
+        cbmc_ret_ty: Type,
+        n: u64,
+    ) -> Stmt {
+        // vector, size n: translated as vector types which cbmc treats as arrays
+        let vec1 = fargs.remove(0);
+        let vec2 = fargs.remove(0);
+        // [u32; n]: translated wrapped in a struct
+        let indexes = fargs.remove(0);
+
```
avanhatt, line 860: nit: assert `fargs` has length 3?

src/test/cbmc/SIMD/Construction/main.rs
```diff
@@ -36,4 +43,17 @@ fn main() {
...
+        assert!(x.0 == 1);
+        assert!(x.1 == 1);
+    }
+    {
+        let a = i64x4(1, 2, 3, 4);
+        let b = i64x4(5, 6, 7, 8);
+        const I: [u32; 4] = [1, 3, 5, 7];
+        let c: i64x4 = unsafe { simd_shuffle4(a, b, I) };
+        assert!(c == i64x4(2, 4, 6, 8));
+    }
```
avanhatt, line 58: What is the expected behavior on passing a negative index?

tedinski: For RMC in principle? CBMC treats these as array indexing, and it would find an array lower bound violation.

For Rust? It requires the indexes are constants, and they're typed `u32`, so it emits an error about the value being out of bounds.

compiler/rustc_codegen_llvm/src/gotoc/intrinsic.rs
tedinski: Good idea

src/test/cbmc/SIMD/Construction/main.rs
avanhatt: Makes sense! Might be useful to add a test with `// rmc-verify-fail` in that case?

src/test/cbmc/SIMD/Construction/main.rs
```diff
@@ -7,9 +7,16 @@
 #[derive(Clone, Copy, PartialEq, Eq)]
 pub struct i64x2(i64, i64);
 
+#[repr(simd)]
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct i64x4(i64, i64, i64, i64);
```
adpaco-aws, line 13: I would move `simd_shuffle` testing into its own file.

src/test/cbmc/SIMD/Construction/main.rs
tedinski: Well, the "For Rust?" one still applies: I can't get past rustc to show the CBMC bound failure, since it emits compiler errors on attempts to use negative values.

compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs
```diff
@@ -373,6 +377,21 @@ impl Expr {
...
+        if let Type::Vector { size, typ: value_typ } = typ.clone() {
+            assert_eq!(size as usize, elems.len());
+            assert!(
+                elems.iter().all(|x| x.typ == *value_typ),
+                "Vector type and value types don't match: \n{:?}\n{:?}",
+                typ,
+                elems
+            );
+        } else {
+            unreachable!("Can't make an vector_val with non-vector target type {:?}", typ);
```
adpaco-aws, line 390:
```suggestion
            unreachable!("Can't make a vector_val with non-vector target type {:?}", typ);
```

compiler/rustc_codegen_llvm/src/gotoc/cbmc/irep/to_irep.rs
```diff
@@ -292,6 +291,11 @@ impl ToIrep for ExprValue {
             ExprValue::UnOp { op, e } => {
                 Irep { id: op.to_irep_id(), sub: vec![e.to_irep(mm)], named_sub: btree_map![] }
             }
```
adpaco-aws, line 293:
```suggestion
            },
```

src/test/cbmc/SIMD/Construction/main.rs
avanhatt: Ah sorry I missed that, thanks! Somewhat remarkable, Rust also catches too-high bounds as compile time (I'm used to C-based shuffle intrinsics which tend to dramatically break at runtime):

```rust
        let a = i64x4(1, 2, 3, 4);
        let b = i64x4(5, 6, 7, 8);
        const I: [u32; 4] = [8, 9, 10, 11]; // too big
        let c: i64x4 = unsafe { simd_shuffle4(a, b, I) };
        assert!(c == i64x4(2, 4, 6, 8));
```

```
error[E0511]: invalid monomorphization of `simd_shuffle4` intrinsic: shuffle index #0 is out of bounds (limit 8)
  --> src/main.rs:63:33
   |
63 |         let c: i64x4 = unsafe { simd_shuffle4(a, b, I) };
```

compiler/rustc_codegen_llvm/src/gotoc/intrinsic.rs
```diff
@@ -3,7 +3,7 @@
 //! this module handles intrinsics
 use tracing::{debug, warn};
 
-use super::cbmc::goto_program::{BuiltinFn, Expr, Location, Stmt, Type};
+use super::cbmc::goto_program::{BuiltinFn, CIntType, Expr, Location, Stmt, Type};
```
adpaco-aws, line 6: Why do you need this?

compiler/rustc_codegen_llvm/src/gotoc/intrinsic.rs
adpaco-aws: Why not have it at the end of the match with an `starts_with` guard?

compiler/rustc_codegen_llvm/src/gotoc/intrinsic.rs
tedinski: I think I don't anymore, will fix...

compiler/rustc_codegen_llvm/src/gotoc/cbmc/irep/to_irep.rs
tedinski: I think this would be a rustfmt violation?

compiler/rustc_codegen_llvm/src/gotoc/intrinsic.rs
adpaco-aws: Still, you can match the prefix, then extract the suffix before calling the codegen function.

tedinski:
I think this has addressed everything except the nits about the match vs preceding if. I think there isn't any obviously better choice (except "if let guards", which isn't stable yet), and this style is what appears in some of the llvm codegen crate. IMO, let's just go with it.

compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs
```diff
@@ -373,6 +377,21 @@ impl Expr {
         expr!(Array { elems }, typ)
     }
 
+    pub fn vector_expr(typ: Type, elems: Vec<Expr>) -> Self {
+        if let Type::Vector { size, typ: value_typ } = typ.clone() {
```
danielsn, line 381: Why is the `clone` needed here

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/intrinsic.rs
```diff
@@ -829,4 +836,47 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        assert!(fargs.len() == 3, "simd_shuffle had unexpected arguments {:?}", fargs);
+        // vector, size n: translated as vector types which cbmc treats as arrays
+        let vec1 = fargs.remove(0);
+        let vec2 = fargs.remove(0);
+        // [u32; n]: translated wrapped in a struct
+        let indexes = fargs.remove(0);
+
+        // An unsigned type here causes an invariant violation in CBMC.
+        // Issue: https://github.com/diffblue/cbmc/issues/6298
+        let st_rep = Type::ssize_t();
```
danielsn, line 865: It might be clearer to just use `Type::ssize_t()` wherever this is used.

compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs
tedinski: The use of `value_typ` later is a partial move out of the value.

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/intrinsic.rs
tedinski: I had that thought too, but I felt like keeping it since it created a place for the comment about the CBMC issue.

Issue #419, opened 2021-08-12
Pass appropriate unwind value for examples extracted from the Rust books.


bdalrhm:
Several example configurations pass `--cbmc-args --unwind 0` to RMC. An unwind value of 0 is passed because, for some reason, Litani does not terminate/kill processes that reach the timeout value (10 seconds). Replace 0 with appropriate values when this issue gets fixed.

zhassan-aws:
@adpaco-aws is this issue still relevant?

PR #420, opened 2021-08-12
Extract examples from other books.

### LINKED ISSUE #385: Extract examples from The Nomicon and Unsafe Rust and display them in the dashboard.

bdalrhm:
### Description of changes: 

This PR adds support for extracting examples from 3 book in addition to The Rust Reference: The Rustonomicon, The Unstable Rust, and Rust by Example. With this PR, the number of examples we run jumps to 700 with 98 failing.

### Resolved issues:

Resolves #385

### Call-outs:

Several example configurations pass `--cbmc-args --unwind 0` to RMC (see #419).

### Testing:

* How is this change tested?
  * The dashboard program now extracts, runs, and displays the results of examples from the 3 books.
* Is this a refactor change?
  * To some degree. Several files were moved/renamed to account for the additional books.

### Checklist


Cargo.lock
```diff
@@ -2,6 +2,16 @@
 # It is not intended for manual editing.
 version = 3
 
+[[package]]
```
tedinski, line 5: Is this going to cause merge conflicts?

bdalrhm:
> Vague idea: If the timeout bug were fixed, could we (instead of having fragile props files) just have a sequence of possible flags to try? (e.g. try default, then try unwind 5, etc...)
> 
> Basically, trading away a little runtime for something that'd hopefully be more maintainable in the future. Just a thought.

That's a good idea. Most examples pass with an unwind value less than 5. So, it will help us reduce the number of props files. Though, we will still need them for outliers. I will run your idea by Adrian next week.

PR #422, opened 2021-08-12
Fix vtable size check for never

### LINKED ISSUE #421: Failed assertion: Correct CBMC vtable size for Never
With some combinations of traits and the `Never` type, the CBMC-time sanity check assert fails because you cannot take an object size of `void`. 

Breaking this into its own PR to land since it's coming up outside of drop work.

avanhatt:
### Description of changes: 

With some combinations of traits and the Never type, the CBMC-time sanity check assert fails because you cannot take an object size of void.

Breaking this into its own PR to land since it's coming up outside of drop work.

### Resolved issues:

Resolves #421

### Call-outs:

Working on getting a small example now, but a little difficulty since this happens in external crate code.

### Testing:

* How is this change tested?

Existing tests, working on minimal example.

* Is this a refactor change?

No.

### Checklist


compiler/rustc_codegen_llvm/src/gotoc/rvalue.rs
```diff
@@ -766,10 +766,17 @@ impl<'tcx> GotocCtx<'tcx> {
         // Insert a CBMC-time size check, roughly:
         //     <Ty> local_temp = nondet();
         //     assert(__CPROVER_OBJECT_SIZE(&local_temp) == vt_size);
-        let temp_var = self.gen_temp_variable(ty, Location::none()).to_expr();
+        let temp_var = self.gen_temp_variable(ty.clone(), Location::none()).to_expr();
         let decl = Stmt::decl(temp_var.clone(), None, Location::none());
-        let check = Expr::eq(Expr::object_size(temp_var.address_of()), vt_size.clone());
-        let assert_msg = format!("Correct CBMC vtable size for {:?}", operand_type.kind());
+        let cbmc_size = if ty.clone().is_empty() {
```
danielsn, line 771: This clone should be unneccessary

compiler/rustc_codegen_llvm/src/gotoc/rvalue.rs
```diff
@@ -766,10 +766,17 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            // CBMC errors on passing a pointer to void to __CPROVER_OBJECT_SIZE, so just pass 0
```
danielsn, line 772: Are we guaranteed this is correct? Or is this a temporary hack?

avanhatt: Added an assert for the `Never` case and a link showing that we do know the size is 0. This way, will get a more detailed error if this fails again for a different type.

danielsn (approved):

PR #423, opened 2021-08-12
Refactor backend code into its own file

danielsn:
### Description of changes: 

mod.rs was a mess. This moves the remaining code into properly named files.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/compiler/rustc_codegen_llvm/src/gotoc/function.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/function.rs
@@ -9,6 +9,28 @@ use rustc_middle::mir::{HasLocalDecls, Local};
 use rustc_middle::ty::{self, Instance, TyS};
 use tracing::{debug, warn};
 
+/// Utility to skip functions that can't currently be successfully codgenned.
+impl<'tcx> GotocCtx<'tcx> {
+    fn should_skip_current_fn(&self) -> bool {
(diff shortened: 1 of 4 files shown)


compiler/rustc_codegen_llvm/src/gotoc/function.rs
```diff
@@ -9,6 +9,28 @@ use rustc_middle::mir::{HasLocalDecls, Local};
 use rustc_middle::ty::{self, Instance, TyS};
 use tracing::{debug, warn};
 
+/// Utility to skip functions that can't currently be successfully codgenned.
```
avanhatt, line 12: I think I agree with Ted that this makes more sense close to where it's used: whether we skip a function is a property of codegening the crate, not the function itself?

tedinski (approved):

compiler/rustc_codegen_llvm/src/gotoc/function.rs
danielsn: Its actually used in function.rs. But I agree that that's maybe not where you would expect to find it.

danielsn: https://github.com/model-checking/rmc/issues/424

PR #430, opened 2021-08-17
Full crate name for allocations (support multiple crate versions)

### LINKED ISSUE #429: RMC should, like Cargo, allow multiple version of dependencies
A few times now, we've seen Cargo RMC fail when dependent crates rely on two different versions of some dependency. You can work around this by deleting one of the resulting GotoC outputs, but this isn't sound if the two versions actually have semantic differences.

We should, like Cargo, figure out how to use the version itself in the mangled symbol name and resolve the right versions of functions. 

https://stephencoakley.com/2019/04/24/how-rust-solved-dependency-hell

avanhatt:
### Description of changes: 

When building with Cargo RMC, you can run into name conflicts with global allocations if two different versions of a crate are pulled into dependencies.

We should use the full crate name, including a mangled versioning info string, when naming allocations.

Example:

Before: 
Name: `getrandom::alloc174`
Error: 
```rust
error: conflicting array sizes for variable 'getrandom::alloc174'
old definition in module '' 
struct getrandom::alloc174::struct {
  unsigned char [10] 0;
}
new definition in module '' 
struct getrandom::alloc174::struct {
  unsigned char [15] 0;
}

reason for conflict at #this.0: array types differ

unsigned char [10]
unsigned char [15]
```

After: `1vemgcjvp3r154gr::getrandom::alloc174`

### Resolved issues:

Resolves #429

### Testing:

* How is this change tested?

Manually running on a (private) codebase with this issue.

* Is this a refactor change?

No.

### Checklist


compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
```diff
@@ -70,10 +71,21 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    /// The short crate name without versioning information.
+    pub fn short_crate_name(&self) -> String {
         self.tcx.crate_name(LOCAL_CRATE).to_string()
     }
 
+    /// The full crate name should use the Codegen Unit builder to include full name resolution,
+    /// for example, the versioning information if a build requires two different versions
+    /// of the same crate.
+    pub fn full_crate_name(&self) -> String {
+        let name_builder = &mut CodegenUnitNameBuilder::new(self.tcx);
```
danielsn, line 83: Do we want to cache this?

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
```diff
@@ -70,10 +71,21 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        let info =
+            name_builder.build_cgu_name(LOCAL_CRATE, &[] as &[String; 0], None as Option<String>);
```
danielsn, line 85: Why are the casts needed

avanhatt: Oops, they were needed for underconstrained generics, but now with the format I think they're not needed.

avanhatt: Whoops no it is needed:
```rust
   --> compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs:288:42
    |
288 |         CodegenUnitNameBuilder::new(tcx).build_cgu_name(LOCAL_CRATE, &[], None),
    |                                          ^^^^^^^^^^^^^^ cannot infer type for type parameter `S` declared on the associated function `build_cgu_name`
```

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
avanhatt: Added as a field, thoughts?

.gitignore
```diff
@@ -80,3 +80,5 @@ src/test/rustdoc-gui/src/**.lock
 /.ninja_deps
 /.ninja_log
 /src/test/dashboard
+**Cargo.lock
```
danielsn, line 83: Do you need `**`

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
```diff
@@ -54,13 +57,15 @@ pub struct GotocCtx<'tcx> {
 impl<'tcx> GotocCtx<'tcx> {
     pub fn new(tcx: TyCtxt<'tcx>) -> GotocCtx<'tcx> {
         let (thks, fhks) = type_and_fn_hooks();
+        let crate_name = full_crate_name(tcx);
```
danielsn, line 60: Why not just call this `full_crate_name`

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
```diff
@@ -70,10 +75,16 @@ impl<'tcx> GotocCtx<'tcx> {
...
 impl<'tcx> GotocCtx<'tcx> {
-    pub fn crate_name(&self) -> String {
+    /// The short crate name without versioning information.
+    pub fn short_crate_name(&self) -> String {
         self.tcx.crate_name(LOCAL_CRATE).to_string()
     }
 
+    /// The full crate name including versioning info
+    pub fn full_crate_name(&self) -> String {
+        self.full_crate_name.clone()
```
danielsn, line 85: Do we want to clone here? Or should we return `&str`?

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
```diff
@@ -268,6 +279,21 @@ impl MetadataLoader for GotocMetadataLoader {
     }
 }
 
+/// The full crate name should use the Codegen Unit builder to include full name resolution,
```
danielsn, line 282: Is this where this function belongs?

src/test/rmc-dependency-test/dependency3/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+pub fn get_int() -> i32 {
+    // Use a constant to force an MIR GlobalAllocation::Memory
+    // Use a non-i32 so there will be a conflict between this
+    // version and the other version.
+    let one = &(1 as i8);
```
danielsn, line 8: It also matters that this is 1 and not 0, right?

src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
danielsn, line 1: https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
```diff
@@ -0,0 +1,39 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+# Test for platform
+platform=`uname -sp`
+if [[ $platform != "Linux x86_64" ]]; then
+  echo "Codegen script only works on Linux x86 platform"
```
danielsn, line 8: What happens on mac?

src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
```diff
@@ -0,0 +1,39 @@
...
+  exit 0
+fi
+
+# Log output
+LOG="/tmp/dependency_test_log.txt"
+
+# Compile crates with RMC backend
+cd $(dirname $0)
+rm -rf build
```
danielsn, line 17: does `cargo rmc` not work

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
avanhatt: I could see an argument for either, where it's used for global names, or `compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/utils/names.rs`. Maybe the right thing is to move both the global name stuff and this to `names.rs`, thoughts?

src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
avanhatt: it fails with:
```
rmc (dependencies) $ ./src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
ERROR: Unexpected number of json outputs: 5
```

It seems like the script for it assumes a single symbol table output, but here we get one per dependency and main.

src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
avanhatt: With the given command, it fails to find `std`:

```
error[E0463]: can't find crate for `std`
  |
  = note: the `x86_64-unknown-linux-gnu` target may not be installed
  = help: consider downloading the target with `rustup target add x86_64-unknown-linux-gnu`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`
```

If you try to build std (using nightly), you hit ICE's within building std. This block to skip is copied from the Firecrackers regression test.

src/test/rmc-dependency-test/dependency3/src/lib.rs
avanhatt: Added comment

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
avanhatt: Changed.

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
avanhatt: Changed.

.gitignore
avanhatt: Changed

src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
avanhatt: Changed!

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/typ.rs
```diff
@@ -374,7 +374,22 @@ impl<'tcx> GotocCtx<'tcx> {
         // code.  See the implementation of pretty_print_region on line 1720 in
         // compiler/rustc_middle/src/ty/print/pretty.rs.
         let name = name.replace(" + \'static", "").replace("\'static ", "");
-        name
+
+        // Crate resolution: mangled names need to be distinct across different versions
+        // of the same crate that could be pulled in by dependencies. However, RMC's
+        // treatment of FFI C calls asssumes that we generate the same name for structs
+        // as the C name, so don't mangle in that case.
+        let is_repr_c_adt = match *t.kind() {
```
danielsn, line 382: This might be a useful helper function

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/typ.rs
```diff
@@ -374,7 +374,22 @@ impl<'tcx> GotocCtx<'tcx> {
         // code.  See the implementation of pretty_print_region on line 1720 in
         // compiler/rustc_middle/src/ty/print/pretty.rs.
         let name = name.replace(" + \'static", "").replace("\'static ", "");
-        name
+
+        // Crate resolution: mangled names need to be distinct across different versions
+        // of the same crate that could be pulled in by dependencies. However, RMC's
+        // treatment of FFI C calls asssumes that we generate the same name for structs
+        // as the C name, so don't mangle in that case.
```
danielsn, line 381: This is a hack. We should note it as such, and have a tracking issue to do the right thing here.

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
```diff
@@ -70,10 +74,16 @@ impl<'tcx> GotocCtx<'tcx> {
...
 impl<'tcx> GotocCtx<'tcx> {
-    pub fn crate_name(&self) -> String {
+    /// The short crate name without versioning information.
+    pub fn short_crate_name(&self) -> String {
         self.tcx.crate_name(LOCAL_CRATE).to_string()
     }
 
+    /// The full crate name including versioning info
+    pub fn full_crate_name(&self) -> &str {
+        &self.full_crate_name
```
danielsn, line 84: Why do we cache this one and not the other one?

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
danielsn: Yeah, I think that makes sense

src/test/rmc-dependency-test/diamond-dependency/dependency3/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+
+pub struct Foo {
+    x: i32,
+    // This field only in this version.
+    y: i32,
+}
+
+// Export a function that takes a struct type which differs between this crate 
+// and the other vesion.
+pub fn take_foo(foo: &Foo) {
```
danielsn, line 12: Would there be an advantage to returning `foo.x + foo.y` so that upstream tests could assert on that?

src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
danielsn: Lets file a ticket for that

src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
```diff
@@ -0,0 +1,39 @@
...
+CARGO_TARGET_DIR=build RUST_BACKTRACE=1 RUSTFLAGS="-Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo build --target x86_64-unknown-linux-gnu 2> $LOG
```
danielsn, line 18: What happens if we use the osx target here and build on mac

danielsn (approved):
Approved modulo comments.

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/context/goto_ctx.rs
avanhatt: One we pull directly from MIR, the other will build from two parts. Maybe with the `to_string` in the first case it's the same, though?

src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
avanhatt: https://github.com/model-checking/rmc/issues/451

src/test/rmc-dependency-test/diamond-dependency/run-dependency-test.sh
avanhatt: IIRC it requires rebuilding the std lib on mac, which then fails. We can look at this again together.

src/test/rmc-dependency-test/diamond-dependency/dependency3/src/lib.rs
avanhatt: Good idea, thanks

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/typ.rs
avanhatt: Good idea, moved.

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/typ.rs
avanhatt: Filed and adding

PR #447, opened 2021-08-23
Add support for modifying examples in the Rust books.

### LINKED ISSUE #408: Look for custom assertions to add to the examples.

bdalrhm:
### Description of changes: 

This PR adds support for modifying examples in the Rust books by adding a diff file with the same name as the example in `src/tools/dashboard/configs`.

### Resolved issues:

Resolves #408

### Call-outs:

To avoid duplicating code, the dashboard now depends on some of `rustdoc`'s internal API, some of which were made public as necessary.

### Testing:

* How is this change tested?
  * Terminal and Litani dashboard still work as expected.
* Is this a refactor change?
  * No.
### Checklist

### DIFF
--- a/src/librustdoc/doctest.rs
+++ b/src/librustdoc/doctest.rs
@@ -35,7 +35,7 @@ use crate::lint::init_lints;
 use crate::passes::span_of_attrs;
 
 #[derive(Clone, Default)]
-crate struct TestOptions {
+pub struct TestOptions {
     /// Whether to disable the default `extern crate my_crate;` when creating doctests.
     crate no_crate_inject: bool,
     /// Whether to emit compilation warnings when compiling doctests. Setting this will suppress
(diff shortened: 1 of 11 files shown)


adpaco-aws (approved):

src/librustdoc/doctest.rs
```diff
@@ -35,7 +35,7 @@ use crate::lint::init_lints;
 use crate::passes::span_of_attrs;
 
 #[derive(Clone, Default)]
-crate struct TestOptions {
+pub struct TestOptions {
```
tedinski, line 38: I think my only concern is the potential for future conflicts from stuff like this, but if it's fine with your Adrian, then it's fine with me.

tedinski (approved):

src/librustdoc/doctest.rs
adpaco-aws: Good point, opened #461 to keep track of this. @bdalrhm had linking issues when not doing this and we agreed to change it in order to have the PR online.

PR #456, opened 2021-08-31
Initial hello world for RMC documentation

tedinski:
### Description of changes: 

This introduces:

1. A candidate CI process for building/publishing RMC documentation.
2. "Hello world" and nothing more to push there.

### Resolved issues:

N/A

### Call-outs:

1. The GitHub Actions documentations is deeply confusing about permissions. This PR will be able to test the building of the documentation, **but not the publishing of it. We may well have to merge it to find out if that part works or not!** (I have several hypotheses about what's going to happen when we try, none of them confirmed or refuted by any documentation I can find.)
2. This just tacks the documentation build onto the end of the `ubuntu-20.04` build, instead of as its own workflow. This is because we want to (not in this PR, but in a later PR) use RMC to test the code examples in the documentation, as well as build and publish Abdal's dashboard. This means we need RMC built and ready to run, so we might as well piggy back on an existing build. (Later, when we have releases of RMC available to download, we can probably separate documentation publishing into its own workflow, instead of piggybacking like this.)

### Testing:

* How is this change tested? We'll have to merge it to see! :(

* Is this a refactor change? -

### Checklist

### DIFF
--- a/.github/workflows/copyright.yml
+++ b/.github/workflows/copyright.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Get paths for files added
         id: git-diff
         run: |
-          ignore='(.toml|.diff|.md|.props|expected|ignore|gitignore)$'
+          ignore='(.diff|.md|.props|expected|ignore|gitignore)$'
           files=$(git diff --ignore-submodules=all --name-only --diff-filter=A ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} | grep -v -E $ignore | xargs)
           echo "::set-output name=paths::$files"
 
--- a/.github/workflows/copyright.yml
+++ b/.github/workflows/copyright.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Get paths for files added
         id: git-diff
         run: |
-          ignore='(.diff|.md|.props|expected|ignore|gitignore)$'
+          ignore='(.toml|.diff|.md|.props|expected|ignore|gitignore)$'
           files=$(git diff --ignore-submodules=all --name-only --diff-filter=A ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} | grep -v -E $ignore | xargs)
           echo "::set-output name=paths::$files"
 
--- a/.github/workflows/rmc.yml
+++ b/.github/workflows/rmc.yml
@@ -46,3 +46,16 @@ jobs:
 
       - name: Execute RMC regression
         run: ./scripts/rmc-regression.sh
+
+      # On one OS only, build the documentation, too.
+      - name: Build Documentation
+        if: ${{ matrix.os == 'ubuntu-20.04' }}
+        run: ./rmc-docs/build-docs.sh
+
+      # When we're pushed to main branch, only then actually publish the docs.
+      - name: Publish Documentation
+        if: ${{ matrix.os == 'ubuntu-20.04' && github.event_name == 'push' && startsWith('refs/heads/main', github.ref) }}
+        uses: JamesIves/github-pages-deploy-action@4.1.4
+        with:
+          branch: gh-pages
+          folder: rmc-docs/book/
--- a/rmc-docs/README.md
+++ b/rmc-docs/README.md
@@ -1,0 +1,10 @@
+## RMC documentation development
+
+A good trick when developing RMC on a remote machine is to SSH forward to test documentation changes.
+
+```
+ssh -t -L 3000:127.0.0.1:3000 rmc-host 'cd rmc/rmc-docs/ && ./mdbook serve'
+```
+
+This command will connect to `rmc-host` where it assumes RMC is checked out in `rmc/` and the documentation has been build once successfully.
+It will automatically detect changes to the docs and rebuild, allowing you to quickly refresh in your browser when you visit: `http://127.0.0.1:3000/`
--- a/rmc-docs/README.md
+++ b/rmc-docs/README.md
@@ -6,5 +6,5 @@ A good trick when developing RMC on a remote machine is to SSH forward to test d
 ssh -t -L 3000:127.0.0.1:3000 rmc-host 'cd rmc/rmc-docs/ && ./mdbook serve'
 ```
 
-This command will connect to `rmc-host` where it assumes RMC is checked out in `rmc/` and the documentation has been build once successfully.
-It will automatically detect changes to the docs and rebuild, allowing you to quickly refresh in your browser when you visit: `http://127.0.0.1:3000/`
+This command will connect to `rmc-host` where it assumes RMC is checked out in `rmc/` and the documentation has been built once successfully.
+It will automatically detect changes to the docs and rebuild, allowing you to quickly refresh in your local browser when you visit: `http://127.0.0.1:3000/`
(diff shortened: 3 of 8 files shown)


.github/workflows/copyright.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Get paths for files added
         id: git-diff
         run: |
-          ignore='(.diff|.md|.props|expected|ignore|gitignore)$'
+          ignore='(.toml|.diff|.md|.props|expected|ignore|gitignore)$'
```
celinval, line 18: Can we actually ignore .toml files?

.github/workflows/rmc.yml
```diff
@@ -46,3 +46,16 @@ jobs:
...
+      - name: Build Documentation
+        if: ${{ matrix.os == 'ubuntu-20.04' }}
+        run: ./rmc-docs/build-docs.sh
+
+      # When we're pushed to main branch, only then actually publish the docs.
+      - name: Publish Documentation
+        if: ${{ matrix.os == 'ubuntu-20.04' && github.event_name == 'push' && startsWith('refs/heads/main', github.ref) }}
+        uses: JamesIves/github-pages-deploy-action@4.1.4
+        with:
+          branch: gh-pages
```
celinval, line 60: Is there any other way to publish github doc? In my previous team, this branch blew up once people added images to the doc. Another possibility is that we could also add a check here to make sure we are not pushing binary files.

.github/workflows/rmc.yml
```diff
@@ -46,3 +46,16 @@ jobs:
...
       - name: Execute RMC regression
         run: ./scripts/rmc-regression.sh
+
+      # On one OS only, build the documentation, too.
+      - name: Build Documentation
+        if: ${{ matrix.os == 'ubuntu-20.04' }}
+        run: ./rmc-docs/build-docs.sh
+
+      # When we're pushed to main branch, only then actually publish the docs.
+      - name: Publish Documentation
```
celinval, line 56: This is great! Less manual work.

rmc-docs/build-docs.sh
```diff
@@ -0,0 +1,26 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+set -eu
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd $SCRIPT_DIR
+
+# Download mdbook release (vs spending time building it via cargo install)
+FILE="mdbook-v0.4.12-x86_64-unknown-linux-gnu.tar.gz"
```
celinval, line 11: nit: can you add a md5sum check here to make sure we have downloaded the correct file?

rmc-docs/build-docs.sh
```diff
@@ -0,0 +1,26 @@
...
+URL="https://github.com/rust-lang/mdBook/releases/download/v0.4.12/$FILE"
+if [ ! -x mdbook ]; then
+    wget -O "$FILE" "$URL"
```
celinval, line 14: Can we download this in a folder outside of the script directory?

rmc-docs/src/getting-started.md
```diff
@@ -0,0 +1,9 @@
+# Getting started with RMC
+
+Hello, World!
+
+```
```
celinval, line 5:
````
```rust
````

**Review by celinval — COMMENTED** [review]
Thanks for doing this!

File: rmc-docs/build-docs.sh (hunk 5 above)
Comment (tedinski) [reply rmc-docs/build-docs.sh:14]: There isn't really anyplace better to put it that I can think of.

File: .github/workflows/rmc.yml (hunk 2 above)
Comment (tedinski) [reply .github/workflows/rmc.yml:60]: The only binary files are the fonts from the mdbook theme. These should change rarely, if ever, and only with changes the to the version of `mdbook` we're using.

I think this will be fine, but the custom action does have a `single-commit` option that ensures the `gh-pages` branch has history wiped out automatically. I don't plan on turning this on for now, but if this becomes a problem in the future, we could switch to using that.

File: rmc-docs/build-docs.sh (hunk 5 above)
Comment (celinval) [reply rmc-docs/build-docs.sh]: You could put it in the current directory, temp or take the directory as a parameter. It's not a blocker but a nice to have. It keeps the repository clean.

File: .github/workflows/rmc.yml (hunk 2 above)
Comment (celinval) [reply .github/workflows/rmc.yml:60]: That's fair. Thanks for checking.

File: rmc-docs/build-docs.sh [hunk 7]
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+set -eu
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
zhassan-aws, line 7: Minor: can use `>&/dev/null` as a shorthand for `>/dev/null 2>&1`.

rmc-docs/README.md
```diff
@@ -0,0 +1,10 @@
+## RMC documentation development
+
+A good trick when developing RMC on a remote machine is to SSH forward to test documentation changes.
+
+```
+ssh -t -L 3000:127.0.0.1:3000 rmc-host 'cd rmc/rmc-docs/ && ./mdbook serve'
+```
+
+This command will connect to `rmc-host` where it assumes RMC is checked out in `rmc/` and the documentation has been build once successfully.
```
zhassan-aws, line 9: build -> built

.github/workflows/rmc.yml
danielsn: Should we branch protect that branch, so only actions can update it

rmc-docs/book.toml
```diff
@@ -0,0 +1,15 @@
...
+authors = ["RMC Developers"]
+src = "src"
+language = "en"
+multilingual = false
+
+[output.html]
+site-url = "/rmc/"
+git-repository-url = "https://github.com/model-checking/rmc"
+# If we get a stable default branch, we can use this feature, but HEAD doesn't work
+#edit-url-template = "https://github.com/model-checking/rmc/edit/HEAD/rmc-docs/{path}"
```
danielsn, line 15: uncomment this

rmc-docs/build-docs.sh
```diff
@@ -0,0 +1,28 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+set -eu
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+cd $SCRIPT_DIR
+
+# Download mdbook release (vs spending time building it via cargo install)
+FILE="mdbook-v0.4.12-x86_64-unknown-linux-gnu.tar.gz"
+URL="https://github.com/rust-lang/mdBook/releases/download/v0.4.12/$FILE"
```
danielsn, line 12: is there a good way to keep up to date with new releases

danielsn: there was a suggestion to use the mdbook action

danielsn (approved):

PR #463, opened 2021-08-31
Codegen unimplemented for missing drop function

avanhatt:
### Description of changes: 

For codegening unimplemented drop functions (which we need when skipping std lib functions), we should generate a pointer to a function with unimplemented instead of calling it directly.

This is blocking the inline Firecracker balloon proof harness: https://github.com/avanhatt/firecracker/blob/add-simple-rmc-harnesses/rmc.sh

### Resolved issues:

Resolves https://github.com/model-checking/rmc/issues/462

### Testing:

* How is this change tested?

Existing regressions.

* Is this a refactor change?

No.

### Checklist


tedinski (approved):
I'm good with this as-is, but there isn't by any chance an easy test case or something for this? Or at least can you comment on what used to fail (and presumably now succeeds)?

avanhatt:
> I'm good with this as-is, but there isn't by any chance an easy test case or something for this? Or at least can you comment on what used to fail (and presumably now succeeds)?

Good point, right now the only reproduction is a proof harness on top of Firecracker, but I will try to minimize it into a smaller test case.

Issue #254, opened 2021-06-22
Training material


danielsn:
Infra:

* [X] Tools and CI "Hello World" for the documentation #456 
* [ ] Extracting and testing the examples from the documentation #635

Docs:

* [X] Getting started guide #471
* [X] Installation guide #471
* [X] Comparison with other tools initial documentation #471
* [X] Using RMC on a single file #525
* [X] Using cargo-rmc #554
* [X] Developer documentation #572
* Migration of CBMC tutorial docs to RMC
  * [X] #474
  * [X] #481
  * [X] #494
  * [X] #525

tedinski:
Here's the work for this issue:

Infra:

* [ ] Tools and CI "Hello World" for the documentation #456 
* [ ] Extracting and testing the examples from the documentation

Docs:

* [ ] Getting started guide
* [ ] Installation guide
* [ ] Comparison with other tools initial documentation
* [ ] Using RMC on a single file
* [ ] Using cargo-rmc
* [ ] Developer documentation
* [ ] Migration of CBMC tutorial docs to RMC

PR #481, opened 2021-09-13
Add second section of the tutorial

tedinski:
### Description of changes: 

This adds section 2/3 of the currently planned RMC tutorial.

### Resolved issues:

Towards #254 

### Call-outs:

* I did not include very extensive pointer material, it seems a lot less natural to talk about raw pointers in Rust. So I reduced that to a subsection of the bounds checking material.

### Testing:

* How is this change tested? Each example in the cargo package fails RMC.

### Checklist
- [n/a] Methods or procedures are documented
- [n/a] Regression or unit tests are included, or existing tests cover the modified code


rmc-docs/src/tutorial-kinds-of-failure.md
```diff
@@ -0,0 +1,161 @@
...
+Consider a different variant on the above function:
+
+```rust
+fn get_wrapped(i: usize, a: &[u32]) -> u32 {
+    return a[i % a.len()];
+}
+```
+
+We've corrected the out-of-bounds access, but now we've omitted the "base case": what to return on an empty list.
+RMC will spot this not as a bound error, but as a mathematical error: on an empty list the modulus operator (`%`) will cause a division by zero.
```
nchong-at-aws, line 84: Nice!

rmc-docs/src/tutorial-kinds-of-failure.md
```diff
@@ -0,0 +1,161 @@
...
+
+Notice the two failures: the Rust-inserted overflow check (`simple_addition.assertion.1`) and RMC's explicit overflow check (`simple_addition.overflow.1`).
+
+> **NOTE:** You could attempt to fix this issue by using Rust's alternative mathematical functions with explicit overflow behavior.
+For instance, instead of `a + b` write `a.wrapping_add(b)`.
+>
+> However, [at the present time](https://github.com/model-checking/rmc/issues/480), while this disables the dynamic assertion that Rust inserts, it does not disable the additional RMC overflow check.
+> As a result, this currently still fails in RMC.
+
+</quote>
```
nchong-at-aws, line 118: ?

rmc-docs/src/tutorial-kinds-of-failure.md
```diff
@@ -0,0 +1,161 @@
...
+It's possible RMC may be extended in the future to find such issues.
+
+2. Rust type invariants.
+For example, it's undefined behavior in Rust to produce a value of type `bool` that isn't `0` or `1`.
+RMC will not spot this error (in presumably unsafe code), yet.
+
+3. Fully generic functions.
+To write a proof harness and call functions, they must be fully "monomorphized."
+This means we can't currently check a generic function (`foo<T>`) generically.
+Proof harnesses have to be written specializing type parameters (`T`) to concrete types (e.g. `u32`), and check those instead.
```
nchong-at-aws, line 150: I'd prefer to link to the dashboard so this information doesn't become stale

rmc-docs/src/SUMMARY.md
```diff
@@ -6,8 +6,11 @@
   - [RMC on a single file]()
   - [RMC on a crate]()
   - [Debugging failures]()
-  
+  - [Debugging non-termination]()
```
zhassan-aws, line 9: Is this section coming in the next batch? If so, should we take it out from this PR?

rmc-docs/src/SUMMARY.md
```diff
@@ -6,8 +6,11 @@
...
+
 - [RMC tutorial](./rmc-tutorial.md)
   - [First steps with RMC](./tutorial-first-steps.md)
+  - [Failures that RMC can spot](./tutorial-kinds-of-failure.md)
+  - [Loops, unwinding, and bounds]()
```
zhassan-aws, line 14: Same for this subsection.

rmc-docs/src/SUMMARY.md
tedinski: mdbook supports "outlining" like this as a specifically added feature. Seems nice to use, and it makes the planned structure visible.

rmc-docs/src/tutorial-kinds-of-failure.md
tedinski: I tried to stick to bigger ideas here than what the dashboard really shows. Set people's expectations appropriately.

rmc-docs/src/tutorial-kinds-of-failure.md
```diff
@@ -133,6 +152,33 @@ RMC immediately spots the bug in the above code.
...
+```
+
+But if you naively try this (try it!), you'll find a new underflow error: `high - low` might result in a negative number, but has type `u32`.
+Hence, the need to add an assumption that would make that impossible.
+(Adding an assumption, though, means there's a new way to "use it wrong." Perhaps we'd like to avoid that!)
+
+After that, you might wonder how to "prove your new implementation correct."
+After all, what does "correct" even mean?
+Often we're using a good approximation of correct, such as the equivalence of two implementations (often one much "simpler" than the other somehow).
+Here's one possible asssertion to make that obvious:
```
zhassan-aws, line 171: assertion has an extra 's'

zhassan-aws (approved):
Great examples, @tedinski!

rmc-docs/src/tutorial-kinds-of-failure.md
nchong-at-aws: Ok, in that case, could we make the statement "does not currently check"?

PR #486, opened 2021-09-14
Partial support for Generator type

avanhatt:
### Description of changes: 

Rust added a new Generator type to support coroutines. This patch lets us codegen code that has the Generator type somewhere, but does not use it in paths that are being verified. This helps unblock crates that depend on the popular Tokio crate. Full support for Generator tracked in https://github.com/model-checking/rmc/issues/416.

### Resolved issues:

Resolves https://github.com/model-checking/rmc/issues/459.

### Call-outs:

- The `Generator` type is a global codegen item, but it does not have a function signature, so we need to make the function signature on our context object optional. 
- This patch includes more helpful debugging assertions in places.

### Testing:

* How is this change tested?

New unit test.

* Is this a refactor change?

Not really.

### Checklist


tedinski (approved):
Like the error message improvements.

Issue #485, opened 2021-09-14
RMC should use rustc monomorphization as is


celinval:
As a short term solution for #295, we're adding a hook to rustc monomorphization to skip functions that RMC has some sort of special handling for them. Conceptually, these are orthogonal changes and we should be able to decouple them. We have also added a special check to skip generating the generation of the runtime start function.

Is this a breaking change? No

Tasks:
- [x] Remove is_rmc check that skips generating the start function.
- [ ] Remove skip_monormophize query.

PR #495, opened 2021-09-17
#295: Use upstream version of the monomorphizer 

### LINKED ISSUE #295: Soundness: Monomorphization
Like C++, Rust handles generic types by “monomorphizing” the code that uses them.  The Rust compiler performs monomorphization in the llvm backend, which we do not have access to in RMC. We solved this by taking a fork of the llvm monomorphizer, but there is no guarantee that the RMC monomorphizer will stay in sync with the standard one from the Rust compiler

## Likelihood

We do not know of any bugs that are currently caused by this, but divergence is likely to occur as the Rust compiler evolves.

## Mitigation

- [ ] We have regression tests that use generic types.

## Path to soundness

- [ ] Work with the Rust compiler team to move monomorphization into a generic module we can use.

celinval:
### Description of changes: 

This is a short term solution to allow us to consume the upstream monomorphizer instead of keeping a full copy on the gotoc codegen. The issue #485 tracks a more permanent solution that will consume a clean version of the monomorphizer.

### Resolved issues:

Resolves #295

### Call-outs:

The changes to rustc files are temporary. I tried to minimize them as much as possible to avoid merge conflicts when synchronizing with rustc repository.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


tedinski:
A few questions to confirm my understanding:

1. We're switching from our custom monomorphizer to the usual one simply by not overriding the functions in `providers` anymore, correct?
2. The changes we're making are:
  1. Skipping monomorphizing for function we hook
  2. Something about entry points I don't quite understand: we should add a comment there about why we're doing that change.

Quick Q: It looks like we have an LLVM implementation of the skip function that checks if we're RMC before looking at hooks. Is there a reason we can't have the default LLVM "return false" provider be overridden with the RMC one? Is there not a place to override the provider or am I misunderstanding something?

celinval:
> A few questions to confirm my understanding:
> 
>     1. We're switching from our custom monomorphizer to the usual one simply by not overriding the functions in `providers` anymore, correct?

Yes

>     2. The changes we're making are:
> 
>     3. Skipping monomorphizing for function we hook

Correct
>     4. Something about entry points I don't quite understand: we should add a comment there about why we're doing that change.

I can definitely do that. We are basically skipping the rust runtime configuration. The monomorphizer is responsible for generating the actual program entry point which we don't use and we actually fail codegen.

> Quick Q: It looks like we have an LLVM implementation of the skip function that checks if we're RMC before looking at hooks. Is there a reason we can't have the default LLVM "return false" provider be overridden with the RMC one? Is there not a place to override the provider or am I misunderstanding something?

The default implementation of skip_monomorphization always returns false (compiler/rustc_codegen_ssa/src/base.rs). I think the confusion comes from the fact that the gotoc codegen is currently inside the llvm codegen crate. I think you do have a point that I don't need to check if is_rmc inside hooks::skip_monomorphization. That is a redundant check.

tedinski (approved):

PR #496, opened 2021-09-17
Use stable hash as unique identifier for types, instead of unstable human-readable names

### LINKED ISSUE #300: Soundness: Type naming relies on pretty-printer
CBMC has a strict typing semantics: each type must be identified by a unique name.  Structurally identical types are not considered identical for CBMC purposes: you must transition between them using an explicit byte_extract.  We currently get the names for Rust types using the Rust pretty-printer. For the most part, this works well. However, in some cases the pretty printer elides necessary data, for e.g. replacing `[i32; 8]` with `[i32, _]`.  This can spuriously cause structurally identical types to appear to differ; more worryingly, it can also cause unexpected type aliasing if both `[i32; 8]` and `[i32; 4]` are reduced to `[i32, _]`.

## Likelihood:

This issue does not occur on most codebases. However, there are known failing regression tests in the RMC test suite where this issue does occur.

## Mitigation: 

* We have special cased pretty-printing of array types to avoid the problem described above
* More generally, the GOTO type-checker we implemented has been very effective in catching these cases, and raising a warning to the user.

## Path to soundness:

Replace the default Rust pretty-printer with a scheme that is guaranteed to produce canonical names.

danielsn:
### Description of changes: 

### Resolved issues:

Resolves #300 

### Call-outs:

Although this sets the `pretty_name` in the symbol table, it seems that `--dump-c` doesn't actually use that.  This will make debugging harder.

### Testing:

* How is this change tested? Existing reg tests, plus a new test that failed before the change and passes now.

### Checklist

### DIFF
--- a/compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/symbol.rs
+++ b/compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/symbol.rs
@@ -100,11 +100,11 @@ impl Symbol {
 
     /// The symbol that defines the type of the struct or union.
     /// For a struct foo this is the symbol "tag-foo" that maps to the type struct foo.
-    pub fn aggr_ty(t: Type) -> Symbol {
+    pub fn aggr_ty(t: Type, pretty_name: Option<String>) -> Symbol {
         //TODO take location
         let base_name = t.tag().unwrap().to_string();
         let name = aggr_name(&base_name);
-        Symbol::new(name, Location::none(), t, SymbolValues::None, Some(base_name), None)
+        Symbol::new(name, Location::none(), t, SymbolValues::None, Some(base_name), pretty_name)
             .with_is_type(true)
     }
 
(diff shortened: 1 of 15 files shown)


compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/utils/utils.rs
```diff
@@ -132,16 +132,18 @@ impl Type {
         // can be reimported.
         // If we don't, we spuriously fail. https://github.com/model-checking/rmc/issues/113
         // TODO: find a better way of checking this https://github.com/model-checking/rmc/issues/152
-        self.type_name().map_or(false, |name| name.contains("boxed::Box"))
+        true
+        //self.type_name().map_or(false, |name| name.contains("boxed::Box"))
     }
 
     /// Checks if the struct represents a Rust "Unique"
     pub fn is_rust_unique_pointer(&self) -> bool {
```
celinval, line 140: Ditto

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/utils/utils.rs
```diff
@@ -132,16 +132,18 @@ impl Type {
         // can be reimported.
         // If we don't, we spuriously fail. https://github.com/model-checking/rmc/issues/113
         // TODO: find a better way of checking this https://github.com/model-checking/rmc/issues/152
-        self.type_name().map_or(false, |name| name.contains("boxed::Box"))
+        true
```
celinval, line 135: Can we just remove this function?

celinval (changes requested):

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/utils/utils.rs
danielsn: Fixed in previous PR

compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/utils/utils.rs
danielsn: Fixed in previous PR

tedinski (approved):

compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/symbol.rs
```diff
@@ -100,11 +100,11 @@ impl Symbol {
 
     /// The symbol that defines the type of the struct or union.
     /// For a struct foo this is the symbol "tag-foo" that maps to the type struct foo.
-    pub fn aggr_ty(t: Type) -> Symbol {
+    pub fn aggr_ty(t: Type, pretty_name: Option<String>) -> Symbol {
         //TODO take location
         let base_name = t.tag().unwrap().to_string();
         let name = aggr_name(&base_name);
-        Symbol::new(name, Location::none(), t, SymbolValues::None, Some(base_name), None)
+        Symbol::new(name, Location::none(), t, SymbolValues::None, Some(base_name), pretty_name)
```
avanhatt, line 107: Maybe a comment here to indicate this is not yet used in `--gen-c`?

src/test/rmc-multicrate/type-mismatch/run-mismatch-test.sh
```diff
@@ -0,0 +1,30 @@
...
+# Add the entry point and remove unused functions
+goto-cc --function main *.out -o a.out 
+goto-instrument --drop-unused-functions a.out b.out 
+
+# Run the solver
+RESULT="/tmp/dependency_test_result.txt"
+cbmc b.out &> $RESULT
+if grep -q "VERIFICATION SUCCESSFUL" $RESULT; then
+  cat $RESULT
+  echo "Successful dependency test"
```
avanhatt, line 24: Maybe include "mismatch" in message?

avanhatt (approved):
LGTM with a few documentation suggestions

Issue #491, opened 2021-09-16
Timeouts have no effect during dashboard generation


adpaco-aws:
When generating the dashboard, individual jobs are added a timeout which does not take into account sub-processes (e.g., CBMC). The dashboard gets stuck because these processes continue to run, so using the `run-build` timeout feature (https://github.com/awslabs/aws-build-accumulator/issues/34) may help here.

adpaco-aws:
Litani has just released version 1.15.0 which is now able to kill subprocesses (tested on the dashboard). We may use the `run-build` flag in the future to assign a general timeout to a dashboard run, but since the fix is available in the newest version it is better to use that for now. Renaming this issue.

PR #505, opened 2021-09-22
Generate and publish dashboard into documentation

### LINKED ISSUE #484: Automatically generate dashboard in CI
Add a new CI job so that dashboard are generated automatically and added to our online documentation. Depends on #482 and #483.

adpaco-aws:
### Description of changes: 

Generates and publishes the dashboard into `rmc-docs`, where we have included a dedicated section and the dashboard itself.

### Resolved issues:

Resolves #484 

### Call-outs:

If approved, the dashboard step will be restricted to run in the same way as the publication step.

### Testing:

* How is this change tested? See the CI run [here](https://github.com/model-checking/rmc/pull/505/checks?check_run_id=3690563156).

### Checklist


tedinski (approved):
1. I believe we discussed running the dashboard generation only on push to release?
2. Do we have any mechanism to timeout a dashboard run overall? I'm worried if it was flaky about nontermination to start then maybe we'd need to worry about that again? I don't want to end up with github actions running for... presumably some default timeout?

tedinski:
Well litani didn't work, so I guess we still don't have a sense of how long it'll take in CI

```
ModuleNotFoundError: No module named 'dataclasses'
```

adpaco-aws:
In response to your questions:
 1. Yes, but I wanted to test this on the PR first (it was worth it).
 2. Unfortunately not (see #491). I was thinking maybe try tools like `timeout <dashboard-command>`?

> Well litani didn't work, so I guess we still don't have a sense of how long it'll take in CI
> 
> ```
> ModuleNotFoundError: No module named 'dataclasses'
> ```

We are actually getting three different errors. For macOS we get the one below:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" }', src/tools/dashboard/src/litani.rs:93:48
```

I think this is due to the MacOS runner being too slow, basically.

The error on Ubuntu 20.04 is:

```
FileNotFoundError: [Errno 2] No such file or directory: 'dot'
```

Because we need to add litani dependencies.

adpaco-aws:
Okay, that worked (50 minutes). Saving the run [here](https://github.com/model-checking/rmc/pull/505/checks?check_run_id=3690563156).

I am going to apply the final changes now.

rmc-docs/src/dashboard.md
```diff
@@ -0,0 +1,11 @@
+# RMC Dashboard
+
+The [RMC Dashboard](./dashboard/index.html) purpose is to show the level of support in RMC for all Rust features.
```
avanhatt, line 3: Nit: grammar, either "RMC Dashboard's purpose" or "the purpose of..."

rmc-docs/src/dashboard.md
```diff
@@ -0,0 +1,11 @@
...
+To this end, we extract all examples from the following books:
```
avanhatt, line 4: Maybe more description, like "all Rust code snippet examples from the following general Rust documentation books"

scripts/setup/ubuntu/install_deps.sh
```diff
@@ -14,7 +14,9 @@ DEPS=(
   g++
   gcc
   git
+  gnuplot # Litani (recommended) dependency
```
avanhatt, line 17: Could you add a line to the README about running `install_deps.sh`?

avanhatt (changes requested):

.github/workflows/rmc.yml
```diff
@@ -49,6 +49,10 @@ jobs:
       - name: Execute RMC regression
         run: ./scripts/rmc-regression.sh
 
+      - name: Generate RMC dashboard
+        if: ${{ matrix.os == 'ubuntu-20.04' && github.event_name == 'push' && startsWith('refs/heads/main', github.ref) }}
+        run: ./x.py run -i --stage 1 dashboard
```
zhassan-aws, line 54: Is incremental (`-i`) needed here?

scripts/setup/ubuntu/install_deps.sh
zhassan-aws, line 17: Do we really need to add `gnuplot` as a dependency to RMC? It brings in a ton of other packages. Since the [Installation ](https://model-checking.github.io/rmc/install-guide.html) section of the tutorial refers to `install_deps.sh`, can we perhaps put `gnuplot` in a separate `install_dashboard_deps.sh` that is only run in CI?

.github/workflows/rmc.yml
adpaco-aws: Yes, otherwise RMC gets re-compiled.

scripts/setup/ubuntu/install_deps.sh
adpaco-aws: I liked this idea so I added it. The script is only run in the workflow when the dashboard is going to be generated.

rmc-docs/src/dashboard.md
adpaco-aws: Please check the new version.

rmc-docs/src/dashboard.md
adpaco-aws: I extended it a bit but would like to extend and review this in a follow-up PR, so opened #509

scripts/setup/ubuntu/install_deps.sh
adpaco-aws: I followed @zhassan-aws advice to put this into its own script. Do you want to use `install_deps.sh` in our README like we do in `rmc-docs`?

zhassan-aws (approved):

tedinski (approved):

adpaco-aws:
Following @avanhatt's advice, I have updated our README to point to `rmc-docs` so the quickstart section is simplified and we have installation instructions in one location. You can see a preview [here](https://github.com/adpaco-aws/rmc/tree/dash-ci), let me know what you think.

avanhatt (approved):

PR #516, opened 2021-09-27
Add clang-format tool, and a script to use it

danielsn:
### Description of changes: 

RMC has some C code, which has no default formatter. Use the `clang-format` spec from s2n. https://github.com/aws/s2n-tls/blob/main/.clang-format

### Resolved issues:

Resolves #ISSUE-NUMBER

### Call-outs:

The script has a `-d` mode to allow dry-runs for CI

### Testing:

* Is this a refactor change? Basically. It reformats code, and adds a check that it stays formatted.

### Checklist

### DIFF
--- a/scripts/rmc-regression.sh
+++ b/scripts/rmc-regression.sh
@@ -20,6 +20,9 @@ check-cbmc-viewer-version.py --major 2 --minor 5
 # Formatting check
 ./x.py fmt --check
 
+# C formatting check
+$SCRIPT_DIR/run-clang-format.sh -d 
+
 # Standalone rmc tests, expected tests, and cargo tests
 ./x.py build -i --stage 1 library/std ${EXTRA_X_PY_BUILD_ARGS}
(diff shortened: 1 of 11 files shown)


scripts/rmc-regression.sh
```diff
@@ -20,6 +20,9 @@ check-cbmc-viewer-version.py --major 2 --minor 5
 # Formatting check
 ./x.py fmt --check
 
+# C formatting check
+$SCRIPT_DIR/run-clang-format.sh -d
```
adpaco-aws, line 24: Not for this PR, but should we keep all formats check here or move them to their own scripts folder/workflow (for instance, with the copyright check)? Of course, the regression workflow should depend on that one.

I am happy to have a discussion about this.

danielsn: I actually ended up needing to move the check to there anyway, since ubuntu-18 has an old clang-format that doesn't support `--dry-run`.  Should I move the other check there too and rename the CI action?

adpaco-aws: Yes, please do that.

PR #523, opened 2021-09-27
Add support to `--function` to rmc script (#169)

### LINKED ISSUE #169: Allow entry-point for `rmc` other than `main`

celinval:
This changes the compilation to use crate type lib instead of binary when
we are running rmc on a single .rs file. This allow us to use
any public function as a verification target.

We have also changed the tests to use pub main so it is exported and it
can be used as the entry point of the proof.

Fix cargo-rmc and the current testcase to support library build.

### Description of changes: 

1- RMC users will need to ensure their main function is public in order to use them as verification harness. Note that this not affect proof harness with attribute `#[no_mangle]` since this attribute actually exports the function.
2- Users can now invoke rmc with the `--function` attribute. They can select any public function as their starting point. Note that the function still has to have the `#[no_mangle]` attribute.

### Resolved issues:

Resolves #169 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/src/tools/dashboard/src/util.rs
+++ b/src/tools/dashboard/src/util.rs
@@ -193,7 +193,7 @@ pub fn add_codegen_job(litani: &mut Litani, test_props: &TestProps) {
             "build/tmp",
             "-Z",
             "human_readable_cgu_names",
-            "-crate-type=lib",
+            "--crate-type=lib",
         ])
         .arg(&test_props.path);
     // TODO: replace `test` with `codegen` when Litani adds support for custom
--- a/src/tools/dashboard/src/util.rs
+++ b/src/tools/dashboard/src/util.rs
@@ -185,7 +185,16 @@ pub fn add_codegen_job(litani: &mut Litani, test_props: &TestProps) {
     let mut rmc_rustc = Command::new("rmc-rustc");
     rmc_rustc
         .args(&test_props.rustc_args)
-        .args(["-Z", "codegen-backend=gotoc", "--cfg=rmc", "--out-dir", "build/tmp"])
+        .args([
+            "-Z",
+            "codegen-backend=gotoc",
+            "--cfg=rmc",
+            "--out-dir",
+            "build/tmp",
+            "-Z",
+            "human_readable_cgu_names",
+            "-crate-type=lib",
+        ])
         .arg(&test_props.path);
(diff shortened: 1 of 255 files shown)


tedinski (approved):
I was **just** writing docs about `--function`.

I'll add the note about `#[no_mangle]` to what I wrote.

src/tools/dashboard/src/util.rs
```diff
@@ -185,7 +185,16 @@ pub fn add_codegen_job(litani: &mut Litani, test_props: &TestProps) {
...
-        .args(["-Z", "codegen-backend=gotoc", "--cfg=rmc", "--out-dir", "build/tmp"])
+        .args([
+            "-Z",
+            "codegen-backend=gotoc",
+            "--cfg=rmc",
+            "--out-dir",
+            "build/tmp",
+            "-Z",
+            "human_readable_cgu_names",
+            "-crate-type=lib",
```
zhassan-aws, line 196: Should be two dashes.

celinval: Good catch... I wonder when this gets run. I should check with @adpaco-aws to make sure we have test coverage in our regression.

tedinski: We have the dashboard set to only run after merges. I think there are concerns right now about prolonging CI times.

celinval: Anyway, I fixed it and I ran the dashboard script locally. The output is a bit hard to parse, but it seemed to work.

zhassan-aws (approved):

PR #529, opened 2021-09-29
optimize checkouts, and time and clean up scripts used in CI

tedinski:
### Description of changes: 

1. We can't shallow checkout our repo, because bootstrap expects commit history, but I added a comment because upstream rust seems to consider this a bug?
2. Don't checkout LLVM in CI, saves a good amount of download time.
3. Shallow checkouts of other submodules. Seems harmless... may save some time?
4. Avoid (seemingly unnecessary?) submodule update in the firecracker regression
5. I got mad at the regression testing scripts and tried to clean them up
6. Among other things, I discovered we weren't failing CI if the firecracker build failed. Resolved that with `set -eu`
7. Added timing and clear "start" "stop" log messages for various sub-components of rmc-regression

Let's see if any CI times are noticably shorter in this PR...

### Checklist
- [n/a] Methods or procedures are documented
- [n/a] Regression or unit tests are included, or existing tests cover the modified code

### DIFF
--- a/.github/workflows/rmc.yml
+++ b/.github/workflows/rmc.yml
@@ -14,6 +14,9 @@ jobs:
     steps:
       - name: Checkout RMC
         uses: actions/checkout@v2
+        # https://github.com/rust-lang/rust/issues/87890
+        # A non-shallow clone is unfortunately necessary, as bootstrap.py searches commit history
+        # to find the correct LLVM artifacts to download. (0 below indicates non-shallow)
         with:
           fetch-depth: 0
 
@@ -29,10 +32,6 @@ jobs:
       - name: Install Rust toolchain
         run: ./scripts/setup/install_rustup.sh
 
-      - name: Install dashboard dependencies
-        if: ${{ matrix.os == 'ubuntu-20.04' && github.event_name == 'push' && startsWith('refs/heads/main', github.ref) }}
-        run: ./scripts/setup/install_dashboard_deps.sh
-
       - name: Set config.toml file
         run: |
           ./configure \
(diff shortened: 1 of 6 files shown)


scripts/codegen-firecracker.sh
```diff
@@ -2,18 +2,29 @@
...
+echo "Starting Firecracker codegen regression..."
+echo
+
 # At the moment, we only test codegen for the virtio module
-git submodule update --init --recursive
 cd $RMC_DIR/firecracker/src/devices/src/virtio/
 RUST_BACKTRACE=1 RUSTFLAGS="-Z trim-diagnostic-paths=no -Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo build --target x86_64-unknown-linux-gnu
+
+echo
+echo "Finished Firefracker codegen regression successfully..."
```
zhassan-aws, line 29: fracker -> cracker 😄

.github/workflows/rmc.yml
```diff
@@ -14,6 +14,9 @@ jobs:
     steps:
       - name: Checkout RMC
         uses: actions/checkout@v2
+        # https://github.com/rust-lang/rust/issues/87890
+        # A non-shallow clone is unfortunately necessary, as bootstrap.py searches commit history
```
zhassan-aws, line 18: ☹️

.github/workflows/rmc.yml
```diff
@@ -42,17 +41,24 @@ jobs:
...
+        run: |
+          # Since we download cached artifacts, we can skip checking out llvm locally in CI
+          git config submodule."src/llvm-project".update none
+          git submodule update --init --depth 1
 
       - name: Build RMC
-        run: ./x.py build -i --stage 1 library/std
+        run: |
+          export RUST_BACKTRACE=1
+          ./x.py build -i --stage 1 library/std
```
zhassan-aws, line 52: Would adding `-j <n>` here help?

.github/workflows/rmc.yml
```diff
@@ -42,17 +41,24 @@ jobs:
...
-        run: export RUST_BACKTRACE=1
+        run: |
+          # Since we download cached artifacts, we can skip checking out llvm locally in CI
+          git config submodule."src/llvm-project".update none
+          git submodule update --init --depth 1
 
       - name: Build RMC
-        run: ./x.py build -i --stage 1 library/std
+        run: |
+          export RUST_BACKTRACE=1
```
celinval, line 51: you could also use env to set the environment variable: https://docs.github.com/en/actions/learn-github-actions/environment-variables

It can be done in a workflow, job or step level.

zhassan-aws (approved):

scripts/std-lib-regression.sh
```diff
@@ -2,20 +2,26 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
+# Deliberately not enabling this, since we expect a failure currently and are failing based on 'grep' later
```
celinval, line 5: You could add a `| true` at the end of the command that we expect to fail for now. Then you can enable `set -eu` to ensure that other errors with is correctly handled.

celinval (approved):
I added a few comments but nothing that blocks this change. Thanks!

celinval:
I'll merge this one since I want to push my upstream merge now. We can address the comments in a follow up PR.

tedinski:
Haha, wow. The submodules changes went from almost 5 minutes to 30 seconds. Awesome. That was most of the time savings for this change though. So a 4ish minute win for now.

PR #535, opened 2021-10-01
use assert-false instead of nondet for missing functions

danielsn:
### Description of changes: 

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


zhassan-aws (commented):
Can you clarify the impact of this change? Will it cause verification to fail if the program includes a missing function that is reachable? If so, can you add a comment explaining this?

celinval:
This is a tough one. I almost feel like this should be an option to how a user runs rmc or even a proof-harness.

src/test/rmc/ForeignItems/missing_fn_fail.rs
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! This tests whether we properly replace missing functions with `assert(false)`
+//! To run this test, do
+//! rmc missing_fn_fail.rs -- lib.c
+
+// rmc-flags: --c-lib src/test/rmc/ForeignItems/lib.c
+
+// https://doc.rust-lang.org/reference/items/external-blocks.html
+// https://doc.rust-lang.org/nomicon/ffi.html
+extern "C" {
```
celinval, line 11: Can we do this without extern "C"? How common is a construct like this in user code?

danielsn: I'd love to do this in pure rust. But I can't easily tell how to generate that case.  Thoughts?

nmanthey:
This is used in CBMC when we run verification of larger projects where we cut off calls to functions, but do not want execution to stop. Essentially, this is part of aggressive slicing - and yes, it's known to be unsound.

PR #536, opened 2021-10-01
Add option to restrict vtable function pointers

### LINKED ISSUE #521: Use vtable cache to restrict CBMC function pointers
Try using the RMC vtable cache to restrict CBMC function pointers via this intertface: http://cprover.diffblue.com/md__home_travis_build_diffblue_cbmc_doc_architectural_restrict-function-pointer.html

avanhatt:
### Description of changes: 

RMC can currently take a long time to verify code with lots of virtual function pointer calls (or, never terminate). This is in part because CBMC uses a very general heuristic to reason about function pointers: it assumes any function of the right type signature could be a target. CBMC does allow you to restrict possible function targets via a command line option and auxiliary json file: http://cprover.diffblue.com/md__home_travis_build_diffblue_cbmc_doc_architectural_restrict-function-pointer.html

Note, using this functionality requires this branch of CBMC, https://github.com/diffblue/cbmc/pull/6376, tracking issue to update RMC once that merges: https://github.com/model-checking/rmc/issues/671

For virtual function pointer through a vtable, we can often do much better than CBMC's default heuristic. In particular, we can virtually guarantee (unless the user transmutes the vtable memory) that a call through a vtable will be one of the trait-defined method for that trait type that we have previously seen in codegen. Because CBMC asserts(false) if the function pointer is not actually within the restricted set, this optimization is sound.

This is especially helpful for `drop_in_place` destructor calls through a vtable. Many functions have types that match `drop` since it takes a single self argument and returns unit. In a Firecracker example, we have seen 300+ function targets identified by the default CBMC heuristic for a single virtual drop call, which otherwise renders CBMC verification basically intractable (think: a massive, 300+ arm `if` block after CBMC performs function pointer elimination). 

This PR add a `VtableCtx` object that build a map from virtual call sites to possible trait-defined method implementations, then writes out the restrictions as a JSON file. 

You can invoke this on single Rust files by passing a `--restrict-vtables` flag to `rmc`, or on full crates with dependencies by adding the following to your harness script (which should eventually be handled by `cargo rmc`:
```
RUST_BACKTRACE=full RUSTFLAGS=" -Z restrict_vtable_fn_ptrs ...
... 
rmc-link-restrictions . <output dir>
...
goto-instrument  --function-pointer-restrictions-file restrictions.json ... 

```

Example:

Say we have two traits that both implement a `noise` method, `Animal` and `Other`. Without the vtable restriction flag, any call to `noise` _and_ any implicit `drop` on a dynamic trait object would unroll to every implementation of either trait:

```rust
  if(!(var_21.vtable->3 == (int (*)(void *))Sheep::noise))
  {
    if(var_21.vtable->3 == (int (*)(void *))i32::noise)
      goto __CPROVER_DUMP_L11;

    if(var_21.vtable->3 == (int (*)(void *))Cow::noise)
      goto __CPROVER_DUMP_L12;

    __CPROVER_assume(0);
  }
```

Now, we see each call site only considers the implementations that match that trait type:
```rust
// Animal::noise 
  if(!(main_1_function_pointer_call_1 == (int (*)(void *))Sheep::noise))
  {
    if(main_1_function_pointer_call_1 == (int (*)(void *))Cow::noise)
      goto __CPROVER_DUMP_L2;

    /* dereferenced function pointer at must be one of [Sheep::noise, Cow::noise] */
    assert(0);
    __CPROVER_assume(0);
  }
  
// Other::noise
  if(!(main_2_function_pointer_call_1 == (int (*)(void *))i32::noise))
  {
    /* dereferenced function pointer at must be i32::noise */
    assert(0);
    __CPROVER_assume(0);
  }
``` 

We also get the benefit of CBMC inserting an `assert(0)` at the end of the block instead of just an `assume(0)`. 

### Resolved issues:

Resolves #521

### Call-outs:

- Using this feature will require this CBMC update to land (which fixes a limitation with command line restriction validation): https://github.com/diffblue/cbmc/pull/6376.

### Testing:

* How is this change tested?

- Manually improved performance on the Firecracker block device example.

* Is this a refactor change?

### Checklist


celinval (changes requested):
Is it possible to make this change less CBMC dependent? Split between collecting potential candidates and doing the CBMC translation after that.

avanhatt:
@celinval Agreed on the goal. I tried to keep the totally CBMC-specific logic to within `get_virtual_function_restrictions`, do you see other place that it would make sense to split off?

celinval:
@avanhatt I was thinking about whether we should split this file: compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs

avanhatt:
@celinval many rebases and switching to serde later, this is now ready for re-review! The CBMC specifics are now in an entirely different crate.

compiler/rustc_codegen_rmc/src/codegen/rvalue.rs
```diff
@@ -721,7 +721,17 @@ impl<'tcx> GotocCtx<'tcx> {
 
         // Lookup in the symbol table using the full symbol table name/key
         let fn_name = self.symbol_name(instance);
-        if let Some(fn_symbol) = self.symbol_table.lookup(&fn_name) {
+
+        if let Some(fn_symbol) = self.symbol_table.clone().lookup(&fn_name) {
```
danielsn, line 725: This clone is going to be super expensive. Please clone after the lookup!

compiler/rustc_codegen_rmc/src/codegen/rvalue.rs
```diff
@@ -760,6 +779,9 @@ impl<'tcx> GotocCtx<'tcx> {
             // for it. Build and insert a function that just calls an unimplemented block
             // to maintain soundness.
             let drop_sym_name = format!("{}_unimplemented", self.symbol_name(drop_instance));
+            if let Some(fn_symbol) = self.symbol_table.clone().lookup(&drop_sym_name) {
```
danielsn, line 782: Same here

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,197 @@
...
+
+/// This structure represents data about the vtable that we construct
+/// Per trait, per method, which functions could virtual call sites
+/// possibly refer to?
+pub struct VtableCtx {
+    // Option to actually enable restrictions
+    pub restrict_vtable_fn_ptrs: bool,
+
+    // Map: (normalized trait name, method index) -> possible implementations
+    possible_methods: FxHashMap<TraitDefinedMethod, Vec<String>>,
```
danielsn, line 33: Should this be internedstring?

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,197 @@
...
+        let fn_type = func_exp.typ().clone();
+        let parameters: Vec<Symbol> = fn_type
+            .parameters()
+            .unwrap()
+            .iter()
+            .enumerate()
+            .map(|(i, parameter)| {
+                let name = format!("{}_{}", wrapper_name.clone(), i);
+                let param = Symbol::variable(
+                    name.to_string(),
```
danielsn, line 142: Is the tostring needed

library/rmc_restrictions/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+
+// Data structures to represent vtable trait function pointer restrictions
+use rustc_data_structures::stable_map::FxHashMap;
+use serde::{Deserialize, Serialize};
+
+/// Trait-defined method: the trait type and the vtable index of the method.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct TraitDefinedMethod {
+    // Needs to be a string to handle both the MIR and Gotoc types
+    pub trait_name: String,
```
danielsn, line 12: Interned string

library/rmc_restrictions/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub vtable_idx: usize,
+}
+
+// Call sites are identified by the trait-defined method and the function location
+// - we use a wrapper function to unambigiously point to call site locations.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct CallSite {
+    pub trait_method: TraitDefinedMethod,
+    pub function_location: String,
```
danielsn, line 21: What is the format of this

src/tools/rmc-link-restrictions/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+use std::io::BufReader;
+
+fn link_function_pointer_restrictions(data_per_crate: Vec<VtableCtxResults>) -> serde_json::Value {
+    let mut output = FxHashMap::default();
+    let mut combined_possible_methods: FxHashMap<TraitDefinedMethod, Vec<std::string::String>> =
+        FxHashMap::default();
+
+    // Combine method possibilities
+    for crate_data in &data_per_crate {
+        for entry in crate_data.possible_methods.clone() {
```
danielsn, line 17: Is this the right place to put the clone?

danielsn (changes requested):

scripts/cargo-rmc
```diff
@@ -59,6 +65,9 @@ def main():
     symbols_filename = os.path.join(args.target_dir, "cbmc.symbols")
     out_files = rmc.symbol_table_to_gotoc(symbol_table_jsons, args.verbose, args.keep_temps, args.dry_run)
 
+    restrictions_filename = base + ".symtab.json"
```
danielsn, line 68: `.restrictions.json`

library/rmc_restrictions/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct PossibleMethodEntry {
+    pub trait_method: TraitDefinedMethod,
+    pub possibilities: Vec<String>,
```
danielsn, line 27: InternedString

compiler/rustc_codegen_rmc/src/codegen/assumptions.rs
```diff
@@ -179,7 +179,7 @@ impl<'tcx> GotocCtx<'tcx> {
                     self.find_function(fname)
                 }
             }
-            ty::Dynamic(_, _) => unimplemented!(),
+            ty::Dynamic(_, _) => None,
```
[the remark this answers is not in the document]
avanhatt: we talked about this in a chime sync, this is just the set of assumptions on the type, so returning `None` is no less safe than `unimplemented!` but let's us not crash.

src/test/rmc/DynTrait/vtable_restrictions_fail_fixme.rs
```diff
@@ -0,0 +1,81 @@
...
+        assert!(s == 2);
+    }
+
+    let other = &5 as &dyn Other;
+
+    // Manually transmute other's vtable to point to a different method
+    unsafe {
+        let vtable_metadata: std::ptr::DynMetadata<dyn std::any::Any> = vtable!(other);
+        let vtable_ptr: *mut usize = std::mem::transmute(vtable_metadata);
+        let noise_ptr = vtable_ptr.offset(3);
```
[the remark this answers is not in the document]
avanhatt: No, but there is no safe/stable way to mangle the vtable AFAIK, and the test would fail if it were to change.

compiler/rustc_codegen_rmc/src/context/goto_ctx.rs
```diff
@@ -58,13 +61,15 @@ impl<'tcx> GotocCtx<'tcx> {
         let fhks = fn_hooks();
         let mm = machine_model_from_session(tcx.sess);
         let symbol_table = SymbolTable::new(mm);
+        let restrict_vtable_fn_ptrs = tcx.sess.opts.debugging_opts.restrict_vtable_fn_ptrs;
```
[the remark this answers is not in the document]
avanhatt: do you like `emit_vtable_restrictions` better?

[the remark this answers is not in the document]
danielsn: Yeah

scripts/rmc.py
```diff
@@ -45,29 +47,37 @@ def match(self, text):
...
     return which(name) is not None
 
+
 def ensure_dependencies_in_path():
     for program in [RMC_RUSTC_EXE, "symtab2gb", "cbmc", "cbmc-viewer", "goto-instrument", "goto-cc"]:
         ensure(is_exe(program), f"Could not find {program} in PATH")
 
 # Assert a condition holds, or produce a user error message.
+
+
```
danielsn, line 62: Why the new whitespace?

avanhatt: We check python formatting with `autopep8` on changed files, it inserted the whitepace

avanhatt: Ah didn't realize quite how copious it was, might need to move some of these to doc comments to make it happy

compiler/rustc_session/src/options.rs
```diff
@@ -1284,6 +1284,8 @@ options! {
         "choose which RELRO level to use"),
     remap_cwd_prefix: Option<PathBuf> = (None, parse_opt_pathbuf, [TRACKED],
         "remap paths under the current working directory to this path prefix"),
+    restrict_vtable_fn_ptrs: bool = (false, parse_bool, [TRACKED],
+        "restrict the targets of virtual table function pointer calls"),
```
tedinski, line 1288: Is there a strong reason to even make this a flag? Can we emit these files unconditionally?

avanhatt: The main reasons are that currently emitting restrictions currently (1) requires wrapper functions at ever vtable call, because CBMC uses fragile naming conventions for specifying the function call site, and (2) the restrictions could be incomplete if the Rust code is doing certain types of casts that we don't handle, meaning the restriction set would be too small.

The 3rd even more transient reason is that the CBMC PR has not landed, so restrictions fail on our minimum required CBMC version.

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// possible targets.
+    ///
+    /// We need to wrap because for the current implemention, CBMC employs
+    /// a hard-to-get-right naming scheme for restrictions: the call site is
+    /// named for its index in  a given function. We don't have a good way to
+    /// track _all_ function pointers within the function, so wrapping the call
+    /// to a function that makes a single virtual function pointer call makes
+    /// the naming unambiguous.
+    ///
+    /// This can be simplified if CBMC implemented label-based restrictions.
```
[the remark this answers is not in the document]
avanhatt: Added!

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,197 @@
...
+                );
+                self.symbol_table.insert(param.clone());
+                param
+            })
+            .collect();
+
+        // Finish constructing the wrapper type
+        let ret_typ = fn_type.return_type().unwrap().clone();
+        let param_typs = parameters.clone().iter().map(|p| p.to_function_parameter()).collect();
+        let new_typ = if fn_type.is_code() {
```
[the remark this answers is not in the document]
avanhatt: Unless something is wrong elsewhere, but added an unreachable assertion.

src/tools/rmc-link-restrictions/src/main.rs
```diff
@@ -0,0 +1,77 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use rmc_restrictions::*;
+use rustc_data_structures::stable_map::FxHashMap;
+use std::env;
+use std::fs::File;
+use std::io::BufReader;
+
+fn link_function_pointer_restrictions(data_per_crate: Vec<VtableCtxResults>) -> serde_json::Value {
```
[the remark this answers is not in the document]
avanhatt: it's a single call to `serde_json::to_value(&output).unwrap()` so I don't feel strongly about that, but happy to move if you prefer?

src/tools/rmc-link-restrictions/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+    let args: Vec<String> = env::args().collect();
+    assert!(args.len() == 3);
+    let path = &args[1];
+    let outpath = &args[2];
+    let md = std::fs::metadata(path).unwrap();
+
+    // Find all files in that path with the expected suffix
+    let mut per_crate_restrictions = Vec::new();
+
+    if md.is_dir() {
```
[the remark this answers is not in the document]
avanhatt: I suspect this is good enough for now, given how our cargo scripts work.

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,216 @@
...
+        if let Some(possibilities) = self.possible_methods.get_mut(&key) {
+            possibilities.push(imp);
+        } else {
+            self.possible_methods.insert(key, vec![imp]);
+        }
+    }
+
+    /// The vtable index for drop is 2
+    pub fn drop_index() -> usize {
+        2
```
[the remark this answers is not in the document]
avanhatt: Switched to named value from MIR.

compiler/rustc_codegen_rmc/src/codegen/rvalue.rs
avanhatt: Totally just a mistake, thanks!

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
avanhatt: Mistake.

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
avanhatt: Not anymore at least

library/rmc_restrictions/src/lib.rs
avanhatt: done

library/rmc_restrictions/src/lib.rs
avanhatt: It's just the name now, renamed the variable to make clear

src/tools/rmc-link-restrictions/src/main.rs
avanhatt: moved

compiler/cbmc/src/irep/serialize.rs
```diff
@@ -98,6 +98,32 @@ impl Serialize for InternedString {
     }
 }
 
+struct InternedStringVisitor;
```
danielsn, line 101: This should really go with InternedString itself.

Cargo.lock
```diff
@@ -3342,6 +3342,25 @@ dependencies = [
 name = "rmc"
 version = "0.1.0"
 
+[[package]]
+name = "rmc-link-restrictions"
+version = "0.1.0"
+dependencies = [
+ "rmc_restrictions",
+ "rustc_data_structures",
```
danielsn, line 3350: What are we using from here?

compiler/rustc_codegen_rmc/src/codegen/rvalue.rs
```diff
@@ -760,6 +779,9 @@ impl<'tcx> GotocCtx<'tcx> {
             // for it. Build and insert a function that just calls an unimplemented block
             // to maintain soundness.
             let drop_sym_name = format!("{}_unimplemented", self.symbol_name(drop_instance));
+            if let Some(fn_symbol) = self.symbol_table.lookup(&drop_sym_name) {
```
danielsn, line 782: using `ensure` would be more canonical here

compiler/rustc_codegen_rmc/src/compiler_interface.rs
```diff
@@ -126,12 +129,22 @@ impl CodegenBackend for GotocCodegenBackend {
             &tcx.sess.opts.debugging_opts.symbol_table_passes,
         );
 
-        let type_map = BTreeMap::from_iter(c.type_map.into_iter().map(|(k, v)| (k, v.to_string())));
+        // Map MIR types to GotoC types
+        let type_map =
+            BTreeMap::from_iter(c.type_map.into_iter().map(|(k, v)| (k, v.to_string().into())));
```
danielsn, line 134: Why do we need to do this, instead of just keeping the `<String, type>` map?

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,207 @@
...
+    // All sites where a virtual call takes place
+    call_sites: Vec<CallSite>,
+
+    // Internal tracing of index needed for call site wrappers
+    call_site_global_idx: usize,
+}
+
+/// Constructor
+impl VtableCtx {
+    pub fn new(restrict_ptrs: bool) -> Self {
```
danielsn, line 45: why not just make this the same name as the struct field?

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,207 @@
...
+impl VtableCtx {
+    /// Add a possible implementation for a virtual method call.
+    pub fn add_possible_method(
+        &mut self,
+        trait_name: InternedString,
+        method: usize,
+        imp: InternedString,
+    ) {
+        assert!(self.emit_vtable_restrictions);
+        let key = TraitDefinedMethod { trait_name: trait_name.clone(), vtable_idx: method };
```
danielsn, line 66: clone not needed here. 
```Rust
        let key = TraitDefinedMethod { trait_name, vtable_idx: method };
```

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// site index.
+    fn add_call_site(
+        &mut self,
+        trait_name: InternedString,
+        method: usize,
+        function_name: InternedString,
+    ) {
+        assert!(self.emit_vtable_restrictions);
+        let site = CallSite {
+            trait_method: TraitDefinedMethod { trait_name: trait_name, vtable_idx: method },
```
danielsn, line 99: don't need to say `trait_name: trait_name`, `trait_name` would do

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,207 @@
...
+        // Crate-based naming scheme for wrappers
+        let full_crate_name = self.full_crate_name().to_string().replace("::", "_");
+        let wrapper_name: InternedString = format!(
+            "restricted_call_{}_{}",
+            full_crate_name,
+            self.vtable_ctx.get_call_site_global_idx()
+        )
+        .into();
+
+        // We only have the Gotoc type, we need to normalize to match the MIR type.
```
danielsn, line 138: Is there any risk here if MIR changes?

danielsn (approved):
Approved with comments

src/tools/rmc-link-restrictions/src/main.rs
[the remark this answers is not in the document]
danielsn: It just feels weird given the function name

compiler/rustc_session/src/options.rs
tedinski: Could we get that broken into an issue, with e.g. links to CBMC issues or PRs we depend on?

This change is awesome, and I'd like to make sure we have a well-lit path to making it the always-on default. e.g. do we know what the "certain types of casts that we don't handle" are, and can we document that?

(This is obviously not a blocker to merging, I just don't want it accidentally dropped or anything!)

compiler/rustc_codegen_rmc/src/compiler_interface.rs
avanhatt: Presumably because `TyS` doesn't have a serialization, the only thing I'm touching here is changing the string to be interned so I would rather not change more.

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
avanhatt: Less so now that we have the `type_map`, we are normalizing the MIR on both sides instead of relying on string editing.

Cargo.lock
avanhatt: The compiler's `FxHashMap`, not strictly necessary/could be replaced with a vanilla `HashMap`

danielsn: If easy, could do that here; if not, lets just do that as a cleanup PR

compiler/rustc_codegen_rmc/src/codegen/rvalue.rs
```diff
@@ -760,37 +779,37 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                let unimplemented = ctx
+                    .codegen_unimplemented(
+                        format!("drop_in_place for {}", drop_sym_name).as_str(),
+                        Type::empty(),
+                        Location::none(),
+                        "https://github.com/model-checking/rmc/issues/281",
+                    )
+                    .as_stmt(Location::none());
+
+                // Declare symbol for the single, self parameter
```
danielsn, line 793: Not for this PR, but this feels like something that would happen often enough to be worth a helper function

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,207 @@
...
+        &mut self,
+        trait_name: InternedString,
+        method: usize,
+        imp: InternedString,
+    ) {
+        assert!(self.emit_vtable_restrictions);
+        let key = TraitDefinedMethod { trait_name, vtable_idx: method };
+
+        if let Some(possibilities) = self.possible_methods.get_mut(&key) {
+            possibilities.push(imp);
```
danielsn, line 69: Is there any reason we might want to keep this list ordered?

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
```diff
@@ -0,0 +1,207 @@
...
+/// Internal tracking helpers
+impl VtableCtx {
+    fn get_call_site_global_idx(&mut self) -> usize {
+        assert!(self.emit_vtable_restrictions);
+        self.call_site_global_idx += 1;
+        self.call_site_global_idx
+    }
+
+    /// Add a given call site for a virtual function, incremementing the call
+    /// site index.
```
danielsn, line 90: Where does it increment the index??

compiler/rustc_codegen_rmc/src/compiler_interface.rs
danielsn: OK

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
avanhatt: This comment was stale, removed.

compiler/rustc_codegen_rmc/src/context/vtable_ctx.rs
avanhatt: Not currently.

celinval (commented):
I haven't finished reviewing this but so far I haven't seen anything critical. Once @danielsn approves this, I'm ok with merging and making the changes as a follow up PR.

Thanks

celinval (approved):

avanhatt:
Will address followups in https://github.com/model-checking/rmc/issues/671, merging to unblock Ted's linked PR.

Issue #109, opened 2021-04-27
Investigate failures to link in standard libraries


adpaco-aws:

nchong-at-aws:
```
cargo new SizeAndAlignOfDstTest #< this test needs mutex and cell (which aren't linked)
cd SizeAndAlignOfDstTest
cp ~/rmc/rust-tests/cbmc-reg/SizeAndAlignOfDst/main_fail.rs src/main.rs 
rustup component add rust-src --toolchain nightly
cargo +nightly run -Z build-std --target x86_64-unknown-linux-gnu #< succeeds (just compiling)
RUSTFLAGS="-Z trim-diagnostic-paths=no -Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo +nightly build -Z build-std --target x86_64-unknown-linux-gnu #< fails trying to compile libc
```

avanhatt:
> cp ~/rmc/rust-tests/cbmc-reg/SizeAndAlignOfDst/main_fail.rs src/main.rs

Is now (after tests moved): 
```
cp ../rmc/src/test/cbmc/SizeAndAlignOfDst/main_fail.rs src/main.rs
```

avanhatt:
We can now successfully codegen all of the crates in the standard library (!) (with some specific functions skipped, in the linked issues). 

However, we now fail with an actual linker error when attempting to compile the local crate, presumably because our backend does not build `.rlib` files:

```rust
 Compiling core v0.0.0 (/home/ubuntu/rmc/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/core)
   Compiling rustc-std-workspace-core v1.99.0 (/home/ubuntu/rmc/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/rustc-std-workspace-core)
   Compiling compiler_builtins v0.1.45
   Compiling libc v0.2.93
   Compiling alloc v0.0.0 (/home/ubuntu/rmc/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/alloc)
   Compiling cfg-if v0.1.10
   Compiling adler v0.2.3
   Compiling rustc-demangle v0.1.18
   Compiling unwind v0.0.0 (/home/ubuntu/rmc/build/x86_64-unknown-linux-gnu/stage1/lib/rustlib/src/rust/library/unwind)
[...]

PR #593, opened 2021-10-27
Fix std lib regression script (#109 and #375)

### LINKED ISSUE #375: Verify that the standard library regression is running in CI
See #346 to get the full picture. Local runs were failing while CI was not. The reasoning behind:

> Take [this run](https://github.com/model-checking/rmc/runs/3220251843?check_suite_focus=true) for example. The whole step takes 2m59s to execute. Out of these, 2m26s are spent in the main regression and 13s in unit tests. It seems unlikely that the stdlib is being codegen'd using the remaining 20s. Opening an issue to investigate.

We should investigate what is going on here. What is the usual runtime?

celinval:
Create a target library instead of a binary.

### Description of changes: 

The current std library build script fails during linking because it's trying to generate a binary file as the last stage. Use a crate with lib type instead.

### Resolved issues:

Resolves #375

### Testing:

### Checklist

### DIFF
--- a/scripts/std-lib-regression.sh
+++ b/scripts/std-lib-regression.sh
@@ -2,14 +2,15 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
-# Deliberately not enabling this, since we expect a failure currently and are failing based on 'grep' later
-#set -eu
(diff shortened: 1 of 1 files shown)


adpaco-aws (approved):

celinval:
Removed the #109 from the resolution list. We should do further testing to ensure that linking works.

Issue #589, opened 2021-10-26
Rename RMC prelude functions to more user-friendly names and to avoid warnings


zhassan-aws:
Compiling pretty much any project with RMC results in warnings related to snake case name and unused variables, e.g.:
```
warning: unused variable: `cond`
 --> src/lib.rs:1:22
  |
1 | fn __VERIFIER_assume(cond: bool) {
  |                      ^^^^ help: if this is intentional, prefix it with an underscore: `_cond`
  |
  = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `cond`
 --> src/lib.rs:5:27
  |
5 | fn __VERIFIER_expect_fail(cond: bool, message: &str) {
  |                           ^^^^ help: if this is intentional, prefix it with an underscore: `_cond`

warning: unused variable: `message`
 --> src/lib.rs:5:39
  |
5 | fn __VERIFIER_expect_fail(cond: bool, message: &str) {
  |                                       ^^^^^^^ help: if this is intentional, prefix it with an underscore: `_message`

warning: function `__VERIFIER_assume` should have a snake case name
 --> src/lib.rs:1:4
  |
1 | fn __VERIFIER_assume(cond: bool) {
  |    ^^^^^^^^^^^^^^^^^ help: convert the identifier to snake case: `__verifier_assume`
  |
  = note: `#[warn(non_snake_case)]` on by default

warning: function `__VERIFIER_expect_fail` should have a snake case name
 --> src/lib.rs:5:4
  |
5 | fn __VERIFIER_expect_fail(cond: bool, message: &str) {
  |    ^^^^^^^^^^^^^^^^^^^^^^ help: convert the identifier to snake case: `__verifier_expect_fail`
```
In addition, the double underscore prefix (e.g. in `__nondet()`) seems unnecessary. We need to rename the functions and update them to avoid those warnings.

Here are example steps to reproduce the above warnings:
1. `cargo new --lib foo`
2. `cd foo`
3. Put the following in `src/lib.rs`:
```rust
fn __VERIFIER_assume(cond: bool) {
    unimplemented!()
}

fn __VERIFIER_expect_fail(cond: bool, message: &str) {
    unimplemented!()
}

fn __nondet<T>() -> T {
    unimplemented!()
}

pub fn main() {
    let x: u32 = __nondet();
    __VERIFIER_assume(x < 10);
    let y: u32 = x * 2;
    assert!(y < 19);
}
```
3. Compile using:
```
[...]

PR #597, opened 2021-10-29
Create an RMC crate and load it as part of rmc-rustc

### LINKED ISSUE #231: Have a crate for RMC prelude instead of using an include macro
#230 is adding a prelude file which ideally would be a crate that we import instead.

celinval:
### Description of changes: 

Users should now be able to use `rmc::nondet()`, `rmc::assume(bool)` and `rmc::expect_fail` as part of their test harnesses. RMC will automatically load the rmc library as part of any crate compilation.

### Resolved issues:

Resolves #231 and #589.

### Call-outs:

This PR also include updates to enable this change.
- I have also updated rmc-rustc to include the flags required by rmc so we don't need to maintain them in so many different places.
- I updated the tests to use the new definition.s
- I fixed an issue in the rmc codegen that was ignoring no-codegen option. This was breaking the compiletest check.

### Testing:

* How is this change tested? In all tests that were updated.

### Checklist


compiler/rustc_codegen_rmc/src/compiler_interface.rs
```diff
@@ -157,18 +157,21 @@ impl CodegenBackend for GotocCodegenBackend {
             .downcast::<GotocCodegenResult>()
             .expect("in link: codegen_results is not a GotocCodegenResult");
 
-        // "path.o"
-        let base_filename = outputs.path(OutputType::Object);
+        // No output should be generated if user selected no_codegen.
+        if !sess.opts.debugging_opts.no_codegen && sess.opts.output_types.should_codegen() {
```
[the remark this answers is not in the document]
celinval: https://github.com/model-checking/rmc/pull/601/files

.github/workflows/rmc.yml
```diff
@@ -51,10 +51,12 @@ jobs:
           export RUST_BACKTRACE=1
           ./x.py build -i --stage 1 library/std
 
+      - name: Build RMC Library
+        run: bash -x ./scripts/setup/build_rmc_lib.sh
```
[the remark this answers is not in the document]
adpaco-aws: Makes sense.

adpaco-aws (approved):
LGTM! My only concern now is that compiling the RMC library is necessary to use these predicates, so we should update the documentation accordingly. But we can have that in a follow-up PR.

celinval:
> LGTM! My only concern now is that compiling the RMC library is necessary to use these predicates, so we should update the documentation accordingly. But we can have that in a follow-up PR.

Good point. I'll submit a PR for that!

library/rmc/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+/// In the snippet below, we are verifying the behavior of the function `fn_under_verification`
+/// under all possible i32 input values.
+///
+/// ```rust
+/// let inputA = rmc::nondet::<i32>();
+/// fn_under_verification(inputA);
+/// ```
+#[inline(never)]
+#[rustc_diagnostic_item = "RmcNonDet"]
+pub fn nondet<T>() -> T {
```
camshaft, line 46: This really needs to be marked as `unsafe` as callers need to ensure all of the type's invariants are met.

library/rmc/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+/// The following code may panic though:
+///
+/// ```rust
+/// let i : i32 = rmc::nondet();
+/// assert!(i < 0, "This may panic and verification should fail.");
+/// rmc::assume(i > 10);
+/// ```
+#[inline(never)]
+#[rustc_diagnostic_item = "RmcAssume"]
+pub fn assume(_cond: bool) {}
```
camshaft, line 30: For a better docs and IDE experience, it would be preferable to do something like:

```rust
pub fn assume(cond: bool) {
    let _ = cond;
}
```

Same goes for the other arguments.

library/rmc/src/lib.rs
celinval: I have pushed this code already, but I created #607 to capture the work to fix this. Thanks

library/rmc/src/lib.rs
celinval: Sure. I'll fix this next time I push some updates to this crate. Thanks for the suggestion.

library/rmc/src/lib.rs
celinval: @camshaft https://github.com/model-checking/rmc/pull/657

PR #606, opened 2021-11-01
Make rmc-rustc errors more user friendly

celinval:
Make rmc-rustc errors more user friendly and also build rmc library as part of the regression script.

### Description of changes: 

1- Change build library script to use global variable as function return instead of the output. This was consuming the error messages and the script failures were rather cryptic.
2- Run the build script as part of the regression.

### Resolved issues:

N/A

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


danielsn:
What is the output before/after?

celinval:
> What is the output before/after?

Before rmc-rustc would fail while trying to compile a file with the following error:
```
error: multiple input filenames provided (first two filenames are `<rmc_path>/target/debug` and `<rmc_path>/target/tmp`)
```
Now the failure is:

```
ERROR: Could not find RMC library.
Looked for: "<rmc_path>/target/*/librmc.rlib"
Was RMC library successfully built first?
```

danielsn (approved):

Issue #533, opened 2021-09-30
Panic in compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs


zhassan-aws:
I tried the following:

1. `cargo new --lib foo`
2. Add `serde` as a dependency in `Cargo.toml`:
```
[dependencies]
serde = "1.0.130"
```
3. Run:
```
RUST_BACKTRACE=1 RUSTFLAGS="-Z trim-diagnostic-paths=no -Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo build
```

with RMC version: `72d7cd412439460a7bc0ad0a2d483424377dbb4d`

```
$ RUST_BACKTRACE=1 RUSTFLAGS="-Z trim-diagnostic-paths=no -Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo build
    Updating crates.io index
   Compiling serde v1.0.130
thread 'rustc' panicked at 'removal index (is 0) should be < len (is 0)', compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs:306:64
stack backtrace:
   0: rust_begin_unwind
             at /home/ubuntu/tools/rmc/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /home/ubuntu/tools/rmc/library/core/src/panicking.rs:103:14
   2: alloc::vec::Vec<T,A>::remove::assert_failed
             at /home/ubuntu/tools/rmc/library/alloc/src/vec/mod.rs:1390:13
   3: alloc::vec::Vec<T,A>::remove
             at /home/ubuntu/tools/rmc/library/alloc/src/vec/mod.rs:1395:13
   4: rustc_codegen_rmc::codegen::intrinsic::<impl rustc_codegen_rmc::context::goto_ctx::GotocCtx>::codegen_intrinsic
             at /home/ubuntu/tools/rmc/compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs:306:58
   5: <rustc_codegen_rmc::overrides::hooks::Intrinsic as rustc_codegen_rmc::overrides::hooks::GotocHook>::handle
             at /home/ubuntu/tools/rmc/compiler/rustc_codegen_rmc/src/overrides/hooks.rs:266:21
   6: rustc_codegen_rmc::codegen::statement::<impl rustc_codegen_rmc::context::goto_ctx::GotocCtx>::codegen_funcall
[...]

PR #610, opened 2021-11-01
Fix panic due to wrong intrinsic arguments (#533)

celinval:
During codegen of function calls, we were ignoring argument of Unit
type. This was causing a failure due to missing arguments when an
intrinsic or closure was being called with an '()' as an argument.

This change modifies how we deal with Unit types during function
definition and function call.

### Description of changes: 

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/src/test/rmc/Iterator/try_fold.rs
+++ b/src/test/rmc/Iterator/try_fold.rs
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-// cbmc-flags: --unwind 20
+// cbmc-flags: --unwind 3
 
 pub fn main() {
     let arr = [(1, 2), (2, 2)];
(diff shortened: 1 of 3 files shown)


src/test/rmc/Iterator/try_fold.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// cbmc-flags: --unwind 20
```
danielsn, line 3: Crazy that we need an unwind of 20 for this example

danielsn (approved):
Nice to have this out of the way!

src/test/rmc/Iterator/try_fold.rs
adpaco-aws: Also non-intuitive for users, which I think is a real issue. We should get this straight when we look into unwinding.

celinval: That was just a random number. I can bring it down if you are worried about that.

Issue #40, opened 2021-04-13
We need to redesign how we handle stubbing rust code


danielsn:

zhassan-aws:
This is now tracked in https://github.com/model-checking/kani/issues/1809. Closing this issue.

celinval:
BTW, I think this is related to hooks, not stubbing.

PR #612, opened 2021-11-02
Remove old vector / hash_map hook (#40)

celinval:
We no longer plan to use these hooks so I am deleting them.

### Description of changes: 

Remove unused legacy code.

### Resolved issues:

None.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


compiler/rustc_codegen_rmc/src/context/goto_ctx.rs
```diff
@@ -56,14 +55,13 @@ pub struct GotocCtx<'tcx> {
 /// Constructor
 impl<'tcx> GotocCtx<'tcx> {
     pub fn new(tcx: TyCtxt<'tcx>) -> GotocCtx<'tcx> {
-        let (thks, fhks) = type_and_fn_hooks();
+        let fhks = fn_hooks();
```
danielsn, line 58: You're not just removing those hooks, but the idea of type hooks entirely.  Is there any case where we might want one?

danielsn (approved):

compiler/rustc_codegen_rmc/src/context/goto_ctx.rs
celinval: I was hoping to get rid of all hooks eventually. :)

I don't think we should keep dead code here. It's fairly small and easy to rewrite if we ever need it again.

PR #615, opened 2021-11-02
Change cargo rmc script to use RUSTC binary during build (#613)

### LINKED ISSUE #613: Failed to run custom build command error with cargo
I'm seeing an error whenever I run `cargo rmc` or `RUSTC=rmc-rustc cargo build`:
```
error: failed to run custom build command for `libc v0.2.106`

Caused by:
  could not execute process `/home/ubuntu/examples/debug/iss533/bar/target/debug/build/libc-8de24690c68dbcab/build-script-build` (never executed)

Caused by:
  No such file or directory (os error 2)
```
Steps to reproduce:
1. `cargo new --lib foo`
2. Add `rand = "0.8.4"` to the `Cargo.toml` dependencies
3. Run `RUST_BACKTRACE=1 RUSTFLAGS="-Z trim-diagnostic-paths=no -Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo build --target x86_64-unknown-linux-gnu -j1`

with RMC version: `a431229efd2`

celinval:
### Description of changes: 

When the crate contains a build script, cargo compiles the build script first without the RUSTFLAGS. Thus, we cannot use rmc-rustc directly, or it will try to compile the build script with rmc enabled. We then fail to generate the script binary.

### Resolved issues:

Resolves #613

### Call-outs:

We still need to add support to --target to be able to fully fix issue #613. I'll leave this PR marked as a draft for now.

### Testing:

* How is this change tested? TODO

* Is this a refactor change?

### Checklist

### DIFF
--- a/scripts/rmc.py
+++ b/scripts/rmc.py
@@ -192,10 +192,24 @@ def compile_single_rust_file(
 def cargo_build(crate, target_dir, verbose=False, debug=False, mangler="v0", dry_run=False, symbol_table_passes=[]):
     ensure(os.path.isdir(crate), f"Invalid path to crate: {crate}")
 
-    rustflags = rustc_flags(mangler, symbol_table_passes)
+    def get_config(option):
+        process = subprocess.run(
+            [RMC_RUSTC_EXE, option],
+            universal_newlines=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            env=os.environ,
+            cwd=crate)
+        if process.returncode != EXIT_CODE_SUCCESS:
+            raise Exception("[Error] Fail to extract rmc configuration.\n{}".format(
+                process.stdout))
+        return process.stdout
+
+    rustflags = rustc_flags(mangler, symbol_table_passes) + get_config("--rmc-flags").split()
+    rustc_path = get_config("--rmc-path").strip()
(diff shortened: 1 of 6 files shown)


scripts/rmc.py
```diff
@@ -192,10 +192,24 @@ def compile_single_rust_file(
...
 
-    rustflags = rustc_flags(mangler, symbol_table_passes)
+    def get_config(option):
+        process = subprocess.run(
+            [RMC_RUSTC_EXE, option],
+            universal_newlines=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            env=os.environ,
+            cwd=crate)
```
adpaco-aws, line 202: What about wrapping this into a `run_cmd` call? Otherwise `--dry-run` will not print this command (not sure if it works with cargo commands anyway).

celinval: That's a good question. I was thinking that this is a different use case. This is just to get the correct tool configuration so I'm not sure we want to include that in the --dry-run output.  The other problem is how `run_cmd` handles the process output. I didn't want to add yet another option to `run_cmd` that would change how the output is handled.

adpaco-aws: Right, this is not something we need in `--dry-run`.

adpaco-aws (approved):

PR #617, opened 2021-11-02
Use interned strings to reduce memory usage

danielsn:
### Description of changes: 

CBMC objects to have a large number of strings which refer to names: symbols, files, etc.
These tend to be reused many times, which causes signifcant memory usage.
If we intern the strings, each unique string is only allocated once, saving memory.### Resolved issues:

Resolves #ISSUE-NUMBER

### Call-outs:

I tried to use generics to allow interfaces not to care if they're called with `&str`, `String` or `InternedString`

Before: 
```
        Maximum resident set size (kbytes): 1024148
```

After:

```
Maximum resident set size (kbytes): 877736
```

### Testing:

* How is this change tested? Existing tests, plus additional unit test

### Checklist

### DIFF
--- a/compiler/cbmc/src/cbmc_string.rs
+++ b/compiler/cbmc/src/cbmc_string.rs
@@ -1,0 +1,118 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use lazy_static::lazy_static;
+use std::sync::Mutex;
+use string_interner::symbol::SymbolU32;
+use string_interner::StringInterner;
+
+/// This class implements an interner for Strings.
+/// CBMC objects to have a large number of strings which refer to names: symbols, files, etc.
+/// These tend to be reused many times, which causes signifcant memory usage.
+/// If we intern the strings, each unique string is only allocated once, saving memory.
+/// On the stdlib test, interning led to a 15% savings in peak memory usage.
+/// Since they're referred to by index, InternedStrings become `Copy`, which simplifies APIs.
+/// The downside is that interned strings live the lifetime of the execution.
+/// So you should only intern strings that will be used in long-lived data-structures, not temps.
+///
+/// We use a single global string interner, which is protected by a Mutex (i.e. threadsafe).
+/// To create an interned string, either do
+/// `let i : InternedString = s.into();` or
+/// `let i = s.intern();`
+#[derive(Debug, Clone, Hash, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct InternedString(SymbolU32);
+
+// Use a `Mutex` to make this thread safe.
+lazy_static! {
+    static ref INTERNER: Mutex<StringInterner> = Mutex::new(StringInterner::default());
+}
+
+impl InternedString {
+    pub fn is_empty(&self) -> bool {
+        self.map(|s| s.is_empty())
+    }
+
+    pub fn len(&self) -> usize {
+        self.map(|s| s.len())
+    }
+
+    /// Apply the function `f` to the interned string, represented as an &str.
+    /// Needed because exporting the &str backing the InternedString is blocked by lifetime rules.
+    /// Instead, this allows users to operate on the &str when needed.
+    pub fn map<T, F: FnOnce(&str) -> T>(&self, f: F) -> T {
+        f(INTERNER.lock().unwrap().resolve(self.0).unwrap())
+    }
+
+    pub fn starts_with(&self, pattern: &str) -> bool {
+        self.map(|s| s.starts_with(pattern))
+    }
+}
+
+impl std::fmt::Display for InternedString {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(fmt, "{}", INTERNER.lock().unwrap().resolve(self.0).unwrap())
+    }
+}
+
+impl<T> From<T> for InternedString
+where
+    T: AsRef<str>,
+{
+    fn from(s: T) -> InternedString {
+        InternedString(INTERNER.lock().unwrap().get_or_intern(s))
+    }
+}
+
+impl<T> PartialEq<T> for InternedString
+where
+    T: AsRef<str>,
+{
+    fn eq(&self, other: &T) -> bool {
+        INTERNER.lock().unwrap().resolve(self.0).unwrap() == other.as_ref()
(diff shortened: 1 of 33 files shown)


compiler/cbmc/src/goto_program/expr.rs
```diff
@@ -559,17 +560,22 @@ impl Expr {
     }
 
     /// `self.field`
-    pub fn member(self, field: &str, symbol_table: &SymbolTable) -> Self {
+    pub fn member<T>(self, field: T, symbol_table: &SymbolTable) -> Self
+    where
+        T: Into<InternedString>,
+    {
+        //TODO DSN
```
tedinski, line 567: What's up here?

compiler/cbmc/src/goto_program/symbol.rs
```diff
@@ -289,12 +323,12 @@ impl Symbol {
 impl Symbol {
     /// Makes a formal function parameter from a symbol.
     pub fn to_function_parameter(&self) -> Parameter {
-        self.typ.clone().as_parameter(Some(self.name.to_string()), self.base_name.clone())
+        self.typ.clone().as_parameter(Some(self.name), self.base_name.clone().map(|n| n.into()))
```
tedinski, line 326: if we're doing a map, do we still need the clone?

compiler/rustc_codegen_rmc/src/context/goto_ctx.rs
```diff
@@ -160,15 +162,18 @@ impl<'tcx> GotocCtx<'tcx> {
     /// Returns: a struct-tag referencing the inserted struct.
     pub fn ensure_struct<F: FnOnce(&mut GotocCtx<'tcx>, &str) -> Vec<DatatypeComponent>>(
         &mut self,
-        struct_name: &str,
-        pretty_name: Option<String>,
+        struct_name: &str, //DSN
```
tedinski, line 165: ?

danielsn: That's my marker that I haven't completed porting this function yet.

compiler/cbmc/src/goto_program/symbol.rs
danielsn: nope!

compiler/cbmc/src/goto_program/expr.rs
danielsn: a todo

compiler/cbmc/src/cbmc_string.rs
```diff
@@ -0,0 +1,117 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use lazy_static::lazy_static;
+use std::sync::Mutex;
+use string_interner::symbol::SymbolU32;
+use string_interner::StringInterner;
+
+/// This class implements an interner for Strings.
+/// CBMC objects to have a large number of strings which refer to names: symbols, files, etc.
+/// These tend to be reused many times, which causes signifcant memory usage.
```
avanhatt, line 11: Instead of/along with "significant", maybe give ballpark reduction?

compiler/cbmc/src/cbmc_string.rs
```diff
@@ -0,0 +1,117 @@
...
+/// If we intern the strings, each unique string is only allocated once, saving memory.
+/// Since they're referred to by index, InternedStrings become `Copy`, which simplifies APIs.
+/// The downside is that interned strings live the lifetime of the execution.
+/// So you should only intern strings that will be used in long-lived data-structures, not temps.
+///
+/// We use a single global string interner, which is protected by a Mutex (i.e. threadsafe).
```
avanhatt, line 17: How much does this slow compilation on crates with many dependencies?

compiler/rustc_codegen_rmc/src/codegen/typ.rs
```diff
@@ -459,16 +464,16 @@ impl<'tcx> GotocCtx<'tcx> {
     /// We handle this by treating it as an incomplete struct.
     fn codegen_foreign(&mut self, ty: Ty<'tcx>, defid: DefId) -> Type {
         debug!("codegen_foreign {:?} {:?}", ty, defid);
-        let name = self.ty_mangled_name(ty);
-        self.ensure(&aggr_name(&name), |ctx, _| {
-            Symbol::incomplete_struct(&name, Some(ctx.ty_pretty_name(ty)))
+        let name = self.ty_mangled_name(ty).intern();
```
avanhatt, line 467: Should `ty_mangled_name` just return an interned string?

compiler/cbmc/src/lib.rs
```diff
@@ -32,3 +32,7 @@ mod machine_model;
 pub mod utils;
 pub use irep::serialize;
 pub use machine_model::{MachineModel, RoundingMode};
+mod cbmc_string;
+pub use cbmc_string::{InternString, InternStringOption, InternedString};
+
+pub const NO_PRETTY_NAME: Option<InternedString> = None;
```
avanhatt, line 38: Is it Rust-ey to use this over just None?

celinval: I believe you have to be explicit about the type here or rustc won't be able to infer the type.

compiler/cbmc/src/cbmc_string.rs
danielsn: Actually seems to slightly speed it up!

compiler/cbmc/src/lib.rs
danielsn: Celina is correct. Rust can't infer the type, so either you use this, or the turbofish.  I'll add a comment.

compiler/cbmc/src/cbmc_string.rs
danielsn: On the stdlib test, this saves 146MB on a 1GB initial.  It used to save more, before we started streaming the ireps.

avanhatt: My suggestion would be to put a rough percentage in the comment, so if someone is looking at this without seeing the PR itself they have that context!

avanhatt (approved):

compiler/cbmc/src/cbmc_string.rs
danielsn: done

compiler/rustc_codegen_rmc/src/codegen/typ.rs
danielsn: done

Issue #608, opened 2021-11-01
Reducing RMC memory usage


tedinski:
I ran under a profiler and fought a bit with the results, I think I'm looking at "total bytes ever allocated" and not "total bytes allocated at one time", but it still gives probably very useful results:

* [ ] We use `BTreeMap` to in `Irep` and `Json`. The `Irep` is our fault, but the `Json` comes from `rustc_serialize`.

A `BTreeMap` seems to be a ~1k allocation by default, even if empty (I'm guessing). In a run with 8GB total allocations, fully 2.7GB of that were `BTreeMap` from `Irep`. I tried replacing these with `Vec` but it turns out we're using `rustc_serialize` which also uses `BTreeMap` for `Json` objects, which stopped my quick attempt, so I'm not sure if there's a win here without replacing how we serialize json too.

* [ ] `codegen_span` is hilarious. If I'm reading this right, this run has 400k calls allocating 1.6GB of... filenames. The filename, over and over again. Lots and lots of copies of the filenames.

...why is that call to `fs:canonicalize` even there? But that's probably a distraction. We really need some way of allocating these names once, instead of having to make so many copies of the string for every `Location`.

After the above we start getting into scattered cloned strings.

I strongly suspect our peak memory usage problem comes from having "MIR -> GotoC IRep -> Json -> String" all pinned in memory at once, with the bulk of the problem coming from inefficient `Irep` and `Json` representations, and tons and tons of duplication of e.g. location strings.

tedinski:
Closed with #625 #621 #620 #617

PR #618, opened 2021-11-02
Swap to Vec instead of BTreeMap in Irep to reduce memory usage

tedinski:
### Description of changes: 

When running a std lib build test with `time -v` I went from

```
    Maximum resident set size (kbytes): 11866204
    Maximum resident set size (kbytes): 11866052
```

to

```
    Maximum resident set size (kbytes): 9417760
    Maximum resident set size (kbytes): 9417552
```

Which is a savings of 2.4 GB of memory. This isn't enough to unblock running this test in CI yet, but it is (a) I think a good change and (b) doesn't get in the way of the real fix Celina is working on.

(The `Json` type we use from `rustc_serialize` also uses `BTreeMap` and it likely accounts for another good 2.5GB (at least, possibly even 5GB) of wasted memory. But that's part of rustc, so let's not change that. So our savings here is despite still having to build all those BTreeMaps.)

### Resolved issues:

Part of #608. Possibly all of it, if we determine string interning there isn't the win I thought it might be.

### Call-outs:

1. We don't actually call `lookup` anywhere, it's only used in `from_irep` which we don't use in RMC normally. So a small part of this code is untested afaict.
2. There's a slight change in theoretical behavior in `with_named_sub`. If the entry was already present it would previously get updated, while now we'd "duplicate" it. I did inspect all uses of this function, and we don't have any place where we're trying to do an "update" with this function, so there's no issue right now, but that's possibly a future footgun?
3. I think we should merge only if we think this is a good representation change. I think it is, but the memory savings won't be as important (I think) once the streaming-serialization fix is in.

### Testing:

* How is this change tested? existing regression suite

### Checklist

### DIFF
--- a/compiler/cbmc/src/irep/to_json.rs
+++ b/compiler/cbmc/src/irep/to_json.rs
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 use super::{Irep, IrepId, Symbol, SymbolTable};
+use std::collections::BTreeMap;
 use crate::btree_map;
 use rustc_serialize::json::*;
 
@@ -11,7 +12,11 @@ impl ToJson for Irep {
             output.insert("sub".to_string(), self.sub.to_json());
         }
         if !self.named_sub.is_empty() {
-            output.insert("namedSub".to_string(), self.named_sub.to_json());
+            let mut d = BTreeMap::new();
+            for (key, value) in &self.named_sub {
+                d.insert(key.to_string(), value.to_json());
+            }
+            output.insert("namedSub".to_string(), Json::Object(d));
         }
         Json::Object(output)
     }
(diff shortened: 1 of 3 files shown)


compiler/cbmc/src/irep/to_json.rs
```diff
@@ -11,7 +12,11 @@ impl ToJson for Irep {
             output.insert("sub".to_string(), self.sub.to_json());
         }
         if !self.named_sub.is_empty() {
-            output.insert("namedSub".to_string(), self.named_sub.to_json());
+            let mut d = BTreeMap::new();
```
danielsn, line 15: You can use `BTreeMap::FromIter()` here to avoid the loop

danielsn:
https://github.com/model-checking/rmc/pull/620 does the same thing, using a `vector_map` crate.

compiler/cbmc/src/irep/to_json.rs
tedinski: Yeah, that required a map though so it seemed like more complicated code

tedinski:
closing in favor of #620

PR #621, opened 2021-11-03
Minimize IRep Serialization memory consumption

celinval:
### Description of changes: 

Compiling the std-library was taking 11+G of memory. We root caused it to the codegen serialization code that was generating the entire json string in memory before dumping to a file.

In this change, I am using serde to serialize the irep with minimum buffering. This reduces the memory consumption to 4.5G while compiling the STD library. This can still be optimized by either improving the irep memory consumption or getting rid of the irep completely.

### Resolved issues:

This solves #622

### Testing:

* How is this change tested? Everywhere

### Checklist


compiler/rustc_codegen_rmc/src/compiler_interface.rs
```diff
@@ -164,13 +160,13 @@ impl CodegenBackend for GotocCodegenBackend {
 
             let symtab_filename = base_filename.with_extension("symtab.json");
             debug!("output to {:?}", symtab_filename);
-            let mut out_file = ::std::fs::File::create(&symtab_filename).unwrap();
-            write!(out_file, "{}", result.symtab.to_irep().to_json().pretty().to_string()).unwrap();
+            let out_file = ::std::fs::File::create(&symtab_filename).unwrap();
+            serde_json::to_writer(out_file, &result.symtab.to_irep()).unwrap();
```
tedinski, line 164: Wrap out_file with a BufWriter and see if that cures your performance problems.

adpaco-aws, line 164: I was experimenting with `serde` by doing:
```
            let out_file = ::std::fs::File::create(&symtab_filename).unwrap();
            let writer = BufWriter::new(out_file);
            serde_json::to_writer(writer, &result.symtab);
```
The main differences are:
 1. Using a BufWriter to handle writing.
 2. Serializing the symbol table directly instead of the irep representation of the single table.

This has brought memory usage down to 1GB if I have not made any mistakes in measuring.

Not sure if this is related to (2), but conceptually I think it is wrong to transform the symbol table with `to_irep` before writing the JSON file. The serialization of a symbol table is the irep representation itself. Of course, this solution would need writing custom serializers for some objects in the symbol table, but I think it is worth it.

celinval: That's a good idea. Let me try.

celinval: For sure, I agree that this is the right path. I used irep because it was the quick and dirty way to unblock me and a fairly simple change. No harm done.

compiler/cbmc/src/irep/serialize.rs
```diff
@@ -0,0 +1,125 @@
...
+    fn serialize_irep() {
+        let irep = Irep::empty();
+        let json_str = serde_json::to_string(&irep);
+        assert!(json_str.unwrap().contains(&IrepId::Empty.to_string()));
+    }
+
+    #[test]
+    fn serialize_sym_table() {
+        let mut sym_table = SymbolTable::new();
+        let symbol = Symbol {
```
danielsn, line 93: Would it make sense to use one of the `Symbol` constructors?

compiler/rustc_codegen_rmc/src/compiler_interface.rs
```diff
@@ -164,13 +161,15 @@ impl CodegenBackend for GotocCodegenBackend {
...
+            let writer = BufWriter::new(out_file);
+            serde_json::to_writer(writer, &result.symtab.to_irep()).unwrap();
 
             let type_map_filename = base_filename.with_extension("type_map.json");
             debug!("type_map to {:?}", type_map_filename);
-            let mut out_file = ::std::fs::File::create(&type_map_filename).unwrap();
-            write!(out_file, "{}", result.type_map.to_json().pretty().to_string()).unwrap();
+            let out_file = ::std::fs::File::create(&type_map_filename).unwrap();
+            let writer = BufWriter::new(out_file);
+            serde_json::to_writer(writer, &result.type_map).unwrap();
```
danielsn, line 172: We do this bit more than once. Should it be a helper function?

scripts/std-lib-regression.sh
```diff
@@ -2,14 +2,15 @@
...
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
-# Deliberately not enabling this, since we expect a failure currently and are failing based on 'grep' later
-#set -eu
+set -eu
 
 # Test for platform
 PLATFORM=$(uname -sp)
-if [[ $PLATFORM == "Linux x86_64" ]]; then
+if [[ $PLATFORM == "Linux x86_64" ]]
```
danielsn, line 9: What auto-formatter are you using, and can we get it enabled in CI so we don't have whitespace wars?

compiler/cbmc/src/irep/serialize.rs
```diff
@@ -0,0 +1,125 @@
...
+            typ: Irep::empty(),
+            value: Irep::empty(),
```
danielsn, line 95: Should we add something here to make sure that it recurses correctly?

compiler/rustc_codegen_rmc/src/compiler_interface.rs
danielsn: Would also help make sure that temporaries were dropped as soon as the helper function exited.

scripts/std-lib-regression.sh
```diff
@@ -2,14 +2,15 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
danielsn, line 2: For git hygine, it would be better to put this as its own PR.

compiler/cbmc/src/irep/serialize.rs
celinval: sure.

compiler/rustc_codegen_rmc/src/compiler_interface.rs
celinval: ok

scripts/std-lib-regression.sh
celinval: For bash scripts I just use `vim`. :)

scripts/std-lib-regression.sh
celinval: Yeah, it's already in it's own PR (#593), but that PR was blocked by the github memory restriction. I could remove this commit from here, push this fix first and get back to the other PR later.

compiler/cbmc/src/irep/serialize.rs
celinval: I think you might be talking about the goto_program::Symbol, but this code is for the irep::Symbol which doesn't have any constructor.

danielsn (approved):

adpaco-aws (approved):

PR #624, opened 2021-11-04
Get rid of skip monomorphizer hack (#485)

### LINKED ISSUE #485: RMC should use rustc monomorphization as is

celinval:
### Description of changes: 

Remove the hack we added to monomorphizer to skip functions that we were not able to handle.

The latest fixes that we added to RMC was enough to allow us to run the monomorphizer as is. We no longer require the hooks to be moved to an mir_transformation. We can investigate that further before pushing to main.

### Resolved issues:

Resolves #485

### Testing:

* How is this change tested?

I added tests to the functions that were causing trouble before.

* Is this a refactor change?

### Checklist

### DIFF
--- a/src/test/rmc/Never/never_return.rs
+++ b/src/test/rmc/Never/never_return.rs
@@ -11,8 +11,7 @@ pub fn err() -> ! {
 // Give an empty main to make rustc happy.
 #[no_mangle]
 pub fn main() {
-    //let var = rmc::nondet::<i32>();
-    let var = 2;
+    let var = rmc::nondet::<i32>();
     if var > 0 {
         err();
     }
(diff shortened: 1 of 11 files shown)


src/test/rmc/Never/never_return.rs
```diff
@@ -0,0 +1,19 @@
...
+
+/// Test using the never type
+pub fn err() -> ! {
+    panic!("EXPECTED FAIL: Function should always fail");
+}
+
+// Give an empty main to make rustc happy.
+#[no_mangle]
+pub fn main() {
+    //let var = rmc::nondet::<i32>();
```
danielsn, line 14: why can't we do nondet

danielsn (approved):

src/test/rmc/Never/never_return.rs
celinval: oops... we can. I modified this test while reading the MIR and I forgot to change it back. Let me fix it.

PR #635, opened 2021-11-10
Run compiletest on rmc documentation code

### LINKED ISSUE #254: Training material

tedinski:
### Description of changes: 

A few findings:

1. Compiletest does weird things and seem very insistent on the convention of putting everything under `src/test/` so I created a symlink. (Github doesn't seem to indicate this very obviously, but that's that `src/test/rmc-docs` is.)
2. Its behavior seems to just be `find . -name "*.rs"` and compile each file separately, but that's fine for the doc tests, since we're only using the `proptest` dependency if `cfg(test)` which we aren't here.
3. All but one doc test is `rmc-verify-fail` but that's fine I think since we mostly want to ensure they still build correctly. It'd be nice if we could be more precise about why they fail, but that's not easy to do, so oh well

### Resolved issues:

Resolves #254 

### Testing:

* How is this change tested? it's new tests

* Is this a refactor change?

### Checklist


scripts/rmc-regression.sh
```diff
@@ -25,7 +25,7 @@ check-cbmc-viewer-version.py --major 2 --minor 5
 ./scripts/setup/build_rmc_lib.sh
 
 # Standalone rmc tests, expected tests, and cargo tests
-./x.py test -i --stage 1 rmc firecracker prusti smack expected cargo-rmc
+./x.py test -i --stage 1 rmc firecracker prusti smack expected cargo-rmc rmc-docs
```
adpaco-aws, line 28: I thought testing `rmc-docs` examples was going to be done by the `build-docs.sh` script (see the `TODO` comment there).

I'm okay with having it here but then we should remove the comment. Does the regression workflow need to change too? Please check.

adpaco-aws (changes requested):
LGTM modulo the comment below.

scripts/rmc-regression.sh
tedinski: Ah good point! I'll remove the comment and add another bit of documentation about how to run the tests for the docs.

The CI workflow runs the `rmc-regression.sh` script.

adpaco-aws: Right, but I recall having a discussion with you about the condition to execute `build-docs.sh`. Anyway, looks like it executes on Ubuntu 20.04 only - that is fine if we run them from `rmc-regression.sh`.

adpaco-aws (approved):

PR #643, opened 2021-11-13
Added a goto-instrument --drop-unused-functions step to rmc and cargo-rmc

### LINKED ISSUE #642: Lots of unrelated assertions reported on any test
I tried this code:

```rust
pub fn main() {
    let x: u8 = rmc::nondet();
    let y = x;
    assert!(x == y);
}
```

```
rmc test.rs
```

with RMC version: 497df3ba1eb

I expected to see this happen: One assertion is checked and reported.

Instead, this happened: RMC reports the results of 64 assertions, 63 of which are unrelated to the test:
```
/home/ubuntu/tools/rmc/library/core/src/fmt/mod.rs function std::fmt::Arguments::<'a>::new_v1
[std::fmt::Arguments::<'a>::new_v1.assertion.1] line 362 attempt to compute `move _13 + const 1_usize`, which would overflow: SUCCESS
[std::fmt::Arguments::<'a>::new_v1.assertion.2] line 362 invalid args: SUCCESS
[std::fmt::Arguments::<'a>::new_v1.overflow.1] line 362 arithmetic overflow on unsigned + in var_13 + 1: SUCCESS
[std::fmt::Arguments::<'a>::new_v1.pointer_arithmetic.1] line 365 pointer arithmetic: dead object in (unsigned char *)&var_17 + 0: SUCCESS
[std::fmt::Arguments::<'a>::new_v1.pointer_dereference.1] line 365 dereference failure: dead object in *((unsigned char **)((unsigned char *)&var_17 + 0)): SUCCESS

/home/ubuntu/tools/rmc/library/core/src/panic.rs function rmc::nondet
[rmc::nondet.assertion.1] line 39 a panicking function std::rt::panic_fmt is invoked: SUCCESS

<builtin-library-__CPROVER_assert> function __CPROVER_assert
[__CPROVER_assert.assertion.1] assertion false: SUCCESS

<builtin-library-__CPROVER_assume> function __CPROVER_assume
[__CPROVER_assume.assertion.1] assertion false: SUCCESS

<builtin-library-__builtin_powi> function __builtin_powi
[__builtin_powi.assertion.1] assertion false: SUCCESS

<builtin-library-__builtin_powif> function __builtin_powif
[__builtin_powif.assertion.1] assertion false: SUCCESS

<builtin-library-calloc> function calloc
[calloc.assertion.1] assertion false: SUCCESS

<builtin-library-ceil> function ceil
[ceil.assertion.1] assertion false: SUCCESS

<builtin-library-ceilf> function ceilf
[ceilf.assertion.1] assertion false: SUCCESS
...
<builtin-library-trunc> function trunc
[trunc.assertion.1] assertion false: SUCCESS

<builtin-library-truncf> function truncf
[truncf.assertion.1] assertion false: SUCCESS

** 0 of 64 failed (1 iterations)
VERIFICATION SUCCESSFUL
```
Even though all the extra asserts are successful, seeing them in the output is very distracting.

zhassan-aws:
### Description of changes: 

Add a `goto-instrument --drop-unused-functions` step to both `rmc` and `cargo rmc` to (conservatively) remove any irrelevant functions and their associated asserts/checks.

### Resolved issues:

Resolves #642 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


danielsn (approved):

scripts/cargo-rmc
```diff
@@ -95,6 +95,12 @@ def main():
             args.verbose, dry_run=args.dry_run):
         return 1
 
+    if EXIT_CODE_SUCCESS != rmc.run_goto_instrument(
+            cbmc_filename, cbmc_filename,
+            ['--drop-unused-functions'],
```
adpaco-aws, line 100: Is there a reason for not adding this argument to the previous `rmc.run_goto_instrument` call? What happens if you do?

zhassan-aws: Good point. I did not see any difference in behavior on the regressions whether I added it to the previous `run_goto_instrument` call or in a separate `run_goto_instrument` call. One thing I'm unsure about though is how `--drop-unused-functions` would interact with other options specified to `goto-instrument` like `--generate-function-body-options` or `--generate-function-body` in terms of the order in which they're applied.

adpaco-aws: Can you run a quick experiment to check how many functions are dropped in each case? I would use a project larger than the ones in our regression.

adpaco-aws:
Isn't the `never_return.rs` test expected to fail at runtime? It does so at least when running it in `rustc`. If that is the case, you could add `// rmc-verify-fail` which checks that you can codegen it but not verify it.

zhassan-aws:
> Isn't the `never_return.rs` test expected to fail at runtime?

Yes, it fails with and without `--drop-unused-functions`. The reason the regression passes without it though is that the CBMC output includes `EXPECTED_FAIL` from the panic inside `err`:
```
/home/ubuntu/tools/rmc/library/std/src/panic.rs function err
[err.assertion.1] line 21 EXPECTED FAIL: Function should always fail: SUCCESS
```
which makes `runtest.rs` consider this test to be an expected failure. With `--drop-unused-functions`, the `EXPECTED_FAIL` does not appear in the output, likely because the `err` function has been emptied out since it never returns.

celinval:
> > Isn't the `never_return.rs` test expected to fail at runtime?
> 
> Yes, it fails with and without `--drop-unused-functions`. The reason the regression passes without it though is that the CBMC output includes `EXPECTED_FAIL` from the panic inside `err`:
> 
> ```
> /home/ubuntu/tools/rmc/library/std/src/panic.rs function err
> [err.assertion.1] line 21 EXPECTED FAIL: Function should always fail: SUCCESS
> ```
> 
> which makes `runtest.rs` consider this test to be an expected failure. With `--drop-unused-functions`, the `EXPECTED_FAIL` does not appear in the output, likely because the `err` function has been emptied out since it never returns.

@zhassan-aws and I inspected the test and the code. During codegen we don't generate the call to the function that returns never. We add a check instead. goto-instrument is correctly identifying it as dead code and dropping the function. I think the correct thing to do here is mark the test as verify fail and add an issue to track the problem.

tedinski:
> With --drop-unused-functions, the EXPECTED_FAIL does not appear in the output, likely because the err function has been emptied out since it never returns.

Do you have a path to resolution on this? This seems weird, not at all expected imo. Wonder if it's exposing some other bug.

tedinski:
So it is a bug, right? We should be generating those calls?

zhassan-aws:
> So it is a bug, right? We should be generating those calls?

Yes. Created #648.

The resolution as @celinval suggested is to add `verify_fail` to the test.

celinval (approved):

scripts/cargo-rmc
zhassan-aws: Thanks for the suggestion. I tried on a couple of bigger tests, and there was no difference in the number of dropped functions reported by `goto-instrument`. The CBMC output had some minor differences that did not seem significant. So overall, it seems that there is no difference in behavior between applying `--drop-unused-functions` in a separate `goto-instrument` call versus applying it in the previous one. With that said, I noticed that the `goto-instrument --drop-unused-functions` call I added to the `rmc` script is incorrectly put under `if args.undefined_function_checks`, while we want it to run independent of that option. The equivalent code in `cargo-rmc` is calling `goto-instrument --add-library` and `goto-instrument --generate-function-body-options assert-false ...` unconditionally, which I think should be fixed?

adpaco-aws (approved):

Issue #646, opened 2021-11-15
Implement arithmetic checks for unchecked operations


celinval:
**Requested feature:** RMC should detect failures when dealing with unchecked arithmetic operations.
**Use case:** Unchecked arithmetic operations assume that the operands respect some invariant. If they don't, the result has an undefined behavior. We should add checks to verify that all the invariants are respected. 
**Link to relevant documentation (Rust reference, Nomicon, RFC):**
https://doc.rust-lang.org/std/primitive.u32.html
https://doc.rust-lang.org/std/primitive.i32.html
https://doc.rust-lang.org/std/primitive.f32.html#method.to_int_unchecked

**Is this a breaking change?** No

celinval:
Removing soundness since we don't claim support to all UBs related to arithmetic operations.

PR #647, opened 2021-11-15
Fix spurious overflow counter examples. (#558)

### LINKED ISSUE #558: Spurious overflow failures with Result/Error types
I tried this code:

```rust
enum MyError {
    Val1,
    Val2,
}

fn foo(x: u32) -> Result<(), MyError> {
    if x > 10 {
        Err(MyError::Val2)
    } else {
        Ok(())
    }
}

fn bar() -> Result<(), MyError> {
    let x = foo(15)?;

    Ok(x)
}

pub fn main() {
    bar();
}
```

```
rmc test.rs
```

with RMC version: `7635cfebfa5`

I expected to see this happen: no failures

Instead, this happened: got 2 arithmetic overflow failures:
```
** Results:
/home/ubuntu/tools/rmc/library/core/src/result.rs function <std::result::Result<T, E> as std::ops::Try>::branch
[<std::result::Result<T, E> as std::ops::Try>::branch.assertion.1] line 1903 unreachable code: SUCCESS
[<std::result::Result<T, E> as std::ops::Try>::branch.overflow.1] line 1903 arithmetic overflow on unsigned - in ((unsigned char *)&self)[0] - 2: FAILURE
[<std::result::Result<T, E> as std::ops::Try>::branch.pointer_arithmetic.1] line 1903 pointer arithmetic: dead object in (unsigned char *)&self + 0: SUCCESS
[<std::result::Result<T, E> as std::ops::Try>::branch.pointer_dereference.1] line 1903 dereference failure: dead object in ((unsigned char *)&self)[0]: SUCCESS
[<std::result::Result<T, E> as std::ops::Try>::branch.pointer_arithmetic.2] line 1904 pointer arithmetic: dead object in (unsigned char *)&var_0 + 0: SUCCESS
[<std::result::Result<T, E> as std::ops::Try>::branch.pointer_dereference.2] line 1904 dereference failure: dead object in ((unsigned char *)&var_0)[0]: SUCCESS

/home/ubuntu/tools/rmc/src/test/rmc/Enum/result4.rs function bar
[bar.assertion.1] line 18 unreachable code: SUCCESS
[bar.overflow.1] line 18 arithmetic overflow on unsigned - in ((unsigned char *)&var_2)[0] - 2: FAILURE
[bar.pointer_arithmetic.1] line 18 pointer arithmetic: dead object in (unsigned char *)&var_2 + 0: SUCCESS
[bar.pointer_dereference.1] line 18 dereference failure: dead object in ((unsigned char *)&var_2)[0]: SUCCESS
[bar.pointer_arithmetic.2] line 20 pointer arithmetic: dead object in (unsigned char *)&var_0 + 0: SUCCESS
[bar.pointer_dereference.2] line 20 dereference failure: dead object in ((unsigned char *)&var_0)[0]: SUCCESS

/home/ubuntu/tools/rmc/src/test/rmc/Enum/result4.rs function foo
[foo.pointer_arithmetic.1] line 13 pointer arithmetic: dead object in (unsigned char *)&var_0 + 0: SUCCESS
[foo.pointer_dereference.1] line 13 dereference failure: dead object in ((unsigned char *)&var_0)[0]: SUCCESS

** 2 of 14 failed (2 iterations)
VERIFICATION FAILED
```
The `- 2` in the check:
```
[bar.overflow.1] line 18 arithmetic overflow on unsigned - in ((unsigned char *)&var_2)[0] - 2: FAILURE
```
seems to be coming from the number of values in the enum. The arithmetic overflow checks are not expected in the first place.

celinval:
### Description of changes: 

RMC was treating all arithmetic operations as non-wrapping which was triggering a lot of spurious counter examples. Instead of relying on cbmc generated overflow checks, we are not explicitly encoding checks for non-wrapping operations.

### Resolved issues:

Resolves #558 

### Call-outs:

I created #646 to track implementing more checks that we currently don't support. Except for shift left and division by 0, I believe CBMC doesn't have any mechanism to check them.

### Testing:

* How is this change tested?

I created a bunch of new tests.

* Is this a refactor change?

I guess.

### Checklist

### DIFF
--- a/compiler/cbmc/src/goto_program/expr.rs
+++ b/compiler/cbmc/src/goto_program/expr.rs
@@ -225,7 +225,7 @@ pub enum UnaryOperand {
     UnaryMinus,
 }
 
-/// The return type for `__builtin_op_overflow` operations
+/// The return type for `__CPROVER_overflow_op` operations
 pub struct ArithmeticOverflowResult {
     /// If overflow did not occur, the result of the operation. Otherwise undefined.
     pub result: Expr,
@@ -815,7 +815,7 @@ impl Expr {
                     && (lhs.typ.is_pointer() || lhs.typ.is_numeric() || lhs.typ.is_vector())
                     || (lhs.typ.is_pointer() && rhs.typ.is_integer())
             }
-            Plus | OverflowMinus | OverflowMult | OverflowPlus => {
+            Plus => {
                 (lhs.typ == rhs.typ && (lhs.typ.is_numeric() || lhs.typ.is_vector()))
                     || (lhs.typ.is_pointer() && rhs.typ.is_integer())
             }
@@ -848,6 +848,15 @@ impl Expr {
             }
             // Floating Point Equalities
             IeeeFloatEqual | IeeeFloatNotequal => lhs.typ == rhs.typ && lhs.typ.is_floating_point(),
+            // Overflow flags
+            OverflowMinus => {
+                (lhs.typ == rhs.typ && (lhs.typ.is_pointer() || lhs.typ.is_numeric()))
+                    || (lhs.typ.is_pointer() && rhs.typ.is_integer())
+            }
+            OverflowMult | OverflowPlus => {
+                (lhs.typ == rhs.typ && lhs.typ.is_integer())
+                    || (lhs.typ.is_pointer() && rhs.typ.is_integer())
+            }
         }
     }
 
--- a/compiler/cbmc/src/goto_program/expr.rs
+++ b/compiler/cbmc/src/goto_program/expr.rs
@@ -815,7 +815,7 @@ impl Expr {
                     && (lhs.typ.is_pointer() || lhs.typ.is_numeric() || lhs.typ.is_vector())
                     || (lhs.typ.is_pointer() && rhs.typ.is_integer())
             }
-            Plus => {
+            Plus | OverflowMinus | OverflowMult | OverflowPlus => {
                 (lhs.typ == rhs.typ && (lhs.typ.is_numeric() || lhs.typ.is_vector()))
                     || (lhs.typ.is_pointer() && rhs.typ.is_integer())
             }
@@ -848,10 +848,6 @@ impl Expr {
             }
             // Floating Point Equalities
             IeeeFloatEqual | IeeeFloatNotequal => lhs.typ == rhs.typ && lhs.typ.is_floating_point(),
-            // Overflow flags
-            OverflowMinus | OverflowMult | OverflowPlus => {
-                lhs.typ == rhs.typ && lhs.typ.is_integer()
-            }
         }
     }
 
--- a/compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs
+++ b/compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs
@@ -146,10 +146,9 @@ impl<'tcx> GotocCtx<'tcx> {
             }};
         }
 
-        // Intrinsics which encode a SIMD arithmetic operation with overflow check
-        // RMC overflow operations accept SIMD vectors but the results are unsound.
-        // TODO: Create a CBMC issue.
-        // For now, we expand the overflow check.
+        // Intrinsics which encode a SIMD arithmetic operation with overflow check.
+        // We expand the overflow check because CBMC overflow operations don't accept array as
+        // argument.
         macro_rules! codegen_simd_with_overflow_check {
             ($op:ident, $overflow:ident) => {{
                 let a = fargs.remove(0);
@@ -182,7 +181,7 @@ impl<'tcx> GotocCtx<'tcx> {
 
         // Intrinsics which encode a simple wrapping arithmetic operation
         macro_rules! codegen_wrapping_op {
-            ($f:ident) => {{ self.binop(p, fargs, |a, b| a.$f(b)) }};
+            ($f:ident) => {{ codegen_intrinsic_binop!($f) }};
         }
 
         // Intrinsics which encode a simple binary operation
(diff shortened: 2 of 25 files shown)


compiler/cbmc/src/goto_program/expr.rs
```diff
@@ -815,7 +815,7 @@ impl Expr {
                     && (lhs.typ.is_pointer() || lhs.typ.is_numeric() || lhs.typ.is_vector())
                     || (lhs.typ.is_pointer() && rhs.typ.is_integer())
             }
-            Plus => {
+            Plus | OverflowMinus | OverflowMult | OverflowPlus => {
```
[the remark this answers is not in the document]
adpaco-aws: The argument types have to remain the same for each operation (i.e., `OverflowMinus` goes with `Minus`).

compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs
```diff
@@ -130,6 +130,61 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    loc,
+                );
+                let expr_place = self.codegen_expr_to_place(p, res.result);
+                Stmt::block(vec![expr_place, check], loc)
+            }};
+        }
+
+        // Intrinsics which encode a SIMD arithmetic operation with overflow check
+        // RMC overflow operations accept SIMD vectors but the results are unsound.
+        // TODO: Create a CBMC issue.
```
adpaco-aws, line 151: Please create the issue and add it here.

src/test/rmc/ArithOperators/wrapping.rs
```diff
@@ -0,0 +1,15 @@
...
+#![feature(core_intrinsics)]
+
+pub fn main() {
+    let a: u8 = rmc::nondet();
+    let b: u8 = rmc::nondet();
+    let sum0 = core::intrinsics::wrapping_add(a, b);
+    let sum1 = a.wrapping_add(b);
+    let sum2 = a.checked_add(b);
+    assert!(sum0 == sum1);
+    assert!(sum1 >= a || sum2.is_none());
```
zhassan-aws, line 14: Do you want to add `sum1 >= b` as well?

zhassan-aws (approved):
Looks great! The new tests are very good!

compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs
```diff
@@ -130,6 +130,61 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                let res = a.$op(b);
+                let expr_place = self.codegen_expr_to_place(p, res);
+                Stmt::block(vec![expr_place, check_stmt], loc)
+            }};
+        }
+
+        // Intrinsics which encode a simple wrapping arithmetic operation
+        macro_rules! codegen_wrapping_op {
+            ($f:ident) => {{ self.binop(p, fargs, |a, b| a.$f(b)) }};
+        }
```
adpaco-aws, line 186: A few lines below there is `codegen_intrinsic_binop` which does the same thing. Why did you add this?

adpaco-aws (changes requested):
The changes here heavily rely on `-C overflow-checks=on`. Can we add a check to our back-end to ensure the flag is always set?

celinval:
> The changes here heavily rely on `-C overflow-checks=on`. Can we add a check to our back-end to ensure the flag is always set?

Should we allow users to turn them off though? What if I add a check with a warning if the flag is off?

compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs
celinval: I was trying to make it explicit that this is a wrapping operation. To avoid others to inadvertently using the wrong operation in the future.

I could modify this to just call the binop one or remove this. Any preference?

compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs
celinval: Oops. Yeah, I need to do that.

compiler/cbmc/src/goto_program/expr.rs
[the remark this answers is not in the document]
celinval: Makes sense. Do you know why the minus has a different retirement?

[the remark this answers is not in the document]
adpaco-aws: Yes, minus allows both operands to be pointers because it is valid to compute offsets between them. This does not apply to plus because adding two pointers will likely result in another pointer to who-knows-where.

compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs
celinval: See my comment above. I did some digging into CBMC and it doesn't look like overflow takes array as argument. I hadn't noticed that CBMC was printing a warning. I wish it would just error out instead.

Anyway, I will update the comment to reflect that.

adpaco-aws (commented):
Looking good! Of course, we can change the check in the near future to support other use cases, but I do not feel comfortable with our codegen depending on this flag and not requiring it.

adpaco-aws (approved):

tedinski:
Doesn't this mean we no longer spot overflows in unsafe code?

celinval:
> Doesn't this mean we no longer spot overflows in unsafe code?

That's a great question! I should definitely add test cases for that.

We are still checking for overflow in unsafe code. Regular arithmetic operations are always checked no matter where they show up in the code.

celinval:
> This should be `check_mul`

I wonder how this test even passed. o.O

adpaco-aws:
> I wonder how this test even passed. o.O

Probably because CBMC is not finding the function to verify and that is understood as a verification failure.

PR #653, opened 2021-11-18
Inherit parent environment when running cargo build

zhassan-aws:
### Description of changes: 

The `cargo rmc` script currently does not pass the environment to the `cargo build` command (unlike the `rmc` script which passes the environment down to the compile command). Thus doing `RUST_BACKTRACE=1 cargo rmc` doesn't result in enabling the backtrace as one would expect.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/scripts/rmc.py
+++ b/scripts/rmc.py
@@ -269,8 +269,7 @@ def cargo_build(
         build_cmd += ["--target", str(build_target)]
     build_env = os.environ
     build_env.update({"RUSTFLAGS": " ".join(rustflags),
-                      "RUSTC": rustc_path,
-                      "PATH": os.environ["PATH"]
+                      "RUSTC": rustc_path
                       })
     if debug:
         add_rmc_rustc_debug_to_env(build_env)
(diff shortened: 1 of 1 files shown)


scripts/rmc.py
```diff
@@ -267,10 +267,11 @@ def get_config(option):
...
     if build_target:
         build_cmd += ["--target", str(build_target)]
-    build_env = {"RUSTFLAGS": " ".join(rustflags),
-                 "RUSTC": rustc_path,
-                 "PATH": os.environ["PATH"]
-                 }
+    build_env = os.environ
+    build_env.update({"RUSTFLAGS": " ".join(rustflags),
+                      "RUSTC": rustc_path,
+                      "PATH": os.environ["PATH"]
```
adpaco-aws, line 273: Isn't `"PATH"` already `os.environ["PATH"]`?

zhassan-aws: Good catch! I'll remove it.

adpaco-aws (approved):

PR #654, opened 2021-11-18
Dashboard: Avoid "field is never read" warnings

### LINKED ISSUE #567: Dashboard example generated with 2 `pub` declarations
The example `The Unstable Book/Language Features/trait_alias/14.rs` fails to check because there are two `pub` declarations before the `main` function. Find what the root cause is and fix it.

adpaco-aws:
### Description of changes: 

Declare each Litani struct field as public to remove "field is never read" warnings.

### Testing:

* How is this change tested? Running dashboard.

### Checklist

### DIFF
--- a/src/tools/dashboard/src/litani.rs
+++ b/src/tools/dashboard/src/litani.rs
@@ -17,19 +17,19 @@ use std::process::{Child, Command};
 /// extend it if advanced features are used (e.g., pools)
 #[derive(Debug, Deserialize)]
 pub struct LitaniRun {
-    aux: Option<HashMap<String, String>>,
-    project: String,
-    version: String,
-    version_major: u32,
-    version_minor: u32,
-    version_patch: u32,
-    release_candidate: bool,
-    run_id: String,
-    start_time: String,
-    parallelism: LitaniParalellism,
-    latest_symlink: Option<String>,
-    end_time: String,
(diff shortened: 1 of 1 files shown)


src/tools/dashboard/src/books.rs
```diff
@@ -283,7 +283,10 @@ fn prepend_props(path: &Path, example: &mut Example, config_paths: &mut HashSet<
 
 /// Make the main function of a test public so it can be verified by rmc.
 fn pub_main(code: String) -> String {
-    code.replace("fn main", "pub fn main")
```
celinval, line 286: I don't think we need this hack anymore. We should be able to verify binary crates now.

celinval: https://github.com/model-checking/rmc/pull/656

adpaco-aws: Thanks!! I have left a review there and hope we can merge it soon. Should we reuse this PR for the warning changes or close it and open a new one with them?

celinval: We can reuse it. Maybe just make sure you edit the title!

adpaco-aws: Updated PR to contain only the warnings changes.

celinval (approved):

PR #656, opened 2021-11-19
Re-enable RMC to allow users to verify non-public main.

### LINKED ISSUE #655: Re-enable support to binary crates verification
We should be able to verify bin crates when the verification target is the main function. I.e., user should be able to verify the following program:

```rust
// my_main.rs
fn main() {
    // code goes here.
}
```

```
rmc my_main.rs
```

with RMC version:

I expected to see this happen: RMC verifies the main function.

Instead, this happened: RMC cannot find main function because it is not marked as public.

celinval:
Users don't usually set their main function to be public and we should enable them to verify their main function. For now, users will have to manually set the create type when running RMC directly. For cargo rmc, this should be automatically set up by cargo.

### Resolved issues:

Resolves #655 

### Call-outs:

We should figure out what is the default behavior of RMC and we might want to add a mechanism to figure out which one to use based on the user configuration. For example, maybe we can check what is the target function for the verification.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/src/test/expected/one-assert/test.rs
+++ b/src/test/expected/one-assert/test.rs
@@ -1,7 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-
-pub fn main() {
+// rmc-flags: --function check_assert
+// compile-flags: --crate-type lib
+#[no_mangle]
+pub fn check_assert() {
     let x: u8 = rmc::nondet();
     let y = x;
     assert!(x == y);
--- a/src/test/rmc/ArithOperators/main.rs
+++ b/src/test/rmc/ArithOperators/main.rs
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-pub fn main() {
+fn main() {
     let a: u32 = rmc::nondet();
     assert!(a / 2 <= a);
     assert!(a / 2 * 2 >= a / 2);
--- a/src/tools/compiletest/src/runtest.rs
+++ b/src/tools/compiletest/src/runtest.rs
@@ -2273,26 +2273,7 @@ impl<'test> TestCx<'test> {
     /// Runs RMC on the test file specified by `self.testpaths.file`. An error
     /// message is printed to stdout if the verification result is not expected.
     fn verify(&self) {
-        // Other modes call self.compile_test(...). However, we cannot call it here for two reasons:
-        // 1. It calls rustc instead of RMC
-        // 2. It may pass some options that do not make sense for RMC
-        // So we create our own command to execute RMC and pass it to self.compose_and_run_compiler(...) directly.
-        let mut rmc = Command::new("rmc");
-        // We cannot pass rustc flags directly to RMC. Instead, we add them
-        // to the current environment through the `RUSTFLAGS` environment
-        // variable. RMC recognizes the variable and adds those flags to its
-        // internal call to rustc.
-        if !self.props.compile_flags.is_empty() {
-            rmc.env("RUSTFLAGS", self.props.compile_flags.join(" "));
-        }
-        // Pass the test path along with RMC and CBMC flags parsed from comments at the top of the test file.
-        rmc.args(&self.props.rmc_flags)
-            .arg("--input")
-            .arg(&self.testpaths.file)
-            .arg("--cbmc-args")
-            .args(&self.props.cbmc_flags);
-        self.add_rmc_dir_to_path(&mut rmc);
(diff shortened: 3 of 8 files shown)


src/tools/dashboard/src/books.rs
```diff
@@ -296,17 +291,15 @@ fn extract(par_from: &Path, par_to: &Path, config_paths: &mut HashSet<PathBuf>)
...
-                None,
-            )
-            .0,
-        );
+        example.code = rustdoc::doctest::make_test(
+            &example.code,
+            None,
+            false,
+            &Default::default(),
+            example.config.edition.unwrap_or(Edition::Edition2018),
```
adpaco-aws, line 299: The default edition was set to 2021 in #652, so this should be:
```suggestion
            example.config.edition.unwrap(),
```

src/test/expected/one-assert/test.rs
```diff
@@ -1,7 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-
-pub fn main() {
+// rmc-flags: --function check_assert
+// compile-flags: --crate-type lib
```
adpaco-aws, line 4: Why is this needed?

src/test/rmc/ArithOperators/main.rs
```diff
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-pub fn main() {
+fn main() {
```
adpaco-aws, line 3: Isn't this change needed for each test as in #523?

src/tools/compiletest/src/runtest.rs
```diff
@@ -2414,16 +2410,7 @@ impl<'test> TestCx<'test> {
...
-        // Arguments to choose specific abstraction are currently provided as
-        // as rmc-flags in the test file
-        rmc.args(&self.props.rmc_flags)
-            .arg("--input")
-            .arg(&self.testpaths.file)
-            .arg("--cbmc-args")
-            .args(&self.props.cbmc_flags);
-        self.add_rmc_dir_to_path(&mut rmc);
-        let proc_res = self.compose_and_run_compiler(rmc, None);
+        let proc_res = self.run_rmc();
```
adpaco-aws, line 2413: Thanks for refactoring this!

adpaco-aws (changes requested):
Does not look ready to me, see comments below. Also, how it comes we are able to codegen binary crates now? The changes in RMC scripts remove all lib-related flags, I do not see any major change anywhere in this PR.

src/test/expected/one-assert/test.rs
celinval: This is needed because binary crates have more checks than library crates. The purpose of this test is to ensure that there is only one assertion in the result.

src/test/rmc/ArithOperators/main.rs
celinval: No. It's not needed as in #523. With this change RMC will accept both public and non public main.

celinval:
> Does not look ready to me, see comments below. Also, how it comes we are able to codegen binary crates now? The changes in RMC scripts remove all lib-related flags, I do not see any major change anywhere in this PR.

Previous fixes is what enabled us to be able to codegen binary crates. This PR is just to remove the hack that we were enforcing the crate type to be a lib.

src/test/rmc/ArithOperators/main.rs
adpaco-aws: Yes, I understand it is not strictly need but nice to have. Anyway, let me take care of that in #659

adpaco-aws (approved):

src/test/rmc/ArithOperators/main.rs
celinval: Thanks

src/test/expected/one-assert/test.rs
zhassan-aws: I'm curious what the extra checks for binary crates are.

zhassan-aws (approved):

src/tools/compiletest/src/runtest.rs
```diff
@@ -2273,26 +2273,7 @@ impl<'test> TestCx<'test> {
...
-        }
-        // Pass the test path along with RMC and CBMC flags parsed from comments at the top of the test file.
-        rmc.args(&self.props.rmc_flags)
-            .arg("--input")
-            .arg(&self.testpaths.file)
-            .arg("--cbmc-args")
-            .args(&self.props.cbmc_flags);
-        self.add_rmc_dir_to_path(&mut rmc);
-        let proc_res = self.compose_and_run_compiler(rmc, None);
+        let proc_res = self.run_rmc();
```
danielsn, line 2276: 💯

PR #657, opened 2021-11-19
Split rmc::nondet into safe and unsafe functions(#607)

### LINKED ISSUE #607: Provide safe and unsafe functions to generate symbolic values
The return value of `rmc::nondet` doesn't always respect the invariant of a type. This can be confusing for a user which may want to cover the search space for all valid values of a type or for all possible values (valid and invalid).

We should create two different functions for each use case and ideally a way for a user to specify invariant of their own types.

celinval:
### Description of changes: 

The return value of rmc::nondet doesn't always respect the invariant of a type which could be confusing for the user. To fix the issue, we decided to provide two functions that create symbolic variables and one new trait that allows user to specify invariant of a type. The `rmc::any::<T>()` function generates a value that must satisfy the type invariant (specified by the new `rmc::Invariant` trait), while the unsafe function, `rmc::any_raw::<T>()` one generates an unconstrained return value which maps to any possible bit value assignment.

### Resolved issues:

Resolves #607 

### Call-outs:

1. I kept the `rmc::nondet::<T>()` for now so I can finish updating the tests in a follow up PR.
2. I am still adding tests to the implementations of the Invariant. I'll update the PR once that it's done.

### Testing:

* How is this change tested? I modified all occurrences of the function in our tests. 

### Checklist


camshaft (approved):
Nice! I think this is the correct approach.

celinval:
A few folks have brought up some concerns with the message `unsafe {}` statement actually provides. Since for some cases RMC does provide assurance that the value is valid.

Instead we decided to create two different functions, one that is safe and it should only represent valid values and another one that we will mark with unsafe that represents all possible bit value for the memory occupied by the object.

camshaft:
> Instead we decided to create two different functions, one that is safe and it should only represent valid values and another one that we will mark with unsafe that represents all possible bit value for the memory occupied by the object.

Is there going to be a marker trait for the safe function? If so that should be `unsafe` to implement (similar to [Send and Sync](https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html)).

src/test/rmc/BinOp_Offset/main_fail.rs
```diff
@@ -4,8 +4,8 @@
 
 pub fn test_offset_in_double_array() {
     //let table: Vec<Vec<u64>> = Vec::with_capacity(1);
-    let table: [[u64; 1]; 1] = [[rmc::nondet::<u64>()]];
-    table[0][rmc::nondet::<usize>()]; // EXPECTED FAIL
+    let table: [[u64; 1]; 1] = [[rmc::any::<u64>()]];
+    table[0][rmc::any::<usize>()]; // EXPCECTED FAIL
```
adpaco-aws, line 8:
```suggestion
    table[0][rmc::any::<usize>()]; // EXPECTED FAIL
```

src/test/expected/one-assert/expected
```diff
@@ -1 +1 @@
-** 0 of 1 failed (1 iterations)
+** 0 of 2 failed (1 iterations)
```
adpaco-aws, line 1: What is going on here? This was supposed to have a single assertion 😆

library/rmc/src/invariant.rs
```diff
@@ -0,0 +1,143 @@
...
+
+unsafe impl Invariant for isize {
+    #[inline(always)]
+    fn is_valid(&self) -> bool {
+        true
+    }
+}
+
+/// Verifies that:
+/// - a char is not outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
```
adpaco-aws, line 108:
```suggestion
/// Validate that a char is not outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
```

library/rmc/src/invariant.rs
```diff
@@ -0,0 +1,162 @@
...
+    fn is_valid(&self) -> bool {
+        if let Ok(v) = self {
+            v.is_valid()
+        } else if let Err(e) = self {
+            e.is_valid()
+        } else {
+            false
+        }
+    }
+}
```
zhassan-aws, line 162: The `assumptions` module handled a lot more types than the ones implemented here, e.g. enums. Do we lose assumptions for those other types?

src/test/firecracker/virtio-block-parse/main.rs
```diff
@@ -42,15 +46,23 @@ impl GuestMemoryMmap {
             if let Some(prev_addr) = TRACK_READ_OBJ {
                 assert!(prev_addr.0 != addr.0);
             }
-            if rmc::nondet() && TRACK_READ_OBJ.is_none() {
+            if rmc::any() && TRACK_READ_OBJ.is_none() {
                 TRACK_READ_OBJ = Some(addr);
             }
         }
-        rmc::nondet()
+        if rmc::any() {
```
zhassan-aws, line 53: Is this change maintaining the old behavior?

src/test/rmc/DynTrait/main.rs
```diff
@@ -34,10 +29,10 @@ fn random_animal(random_number: i64) -> Box<dyn Animal> {
 }
 
 fn main() {
-    let random_number = rmc::nondet();
+    let random_number = rmc::any();
     let animal = random_animal(random_number);
     let s = animal.noise();
-    if random_number < 5 {
+    if (random_number < 5) {
```
zhassan-aws, line 35: Changed by mistake?

src/test/rmc/Pointers_OtherTypes/main.rs
```diff
@@ -21,7 +21,7 @@ fn main() {
     make_true(&mut z);
     assert!(z);
 
-    let mut a: f32 = rmc::nondet();
+    let mut a: f32 = unsafe { rmc::any_raw() };
```
zhassan-aws, line 24: Should this be `rmc::any()`?

src/test/rmc/TypeInvariant/char.rs_ignore
```diff
@@ -0,0 +1,9 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
zhassan-aws, line 1: Why is this test `ignore`?

zhassan-aws (commented):
Looks great! What is unclear to me is my comment on the removal of the `assumptions` module.

src/test/rmc/BinOp_Offset/main_fail.rs
danielsn, line 8: Is the turbofish needed here?

src/test/rmc/Cast/from_be_bytes.rs
```diff
@@ -3,14 +3,14 @@
...
-        rmc::nondet(),
+        rmc::any(),
+        rmc::any(),
+        rmc::any(),
+        rmc::any(),
+        rmc::any(),
+        rmc::any(),
+        rmc::any(),
+        rmc::any(),
     ];
```
danielsn, line 14: Could we use `any_slice`

src/test/rmc/Enum/result3.rs
```diff
@@ -11,10 +11,10 @@ fn foo(input: &Result<u32, Unit>) -> u32 {
 }
 
 fn main() {
-    let input: Result<u32, Unit> = rmc::nondet();
+    let input: Result<u32, Unit> = unsafe { rmc::any_raw() };
     let x = foo(&input);
     assert!(x == 3 || input != Err(Unit::Unit));
-    let input: Result<u32, Unit> = if rmc::nondet() { Ok(0) } else { Err(Unit::Unit) };
+    let input: Result<u32, Unit> = if unsafe { rmc::any_raw() } { Ok(0) } else { Err(Unit::Unit) };
```
danielsn, line 17: why raw

src/test/firecracker/micro-http-parsed-request/ignore-main.rs
```diff
@@ -5,26 +5,26 @@
 
 // Should return a nondet string of up to n characters
 // Currently RMC does not support strings
-fn rmc::nondet_string(n: u32) -> String {
+fn rmc::any_string(n: u32) -> String {
```
danielsn, line 8: Should this belong in the prelude?

library/rmc/src/invariant.rs
```diff
@@ -0,0 +1,143 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+/// This module introduces the Invariant trait and as well as implementation for commonly used types.
```
danielsn, line 4: Are there any standard types missing here?

celinval: From the list of primitive types [here](https://doc.rust-lang.org/std/index.html#primitives), the types missing are:

- never
- fn
- pointer
- reference
- slice
- str
- tuple
- unit

Except for tuple, I don't see a use case for invariant. The tuple is a bit more painful to generate since rust doesn't have variadic templates.

In the next revision, I will include array which I was missing.

The standard library have a lot of other types that we are not implementing yet. Things with pointers and references are a bit trickier to implement. I think having a builder pattern for them will make more sense.

library/rmc/src/invariant.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// Types that implement a check to ensure its value is valid and safe to be used. See
+/// https://doc.rust-lang.org/stable/nomicon/what-unsafe-does.html for examples of valid values.
+///
+/// Implementations of Invariant traits must ensure that the current bit values of the given type
+/// is valid and that all its invariants hold.
+pub unsafe trait Invariant {
```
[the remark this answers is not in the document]
celinval: That was a suggestion from @camshaft which I thought made a lot of sense. So I would prefer keeping it.
> 
> Is there going to be a marker trait for the safe function? If so that should be `unsafe` to implement (similar to [Send and Sync](https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html)).

library/rmc/src/invariant.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn is_valid(&self) -> bool {
+        let byte = u8::from(*self);
+        byte == 0 || byte == 1
+    }
+}
+
+unsafe impl Invariant for u8 {
+    #[inline(always)]
+    fn is_valid(&self) -> bool {
+        true
```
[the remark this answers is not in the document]
celinval: I'll generate these with a macro. I don't know if we should provide a default implementation that returns `true`. Is the macro option OK?

library/rmc/src/invariant.rs
celinval: Yes, we do. For enums and structs, @danielsn and I were thinking about adding a custom derive: #726

For other cases, I would prefer being safe. I don't think we should infer anything based on the type of fields without any sort of user intervention.

library/rmc/src/slice.rs
```diff
@@ -53,8 +53,8 @@ pub struct NonDetSlice<T, const MAX_SLICE_LENGTH: usize> {
 
 impl<T, const MAX_SLICE_LENGTH: usize> NonDetSlice<T, MAX_SLICE_LENGTH> {
     fn new() -> Self {
-        let arr: [T; MAX_SLICE_LENGTH] = nondet();
-        let slice_len: usize = nondet();
+        let arr: [T; MAX_SLICE_LENGTH] = unsafe { any_raw() };
```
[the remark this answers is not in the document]
celinval: Yes. I am adding a variant of that. Thanks!

src/test/expected/one-assert/expected
celinval: I know... we just can't keep this going, eh? We really need that report filter. This is due to a compiler sanity check. I could just use `any_raw` here to keep the 1 assertion.

src/test/firecracker/micro-http-parsed-request/ignore-main.rs
celinval: We don't currently support generating strings. We should consider doing something similar to what @zhassan-aws did for slices. But I think this is out of the scope of this PR.

src/test/rmc/Pointers_OtherTypes/main.rs
celinval: Yes. Sorry, I fixed this before I implemented the Invariant for f32.

src/test/rmc/TypeInvariant/char.rs_ignore
celinval: oh... I think that was a mistake. :)

src/test/rmc/i32-Unary-/main.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 fn main() {
-    let a: i32 = rmc::nondet();
+    let a: i32 = rmc::any();
     if -100000 < a && a < 100000 {
```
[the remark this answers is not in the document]
celinval: Can you please clarify?

src/test/smack/structures/point_fail.rs
```diff
@@ -38,10 +38,10 @@ impl AddAssign for Point {
...
-    let x = rmc::nondet();
-    let y = rmc::nondet();
-    let z = rmc::nondet();
+    let w = rmc::any();
+    let x = rmc::any();
+    let y = rmc::any();
+    let z = rmc::any();
 
     if w <= std::u64::MAX / 2 // avoid overflow
         && x <= std::u64::MAX / 2 // avoid overflow
```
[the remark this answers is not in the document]
celinval: Yes. I think so.

src/test/stub-tests/HashSet/concrete.rs
```diff
@@ -8,9 +8,9 @@ fn main() {
     // TODO: This test should ideally work with nondeterminstic values but for
     // for the moment it does not.
```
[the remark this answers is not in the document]
celinval: Have we enabled these tests? I thought the stub-tests weren't running.

src/test/stub-tests/HashSet/ignore-nondet.rs
```diff
@@ -7,9 +7,9 @@ fn main() {
 
     // TODO: This test should ideally work with nondeterminstic values but for
     // for the moment it does not.
```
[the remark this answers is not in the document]
celinval: I don't know.

library/rmc/src/invariant.rs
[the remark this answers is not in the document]
camshaft: I was originally thinking it would be something like:

```rust
/// Types that implement this can be generated without any constraints or invariants
pub unsafe trait NondetTransmute { }

pub fn any<T: NondetTransmute>() -> T {
    unsafe { any_raw() }
}
```

You would then implement that for the primitives.

I'm trying to decide if the unsafe makes sense in its current form, though... On the one hand the `&self` argument is completely unvalidated and without knowing the RMC compiler internals, may result in UB when using it.

Is there a reason the trait method doesn't return an owned value instead? It would make the trait safe to implement and each implementation may or may not need to use unsafe to generate a value:

```rust
pub trait Arbitrary {
    fn any() -> Self;
}

impl Arbitrary for u8 {
    fn any() -> Self }
        unsafe { any_raw() }
    }
}
```

src/test/firecracker/virtio-block-parse/main.rs
celinval: From what I understood... any insight?

library/rmc/src/invariant.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+unsafe impl<T> Invariant for Option<T>
+where
+    T: Invariant,
+{
+    #[inline(always)]
+    fn is_valid(&self) -> bool {
+        match self {
```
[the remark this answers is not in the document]
celinval: I have updated the implementation to avoid the issue I believe you are talking about.

src/test/rmc/DynTrait/main.rs
celinval: Probably! Some of these changes are from a couple months ago.

library/rmc/src/invariant.rs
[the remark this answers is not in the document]
camshaft: Another concern is: in its current form, it's _very_ easy to miss validating a field when implementing the trait:

```rust
struct MyThing {
    field: char,
    other: char,
}

impl Invariant for MyThing {
    fn is_valid(& self) -> bool {
        self.field.is_valid() // missing `other`!
    }
}
```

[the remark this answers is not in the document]
celinval: Thanks for clarifying. This is another possibility which uses a builder pattern. I thought the Invariant approach was a bit cleaner and easier to use. It also adds a method that can be used in assert statements which I thought was nice. I do agree that it has the downside of having a potential invalid `Self` and I think the unsafe impl is proper in this case.

I think they are somewhat equivalent though. For an arbitrary enum, it would go from:

```rust
unsafe impl Invariant for MyEnum {
    #[inline(always)]
    fn is_valid(&self) -> bool {
        matches!(*self, MyEnum::V1 | MyEnum::V2 | MyEnum::V3)
    }
}
```

to

```rust
impl Arbitrary for MyEnum {
    #[inline(always)]
    fn any() -> Self {
        match u8::any() {
            0 => MyEnum::V1,
            1 => MyEnum::V2,
            _ => MyEnum::V3,
        }
    }
}
```

Maybe we should check which one performs better or even offer both.

library/rmc/src/invariant.rs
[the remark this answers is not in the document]
celinval: No, this form didn't quite work so I replace it in the new revision.

library/rmc/src/invariant.rs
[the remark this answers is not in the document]
celinval: I understand the concern, and I debated between the two options for a long time. Maybe we can have a quick sync tomorrow to discuss the options.

Another possibility is to add both methods to the type with default implementations. Something like:

```rust
pub unsafe trait Arbitrary {
    fn is_valid(&self) -> bool { true }
    fn any() -> Self {
        let s = unsafe { rmc::any_raw::<Self>() };
        rmc::assume(s.is_valid());
        s
    } 
}

pub fn any<T: Arbritary>() -> T {
    T::any()
}
```

Users can pick and choose which one to implement / use.

library/rmc/src/invariant.rs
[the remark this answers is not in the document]
zhassan-aws: A macro is fine.

library/rmc/src/invariant.rs
zhassan-aws: My two cents: it would be inconvenient if the user needs to annotate all the enums in their code with `#[derive(Invariant)]`. Automatically handling that during codegen is more user friendly.

src/test/rmc/Invariants/nonzero.rs
```diff
@@ -0,0 +1,39 @@
...
+        assert!(v1.get() != 0, "Any should not generate value zero");
+
+        let option = Some(v1);
+        assert!(option.is_some(), "Niche optimization works well.");
+
+        let v2 = unsafe { rmc::any_raw::<$type>() };
+        rmc::expect_fail(v2.get() != 0, "Any raw may generate invalid value.");
+
+        let option = Some(v2);
+        rmc::expect_fail(option.is_some(), "Yep. This can happen due to niche optimization");
```
zhassan-aws, line 21: This is surprising!

src/test/expected/one-assert/expected
zhassan-aws: My bad :( The purpose of this test was to make sure we don't get all the noisy successes for builtin functions, e.g.:
```
<builtin-library-calloc> function calloc
[calloc.assertion.1] assertion false: SUCCESS

<builtin-library-ceil> function ceil
[ceil.assertion.1] assertion false: SUCCESS

<builtin-library-ceilf> function ceilf
[ceilf.assertion.1] assertion false: SUCCESS

<builtin-library-copysign> function copysign
[copysign.assertion.1] assertion false: SUCCESS
...
```
which we got rid of with `--drop-unused-functions`. So to make the test more robust, we need to check if the output does **not** include those lines. Is there a way to do that with our testing infrastructure?

src/test/firecracker/virtio-block-parse/main.rs
zhassan-aws: If we want to replace this with just:
```rust
rmc::any()
```
we would have to add an `Invariant` trait bound to `T`, is that correct? Not sure which option is better (assuming what I mentioned is correct).

src/test/expected/one-assert/expected
adpaco-aws: At present, there is no way to do that. Not that it would be hard to, but I would wait until we get structured output to implement this kind of checks. The test is still ensuring there are only a couple of successful assertions so I would leave as is.

src/test/firecracker/virtio-block-parse/main.rs
celinval: I see what you mean. This does remove some constraints from the proof. That said, the proof still succeeds, so I don't think there is any harm done. 

I'll add the Invariant and use rmc::any since that is the prefeed method.

src/test/rmc/Pointers_OtherTypes/main.rs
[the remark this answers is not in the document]
celinval: I updated this.

src/test/rmc/Invariants/enum.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// Check that we correctly generate invariants for enums.
```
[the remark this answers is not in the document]
celinval: Niche optimization is covered in the nonzero.rs testcase.

src/test/rmc/FunctionCall_Ret-NoParam/main.rs
```diff
@@ -37,15 +37,15 @@ fn return_bool() -> bool {
...
+    let y: bool = rmc::any();
     if y {
         return x / 2.0;
     } else {
         return x * 2.0;
     }
 }
 fn return_f64() -> f64 {
-    let x: f64 = rmc::nondet();
+    let x: f64 = unsafe { rmc::any_raw() };
```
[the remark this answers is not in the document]
celinval: By mistake. I hadn't implemented the float invariant yet when I updated the tests and I forgot to update them later.

src/test/rmc/Invariants/enum_raw_fail.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// rmc-verify-fail
+// Check that rmc::raw_any for enums will generate invalid enums.
+// This code should fail due to unreachable code being reached.
```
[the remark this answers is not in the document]
celinval: Sure.

src/test/rmc/Enum/result3.rs
celinval: Fixed it.

celinval:
I created the following issues to capture the mitigation tasks for the drawback of the current approach:
#726: Create custom derive attribute. 
#735: Validate all fields and enum variants are checked in the Invariant implementation.

celinval:
I added the Arbitrary trait with an implementation for the Invariant trait per @camshaft and @danielsn suggestion. Thanks!

camshaft (approved):
I really like how this turned out! nice work!

zhassan-aws (approved):
Looks great!

PR #663, opened 2021-11-23
Added new APIs to the RMC crate for creating non-deterministic slices

### LINKED ISSUE #614: An API for creating a non-deterministic slice
Requested feature: Creating a non-deterministic slice (with non-deterministic content as well as a non-deterministic length) currently requires multiple steps. For example, the following code creates a non-deterministic slice of `u8` with a maximum length of 10:
```rust
let arr: [u8; 10] = rmc::nondet();
let slice_len: usize = rmc::nondet();
rmc::assume(slice_len <= 10);
&arr[..slice_len]
```
It would be useful to wrap this in some API to simplify creating non-deterministic slices.
Use case: To test functions that take a slice.
Is this a breaking change?

zhassan-aws:
### Description of changes: 

Added several convenience APIs to the RMC crate for creating non-deterministic slices.

### Resolved issues:

Resolves #614 

### Testing:

* How is this change tested? Added two new tests

### Checklist


library/rmc/src/lib.rs
```diff
@@ -51,3 +53,82 @@ pub fn nondet<T>() -> T {
...
+/// ```rust
+/// let slice = rmc::NonDetSlice::<u8, 5>::new();
+/// foo(&slice); // where foo is a function that takes a slice and verifies a property about it
+/// ```
+pub struct NonDetSlice<T, const MAX_SLICE_LENGTH: usize> {
+    arr: [T; MAX_SLICE_LENGTH],
+    slice_len: usize,
+}
+
+impl<T, const MAX_SLICE_LENGTH: usize> NonDetSlice<T, MAX_SLICE_LENGTH> {
```
celinval, line 105: Looking at the example made me wonder, how is this different from a non-deterministic array?

library/rmc/src/lib.rs
```diff
@@ -51,3 +53,82 @@ pub fn nondet<T>() -> T {
...
+    pub fn new() -> Self {
```
celinval, line 106: Instead of new, I would recommend calling this from a function like the others in this module:

```rust
pub fn nondet_slice<T, const N: usize>() -> NonDetSlice<T, N>
```

Or we can just generate an array here too.
```rust
pub fn nondet_array<T, const N: usize>() -> [T; N]
```

celinval (changes requested):
Should we rename these functions to `any_*` to follow the same new name of `nondet` that we discussed offline?

zhassan-aws:
> Should we rename these functions to `any_*` to follow the same new name of `nondet` that we discussed offline?

I prefer if we rename all of them at the same time once we settle on the new name.

library/rmc/src/lib.rs
zhassan-aws: A non-deterministic array has a fixed length, whereas this API provides a slice with a non-deterministic length up to the given bound.

library/rmc/src/lib.rs
zhassan-aws: > Instead of new, I would recommend calling this from a function 

Good idea! I will add a function.

> Or we can just generate an array here too

Generating an array is already possible with `nondet`:
```rust
let arr: [u8; 5] = rmc::nondet();
```

src/test/rmc/NondetSlices/test1.rs
```diff
@@ -0,0 +1,18 @@
...
+fn check(slice: &[u8]) {
+    let len = slice.len();
+    assert!(len >= 0 && len <= 3);
+    if len > 0 {
+        let elem = slice[0];
+        assert!(elem == 1 || elem == 2 || elem == 3);
+    }
+}
+
+#[no_mangle]
```
adpaco-aws, line 13: This `#[no_mangle]` is not need, right?

zhassan-aws: Indeed! I'll take it out.

src/test/rmc/NondetSlices/test1.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+fn check(slice: &[u8]) {
```
celinval, line 4: can you add a comment to this test please?

src/test/rmc/NondetSlices/test1.rs
```diff
@@ -0,0 +1,17 @@
...
+    let len = slice.len();
+    assert!(len >= 0 && len <= 3);
```
celinval, line 6: can you also add a message to the assert?

src/test/rmc/NondetSlices/test2.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+fn check(s: &[u8]) {
+    let len = s.len();
+    assert!(len >= 0 && len < 6);
```
celinval, line 6: can you also please add a message here?

celinval (approved):
Just a few minor comments.

PR #665, opened 2021-11-23
Add CBMC Response parsing and Display controlled output

jaisnan:
### Description of changes: 

The Old Output from rmc had non user friendly elements including - 

1. Confusing output messages - See #189
2. Uncontrolled output from CBMC
3. Visually unappealing

To solve all the above issues and to setup future features, the CBMC Parser can now -

1. Process the response blob that is thrown out by CBMC 
2. Store it in temporary files for easy reading.
3. Restructure the blob to make it easily parseable
4. Display a controlled output depending on user inputted flags.

The flag --output-format provides the following options [old | terse| regular] and can be passed as following - 

`rmc test.rs --output-format regular`

This is how the previous output from rmc looks like - 

image

This is how the controlled output looks like -

image

This change also comes with a pipeline to easily add more styles of output in the future per user requirements and flag based inputs. One such option already provided is the '--terse' flag that displays a summarized output with error messages.

image

Since final output displays are subject to a lot of user experience feedback, the parser seperates out functionality from the result and can be easily changed to match customer tastes.

### Call-outs:

1. Integrating CBMC Viewer into the parser
2. Introduce three new formats of output via optional flags including report, log format, machine readable format
3. Handle non Json CBMC outputs with viewer
4. Test suite for the parser and Data Classes for each output format.

### Testing:

* How is this change tested?
Tested using regression tests, all but "Expected" suite passes tests with new output, but it still needs it's own testing suite which is in development.
* Is this a refactor change?
Refactor and UX/UI change

### Checklist


scripts/rmc
```diff
@@ -112,17 +112,33 @@ def main():
...
+            # This gives us the json output from CBMC which is passed into the parser
+            if "--json-ui" not in args.cbmc_args:
+                args.cbmc_args.extend(["--json-ui"])
+            if args.terse == True:
+                output_style = '1'
+            else:
+                output_style = '0'
+        else:
+            output_style = '2'
+
```
jaisnan, line 129: Use Enums and Constants instead of Magic Numbers.

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,398 @@
...
+
+# Get Information about the Solver from the Json file written
+def extract_solver_information(json_object):
+    
+    # Solver information is all the fields in the json object which are not related to the results
+    solver_information = []
+    try:
+        responses = json_object["response"]
+    except:
+        print("No such Value called \"response\"")
```
celinval, line 177: There is no point of keeping executing this function if there was an exception here, right?

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,398 @@
...
+                    pass
+            else:
+                solver_information_message += ''
+        except KeyError as e:
+            print("Key Error, Missing Properties in reading Solver Information from JSON")
+        solver_information_message += '\n'
+    return solver_information_message
+
+# Get property tests and results from the Json file written
+def extract_properties(json_object):
```
celinval, line 220: This function body looks a lot like the function `extract_solver_information(json_object)`

celinval (changes requested):
Here are a few comments.. I'll keep reviewing the other files.

scripts/temp_json/.gitignore
```diff
@@ -0,0 +1,7 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
celinval, line 1: Can you please remove this file.

scripts/rmc_flags.py
```diff
@@ -179,6 +179,16 @@ def add_visualizer_flags(make_group, add_flag, config):
                   this is generally the location from which rmc is currently being invoked
                   """)
 
+# Add flags needed for toggling and switching between outputs.
+def add_output_flags(make_group, add_flag, config):
+    group = make_group("Output flags", "Toggle between different styles of output")
+    add_flag(group, "--newOutput", default=False, action=BooleanOptionalAction,
+             help="Turn on new output")    
+    add_flag(group, "--oldOutput", default=True, action=BooleanOptionalAction,
```
celinval, line 187: the name convention is to separate words with '-'. So this should be `--old-output`.

scripts/cbmc_json_parser.py
jaisnan: Will take out the common functionality into it's own function

scripts/rmc
jaisnan: Working on fixing this

scripts/rmc.py
```diff
@@ -310,22 +319,24 @@ def link_c_lib(srcs, dst, c_lib, verbose=False, quiet=False, function="main", dr
...
-def run_cbmc(cbmc_filename, cbmc_args, verbose=False, quiet=False, dry_run=False):
+def run_cbmc(cbmc_filename, cbmc_args, verbose=False, quiet=False, dry_run=False,output_style='0'):
     cbmc_cmd = ["cbmc"] + cbmc_args + [cbmc_filename]
     scanners = []
     if "--unwinding-assertions" in cbmc_args:
         # Pass a scanner that shows a tip if the CBMC output contains unwinding failures
         unwind_asserts_pattern = ".*unwinding assertion.*: FAILURE"
         unwind_asserts_scanner = Scanner(unwind_asserts_pattern, append_unwind_tip)
-        scanners.append(unwind_asserts_scanner)
+        scanners.append(unwind_asserts_scanner)
```
celinval, line 329: can you please configure your editor to remove trailing spaces?

scripts/rmc.py
```diff
@@ -187,7 +189,14 @@ def run_cmd(
 
     # Write to stdout if specified, or if failure, or verbose or debug
     if (output_to == "stdout" or process.returncode != EXIT_CODE_SUCCESS or verbose or debug) and not quiet:
-        print(stdout)
+        # '2' represents the old default output and '0' and '1' are for the transformed output styles
+        if (output_style != '2'):
```
celinval, line 193: please use descriptive string or a constant instead of '2'.

scripts/rmc
```diff
@@ -112,17 +112,35 @@ def main():
                 args.verbose, dry_run=args.dry_run):
             return 1
 
+    # Control final output style using output_style variable which takes '0', '1', '2' as values
+    if "newOutput" in args:
```
celinval, line 116: Isn't this check redundant?

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,400 @@
...
+        # Print using an Interfact function
+        print_to_terminal(output_message)
+
+        # Delete temp Json file after displaying result
+        clear_json_file(json_file)
+    else:
+        # DynTrait tests generate a non json output due to "Invariant check failed" error
+        # For these cases, we just produce the cbmc output unparsed
+        # TODO: Parse these non json outputs from CBMC
+        non_json_cbmc_output_handler(log_file)
```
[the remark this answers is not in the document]
jaisnan: The following files in the regression tests (all related to DynTraits) give non json cbmc output, so for now the handling is to just print out the direct output (for the regression tests). Raising an exception would fail the regression tests. I thought I could handle these as a seperate Issue.
```
rmc/DynTrait/drop_nested_boxed_dyn_fail.rs
rmc/DynTrait/dyn_fn_mut_fail.rs
 rmc/DynTrait/dyn_fn_once_fail.rs
rmc/DynTrait/nested_closures_fail.rs
```

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,398 @@
...
+    return
+
+# Write the json file to disk for easy parsing
+def save_json_file(log_file):
+
+    # TODO : Change to use the source file name as the temp file name too
+    # Create a temp json file for every source file as testing can be parallelized
+    temp_file_name = str(random.randint(10,1000000))
```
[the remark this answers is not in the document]
jaisnan: Thanks!

scripts/rmc_flags.py
```diff
@@ -179,6 +179,16 @@ def add_visualizer_flags(make_group, add_flag, config):
                   this is generally the location from which rmc is currently being invoked
                   """)
 
+# Add flags needed for toggling and switching between outputs.
+def add_output_flags(make_group, add_flag, config):
+    group = make_group("Output flags", "Toggle between different styles of output")
```
celinval, line 184: I suggest you to replace these flags by the one command options, such as: `--output-format [old | regular | terse]`. You can use https://docs.python.org/3.6/library/argparse.html#choices to specify the options.

scripts/rmc-regression.sh
```diff
@@ -39,6 +39,10 @@ if [ -d "$RMC_DIR/firecracker/build" ]; then
   rm -rf "$RMC_DIR/firecracker/build"
 fi
 
+if [ -d "$RMC_DIR/scripts/temp_json" ]; then
```
celinval, line 42: Please use the project folder for temporary like other functions do.

scripts/rmc_flags.py
jaisnan: Yes, refactoring a lot of code to use Enums and Choices.

scripts/cbmc_json_parser.py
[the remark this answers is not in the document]
celinval: Can you add to these tests the option to use the old output format? You can set rmc flags by adding a comment at the top of the rust file:

```rust
// rmc-flags: <option>
```

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,400 @@
...
+import subprocess
+import os
+from colorama import Fore, Back, Style
+import random
+
+def main():
+    
+    # parser = argparse.ArgumentParser()
+    # Sample command run
+    command = 'rmc /home/ubuntu/rmc/src/test/rmc/Assume/main.rs'.split()
```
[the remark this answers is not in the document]
jaisnan: @tedinski

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,400 @@
...
+
+def transform_cbmc_output(log_file, output_style):
+    """
+    Take Unstructured CBMC Response object, parse the blob and gives structured 
+    and formatted output depending on User Inputted Output Style
+
+    Parameters -
+        log_file : str
+            A response blob that is given to the function from CBMC
+        output_style : int
```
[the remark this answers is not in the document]
jaisnan: Will add argparse as a dependency after integrating CBMC-Viewer.

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,408 @@
...
+            # Extract property messages from the json file
+            # Combine both messages to give as final output
+            output_message += construct_property_message(properties)
+
+        elif output_style == output_style_switcher["terse"]:
+
+            # Construct only summarized result and display output
+            output_message = construct_terse_property_message(properties)
+
+        # Print using an Interfact function
```
zhassan-aws, line 113:
```suggestion
        # Print using an Interface function
```

scripts/rmc
```diff
@@ -110,6 +110,17 @@ def main():
...
+    # if oldOutput flag is passed, then prevent output from going into the json parser
+    # else then by default and for terse output, display the result through the json parser
+    if(args.output_format == rmc_flags.OutputStyle.REGULAR) or (args.output_format == rmc_flags.OutputStyle.TERSE):
+        # This gives us the json output from CBMC which is passed into the parser
+        if "--json-ui" not in args.cbmc_args:
+            args.cbmc_args.extend(["--json-ui"])
+    elif args.output_format == rmc_flags.OutputStyle.OLD:
+        pass
+    else:
+        raise Exception('Contradrictory Flag outputs passed')
```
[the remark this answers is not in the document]
jaisnan: rmc_flags.OutputStyle.DEFAULT is the current output so it's value is 'old'. After this PR is merged, DEFAULT needs to be changed to 'regular' and the default output becomes the new output. 

It's value will be one of the other three options (or more in the future). if we are in the else branch, it means that possibly , more than one flag was passed.

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,408 @@
...
+        raise Exception("No Response Object from CBMC")
+
+    # Check for the objects which are not related to the result object
+    for response_object in responses:
+        if GlobalMessages.RESULT not in response_object.keys():
+            solver_information.append(response_object)
+        elif GlobalMessages.RESULT in response_object.keys():
+            properties = response_object["result"]
+        else:
+            return properties, solver_information
```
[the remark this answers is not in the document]
jaisnan: Yes, will clean it up.

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,408 @@
...
+
+    # Output Message is the final output that is printed to the user
+    output_message = ""
+
+    # Check if the output given by CBMC is in Json format or not
+    is_json_bool, json_file = is_json(log_file)
+    if is_json_bool:
+
+        # Write cbmc output to a json file in scripts/temp_json for parallel I/O
+        json_file = save_json_file(json_file)
```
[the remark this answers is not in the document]
jaisnan: `log_file` is the raw unstructured output (mostly a JSON Array) given by CBMC, `json_file` is  the JSON Array (log_file) converted to a JSON Object. save_json_file writes the json to disk as a temp file. Each temp file has a unique filename so when we run regression tests, multiple instances of the script can run the tests parallelly 
without all of the instances reading and writing into the same json file on disk. This temp file is removed after we're done reading from it.

scripts/rmc_flags.py
```diff
@@ -4,6 +4,8 @@
 
 import argparse
 import pathlib as pl
+from enum import Enum
+from random import choices
```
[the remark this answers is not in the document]
jaisnan: dead code, will clean it up.

scripts/rmc_flags.py
```diff
@@ -92,6 +94,49 @@ def __call__(self, parser, namespace, values, option_string=None):
         items.extend(values)
         setattr(namespace, self.dest, items)
 
+class EnumAction(argparse.Action):
```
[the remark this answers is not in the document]
jaisnan: It allows us to pass the Enum OutputStyle into existing flags by passing it as an action argument , ex -     `add_flag(group,"--output-format", default=OutputStyle.OLD,type=OutputStyle, action=EnumAction,help="Select the format for output")`. By declaring it as an option and setting it's properties, we can use the Enum effectively as a menu.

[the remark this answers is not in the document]
jaisnan: Will add further comments explaining the class and what it allows us to do.

scripts/rmc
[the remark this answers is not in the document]
jaisnan: Will fix the Exception and add comments to explain the `else` branch.

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,409 @@
...
+        print("JSON File not being parsed")
+        return {}
+
+    # TODO: Check if passing dictionaries around is optimal or if there are alternate ways to
+    # restructure and parse cbmc blob.
+    return temp_dict
+
+def extract_solver_information(json_object):
+    """
+    Get Information about the Solver from the Json file written
```
tedinski, line 194: This function does something important! It should get better docs. :)

Seems to be sorting these into two buckets based on whether `result` is a key.

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,409 @@
...
+        raise Exception("No Response Object from CBMC")
+
+    # Check for the objects which are not related to the result object
+    for response_object in responses:
+        if GlobalMessages.RESULT not in response_object.keys():
+            solver_information.append(response_object)
+        elif GlobalMessages.RESULT in response_object.keys():
+            properties = response_object["result"]
+        else:
+            raise KeyError
```
tedinski, line 213: This else shouldn't ever be reachable right?

(We might want some validation logic here to ensure properties isn't already set if we're just overriding it here. Otherwise some weird output might confuse this code later. Also: `[]` isn't the right initializer if it's not an array, seems like it should be `None`. Unless this is a bug and it should be an array...)

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,409 @@
...
+    return output_message
+
+# Method provides a handler for non json cbmc outputs, for example
+# with DynTrait tests
+def non_json_cbmc_output_handler(logfile):
+    # Basic handling is just printing the output provided by CBMC
+    raise Exception("CBMC Crashed - Unable to present Result")
+
+# Method provides an Interface to printing, can be expanded upon
+def print_to_terminal(output_message):
```
tedinski, line 395: Just print it, this function isn't necessary.

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,409 @@
...
+
+    # TODO: Change this to cProver status
+    # TODO: Extract information from CBMC about iterations
+    output_message += f"\nVERIFICATION:- {verification_status}\n"
+
+    return output_message
+
+# Method provides a handler for non json cbmc outputs, for example
+# with DynTrait tests
+def non_json_cbmc_output_handler(logfile):
```
tedinski, line 390: Inline this, doesn't need to be a function.

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,409 @@
...
+        CBMC version 5.36.0 (cbmc-5.36.0) 64-bit x86_64 linux
+        Reading GOTO program from file
+        ...
+    """
+
+    solver_information_message = ""
+    for message_object in solver_information:
+        # 'Program' and 'messageText' fields give us the information about the solver
+        try:
+            if GlobalMessages.PROGRAM in message_object.keys():
```
tedinski, line 232: Before each case, you should try to give 1-2 short abbreviated examples of what this case is intended to handle.

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,409 @@
...
+    if is_json_bool:
+
+        # Write cbmc output to a json file in scripts/temp_json for parallel I/O
+        json_file = save_json_file(json_file)
+
+        # Read from the saved json file and restructure it to make it parseable
+        parseable_json_dictionary = restructure_json_file(json_file)
+
+        # Extract property information from the restructured JSON file
+        properties, solver_information = extract_solver_information(parseable_json_dictionary)
```
tedinski, line 95: Better names needed. I'm actually unclear on what these are but if I'm guessing right then it almost seems backwards. Like, `properties` seems to be summary information from the solver while `solver_information` seems to be each individual property?

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,400 @@
...
+        '1' : 'terse',
+        '2' : 'old'
+    }
+
+    # Output Message is the final output that is printed to the user
+    output_message = ""
+
+    # Check if the output given by CBMC is in Json format or not
+    if is_json(log_file):
+        json_file = json.loads(log_file)
```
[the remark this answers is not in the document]
jaisnan: Will add as a feature in later iterations.

scripts/cbmc_json_parser.py
```diff
@@ -0,0 +1,359 @@
...
+
+    Parameters -
+        cbmc_response_json_array : JSON Array
+            A response JSON Array that contains the Result Object from CBMC and
+            Solver Information
+
+            Input Example -
+                [{'program': 'CBMC 5.44.0 (cbmc-5.44.0)'},
+                {'messageText': 'CBMC version 5.44.0 (cbmc-5.44.0) 64-bit x86_64 linux', 'messageType': 'STATUS-MESSAGE'},
+                {'messageText': 'Reading GOTO program from file', 'messageType': 'STATUS-MESSAGE'},..]
```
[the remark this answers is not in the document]
jaisnan: Sure fixing this

celinval:
> The comments I have are about... comments. So I think this can be merged. Hopefully after those quick fixes. @celinval do you have any further comments?

I haven't done a last pass, but I think your review should be enough and I don't want to delay this any longer. Thanks for checking!

celinval (approved):

PR #670, opened 2021-11-26
Refactor with `Book` struct and set default edition for each book

### LINKED ISSUE #666: Dashboard examples should be run using the default edition for their book
Contrary to what we had thought, dashboard examples should be run using Rust 2015 edition by default, unless otherwise noted:
https://doc.rust-lang.org/edition-guide/rust-2015/index.html

> Since it's the default edition, there's no way to port your code to Rust 2015; it just is. You'll be transitioning away from 2015, but never really to 2015. As such, there's not much else to say about it!

This is a bit counter-intuitive since only 4 examples out of the ~700 ones that we run specify the 2015 edition as a requirement. The rest do not specify the edition. Take for example `Rust by Example/Functions/Closures/Examples in std/Iterator::any/22.rs`, which is currently failing while using edition 2021:

```
stderr:
            
--> src/test/dashboard/books/Rust by Example/Functions/Closures/Examples in std/Iterator::any/22.rs:19:57
|
19 |     println!("2 in array2: {}", array2.into_iter().any(|&x| x == 2));
|                                                         ^--
|                                                         ||
|                                                         |expected due to this
|                                                         expected integer, found reference
|                                                         help: did you mean `x`: `&{integer}`
|
= note:   expected type `{integer}`
found reference `&_`

```

This does not happen when using the 2015 edition. So it is probably the best to:
 1. Not use any edition by default (this is essentially defaulting to 2015).
 2. Use an edition only if specified in the example (such cases do not exist at the moment except for the 4 mentioned examples, which use Rust 2015 anyway).

adpaco-aws:
### Description of changes: 

Turns out default editions are different for each book: https://rust-lang.zulipchat.com/#narrow/stream/237824-t-lang.2Fdoc

Refactored the dashboard code to include an extensible `Book` structure which contains all information related to a particular book. This allows us to set the correct default edition for the examples in each book.

### Resolved issues:

Resolves #666 

### Testing:

* How is this change tested? Running dashboard locally.

* Is this a refactor change? No, but includes some refactoring.

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -994,6 +994,7 @@ dependencies = [
  "rustdoc",
  "serde",
  "serde_json",
+ "toml",
  "walkdir",
 ]
 
--- a/src/tools/dashboard/Cargo.toml
+++ b/src/tools/dashboard/Cargo.toml
@@ -13,3 +13,4 @@ rustdoc = { path = "../../librustdoc" }
 walkdir = "2.3.2"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+toml = "0.5"
--- a/src/tools/dashboard/src/books.rs
+++ b/src/tools/dashboard/src/books.rs
@@ -43,7 +43,7 @@ struct SummaryData {
     summary_start: String,
 }
 
-// Data needed for parsing book without a `SUMMARY.md` file
+// Data needed for parsing book without a summary file
 struct DirectoryData {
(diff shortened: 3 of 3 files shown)


zhassan-aws (approved):
LGTM

adpaco-aws:
Thanks @zhassan-aws ! Rebased and added two minor comments, merging now.

Issue #671, opened 2021-11-29
Update vtable restrictions once CBMC side lands. 


avanhatt:
Once this lands: https://github.com/diffblue/cbmc/pull/6376

- [ ] Update minimum CBMC version.
- [ ] Enable fixme unit tests.
- [ ] Add unit test for InternedString deserialization. 
- [ ] Move flag check inside block, as listed here: https://github.com/model-checking/rmc/pull/536

tautschnig:
diffblue/cbmc#6376 has finally been merged, and will be part of the next CBMC release (5.46.0, due to be released 2021-12-09).

tautschnig:
CBMC 5.46.0 has now been released.

PR #685, opened 2021-12-08
add regression for link restrictions tool

### LINKED ISSUE #684: Add building restriction link to regression script
We can't land proper tests for the link restrictions feature until the CBMC release is out (RMC issue https://github.com/model-checking/rmc/issues/671), but we should at least build the code.

avanhatt:
### Description of changes: 

We can't land proper tests for the link restrictions feature until the CBMC release is out (RMC issue #671), but we should at least build the code. Fix build and add to regression script.

### Resolved issues:

Resolves #684

### Call-outs:

To build in tree, this removes some unused rustc dependencies from the cbmc crate.

### Testing:

* How is this change tested?

Adding build command to regression tests.

* Is this a refactor change?

### Checklist

### DIFF
--- a/scripts/rmc-regression.sh
+++ b/scripts/rmc-regression.sh
@@ -27,6 +27,9 @@ check-cbmc-viewer-version.py --major 2 --minor 5
 (cd src/rmc-compiler/cbmc; cargo test)
 (cd src/rmc-compiler; cargo test)
 
+# Build tool for linking RMC pointer restrictions
+cargo build --release --manifest-path src/tools/rmc-link-restrictions/Cargo.toml 
+
 # Standalone rmc tests, expected tests, and cargo tests
 ./x.py build -i src/tools/compiletest --stage 0
 export COMPILETEST_FORCE_STAGE0=1  # We don't care about the stage anymore. Remove this once we replace ./x.py test
--- a/library/rmc_restrictions/Cargo.toml
+++ b/library/rmc_restrictions/Cargo.toml
@@ -9,6 +9,5 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-rustc_data_structures = { path = "../../compiler/rustc_data_structures" }
 serde = {version = "1", features = ["derive"]}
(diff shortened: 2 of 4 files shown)


scripts/rmc-regression.sh
```diff
@@ -24,6 +24,10 @@ check-cbmc-viewer-version.py --major 2 --minor 5
 ./x.py build -i --stage 1 library/std ${EXTRA_X_PY_BUILD_ARGS}
 ./scripts/setup/build_rmc_lib.sh
 
+# Build tool for linking RMC pointer restrictions
+rustup toolchain install nightly
+cargo +nightly build --release --manifest-path src/tools/rmc-link-restrictions/Cargo.toml
```
danielsn, line 29: why do we need nightly?

danielsn (approved):

scripts/rmc-regression.sh
avanhatt: The `smallvec` dependency of cbmc requires it, trying without that now

avanhatt:
@danielsn sorry for the delay, this is now ready to merge! I tried without nightly and the regressions fail, it might be doable if we remove more dependencies from the `cbmc` crate, but I won't be able to try again for a bit.

I no longer have write access so won't be able to merge myself.

scripts/rmc-regression.sh
celinval: Maybe create a symbolic link to `src/rmc-compiler/rust-toolchain.toml` inside `src/tools/rmc-link-restrictions/`. We want to lock in a specific version.

I would use a link so we track the same version.

The only caveat is that I think you need to execute cargo from the folder where the toolchain file is.

celinval (changes requested):
Thanks for doing this Alexa! I would highly prefer to specify a nightly version in the regression to keep our tests consistent.

avanhatt:
> Thanks for doing this Alexa! I would highly prefer to specify a nightly version in the regression to keep our tests consistent.

Looks like this commit b515875a25f7e74eb7b379b66915be56ddd82c91 from DSN removes the need for nightly! Pushed a new version that should hopefully work without it.

celinval (approved):
Sweet! That works for me. :)

Issue #545, opened 2021-10-07
RMC doesn't seem to handle stack unwind logic correctly


celinval:
I was playing with the assertion codegen to understand how we were handling cleanup after a failure. I created the following test case:

```rust
// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0 OR MIT
//
// This test ensures that rmc follows the correct CFG for assertion failures.
// - Statements that succeeds an assertion failure should be unreachable.
// - Cleanup statements should still be executed though.
// - Note that failures while unwinding actually crashes the process. So drop may only be called
//   once.
// File: cleanup.rs

#[derive(PartialEq, Eq)]
struct S {
    a: u8,
    b: u16,
}

impl Drop for S {
    fn drop(&mut self) {
        assert!(false, "A1: This should still fail during cleanup");
    }
}

pub fn main() {
    let lhs = S { a: 42, b: 42 };
    let rhs = S { a: 0, b: 0 };
    assert!(lhs == rhs, "A2: A very false statement. Always fail.");
    assert!(false, "A3: Unreachable assert. Code always panic before this line.");
}
```

using the following command line invocation:

```
rmc cleanup.rs
```

**I expected to see this happen:** Assertions A1 and A2 should fail, but assertions A3 should be unreachable.

**Instead, this happened:** RMC entered an infinite loop.

I also tested the code with --unwind 0. The result is incorrect. A2 and A3 fail, but A1 doesn't.

```
[<S as std::ops::Drop>::drop.assertion.1] line 21 A1: This should still fail during cleanup: SUCCESS
....
[main.assertion.3] line 23 resume instruction: SUCCESS
[main.assertion.1] line 26 A2: A very false statement. Always fail.: FAILURE
[main.assertion.2] line 26 A3: Unreachable assert. Code always panic before this line.: FAILURE
```

celinval:
There are actually two different bugs reported in this issue. 
1. Infinite loop.
2. Incorrect cleanup logic.

We'll handle the bug `#1` on issue #636 and we'll leave this issue to track bug `#2`

tedinski:
I don't think this is tracking anything not in #692 and #267, closing.

PR #687, opened 2021-12-10
Fix assert!() and panic!() code generation

### LINKED ISSUE #67: Soundness: `codegen_panic` makes a lot of assumptions
`codegen_panic` makes a lot of assumptions about the type of input it gets

### LINKED ISSUE #466: Code location for `assert!(false)` is incorrect.
RMC is giving the wrong code location of assert!(false) statements.

I tried this code:

```rust
fn magic() -> i32 {     // L1
    5                   // L2
}                       // L3
                        // L4
pub fn main() {         // L5
    if magic() == 5 {   // L6
        assert!(false); // L7
    }                   // L8
    println!("Magic!"); // L9
}                       // L10
```

```
rmc --input dummy.rs
```

with RMC version:

The following output is printed:
```
[main.assertion.1] line 6 assertion failed: false: FAILURE
```

Note that the MIR is correct:
```
> rustc dummy.rs --emit mir
> grep panic dummy.mir
     core::panicking::panic(const "assertion failed: false"); // scope 0 at dummy.rs:7:9: 7:24
```

Root cause:

The assert!() macro is translated to `if !cond { begin_panic; }`. However, for `assert!(false)`, rustc optimizes the generated code and it only emit a `panic` statement.

During RMC code generation of panic statements, we assume that if the basic block only has one predecessor, it is an assertion. There are many cases that this is not true. See [this code](https://github.com/model-checking/rmc/blob/ef758e29c388de215fc88e30b6b39c11db1e1d0f/compiler/rustc_codegen_llvm/src/gotoc/mir_to_goto/codegen/statement.rs#L488).

### LINKED ISSUE #543: Panic and abort should stop the program execution
I tried this code:

```rust
//
// rmc-verify-fail

pub fn main() {
    for i in 0..4 {
        if i == 1 {
            panic!("This comes first");
        }
        if i == 2 {
            panic!("This should never happen");
        }
    }
}
```

```
rmc loop.rs
```

I expected to see this happen: *The first assertion should fail but the second one should be unreachable. The panic call should abort the execution in the second iteration of the loop.*

Instead, this happened: *Both assertions fail. The trace for the second assertion includes the first panic call.*

Note that the same behavior happens if we replace the `panic!()` statement by `process::abort()`.

### LINKED ISSUE #636: An example with a false assertion doesn't terminate
I tried this code:

```rust
#[no_mangle]
fn main() {
    let x: bool = rmc::nondet();
    if x {
        assert!(1 + 1 == 1);
    }
    assert!(1 + 1 == 3);
}
```

```
rmc test.rs
```

with RMC version: `2141a76f98d`

I expected to see this happen: both assertions fail.

Instead, this happened: it runs forever (CBMC keeps unwinding forever). If I specify an unwinding depth, the unwinding assertion fails:
```
$ rmc test.rs --cbmc-args --unwind 10
.
.
.

** 3 of 78 failed (2 iterations)
VERIFICATION FAILED
[RMC] info: Verification output shows one or more unwinding failures.
[RMC] tip: Consider increasing the unwinding value or disabling `--unwinding-assertions`.
```

celinval:
### Description of changes: 

I refactored how we codegen panic statements so it now it terminate the program per its spec. I have also removed the hack we had to try to get the assert location.

Since we currently do not support stack unwinding, the panic codegen will always terminate immediately and it will not try to unwind the stack. I added an option to RMC to force the compiler to use abort as the panic strategy and a check to rmc codegen that will fail if the user tries to override that.

Another note is that we currently do not support `#[track_caller]` and I have not changed that.

This change fixes #67, fixes #466, fixes #543, and fixes #636. This change also mitigates #545.

### Resolved issues:

Resolves #67
Resolves #466 
Resolves #543 
Resolves #636 

### Call-outs:

I split this into two commits. One with the changes to the compiler and new test cases. The second commit is adjustments to our existing tests.

### Testing:

* How is this change tested? New and old test cases.

### Checklist

### DIFF
--- a/compiler/cbmc/src/goto_program/stmt.rs
+++ b/compiler/cbmc/src/goto_program/stmt.rs
@@ -207,10 +207,6 @@ impl Stmt {
         )
     }
 
-    pub fn assume_false(loc: Location) -> Self {
-        Stmt::assume(Expr::bool_false(), loc)
-    }
-
     /// `__CPROVER_assume(cond);`
     pub fn assume(cond: Expr, loc: Location) -> Self {
         assert!(cond.typ().is_bool(), "Assume expected bool, got {:?}", cond);
--- a/compiler/cbmc/src/goto_program/stmt.rs
+++ b/compiler/cbmc/src/goto_program/stmt.rs
@@ -207,6 +207,10 @@ impl Stmt {
         )
     }
 
+    pub fn assume_false(loc: Location) -> Self {
+        Stmt::assume(Expr::bool_false(), loc)
+    }
+
     /// `__CPROVER_assume(cond);`
     pub fn assume(cond: Expr, loc: Location) -> Self {
         assert!(cond.typ().is_bool(), "Assume expected bool, got {:?}", cond);
--- a/compiler/rustc_codegen_rmc/src/codegen/span.rs
+++ b/compiler/rustc_codegen_rmc/src/codegen/span.rs
@@ -27,6 +27,16 @@ impl<'tcx> GotocCtx<'tcx> {
         )
     }
 
+    // Get the location of the caller. This will attempt to reach the macro caller.
+    pub fn codegen_caller_span(&self, sp: &Option<Span>) -> Location {
+        if let Some(span) = sp {
+            let topmost = span.ctxt().outer_expn().expansion_cause().unwrap_or(*span);
+            self.codegen_span(&topmost)
+        } else {
+            Location::none()
+        }
+    }
+
     pub fn codegen_span_option(&self, sp: Option<Span>) -> Location {
         sp.map_or(Location::none(), |x| self.codegen_span(&x))
     }
--- a/compiler/rustc_codegen_rmc/src/codegen/span.rs
+++ b/compiler/rustc_codegen_rmc/src/codegen/span.rs
@@ -27,7 +27,10 @@ impl<'tcx> GotocCtx<'tcx> {
         )
     }
 
-    // Get the location of the caller. This will attempt to reach the macro caller.
+    /// Get the location of the caller. This will attempt to reach the macro caller.
+    /// This function uses rustc_span methods designed to returns span for the macro which
+    /// originally caused the expansion to happen.
+    /// Note: The API stops backtracing at include! boundary.
     pub fn codegen_caller_span(&self, sp: &Option<Span>) -> Location {
         if let Some(span) = sp {
             let topmost = span.ctxt().outer_expn().expansion_cause().unwrap_or(*span);
--- a/compiler/rustc_codegen_rmc/src/codegen/statement.rs
+++ b/compiler/rustc_codegen_rmc/src/codegen/statement.rs
@@ -3,7 +3,7 @@
 use super::typ::TypeExt;
 use super::typ::FN_RETURN_VOID_VAR_NAME;
 use crate::{GotocCtx, VtableCtx};
-use cbmc::goto_program::{BuiltinFn, Expr, Location, Stmt, Type};
+use cbmc::goto_program::{BuiltinFn, Expr, ExprValue, Location, Stmt, Type};
 use rustc_hir::def_id::DefId;
 use rustc_middle::mir;
 use rustc_middle::mir::{
@@ -485,51 +485,34 @@ impl<'tcx> GotocCtx<'tcx> {
     pub fn codegen_panic(&mut self, span: Option<Span>, fargs: Vec<Expr>) -> Stmt {
         // CBMC requires that the argument to the assertion must be a string constant.
         // If there is one in the MIR, use it; otherwise, explain that we can't.
-        // TODO: give a better message here.
-        let arg = match fargs[0].struct_expr_values() {
-            Some(values) => values[0].clone(),
-            _ => Expr::string_constant(
-                "This is a placeholder assertion message; the rust message requires dynamic string formatting, which is not supported by CBMC",
-            ),
-        };
-
-        let loc = self.codegen_span_option(span);
-        let cbb = self.current_fn().current_bb();
-
-        // TODO: is it proper?
-        //
-        // [assert!(expr)] generates code like
-        //     if !expr { panic() }
-        // thus when we compile [panic], we would like to continue with
-        // the code following [assert!(expr)] as well as display the panic
-        // location using the assert's location.
-        let preds = &self.current_fn().mir().predecessors()[cbb];
-        if preds.len() == 1 {
-            let pred: &BasicBlock = preds.first().unwrap();
-            let pred_bbd = &self.current_fn().mir()[*pred];
-            let pterm = pred_bbd.terminator();
-            match pterm.successors().find(|bb| **bb != cbb) {
-                None => self.codegen_assert_false(arg, loc),
-                Some(alt) => {
-                    let loc = self.codegen_span(&pterm.source_info.span);
-                    Stmt::block(
-                        vec![
-                            self.codegen_assert_false(arg, loc.clone()),
-                            Stmt::goto(self.current_fn().find_label(alt), Location::none()),
-                        ],
-                        loc,
-                    )
+        let mut msg = String::from(
+            "This is a placeholder message; RMC doesn't support message formatted at runtime",
+        );
+        if let Some(values) = fargs[0].struct_expr_values() {
+            if !values.is_empty() {
+                if let ExprValue::AddressOf(msg_value) = values[0].value() {
+                    if let ExprValue::Index { array, .. } = msg_value.value() {
+                        if let ExprValue::StringConstant { s } = array.value() {
+                            msg = s.to_string();
+                        }
+                    }
+                } else {
+                    tracing::error!(?values, "Oh no");
                 }
             }
-        } else {
-            self.codegen_assert_false(arg, loc)
-        }
+        };
+
+        self.codegen_fatal_error(&msg, span)
     }
 
-    // By the time we get this, the string constant has been codegenned.
-    // TODO: make this case also use the Stmt::assert_false() constructor
-    pub fn codegen_assert_false(&mut self, err: Expr, loc: Location) -> Stmt {
-        BuiltinFn::CProverAssert.call(vec![Expr::bool_false(), err], loc.clone()).as_stmt(loc)
+    // Generate code for fatal error which should trigger an assertion failure and abort the
+    // execution.
+    pub fn codegen_fatal_error(&self, msg: &str, span: Option<Span>) -> Stmt {
+        let loc = self.codegen_caller_span(&span);
+        Stmt::block(
+            vec![Stmt::assert_false(msg, loc.clone()), Stmt::assume_false(loc.clone())],
+            loc,
+        )
     }
 
     pub fn codegen_statement(&mut self, stmt: &Statement<'tcx>) -> Stmt {
--- a/compiler/rustc_codegen_rmc/src/codegen/statement.rs
+++ b/compiler/rustc_codegen_rmc/src/codegen/statement.rs
@@ -482,25 +482,13 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
-    pub fn codegen_panic(&mut self, span: Option<Span>, fargs: Vec<Expr>) -> Stmt {
+    pub fn codegen_panic(&self, span: Option<Span>, fargs: Vec<Expr>) -> Stmt {
         // CBMC requires that the argument to the assertion must be a string constant.
         // If there is one in the MIR, use it; otherwise, explain that we can't.
-        let mut msg = String::from(
+        assert!(!fargs.is_empty(), "Panic requires a string message");
+        let msg = extract_const_message(&fargs[0]).unwrap_or(String::from(
             "This is a placeholder message; RMC doesn't support message formatted at runtime",
-        );
-        if let Some(values) = fargs[0].struct_expr_values() {
-            if !values.is_empty() {
-                if let ExprValue::AddressOf(msg_value) = values[0].value() {
-                    if let ExprValue::Index { array, .. } = msg_value.value() {
-                        if let ExprValue::StringConstant { s } = array.value() {
-                            msg = s.to_string();
-                        }
-                    }
-                } else {
-                    tracing::error!(?values, "Oh no");
-                }
-            }
-        };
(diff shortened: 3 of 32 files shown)


compiler/cbmc/src/goto_program/stmt.rs
```diff
@@ -207,6 +207,10 @@ impl Stmt {
         )
     }
 
+    pub fn assume_false(loc: Location) -> Self {
```
danielsn, line 210: doc comment

compiler/rustc_codegen_rmc/src/codegen/span.rs
```diff
@@ -27,6 +27,16 @@ impl<'tcx> GotocCtx<'tcx> {
         )
     }
 
+    // Get the location of the caller. This will attempt to reach the macro caller.
+    pub fn codegen_caller_span(&self, sp: &Option<Span>) -> Location {
+        if let Some(span) = sp {
+            let topmost = span.ctxt().outer_expn().expansion_cause().unwrap_or(*span);
```
danielsn, line 33: what is this doing?

compiler/rustc_codegen_rmc/src/codegen/statement.rs
```diff
@@ -485,51 +485,34 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        if let Some(values) = fargs[0].struct_expr_values() {
+            if !values.is_empty() {
+                if let ExprValue::AddressOf(msg_value) = values[0].value() {
+                    if let ExprValue::Index { array, .. } = msg_value.value() {
+                        if let ExprValue::StringConstant { s } = array.value() {
+                            msg = s.to_string();
+                        }
+                    }
+                } else {
+                    tracing::error!(?values, "Oh no");
```
danielsn, line 500: More informative message please?

compiler/rustc_codegen_rmc/src/codegen/statement.rs
```diff
@@ -485,51 +485,34 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                            Stmt::goto(self.current_fn().find_label(alt), Location::none()),
-                        ],
-                        loc,
-                    )
+        let mut msg = String::from(
+            "This is a placeholder message; RMC doesn't support message formatted at runtime",
+        );
+        if let Some(values) = fargs[0].struct_expr_values() {
+            if !values.is_empty() {
+                if let ExprValue::AddressOf(msg_value) = values[0].value() {
```
danielsn, line 493: Are there any other times we might want to do this? Could be a utility function?

compiler/rustc_codegen_rmc/src/codegen/statement.rs
```diff
@@ -485,51 +485,34 @@ impl<'tcx> GotocCtx<'tcx> {
...
-    // By the time we get this, the string constant has been codegenned.
-    // TODO: make this case also use the Stmt::assert_false() constructor
-    pub fn codegen_assert_false(&mut self, err: Expr, loc: Location) -> Stmt {
-        BuiltinFn::CProverAssert.call(vec![Expr::bool_false(), err], loc.clone()).as_stmt(loc)
+    // Generate code for fatal error which should trigger an assertion failure and abort the
+    // execution.
+    pub fn codegen_fatal_error(&self, msg: &str, span: Option<Span>) -> Stmt {
+        let loc = self.codegen_caller_span(&span);
+        Stmt::block(
+            vec![Stmt::assert_false(msg, loc.clone()), Stmt::assume_false(loc.clone())],
```
danielsn, line 513: should we `assume(false)` or `abort()`

compiler/rustc_codegen_rmc/src/compiler_interface.rs
```diff
@@ -187,6 +186,16 @@ impl CodegenBackend for GotocCodegenBackend {
     }
 }
 
+fn check_options(session: &Session) {
+    if !session.overflow_checks() {
+        session.err("RMC requires overflow checks in order to provide a sound analysis.")
+    }
+
+    if session.panic_strategy() != PanicStrategy::Abort {
+        session.err("RMC can only handle abort panic strategy (-C panic=abort).")
```
danielsn, line 195: Link to issue they can +1

compiler/rustc_codegen_rmc/src/overrides/hooks.rs
```diff
@@ -197,9 +197,10 @@ struct Panic;
 
 impl<'tcx> GotocHook<'tcx> for Panic {
     fn hook_applies(&self, tcx: TyCtxt<'tcx>, instance: Instance<'tcx>) -> bool {
-        output_of_instance_is_never(tcx, instance)
-            && (instance_name_is(tcx, instance, "begin_panic")
-                || instance_name_is(tcx, instance, "panic"))
+        let def_id = instance.def.def_id();
+        Some(def_id) == tcx.lang_items().panic_fn()
+            || Some(def_id) == tcx.lang_items().panic_display()
+            || Some(def_id) == tcx.lang_items().begin_panic_fn()
```
danielsn, line 203: 👍

compiler/rustc_codegen_rmc/src/overrides/hooks.rs
```diff
@@ -231,15 +232,11 @@ impl<'tcx> GotocHook<'tcx> for Nevers {
...
-        // _target must be None due to how rust compiler considers it
-        Stmt::assert_false(
-            &format!(
-                "a panicking function {} is invoked",
-                with_no_trimmed_paths(|| tcx.tcx.def_path_str(instance.def_id()))
-            ),
-            loc,
-        )
+        let msg = format!(
+            "a panicking function {} is invoked",
```
danielsn, line 236: I wonder if we can give a clearer message here

src/test/expected/binop/main.rs
```diff
@@ -56,19 +56,22 @@ fn ibxor_test(a: i32, b: i32, correct: i32, wrong: i32) {
 }
 
 fn main() {
```
danielsn, line 58: IS this to prevent one failure from hiding others?

src/test/expected/float-nan/main.rs
```diff
@@ -6,15 +6,17 @@ fn main() {
...
+        0 => assert!(1.0 / f != 0.0 / f),
+        // should succeed
+        1 => assert!(!(1.0 / f == 0.0 / f)),
+        // should fail
+        2 => assert!(1.0 / f == 0.0 / f),
+        // should fail
+        3 => assert!(0.0 / f == 0.0 / f),
+        // should suceed
+        _ => assert!(0.0 / f != 0.0 / f),
+    }
```
danielsn, line 21: would it make sense to have a macro for this pattern

src/test/rmc/DynTrait/dyn_fn_param_fail.rs
```diff
@@ -10,14 +10,18 @@
...
+        rmc::expect_fail(x != 5, "Wrong return");
+    }
 
     /* The function dynamic object has no associated data */
     rmc::expect_fail(size_from_vtable(vtable!(fun)) != 0, "Wrong size");
 }
 
 pub fn unit_to_u32() -> u32 {
-    assert!(false);
+    if rmc::nondet::<bool>() {
```
danielsn, line 22: do we need the turbofish here

compiler/rustc_codegen_rmc/src/codegen/statement.rs
celinval: ooops... That was a debug statement. =)

compiler/rustc_codegen_rmc/src/codegen/span.rs
celinval: It is getting the location of the macro caller. It gets the outer most expansion and get its cause.

compiler/rustc_codegen_rmc/src/codegen/statement.rs
celinval: I don't know. I figured that this code has a lot of assumptions about the format of the input, so I was thinking about leaving here for now and pulled it out if we start seeing this pattern in other places.

compiler/rustc_codegen_rmc/src/codegen/statement.rs
celinval: That's a good question. What do you think? I can try it out with abort and see how CBMC behaves.

compiler/rustc_codegen_rmc/src/compiler_interface.rs
celinval: Sure.

compiler/rustc_codegen_rmc/src/overrides/hooks.rs
celinval: Sure. Can I do it in a follow up PR since this will likely impact a lot of expected tests. I think this PR is already big enough.

src/test/expected/binop/main.rs
celinval: Yes. Otherwise everything passed `iadd_test()` becomes unreachable.

src/test/expected/float-nan/main.rs
celinval: I figured it wasn't worth the effort. There are a few differences in them and a macro would probably make it harder to read.

src/test/rmc/DynTrait/dyn_fn_param_fail.rs
celinval: Oh... that's a good point. I guess because it's inside an `if`, we probably don't. Let me check.

compiler/cbmc/src/goto_program/stmt.rs
celinval: For sure!

compiler/rustc_codegen_rmc/src/codegen/statement.rs
celinval: I'll remove it!

compiler/rustc_codegen_rmc/src/codegen/span.rs
celinval: I think the important part is the expansion_cause() method. This is the doc:
```
    /// Returns span for the macro which originally caused this expansion to happen.
    ///
    /// Stops backtracing at include! boundary.
    pub fn expansion_cause(mut self) -> Option<Span> 
```

This is how other backend engines assemble the Location in PanicInfo when the current function does not have `#[track_caller]` attribute which we currently don't support.

compiler/rustc_codegen_rmc/src/codegen/statement.rs
celinval: I replaced this by calling `abort()` instead of `assume(false)`, but I still see this message on CBMC output: 
```
aborting path on assume(false) at file <builtin-library-abort> line 6 function abort thread 0
```
So I don't think it will make much difference. I'll use abort() to improve readability.

compiler/cbmc/src/goto_program/stmt.rs
celinval: I'm actually removing this since I am adding a builtin abort()

compiler/rustc_codegen_rmc/src/codegen/statement.rs
tedinski: I think Jai had a utility function that does this in his first PR (that's still open). Maybe just steal it from there?

src/test/rmc/Options/check_tests.rs
```diff
@@ -4,7 +4,7 @@
 // Check that we can verify test harnesses using the --tests argument.
 // Note: We need to provide the compile-flags because compile test runs rustc directly and via rmc.
 
-// compile-flags: --test
+// compile-flags: --test --C panic=abort
```
tedinski, line 7: Is this needed?

compiler/rustc_codegen_rmc/src/codegen/span.rs
tedinski: I agree this function could use an extensive doc comment. It looks like you found a really great solution compared to our original hack, so helping us use this function more in the future would be awesome.

src/test/expected/float-nan/main.rs
tedinski: I wonder if we should have the original, primitive cbmc assert available still.

Then our tests could still be straight-line code, just using `rmc::assert()` instead of `assert!`

celinval: I was thinking about adding a function rmc::check() that had the behaviour you mentioned. What do you think? I would prefer not having two types of assert.

That said, I would prefer adding it in a different PR.

celinval: Also, once we have the `#[proof]` attribute integrated, we should be able to easily split tests like this into multiple harnesses in the same file.

src/test/rmc/Options/check_tests.rs
celinval: Let me double check. I think you are correct though.

src/test/expected/float-nan/main.rs
tedinski: I'm cool with a different name (brainstorming: `rmc::probe`? `check` also seems good) and in a follow up PR.

src/test/rmc/DynTrait/dyn_fn_param_fail.rs
tedinski: was it necessary?

src/test/rmc/Options/check_tests.rs
tedinski: was it necessary?

tedinski (approved):
I think this is good to merge. I really want `rmc::check` or whatever we call it, but whenever we get to it I guess. :)

There were two very minor things I didn't see addressed that I added comments to, also.

src/test/expected/float-nan/main.rs
celinval: I created https://github.com/model-checking/rmc/issues/695 to capture the work.

celinval:
> I think this is good to merge. I really want `rmc::check` or whatever we call it, but whenever we get to it I guess. :)
> 
> There were two very minor things I didn't see addressed that I added comments to, also.

My bad. I had missed them. Thanks for pointing it out.

danielsn (approved):

PR #688, opened 2021-12-10
fix tupled args

danielsn:
### Description of changes: 

Currently, we marshall arguments to closures using a positional constructor. This is problematic if rustc has reordered the fields of the tuple they are being marshalled into.  Instead, use the field names.

### Resolved issues:

Resolves https://github.com/model-checking/rmc/issues/678

### Call-outs:

The whole code around spread arguments is confusing. I tried to rewrite the comments to better reflect what I figured out as I was debugging the issue.

### Testing:

* Is this a refactor change? Partially, cleaned up the code as I worked on it.

### Checklist


zhassan-aws (approved):
LGTM

tedinski (approved):
Tupling still confuses me, but looks fine as far as I can tell.

Needs merge conflicts resolved, fyi.

src/test/rmc/Closure/tupled_closure.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! Tests that we cover closures with tuples correctly
```
celinval, line 3: If you haven't solved the merge conflicts yet, can you please clarify this comment? I know what you mean, but I think this comment might be confusing if you don't have good knowledge on how the compiler handles closures, because there is no actual tuples in this test.

That said, it is a minor comment, so please don't let that stop you from merging.

compiler/rustc_codegen_rmc/src/codegen/function.rs
```diff
@@ -158,30 +148,49 @@ impl<'tcx> GotocCtx<'tcx> {
             ),
         };
 
-        // Now that we have the tuple, write the individual component locals
-        // back to it as a GotoC struct.
+        // When we codegen the function signature elsewhere, we will codegen the untupled version.
```
celinval, line 151: Let me see if I understand, the fix here is to use a BTreeMap to sort the arguments according to their name instead of the order of occurrence.

If that is the case, can you please clarify that in the comment inside the code? In the example you posted, both positional or sorted arguments would yield the same result.

compiler/rustc_codegen_rmc/src/codegen/function.rs
```diff
@@ -158,30 +148,49 @@ impl<'tcx> GotocCtx<'tcx> {
...
-        let tuple_expr = Expr::struct_expr_from_values(tup_typ, arg_exprs, &self.symbol_table)
-            .with_location(loc.clone());
-        self.current_fn_mut().push_onto_block(Stmt::decl(tup_sym.to_expr(), Some(tuple_expr), loc));
+        let marshalled_tuple_fields =
+            BTreeMap::from_iter(args.iter().enumerate().map(|(arg_i, arg_t)| {
+                // The components come at the end, so offset by the untupled length.
+                let lc = Local::from_usize(arg_i + starting_idx);
+                let (name, base_name) = self.codegen_spread_arg_name(&lc);
+                let sym = Symbol::variable(name, base_name, self.codegen_ty(arg_t), loc.clone());
+                // TODO, I'm concerned that these are never declared inside the function?
```
celinval, line 176: Have you tried --gen-c to see what's being generated?

celinval (commented):
Nothing blocking, but nice to have.

compiler/rustc_codegen_rmc/src/codegen/function.rs
```diff
@@ -158,30 +148,49 @@ impl<'tcx> GotocCtx<'tcx> {
             ),
         };
 
-        // Now that we have the tuple, write the individual component locals
-        // back to it as a GotoC struct.
+        // When we codegen the function signature elsewhere, we will codegen the untupled version.
+        // We then marshall the arguments into a local variable holding the expected tuple.
```
avanhatt, line 152: nit: I personally don't associate the work "marshall" with what's happening here - is this a common definition?

danielsn: https://docs.microsoft.com/en-us/cpp/dotnet/overview-of-marshaling-in-cpp?view=msvc-170
It is a term used to indicate translating data between one calling convention and another.

compiler/rustc_codegen_rmc/src/codegen/function.rs
danielsn: Added comments based on the results of `--gen-c`

compiler/rustc_codegen_rmc/src/codegen/function.rs
danielsn: Done

src/test/rmc/Closure/tupled_closure.rs
danielsn: done

PR #704, opened 2021-12-17
Use NULL constant for null pointer IRep

danielsn:
### Description of changes: 

CBMC has an explicit NULL pointer value, in addition to a pointer with value 0 being an implicit null pointer. The CBMC constant propagator does a better job with the first case. This is being fixed on the CBMC side, but till then this improves performance, and doesn't cost much.   

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/compiler/cbmc/src/irep/to_irep.rs
+++ b/compiler/cbmc/src/irep/to_irep.rs
@@ -228,6 +228,11 @@ impl ToIrep for ExprValue {
                 ],
             },
             ExprValue::Nondet => side_effect_irep(IrepId::Nondet, vec![]),
+            ExprValue::PointerConstant(0) => Irep {
+                id: IrepId::Constant,
+                sub: vec![],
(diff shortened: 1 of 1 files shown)


tautschnig:
Just noting that https://github.com/diffblue/cbmc/pull/6532 will also fix the problem in that CBMC will then treat 0 and `NULL` interchangeably (which most places already did) whenever the platform configuration has `NULL_is_zero` is true, which RMC does set.

tedinski (approved):
I like less code and @tautschnig's alternative of just updating cbmc, but this is clean so I have no objection.

But what issue does this resolve @danielsn? Your pull request description is not very enlightening. :P

zhassan-aws:
@tedinski, this is an example program that was not solved by CBMC after running for more than 30 hours:
```rust
fn main() {
    let mut v = Vec::<u8>::with_capacity(2);
    v.push(0);
    v.push(0);
}
```
With this change, it is now solved in ~1 second.

Issue #42, opened 2021-04-13
Compiler interface: remove need to patch upstream API


danielsn:

celinval:
This is the current result of a `git diff` on the `compiler` folder that ignores `rustc_codegen_rmc` and `cbmc` folders.

```diff
diff --git a/compiler/rustc_codegen_llvm/src/lib.rs b/compiler/rustc_codegen_llvm/src/lib.rs
index 64fedb7bc1a..02d46414fd9 100644
--- a/compiler/rustc_codegen_llvm/src/lib.rs
+++ b/compiler/rustc_codegen_llvm/src/lib.rs
@@ -323,27 +323,24 @@ fn join_codegen(
         &self,
         ongoing_codegen: Box<dyn Any>,
         sess: &Session,
-    ) -> Result<(CodegenResults, FxHashMap<WorkProductId, WorkProduct>), ErrorReported> {
+    ) -> Result<(Box<dyn Any>, FxHashMap<WorkProductId, WorkProduct>), ErrorReported> {
         let (codegen_results, work_products) = ongoing_codegen
             .downcast::<rustc_codegen_ssa::back::write::OngoingCodegen<LlvmCodegenBackend>>()
             .expect("Expected LlvmCodegenBackend's OngoingCodegen, found Box<Any>")
             .join(sess);
-
-        sess.time("llvm_dump_timing_file", || {
-            if sess.opts.debugging_opts.llvm_time_trace {
-                llvm_util::time_trace_profiler_finish("llvm_timings.json");
-            }
-        });
-
-        Ok((codegen_results, work_products))
+        Ok((Box::new(codegen_results), work_products))
     }
 
     fn link(
         &self,
         sess: &Session,
-        codegen_results: CodegenResults,
+        codegen_results: Box<dyn Any>,
         outputs: &OutputFilenames,
     ) -> Result<(), ErrorReported> {
+        let codegen_results = codegen_results
+            .downcast::<CodegenResults>()
+            .expect("Expected CodegenResults, found Box<Any>");
+
         use crate::back::archive::LlvmArchiveBuilder;
         use rustc_codegen_ssa::back::link::link_binary;
 
diff --git a/compiler/rustc_codegen_ssa/src/traits/backend.rs b/compiler/rustc_codegen_ssa/src/traits/backend.rs
index 9c8bc3b2109..e9323b98f82 100644
--- a/compiler/rustc_codegen_ssa/src/traits/backend.rs
[...]

PR #710, opened 2021-12-20
Revert change to rustc SwitchTargets type.

celinval:
### Description of changes: 

We have changed the visibility of one of the members of SwitchTargets to have access to their values during codegen. I re-wrote the code to use standard API so we don't need to keep using a forked version of the compiler. Issue #42.

The only changes left are the ones to enable rmc codegen and a a few options we enabled. Those should be handled once we have a new driver.

### Resolved issues:

None

### Testing:

* How is this change tested? Added new test + all older tests.

### Checklist

### DIFF
--- a/src/test/rmc/Match/match_bool.rs
+++ b/src/test/rmc/Match/match_bool.rs
@@ -1,0 +1,28 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// compile-flags: --crate-type lib
+// rmc-flags: --function match_bool
+
+#[rmc::proof]
+pub fn match_bool(arg: bool) {
+    let var = match arg {
+        true => !arg,
+        _ => arg,
+    };
+
+    let var2 = match arg {
+        _ => false,
+    };
+
+    assert!(var == var2);
+
+    let mut i = 0;
+
+    match arg {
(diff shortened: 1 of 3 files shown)


src/test/rmc/Match/match_bool.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// compile-flags: --crate-type lib
+// rmc-flags: --function match_bool
+
+#[rmc::proof]
+pub fn match_bool(arg: bool) {
```
tedinski, line 7: I wouldn't use proof annotations on functions with arguments. This isn't really supported (yet!) and only sometimes sorta works.

tedinski (approved):

src/test/rmc/Match/match_bool.rs
celinval: I kind of prefer using the proof annotation than the `#[no_mangle]`. What are the downsides to it?

tedinski: I'm just suggesting switching from an argument to `let arg = rmc::nondet();` at the start.

celinval: I misunderstood you. Sorry about that. Sure, I can do that

Issue #708, opened 2021-12-20
Indexing into a slice of an array with a non-zero offset causes a CBMC invariant violation


zhassan-aws:
I tried this code:

```rust
fn main() {
    let arr = [1, 2, 3];
    let s = &arr[1..];
    assert!(s[0] == 2);
    assert!(s[1] == 3);
}
```

using the following command line invocation:

```
rmc test.rs
```

with RMC version: `371faaebc7c`

I expected to see this happen: Verification successful

Instead, this happened: CBMC invariant violation:
```
$ rmc test.rs 
CBMC version 5.46.0 (cbmc-5.46.0) 64-bit x86_64 linux
Reading GOTO program from file
Reading: test.goto
Generating GOTO Program
Adding CPROVER library (x86_64)
Removal of function pointers and virtual functions
Generic Property Instrumentation
Running with 16 object bits, 48 offset bits (user-specified)
Starting Bounded Model Checking
aborting path on assume(false) at file <builtin-library-abort> line 6 function abort thread 0
Runtime Symex: 0.00918048s
size of program expression: 318 steps
simple slicing removed 9 assignments
Generated 35 VCC(s), 7 remaining after simplification
Runtime Postprocess Equation: 1.5481e-05s
Passing problem to propositional reduction
converting SSA
--- begin invariant violation report ---
Invariant check failed
../src/solvers/flattening/boolbv_add_sub.cpp:68 function: convert_add_sub
Condition: it->type() == type
Reason: add/sub with mixed types:
+
  * type: signedbv
      * width: 64
      * #c_type: signed_long_int
  0: constant
      * type: signedbv
          * width: 64
          * #c_type: signed_long_int
      * value: 1
  1: constant
      * type: unsignedbv
          * #source_location: 
            * file: <built-in-additions>
            * line: 1
            * working_directory: /home/ubuntu/examples/new-iss
          * width: 64
          * #typedef: __CPROVER_size_t
          * #c_type: unsigned_long_int
      * value: 1
Backtrace:
cbmc(print_backtrace(std::ostream&)+0x50) [0x55863b8aa5e0]
cbmc(get_backtrace[abi:cxx11]()+0x169) [0x55863b8aab89]
[...]

PR #711, opened 2021-12-21
Add missing type constructors

danielsn:
### Description of changes: 

While working on #708 I wanted to make an element with type `ssize_t`. I discovered that there was no constructor for that type. Adding it.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Call-outs:

We should really audit this code at some point to find the rest of the missing constructors. There are probably more.

### Testing:

* How is this change tested?  Regressions still pass. Really, we should have unit tests for these things, but that's a future task.

### Checklist

### DIFF
--- a/compiler/cbmc/src/goto_program/typ.rs
+++ b/compiler/cbmc/src/goto_program/typ.rs
@@ -369,7 +369,7 @@ impl Type {
         }
     }
 
-    pub fn is_c_sizet(&self) -> bool {
+    pub fn is_c_size_t(&self) -> bool {
         match self {
             Type::CInteger(CIntType::SizeT) => true,
             _ => false,
--- a/compiler/cbmc/src/goto_program/typ.rs
+++ b/compiler/cbmc/src/goto_program/typ.rs
@@ -369,6 +369,20 @@ impl Type {
         }
     }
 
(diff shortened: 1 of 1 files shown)


tedinski (approved):
I assume you will be using these in an upcoming PR?

compiler/cbmc/src/goto_program/typ.rs
```diff
@@ -369,6 +369,20 @@ impl Type {
         }
     }
 
+    pub fn is_c_sizet(&self) -> bool {
```
zhassan-aws, line 372: Minor:
```suggestion
    pub fn is_c_size_t(&self) -> bool {
```

compiler/cbmc/src/goto_program/typ.rs
```diff
@@ -369,6 +369,20 @@ impl Type {
...
+        match self {
+            Type::CInteger(CIntType::SizeT) => true,
+            _ => false,
+        }
+    }
+
+    pub fn is_c_ssizet(&self) -> bool {
```
zhassan-aws, line 379:
```suggestion
    pub fn is_c_ssize_t(&self) -> bool {
```

zhassan-aws (approved):

PR #712, opened 2021-12-21
Add a hint for how to fix missing RMC library

danielsn:
### Description of changes: 

Currently, when the RMC library hasn't been built, we fail with an error message, but no indication to the user of how to fix it. Add to the error message the script they can run to fix it.
### Resolved issues:

Resolves #ISSUE-NUMBER

### Call-outs:
This only works if the user has built RMC from source, and knows where the repo with the scripts are.  Could we give them a reference to the absolute path to run the script from?

### Testing:

* How is this change tested? Ran it locally, saw the new message.

* Is this a refactor change? Documentation / error message fix.

### Checklist

### DIFF
--- a/scripts/rmc-rustc
+++ b/scripts/rmc-rustc
@@ -41,6 +41,7 @@ set_rmc_lib_path() {
             echo "ERROR: Could not find RMC library."
             echo "Looked for: \"$REPO_DIR/target/*/librmc.rlib\""
             echo "Was RMC library successfully built first?"
+            echo "Try running <RMC_ROOT>/scripts/setup/build_rmc_lib.sh"
             exit 1
         fi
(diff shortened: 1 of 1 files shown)


tedinski (approved):

scripts/rmc-rustc
```diff
@@ -41,6 +41,7 @@ set_rmc_lib_path() {
             echo "ERROR: Could not find RMC library."
             echo "Looked for: \"$REPO_DIR/target/*/librmc.rlib\""
             echo "Was RMC library successfully built first?"
+            echo "Try running <RMC_ROOT>/scripts/setup/build_rmc_lib.sh"
```
tautschnig, line 44: Should the line above ("Was RMC library successfully built first?") be removed? It obviously wasn't, hence the newly added guidance.

danielsn: Good point, will remove that.

PR #714, opened 2021-12-22
Documentation: Limitations

adpaco-aws:
### Description of changes: 

No changes to RMC. Adds a new section "Limitations" to documentation in order to summarize the current status in regards to Rust feature support using a table.

Preview [here](https://github.com/adpaco-aws/rmc/blob/support-table/rmc-docs/src/limitations.md)

### Call-outs:

Open as draft so I can get some feedback.

### Testing:

* How is this change tested? Running remote server for documentation.

### Checklist

### DIFF
--- a/rmc-docs/src/limitations.md
+++ b/rmc-docs/src/limitations.md
@@ -142,9 +142,8 @@ was sequential code.
 ### Standard library functions
 
 At present, RMC is able to link in functions from the standard library but the
-generated code will not contain them unless they are inlined. This results in
-verification failures if the code under verification, for example, includes a
-`println!` statement.
+generated code will not contain them. This results in verification failures if
+the code under verification, for example, includes a `println!` statement.
 
 We have done some experiments to embed the standard library into the generated
 code, but this causes verification times to increase significantly. As of now,
@@ -154,11 +153,10 @@ for future work in this direction.
 
 ### Advanced features
 
-Advanced features (traits, types, etc.) from Rust have proven to particularly
-difficult to model. Even though the code handling these features has been
-reviewed in multiple occasions, we often find new corner cases that are not
-covered by RMC. We are especially interested in bug reports concerning these
-features, so please [file a bug
+The semantics around some advanced features (traits, types, etc.) from Rust are
+not formally defined which makes it harder to ensure that we can properly model
+all their use cases. We are particularly interested in bug reports concerning
+these features, so please [file a bug
 report](https://github.com/model-checking/rmc/issues/new?assignees=&labels=bug&template=bug_report.md)
 if you are aware of one.
 
--- a/rmc-docs/src/limitations.md
+++ b/rmc-docs/src/limitations.md
@@ -2,6 +2,14 @@
 
 ## Rust feature support
 
+The table below tries to summarize the current support in RMC for
+the Rust language features according to the [Rust Reference](https://doc.rust-lang.org/stable/reference/).
+We use three values to indicate the level of support:
+ * *Yes*: The feature is fully supported. There may
+ * *Partial*: The feature is at least partially supported. This means we are aware of some issue
+ with the feature.
+ * *No*:
+
 *Define Yes, No and Partial*
 
 Reference | Feature | Support | Observations |
@@ -90,7 +98,7 @@ Reference | Feature | Support | Observations |
 15.3 | Behavior considered undefined | Partial | |
 | | Data races | No | [Concurrency](#concurrency) |
 | | Dereferencing dangling raw pointers | Yes | |
-| | Dereferencing unaligned  raw pointers | No | |
+| | Dereferencing unaligned raw pointers | No | |
 | | Breaking pointer aliasing rules | No | |
 | | Mutating immutable data | No | |
 | | Invoking undefined behavior via compiler intrinsics | Partial | [Intrinsics](#intrinsics) |
--- a/rmc-docs/src/limitations.md
+++ b/rmc-docs/src/limitations.md
@@ -144,11 +144,11 @@ compiles as if it was sequential code.
 ### Standard library functions
 
 At present, RMC is able to link in functions from the standard library but the
-generated code will not contain them unless they generic, intrinsics, inlined or
-macros. Missing functions are treated in a similar way to unsupported features
-(i.e., replacing the function body with an `assert(false)` statement). This
-results in verification failures if the code under verification, for example,
-includes a reachable `println!` statement.
+generated code will not contain them unless they are generic, intrinsics,
+inlined or macros. Missing functions are treated in a similar way to unsupported
+features (i.e., replacing the function body with an `assert(false)` statement).
+This results in verification failures if the code under verification, for
+example, includes a reachable `println!` statement.
 
 We have done some experiments to embed the standard library into the generated
 code, but this causes verification times to increase significantly. As of now,
@@ -420,10 +420,10 @@ write_bytes | Yes | |
 #### Atomics
 
 All atomic intrinsics are compiled as an atomic block where the operation is
-performed. But as noted in [Notes - Concurrency](#concurrency), CBMC
-support for concurrent verification is limited and not used by default.
-Verification on code containing atomic intrinsics should not be trusted given
-that CBMC assumes the code to be sequential.
+performed. But as noted in [Notes - Concurrency](#concurrency), RMC support for
+concurrent verification is limited and not used by default. Verification on code
+containing atomic intrinsics should not be trusted given that RMC assumes the
+code to be sequential.
 
 #### SIMD instructions
 
--- a/rmc-docs/src/limitations.md
+++ b/rmc-docs/src/limitations.md
@@ -1,0 +1,120 @@
+# Limitations
+
+## Rust feature support
+
+*Define Yes, No and Partial*
+
+Reference | Feature | Support | Observations |
+--- | --- | --- | --- |
+3.1 | Macros By Example | Yes | |
+3.2 | Procedural Macros | Yes | |
+4 | Crates and source files | Yes | |
+5 | Conditional compilation | Yes | |
+6.1 | Modules | Yes | |
+6.2 | Extern crates | Yes | |
+6.3 | Use declarations | Yes | |
+6.4 | Functions | Yes | |
+6.5 | Type aliases | Yes | |
+6.6 | Structs | Yes | |
+6.7 | Enumerations | Yes | |
+6.8 | Unions | Yes | |
+6.9 | Constant items | Yes | |
+6.10 | Static items | Yes | |
+6.11 | Traits | Yes | |
+6.12 | Implementations | Yes | |
+6.13 | External blocks | Yes | |
+6.14 | Generic parameters | Yes | |
+6.15 | Associated Items | Yes | |
+7 | Attributes | Yes | |
+8.1 | Statements | Yes | |
+8.2.1 | Literal expressions | Yes | |
+8.2.2 | Path expressions | Yes | |
+8.2.3 | Block expressions | Yes | |
+8.2.4 | Operator expressions | Yes | |
+8.2.5 | Grouped expressions | Yes | |
+8.2.6 | Array and index expressions | Yes | |
+8.2.7 | Tuple and index expressions | Yes | |
+8.2.8 | Struct expressions | Yes | |
+8.2.9 | Call expressions | Yes | |
+8.2.10 | Method call expressions | Yes | |
+8.2.11 | Field access expressions | Yes | |
+8.2.12 | Closure expressions | Yes | |
+8.2.13 | Loop expressions | Yes | |
+8.2.14 | Range expressions | Yes | |
+8.2.15 | If and if let expressions | Yes | |
+8.2.16 | Match expressions | Yes | |
+8.2.17 | Return expressions | Yes | |
+8.2.18 | Await expressions | No | [Concurrency](#concurrency) |
+9 | Patterns | Partial | [Patterns](#patterns) |
+10.1.1 | Boolean type | Yes | |
+10.1.2 | Numeric types | Yes | |
+10.1.3 | Textual types | Yes | |
+10.1.4 | Never type | Yes | |
+10.1.5 | Tuple types | Yes | |
+10.1.6 | Array types | Yes | |
+10.1.7 | Slice types | Yes | |
+10.1.8 | Struct types | Yes | |
+10.1.9 | Enumerated types | Yes | |
+10.1.10 | Union types | Yes | |
+10.1.11 | Function item types | Yes | |
+10.1.12 | Closure types | Partial | |
+10.1.13 | Pointer types | Partial | |
+10.1.14 | Function pointer types | Partial | |
+10.1.15 | Trait object types | Partial | |
+10.1.16 | Impl trait type | Partial | |
+10.1.17 | Type parameters | Partial | |
+10.1.18 | Inferred type | Partial | |
+10.2 | Dynamically Sized Types | Partial | |
+10.3 | Type layout | Yes | |
+10.4 | Interior mutability | Yes | |
+10.5 | Subtyping and Variance | Yes | |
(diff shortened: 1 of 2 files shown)


rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,401 @@
...
+Because of that, the general rule is that RMC generates an `assert(false)`
+statement when compiling any unsupported feature. This will cause proofs to fail
+if the statement is reachable during the verification stage. However, the
+analysis will not affected if the statement is not reachable from the code under
+verification, so users can still verify components of their code not using
+unsupported features.
+
+### Assembly
+
+CBMC support for assembly code is quite limited. It may be extended in the
```
celinval, line 126: I would say RMC instead of CBMC here. At this point, I would only mention CBMC if you want to add some link to CBMC assembly support.

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,401 @@
...
+Check out the tracking issues for [inline assembly (`asm!`
+macro)](https://github.com/model-checking/rmc/issues/2) and [global assembly
+(`asm_global!` macro)](https://github.com/model-checking/rmc/issues/316) to know
+more about the current status.
+
+### Concurrency
+
+Concurrent features are out of scope for RMC. In general, the verification of
+concurrent programs continues to be an open research problem. In particular,
+support for concurrent verification in CBMC is limited and RMC emits a warning
```
celinval, line 138: The same here.

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,401 @@
...
+whenever it encounters concurrent code.
+
+### Not inlined standard library functions
```
celinval, line 141: Maybe change the title to standard library functions.

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,401 @@
...
+
+At present, RMC is able to link in functions from the standard library but the
+generated code will not contain them unless they are inlined. This results in
```
celinval, line 144: We should add that we support to generic functions and macros at this point. They get monomorphized / expanded during the compilation. BTW, I don't think we are running the rustc pass to inline functions.

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,401 @@
...
+verification failures if the code under verification, for example, includes a
+`println!` statement.
+
+We have done some experiments to embed the standard library into the generated
```
celinval, line 148: Did we run these experiments before or after enabling the code slicing?

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,401 @@
...
+code, but this causes verification times to increase significantly. As of now,
+we have not been able to find a simple solution for [this
+issue](https://github.com/model-checking/rmc/issues/581), but we have some ideas
+for future work in this direction.
+
+### Advanced features
+
+Advanced features (traits, types, etc.) from Rust have proven to particularly
```
celinval, line 156: I would say something like: the semantics around some of these rust advanced features are not formally defined which makes it harder to ensure that we can properly model all their use cases. If you discover any case that cannot be handled properly by RMC, please file a bug report.

celinval (changes requested):
Thanks for doing this. I just have a few minor comments.

Can you please add that we currently do not support stack unwinding?

rmc-docs/src/limitations.md
adpaco-aws: Not sure why you want to add it here. I think the table already includes this information and I'm not sure what it has to do with the standard library.

Removed "unless they are inlined" from this block.

rmc-docs/src/limitations.md
adpaco-aws: If you refer to `--drop-unused-functions`, I tried both when doing these experiments. It reduced the generated code considerably, but still kept many other functions around which made this approach unfeasible.

rmc-docs/src/limitations.md
adpaco-aws: Thanks! I liked the first part a lot so I included it in here. Let me know if you want to change it back.

adpaco-aws:
Thanks @celinval ! Updated the PR with some of your suggestions, let me know if you have other comments 😄

adpaco-aws:
Forgot to mention stack unwinding - Added a new section on panic strategies just now.

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+### Code generation for unsupported features
+
+RMC aims to be a industrial verification tool. Most industrial crates may
+include unsupported features in parts of their code that do not need to be
+verified. In general, this should not prevent users from verifying their code.
+
+Because of that, the general rule is that RMC generates an `assert(false)`
+statement when compiling any unsupported feature. This will cause proofs to fail
+if the statement is reachable during the verification stage. However, the
+analysis will not affected if the statement is not reachable from the code under
```
tedinski, line 120: analysis will not *be* affected

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+verification, so users can still verify components of their code not using
```
tedinski, line 121: ~not using~ that do not use

(I think your phrasing is technically fine, it is just potentially ambiguous what 'not using' refers to (the code, not the verification) until you read after it in the sentence.)

(Hopefully not too nitpicky here, but it's public facing documentation, so I'm looking for these things)

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+performed. But as noted in [Observations - Concurrency](#concurrency), CBMC
+support for concurrent verification is limited and not used by default.
+Verification on code containing atomic intrinsics should not be trusted given
+that CBMC assumes the code to be sequential.
+
+#### SIMD instructions
+
+While RMC is capable of generating code for SIMD instructions. Unfortunately,
+CBMC does not provide support for some operations like vector comparison (e.g.,
+`simd_eq`) or shuffling (e.g., `simd_shuffle8`).
```
tedinski, line 413: wait, I thought I added shuffle.

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+generated code will not contain them. This results in verification failures if
+the code under verification, for example, includes a `println!` statement.
+
+We have done some experiments to embed the standard library into the generated
+code, but this causes verification times to increase significantly. As of now,
+we have not been able to find a simple solution for [this
+issue](https://github.com/model-checking/rmc/issues/581), but we have some ideas
+for future work in this direction.
+
+### Advanced features
```
tedinski, line 154: I worry this section might be too vague. It might be a good idea to back-reference what we say in the table? (E.g. there's a lot of links from the table to here that seem to be about projections, and I think our goal is to have to solid by 1.0.)

I think over-technical would be better than vague. If this section answered for us "what would it take for us to remove this section from this doc because we know we support it now" I think that's be better than being vague.

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+We use the following values to indicate the level of support:
+ * **Yes**: The feature is fully supported. We are not aware of any issue with it.
+ * **Partial**: The feature is at least partially supported. We are aware of some issue with
+ with it.
+ * **No**: The feature is not supported. Some support may be available but analyses should not be trusted.
+
+As with all software, bugs may be found anywhere regardless of the level of support. In such cases, we
+would greatly appreciate that you [filed a bug report](https://github.com/model-checking/rmc/issues/new?assignees=&labels=bug&template=bug_report.md).
+
+Reference | Feature | Support | Observations |
```
zhassan-aws, line 16: Would "Remarks" or "Notes" be more suitable than "Observations"?

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+| | Mutating immutable data | No | |
+| | Invoking undefined behavior via compiler intrinsics | Partial | See [Observations - Intrinsics](#intrinsics) |
+| | Executing code compiled with platform features that the current platform does not support | No | |
+| | Producing an invalid value, even in private fields and locals | No | |
+
+## Observations on partially or unsupported features
+
+### Code generation for unsupported features
+
+RMC aims to be a industrial verification tool. Most industrial crates may
```
zhassan-aws, line 113:
```suggestion
RMC aims to be an industrial verification tool. Most industrial crates may
```
Not sure I like the word "industrial", but can't think of an alternative.

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+include unsupported features in parts of their code that do not need to be
+verified. In general, this should not prevent users from verifying their code.
```
zhassan-aws, line 115:
```suggestion
verified. In general, this should not prevent users from using RMC to verify their code.
```

rmc-docs/src/limitations.md
celinval: Sorry if I wasn't clear. I was talking about the "unless they are inlined". The correct would be "unless they are generic, intrinsics, inlined or macros." :)

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+
+Because of that, the general rule is that RMC generates an `assert(false)`
```
zhassan-aws, line 117: Technically, it generates `assert(false)` followed by `assume(false)`, I think this piece of detail is worth mentioning since it may cause some asserts to become unreachable, and hence vacuously pass.

rmc-docs/src/limitations.md
celinval: did your experiments include tests with `println`?

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+statement when compiling any unsupported feature. This will cause proofs to fail
```
zhassan-aws, line 118:
```suggestion
statement when compiling any unsupported feature. This will cause verification to fail
```

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+if the statement is reachable during the verification stage. However, the
+analysis will not affected if the statement is not reachable from the code under
+verification, so users can still verify components of their code not using
+unsupported features.
+
+### Assembly
```
zhassan-aws, line 124: I think the official term is "Inline Assembly".

celinval (approved):
It looks good to me. Please make sure you address Ted's feedback before pushing. :)

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+present there are no plans to do so.
+
+Check out the tracking issues for [inline assembly (`asm!`
+macro)](https://github.com/model-checking/rmc/issues/2) and [global assembly
+(`asm_global!` macro)](https://github.com/model-checking/rmc/issues/316) to know
+more about the current status.
+
+### Concurrency
+
+Concurrent features are out of scope for RMC. In general, the verification of
```
zhassan-aws, line 136: Should we add "currently"?
```suggestion
Concurrent features are currently out of scope for RMC. In general, the verification of
```

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+concurrent programs continues to be an open research problem where most tools
+that analyze concurrent code lack support for other features. Because of this,
+RMC emits a warning whenever it encounters concurrent code and compiles as if it
+was sequential code.
```
zhassan-aws, line 140: Does RMC pick some order for execution of concurrent code? If so, we should point out that this may lead to RMC missing bugs that may occur with a different execution order.

zhassan-aws (commented):
First batch of comments.

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+
+### Standard library functions
+
+At present, RMC is able to link in functions from the standard library but the
+generated code will not contain them. This results in verification failures if
```
zhassan-aws, line 145: > but the generated code will not contain them

This sentence a bit vague. Can we clarify it? Does RMC generate these functions with an empty body?

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,413 @@
...
+
+All atomic intrinsics are compiled as an atomic block where the operation is
+performed. But as noted in [Observations - Concurrency](#concurrency), CBMC
+support for concurrent verification is limited and not used by default.
+Verification on code containing atomic intrinsics should not be trusted given
+that CBMC assumes the code to be sequential.
+
+#### SIMD instructions
+
+While RMC is capable of generating code for SIMD instructions. Unfortunately,
```
zhassan-aws, line 411:
```suggestion
While RMC is capable of generating code for SIMD instructions, unfortunately,
```

rmc-docs/src/limitations.md
adpaco-aws: AFAIK it assumes all code to be sequential. I think there are some CBMC options for concurrent code verification but we do not make use of them at the moment.

rmc-docs/src/limitations.md
adpaco-aws: Added a sentence to mention these are treated in a similar way to unsupported features. Let me know what you think!

rmc-docs/src/limitations.md
adpaco-aws: Yes, in fact they just included `println`.

rmc-docs/src/limitations.md
adpaco-aws: I was not sure how to make this more technical, so I highlighted some outstanding issues here.

Please let me know how it looks and if it does not look good to you we can work together on this section.

rmc-docs/src/limitations.md
adpaco-aws: Sorry, I thought this was not supported by CBMC but after reviewing related issues I see that you implemented shuffle operations in RMC. So I removed shuffling from here.

tedinski (approved):

rmc-docs/src/limitations.md
zhassan-aws: One possibility (I don't feel strongly about it either): RMC is intended to be used on real-world Rust code. Most real-world crates ...

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,432 @@
...
+Concurrent features are currently out of scope for RMC. In general, the
+verification of concurrent programs continues to be an open research problem
+where most tools that analyze concurrent code lack support for other features.
+Because of this, RMC emits a warning whenever it encounters concurrent code and
+compiles as if it was sequential code.
+
+### Standard library functions
+
+At present, RMC is able to link in functions from the standard library but the
+generated code will not contain them unless they generic, intrinsics, inlined or
```
zhassan-aws, line 147:
```suggestion
generated code will not contain them unless they are generic, intrinsics, inlined or
```

rmc-docs/src/limitations.md
```diff
@@ -0,0 +1,432 @@
...
+volatile_store | No | See [Notes - Concurrency](#concurrency) |
+wrapping_add | Yes | |
+wrapping_mul | Yes | |
+wrapping_sub | Yes | |
+write_bytes | Yes | |
+
+#### Atomics
+
+All atomic intrinsics are compiled as an atomic block where the operation is
+performed. But as noted in [Notes - Concurrency](#concurrency), CBMC
```
zhassan-aws, line 423: Should we replace the two occurrences of CBMC in this section by RMC?

zhassan-aws (approved):

rmc-docs/src/limitations.md
adpaco-aws: Yes, done.

rmc-docs/src/limitations.md
adpaco-aws: Thanks!

PR #718, opened 2021-12-29
Add a new issue template for code improvements

celinval:
### Description of changes: 

New template for code improvements.

### Resolved issues:

n/a

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


.github/ISSUE_TEMPLATE/code_improvement.md
```diff
@@ -0,0 +1,28 @@
...
+
+<!--
+Thank you for requesting an RMC improvment!
+Please include the following information if it applies to your case:
+-->
+**Requested improvement:**
+
+**Use case:**
+
+**Is this a breaking change?**
```
adpaco-aws, line 19: Do this kind of issues
 1. require a use case?
 2. cause breaking changes?

.github/ISSUE_TEMPLATE/code_improvement.md
```diff
@@ -0,0 +1,28 @@
...
+about: Create a request for improving the code quality of RMC
+labels: Cat: cleanup, Cat: enhancement 
+---
+<!--
+Please report security concerns to us on the
+[AWS vulnerability reporting](http://aws.amazon.com/security/vulnerability-reporting/) page.
+-->
+
+<!--
+Thank you for requesting an RMC improvment!
```
adpaco-aws, line 12:
```suggestion
Thank you for requesting an RMC improvement!
```

.github/ISSUE_TEMPLATE/code_improvement.md
celinval: The use case I was thinking as existing use cases that trigger this code not new ones. I can remove it or rename it. Suggestions?

I think there may be cases where a code improvement can cause breaking changes. For example, deprecating some ugly options and stuff.

adpaco-aws: I see, thanks. Maybe we could use "Related features" instead of "Use case", but I would expect "Requested improvement" to already include that information in most cases. But we can try with this and see how it goes.

adpaco-aws (approved):

Issue #721, opened 2021-12-29
Improve rmc-compiler configuration when using cargo rmc


celinval:
**Requested change:** Figure out a better mechanism to pass arguments from cargo rmc to rmc-compiler.

**Use case:** The rmc-rustc  handles the hack to connect cargo rmc to rmc. We should investigate a more reliable mechanism that doesn't rely on environment variables.

Note: I tried to bundle the options as a sub-option but cargo doesn't seem to handle quotes very well.

**Is this a breaking change?** No

tedinski:
Is this still an issue? Certainly `rmc-rustc` is gone...

celinval:
Yes, this is still a problem. We should be passing arguments via RUSTC_FLAGS since `cargo` uses that information to compute whether a crate should be recompiled or not.

We will need to fix this in order to properly support incremental compilation.

celinval:
Closing this since it will be fixed as part of #2137

Issue #720, opened 2021-12-29
RMC libraries should be build and put into a standard place


celinval:
**Requested change:** We should ensure that RMC libraries are built and placed in a standard place for rmc-compiler find them without relying on rmc-rustc script. Ideally, this solution should also work with `cargo install`. We should keep supporting the --rmc-lib argument in the rare cases the user wants to override this configuration.

**Use case:** Improve installation of RMC.

**Is this a breaking change?** No

This issue depends on #719

tedinski:
I believe this was resolved when we built the installation process.

Issue #719, opened 2021-12-29
Improve project layout


celinval:
**Requested change:** Create a cargo workspace for RMC that is not dependent on RUSTC workspace.

**Use case:** This will allow us to verify that we no longer depend on any of the bootstrap script functionality to keep RMC infrastructure. This will also allow us to define which rust toolchain to use for all RMC packages and tools.

**Is this a breaking change?** No

The final layout proposed for this issue is:
```rust
.
├── Cargo.lock
├── Cargo.toml
├── rust-toolchain.toml
├── rustfmt.toml
├── README.md
├── ...    // Other files such as .gitignore, .github/*`
├── src/   // Kani source code split into crates per component.
│ ├── cargo-kani/ // Kani entry point for standalone kani and cargo-kani.
│ │ ├── Cargo.toml
│ │ ├── src/
│ │ └── ...
│ ├── kani_driver/
│ │ ├── Cargo.toml
│ │ ├── src
│ │ └── ...
│ ├── kani_compiler/
│ │ ├── Cargo.toml
│ │ ├── src
│ │ └── ...
│ ├── kani_report/
│ │ ├── Cargo.toml
│ │ ├── src
│ │ └── ...
│ └── ...
├── docs/
│ ├── book.toml
│ ├── src/
│ ├── SUMMARY.md
│ └── ...
├── library/
│ ├── kani/
│ └── kani-macros/
├── scripts/
│ ├── setup/
│ ├── regression.sh
│ └── ...
├── tests/  // End to end tests that are executed via compiletest
│ ├── expected/
│ ├── cargo-kani/
│ └── kani/
└── tools/
   ├── compiletest/
   └── bookrunner/
```

celinval:
We should create a workspace and update our format checker script.

celinval:
We might also want to get rid of all the Cargo.toml inside src/ so we can pack everything as part of one package.

Issue #658, opened 2021-11-19
Create RMC compiler driver


celinval:
**Requested feature:** We should be able to configure the compiler without having to modify its code. 
**Use case:** Provide a robust and clean compilation module for RMC.
**Link to relevant documentation (Rust reference, Nomicon, RFC):**
**Is this a breaking change?** No

**Success criteria:** We replace rmc-rustc by a rust application call. I.e., this issue only covers the creation of the RMC compiler driver on the top of `rustc`. Follow up issues will track moving more logic out of our python scripts into an RMC application.

celinval:
We still have rmc-rustc wrapper, but it's mostly needed to overcome a limitation of `cargo rmc`. I'll close this issue since it has captured the most important issue which is the dependency on a modified rustc and its bootstrap script.

PR #722, opened 2021-12-29
Add rmc-compiler binary to drive rust compilation

### LINKED ISSUE #42: Compiler interface: remove need to patch upstream API

celinval:
### Description of changes: 

I have created a new compilation driver (rmc-compiler) that no longer depends on a modified version of the rust compiler, neither depends on the bootstrap script. 

### Resolved issues:

Resolves #42 and #658

### Call-outs:

I'm not quite happy with a few things in this change, but it is getting way too big to manage in a separate branch. It is also getting quite big to be reviewed. Thus, I created the following issues to track a few improvements:

- #719: Improve the layout of our packages and create a new workspace.
- #720: Improve how we are building our libraries. For now, the rmc-compiler has a build-dependency that ensures the libraries get built with the same toolchain as rmc-compiler.
- #721: Improve the interface between cargo rmc and the rmc-compiler. The solution here is a bit hacky. There are a few tricky things here. Cargo invoke rustc to retrieve its version as well as to compile build scripts which shouldn't use the goto-c backend.

### Testing:

* How is this change tested? All our tests.

### Checklist


src/rmc-compiler/build.rs
```diff
@@ -0,0 +1,15 @@
...
+use std::path::PathBuf;
+
+/// Configure the compiler to build rmc-compiler binary. We currently support building
+/// rmc-compiler with nightly only. We also link to the rustup rustc_driver library for now.
+pub fn main() {
+    // Add rustup to the rpath in order to properly link with the correct rustc version.
+    let rustup_home = env::var("RUSTUP_HOME").unwrap();
+    let rustup_tc = env::var("RUSTUP_TOOLCHAIN").unwrap();
+    let lib_path = [&rustup_home, "toolchains", &rustup_tc, "lib"].iter().collect::<PathBuf>();
+    println!("cargo:rustc-link-arg-bin=rmc-compiler=-Wl,-rpath,{}", lib_path.to_string_lossy());
```
tedinski, line 14: Just checking in here, because I'll have to work on the install later, that you're dynamic linking against a `rustup` provided rust compiler library?

Is there documentation on how this works somewhere? (e.g. I'm used to rust's normal static linking, so dynamic linking to a rust library is not something I'm familiar with)

Also, this might complication installation later. For one, we need the correct relative RPATH here not an absolute path like it is now, but also we'll need to distribute the correct rustc version, right? Or is there a stable-ish ABI here?

If there's not documentation about this, I think we need to write one.

src/rmc-compiler/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+        "--extern",
+        "rmc",
+        "-L",
+        rmc_deps.as_str(),
+    ];
+    args.iter().map(|s| s.to_string()).collect()
+}
+
+fn parser<'a, 'b>() -> App<'a, 'b> {
+    app_from_crate!()
```
tedinski, line 54: Do you have a preference for `clap`? I was starting work on cargo-rmc with `structopt` which is a wrapper around `clap`. Just curious if you think I shouldn't be using that for some reason (it's not too late for me to abort and turn around...)

src/rmc-compiler/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+    app_from_crate, crate_authors, crate_description, crate_name, crate_version, App, AppSettings,
+    Arg, ArgMatches,
+};
+use rmc_queries::{QueryDb, UserInput};
+use rustc_driver::{init_env_logger, install_ice_hook, Callbacks, RunCompiler};
+use std::path::PathBuf;
+use std::rc::Rc;
+
+/// This function generates all rustc configurations required by our goto-c codegen.
+fn rustc_gotoc_flags(lib_path: &str) -> Vec<String> {
```
tedinski, line 25: Do you have any objection to my moving this into `cargo-rmc` in the future instead of having a `--goto-c` flag to turn them on? That was my original plan. (This is fine for now ofc)

src/rmc-compiler/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+        }
+        _ => None,
+    }
+}
+
+/// Get the sysroot, following the order bellow:
+/// - "--sysroot" command line argument
+/// - compile-time environment
+///    - $SYSROOT
+///    - $RUSTUP_HOME/toolchains/$RUSTUP_TOOLCHAIN
```
tedinski, line 183: are these environment variables typically provided by cargo? Or is this just another option we're creating to supply them?

scripts/rmc-rustc
```diff
@@ -16,72 +16,36 @@ RMC_PATH=${RMC_PATH:-""}
...
-set_rmc_path
-if [ "${1:-''}" == "--rmc-path" ]
+set_paths
+
+# Hack to enable cargo rmc
+args="${*#--rmc-flags}"
+if [ "$args" != "$*" ]
 then
-    echo ${RMC_PATH}
+    FLAGS=("--rmc-lib ${RMC_LIB_PATH} ${RMCFLAGS} $args")
```
tedinski, line 46: what is `${RMCFLAGS}` here. I mean, I can search this diff, but there should be a comment saying what's happening here.

library/rmc_macros/src/lib.rs
```diff
@@ -34,7 +34,7 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
     // )
 }
 
-#[cfg(rmc)]
+#[cfg(feature = "rmc")]
```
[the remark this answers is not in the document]
celinval: No. This is actually the only I could find to turn on and off this configuration via cargo. I thought the plan was for users to be able to use cargo here.

src/rmc-compiler/build.rs
celinval: Yes, you are correct. It is linking to a rustup shared library.  We will need to make it relative if we want to distribute the binaries or if we eventually install it via rustup.

This should work if the user wants to build / install it with cargo.

src/rmc-compiler/src/main.rs
celinval: I don't have any strong preference. I haven't done much research since this code is throw away code. Whatever works.

src/rmc-compiler/src/main.rs
celinval: Let's talk more about this one. I'm not 100% sure what the solution that you are proposing is.

src/rmc-compiler/src/main.rs
celinval: Yes, they are usually provided by cargo. This is similar to code from clippy and Miri.

scripts/rmc-rustc
celinval: Sure

celinval: It's set by rmc.py because cargo does some funny stuff with RUSTFLAGS.

tedinski:
Also, if I'm not mistaken, if we eliminate our remaining use of `x.py` (just compiletest and dashboard/bookrunner, right?) CI runtimes are about to be _outstanding_!

library/rmc_macros/src/lib.rs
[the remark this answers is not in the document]
tedinski: So users should be putting `[dev-dependencies] rmc = "1.0"` or whatever into their Cargo.toml, but that means when cargo-rmc does its build/codegen of the project `cfg(rmc)` would be set during that build.

So I was suggesting the current approach of building this library at the same time as `rmc-compiler` is a temporary hack (to ensure it's already built to support the current `rmc` script).

src/rmc-compiler/build.rs
tedinski: Do you know where docs about this are? I searched "rust shared library rustc_driver" and I'm not finding anything helpful. I'm not sure how fragile this approach is. (But I sure like what it does for CI times! Just trying to understand the implications for installation later...)

library/rmc_macros/src/lib.rs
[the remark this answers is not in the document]
celinval: That was a suggestion from Daniel to pre-build the rmc library. I can pre-build with the rmc config enable using RUSTFLAGS then. Let me try this and update the PR.

src/rmc-compiler/build.rs
celinval: I couldn't find any doc either. I did a lot of reverse engineering in other tools that use rustc to figure this out and I also talked to a few folks.

This seems to be the de facto way to build tools on the top of rustc. See https://github.com/rust-lang/rust-clippy/blob/830f2205d4002b26f7fd36d8467fb491eb2306ac/src/driver.rs, https://github.com/rust-lang/miri/blob/035933186957cf81c488261fb48a98bf523e8006/src/lib.rs and this little example: https://doc.rust-lang.org/beta/unstable-book/language-features/plugin.html?highlight=rustc_private#lint-plugins

celinval:
> Also, if I'm not mistaken, if we eliminate our remaining use of `x.py` (just compiletest and dashboard/bookrunner, right?) CI runtimes are about to be _outstanding_!

For sure! This change reduced the CI in about 10min. @adpaco-aws and I are working on the second piece. After that, we should be able to prune our repository, remove the bootstrap script and no longer depend on the rust upstream. :)

src/rmc-compiler/rust-toolchain.toml
```diff
@@ -0,0 +1,6 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+[toolchain]
+channel = "nightly-2021-12-27"
```
tedinski, line 5: Does this cause cargo to download this toolchain, or does it fail or something if it's not present?

celinval: It will download it.

celinval: I believe any rustup command will trigger the download

src/rmc-compiler/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+        )
+        .arg(
+            // TODO: Move this to a cargo wrapper. This should return rmc version.
+            Arg::with_name("rustc-version")
+                .short("V")
+                .long("--version")
+                .help("Gets underlying rustc version."),
+        )
+        .arg(
+            Arg::with_name("rustc-options")
```
tedinski, line 101: ah, I see, so here's where we have the rustc flags after our flags, and we have the problem with cargo mixing them up. And that's why we need `RMCFLAGS` in `rmc-rustc` for now.

tedinski (approved):
LGTM. Think DSN should review too?

celinval:
> LGTM. Think DSN should review too?

I think it would be nice to have someone else reviewing it. Let me ping @danielsn.

src/rmc-compiler/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+    // Configure and run compiler.
+    let mut callbacks = RmcCallbacks {};
+    install_ice_hook();
+    let mut compiler = RunCompiler::new(&rustc_args, &mut callbacks);
+    if matches.is_present("goto-c") {
+        compiler.set_make_codegen_backend(Some(Box::new(move |_cfg| {
+            rustc_codegen_rmc::GotocCodegenBackend::new(&Rc::new(queries))
+        })));
+    }
+    compiler.run().or(Err("Failed to compile crate."))
```
danielsn, line 133: Will this give enough debug info?

src/rmc-compiler/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+/// Empty struct since we don't support any callbacks yet.
+struct RmcCallbacks {}
```
danielsn, line 137: How will this be used once we do?

src/rmc-compiler/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+    rustc_args.append(
+        &mut args
+            .values_of("rustc-options")
+            .unwrap_or(clap::Values::default())
+            .map(|s| s.to_string())
+            .collect::<Vec<_>>(),
+    );
+    let sysroot = sysroot_path(args.value_of("sysroot")).unwrap();
+    rustc_args.push(String::from("--sysroot"));
+    rustc_args.push(sysroot.to_string_lossy().to_string());
```
danielsn, line 164: Is `to_string_lossy` what we want here? If there are non-valid characters, shouldn't we just fail?

src/rmc-compiler/README.md
```diff
@@ -0,0 +1,8 @@
+This is a temporary wrapper that can be used to compiler rust into gotoc. This
+binary should not be used on its own and it should be used via `rmc` or
+`cargo-rmc` commands.
+
+### Notes for developers:
+
+This binary can be build like a regular cargo package. There is no need to
```
danielsn, line 7: built

danielsn (approved):
Some questions, but LGTM

scripts/cargo-rmc
```diff
@@ -26,6 +26,9 @@ def main():
     if args.gen_c_runnable:
         rmc.cargo_build(args.crate, args.target_dir, args, ["gen-c"])
 
+        if args.only_codegen:
```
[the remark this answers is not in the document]
celinval: This was just to make it easier to debug codegen... I was also thinking about using that in our tests, since today we call rustc-rmc directly.

src/rmc-compiler/cbmc/src/lib.rs
```diff
@@ -35,5 +36,7 @@ pub use machine_model::{MachineModel, RoundingMode};
 mod cbmc_string;
 pub use cbmc_string::{InternString, InternStringOption, InternedString};
 
+extern crate rustc_data_structures;
```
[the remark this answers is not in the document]
celinval: I agree. We should create a jira for that.

src/rmc-compiler/rustc_codegen_rmc/src/lib.rs
```diff
@@ -11,13 +11,30 @@
...
+extern crate rustc_fs_util;
+extern crate rustc_hir;
+extern crate rustc_index;
+extern crate rustc_llvm;
+extern crate rustc_metadata;
+extern crate rustc_middle;
+extern crate rustc_serialize;
+extern crate rustc_session;
+extern crate rustc_span;
+extern crate rustc_target;
```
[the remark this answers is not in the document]
celinval: Yes. Unfortunately it is.

src/rmc-compiler/rmc_queries/src/lib.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use std::sync::atomic::{AtomicBool, Ordering};
+
+pub trait UserInput {
+    fn set_symbol_table_passes(&mut self, passes: Vec<String>);
```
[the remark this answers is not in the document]
celinval: Yes. I don't have to add them now. But I wanted to start trying the concept sooner rather than later.

src/rmc-compiler/rustc_codegen_rmc/src/context/goto_ctx.rs
```diff
@@ -59,11 +59,11 @@ pub struct GotocCtx<'tcx> {
 
 /// Constructor
 impl<'tcx> GotocCtx<'tcx> {
-    pub fn new(tcx: TyCtxt<'tcx>) -> GotocCtx<'tcx> {
+    pub fn new(tcx: TyCtxt<'tcx>, emit_vtable_restrictions: bool) -> GotocCtx<'tcx> {
```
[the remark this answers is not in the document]
celinval: That's a good point. I could provide the query here.

src/rmc-compiler/rustc_codegen_rmc/src/compiler_interface.rs
```diff
@@ -58,7 +62,7 @@ impl CodegenBackend for GotocCodegenBackend {
         check_options(&tcx.sess, need_metadata_module);
 
         let codegen_units: &'tcx [CodegenUnit<'_>] = tcx.collect_and_partition_mono_items(()).1;
-        let mut c = GotocCtx::new(tcx);
+        let mut c = GotocCtx::new(tcx, self.queries.get_emit_vtable_restrictions());
```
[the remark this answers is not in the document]
celinval: Is this related to one of your next comments that we should pass the queries instead, right? If so, I like the suggestion.

src/rmc-compiler/rustc_codegen_rmc/src/context/goto_ctx.rs
```diff
@@ -59,11 +59,11 @@ pub struct GotocCtx<'tcx> {
...
         let fhks = fn_hooks();
         let mm = machine_model_from_session(tcx.sess);
         let symbol_table = SymbolTable::new(mm);
-        let emit_vtable_restrictions = tcx.sess.opts.debugging_opts.emit_vtable_restrictions;
+        let emit_vtable_restrictions = emit_vtable_restrictions;
```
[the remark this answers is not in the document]
celinval: I didn't dive deep here, but this variable is passed to the constructor of VtableCtx.

src/rmc-compiler/src/main.rs
celinval: This won't suppress the output of rustc. The error should've been printed already.

src/rmc-compiler/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+    let mut queries = QueryDb::default();
+    queries
+        .set_symbol_table_passes(matches.values_of_lossy("symbol-table-passes").unwrap_or(vec![]));
+    queries.set_emit_vtable_restrictions(matches.is_present("restrict-vtable-fn-ptrs"));
+
+    // Configure and run compiler.
+    let mut callbacks = RmcCallbacks {};
+    install_ice_hook();
+    let mut compiler = RunCompiler::new(&rustc_args, &mut callbacks);
+    if matches.is_present("goto-c") {
```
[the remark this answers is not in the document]
celinval: If it's not present, we will invoke `rustc` compilation without `RMC` arguments. This is required due to our current implementation of cargo rmc. Cargo calls rustc to query its version as well as to build `build.rs` scripts.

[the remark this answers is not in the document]
celinval: So we need to support both options, with and without --goto-c.

src/rmc-compiler/src/main.rs
celinval: Callbacks can be used to analyze each compilation step and it can also be used to stop compilation in earlier stages. For example, if we want to traverse the AST after parsing, we can implement a callback.

src/rmc-compiler/src/main.rs
celinval: Sure. We can fail.

src/rmc-compiler/rustc_codegen_rmc/src/context/goto_ctx.rs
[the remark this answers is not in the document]
danielsn: But I think we don't need to assign it to itself first

src/rmc-compiler/rust-toolchain.toml
celinval: The plan is to update this every two weeks for now. Instead of merging with the upstream repository, we will keep up with the latest compiler by updating the toolchain file.

I don't know long they plan to keep old toolchains, but from what I've seen, they have been keeping them all on S3 and some linux distro uses old versions of rustup. For documentation purpose, we could also track the commit that is associated with the current version.

Unfortunately, we still rely on a lot of nightly features, so we can't just use stable releases.

src/rmc-compiler/cbmc/src/lib.rs
```diff
@@ -24,6 +24,7 @@
 //!
 //! Speical [id]s include:
 //! 1. [Empty] and [Nil] behaves like [null].
+#![feature(rustc_private)]
```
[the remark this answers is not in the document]
celinval: It allow us to use rustc crates.

src/rmc-compiler/rustc_codegen_rmc/src/context/goto_ctx.rs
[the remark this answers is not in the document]
celinval: ah, I see. Yeah, this is garbage. :)

src/rmc-compiler/build.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use std::env;
+use std::path::PathBuf;
+use std::process::Command;
+
+macro_rules! path_str {
+    ($input:expr) => {
+        &$input.iter().collect::<PathBuf>().to_string_lossy().into_owned()
```
danielsn, line 10: Did we want to do a lossy conversion, or to fail fast here?

src/rmc-compiler/build.rs
```diff
@@ -0,0 +1,50 @@
...
+    };
+}
+
+/// Build the target library, and setup cargo to rerun them if the source has changed.
+fn setup_lib(out_dir: &str, lib_out: &str, lib: &str) {
+    let rmc_lib = vec!["..", "..", "library", lib];
```
danielsn, line 16: why `..` twice? Should one be `.`

danielsn (changes requested):
I think there is an issue where you meant to say `.` and write `..` for a directory path.  Not sure, but blocking merge till we look at it

src/rmc-compiler/build.rs
celinval: Because this is inside src/rmc-compiler/ so we need to walk back two directories.

src/rmc-compiler/build.rs
celinval: Sure, I can fall fast.

danielsn (approved):

src/rmc-compiler/build.rs
danielsn: Ahh, I thought this was a search path, not a set of folders

PR #736, opened 2022-01-07
Handle non returning intrinsics

### LINKED ISSUE #734: Non returning instrinsics should be handled properly
The `hooks.rs` file handles the abort intrinsic before even calling `codegen_intrinsic` in this piece of code:

```rust
    fn handle(
        &self,
        tcx: &mut GotocCtx<'tcx>,
        instance: Instance<'tcx>,
        fargs: Vec<Expr>,
        assign_to: Option<Place<'tcx>>,
        target: Option<BasicBlock>,
        span: Option<Span>,
    ) -> Stmt {
        if tcx.symbol_name(instance) == "abort" {
            Stmt::assert_false("abort intrinsic reached", loc)
        } else {
            let p = assign_to.unwrap();
            let target = target.unwrap();
            Stmt::block(
                vec![
                    tcx.codegen_intrinsic(instance, fargs, &p, span),
                    Stmt::goto(tcx.current_fn().find_label(&target), loc.clone()),
                ],
                loc,
            )
        }
    }
}
```

This special case should be removed to handle abort in `codegen_intrinsic`

adpaco-aws:
### Description of changes: 

Adds a specific function for handling of non returning intrinsics, which currently are `abort` and a special case of `transmute`.

### Resolved issues:

Resolves #734 

### Call-outs:

Wrote a function for these special intrinsics but they can also be handled within the `codegen_intrinsic` function. That would require passing `p` as an option (i.e., not unwrap before call to `codegen_intrinsic`) and match on it at the beginning of `codegen_intrinsic`. Let me know if you prefer it that way.

### Testing:

* How is this change tested? Existing regression + new test for transmute.

### Checklist


src/rmc-compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs
```diff
@@ -29,6 +29,24 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        &mut self,
+        instance: Instance<'tcx>,
+        span: Option<Span>,
+    ) -> Stmt {
+        let intrinsic = self.symbol_name(instance);
+        let intrinsic = intrinsic.as_str();
+        let loc = self.codegen_span_option(span);
+        debug!("codegen_nonret_intrinsic:\n\tinstance {:?}\n\tspan {:?}", instance, span);
+        match intrinsic {
+            "abort" => Stmt::assert_false("reached intrinsic::abort", loc),
```
celinval, line 44: I recommend using `GotocCtx::codegen_fatal_error()` here. The `Stmt::assert_false` will not abort the program execution path.

src/rmc-compiler/rustc_codegen_rmc/src/codegen/intrinsic.rs
```diff
@@ -29,6 +29,24 @@ impl<'tcx> GotocCtx<'tcx> {
         self.codegen_expr_to_place(p, e)
     }
 
+    /// Handles codegen for non returning intrinsics
+    /// Non returning intrinsics are not associated with a destination
+    pub fn codegen_nonret_intrinsic(
```
celinval, line 34: nit: I would rename it to `codegen_never_return_intrinsic()`

celinval (changes requested):
It looks good. We do need to add either an `assume(false)` or `abort()` call after the `assert(false)` though. In gotoc, an assertion failure does not interrupt the execution flow which may cause some odd behavior.

adpaco-aws:
Thanks @celinval ! Renamed and replaced `assert(false)` by `codegen_fatal_error()` in both cases, since they have similar behaviors.

celinval (approved):
Perfect! Thanks

PR #737, opened 2022-01-08
Change the results of all successful checks to "undetermined" if an unsupported construct was found to be reachable

### LINKED ISSUE #725: Incorrect success reported due to unsupported constructs
I tried this code:

```rust
#![feature(asm)]
fn foo(x: &mut u8, y: &mut u8) {
    unsafe {
        std::arch::asm!("nop");
    }
    *x = 0;
}
 fn main() {
    let mut x: u8 = rmc::nondet();
    let mut y = x;
    if x > 100 {
        foo(&mut x, &mut y);
    } else {
        x /= 2;
        y /= 2;
    }
    assert!(x == y);
}
```

```
rmc test.rs
```

with RMC version: `ab096b6ce6a`

I expected to see this happen: Since an unsupported construct (inline assembly) is reachable, I expected RMC to not report results for other assertions since their results are questionable (and in the example above is actually wrong).

Instead, this happened: RMC reported SUCCESS for `assert!(x == y)`:
```
/home/ubuntu/examples/unsupported/test.rs function foo
[foo.assertion.1] line 4 InlineAsm is not currently supported by RMC. Please post your example at https://github.com/model-checking/rmc/issues/2 : FAILURE

/home/ubuntu/examples/unsupported/test.rs function main
[main.assertion.1] line 17 assertion failed: x == y: SUCCESS

** 1 of 10 failed (2 iterations)
VERIFICATION FAILED
```

zhassan-aws:
### Description of changes: 

If the program contains reachable Rust constructs that are not supported by RMC, report all SUCCESS check results as "UNDETERMINED" since their results cannot be trusted; the modeling for unsupported constructs (`assert(false); assume(false);`) could result in missing execution paths in which some of the checks can fail.

For example, the output of running the following test:
```rust
#![feature(asm)]
fn unsupp(x: &mut u8) {
    unsafe {
        std::arch::asm!("nop");
    }
}

fn main() {
    let mut x = 0;
    unsupp(&mut x);
    assert!(x == 0);
}
```
using `rmc --output-format regular` changes from:
```
RESULT:
Check 1: main.assertion.1
         - Status: SUCCESS
         - Description: "assertion failed: x == 0"

Check 2: unsupp.assertion.1
         - Status: FAILURE
         - Description: "InlineAsm is not currently supported by RMC. Please post your example at https://github.com/model-checking/rmc/issues/2 "

Check 3: assertion.1
         - Status: SUCCESS
         - Description: "Code generation sanity check: Correct CBMC vtable size for StructTag("tag-_2692652701164773846") (MIR type Closure(DefId(1:9084 ~ std[c9dd]::rt::lang_start::{closure#0}), [(), i8, extern "rust-call" fn(()) -> i32, (fn(),)])). Please report failures:
https://github.com/model-checking/rmc/issues/new?template=bug_report.md"

Check 4: pointer_primitives.1
         - Status: SUCCESS
         - Description: "dead object in OBJECT_SIZE(&temp_0)"


SUMMARY: 
 ** 1 of 4 failed
Failed Tests: InlineAsm is not currently supported by RMC. Please post your example at https://github.com/model-checking/rmc/issues/2 
 File: "/home/ubuntu/git/rmc/src/test/expected/report/unsupported/reachable/ttt.rs", line 4, in unsupp
VERIFICATION:- FAILED
```
to:
```
RESULT:
Check 1: main.assertion.1
         - Status: UNDETERMINED
         - Description: "assertion failed: x == 0"

Check 2: unsupp.assertion.1
         - Status: FAILURE
         - Description: "InlineAsm is not currently supported by RMC. Please post your example at https://github.com/model-checking/rmc/issues/2 "

Check 3: assertion.1
         - Status: UNDETERMINED
         - Description: "Code generation sanity check: Correct CBMC vtable size for StructTag("tag-_10210028484619694427") (MIR type Closure(DefId(1:9084 ~ std[19b1]::rt::lang_start::{closure#0}), [(), i8, extern "rust-call" fn(()) -> i32, (fn(),)])). Please report failures:
https://github.com/model-checking/rmc/issues/new?template=bug_report.md"

Check 4: pointer_primitives.1
         - Status: UNDETERMINED
         - Description: "dead object in OBJECT_SIZE(&temp_0)"


SUMMARY: 
 ** 1 of 4 failed
Failed Checks: InlineAsm is not currently supported by RMC. Please post your example at https://github.com/model-checking/rmc/issues/2 
 File: "/home/ubuntu/git/rmc/src/test/expected/report/unsupported/reachable/test.rs", line 8, in unsupp
VERIFICATION:- FAILED

** WARNING: A Rust construct that is not currently supported by RMC was found to be reachable. Check the results for more details.
```

### Resolved issues:

Resolves #725 

### Call-outs:

1. It would be cleaner if the new result post-processing function uses a class that holds information on the string in the output that needs to be checked, which result (SUCCESS/FAILURE) is relevant, etc. However, since we plan to replace this script soon, I used a less structured approach.
2. The result post-processing functionality is only invoked if one of the new output formats (`--output-format` `regular|terse`) is used.

### Testing:

* How is this change tested? Added 2 tests.

### Checklist


jaisnan:
I think adding a few screenshots (before and after , if you have them) to the PR description would be visually informative.

jaisnan (approved):

scripts/cbmc_json_parser.py
```diff
@@ -165,6 +171,60 @@ def extract_solver_information(cbmc_response_json_array):
...
     return properties, solver_information
 
+def postprocess_results(properties):
+    """
+    Check for certain cases, e.g. a reachable unsupported construct or a failed
+    unwinding assertion, and update the results of impacted checks accordingly.
+    1. Change all "SUCCESS"/"FAILURE" results to "UNKNOWN" if the reachability
+    check for a Rust construct that is not currently supported by RMC failed,
+    since the missing modeling of the unsupported construct may impact results.
+    2. TODO: Change results from "SUCCESS" to "UNKNOWN" if an unwinding
```
celinval, line 181: Do you have an issue for this? If you do, can you please paste the link?

scripts/cbmc_json_parser.py
```diff
@@ -165,6 +171,60 @@ def extract_solver_information(cbmc_response_json_array):
...
+    messages = ""
+    if has_reachable_unsupported_constructs:
+        messages += "** WARNING: A Rust construct that is not currently supported " \
+                    "by RMC was found to be reachable. Check the results for " \
+                    "more details."
+
+    return properties, messages
+
+
+def check_special_cases(properties):
```
celinval, line 208: nit: I would change this to be:

```python
def has_check_failure(properties, message):
    for property in properties:
        if message in property["description"] and property["status"] == "FAILURE":
                return True
    return False
```
Then change the caller to invoke this twice. I think it will make this easier to read.

scripts/cbmc_json_parser.py
```diff
@@ -165,6 +171,60 @@ def extract_solver_information(cbmc_response_json_array):
...
+    """
+
+    has_reachable_unsupported_constructs, has_failed_unwinding_asserts = check_special_cases(properties)
+
+    for property in properties:
+        if has_reachable_unsupported_constructs:
+            # Change status of all properties except failed unsupported
+            # construct reachability checks to UNKNOWN
+            description = property["description"]
+            if GlobalMessages.UNSUPPORTED_CONSTRUCT_DESC not in description or property["status"] == "SUCCESS":
```
celinval, line 195: Can we keep the FAILED properties as FAILED?

celinval (changes requested):
My main question is regarding changing FAILURE by UNKNOWN. Can you please remind me again whether we can keep their status?

zhassan-aws:
> My main question is regarding changing FAILURE by UNKNOWN. Can you please remind me again whether we can keep their status?

Good catch! Since the case we're checking for in this PR is codegen as `assert(false); assume(false);`, any FAILURE is still valid, and should not be changed to UNKNOWN. I'll fix the PR.

scripts/cbmc_json_parser.py
zhassan-aws: I was trying to avoid iterating over all properties (which could be in the tens of thousands for larger tests) multiple times. I could refactor the loop so that it iterates over checks passed by the caller, e.g.
```python
def check_special_cases(properties, cases):

    for property in properties:
        for case in cases:
            if case.description in property["description"] and property["status"] == case.result:
            ...
```
but, I'm not sure if it's worth it given that we're likely getting rid of this script soon. Let me know your thoughts.

scripts/cbmc_json_parser.py
zhassan-aws: Yes! Done.

scripts/cbmc_json_parser.py
```diff
@@ -165,6 +171,60 @@ def extract_solver_information(cbmc_response_json_array):
...
+                property["status"] = "UNKNOWN"
```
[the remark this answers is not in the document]
zhassan-aws: Yes, Enums make sense to use here. Should we do it in a separate PR though? @jaisnan: was this something you were planning to address?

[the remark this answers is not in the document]
jaisnan: Yes @zhassan-aws , including putting the parser itself in a structure or class as well.

src/test/expected/report/unsupported/reachable/expected
```diff
@@ -0,0 +1,5 @@
+UNKNOWN
+assertion failed: x == 0
+ ** 1 of 3 failed
```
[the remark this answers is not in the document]
zhassan-aws: OK.

src/test/expected/report/unsupported/reachable/expected
```diff
@@ -0,0 +1,5 @@
+UNKNOWN
```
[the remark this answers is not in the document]
zhassan-aws: Just updated it to `UNDETERMINED`. Yes, unfortunately, I couldn't find a better way of checking given how the `expected` suite works in `compiletest`.

src/rmc-compiler/rustc_codegen_rmc/src/utils/utils.rs
```diff
@@ -53,6 +55,10 @@ impl<'tcx> GotocCtx<'tcx> {
 
         Expr::statement_expression(body, t).with_location(loc)
     }
+
+    pub fn unsupported_msg() -> &'static str {
```
celinval, line 59: Not a blocker, but I was expecting a function that gets the name of the function and generates the error message.

celinval (approved):
Just a minor comment, not a blocker.

scripts/cbmc_json_parser.py
celinval: The current code is iterating over all properties though, isn't it? But I don't think it matters. It should still be pretty fast and you can't avoid it in successful cases anyway.

Issue #701, opened 2021-12-15
cargo-rmc tracking issue


tedinski:
This is an overall tracking issue for `cargo-rmc`. The milestone of closing this issue should correspond to "cargo-rmc is good enough to make our first real release of RMC."

**Core critical path:** (The things needing tackling first, to enable other dependent work)
* [x] Replace single-file `rmc` script with scripty rust.
  * Command-line flag parsing. Invoking `rmc-rustc`. Invoking `symtab2gb`, `goto-cc`, `cbmc`.
* [x] Replace python `cargo-rmc` script with scripty rust.
* [x] Implement proof harness parsing / handling. (out of scope: good output. For first pass, just run all of them in sequence and maybe summarize at the end.)

**Followup tasks not on the critical path:** (Continuation of the above, but doesn't impact other dependent work)
* [x] Eliminate the "shell-scripty-ness" of the rust implementation. (e.g. do not shell out to `parallel`, invoke commands in parallel ourselves. Necessary to remove OS dependencies we don't want to ask users to install.)
* [x] Ensure `rmc` correctly re-builds the `rmc` prelude as part of the build. (`cargo-rmc` should be able to expect `rmc` as a `dev-dependency`)
* [x] We need expanded tests written for `cargo-rmc`

**"Nice to have"**:
* [ ] Do the build ourselves instead of invoking `cargo test --no-run`. (i.e. look into `cargo metadata` etc)
* [ ] [probably defer this until after release] `cargo rmc plan` to help with a litani integration. Perhaps a `--ci` mode that changes output defaults.

**Output improvements**:
* The initial state of the rust port will save cbmc output to a file and invoke `cbmc_json_parser.py`
* [x] Port that python to rust
* [ ] Improve output handling for multiple proof harnesses (Jai)
* [x] Introduce handling for asserts so we can report e.g. unknown/unreachable/always-fails (Zyad)
* [ ] Get "warnings" information through `rmc-metadata.json` and report potential soundness issues somehow (DSN)
* [x] Clarifying properties that mention temporary variables (Celina)

**Installation experience dependencies:**
* [x] Get `cargo-rmc` into crates.io.
* [x] Implement "download binaries on first run" (and `cargo-rmc prepare` command to explicitly trigger that)

---

**Tracking potential improvements from #738**
* [x] Refactor `InstallType` once we're sure of how we should be detecting/handling different installation types.
[...]

PR #738, opened 2022-01-08
Introduce native kani / cargo-kani

### LINKED ISSUE #755: `cargo-rmc` tests fail if we do not add `$(pwd)` to `--build-base`
`cargo-rmc` tests fail if we do not add `$(pwd)` to `--build-base`.

The issue appeared in #753 but it had not been detected before because full paths were being passed from `bootstrap`. Moreover, `--target` refers to the output directory instead of the build triple, so a better name would be `--target-dir`.

tedinski:
### Description of changes: 

This is a port of our python/bash scripts to native Rust implementations.

1. There are two entry points to the binary: `kani` and `cargo-kani`.
2. Argument parsing is in `arg.rs`
3. We set up a `KaniSession` from `session.rs` to serve as the ambient background data/state for an execution.
4. All calls out to other binaries are organized around `call_*.rs` files.

Major TODOs:

* [x] I need to double check the behavior for `cargo-kani --visualize`
* [x] I need to audit the old code to be sure I know everything that's been dropped that major.

Currently known to be missing:

* `--gen-c-runnable` (producing `--symbol-table-passes`)
* "unwind tip" (we should do this over using the new output parsing code instead of what we did previously, possibly create an issue to track this)
* Checking for dependencies in PATH (we should install them instead)

### Resolved issues:

* some progress should be entered into the major tracking issue #701
* I believe this fixes #755. I'm not sure how to interpret that issue, but I think we switch to changing the directory that `cargo` gets called in, which maybe resolves that issue?

### Testing:

* How is this change tested? existing regressions

### Checklist


src/cargo-kani/src/context.rs
```diff
@@ -0,0 +1,122 @@
...
+            bail!(
+                "Unable to determine installation location. {} doesn't look typical",
+                exe.display()
+            )
+        }
+    }
+
+    pub fn kani_rustc(&self) -> Result<PathBuf> {
+        match self {
+            Self::DevRepo(repo) => {
```
celinval, line 83: Can you create a utility function? This seemed to be repeated in the functions bellow.

src/cargo-kani/src/context.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+impl InstallType {
+    pub fn new() -> Result<Self> {
+        let mut exe = std::env::current_exe()
+            .context("cargo-kani was unable to determine where its executable was located")?;
+        // Remove the executable name, so we're in the directory we care about
+        exe.pop();
+
+        println!("{:?}", exe);
```
celinval, line 65: `tracing::debug!()`

src/cargo-kani/src/context.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub kani_lib_c: PathBuf,
+    /// The location we found 'cbmc_json_parser.py'
+    pub cbmc_json_parser_py: PathBuf,
+
+    /// The temporary files we littered that need to be cleaned up at the end of execution
+    pub temporaries: RefCell<Vec<PathBuf>>,
+}
+
+/// Represents where we detected Kani, with helper methods for using that information to find critical paths
+enum InstallType {
```
celinval, line 26: Why do we need this? Can we use a build.rs instead?

src/cargo-kani/src/context.rs
```diff
@@ -0,0 +1,122 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use crate::args::KaniArgs;
+use anyhow::{bail, Context, Result};
+use std::cell::RefCell;
+use std::path::PathBuf;
+
+/// Contains information about the execution environment and arguments that affect operations
+pub struct KaniContext {
```
celinval, line 10: nit: I prefer the name Session like rustc uses.

src/cargo-kani/src/call_single_file.rs
```diff
@@ -0,0 +1,44 @@
...
+        let mut args = self.kani_rustc_flags();
+        args.push(file.to_owned().into_os_string());
+
+        let result = Command::new(&self.kani_rustc)
+            .args(args)
+            .status()
+            .context("Failed to invoke kani-rustc")?;
+
+        if !result.success() {
+            bail!("kani-rustc exited with status {}", result);
```
celinval, line 34: This is neat.

src/cargo-kani/src/call_goto_instrument.rs
```diff
@@ -0,0 +1,66 @@
...
+        self.add_library(file)?;
+        self.undefined_functions(file)?;
+
+        Ok(())
+    }
+
+    fn add_library(&self, file: &Path) -> Result<()> {
+        let args: Vec<OsString> = vec![
+            "--add-library".into(),
+            file.to_owned().into_os_string(),
```
celinval, line 23: why do we need it twice?

celinval (commented):
I took a quick look. It's a nice start! Thanks for doing this. Here are a few suggestions:
1. I think it would be cleaner if we create two modules, one that does the external calls and one that does the parsing. You might as well split the objects too.
2. Can you also add a bit more documentation to the files and functions?

src/cargo-kani/src/args.rs
```diff
@@ -0,0 +1,196 @@
...
+                 help="Choose abstraction for modules of standard library if available")
+        add_flag(group, "--restrict-vtable", default=False, action=BooleanOptionalAction,
+                 help="Restrict the targets of virtual table function pointer calls")
+
+        */
+}
+
+#[derive(Debug, StructOpt)]
+pub struct CheckArgs {
+    // Rust argument parsers (/clap) don't have the convenient '--flag' and '--no-flag' boolean pairs, so approximate
```
celinval, line 133: Do we need them?

src/cargo-kani/src/context.rs
tedinski: (same as earlier comment: I'll fix it once we have another install type?)

tests/cargo-kani/simple-config-toml/Cargo.toml
```diff
@@ -9,13 +9,6 @@ edition = "2018"
 
 [workspace]
 
-[kani.flags.loudness]
-quiet = false
-
-[kani.flags.artifacts.extra.suffixes]
```
tedinski, line 15: Just a note for anyone reviewing: these things under `kani.flags` were pointless and I'm not sure why they were ever there. So I removed these "sections" that shouldn't be there.

Also, I removed `--no-quiet` and `--no-visualize` and `mangler` as options, so I removed those here too.

tools/compiletest/src/runtest.rs
```diff
@@ -329,8 +329,7 @@ impl<'test> TestCx<'test> {
             .args(["--function", function_name])
             .arg("--target-dir")
             .arg(self.output_base_dir().join("target"))
-            .arg("--crate")
-            .arg(&parent_dir);
+            .current_dir(&parent_dir);
```
tedinski, line 332: I removed the `--crate` flag in favor of just changing directories here.

tools/compiletest/src/runtest.rs
```diff
@@ -352,9 +351,8 @@ impl<'test> TestCx<'test> {
             kani.env("RUSTFLAGS", self.props.compile_flags.join(" "));
         }
         // Pass the test path along with Kani and CBMC flags parsed from comments at the top of the test file.
-        kani.args(&self.props.kani_flags)
-            .arg("--input")
-            .arg(&self.testpaths.file)
+        kani.arg(&self.testpaths.file)
```
tedinski, line 354: I removed the `--input` flag in favor of just passing the argument here.

src/cargo-kani/src/context.rs
tedinski: I like that, will change KaniContext to KaniSession.

src/cargo-kani/src/args.rs
celinval: I am in favor of revisiting. This seems to be adding unnecessary complexity and it doesn't seem to be a common practice in rust anyway.

src/cargo-kani/src/call_cargo.rs
```diff
@@ -0,0 +1,44 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use anyhow::{bail, Context, Result};
+use std::ffi::OsString;
+use std::path::PathBuf;
+use std::process::Command;
+
+use crate::context::KaniContext;
+
+impl KaniContext {
```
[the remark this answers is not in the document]
celinval: This is a general comment and I think we should address in a separate PR, but I would prefer breaking some of this logic down into more than one struct. I think it would be cleaner if we had smaller structs. We are using self a lot, and that could possible be broken down by calling other objects.

src/cargo-kani/src/call_goto_instrument.rs
```diff
@@ -0,0 +1,133 @@
...
+            "assert-false".into(),
+            "--generate-function-body".into(),
+            ".*".into(),
+            "--drop-unused-functions".into(),
+            file.to_owned().into_os_string(), // input
+            file.to_owned().into_os_string(), // output
+        ];
+
+        // TODO get goto-instrument path from self
+        let mut cmd = Command::new("goto-instrument");
```
zhassan-aws, line 90: Nit: to avoid the code duplication with `just_drop_unused_functions`, might be better to separate the construction of the arguments from running the command.

src/cargo-kani/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+    args.validate();
+    let ctx = session::KaniSession::new(args.common_opts)?;
+
+    let symtabs = ctx.cargo_build()?;
+    if ctx.args.only_codegen {
+        return Ok(());
+    }
+    let mut goto_objs: Vec<PathBuf> = Vec::new();
+    for symtab in &symtabs {
+        goto_objs.push(ctx.symbol_table_to_gotoc(symtab)?);
```
zhassan-aws, line 42: This is a candidate for parallelization (in a future PR).

zhassan-aws (commented):
Awesome work, Ted!

docs/src/tutorial/loops-unwinding/main.expected
```diff
@@ -1,5 +1,2 @@
 [initialize_prefix.unwind.0] line 11 unwinding assertion loop 0: FAILURE
 VERIFICATION FAILED
-[Kani] info: Verification output shows one or more unwinding failures.
-[Kani] tip: Consider increasing the unwinding value or disabling
```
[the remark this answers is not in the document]
tedinski: Awesome TODO. Yeah, I think that output parsing code is the better approach.

scripts/cargo-kani
```diff
@@ -1,359 +1,21 @@
-#!/usr/bin/env python3
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
tedinski: My current plan: it won't be installed at all, but we'll probably still use it as Kani developers. It's an opportunity to say "hey, build first" which definitely was a point of confusion at one point last summer with interns...

src/cargo-kani/src/call_cbmc.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        if self.args.output_format == crate::args::OutputFormat::Old {
+            let result = self.run_terminal(cmd);
+            if !self.args.allow_cbmc_verification_failure {
+                result?;
+            }
+        } else {
+            // extra argument
+            cmd.arg("--json-ui");
+            let result = self.run_redirect(cmd, &output_filename)?;
```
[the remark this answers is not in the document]
tedinski: Possible future improvement, yeah, but that'll take more changes to how the python code works in this PR and I was trying to keep that to an absolute minimum.

src/cargo-kani/src/call_cbmc_formatter.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
tedinski: Hmm. `call_display_results` possibly?

src/cargo-kani/src/call_cbmc_viewer.rs
```diff
@@ -0,0 +1,72 @@
...
+        let mut cmd = Command::new("cbmc-viewer");
+        cmd.args(args);
+
+        self.run_suppress(cmd)?;
+
+        Ok(())
+    }
+
+    fn cbmc_variant(&self, file: &Path, extra_args: &[&str], output: &Path) -> Result<()> {
+        let mut args = self.cbmc_flags(file)?;
```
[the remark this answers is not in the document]
tedinski: Yeah, that's a refactor that needs to happen here that I'd like to try to make a followup PR. I think there's a TODO in `call_cbmc.rs` about it. I'm planning a follow up that will fix the flags we send to cbmc for the 3 runs so that they aren't so much of a mess.

src/cargo-kani/src/main.rs
tedinski: Yep! There are a couple others too. For instance, the multiple runs of cbmc when invoking viewer.

src/cargo-kani/src/call_goto_instrument.rs
tedinski: Done.

src/cargo-kani/src/call_cbmc_formatter.rs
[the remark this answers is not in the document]
zhassan-aws: That works too, or `call_generate_results`/`call_generate_report`.

src/cargo-kani/src/call_goto_instrument.rs
zhassan-aws: Thanks.

scripts/cargo-kani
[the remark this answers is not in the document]
celinval: I think we should just remove these scripts. We can use `cargo run` if we want to ensure that the binaries are properly build.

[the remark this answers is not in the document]
tedinski: > I think we should just remove these scripts. We can use cargo run if we want to ensure that the binaries are properly build.

Disagree. People developing kani (us!) need to be able to run `cargo kani` on other projects and such. The alternative isn't `cargo run` it's to put symlinks here to `../target/debug/cargo-kani` or something that might give people confusing error messages.

BTW: `cargo run` isn't suitable for anything that gets invoked frequently. (I hit this with `kani-link-restrictions` where 30 second test suite runs turned into 20 minutes.)

celinval (commented):
Nice work Ted! This is much better than the mix of python scripts.

I think most comments can be addressed in follow up PRs. Can you just please clarify if this implementation would work with the release build. I would prefer if we at least add a high priority issue to ensure that we don't forget to fix that.

Thanks

src/cargo-kani/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+[package]
+name = "cargo-kani"
+version = "0.1.0"
+edition = "2021"
+description = "Build a project with Kani and run all proof harnesses"
+license = "MIT OR Apache-2.0"
+homepage = "https://github.com/model-checking/kani"
+repository = "https://github.com/model-checking/rmc"
```
celinval, line 11: Another one. s/rmc/kani

scripts/codegen-firecracker.sh
```diff
@@ -27,7 +27,7 @@ cd $KANI_DIR/firecracker/src/devices/src/virtio/
 export RUSTC_LOG=error
 export RUST_BACKTRACE=1
 # Kani cannot locate Cargo.toml correctly: https://github.com/model-checking/kani/issues/717
-cargo kani --only-codegen --no-config-toml
+cargo kani --only-codegen
```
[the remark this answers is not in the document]
tedinski: We do still. I changed the behavior to not fail if it can't find `./Config.toml` but there's actually a todo in the code where we should switch to using locate-project.

tests/expected/unwind_tip/expected
```diff
@@ -1,2 +0,0 @@
-[Kani] info: Verification output shows one or more unwinding failures.
```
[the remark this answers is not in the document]
tedinski: Created: https://github.com/model-checking/kani/issues/839

src/cargo-kani/src/call_cbmc.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        if !result.success() {
+            bail!("cbmc exited with status {}", result);
+        }
+        // TODO: We 'bail' above, but then ignore it in 'call_cbmc_viewer' ...
+
+        Ok(())
+    }
+
+    /// "Internal," but also used by call_cbmc_viewer
```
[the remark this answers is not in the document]
tedinski: I suppose everything could be `pub(crate)` instead of `pub` really. Is there an advantage to doing so?

src/cargo-kani/src/session.rs
```diff
@@ -0,0 +1,234 @@
...
+
+impl InstallType {
+    pub fn new() -> Result<Self> {
+        let mut exe = std::env::current_exe()
+            .context("cargo-kani was unable to determine where its executable was located")?;
+        // Remove the executable name, so we're in the directory we care about
+        exe.pop();
+
+        // Case 1: We've checked out the development repo and we're built under `target/`
+        if exe.ends_with("target/debug") {
```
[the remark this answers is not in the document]
tedinski: True. (Only for when this logic is applicable: building Kani from the repo. So this shouldn't affect the typical user who will get things from the installer.)

Do you use release builds? I'm open to adding more logic to detect release too but then there should also be something to do the right thing when there's both builds available (the newest?)

scripts/cargo-kani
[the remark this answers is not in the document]
celinval: I meant to say cost is low.

src/cargo-kani/src/call_cbmc.rs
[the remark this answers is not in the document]
celinval: It was more like a question whether you wanted to make it explicit here given the internal comment.

I'm not sure if it makes any difference when it comes to dead code analysis since this is only compiled into a binary.

src/cargo-kani/src/session.rs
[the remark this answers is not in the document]
celinval: I think release build will be important for us when we add benchmarks.

What if we use a relative path to the current executable folder? This would keep it consistent with the cargo-kani build version.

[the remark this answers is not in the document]
celinval: Sure. Can you please create an issue to fix scripts and the cargo-kani at the same time. They should all behave the same way.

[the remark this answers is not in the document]
tedinski: Sure https://github.com/model-checking/kani/issues/841

src/cargo-kani/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+    // Look where the symtabs are generated to find the `target/$TARGET/debug/deps` directory we're generating things in
+    let discovered_outdir = symtabs.get(0).unwrap().parent().unwrap().to_path_buf();
```
celinval, line 46: Can we please assert that symtab is not empty?

src/cargo-kani/src/call_cargo.rs
```diff
@@ -0,0 +1,77 @@
...
+        if self.args.debug {
+            cmd.env("KANI_LOG", "rustc_codegen_kani");
+        }
+
+        self.run_terminal(cmd)?;
+
+        if self.args.dry_run {
+            // mock an answer
+            return Ok(vec![
+                format!(
```
celinval, line 55: Can you please use PathBuf to build the string?

src/cargo-kani/src/call_cargo.rs
```diff
@@ -0,0 +1,77 @@
...
+                    "{}/{}/debug/deps/dry-run.symtab.json",
+                    target_dir.into_os_string().to_string_lossy(),
+                    build_target
+                )
+                .into(),
+            ]);
+        }
+
+        let build_glob = format!(
+            "{}/{}/debug/deps/*.symtab.json",
```
celinval, line 65: Same here

src/cargo-kani/src/call_cargo.rs
```diff
@@ -0,0 +1,77 @@
...
+            target_dir.into_os_string().to_string_lossy(),
```
celinval, line 66: I would prefer if we treat inaccurate conversion as an error.

src/cargo-kani/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    // here on almost identical to below
+    ctx.link_c_lib(&goto_objs, &linked_obj, &ctx.args.function)?;
+    if ctx.args.restrict_vtable() {
+        ctx.apply_vtable_restrictions(&linked_obj, &discovered_outdir)?;
+    }
+    ctx.run_goto_instrument(&linked_obj)?;
+
+    if ctx.args.visualize {
+        ctx.run_visualize(&linked_obj, "target/report")?;
```
celinval, line 61: Can you please use PathBuf?

src/cargo-kani/src/args.rs
```diff
@@ -0,0 +1,313 @@
...
+    #[structopt(long)]
+    pub dry_run: bool,
+    /// Generate C file equivalent to inputted program
+    #[structopt(long)]
+    pub gen_c: bool,
+
+    // TODO: currently only cargo-kani pays attention to this.
+    /// Directory for all generated artifacts
+    #[structopt(long, parse(from_os_str))]
+    pub target_dir: Option<PathBuf>,
```
celinval, line 64: Is it possible to set a default value here to avoid hard coding the default value in other places?

src/cargo-kani/src/args.rs
```diff
@@ -0,0 +1,313 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use anyhow::bail;
+use clap::{arg_enum, Error, ErrorKind};
+use std::ffi::OsString;
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+#[structopt(
```
celinval, line 11: As discussed offline, we should plan to migrate out of structopt since it is now in maintenance mode: https://docs.rs/structopt/latest/structopt.

src/cargo-kani/src/args.rs
```diff
@@ -0,0 +1,313 @@
...
+)]
+pub struct CargoKaniArgs {
+    #[structopt(flatten)]
+    pub common_opts: KaniArgs,
+}
+
+// Common arguments for invoking Kani. This gets put into KaniContext, whereas
+// anything above is "local" to "main"'s control flow.
+#[derive(Debug, StructOpt)]
+pub struct KaniArgs {
```
celinval, line 37: What is the output for `kani --help` after this change?

src/cargo-kani/src/call_single_file.rs
```diff
@@ -0,0 +1,99 @@
...
+impl KaniSession {
+    /// Used by `kani` and not `cargo-kani` to process a single Rust file into a `.symtab.json`
+    pub fn compile_single_rust_file(&self, file: &Path) -> Result<PathBuf> {
+        let output_filename = alter_extension(file, "symtab.json");
+
+        {
+            let mut temps = self.temporaries.borrow_mut();
+            temps.push(output_filename.clone());
+            temps.push(alter_extension(file, "type_map.json"));
+            temps.push(alter_extension(file, "kani-metadata.json"));
```
celinval, line 21: Can we create a little utility class to generate these extensions for us? Another alternative is to create a constant for each type of extension. I noticed that they are hard-coded in different places.

celinval (commented):
A few more comments. I'm OK leaving the comments related to polishing the code to a separate PR.

src/cargo-kani/src/args.rs
tedinski: Right now a big jumbled mess:

```
$ cargo-kani --help
cargo-kani 0.1.0
Verify a Rust crate. For more information, see https://github.com/model-checking/kani

USAGE:
    cargo-kani [FLAGS] [OPTIONS]

FLAGS:
        --allow-cbmc-verification-failure    Do not produce error return code on CBMC verification failure
        --assertion-reach-checks             Turn on assertion reachability checks
        --auto-unwind                        Turn on automatic loop unwinding
        --debug                              Produce full debug information
        --default-checks                     Turn on all default checks
        --dry-run                            Print commands instead of running them
        --gen-c                              Generate C file equivalent to inputted program
    -h, --help                               Prints help information
        --keep-temps                         Keep temporary files generated throughout Kani process
        --memory-safety-checks               Turn on default memory safety checks
        --no-default-checks                  Turn off all default checks
        --no-memory-safety-checks            Turn off default memory safety checks
        --no-overflow-checks                 Turn off default overflow checks
        --no-restrict-vtable                 Disable restricting the targets of virtual table function pointer calls
        --no-undefined-function-checks       Turn off undefined function checks
        --no-unwinding-checks                Turn off default unwinding checks
        --only-codegen                       Kani will only compile the crate
        --overflow-checks                    Turn on default overflow checks
    -q, --quiet                              Produces no output, just an exit code and requested artifacts; overrides
                                             --verbose
        --restrict-vtable                    Restrict the targets of virtual table function pointer calls
        --tests                              Enable test function verification. Only use this option when the entry
                                             point is a test function
        --undefined-function-checks          Turn on undefined function checks
        --unwinding-checks                   Turn on default unwinding checks
        --use-abs                            Use abstractions for the standard library
    -V, --version                            Prints version information
    -v, --verbose                            Output processing stages and commands, along with minor debug information
        --visualize                          Generate visualizer report to <target-dir>/report/html/index.html

OPTIONS:
        --abs-type <abs-type>              Choose abstraction for modules of standard library if available [default:
                                           std]  [possible values: std, kani, c-ffi, no-back]
        --c-lib <c-lib>...                 Link external C files referenced by Rust code
        --cbmc-args <cbmc-args>...         Pass through directly to CBMC; must be the last flag
        --function <function>              Entry point for verification [default: main]
        --object-bits <object-bits>        Specify the number of bits used for representing object IDs in CBMC [default:
                                           16]
        --output-format <output-format>    Toggle between different styles of output [default: regular]  [possible
                                           values: Regular, Terse, Old]
        --target-dir <target-dir>          Directory for all generated artifacts
        --unwind <unwind>                  Specify the value used for loop unwinding in CBMC
```

IMO, cleaning it up is another PR.

src/cargo-kani/src/call_single_file.rs
tedinski: While they're hardcoded, they each appear exactly once (except "restrictions.json" which is a design flaw I might try to correct now...)

src/cargo-kani/src/args_toml.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use anyhow::{bail, Result};
+use std::collections::BTreeMap;
+use std::ffi::OsString;
+use toml::Value;
+
+/// Produces a set of arguments to pass to "self" (cargo-kani) from a Cargo.toml project file
+pub fn config_toml_to_args() -> Result<Vec<OsString>> {
+    // TODO: `cargo locate-project` maybe?
+    let file = std::fs::read_to_string("Cargo.toml");
```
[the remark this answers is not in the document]
tedinski: Nothing is using that flag, so I dropped it.

src/cargo-kani/src/args.rs
tedinski: No: the default is different for standalone/cargo. The "hard coding in other places" is just the two different defaults.

I agree there's a design problem there, but it's not an obvious one to fix. I've deliberately avoided letting a `KaniSession` know if it's `cargo-kani` or `kani` because it's just supposed to represent the flags, but here's an area where it matters, and it gets a little annoying.

src/cargo-kani/src/call_single_file.rs
celinval: Ok. Maybe once you merge this I can add a little utility module that can be used by kani / cargo-kani and the kani-compiler.

celinval (commented):
I think I am OK signing off this PR the way it is since it is a clear improvement to what we have today. Thanks for that.

What's your plan to address the issues that we are leaving for follow up PRs though?

Some things like improving the --help output I think is high priority but I also see a lot of TODO statements that don't have any issue attached to them. There are also a lot of comments here that are going to be pushed to follow up PRs.

src/cargo-kani/src/args_toml.rs
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    args.extend(suffixed_args);
+
+    Ok(args)
+}
+
+/// Translates one toml entry into arguments and inserts it into args
+fn insert_arg_from_toml(flag: &str, value: &Value, args: &mut Vec<OsString>) -> Result<()> {
```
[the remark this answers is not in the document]
tedinski: It should be the same thing as happens on the command line so:

```
$ cargo-kani --verbose --verbose
error: The argument '--verbose' was provided more than once, but cannot be used multiple times

USAGE:
    cargo-kani --abs-type <abs-type> --function <function> --object-bits <object-bits> --output-format <output-format> --verbose

For more information try --help
```

tedinski:
I combed the PR and updated #701 with all follow up issues we've mentioned.

src/cargo-kani/src/args_toml.rs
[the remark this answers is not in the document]
celinval: The command line argument should take precedence over the cargo configuration though. Like it does today. I believe this works today because the python implementation allows you to use an argument more than once.

On clap you can set the following setting [AllArgsOverrideSelf](https://docs.rs/clap/2.33.0/clap/enum.AppSettings.html) to enable that behavior.

celinval (approved):
Can you please add to your TODO list that command line arguments should take precedence over flags on Cargo.toml?

Thanks

src/cargo-kani/src/main.rs
```diff
@@ -76,7 +82,12 @@ fn standalone_main() -> Result<()> {
     if ctx.args.visualize {
         ctx.run_visualize(&linked_obj, "report")?;
     } else {
-        ctx.run_cbmc(&linked_obj)?;
+        let result = ctx.run_cbmc(&linked_obj)?;
+        if result == VerificationStatus::Failure {
```
celinval, line 86: Nit: next time you can use `if let`

celinval (commented):
Good catch

PR #767, opened 2022-01-20
Switch --restrict-vtable to use CBMC labels instead of wrapper functions

### LINKED ISSUE #651: Use labels for function pointer restrictions if/when CBMC supports
If CBMC supports specifying function pointer restrictions using labels in the future (https://github.com/diffblue/cbmc/issues/6464), we can edit our `vtable_ctx` to not use wrapper functions.

tedinski:
### Description of changes: 

This PR no longer switches the default to on, due to #802 

It does however:

1. Switch to using labels instead of wrapper functions.
2. Work when enabled by default, which was not previously the case (various tests would fail, likely due to the wrapper functions not working perfectly correctly.)

### Resolved issues:

Fixes #651

### Call-outs:

* This removes the old "wrapping vtable calls with functions" code in favor of using cbmc labels.
* ~A bug was fixed where our scripts tried to apply function restrictions twice. (The second would fail)~ (No longer relevant to the new cargo-kani.)
* ~A file naming bug was fixed where multiple parallel runs of `kani` would choose the same filename and fail.~ (Ditto)

### Testing:

* How is this change tested? existing regressions

* Is this a refactor change? sort of, not really.

### Checklist


scripts/kani.py
```diff
@@ -431,8 +431,8 @@ def run_goto_instrument(
 
 # Processes vtable restrictions to the format CBMC expects
 def process_vtable_restrictions(restrictions_filename, verbose=False, dry_run=False):
-    cmd = ["./target/release/kani-link-restrictions"]
-    outname = os.path.join(os.path.dirname(os.path.abspath(restrictions_filename)), "linked-restrictions.json")
+    cmd = ["/home/ubuntu/rmc/target/release/kani-link-restrictions"]
```
celinval, line 434: Probably forgot to update this.

tests/kani/DynTrait/any_cast_int.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+// kani-flags: --no-restrict-vtable
+// TODO: Create a github issue about this?
```
celinval, line 5: Yes, please! :)

tests/kani/DynTrait/any_cast_int_fail.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-verify-fail
```
celinval, line 3: Is this expected?

celinval (commented):
There are a lot of things going on in this PR now. Please update the description when it's ready for review. Thanks

scripts/kani.py
tedinski: Yeah, I'm trying to just get things through regression on my local machine. This PR won't be ready for review/merge until a PR is merged into cbmc and released and fully available. It'll be a few weeks!

tests/kani/DynTrait/any_cast_int_fail.rs
tedinski: It does strike me as odd that we had to change this. I think someone will have to take some time to figure out why it was working before (_fail file but no fail directive?) and if this is definitely the correct fix.

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
```diff
@@ -436,17 +436,14 @@ impl<'tcx> GotocCtx<'tcx> {
...
         // Virtual function call and corresponding nonnull assertion.
-        let call = if self.vtable_ctx.emit_vtable_restrictions {
-            self.virtual_call_with_restricted_fn_ptr(
-                trait_fat_ptr.typ().clone(),
-                idx,
-                fn_ptr,
-                fargs.to_vec(),
-            )
+        let call = fn_ptr.dereference().call(fargs.to_vec());
+        let call_stmt = self.codegen_expr_to_place(place, call).with_location(loc.clone());
```
zhassan-aws, line 440: The changes do not seem to preserve the previous behavior: previously it was calling `self.codegen_expr_to_place` on `call` (which is the restricted function pointer if vtable restrictions is enabled), but now it is calling `self.virtual_call_with_restricted_fn_ptr` on the return value of `self.codegen_expr_to_place`. Is this intentional?

tools/kani-link-restrictions/src/main.rs
```diff
@@ -23,10 +23,7 @@ fn link_function_pointer_restrictions(data_per_crate: Vec<VtableCtxResults>) ->
     // Iterate call sites
     for crate_data in data_per_crate {
         for call_site in crate_data.call_sites {
-            // CBMC index is 1-indexed:
-            // http://cprover.diffblue.com/md__home_travis_build_diffblue_cbmc_doc_architectural_restrict-function-pointer.html
-            let cbmc_call_site_name =
-                format!("{}.function_pointer_call.1", call_site.function_name);
+            let cbmc_call_site_name = format!("{}.{}", call_site.function_name, call_site.label);
```
zhassan-aws, line 26: Can you add a comment that references CBMC's support for labeling function pointer restrictions in the CBMC documentation or PR?

tests/kani/DynTrait/any_cast_int.rs
```diff
@@ -1,6 +1,10 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+// kani-flags: --no-restrict-vtable
```
zhassan-aws, line 4: Should this changed be postponed to when function pointer restrictions are turned on by default?

tests/kani/DynTrait/any_cast_int_fail.rs
```diff
@@ -1,6 +1,10 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+// kani-flags: --no-restrict-vtable
```
zhassan-aws, line 4: Same for this.

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
tedinski: If you'd like a more definitive answer, I could ping Alexa about it. But I believe the answer is yes.

It was previously translating the function call (`foo(...)`) to a new function (calling a generated wrapper function) call expression. Now it's doing the assignment (`place = foo(...)`) but inserting a label beforehand to reference the call site. The label has to be at the statement-level, rather than before where we were transforming something at the expression level.

Notice that `virtual_call_with_restricted_fn_ptr` changed from `Expr` to `Stmt` return type.

tests/kani/DynTrait/any_cast_int.rs
tedinski: IMO no: I like that the test suite passes with it enabled and disabled by default, and we have the tracking issue to track this particular issue. While I think we should fix this issue before enabling by default, I actually don't feel super strongly about it, so if we decided to do otherwise, I'd go along with it. I don't think we'd need a test suite failure to remind us?

PR #770, opened 2022-01-21
Fix how we handle files with space in their names

### LINKED ISSUE #748: Book runner is broken
Book runner results show 0 examples pass and the rest fail. Looks like this has been happening since #722.

celinval:
### Description of changes: 

Our bash scripts was not taking into account that file names may have spaces in them. Fix how we handle that in `kani-rustc` and `kani-fmt.sh`

### Resolved issues:

Fixes #748. 

### Testing:

* How is this change tested? Added a new testcase.

### Checklist


src/test/kani/FileNameWithSpace/my src/hi.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
adpaco-aws, line 1: Nit: The path can be simpler (`src/test/kani/FileNameWithSpace/hello there.rs` or similar)

adpaco-aws (approved):

src/test/kani/FileNameWithSpace/my src/hi.rs
celinval: I tried that, but it doesn't actually work. Rust doesn't allow space in their crate name. I get the following error even while running stable `rustc`:
```sh
$ rustc "hello there.rs" 
error: invalid character ` ` in crate name: `hello there`

```

Issue #696, opened 2021-12-14
Create a `cover` statement to RMC


celinval:
**Requested feature:** Provide a functionality to RMC to allow users to prove reachability. I.e., prove that there is at least one path where the provided statement is true. 
**Use case:** Provides more insight to proof harnesses allowing users to ensure that their harnesses are not over-constrained.
**Link to relevant documentation (Rust reference, Nomicon, RFC):**
**Is this a breaking change?** No

**Test case:**

```rust

#[proof]
pub fn check_update() {
    // Write any value to the database.
    let database = rmc::nondet::<MyDatabase>();
    let key = rmc::nondet::<Key>();
    let v1 = rmc::nondet::<Value>();
    database.write(key, v1);

    // Write any other value to the database under the same key.
    let v2 = rmc::nondet::<Value>();
    database.write(key, v2);

    // Make sure v1 may be different than v2. It' OK if it's equal though.
    // Proof should fail if this is always false though.
    rmc::cover(v1 != v2);

    // Verify read gets the latest value.
    assert!(database.read(key) == v2);
}
```

This dummy harness shows how the harness runs under every single value combination, and that the last read is reachable under scenarios where v1 is different than v2. This is different than using an `assume` or another `assert` statement, since it won't restrict the search space.

Note that the verification should fail if the `cover` statement is unreachable.

celinval:
You can enable __CPROVER_cover in cbmc:

```
cbmc <target> --cover cover
```

celinval:
There are a couple issues with `__CPROVER_cover` which is being tracked here: https://github.com/diffblue/cbmc/issues/6613

PR #774, opened 2022-01-25
Hooked up __CProver_cover

zhassan-aws:
### Description of changes: 

Hook up CBMC's `CProverCover` builtin-in function. Enabling it currently requires running CBMC with the `--cover cover` option, but nevertheless it may be helpful for implementing #696.

### Resolved issues:

Partially addresses #696 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


tedinski:
`--visualize` for instance turns on `--cover location`. Can we have `--cover cover` at the same time? Do they conflict?

zhassan-aws:
> `--visualize` for instance turns on `--cover location`. Can we have `--cover cover` at the same time? Do they conflict?

Unfortunately, `--cover cover` conflicts with other options that we use:
```
$ cbmc --cover cover --unwinding-assertions cover.c 
--cover and --unwinding-assertions must not be given together
```

tedinski:
> Unfortunately, `--cover cover` conflicts with other options that we use:

Already an issue: https://github.com/diffblue/cbmc/issues/6561

celinval:
I was chatting with @zhassan-aws about this. The `--cover cover` option turns off the properties verification and it conflicts with `--cover assertions`. So that would force us to run cbmc twice. @zhassan-aws was going to chat with @tautschnig to see if it's possible to enable both and when could this be done.

Another odd thing is that if the `--cover cover` is not enabled, CBMC prints the following warning:
```
WARNING: no body for function __CPROVER_cover
```

celinval (approved):
Despite the limitations we've found so far which blocks us from taking advantage of CProver cover, this PR only adds the cover to the `cbmc` crate.

Issue #778, opened 2022-01-26
Ensure testing for `bitreverse` when support in CBMC is available


adpaco-aws:
Soon we will add partial support for the `bitreverse` intrinsic in Kani. CBMC does not support it but is being implemented in https://github.com/diffblue/cbmc/pull/6581

When support is available, fix any issues with our implementation and enable testing.

tautschnig:
https://github.com/diffblue/cbmc/pull/6638 has been issued, so a release that includes support for `bitreverse` should be out before the end of the day.

PR #779, opened 2022-01-26
Partial support for `bitreverse`

### LINKED ISSUE #745: Unimplemented intrinsic: `bitreverse`
`bitreverse` is not supported at the moment.

It appears CBMC does not have available the `bitreverse` builtin according to [the irep IDs definitions](https://github.com/diffblue/cbmc/blob/develop/src/util/irep_ids.def). I will talk to CBMC folks to see what can be done about this, but another approach here is to implement the logic for bit reversal it in RMC.

adpaco-aws:
### Description of changes: 

Adds partial support for the `bitreverse` intrinsic.

It is supposed to work until support is added in CBMC, but created #778 for tracking and doing the remaining work then.

### Resolved issues:

Resolves #745 
Opens #778 

### Testing:

* How is this change tested? Regression + new "fixme" test.

### Checklist


celinval (changes requested):
Can you please describe what partial support means here? From the user perspective, what has changed with this PR?

adpaco-aws:
> Can you please describe what partial support means here? From the user perspective, what has changed with this PR?

Right, it would be best to claim support in the work for #778 and leave the current support value as it was. What do you think?

adpaco-aws:
Dropped the change on the table and sync'd with `main`.

celinval (approved):
Thanks

PR #782, opened 2022-01-28
[Not ready for review] Inject reachability checks for assertions

### LINKED ISSUE #637: RMC should distinguish between reachable assertions that hold and unreachable ones
Requested feature: Consider the following example:
```rust
fn foo(x: u8) -> u8 {
    if x < 10 {
        17
    } else {
        5
    }
}

#[no_mangle]
fn main() {
    let x: u8 = rmc::nondet();
    let y = foo(x);
    if y > 200 {
        assert!(x != y);
    } else if y > 100 {
        assert!(x == y);
    } else {
        assert!(y != 10);
    }
}
```
RMC reports the following when run with `rmc test.rs`:
```
/home/ubuntu/tools/rmc/src/test/report/vacuity/test.rs function main
[main.assertion.1] line 14 assertion failed: x != y: SUCCESS
[main.assertion.2] line 16 assertion failed: x == y: SUCCESS
[main.assertion.3] line 18 assertion failed: y != 10: SUCCESS
...
** 0 of 66 failed (1 iterations)
VERIFICATION SUCCESSFUL
```
This output is misleading in the following sense:
1. Even though the first assert holds, it is actually unreachable
2. The second assert doesn't hold and is unreachable, but there's no indication of its unreachability in the output

The third assert is the only reachable one, so it's the only one that it makes sense to report "SUCCESS" for.

It would be useful if RMC automatically generates reachability checks for such asserts and takes their result into account when reporting results for asserts: if an assert is unreachabe, it would be reported as UNREACHABLE and not SUCCESS/FAILURE.

This can perhaps be put under an option that can be turned on/off, similar to CBMC's `--cover assertions` option (so that the user can turn them off if performance is an issue).

zhassan-aws:
### Description of changes: 

Inject a reachability check for every assertion so that unreachable checks are reported as "UNREACHABLE" as opposed to "SUCCESS". The reachability checks are currently enabled by a new option, `--assertion-reach-checks` that is default off.

### Resolved issues:

Resolves #637 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -100,3 +100,54 @@ pub fn expect_fail(_cond: bool, _message: &str) {}
 
 /// Kani proc macros must be in a separate crate
 pub use kani_macros::*;
+
+pub use std::*;
+
+#[rustc_diagnostic_item = "KaniOptionalCover"]
+#[inline(never)]
+#[doc(hidden)]
+pub fn __optional_cover() {
+    unimplemented!("Kani optional_cover")
+}
+
+#[macro_export]
+macro_rules! assert {
+    ($cond:expr $(,)?) => {
+        //core::assert!($crate::any::<bool>());
+        //$crate::__optional_cover(concat!("reachability check for assert in file ", file!(), " line ", line!()));
+        $crate::__optional_cover();
+        core::assert!($cond);
+    };
+}
+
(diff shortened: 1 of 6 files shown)


zhassan-aws:
Closing obsolete PR.

PR #784, opened 2022-01-28
Simplify how we invoke compiletest

celinval:
### Description of changes: 

I removed stuff that wasn't needed and I have also added default values for stuff that doesn't really change for Kani. Now only --suite and --mode is requied.

This will make it easier for us to manually invoke a single test suite.

### Resolved issues:

N/A

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/scripts/kani-regression.sh
+++ b/scripts/kani-regression.sh
@@ -8,16 +8,6 @@ fi
 set -o pipefail
 set -o nounset
 
-# Test for platform
-PLATFORM=$(uname -sp)
-if [[ $PLATFORM == "Linux x86_64" ]]
-then
-  TARGET="x86_64-unknown-linux-gnu"
-elif [[ $PLATFORM == "Darwin i386" ]]
-then
-  TARGET="x86_64-apple-darwin"
-fi
-
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 export PATH=$SCRIPT_DIR:$PATH
 EXTRA_X_PY_BUILD_ARGS="${EXTRA_X_PY_BUILD_ARGS:-}"
--- a/scripts/kani-regression.sh
+++ b/scripts/kani-regression.sh
@@ -55,15 +55,10 @@ for testp in "${TESTS[@]}"; do
   testl=($testp)
   suite=${testl[0]}
   mode=${testl[1]}
-  echo "Check compiletest suite=$suite mode=$mode ($TARGET -> $TARGET)"
+  echo "Check compiletest suite=$suite mode=$mode"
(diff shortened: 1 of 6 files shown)


adpaco-aws (commented):
It would be nice to have the `-f` option in `kani-regression.sh` since it will continue to be the "entry point" for testing our changes.

celinval:
I updated this PR to follow @tedinski advice. Instead of adding yet another script, I am simplifying how we run compiletest.

tools/compiletest/src/main.rs
```diff
@@ -175,17 +173,21 @@ pub fn parse_config(args: Vec<String>) -> Config {
         Some(x) => panic!("argument for --color must be auto, always, or never, but found `{}`", x),
     };
 
-    let src_base = opt_path(matches, "src-base");
+    let suite = matches.opt_str("suite").unwrap();
+    let src_base = opt_path(matches, "src-base", &["tests", suite.as_str()]);
```
celinval, line 177: @adpaco-aws I was wondering if I should try to find the fit root folder or just assume we're in the top folder. What do you think?

scripts/kani-regression.sh
```diff
@@ -55,15 +55,10 @@ for testp in "${TESTS[@]}"; do
   testl=($testp)
   suite=${testl[0]}
   mode=${testl[1]}
-  echo "Check compiletest suite=$suite mode=$mode ($TARGET -> $TARGET)"
```
adpaco-aws, line 58: I think we are not using `$TARGET` anywhere else so the platform test above can be removed.

scripts/kani-regression.sh
```diff
@@ -55,15 +45,10 @@ for testp in "${TESTS[@]}"; do
...
+  echo "Check compiletest suite=$suite mode=$mode"
   # Note: `cargo-kani` tests fail if we do not add `$(pwd)` to `--build-base`
   # Tracking issue: https://github.com/model-checking/kani/issues/755
-  cargo run -p compiletest --  --kani-dir-path scripts --src-base tests/$suite \
-                               --build-base $(pwd)/build/$TARGET/tests/$suite \
-                               --stage-id stage1-$TARGET \
-                               --suite $suite --mode $mode \
-                               --target $TARGET --host $TARGET \
-                               --quiet --channel dev
+  cargo run -p compiletest --quiet -- --suite $suite --mode $mode --quiet
```
adpaco-aws, line 51: What does the other `--quite` do here?

adpaco-aws (approved):
Thanks!

scripts/kani-regression.sh
celinval: It suppress cargo log messages. I think this makes the logs cleaner since we have already built compiletest.

Issue #786, opened 2022-02-01
Overflow checking does not do the right thing on pointers


danielsn:
The overflow operations in CBMC operate on integers, which does the wrong thing on pointers

tedinski:
`x+1` on a pointer is `x+sizeof(*x)*1`. The overflow check is treating it as `(int)x+1`

To clarify "the wrong thing"

danielsn:
https://github.com/diffblue/cbmc/issues/6627

danielsn:
```
pub fn main() {
    let a = [0; 5];
    let i : i32 = 0;
    let ptr1: *const i32 = &a[1];
    let ptr2: *const i32 = &i;
    let ptr_overflow1 = unsafe { ptr1.offset(isize::MAX) }; // This succeeds weirdly
    let ptr_overflow2 = unsafe { ptr2.offset(isize::MAX) }; //this fails
}
```

danielsn:
C reproducer

```C
#include <limits.h>
#include <stdint.h>

int main() {
    int i[5];
    int*p = &i[1];
    int*q = p + SSIZE_MAX;
}
```

danielsn:
works on offset of [1], fails for all other array offsets

danielsn:
https://github.com/diffblue/cbmc/issues/6631

nchong-at-aws:
I have another testcase:

```rust
fn main() {
    let mut xs = [0;8];
    let p = xs.as_mut_ptr();
    unsafe {
        let q = p.offset(3);
        assert!(*q == 0);
        let r = q.offset(-3);
        assert!(p == r);
    }
}
```

Currently fails with "attempt to compute offset which would overflow" (computing `r`)

PR #791, opened 2022-02-01
Don't use the broken overflow builtins for pointers

danielsn:
### Description of changes: 

Currently, we emit `__builtin_overflow_p` operations for pointer arithmetic.  This is basically broken, since CBMC converts to integers and does integer arithmetic under the hood, which is not what we want. 

### Resolved issues:

Initial step towards addressing https://github.com/model-checking/kani/issues/786

### Call-outs:

1. This relies on `--pointer-overflow-check` being used on CBMC for soundness.  We currently do so, but it would be better to have a `__pointer_overflow` builtin we could use when we need this behaviour. https://github.com/diffblue/cbmc/issues/6627

2. Also fixed some confusing missing brackets in the typechecking code for binops.

### Testing: 

* How is this change tested? Existing regression tests. 

### Checklist

### DIFF
--- a/src/kani-compiler/cbmc/src/goto_program/expr.rs
+++ b/src/kani-compiler/cbmc/src/goto_program/expr.rs
@@ -854,6 +854,12 @@ impl Expr {
             // Overflow flags
             // While one can technically use these on pointers, the result is treated as an integer.
             // This is almost never what you actually want.
+            // In particular, this check can unsoundly report no overflow on pointer operations:
+            // ```
+            //     uint32_t *p = (SIZE_MAX-4);
+            //     uint32_t *q = p+1; //overflows
+            //     OverflowPlus(p,1); //calculates SIZE_MAX-3, reports NO OVERFLOW
+            // ```
             OverflowMinus | OverflowMult | OverflowPlus => {
                 lhs.typ == rhs.typ && lhs.typ.is_integer()
             }
@@ -1205,7 +1211,9 @@ impl Expr {
         let result = self.clone().plus(e.clone());
         // FIXME: https://github.com/model-checking/kani/issues/786
         // Overflow checking on pointers is hard to do at the IREP level.
-        // Instead, we rely on `--pointer-overflow-check` in CBMC.
+        // In particular, the `__overflow` primitives check INTEGER (not pointer) arithmatic.
+        // So the `__add_overflow_p` value really only makes sense for integer arithmatic.
+        // For pointers, we rely on `--pointer-overflow-check` in CBMC.
         let overflowed = if self.typ.is_pointer() || e.typ.is_pointer() {
             warn!(
                 "Overflow operations are not properly supported on pointer types {:?} {:?}",
--- a/src/kani-compiler/cbmc/src/goto_program/expr.rs
+++ b/src/kani-compiler/cbmc/src/goto_program/expr.rs
@@ -9,6 +9,7 @@ use crate::InternedString;
 use num::bigint::BigInt;
 use std::collections::BTreeMap;
 use std::fmt::Debug;
+use tracing::warn;
 
 ///////////////////////////////////////////////////////////////////////////////////////////////
 /// Datatypes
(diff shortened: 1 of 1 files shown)


src/kani-compiler/cbmc/src/goto_program/expr.rs
```diff
@@ -1205,7 +1203,18 @@ impl Expr {
     /// `ArithmeticOverflowResult r; >>>r.overflowed = builtin_add_overflow(self, e, &r.result)<<<`
     pub fn add_overflow(self, e: Expr) -> ArithmeticOverflowResult {
         let result = self.clone().plus(e.clone());
-        let overflowed = self.add_overflow_p(e);
+        // FIXME: https://github.com/model-checking/kani/issues/786
+        // Overflow checking on pointers is hard to do at the IREP level.
+        // Instead, we rely on `--pointer-overflow-check` in CBMC.
+        let overflowed = if self.typ.is_pointer() {
```
adpaco-aws, line 1209: Don't we have to check if `e` is a pointer here?

src/kani-compiler/cbmc/src/goto_program/expr.rs
```diff
@@ -851,13 +852,10 @@ impl Expr {
...
-                (lhs.typ == rhs.typ && (lhs.typ.is_pointer() || lhs.typ.is_numeric()))
-                    || (lhs.typ.is_pointer() && rhs.typ.is_integer())
-            }
-            OverflowMult | OverflowPlus => {
-                (lhs.typ == rhs.typ && lhs.typ.is_integer())
-                    || (lhs.typ.is_pointer() && rhs.typ.is_integer())
+            // While one can technically use these on pointers, the result is treated as an integer.
+            // This is almost never what you actually want.
+            OverflowMinus | OverflowMult | OverflowPlus => {
+                lhs.typ.is_integer() && rhs.typ.is_integer()
```
adpaco-aws, line 858: Suggestion: Follow the same style as in other arms.
```suggestion
                (lhs.typ == rhs.typ && lhs.typ.is_integer())
```

src/kani-compiler/cbmc/src/goto_program/expr.rs
```diff
@@ -851,13 +852,10 @@ impl Expr {
             // Floating Point Equalities
             IeeeFloatEqual | IeeeFloatNotequal => lhs.typ == rhs.typ && lhs.typ.is_floating_point(),
             // Overflow flags
-            OverflowMinus => {
-                (lhs.typ == rhs.typ && (lhs.typ.is_pointer() || lhs.typ.is_numeric()))
```
adpaco-aws, line 855: Good to restrict this case to integers because I do not think we were handling floating point numbers anyway.

celinval:
I don't have a lot of context here, but I thought that was the goal of these checks. They were complementary to CBMC pointer checks. These were just making sure the arithmetic operations would not overflow.

One could argue that they are redundant but I don't know if broken is how I would describe them.

celinval:
Can you please add a test that would trigger the overflow to ensure the verification will still fail due to other checks?

src/kani-compiler/cbmc/src/goto_program/expr.rs
danielsn: done

src/kani-compiler/cbmc/src/goto_program/expr.rs
danielsn: Technically, that we'd notice any cases cause type-checking would fail. But making the test is clearer.

danielsn:
> Can you please add a test that would trigger the overflow to ensure the verification will still fail due to other checks?

Done

danielsn:
> I don't have a lot of context here, but I thought that was the goal of these checks. They were complementary to CBMC pointer checks. These were just making sure the arithmetic operations would not overflow.
> 
> One could argue that they are redundant but I don't know if broken is how I would describe them.

The issue is that the semantics of the arithmetic operation being done, and the semantics of the check, don't line up.  The arithmetic operation is doing *pointer* arithmetic, while the overflow check is doing *integer* arithmetic. So its possible for the actual addition to overflow, but the overflow check to report no overflow.

celinval:
> > I don't have a lot of context here, but I thought that was the goal of these checks. They were complementary to CBMC pointer checks. These were just making sure the arithmetic operations would not overflow.
> > One could argue that they are redundant but I don't know if broken is how I would describe them.
> 
> The issue is that the semantics of the arithmetic operation being done, and the semantics of the check, don't line up. The arithmetic operation is doing _pointer_ arithmetic, while the overflow check is doing _integer_ arithmetic. So its possible for the actual addition to overflow, but the overflow check to report no overflow.

Can we account for the type size?

celinval:
I believe this change is currently blocked waiting for a CBMC fix to be released. @danielsn Please let us know once this is ready for review again. Cheers!

Issue #727, opened 2022-01-05
Intrinsics Audit


adpaco-aws:
The intrinsics list below is based on the table recently published in #714.

The goal is for each intrinsics to:
1. Identify what is needed to claim full support.
2. Implement support for it if possible.

Note: We are now focusing on fixing issues with intrinsics that we already support. New intrinsics will be added on demand.

This issue will keep track of the work done on intrinsics (a check on intrinsics will mark it as "done") and link other issues related to the implementation/rework/cleanup of intrinsics.

 - [x] `abort`: Handled separately as non returning intrinsic.
 - [x] `add_with_overflow`: Reviewed and tested in #1083 
 - [x] `arith_offset`: Disabled in #1081 
 - [x] `assert_inhabited`: Changed to emit a panic or no-op depending on type.
 - [x] `assert_uninit_valid`: Changed to emit a panic or no-op depending on type.
 - [x] `assert_zero_valid`: Changed to emit a panic or no-op depending on type.
 - [x] `assume`: Added two tests in #1015 
 - [x] `atomic_and`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_and_acq`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_and_acqrel`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_and_rel`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_and_relaxed`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_cxchg`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_cxchg_acq`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_cxchg_acq_failrelaxed`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_cxchg_acqrel`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_cxchg_acqrel_failrelaxed`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_cxchg_failacq`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_cxchg_failrelaxed`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_cxchg_rel`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_cxchg_relaxed`: Reviewed in #850. Possible improvement: #25
 - [x] `atomic_cxchgweak`: Reviewed in #850. Possible improvement: #25
[...]

PR #794, opened 2022-02-03
Move `volatile_store` and ensure alignment

adpaco-aws:
### Description of changes: 

Adds an `is_aligned` function to determine if `dst` is aligned.
Moves `volatile_store` out of the `PtrWrite` hook. Now emits a warning during generation and adds a check for alignment.

### Resolved issues:

Part of #727

### Testing:

* Is this a refactor change? Almost.

### Checklist


src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
```diff
@@ -963,4 +964,30 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    fn codegen_volatile_store(
+        &mut self,
+        instance: Instance<'tcx>,
+        intrinsic: &str,
+        mut fargs: Vec<Expr>,
+        loc: Location,
+    ) -> Stmt {
+        warn!(
+            "Kani does not support concurrency for now. {} treated as a sequential operation.",
+            intrinsic
```
danielsn, line 983: Can we dump a srcloc too, to help the user figure out where this happened?

src/kani-compiler/rustc_codegen_kani/src/utils/utils.rs
```diff
@@ -48,6 +50,16 @@ impl<'tcx> GotocCtx<'tcx> {
         Expr::statement_expression(body, t).with_location(loc)
     }
 
+    /// Generates an expression `((dst as usize) % align_of(T) == 0`
+    /// to determine if `dst` is aligned.
+    pub fn is_aligned(&mut self, typ: &'tcx TyS<'_>, dst: Expr) -> Expr {
```
danielsn, line 55: @tautschnig  Will CBMC do the right thing and lower this to a bit operation? Or should we do that ourselves?

src/kani-compiler/rustc_codegen_kani/src/utils/utils.rs
```diff
@@ -48,6 +50,16 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        let layout = self.layout_of(typ);
+        let align = Expr::int_constant(layout.align.abi.bytes(), Type::size_t());
+        let cast_dst = dst.cast_to(Type::size_t());
+        let zero = Type::size_t().zero();
+        cast_dst.rem(align).eq(zero)
```
danielsn, line 60: What happens for 0 size operations? Does this divide by 0?

```
Note that even if T has size 0, the pointer must be non-null and properly aligned.
```

adpaco-aws: Note that we are using the alignment value, not the size. Alignment is always at least 1: https://doc.rust-lang.org/nomicon/repr-rust.html

src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
adpaco-aws: Yes, I will also wrap it as a util function because we are using this message in many places now.

adpaco-aws: Done!

src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
```diff
@@ -963,4 +952,27 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        intrinsic: &str,
+        mut fargs: Vec<Expr>,
+        loc: Location,
+    ) -> Stmt {
+        emit_concurrency_warning(intrinsic, loc);
+        let dst = fargs.remove(0);
+        let src = fargs.remove(0);
+        let typ = instance.substs.type_at(0);
+        let align = self.is_aligned(typ, dst.clone());
+        let align_check = Stmt::assert(align, "`dst` is properly aligned", loc.clone());
```
danielsn, line 974: Is this something we might want to reuse? Should it be a function?

src/kani-compiler/rustc_codegen_kani/src/utils/utils.rs
```diff
@@ -16,6 +18,14 @@ pub fn dynamic_fat_ptr(typ: Type, data: Expr, vtable: Expr, symbol_table: &Symbo
     Expr::struct_expr(typ, btree_string_map![("data", data), ("vtable", vtable)], symbol_table)
 }
 
+pub fn emit_concurrency_warning(intrinsic: &str, loc: Location) {
```
danielsn, line 21: Does moving this here affect our ability to filter warnings? Might be better as a macro.

src/kani-compiler/rustc_codegen_kani/src/utils/utils.rs
danielsn: 👍

src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
adpaco-aws: I think `is_aligned` strikes a good balance here. The first version was passing `instance` as an argument and generating the assert but that seemed too much work for a utils function.

src/kani-compiler/rustc_codegen_kani/src/utils/utils.rs
adpaco-aws: It does not affect filtering, that depends on the `warn` macro. I was actually trying to avoid adding macros to the intrinsics work given the various readability/debugging issues they introduce, but if there is a good reason to have it as macro instead we can make the change.

src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
```diff
@@ -1005,4 +994,27 @@ impl<'tcx> GotocCtx<'tcx> {
...
```
zhassan-aws, line 1016: Is it possible to add a test where this assert fails, to make sure it's working correctly and does not break?

src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
```diff
@@ -1005,4 +994,27 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        &mut self,
+        instance: Instance<'tcx>,
+        intrinsic: &str,
+        mut fargs: Vec<Expr>,
+        loc: Location,
+    ) -> Stmt {
+        emit_concurrency_warning(intrinsic, loc);
+        let dst = fargs.remove(0);
+        let src = fargs.remove(0);
+        let typ = instance.substs.type_at(0);
```
zhassan-aws, line 1014: Does this return the type of the pointer or that of the object it's pointing to?

src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
```diff
@@ -519,6 +519,7 @@ impl<'tcx> GotocCtx<'tcx> {
             "volatile_copy_memory" => codegen_intrinsic_copy!(Memmove),
             "volatile_copy_nonoverlapping_memory" => codegen_intrinsic_copy!(Memcpy),
             "volatile_load" => self.codegen_expr_to_place(p, fargs.remove(0).dereference()),
+            "volatile_store" => self.codegen_volatile_store(instance, intrinsic, fargs, loc),
```
zhassan-aws, line 522: When @danielsn looked at this recently, he suggested adding the following assert:
```rust
assert!(self.place_ty(p).is_unit());
```

src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
danielsn: +1

src/kani-compiler/rustc_codegen_kani/src/utils/utils.rs
adpaco-aws: That is a valid concern and I agree with it, so the new version rewrites this as a macro.

src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
adpaco-aws: Sorry, I should have included at least a negative test in the first place. I have added now two tests.

src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
adpaco-aws: Added.

src/kani-compiler/rustc_codegen_kani/src/codegen/intrinsic.rs
adpaco-aws: The latter.

zhassan-aws (approved):
LGTM

PR #797, opened 2022-02-04
Override assert macro

### LINKED ISSUE #189: Can we remove "assertion failed" from successful assertions?

zhassan-aws:
### Description of changes: 

Override the `assert` macro via calling a new `kani::assert` function that is implemented in codegen (via a hook). This gives us control over the implementation of the macro and allows us to inject any additional checks (e.g. reachability).

For assertions with a message, e.g. `assert!(a == b, "a and b have the values {} and {}, respectively", a, b)`, it produces `"a and b have the values {} and {}, respectively", a, b` instead of the current `This is a placeholder message; Kani doesn't support message formatted at runtime`.

### Call-outs:

The handling of assertions with messages results in bypassing some compile-time checks. See #803 for details.

### Testing:

* How is this change tested? Current regressions

### Checklist

### DIFF
--- a/library/std/src/lib.rs
+++ b/library/std/src/lib.rs
@@ -11,3 +11,33 @@
 
 // re-export all std symbols
 pub use std::*;
+
+/// This assert macro calls kani's assert function passing it down the condition
+/// as well as a message that will be used when reporting the assertion result.
+///
+/// For the first form that does not involve a message, the macro will generate the following message:
+/// assertion "cond"
+/// where `cond` is the stringified condition. For example, for
+/// ```rust
+/// assert!(1 + 1 == 2);
+/// ```
+/// the message will be:
+/// assertion: 1 + 1 == 2
+///
+/// For the second form that involves a custom message possibly with arguments,
+/// the macro will generate a message that is a concat of the custom message
+/// along with all the arguments. For example, for
+/// ```rust
+/// assert!(a + b == c, "The sum of {} and {} is {}", a, b, c);
+/// ```
+/// the assert message will be:
+/// assertion "The sum of {} and {} is {}", 1, 1, 2
+#[macro_export]
+macro_rules! assert {
+    ($cond:expr $(,)?) => {
+        kani::assert($cond, concat!("assertion: ", stringify!($cond)));
+    };
+    ($cond:expr, $($arg:tt)+) => {
+        kani::assert($cond, concat!("assertion: ", stringify!($($arg)+)));
+    };
+}
--- a/library/std/src/lib.rs
+++ b/library/std/src/lib.rs
@@ -38,6 +38,10 @@ macro_rules! assert {
         kani::assert($cond, concat!("assertion failed: ", stringify!($cond)));
     };
     ($cond:expr, $($arg:tt)+) => {
+        // Note that by stringifying the arguments to the custom message, any
+        // compile-time checks on those arguments (e.g. checking that the symbol
+        // is defined and that it implements the Display trait) are bypassed:
+        // https://github.com/model-checking/kani/issues/803
         kani::assert($cond, concat!(stringify!($($arg)+)));
     };
 }
--- a/library/std/src/lib.rs
+++ b/library/std/src/lib.rs
@@ -16,13 +16,13 @@ pub use std::*;
 /// as well as a message that will be used when reporting the assertion result.
 ///
 /// For the first form that does not involve a message, the macro will generate the following message:
-/// assertion "cond"
+/// assertion failed: cond
 /// where `cond` is the stringified condition. For example, for
 /// ```rust
 /// assert!(1 + 1 == 2);
 /// ```
 /// the message will be:
-/// assertion: 1 + 1 == 2
+/// assertion failed: 1 + 1 == 2
 ///
 /// For the second form that involves a custom message possibly with arguments,
 /// the macro will generate a message that is a concat of the custom message
@@ -31,13 +31,13 @@ pub use std::*;
 /// assert!(a + b == c, "The sum of {} and {} is {}", a, b, c);
 /// ```
 /// the assert message will be:
-/// assertion "The sum of {} and {} is {}", 1, 1, 2
+/// "The sum of {} and {} is {}", 1, 1, 2
 #[macro_export]
 macro_rules! assert {
     ($cond:expr $(,)?) => {
-        kani::assert($cond, concat!("assertion: ", stringify!($cond)));
+        kani::assert($cond, concat!("assertion failed: ", stringify!($cond)));
     };
     ($cond:expr, $($arg:tt)+) => {
-        kani::assert($cond, concat!("assertion: ", stringify!($($arg)+)));
+        kani::assert($cond, concat!(stringify!($($arg)+)));
     };
 }
--- a/src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
+++ b/src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
@@ -166,20 +166,19 @@ impl<'tcx> GotocHook<'tcx> for Assert {
         let msg = fargs.remove(0);
         let msg = utils::extract_const_message(&msg).unwrap();
         let target = target.unwrap();
-        let loc = tcx.codegen_span_option(span);
         let caller_loc = tcx.codegen_caller_span(&span);
 
         // Since `cond` might have side effects, assign it to a temporary
         // variable so that it's evaluated once, then assert and assume it
-        let tmp = tcx.gen_temp_variable(cond.typ().clone(), loc.clone()).to_expr();
+        let tmp = tcx.gen_temp_variable(cond.typ().clone(), caller_loc.clone()).to_expr();
         Stmt::block(
             vec![
-                Stmt::decl(tmp.clone(), Some(cond), loc.clone()),
+                Stmt::decl(tmp.clone(), Some(cond), caller_loc.clone()),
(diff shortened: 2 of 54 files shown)


library/std/src/lib.rs
```diff
@@ -11,3 +11,33 @@
...
+/// ```
+/// the assert message will be:
+/// assertion "The sum of {} and {} is {}", 1, 1, 2
+#[macro_export]
+macro_rules! assert {
+    ($cond:expr $(,)?) => {
+        kani::assert($cond, concat!("assertion: ", stringify!($cond)));
+    };
+    ($cond:expr, $($arg:tt)+) => {
+        kani::assert($cond, concat!("assertion: ", stringify!($($arg)+)));
```
celinval, line 41: I guess in the future we should ensure that any variable access is valid for message arguments. Can you please add a fixme comment and an issue capturing that?

src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
```diff
@@ -146,13 +146,51 @@ impl<'tcx> GotocHook<'tcx> for Assume {
...
+        let caller_loc = tcx.codegen_caller_span(&span);
+
+        // Since `cond` might have side effects, assign it to a temporary
+        // variable so that it's evaluated once, then assert and assume it
+        let tmp = tcx.gen_temp_variable(cond.typ().clone(), loc.clone()).to_expr();
+        Stmt::block(
+            vec![
+                Stmt::decl(tmp.clone(), Some(cond), loc.clone()),
+                Stmt::assert(tmp.clone(), &msg, caller_loc.clone()),
+                Stmt::assume(tmp, loc.clone()),
```
celinval, line 179: I'm assuming the assumption here is to abort when the assertion fails, right?

library/std/src/lib.rs
```diff
@@ -11,3 +11,33 @@
...
+/// assertion: 1 + 1 == 2
+///
+/// For the second form that involves a custom message possibly with arguments,
+/// the macro will generate a message that is a concat of the custom message
+/// along with all the arguments. For example, for
+/// ```rust
+/// assert!(a + b == c, "The sum of {} and {} is {}", a, b, c);
+/// ```
+/// the assert message will be:
+/// assertion "The sum of {} and {} is {}", 1, 1, 2
```
celinval, line 34: Why not keep it compatible with the std message?

src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
```diff
@@ -146,13 +146,51 @@ impl<'tcx> GotocHook<'tcx> for Assume {
...
+        _assign_to: Option<Place<'tcx>>,
+        target: Option<BasicBlock>,
+        span: Option<Span>,
+    ) -> Stmt {
+        assert_eq!(fargs.len(), 2);
+        let cond = fargs.remove(0).cast_to(Type::bool());
+        let msg = fargs.remove(0);
+        let msg = utils::extract_const_message(&msg).unwrap();
+        let target = target.unwrap();
+        let loc = tcx.codegen_span_option(span);
```
celinval, line 169: Have you checked how the trace looks like in viewer? I'm wondering if this will look weird since we are using two different locations.

library/std/src/lib.rs
zhassan-aws: You mean that the variable implements the `Display` trait for example?

src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
zhassan-aws: Yes, this is to maintain the current semantics of assert, which are currently roughly implemented as:
```rust
if !cond {
    assert(false);
    abort();
}
// cond must be true here

File: library/std/src/lib.rs (hunk 3 above)
Comment (zhassan-aws) [reply library/std/src/lib.rs:34]: You mean remove the `assertion: ` prefix? If so, the reason I added this prefix is to make it easy to match a reachability check with its assert. If not, can you clarify what you mean?

File: src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs (hunk 4 above)
Comment (zhassan-aws) [reply src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs:169]: The trace does look a bit weird in viewer (notice the missing files/lines):
```
Step 38: Function foo, File reach.rs, Line 4
assert!(x < 4);
var_5 = FALSE 
Step 39: Function MISSING, File MISSING, Line 0
temp_0 = FALSE 
Step 40: Function MISSING, File MISSING, Line 0
temp_0 = FALSE 
Step 41: Function foo, File reach.rs, Line 4
assert!(x < 4);
failure: foo.assertion.1: assertion: x < 4
```
I'll see if changing all the locations to the "caller" location fixes it.

File: library/std/src/lib.rs (hunk 1 above)
Comment (celinval) [reply library/std/src/lib.rs:41]: Yeah, something like that. However, I was just checking, it looks like the message side of the expression only executes if the condition fails. So I don't think it's that important.

File: src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs (hunk 2 above)
Comment (celinval) [reply src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs:179]: Can you use the original method `codegen_fatal_error()` instead?

File: library/std/src/lib.rs (hunk 3 above)
Comment (celinval) [reply library/std/src/lib.rs:34]: The std library adds "assertion failed:" prefix to things that don't have a message. So I was wondering if we should match that.

Not sure I understand how you are planning to match the check with its assert though.

File: library/std/src/lib.rs (hunk 1 above)
Comment (zhassan-aws) [reply library/std/src/lib.rs:41]: I see. In some sense, kani would be more lenient than rustc. For example, for
```rustc
fn main() {
    assert!(1 + 1 == 3, "Bad arithmetic: {}", foo);
}
```
rustc reports:
```
error[E0425]: cannot find value `foo` in this scope
 --> no_display.rs:2:47
  |
2 |     assert!(1 + 1 == 3, "Bad arithmetic: {}", foo);
  |                                               ^^^ not found in this scope
```
whereas kani doesn't complain:
```
$ kani no_display.rs 
CBMC version 5.48.0 (cbmc-5.48.0) 64-bit x86_64 linux
Reading GOTO program from file no_display.goto
Generating GOTO Program
Runtime Symex: 0.00102581s
size of program expression: 75 steps
simple slicing removed 3 assignments
Generated 3 VCC(s), 1 remaining after simplification
Runtime Postprocess Equation: 7.25e-06s
converting SSA
Runtime Convert SSA: 0.000213004s
Post-processing
Runtime Post-process: 2.594e-05s
100 variables, 33 clauses
Runtime Solver: 9.1981e-05s
Runtime decision procedure: 0.000343287s

** Results:
[assertion.1] Code generation sanity check: Correct CBMC vtable size for StructTag("tag-_4317141983261250885") (MIR type Closure(DefId(2:9147 ~ std[2f3c]::rt::lang_start::{closure#0}), [(), i8, extern "rust-call" fn(()) -> i32, (fn(),)])). Please report failures:
https://github.com/model-checking/kani/issues/new?template=bug_report.md: SUCCESS
[pointer_primitives.1] dead object in OBJECT_SIZE(&temp_0): SUCCESS
/home/ubuntu/examples/no_display.rs function main
[main.assertion.1] line 2 assertion: "Bad arithmetic: {}", foo: FAILURE

** 1 of 3 failed (2 iterations)
VERIFICATION FAILED
```
I'll file an issue and add a comment.

src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
zhassan-aws: As far as I understand, `codegen_fatal_error` is meant for a `panic` statement, which unconditionally fails, whereas here, there is a condition involved.

The alternative I was considering is to do something similar to how rustc expands the `assert` macro, e.g. 
```rust
cond.not().if_then_else(codegen_fatal_error(...), None)
```
However, based on @danielsn's feedback, with an "if", we lose the structure of the original assert.

library/std/src/lib.rs
zhassan-aws: Makes sense. The matching should be doable without the `assertion: ` prefix. I'll remove it.

src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
celinval: Cool. BTW, the `MISSING` is a different problem. We should chat with @markrtuttle about this. Viewer doesn't seem to handle files that are not inside the current folder.

src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
celinval: Gotcha. I was just hoping that we could encapsulate and formalized what an assert in the rust / kani world means, which is different from the goto-c semantics. But we can do that in a follow up PR.

Maybe we can do that when we fix codegen `TerminatorKind::Assert()`.

celinval (approved):

PR #798, opened 2022-02-04
Add an option to turn on reachability checks for assertions

zhassan-aws:
### Description of changes: 

A new option, `--assertion-reach-checks`, turns on reachability checks for assertions specified via the `assert!` macro. The results of reachability checks can aid in identifying vacuously passing properties that are reported as "SUCCESS" due to their unreachability (see #637). When this option is turned on, unreachable asserts would be reported as "UNREACHABLE".

### Call-outs:

1. The option currently turns on reachability checks for assertions specified via the `assert!` macro only. 
2. The `UNREACHABLE` result is only reported if the new output format (`--output-format regular`) is used. Without this option, the output would contain some raw unprocessed output that is not supposed to be user consumable.
3. Reachability checks are implemented via `__CPROVER_assert(false)` as opposed to the cleaner `__CPROVER_cover`. But using `__CPROVER_cover` is currently blocked by https://github.com/diffblue/cbmc/issues/6613. Because of using `__CPROVER_assert(false)`, the way to interpret the results of reachability checks is:
  - "SUCCESS": Assert is unreachable
  - "FAILURE": Assert is reachable

### Testing:

* How is this change tested? A standalone test and a cargo kani test were added 

### Checklist


src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
```diff
@@ -169,18 +170,24 @@ impl<'tcx> GotocHook<'tcx> for Assert {
         let loc = tcx.codegen_span_option(span);
         let caller_loc = tcx.codegen_caller_span(&span);
 
+        let mut stmts: Vec<Stmt> = Vec::new();
+
+        if tcx.queries.get_check_assertion_reachability() {
+            // inject a reachability (cover) check to the current location
+            let reach_msg = format!("kani_reachability_check for {}", msg);
```
celinval, line 177: Nit: Is this meant to be user visible?

celinval (changes requested):
It looks good to me, but I would prefer blocking this last PR till we have the results parser changes in place. Till then, turning this on with basically fail any verification.

zhassan-aws:
> I would prefer blocking this last PR till we have the results parser changes in place.

Sure, I'll add the parser changes to this PR.

src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
zhassan-aws: No, this is meant to make it easy for the CBMC result parser to match the reachability check with the assertion.

celinval: Ok. In that case, I would replace the `for` by either `:` or `[]`:
```
let reach_msg = format!("[KANI_REACHABILITY_CHECK] {}", msg);
```
or
```
let reach_msg = format!("KANI_REACHABILITY_CHECK: {}", msg);
```

I think it makes it more clear that this is some sort of tag / category.

zhassan-aws:
Added parser changes.

zhassan-aws:
Addressed an issue that @celinval caught, which is that `kani` was terminating with a non-zero exit code even if the user-level checks were all successful. This happened because of the failure of the `assert(false)` statements that check reachability.

scripts/cbmc_json_parser.py
```diff
@@ -469,6 +474,13 @@ def construct_property_message(properties):
 
     return output_message
 
+def count_failures(properties):
```
celinval, line 477: Don't we have this counter logic somewhere else where we print the execution summary? We should keep only one source of truth to the number of failures.

zhassan-aws: Deleted this function, and returned the number of failures from the functions that construct the summary message.

celinval (approved):
Can you please add a testcase or a line in the expected files that a reachable property is still reported as success? Maybe also add a testcase for the failure case too.

That can be in a follow up PR if you prefer.

Thanks

zhassan-aws:
> Can you please add a testcase or a line in the expected files that a reachable property is still reported as success? Maybe also add a testcase for the failure case too.
>

Done.

PR #800, opened 2022-02-05
Adding Property Class and Comment Fields to Source Location for Expect_Fail asserts

### LINKED ISSUE #793: Feature - Adding Property Classes Fields to Assertion Metadata
Requested feature: Add Property Classes to Assertion Metadata

Use case: 1) Allow Kani to dictate behavior and handling of different types of assertions depending on their property_class instead of handling all the assertions the same way. 
                  2) Users can query with unique assertion id's like - [main.property_class.1]

This requires changing how we codegen asserts. The symtab currently looks like - 

```json

"id": "side_effect",
"sub": [
{
"id": "symbol",
"namedSub": {
    "identifier": {
        "id": "__CPROVER_assert"
    },
    "#source_location": {
        "id": "",
        "namedSub": {
            "file": {
                "id": "/home/ubuntu/kani/library/kani/src/lib.rs"
            },
            "line": {
                "id": "86"
            },
            "column": {
                "id": "5"
            },
            "function": {
                "id": "kani::any_raw"
            },
        }
    },
    "type": {
        "id": "code",
        "namedSub": {
            "parameters": {
                "id": "",
                "sub": [
                    {
                        "id": "parameter",
                        "namedSub": {
                            "type": {
                                "id": "bool"
                            }
                        }
                    },
                    {
                        "id": "parameter",
                        "namedSub": {
                            "type": {
                                "id": "pointer",
                                "sub": [
                                    {
                                        "id": "signedbv",
                                        "namedSub": {
                                            "width": {
                                                "id": "8"
                                            }
                                        }
                                    }
                                ],
                                "namedSub": {
                                    "width": {
                                        "id": "64"
                                    }
                                }
                            }
                        }
                    }
                ]
            },
            "return_type": {
                "id": "empty"
            }
        }
    }
}
                
```

This is generated by a function that has a `function_call` statement - 

```rust
pub fn assert(cond: Expr, msg: &str, loc: Location) -> Self {
    BuiltinFn::CProverAssert
        .call(vec![cond, Expr::string_constant(msg)], loc.clone())
        .as_stmt(loc)
}

```

This needs to be replaced by `assert` statement instead.

jaisnan:
### Description of changes: 

Adds Property_classes and Comment Fields to the Source Location Metadata for Expect_fail asserts only (for now). With this change, the output for the rust line - 

` kani::expect_fail(i > 20, "Blocked by assumption above.");` changes from

Before the change -
`[main.assertion.1] line 7 EXPECTED FAIL: FAILURE` 

After the change -
`[main.expect_fail.1] line 7 EXPECTED FAIL: FAILURE`

It's now possible to customize the assertion ID strings as well as control descriptions for various types of assertions (which currently default to `assertion`.

### Resolved issues:

Resolves #793

### Call-outs:

Wherever needed, `Stmt::assert` should be replaced by `Stmt:assert_stmt` which takes in additional arguments for `property_class` and `comment`.

As of now, both of these fields default to "" or empty strings so they don't interfere with the current flow but that could cause potential problems and so the new symtab structure must be verified against existing symtabs.

### Testing:

* How is this change tested?

### Checklist

### DIFF
--- a/src/kani-compiler/cbmc/src/goto_program/builtin.rs
+++ b/src/kani-compiler/cbmc/src/goto_program/builtin.rs
@@ -7,6 +7,7 @@ use super::{Expr, Location, Symbol, Type};
 #[derive(Debug, Clone, Copy)]
 pub enum BuiltinFn {
     Abort,
+    Assert,
     CProverAssert,
     CProverAssume,
     CProverCover,
@@ -66,6 +67,7 @@ impl ToString for BuiltinFn {
     fn to_string(&self) -> String {
         match self {
             Abort => "abort",
+            Assert => "assert",
             CProverAssert => "__CPROVER_assert",
             CProverAssume => "__CPROVER_assume",
             CProverCover => "__CPROVER_cover",
@@ -129,6 +131,7 @@ impl BuiltinFn {
     pub fn param_types(&self) -> Vec<Type> {
         match self {
             Abort => vec![],
+            Assert => vec![Type::bool()],
             CProverAssert => vec![Type::bool(), Type::c_char().to_pointer()],
             CProverAssume => vec![Type::bool()],
             CProverCover => vec![Type::bool()],
@@ -185,6 +188,7 @@ impl BuiltinFn {
     pub fn return_type(&self) -> Type {
         match self {
             Abort => Type::empty(),
+            Assert => Type::empty(),
             CProverAssert => Type::empty(),
             CProverAssume => Type::empty(),
             CProverCover => Type::empty(),
@@ -244,6 +248,7 @@ impl BuiltinFn {
     pub fn list_all() -> Vec<BuiltinFn> {
         vec![
             Abort,
+            Assert,
             CProverAssert,
             CProverAssume,
             CProverCover,
--- a/src/kani-compiler/cbmc/src/goto_program/location.rs
+++ b/src/kani-compiler/cbmc/src/goto_program/location.rs
@@ -13,7 +13,14 @@ pub enum Location {
     BuiltinFunction { function_name: InternedString, line: Option<u64> },
     /// Location in user code.
     /// `function` is `None` for global, `Some(function_name)` for function local.
-    Loc { file: InternedString, function: Option<InternedString>, line: u64, col: Option<u64> },
+    Loc {
+        file: InternedString,
+        function: Option<InternedString>,
+        line: u64,
+        col: Option<u64>,
+        comment: InternedString,
+        property_class: InternedString,
+    },
 }
 
 /// Getters and predicates
@@ -46,15 +53,43 @@ impl Location {
         }
     }
 
+    pub fn function_name(&self) -> Option<InternedString> {
+        match self {
+            Location::Loc { function, .. } => *function,
+            _ => None,
+        }
+    }
+
+    pub fn get_column_number(&self) -> Option<u64> {
+        match self {
+            Location::Loc { col, .. } => *col,
+            _ => None,
+        }
+    }
+
+    pub fn get_comment(&self) -> Option<String> {
+        match self {
+            Location::Loc { comment, .. } => Some(comment.to_string()),
+            _ => None,
+        }
+    }
+
+    pub fn get_property_class(&self) -> Option<String> {
+        match self {
+            Location::Loc { property_class, .. } => Some(property_class.to_string()),
+            _ => None,
+        }
+    }
+
     /// Convert a location to a short string suitable for (e.g.) logging.
     /// Goal is to return just "file:line" as clearly as possible.
     pub fn short_string(&self) -> String {
         match self {
             Location::None => "<none>".to_string(),
-            Location::BuiltinFunction { function_name, line: Some(line) } => {
+            Location::BuiltinFunction { function_name, line: Some(line), .. } => {
                 format!("<{}>:{}", function_name, line)
             }
-            Location::BuiltinFunction { function_name, line: None } => {
+            Location::BuiltinFunction { function_name, line: None, .. } => {
                 format!("<{}>", function_name)
             }
             Location::Loc { file, line, .. } => format!("{}:{}", file, line),
@@ -69,6 +104,8 @@ impl Location {
         function: Option<V>,
         line: T,
         col: Option<T>,
+        comment: U,
+        property_name: U,
     ) -> Location
     where
         T: TryInto<u64>,
@@ -78,7 +115,9 @@ impl Location {
         let line = line.try_into().unwrap();
         let col = col.map(|x| x.try_into().unwrap());
         let function = function.intern();
-        Location::Loc { file, function, line, col }
+        let property_class = property_name.into();
+        let comment = comment.into();
+        Location::Loc { file, function, line, col, comment, property_class }
     }
 
     pub fn none() -> Location {
--- a/src/kani-compiler/cbmc/src/goto_program/stmt.rs
+++ b/src/kani-compiler/cbmc/src/goto_program/stmt.rs
@@ -40,6 +40,10 @@ pub enum StmtBody {
         lhs: Expr,
         rhs: Expr,
     },
+    /// `assert(cond)`
+    Assert {
+        cond: Expr,
+    },
     /// `__CPROVER_assume(cond);`
     Assume {
         cond: Expr,
@@ -174,6 +178,22 @@ impl Stmt {
             return Stmt::block(vec![assert_stmt, nondet_assign_stmt], loc);
         }
         stmt!(Assign { lhs, rhs }, loc)
+    }
+
+    /// `assert(cond);`
+    pub fn assert_stmt(cond: Expr, prop_class: &str, msg: &str, loc: Location) -> Self {
+        assert!(cond.typ().is_bool());
+
+        let file = loc.filename().unwrap();
+        let line = loc.line().unwrap();
+        let function = loc.function_name();
+        let col = loc.get_column_number().unwrap();
+        let comment = msg.to_string();
+        let property_class = prop_class.to_string();
+
+        let loc = Location::new(file, function, line, Some(col), comment, property_class);
+
+        stmt!(Assert { cond }, loc)
     }
 
     /// `__CPROVER_assert(cond, msg);`
--- a/src/kani-compiler/cbmc/src/goto_program/stmt.rs
+++ b/src/kani-compiler/cbmc/src/goto_program/stmt.rs
@@ -184,6 +184,7 @@ impl Stmt {
     pub fn assert_stmt(cond: Expr, prop_class: &str, msg: &str, loc: Location) -> Self {
         assert!(cond.typ().is_bool());
 
+        // TODO: Extract location metadata in a safer manner
         let file = loc.filename().unwrap();
         let line = loc.line().unwrap();
         let function = loc.function_name();
--- a/src/kani-compiler/cbmc/src/goto_program/symtab_transformer/transformer.rs
+++ b/src/kani-compiler/cbmc/src/goto_program/symtab_transformer/transformer.rs
@@ -496,6 +496,7 @@ pub trait Transformer: Sized {
     fn transform_stmt(&mut self, stmt: &Stmt) -> Stmt {
         match stmt.body() {
             StmtBody::Assign { lhs, rhs } => self.transform_stmt_assign(lhs, rhs),
+            StmtBody::Assert { cond } => self.transform_stmt_assert(cond),
             StmtBody::Assume { cond } => self.transform_stmt_assume(cond),
             StmtBody::AtomicBlock(block) => self.transform_stmt_atomic_block(block),
             StmtBody::Block(block) => self.transform_stmt_block(block),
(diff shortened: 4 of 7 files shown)


src/kani-compiler/cbmc/src/goto_program/symtab_transformer/transformer.rs
```diff
@@ -529,6 +530,14 @@ pub trait Transformer: Sized {
...
     }
 
+    /// Transforms a CPROVER assume stmt (`__CPROVER_assert(cond);`)
+    fn transform_stmt_assert(&mut self, cond: &Expr) -> Stmt {
+        let transformed_cond = self.transform_expr(cond);
+        let msg = "test_message";
+        let property_class = "test_property_class";
+        Stmt::assert_stmt(transformed_cond, property_class, msg, Location::none())
+    }
+
```
jaisnan, line 540: I'm not completely clear on the transformations and how they operate and need someone to review if this is safe to do. While it doesn't present any immediate problems, I also don't see this being the best way to write a transformer.

src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
```diff
@@ -98,7 +98,13 @@ impl<'tcx> GotocHook<'tcx> for ExpectFail {
         let target = target.unwrap();
         let cond = fargs.remove(0).cast_to(Type::bool());
         //TODO: actually use the error message passed by the user.
+
+        // msg is the comment field that's set on source location
         let msg = "EXPECTED FAIL";
+
+        // property_class is used as a unique identifier for this assert
+        let _property_class = "expect_fail";
+
```
jaisnan, line 107: property_class is unused for now, but when we use `assert_stmt` instead of `assert`, it needs to be passed as the second variable.

src/kani-compiler/cbmc/src/goto_program/builtin.rs
```diff
@@ -7,6 +7,7 @@ use super::{Expr, Location, Symbol, Type};
 #[derive(Debug, Clone, Copy)]
 pub enum BuiltinFn {
     Abort,
+    Assert,
```
zhassan-aws, line 10: All the changes to this file are no longer needed, right? The new `assert` generates a statement, rather than a built-in function call.

src/kani-compiler/cbmc/src/goto_program/location.rs
```diff
@@ -13,7 +13,14 @@ pub enum Location {
     BuiltinFunction { function_name: InternedString, line: Option<u64> },
     /// Location in user code.
     /// `function` is `None` for global, `Some(function_name)` for function local.
-    Loc { file: InternedString, function: Option<InternedString>, line: u64, col: Option<u64> },
+    Loc {
```
danielsn, line 16: Other than asserts, do we ever want to have `comment` or `property_classes` on statements/expressions. What do we do if there are statements with no property classes?

src/kani-compiler/cbmc/src/goto_program/builtin.rs
tautschnig: +1, and shouldn't the opposite be happening in that `CProverAssert` (and `CProverAssume`) should be removed?

src/kani-compiler/cbmc/src/goto_program/location.rs
```diff
@@ -13,7 +13,14 @@ pub enum Location {
...
+        file: InternedString,
+        function: Option<InternedString>,
+        line: u64,
+        col: Option<u64>,
+        comment: InternedString,
+        property_class: InternedString,
```
danielsn, line 22: Just because CBMC 

1. Puts this on the location and 
2. Uses a string, doesn't mean we need to do either. Make this a propert of the `assert`, and use an `enum` for the different property classes. Then, make the `IRep` generator translate that into the anotations.

src/kani-compiler/cbmc/src/goto_program/symtab_transformer/transformer.rs
```diff
@@ -529,6 +530,14 @@ pub trait Transformer: Sized {
         transformed_lhs.assign(transformed_rhs, Location::none())
     }
 
+    /// Transforms a CPROVER assume stmt (`__CPROVER_assert(cond);`)
+    fn transform_stmt_assert(&mut self, cond: &Expr) -> Stmt {
+        let transformed_cond = self.transform_expr(cond);
+        let msg = "test_message";
+        let property_class = "test_property_class";
```
tautschnig, line 537: I'm lacking context: is this test code?

src/kani-compiler/cbmc/src/goto_program/location.rs
```diff
@@ -46,15 +53,43 @@ impl Location {
         }
     }
 
+    pub fn function_name(&self) -> Option<InternedString> {
+        match self {
+            Location::Loc { function, .. } => *function,
```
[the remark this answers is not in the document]
danielsn: Don't builtin functions also have names?

src/kani-compiler/cbmc/src/irep/to_irep.rs
```diff
@@ -321,12 +321,19 @@ impl ToIrep for Location {
...
+                    (IrepId::File, Irep::just_string_id(file.to_string())),
+                    (IrepId::Line, Irep::just_int_id(*line)),
+                ])
+                .with_named_sub_option(IrepId::Column, col.map(Irep::just_int_id))
+                .with_named_sub_option(IrepId::Function, function.map(Irep::just_string_id))
+                .with_named_sub(IrepId::Comment, Irep::just_string_id(comment.to_string()))
+                .with_named_sub(
+                    IrepId::PropertyClass,
+                    Irep::just_string_id(property_class.to_string()),
+                )
```
tautschnig, line 335: Why are these note `with_named_sub_option`?

src/kani-compiler/cbmc/src/goto_program/location.rs
```diff
@@ -46,15 +53,43 @@ impl Location {
...
+    }
+
+    pub fn get_column_number(&self) -> Option<u64> {
+        match self {
+            Location::Loc { col, .. } => *col,
+            _ => None,
+        }
+    }
+
+    pub fn get_comment(&self) -> Option<String> {
```
danielsn, line 70: Why `String` and not `InternedString`

danielsn, line 70: CBMC tracks comments on source locations. Do we want to?

src/kani-compiler/cbmc/src/goto_program/location.rs
```diff
@@ -46,15 +53,43 @@ impl Location {
...
+        match self {
+            Location::Loc { comment, .. } => Some(comment.to_string()),
+            _ => None,
+        }
+    }
+
+    pub fn get_property_class(&self) -> Option<String> {
```
danielsn, line 77: Why `String`?

src/kani-compiler/cbmc/src/goto_program/location.rs
```diff
@@ -46,15 +53,43 @@ impl Location {
...
+        }
+    }
+
     /// Convert a location to a short string suitable for (e.g.) logging.
     /// Goal is to return just "file:line" as clearly as possible.
     pub fn short_string(&self) -> String {
         match self {
             Location::None => "<none>".to_string(),
-            Location::BuiltinFunction { function_name, line: Some(line) } => {
+            Location::BuiltinFunction { function_name, line: Some(line), .. } => {
```
danielsn, line 89: Why don't we handle the non case here?

src/kani-compiler/cbmc/src/goto_program/stmt.rs
```diff
@@ -176,6 +180,23 @@ impl Stmt {
...
+
+        // TODO: Extract location metadata in a safer manner
+        let file = loc.filename().unwrap();
+        let line = loc.line().unwrap();
+        let function = loc.function_name();
+        let col = loc.get_column_number().unwrap();
+        let comment = msg.to_string();
+        let property_class = prop_class.to_string();
+
+        let loc = Location::new(file, function, line, Some(col), comment, property_class);
```
danielsn, line 195: Why are you destructuring this just to build it up again?

src/kani-compiler/cbmc/src/goto_program/symtab_transformer/transformer.rs
```diff
@@ -529,6 +530,14 @@ pub trait Transformer: Sized {
         transformed_lhs.assign(transformed_rhs, Location::none())
     }
 
+    /// Transforms a CPROVER assume stmt (`__CPROVER_assert(cond);`)
+    fn transform_stmt_assert(&mut self, cond: &Expr) -> Stmt {
+        let transformed_cond = self.transform_expr(cond);
+        let msg = "test_message";
```
danielsn, line 536: Why not use the actual message here?

danielsn (changes requested):

src/kani-compiler/cbmc/src/goto_program/location.rs
zhassan-aws, line 89: Unintended change?

zhassan-aws (changes requested):

src/kani-compiler/cbmc/src/goto_program/location.rs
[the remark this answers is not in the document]
jaisnan: So if I'm understanding it correctly, I should add a new variant for Location called Stmt::Assert instead of using Loc (local functions) to store information about property_class and comment?

src/kani-compiler/cbmc/src/goto_program/location.rs
jaisnan: Addressing both comments here - I think comment is a field that's common to statements, expressions as well but property_classes is not something that can be added to other things apart from `assert`. I'll change the flow to add these as properties to `assert` only. I was unclear about this earlier.

src/kani-compiler/cbmc/src/goto_program/location.rs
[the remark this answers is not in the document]
zhassan-aws: No, what I meant is to add new a property class field to the new `StmtBody::Assert` variant (in `cbmc/src/goto_program/stmt.rs`).

src/kani-compiler/cbmc/src/goto_program/location.rs
jaisnan: We might only want it on asserts for now, so I'll change the assert properties instead.

src/kani-compiler/cbmc/src/goto_program/stmt.rs
jaisnan: So, we need to build a new Location struct object but change just the property_class field. (We are basically setting the field). 

Between adding a setter function and adding the property_class and creating a new object, this seemed to be the safer option as setting even one variable involved making the entire Location struct as mutable. This didn't seem safe to me.

src/kani-compiler/rustc_codegen_kani/src/codegen/span.rs
```diff
@@ -19,11 +19,17 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        // Default to empty strings for both Comment and Property_class
+        let comment = "".to_string();
+        let property_class = "".to_string();
         Location::new(
             filename1,
             self.current_fn.as_ref().map(|x| x.readable_name().to_string()),
             line,
             Some(col),
+            comment,
+            property_class,
```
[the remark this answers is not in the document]
jaisnan: Yes, that needs to be refleced

src/kani-compiler/cbmc/src/irep/to_irep.rs
jaisnan: Comment and Property class weren't set as Options, i'll change that now.

src/kani-compiler/cbmc/src/goto_program/location.rs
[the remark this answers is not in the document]
jaisnan: That seems a much better solution, will make those changes.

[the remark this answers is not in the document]
celinval: Isn't this crate trying to match what CBMC is doing though? Also, the `propery_class` might be relevant at least for cover statements, right?

src/kani-compiler/cbmc/src/goto_program/location.rs
celinval: @danielsn I though this package was trying to match CBMC's internal structure though.

That said, these new fields should be optional.

src/kani-compiler/cbmc/src/goto_program/stmt.rs
celinval: I would recommend to remove the getter methods and to use an `if let` expression to extract the fields you need. Something like:

```rust
let new_loc = if let Location::Loc{file, line, function, col, .. } = loc {
        Location::Loc{file, line, function, col, comment: msg.to_string(), property_class: prop_class.to_string());
    } else {
        loc
    }
```

src/kani-compiler/cbmc/src/goto_program/symtab_transformer/transformer.rs
celinval: Can you explain what you are trying to accomplish?

src/kani-compiler/rustc_codegen_kani/src/codegen/span.rs
```diff
@@ -19,11 +19,17 @@ impl<'tcx> GotocCtx<'tcx> {
             Ok(pathbuf) => pathbuf.to_str().unwrap().to_string(),
             Err(_) => filename0,
         };
+
```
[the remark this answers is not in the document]
celinval: You should make these fields optional. If they are not set, you don't add them to the irep.

src/kani-compiler/cbmc/src/goto_program/location.rs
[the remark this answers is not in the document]
jaisnan: It is trying to match CBMC's structure but maybe we only need to map it at the end when converting to Irep's

jaisnan:
Closing this PR to start afresh

PR #804, opened 2022-02-07
Fast math

### LINKED ISSUE #788: Unimplemented intrinsics: Fast math for floating point numbers
Fast math intrinsics such as `[fadd_fast](https://doc.rust-lang.org/std/intrinsics/fn.fadd_fast.html)` are not implemented.

adpaco-aws:
### Description of changes: 

Models fast math intrinsics as regular floating point operations. Includes function to check inputs are finite arguments. Adds 10 new test cases (5 positive, 5 negative) for testing fast math intrinsics.

### Resolved issues:

Resolves #788 

### Testing:

* How is this change tested? Existing regression + 10 new tests.

### Checklist


src/kani-compiler/cbmc/src/goto_program/stmt.rs
```diff
@@ -129,6 +129,13 @@ impl Stmt {
             _ => None,
         }
     }
+
+    pub fn get_stmts(&self) -> Option<&Vec<Stmt>> {
```
adpaco-aws, line 133: This is part of an earlier version where the intrinsics where encoded in a block statement, so I needed to retrieve statements within in order to prepend the finite checks. This version does not use it, but I figured it could be useful for something else. We can remove it otherwise.

tests/kani/Intrinsics/FastMath/add.rs
```diff
@@ -0,0 +1,19 @@
...
+    let y: f64 = kani::any();
+
+    kani::assume(x.is_finite());
+    kani::assume(y.is_finite());
+    match (x.is_sign_positive(), y.is_sign_positive()) {
+        (true, true) => kani::assume(x < f64::MAX - y),
+        (false, false) => kani::assume(x > f64::MIN - y),
+        _ => (),
+    }
+    let _z = unsafe { std::intrinsics::fadd_fast(x, y) };
```
celinval, line 18: Can you please assert that the result is equal to regular add? The same for the other tests.

src/kani-compiler/cbmc/src/goto_program/stmt.rs
celinval: If you do keep it, can you please add a comment?

celinval (commented):
Nice test coverage! Just minor comments.

src/kani-compiler/cbmc/src/goto_program/stmt.rs
adpaco-aws: Added a comment.

tests/kani/Intrinsics/FastMath/add.rs
adpaco-aws: Changed it so we assert it in the case of addition and substraction, as other operations show serious performance issues. Created #809 to revisit this.

adpaco-aws:
Some changes I made:
 1. Added tests for `f32` values so we now cover all floating point types.
 2. Changed tests to assert equality (w.r.t. regular operations) in the case of addition and substraction, but other operations show serious performance issues. Created #809 to revisit this topic later.
 3. Dropped support for `frem_fast` because testing showed unexpected results. Created #810 to explain the issue and tackle it in the future.

celinval (approved):

PR #828, opened 2022-02-15
Change Default UI to New Display and Add UI Test Suite

jaisnan:
### Description of changes: 

1. Change Default Value to the new display.
2. Add test suite for the output formats.
3. Change `expected` test suite to use old format as a temporary fix.
4. Fix issue that `dry-run` didn't run well with `expected` 

This is what calling `kani` on a file looks like by default - 

Screen Shot 2022-02-15 at 1 13 29 AM

### Call-outs:

To get the old output display, pass the flag `--output-format old`.

The new output display has an issue where `--visualize` doesn't act well with the `--output-format` flag. A fix is in the works.

### Testing:

* How is this change tested? Added a new test suite

### Checklist


celinval (approved):
Sweet! Thanks

jaisnan:
> Yes, that change makes sense. That is coming up next.

tedinski:
> Does it make sense to add --output-format old in compiletest itself

This might make it difficult to write tests against other output formats. We'd have to see if multiple `--output-format` would mean last one wins on my new code, or if it'd just error.

PR #834, opened 2022-02-17
Add checks to ensure that we are safely using transmute for transparent types

### LINKED ISSUE #822: Error compiling new std version (2021-02-09)
While trying to update our toolchain to a newer version, I bumped into an issue compiling the new version of the std. I reduced the testcase which I am adding to the fixme suite.

To reproduce after my test PR gets merged, go the folder `tests/kani/Repr` and run:

```
RUSTFLAGS="--crate-type lib" kani check_repr_fixme.rs --function find_error
```

Kani is crashing with the following error:

```
thread 'rustc' panicked at 'Can't cast

Expr { value: BinOp { op: Plus, lhs: Expr { value: Typecast(Expr { value: AddressOf(Expr { value: Symbol { identifier: "check_repr_fixme.5aa6d883::check_repr_fixme::alloc10" }, typ: StructTag("tag-check_repr_fixme.5aa6d883::check_repr_fixme::alloc10::struct"), location: None }), typ: Pointer { typ: StructTag("tag-check_repr_fixme.5aa6d883::check_repr_fixme::alloc10::struct") }, location: None }), typ: Pointer { typ: Unsignedbv { width: 8 } }, location: None }, rhs: Expr { value: IntConstant(0), typ: Unsignedbv { width: 64 }, location: None } }, typ: Pointer { typ: Unsignedbv { width: 8 } }, location: None }

StructTag("tag-_14501000783433242740")', src/kani-compiler/cbmc/src/goto_program/expr.rs:433:9

```

danielsn:
### Description of changes: 

Rust sometimes directly assigns a scalar to a structurally identical transparent struct.
We currently handle this through the highly unsafe hammer of transmute.
Instead, check that the type really is structurally identical, and then generate a proper rvalue. 
### Resolved issues:

Resolves #822 #837 

### Call-outs:

I wasn't 100% sure if we wanted this in the cprover or kani crate. Put it in cprover as it may be useful to others.

### Testing:

* How is this change tested?
  * Existing unit tests including the one for #822 
  * A test for #837 

### Checklist


cprover_bindings/src/goto_program/expr.rs
```diff
@@ -1315,6 +1315,57 @@ impl Expr {
...
+                if let DatatypeComponent::Field { typ, .. } = &components[0] {
+                    Expr::struct_expr_from_values(
+                        wrapper_typ.aggr_tag().unwrap(),
+                        vec![self.wrap_in_transparent_type(typ, st)],
+                        st,
+                    )
+                } else {
+                    unreachable!()
+                }
+            }
```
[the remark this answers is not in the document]
celinval: Can you use a loop here instead of recursion?

src/kani-compiler/rustc_codegen_kani/src/codegen/operand.rs
```diff
@@ -389,14 +389,21 @@ impl<'tcx> GotocCtx<'tcx> {
...
-            // The `cast_to` operation causes https://github.com/model-checking/kani/issues/822
-            // FIXME: Changing to a transmute to temporarily unblock, proper fix is to
-            // unwrap the transparent struct.
-            .transmute_to(res_t, &self.symbol_table)
-        //.cast_to(res_t)
+            .plus(Expr::int_constant(offset.bytes(), Type::unsigned_int(64)));
+
+        // In some cases, Rust uses a transparent type here. Convert the pointer to an rvalue
+        // of the type expected. https://github.com/model-checking/kani/issues/822
+        if let Some(wrapped_type) = res_t.unwrap_transparent_type(&self.symbol_table) {
```
celinval, line 399: Should we use the `ReprOptions::transparent()` here?

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -570,6 +570,87 @@ impl Type {
...
+            | Unsignedbv { .. } => Some(self.clone()),
+
+            // Follow tags to get the underlying structure
+            StructTag(tag) | UnionTag(tag) => {
+                st.lookup(*tag).unwrap().typ.unwrap_transparent_type(st)
+            }
+
+            // Recursively check: does this only have one field, which is either a wrapper or base type.
+            Struct { components, .. } | Union { components, .. } => {
+                if components.len() != 1 {
```
celinval, line 642: Can you please check if that's the case for transparent types that have extra zero-sized fields as described here? https://doc.rust-lang.org/1.26.2/unstable-book/language-features/repr-transparent.html#enter-reprtransparent

celinval (changes requested):
I think the recursion in this code is a bit hard to follow. It might be worth trying to simplify it a bit.

src/kani-compiler/rustc_codegen_kani/src/codegen/operand.rs
danielsn: At this point in the codegen, we only have access to the goto type, not the MIR.  Longer term, it would be nice to be able to map back from the cbmc type to the original MIR type and ask these kinds of questions.

cprover_bindings/src/goto_program/typ.rs
danielsn: Added a TODO and a tracking ticket

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -539,6 +568,13 @@ impl Type {
         }
     }
 
+    pub fn is_union_like(&self) -> bool {
+        match self {
```
celinval, line 572: nit: This could be converted into:

```rust
matches!(self, IncompleteUnion { .. } | Union { .. } | UnionTag(_))
```

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -578,76 +614,30 @@ impl Type {
...
-            | IncompleteStruct { .. }
-            | IncompleteUnion { .. }
-            | InfiniteArray { .. }
-            | Pointer { .. }
-            | Signedbv { .. }
-            | Unsignedbv { .. }
-            | VariadicCode { .. }
-            | Vector { .. } => false,
-        }
+        (self.is_struct_like() || self.is_union_like())
```
celinval, line 617: Why isn't `self.unwrap_transparent_type(st).is_some()` enough?

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -621,11 +621,11 @@ impl Type {
     /// Given a transparent type (see comment on `Type::is_transparent_type()`),
     /// extract the type it wraps.
     pub fn unwrap_transparent_type(&self, st: &SymbolTable) -> Option<Type> {
-        // If the type has components, i.e. is either a union or a
+        // If the type has components, i.e. is either a union or a struct, recurse into them
         if self.is_struct_like() || self.is_union_like() {
             let components = st.lookup_components_in_type(self).unwrap();
             // For now, only handle the case of transparent structs with one field.
-            // TODO ticket for this
+            // TODO https://github.com/model-checking/kani/issues/837
```
celinval, line 628: Can we add a fixme test for this issue? I think it would help to know how this would fail. It can be a follow up PR.

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -570,6 +606,41 @@ impl Type {
...
+    ///     struct TransparentWrapper { int x };
+    /// ```
+    /// We define it recursively as a type which has exactly one field, which it itself either
+    /// a transparent type, or is a scalar type.
+    pub fn is_transparent_type(&self, st: &SymbolTable) -> bool {
+        (self.is_struct_like() || self.is_union_like())
+            && self.unwrap_transparent_type(st).is_some()
+    }
+
+    /// Given a transparent type (see comment on `Type::is_transparent_type()`),
```
celinval, line 621: nit: Can you please update this to something like: "If type is transparent, extract the type it wraps"? The current documentation makes it sound like there is a pre-condition that `self` must be transparent.

celinval (commented):
Mostly small comments. As we discussed offline, I was wondering if we should add the extra checks that you implemented, but still translate this to transmute instead of generating the unwrap + wrap expressions.

cprover_bindings/src/goto_program/typ.rs
danielsn: That would match scalar types too.

cprover_bindings/src/goto_program/typ.rs
danielsn: My concern with that pattern is that the compiler wouldn't warn if we add a new variant to the `Type` enum

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -570,6 +620,121 @@ impl Type {
...
+    /// But, `struct foo` is not structurally equivilent to:
+    /// ```
+    /// __attribute__((packed))
+    /// struct baz {}
+    ///     char x;
+    ///     int y;
+    /// }
+    /// ```
+    /// Since they have different padding.
+    pub fn is_structurally_equivalent_to(&self, other: &Type, st: &SymbolTable) -> bool {
```
tedinski, line 698: IMO, break the branches of this function into 2 helper functions: scalar,other and other,other. Eliminates code duplication, shortens the functions.

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -570,6 +620,121 @@ impl Type {
...
+        st: &'a SymbolTable,
+    ) -> Option<Vec<&'a DatatypeComponent>> {
+        if let Some(components) = st.lookup_components_in_type(self) {
+            Some(components.iter().filter(|x| x.sizeof_in_bits(st) != 0).collect())
+        } else {
+            None
+        }
+    }
+
+    /// Two types are structurally equivilent if one can be cast into the other without changing
```
danielsn, line 670: https://github.com/diffblue/cbmc/blob/develop/src/solvers/lowering/byte_operators.cpp#L1093..L1136

celinval (approved):
Thanks

cprover_bindings/src/goto_program/typ.rs
danielsn: done

PR #838, opened 2022-02-17
Add reachability checks for some auto-generated asserts

zhassan-aws:
### Description of changes: 

Generate reachability checks (under `--assertion-reach-checks`) for asserts generated through `TerminatorKind::Assert` which includes out-of-bound checks, overflow checks, overflow negation checks, and division/remainder by zero checks.

### Testing:

* How is this change tested? Added unreachable and reachable tests that cover the above checks.

### Checklist


src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
```diff
@@ -78,18 +79,24 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    cond.cast_to(Type::bool()).if_then_else(
                         Stmt::goto(self.current_fn().find_label(target), loc.clone()),
-                    ],
-                    loc,
-                )
+                        None,
+                        loc.clone(),
+                    ),
+                    Stmt::assert_false(&msg_str, loc.clone()),
+                    Stmt::goto(self.current_fn().find_label(target), loc.clone()),
```
danielsn, line 97: we no longer need to clone locations

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
```diff
@@ -78,18 +79,24 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                            None,
-                            loc.clone(),
-                        ),
-                        Stmt::assert_false(&format!("{:?}", msg), loc.clone()),
+                let mut msg_str = format!("{:?}", msg);
+                let mut stmts: Vec<Stmt> = Vec::new();
+                if self.queries.get_check_assertion_reachability() {
+                    let check_id = self.next_check_id();
+                    msg_str = format!("[{}] {}", check_id, msg_str);
+                    let reach_msg = GotocCtx::reach_check_msg(&check_id);
```
danielsn, line 87: Why is one of these behind a function and the other not?

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
```diff
@@ -78,18 +79,24 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                Stmt::block(
-                    vec![
-                        cond.cast_to(Type::bool()).if_then_else(
-                            Stmt::goto(self.current_fn().find_label(target), loc.clone()),
-                            None,
-                            loc.clone(),
-                        ),
-                        Stmt::assert_false(&format!("{:?}", msg), loc.clone()),
+                let mut msg_str = format!("{:?}", msg);
+                let mut stmts: Vec<Stmt> = Vec::new();
```
danielsn, line 83: Do these need to be mutable?

src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
```diff
@@ -178,7 +178,7 @@ impl<'tcx> GotocHook<'tcx> for Assert {
             // [KANI_REACHABILITY_CHECK] <check ID>
             // for reachability checks
             msg = format!("[{}] {}", assert_id, msg);
-            let reach_msg = format!("[KANI_REACHABILITY_CHECK] {}", assert_id);
+            let reach_msg = GotocCtx::reach_check_msg(&assert_id);
```
danielsn, line 181: Long term do we want to keep this in the hooks? Or is this solved by writing our own assertion macro that we inject?

src/kani-compiler/rustc_codegen_kani/src/utils/names.rs
```diff
@@ -109,6 +109,10 @@ impl<'tcx> GotocCtx<'tcx> {
         // TODO: use def_id https://github.com/model-checking/kani/issues/364
         idx.to_string().into()
     }
+
+    pub fn reach_check_msg(check_id: &str) -> String {
```
danielsn, line 113: What does the `check_id` represent?

danielsn, line 113: Also, this function name is confusing.

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
zhassan-aws: Making them mutable was the only way I could think of to avoid code duplication. I'm open for suggestions.

src/kani-compiler/rustc_codegen_kani/src/overrides/hooks.rs
zhassan-aws: I'm thinking we should keep this in the hooks for the long term. Our own assertion macro calls this hook here: https://github.com/model-checking/kani/blob/541a3100d88d42bbad941060490c723f87313c53/library/std/src/lib.rs#L38

src/kani-compiler/rustc_codegen_kani/src/utils/names.rs
zhassan-aws: This is a unique identifier for each check generated by codegen. I added documentation for the function that should hopefully clarify.

zhassan-aws: The documentation I added should hopefully clarify what this function is doing. I'm not good with picking names, so I'm open for suggestions.

tests/expected/reach/bounds/reachable_fail/test.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: --assertion-reach-checks --output-format regular --no-default-checks
+
+// This test checks that kani injects a reachability check for
+// index-out-of-bounds checks and that it reports ones that are unreachable.
+// The check in this test is unreachable and doesn't hold, so should be reported
+// as FAILURE
```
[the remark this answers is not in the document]
zhassan-aws: Yes, this test is checking that with `--assertion-reach-checks` turned on, we still report reachable failures as "FAILURE" (and don't show the reachability check in the results). Under the `tests/expected/reach` directory, there are different check types (e.g. bounds, overflow, div-zero, etc.), and underneath each, there are 3 tests: one where the check is unreachable, one where it is reachable and passes, and one where it is reachable and fails. This is based on a suggestion from @celinval in the previous PR: https://github.com/model-checking/kani/pull/798#pullrequestreview-882408976

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
zhassan-aws: Removed clones.

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
zhassan-aws: Good point! I added another function.

tests/expected/reach/bounds/reachable_fail/test.rs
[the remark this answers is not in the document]
celinval: Does this comment needs to be updated to say that the check is *reachable*?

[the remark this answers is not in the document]
zhassan-aws: I now see what @danielsn meant 😄 Fixed.

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
```diff
@@ -78,18 +79,24 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                            Stmt::goto(self.current_fn().find_label(target), loc.clone()),
-                            None,
-                            loc.clone(),
-                        ),
-                        Stmt::assert_false(&format!("{:?}", msg), loc.clone()),
-                        Stmt::goto(self.current_fn().find_label(target), loc.clone()),
-                    ],
-                    loc,
-                )
+                let mut msg_str = format!("{:?}", msg);
```
danielsn, line 82: What if we tagged EVERY assertion with a check id?

src/kani-compiler/rustc_codegen_kani/src/utils/names.rs
danielsn: reachability_check_message?

tests/expected/reach/bounds/unreachable/expected
```diff
@@ -0,0 +1 @@
+UNREACHABLE
```
danielsn, line 1: This only guarantees that there is at least one check that is unreachable, not that all expected checks are

src/kani-compiler/rustc_codegen_kani/src/utils/names.rs
celinval: Have you considered using the metadata file instead to do the mapping?

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
zhassan-aws: Currently, this would be a bit problematic because the `expected` test suite mostly uses the old format for which the CBMC parser is not run. And it looks like CBMC currently doesn't have a way to tag assertions (see https://github.com/diffblue/cbmc/issues/6692).

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
zhassan-aws: I changed them to immutable by using the `let x = if {...} else {...};` pattern in which I used the `Stmt::Skip` type. Let me know if this looks better.

src/kani-compiler/rustc_codegen_kani/src/utils/names.rs
zhassan-aws: @danielsn and I discussed this option yesterday. The difficulty here is that the output from CBMC wouldn't include this metadata, so the CBMC parser script will need to parse the metadata file and use it when parsing the CBMC output.

zhassan-aws: Updated.

tedinski (approved):

Issue #689, opened 2021-12-10
Introduce rmctool attribute validation


tedinski:
We presently do not do any validation of `rmctool` attributes in the rmc compiler.

Presently, we only have `rmctool::proof`, so the only real missing validations are:

* [x] Check to ensure there aren't multiple proof attributes on the function.
* [x] Check there's no parameters to proof (`#[rmc::proof(stuff, here)]`)
* [ ] Check the function being annotated to ensure it has no parameters. (Initially: in the future, we might try to handle parameters as `nondet`. But this is something to design to handle the assumptions problem...)

When we add more attributes, those need validation too.

Possibly we should do this at the `proc_macro` level, instead of in the rmc backend of the compiler?

Issue #600, opened 2021-10-29
Add support for common proof options


celinval:
Requested feature: We would like to allow users to customize rmc behavior for specific proof harnesses by adding attributes to the harnesses themselves. Today, users have to change how they invoke RMC.

Use case:

```rust
#[rmc::unwind(value=10)]
#[rmc::expect_failure("<message>")]
fn my_harness() {
    // code
}
```

Proposal: We should create macro attributes in our `rmc_macros` crate that allows the user to specify unwind values as well as proof expectations. Our compiler component will any proof harness attribute and dump it to the metadata file introduced in this PR: #668. Then we will use those values to control CBMC (for unwind) and the output results (for expect_failure).

If the user provides `--unwind <value>` argument, it should have precedence over the attribute. 

Tasks breakdown:

- [x] Add support to `rmc::unwind`.
- [ ] Add support to `rmc::expect_failure`.

jaisnan:
Will look at this.

adpaco-aws:
We had an exciting discussion about this today (cc @karkhaz ).

Let's assume for a moment that we have an annotation like `#[rmc::expect_failure("<message>")]` implemented. What would be its semantics?
 1. All properties whose description contains `message` are expected to fail.
 2. At least one property whose description contains `message` is expected to fail.

Now, let's discuss their usefulness: 
 * With (1), I suspect that the annotation wouldn't be useful to anyone since that'd require ALL properties with a given description to fail. For example, let's suppose we care about an overflow property in our code. We cannot expect all overflow properties to fail, and it'd be hard or impossible to even come up with such a program (even more with Kani, since it'll attempt to block executions from any failure).
 * (2) would probably be more useful than (1), since we can detect that a particular failure happened. But this presents some problems: what "verification result" do we return if there are other failures? what if the code changes so the failure persists, but its origin is different?

The `#[should_fail(...)]` annotation for Rust unit tests doesn't have these problems because the "error" comes from a single execution path, and there will always be at most *one* error in your code, so matching it with the panic message is a good idea.

But execution paths in the context of verification work differently: conditions will split them, potentially generating several failures for a program. So we should consider a set of messages, instead of a single one.

Now, should we consider a string message? That's also not clear to me. Character strings aren't reliable in general.

In principle, we've identified three levels of granularity for this feature:
 1. the proof fails (no need for arguments).
 2. the proof fails and a given property fails.
 3. the proof fails and a property related to a given line of code fails.

[...]

PR #846, opened 2022-02-19
Unwind Attribute without Python Changes

jaisnan:
### Description of changes: 

Adds Unwind Attribute without Python Changes. The `kani-metadata.json` file now has the following structure

``` json
{
    "proof_harnesses": [
        {
            "pretty_name": "harness",
            "mangled_name": "harness",
            "original_file": "/home/ubuntu/kani/tests/cargo-kani/simple-unwind-annotation/src/main.rs",
            "original_line": "17",
            "unwind_value": 10
        },
        {
            "pretty_name": "harness_2",
            "mangled_name": "harness_2",
            "original_file": "/home/ubuntu/kani/tests/cargo-kani/simple-unwind-annotation/src/main.rs",
            "original_line": "26",
            "unwind_value": null
        }
    ]
}
```

Given that the rust library looks like -

``` rust
#[kani::proof]
#[kani::unwind(10)]
fn harness() {
...
}

#[kani::proof]
fn harness_2() {
...
}
```
Update - The PR has been updated to include handling for the following scenarios, all of which produce user errors which point to suggestions on how to fix the attribute.

``` rust
#[kani::proof]
#[kani::unwind(10,5)]
fn harness_3() {
...
}

#[kani::unwind(8)]
fn harness_4() {
...
}

#[kani::proof]
#[kani::proof]
fn harness_5() {
...
}

#[kani::proof(some, argument2)]
fn harness_6() {
...
}
```

### Resolved issues:

Resolves - partly https://github.com/model-checking/kani/issues/600 and https://github.com/model-checking/kani/issues/689.

### Call-outs:

The scripting needs to parse the metadata file and apply the unwind value to the target function.

### Testing:

* How is this change tested? Somewhat

### Checklist


library/kani_macros/src/lib.rs
```diff
@@ -49,3 +50,29 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+#[proc_macro_attribute]
+pub fn unwind(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut result = TokenStream::new();
+
+    // Translate #[kani::unwind(arg)] to #[kanitool::unwind_arg_] for easier handling
+    let insert_string = "#[kanitool::unwind_".to_owned() + &attr.clone().to_string() + "]";
+
+    // Add the string that looks like - #[kanitool::unwind_value_]
+    result.extend(insert_string.parse::<TokenStream>().unwrap());
+    // No mangle seems to be necessary as removing it prevents all the attributes in a lib from being read
```
tedinski, line 71: I don't understand this comment. Unwind should be on something that already has `#[proof]` surely?

library/kani_macros/src/lib.rs
```diff
@@ -49,3 +50,29 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    // Not-RMC, Not-Test means this code shouldn't exist, return nothing.
+    TokenStream::new()
+}
+
+#[cfg(kani)]
+#[proc_macro_attribute]
+pub fn unwind(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut result = TokenStream::new();
+
+    // Translate #[kani::unwind(arg)] to #[kanitool::unwind_arg_] for easier handling
```
tedinski, line 66: Is this because you're not sure how to parse the attributes, or because you figured it out and it looked monstrous?

src/kani-compiler/rustc_codegen_kani/src/context/metadata.rs
```diff
@@ -19,8 +19,19 @@ pub struct HarnessMetadata {
...
+    pub mangled_name: String,
+    /// The value of the unwind attribute that the user wants to set
+    pub unwind_value: u32,
+}
+
 /// The structure of `.kani-metadata.json` files, which are emitted for each crate
 #[derive(Serialize)]
 pub struct KaniMetadata {
     pub proof_harnesses: Vec<HarnessMetadata>,
+    pub unwind_metadata: Vec<UnwindMetadata>,
```
tedinski, line 36: IMO, the unwind value should be an `Option<u32>` on `HarnessMetadata`, not its own structure here.

adpaco-aws (changes requested):
Unwinding values should be part of the harness metadata and not a separate struct. For me it is okay to keep it a struct (not just a `u32` value) since long term we want to support unwinding values for individual loops in addition to this one.

library/kani_macros/src/lib.rs
jaisnan: I thought it created safer parsing conditions.

library/kani_macros/src/lib.rs
jaisnan: This was a design choice made to separate unwind as an attribute from proof i.e, the user should be able to call unwind on harnesses without the proof attribute. This might be a disagreement in design that should be resolved before we add other attributes as well.

src/kani-compiler/rustc_codegen_kani/src/context/metadata.rs
jaisnan: This too comes from a fork in the design i think.

library/kani_macros/src/lib.rs
```diff
@@ -49,3 +50,29 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
     //     $item
     // )
 }
+
+#[cfg(not(kani))]
+#[proc_macro_attribute]
+pub fn unwind(_attr: TokenStream, _item: TokenStream) -> TokenStream {
+    // Not-RMC, Not-Test means this code shouldn't exist, return nothing.
```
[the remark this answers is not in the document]
jaisnan: My brain is still stuck on the old name sadly. Yes, i can do that , it might make the documentation clearer.

src/kani-compiler/rustc_codegen_kani/src/context/goto_ctx.rs
```diff
@@ -59,6 +59,7 @@ pub struct GotocCtx<'tcx> {
     pub current_fn: Option<CurrentFnCtx<'tcx>>,
     pub type_map: FxHashMap<InternedString, Ty<'tcx>>,
     pub proof_harnesses: Vec<HarnessMetadata>,
+    pub unwind_metadata: Vec<UnwindMetadata>,
```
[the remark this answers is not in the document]
jaisnan: This was an active design choice that I think needs to be discussed. I was under the opinion that each attribute needs to be able to be called separately.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -243,11 +245,33 @@ impl<'tcx> GotocCtx<'tcx> {
...
-            match kanitool_attr_name(attr).as_deref() {
-                Some("proof") => self.handle_kanitool_proof(),
-                _ => {}
-            }
+            if let Some(attribute_string) = kanitool_attr_name(attr).as_deref() {
+                // This regex capture operation returns a vector of strings that can be matched for our purpose
+                // In later Attributes, it will be easy to parse the regex for the attribute arguments, regardless
+                // of the type of the argument itself if we capture all the patterns as a vector of strings of the format
+                // [command, argument]
+                let captures = input_re.captures(attribute_string).map(|captures| {
```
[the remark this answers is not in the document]
jaisnan: It makes parsing convenient. I tried searching for better solutions to parsing the but I think this solves it the easiest way. Happy to listen to any alternate ways of doing this.

jaisnan:
Updated the PR as per suggestions.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -243,16 +244,65 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                // Push to attribute vector that can be expanded to a map when more options become available
+                else {
+                    attribute_vector.push((attribute_string.to_string(), attr));
+                }
             }
         }
+
+        // In the case when there's only one proof attribute (correct behavior), create harness and modify it
+        // depending on each subsequent attribute that's being called by the user.
+        if proof_attribute_vector.len() == 1 {
```
adpaco-aws, line 270: These cases do not seem to overlap so I am not sure why you are using if-then-else statements here. I would also recommend handling them in small functions if possible.

src/kani-compiler/rustc_codegen_kani/src/context/metadata.rs
```diff
@@ -17,6 +17,8 @@ pub struct HarnessMetadata {
     pub original_file: String,
     /// The line in that file where the proof harness begins
     pub original_line: String,
+    /// Optional data to store unwind value
+    pub unwind_value: Option<u128>,
```
adpaco-aws, line 21: I thought we agreed having a specific struct for unwind data where this would be an optional value. Also, `Option<u32>` should be enough.

tests/cargo-kani/simple-unwind-annotation/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+
+[package]
+name = "simple-unwind-annotation"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+
+[kani.flags]
+output-format = "old"
```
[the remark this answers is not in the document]
adpaco-aws: No newline here yet.

tests/cargo-kani/simple-unwind-annotation/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+#[kani::proof]
+fn harness_2() {
+    let mut counter = 0;
+    loop {
+        counter += 1;
+        assert!(counter < 10);
+    }
+}
+
+// #[kani::proof]
```
adpaco-aws, line 34: Why is this commented code included? If there is a reason to, add a comment explaining why.

tests/kani/Unwind-Attribute/fixme_lib.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// rmc-flags: --no-unwinding-checks
+
+// Fix me
```
adpaco-aws, line 6: Please add the reason as to why this is a fix me test, including relevant issues if possible.

adpaco-aws (changes requested):

src/kani-compiler/rustc_codegen_kani/src/context/metadata.rs
jaisnan: A struct didn't seem necessary for now as we're just targeting a default value for the entire harness. The struct and related changes will be added with the changes to map each loop to it's own harness.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
jaisnan: They do overlap as it's necessary to check to how many proof attributes have been declared per harness and depending on the number, there's different handling that needs to happen. The entire function has a single purpose so it didn't seem necessary to factor it out into smaller functions.

tests/cargo-kani/simple-unwind-annotation/Cargo.toml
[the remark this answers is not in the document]
jaisnan: will add, thanks!

library/kani_macros/src/lib.rs
```diff
@@ -49,3 +51,27 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
     //     $item
     // )
 }
+
+#[cfg(not(kani))]
+#[proc_macro_attribute]
+pub fn unwind(_attr: TokenStream, _item: TokenStream) -> TokenStream {
+    // This code is never called by default as the config value is set to kani
+    TokenStream::new()
+}
```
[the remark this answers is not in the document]
jaisnan: From what i understand, it's necessary to declare for when the configuration is set to anything but Kani (either deliberately or accidentally). in that case, we don't want the unwind attributes to cause a panic so we just return an empty tokenstream.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
celinval: @jaisnan, if I understand this code correctly, you are trying to restrict any other `kani` attribute to things that are tagged with the `proof` attribute, right?

If that is the case, I would recommend to change this to have something like the following structure:
```rust
        let attrs: kani_attributes(self.tcx.get_attrs(instance.def_id()));  // BTreeMap<&str, &rustc_ast::Attribute>
        if attrs.contains_key("proof") {
            // Handle attributes here.
        } else {
            if attrs.len() > 0 {
                // Warning: We currently only support attributes on proof harnesses. The attribute {} will be ignored.
            }
        }
```

tests/cargo-kani/simple-unwind-annotation/src/main.rs
celinval: Can we change these to be UI tests that we ensure that Kani prints the correct error message?

library/kani_macros/src/lib.rs
```diff
@@ -36,9 +37,10 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
 
 #[cfg(kani)]
 #[proc_macro_attribute]
-pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
+pub fn proof(attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut result = TokenStream::new();
 
+    assert!(attr.to_string().len() == 0, "#[kani::proof] does not take any arguments");
```
celinval, line 43: Nice!

library/kani_macros/src/lib.rs
```diff
@@ -49,3 +51,27 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    // This code is never called by default as the config value is set to kani
+    TokenStream::new()
+}
+
+#[cfg(kani)]
+#[proc_macro_attribute]
+pub fn unwind(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut result = TokenStream::new();
+
+    // Translate #[kani::unwind(arg)] to #[kanitool::unwind(8)]
```
celinval, line 67: s/ #[kanitool::unwind(8)]/ #[kanitool::unwind(arg)]

library/kani_macros/src/lib.rs
```diff
@@ -49,3 +51,27 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+#[cfg(not(kani))]
+#[proc_macro_attribute]
+pub fn unwind(_attr: TokenStream, _item: TokenStream) -> TokenStream {
+    // This code is never called by default as the config value is set to kani
+    TokenStream::new()
+}
+
+#[cfg(kani)]
+#[proc_macro_attribute]
+pub fn unwind(attr: TokenStream, item: TokenStream) -> TokenStream {
```
celinval, line 64: Can you please document how this should be used? What kind of arguments does it accept and mention that today we restrict its usage to proof_harnesses only.

library/kani_macros/src/lib.rs
```diff
@@ -49,3 +51,27 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    let mut result = TokenStream::new();
+
+    // Translate #[kani::unwind(arg)] to #[kanitool::unwind(8)]
+    let insert_string = "#[kanitool::unwind(".to_owned() + &attr.clone().to_string() + ")]";
+
+    // Add the string that looks like - #[kanitool::unwindvalue]
+    result.extend(insert_string.parse::<TokenStream>().unwrap());
+    // No mangle seems to be necessary as removing it prevents all the attributes in a lib from being read
+    result.extend("#[no_mangle]".parse::<TokenStream>().unwrap());
```
celinval, line 73: Do we still need this?

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -263,20 +313,64 @@ impl<'tcx> GotocCtx<'tcx> {
...
             original_line: loc.line().unwrap().to_string(),
+            unwind_value: None,
         };
 
-        self.proof_harnesses.push(harness);
+        harness
+    }
+
+    /// Unwind strings of the format 'unwind_x' and the mangled names are to be parsed for the value 'x'
+    fn handle_kanitool_unwind(&mut self, attr: &Attribute, harness: &mut HarnessMetadata) {
```
celinval, line 323: Since we are planning to eventually support unwind outside of the metadata, I would make this function independent from harness metadata. I.e.:

```rust
fn handle_unwind(&mut self, attrs: &BTreeMap<&str, &Attribute>) -> Option<UnwindMetadata> {
}
```

BTW, see my other comment about `UnwindMetadata` type.

src/kani-compiler/rustc_codegen_kani/src/context/metadata.rs
celinval: Maybe create a type alias here that we can extend later.

```rust
type UnwindMetadata = usize;
```

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -263,20 +313,64 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        }
     }
 }
 
 /// If the attribute is named `kanitool::name`, this extracts `name`
 fn kanitool_attr_name(attr: &ast::Attribute) -> Option<String> {
     match &attr.kind {
         ast::AttrKind::Normal(ast::AttrItem { path: ast::Path { segments, .. }, .. }, _)
-            if segments.len() == 2 && segments[0].ident.as_str() == "kanitool" =>
+            if (!segments.is_empty()) && segments[0].ident.as_str() == "kanitool" =>
```
celinval, line 346: +1

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -263,20 +313,64 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            }
+        } else {
+            // User warning for when there's more than one unwind attribute, in this case, only the first value will be
+            self.tcx.sess.span_err(attr.span, "Use only one Unwind Annotation per Harness");
+        }
     }
 }
 
 /// If the attribute is named `kanitool::name`, this extracts `name`
 fn kanitool_attr_name(attr: &ast::Attribute) -> Option<String> {
```
celinval, line 343: maybe this should return `Option<&str>` since we keep converting it to `&str`.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -263,20 +313,64 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            for index in 1..segments.len() {
+                new_string.push_str(segments[index].ident.as_str());
+            }
+            Some(new_string)
         }
         _ => None,
     }
 }
+
+/// Extracts the integer value argument from the unwind attribute
```
celinval, line 358: Can you please update the comments in this function? The function is generic but the comments are only related to unwind.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -263,20 +313,64 @@ impl<'tcx> GotocCtx<'tcx> {
...
+fn extract_integer_argument(attr: &Attribute) -> Option<u128> {
+    // Vector of meta items , that contain metadata about the annotation
+    let attr_args = attr.meta().and_then(|x| x.meta_item_list().map(|x| x.to_vec()))?;
```
celinval, line 361: you can call `attr.meta_item_list()` directly.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -263,20 +313,64 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            // User warning for when there's more than one unwind attribute, in this case, only the first value will be
+            self.tcx.sess.span_err(attr.span, "Use only one Unwind Annotation per Harness");
+        }
     }
 }
 
 /// If the attribute is named `kanitool::name`, this extracts `name`
 fn kanitool_attr_name(attr: &ast::Attribute) -> Option<String> {
     match &attr.kind {
         ast::AttrKind::Normal(ast::AttrItem { path: ast::Path { segments, .. }, .. }, _)
```
celinval, line 345: Since you are changing this function, can you please replace this logic by calling `attr.name_or_empty()`?

celinval (commented):
Thanks for doing this! I have a few suggestions to clean the code a bit. Let me know if you disagree with any of them. :)

src/kani-compiler/rustc_codegen_kani/src/context/metadata.rs
jaisnan: do we want`usize` or `u32/64`. Doesnt `usize` tie the limit of the unwind value to the target architecture?

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -243,16 +244,65 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                .span_err(proof_attribute_vector[0].span, "Only one Proof Attribute allowed");
+        }
+        // User error handling for when there's an attribute being called without #kani::tool
+        else if proof_attribute_vector.len() == 0 && attribute_vector.len() > 0 {
+            self.tcx.sess.span_err(
+                attribute_vector[0].1.span,
+                format!("Please use '#kani[proof]' above the annotation {}", attribute_vector[0].0)
+                    .as_str(),
+            );
+        } else {
```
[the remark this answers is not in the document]
jaisnan: When neither proof or any other attributes are used. This function is called on all instances in `codegen_function` so raising an error might break a lot of cases. Hence, the empty handling.

src/kani-compiler/rustc_codegen_kani/src/context/metadata.rs
celinval: I suggested `usize` because that's usually used to restrict structure sizes, but in  practice, I highly doubt someone is going to use a large unwind parameter. So `u32` is fine too.

tests/cargo-kani/simple-unwind-annotation/src/main.rs
jaisnan: sure

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
jaisnan: I think using a BTreeMap would prevent us from checking if there are duplicate attributes in the same harness or not. If someone calls two `unwind` annotations then, the last one is going to be read and parsed if we use a BTreeMap I think. 

We ideally should not just check if a key exists but also how many instances of them. Maybe I can use a multimap? https://docs.rs/btreemultimap/latest/btreemultimap/ ?

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
tedinski: Not sure I understand this comment, if we're putting unwind attributes elsewhere from proof harnesses (imo, still an 'if'), that's going to require a ton of design work, the choices of which we can't really anticipate. This approach seems simple. Trying to anticipate what we might do later is overly complex for the time being.

src/kani-compiler/rustc_codegen_kani/src/context/metadata.rs
tedinski: IMO stay an integer. Any future unwinding information (e.g. unwindset) can be provided in separate fields. We're never going to change `unwind_value` because it's just the thing we pass to cbmc as `--unwind` and that's it.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
jaisnan: Thanks!

library/kani_macros/src/lib.rs
jaisnan: No, not anymore. let me update that.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
jaisnan: Returning Option<&str> would return a dangling pointer as the original string would be destroyed once the function returns. We need to return an Owned String so I don't think we can replace the return type.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
jaisnan: Will do!

library/kani_macros/src/lib.rs
jaisnan: Update :  We do need this. Without it, all unwind annotations without `kani::proof` just get ignored and this prevents us from showing an error to the user. IMO, that would confuse the user as they would not know what the correct syntax for the annotation is. So, I think we should keep no_mangle on.

celinval: No mangle has a side effect of making a function public and potentially causing name conflicts (E.g.: https://github.com/rust-lang/rust/issues/28179).

BTW, how did you test this behavior? Can you please make sure that the function you are annotating isn't dead code. Feel free to ping me offline.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
celinval: Indeed, I didn't notice that you changed the code to join the segments. For the original code, the return value would have the same lifetime as the argument `attr`.

BTW, why did you change it? This code might actually be doing the wrong thing. I.e., let's say you have the following attribute:
```rust
#[kanitool::type::sub_type]
```
I believe that you are going to return the string `typesub_type`.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -243,16 +245,64 @@ impl<'tcx> GotocCtx<'tcx> {
     fn handle_kanitool_attributes(&mut self) {
         let instance = self.current_fn().instance();
 
+        // Vectors for storing instances of each attribute type,
+        // TODO: This can be modifed to use Enums when more options are provided
+        let mut attribute_vector = vec![];
+        let mut proof_attribute_vector = vec![];
```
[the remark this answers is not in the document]
jaisnan: We need to maintain a count of the instances apart from the attributes themselves, hence vectors.

library/kani_macros/src/lib.rs
celinval: Talked offline. We'll remove this.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -243,16 +245,64 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                }
+            }
+            // self.proof_harnesses contains the final metadata that's to be parsed
+            self.proof_harnesses.push(harness_metadata);
+        }
+        // User error handling for when there's more than one proof attribute being called
+        else if proof_attribute_vector.len() > 1 {
+            self.tcx
+                .sess
+                .span_err(proof_attribute_vector[0].span, "Only one Proof Attribute allowed");
```
celinval, line 292: Reduce this to warning.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
tedinski: Oh now I see what you mean. I'm not sure if it is cleaner though. The need to give good error messages demands we spot duplicates and things like that. If we try to make this function "cleaner" then we end up pushing a lot of unwind-specific logic (like detecting duplicate unwind attributes) up to the calling function instead of here in the 'unwind' handler.

This organization lets that calling function have that nice, clean "loop over kanitool attributes, call handler function" and that's it.

src/kani-compiler/rustc_codegen_kani/src/context/metadata.rs
tedinski: I don't know why any future support for those wouldn't just add an `unwind_max_value` as well. There doesn't seem to be a value to nesting it. We control the producer (kani-compiler) and consumer (cargo-kani) of this structure, so if it's looks like a mistake later, we can just fix it. So simplest choice seems best to me.

But I don't feel strongly about it.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -263,20 +313,68 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    // Only accept unwind attributes with one integer value as argument
+    if attr_args.len() == 1 {
+        // Returns the integer value that's the argument for the unwind
+        let x = attr_args[0].literal()?;
+        match x.kind {
+            LitKind::Int(y, ..) => Some(y),
+            _ => None,
+        }
+    }
+    // Return none if there are no unwind attributes or if there's too many unwind attributes
```
celinval, line 376: unwind comment in the generic function.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
celinval: Anyway, not a biggie. Let's move on.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
celinval: Talked offline, this is no longer needed. Going to revert the change.

celinval (commented):
Here are some notes from our chat.

jaisnan:
1. Removed `no_mangle`
2. Refactored the parsing into functions, added utility functions
3. Added documentation for all user cases
4. Added a test to check failure of harness assert instead of `main`.

library/kani_macros/src/lib.rs
```diff
@@ -9,6 +9,7 @@
 //   RUSTFLAGS="-Zcrate-attr=feature(register_tool) -Zcrate-attr=register_tool(kanitool)"
 
 // proc_macro::quote is nightly-only, so we'll cobble things together instead
+extern crate proc_macro;
```
tedinski, line 12: was this necessary?

library/kani_macros/src/lib.rs
```diff
@@ -49,3 +51,27 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+#[cfg(not(kani))]
+#[proc_macro_attribute]
+pub fn unwind(_attr: TokenStream, _item: TokenStream) -> TokenStream {
+    // This code is never called by default as the config value is set to kani
+    TokenStream::new()
+}
+
+// The attribute '#[kani::unwind(arg)]' can only be called alongside '#[kani::proof]'.
+// arg - Takes in a integer value (u32) that represents the unwind value for the harness.
+// Usage is restricted to proof harnesses only currently.
```
tedinski, line 64: Doc-comments in Rust use 3 `/`s not 2, check all your function comments for this.

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -241,18 +243,54 @@ impl<'tcx> GotocCtx<'tcx> {
     /// Currently, this is only proof harness annotations.
     /// i.e. `#[kani::proof]` (which kani_macros translates to `#[kanitool::proof]` for us to handle here)
     fn handle_kanitool_attributes(&mut self) {
-        let instance = self.current_fn().instance();
+        let attributes_list = self.tcx.get_attrs(self.current_fn().instance().def_id());
+        let (proof_attribute_vector, attribute_vector) = create_attribute_vectors(attributes_list);
```
tedinski, line 247: Naming suggestion: don't put the type into the name really.

IMO, these names are clearer as "all_attributes" "proof_attributes" "other_attributes".

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -241,18 +243,54 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        let proof_attribute_count = proof_attribute_vector.len();
 
-        for attr in self.tcx.get_attrs(instance.def_id()) {
-            match kanitool_attr_name(attr).as_deref() {
-                Some("proof") => self.handle_kanitool_proof(),
-                _ => {}
+        // Handle the attributes only if there is one proof attribute
+        if proof_attribute_count == 1 {
+            self.create_proof_harness(attribute_vector);
+        }
```
tedinski, line 253: There's a very common programming convention to write functions like:

```
fn foo() {
  if error condition {
    // bail, whatever
  }
  if other error {
    // bail
  }
  // etc
  // no errors:
  the conventional, expected case.
}
```

So try that. Spot errors, raise them and return. Then just unconditionally `create_proof_harness` at the end of the function (no need for any empty else). By following the convention, functions are usually a lot easier to read (if only because people expect it)

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -241,18 +243,54 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                }
             }
         }
+        // self.proof_harnesses contains the final metadata that's to be parsed
+        self.proof_harnesses.push(harness_metadata);
     }
 
     /// Update `self` (the goto context) to add the current function as a listed proof harness
-    fn handle_kanitool_proof(&mut self) {
+    fn handle_kanitool_proof(&mut self) -> HarnessMetadata {
```
tedinski, line 293: Function doc comment needs updating. Also, we should rename this, maybe. "default_kanitool_proof" maybe?

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -263,20 +301,93 @@ impl<'tcx> GotocCtx<'tcx> {
...
         };
 
-        self.proof_harnesses.push(harness);
+        harness
+    }
+
+    /// Unwind strings of the format 'unwind(x)' and the mangled names are to be parsed for the value 'x'
+    fn handle_kanitool_unwind(&mut self, attr: &Attribute, harness: &mut HarnessMetadata) {
+        // Check if some unwind value doesnt already exist
+        if harness.unwind_value.is_none() {
```
tedinski, line 313: Same suggestion as previous comment on handling errors, then the expected case.

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -263,20 +301,93 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            // Get Attribute value and if it's not none, assign it to the metadata
+            if let Some(integer_value) = extract_integer_argument(attr) {
+                // Convert the extracted u128 and convert to u32
+                assert!(
+                    integer_value < u32::MAX.into(),
+                    "Value above maximum permitted value - u32::MAX"
+                );
```
tedinski, line 320: Good call. But we might want to `span_err` instead of assert.

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -263,20 +301,93 @@ impl<'tcx> GotocCtx<'tcx> {
...
         {
             Some(segments[1].ident.as_str().to_string())
         }
         _ => None,
     }
 }
+
+// Create two vectors, proof_vector and attribute_vector which seperates the list of attributes into
+// proof and non-proof for easier parsing
+fn create_attribute_vectors(
```
tedinski, line 349: Rename `partition_kanitool_attributes` maybe?

Second suggestion: try improving the doc-comment enough that all the comments in the body aren't necessary anymore. This function is small enough this is possible and would very likely be an improvement.

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -241,18 +243,54 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            // No op for instances without any annotations declared
+        }
+    }
+
+    // Create the proof harness struct using the handler methods for various attributes
+    fn create_proof_harness(&mut self, attribute_vector: Vec<(String, &Attribute)>) {
+        // In the case when there's only one proof attribute (correct behavior), create harness and modify it
+        // depending on each subsequent attribute that's being called by the user.
+        let mut harness_metadata = self.handle_kanitool_proof();
+        if attribute_vector.len() > 0 {
```
tedinski, line 276: unnecessary if

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -241,18 +243,54 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            // loop through all subsequent attributes
+            for attribute_tuple in attribute_vector.iter() {
+                // match with "unwind" attribute and provide the harness for modification
+                match attribute_tuple.0.as_str() {
+                    "unwind" => {
+                        self.handle_kanitool_unwind(attribute_tuple.1, &mut harness_metadata)
+                    }
+                    _ => {}
```
tedinski, line 284: maybe we could error or warn?

tedinski (commented):
Almost there!

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
jaisnan: Going to wrap the behavior with a match statement.

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
jaisnan: For the no-op case, can we now early return with the empty parenthesis?

tests/ui/unsupported-annotation/main.rs
```diff
@@ -0,0 +1,21 @@
...
+
+// This test is to check Kani's error handling for harnesses that have multiple proof annotations
+// declared.
+
+fn main() {
+    assert!(1 == 2);
+}
+
+#[kani::proof]
+#[kani::test_annotation]
```
tedinski, line 14: FYI: this isn't testing your code, you'd have to write `kanitool::` for that

src/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -241,18 +243,54 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        // User error handling for user cases where there's more than one proof attribute being called
+        if proof_attributes.len() > 1 {
+            self.tcx.sess.span_warn(proof_attributes[0].span, "Only one Proof Attribute allowed");
+            return;
+        }
+        // User error handling for the user case where there's an attribute being called without #kani::tool
+        if proof_attributes.is_empty() && !other_attributes.is_empty() {
+            self.tcx.sess.span_err(
+                other_attributes[0].1.span,
+                "Please use '#kani[proof]' above the annotation",
```
tedinski, line 262: typo

library/kani_macros/src/lib.rs
```diff
@@ -49,3 +50,26 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
     //     $item
     // )
 }
+
+#[cfg(not(kani))]
+#[proc_macro_attribute]
+pub fn unwind(_attr: TokenStream, _item: TokenStream) -> TokenStream {
+    // When the config is not kani, we should leave the function alone
+    _item
```
tedinski, line 58: Does it issue a warning or anything about the _? (That usually indicated unused, so maybe this should just be `item` not `_item`)

tedinski (approved):
Assuming CI passes, looks largely good. Maybe one last typo.

PR #854, opened 2022-02-22
Negative test case for `variant_count`

adpaco-aws:
### Description of changes: 

Adds a negative test for the `variant_count` intrinsic. Can easily converted into a positive one when support is added.

### Resolved issues:

Part of #727 

### Testing:

* How is this change tested? Adding another test.

### Checklist


tedinski (approved):

celinval:
Should we check what kind of error is given to the user?

adpaco-aws:
> Should we check what kind of error is given to the user?

The test is intentionally written this way to avoid error checking and to enable a smooth transition into a positive test. This is a good candidate for the newer testing suite we were discussing, and I want to move all these negative tests into that suite once it is available. But we can move it to `expected` for now if this is a concern, I am just afraid I will lose track of them in there.

celinval:
Can you add the version without comments as a fixme test?

adpaco-aws:
@celinval just added the "fixme" test as requested and updated the negative one (removed commented out code to make it simpler).

celinval (approved):
Thanks!

PR #858, opened 2022-02-24
Update rust toolchain

### LINKED ISSUE #857: Sprint 2022-03-01: Update rust toolchain version
We update the rust toolchain version once every sprint. In this sprint we will update to version `nightly-2022-02-22`

celinval:
### Description of changes: 

There were a few fixes required due to the following changes:

1. TyS is no longer public. Instead we have to use Ty directly. This also caused a bunch of little updates since TyS is used as a reference and Ty is by value.
2. TyKind::Tuple now uses List<Ty>. We no longer need to convert subst to Ty.
3. Const fields are no longer public. Use methods instead.
4. with_no_trimmed_paths is now a macro and it takes an object instead of closure.
5. Res::SelfTy is now a struct. Update match branches.
6. `rustc` can now prune matches!() statements when all possible enum values are provided. Updated our testcases to use 2 matches instead. Need to figure out a more robust solution but this unblocks the update.

### Resolved issues:

Resolves #857 

### Testing:

* How is this change tested? Regression

### Checklist

### DIFF
--- a/src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
+++ b/src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
@@ -9,7 +9,7 @@ use cbmc::goto_program::{Expr, Stmt, Symbol};
 use cbmc::InternString;
 use rustc_ast::ast;
 use rustc_middle::mir::{HasLocalDecls, Local};
-use rustc_middle::ty::{self, Instance, TyS};
+use rustc_middle::ty::{self, Instance};
 use std::collections::BTreeMap;
 use std::iter::FromIterator;
 use tracing::{debug, warn};
@@ -184,8 +184,8 @@ impl<'tcx> GotocCtx<'tcx> {
         // ```
 
         let tupe = sig.inputs().last().unwrap();
-        let args: Vec<&TyS<'tcx>> = match tupe.kind() {
-            ty::Tuple(substs) => substs.iter().map(|s| s.expect_ty()).collect(),
+        let args = match tupe.kind() {
+            ty::Tuple(substs) => *substs,
             _ => unreachable!("a function's spread argument must be a tuple"),
         };
         let starting_idx = sig.inputs().len();
--- a/src/kani-compiler/rustc_codegen_kani/src/codegen/rvalue.rs
+++ b/src/kani-compiler/rustc_codegen_kani/src/codegen/rvalue.rs
@@ -158,7 +158,7 @@ impl<'tcx> GotocCtx<'tcx> {
     fn codegen_rvalue_repeat(
         &mut self,
         op: &Operand<'tcx>,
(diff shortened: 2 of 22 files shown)


src/kani-compiler/rustc_codegen_kani/src/codegen/rvalue.rs
```diff
@@ -890,15 +886,15 @@ impl<'tcx> GotocCtx<'tcx> {
                     COMMON_VTABLE_ENTRIES
                 };
 
-                let (vt_size, vt_align) = ctx.codegen_vtable_size_and_align(&src_mir_type);
-                let size_assert = ctx.check_vtable_size(&src_mir_type, vt_size.clone());
+                let (vt_size, vt_align) = ctx.codegen_vtable_size_and_align(*&src_mir_type);
```
tedinski, line 889: `*&`?

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
```diff
@@ -184,8 +184,8 @@ impl<'tcx> GotocCtx<'tcx> {
         // ```
 
         let tupe = sig.inputs().last().unwrap();
-        let args: Vec<&TyS<'tcx>> = match tupe.kind() {
-            ty::Tuple(substs) => substs.iter().map(|s| s.expect_ty()).collect(),
+        let args = match tupe.kind() {
+            ty::Tuple(substs) => *substs,
             _ => unreachable!("a function's spread argument must be a tuple"),
         };
```
tedinski, line 190: You don't have to change this, obviously, but I see this pattern a couple of times and wonder if there's actually already a helper to do this in rustc...

tedinski (approved):
lgtm. Were the changes to bookrunner a huge bother? How annoying was that part?

src/kani-compiler/rustc_codegen_kani/src/codegen/rvalue.rs
celinval: Oops. Let me fix that

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
celinval: Let me check.

celinval:
> lgtm. Were the changes to bookrunner a huge bother? How annoying was that part?

It was annoying. The TyS change was by far the most annoying one though. Everything else was somewhat the same.

src/kani-compiler/rustc_codegen_kani/src/codegen/function.rs
celinval: I can only find these two occurrences, so I won't bother with a helper method at this point.

tedinski:
lgtm

PR #864, opened 2022-02-24
Add documentation of Kani result types

### LINKED ISSUE #863: Missing explanation of result types in docs
The documentation currently lacks an explanation of the result types that Kani produces (SUCCESS, FAILURE, UNREACHABLE, etc.)

zhassan-aws:
### Description of changes: 

Document the different results a check can have in Kani (SUCCESS, FAILURE, UNDETERMINED, UNREACHABLE).

### Resolved issues:

Resolves #863 

### Testing:

* How is this change tested? New doc tests.

### Checklist


docs/src/kani-result-types.md
```diff
@@ -0,0 +1,73 @@
...
+```
+Check 4: success_example.assertion.4
+         - Status: SUCCESS
+         - Description: "assertion failed: sum == 6"
+```
+
+2. `FAILURE`: This indicates that the check failed, i.e., the property doesn't
+hold. In this case, you can view the failure trace by re-running with the
+`--visualize` option.  This generates a `report` directory that contains the
+trace information. Load the `report/html/index.html` file in a browser to
```
tedinski, line 24: The new cargo-kani prints where the report was generated, and this is under `target/` for `cargo-kani` vs `kani`, so I recommend not stating the path here.

zhassan-aws: Done.

docs/src/kani-result-types.md
```diff
@@ -0,0 +1,80 @@
...
+    5. Divide/remainder-by-zero checks
+
+
+Example:
+
+```rust
+{{#include getting-started/result-types/src/main.rs:unreachable_example}}
+```
+
+The output from Kani when run with `--assertion-reach-checks` indicates that
```
tedinski, line 56: IMO we should emphasize this flag more, right now it seems a bit buried that this flag has to be enabled to get this result. I think burying the flag down here only makes sense if it's no by default, which it isn't yet right?

 Maybe by adding "(requires `--assertion-reach-checks`)" after UNREACHABLE above?

tedinski (approved):
One little thing, otherwise lgtm.

docs/src/kani-result-types.md
zhassan-aws: Makes sense. Added.

PR #865, opened 2022-02-24
Add basic test for --visualize

tedinski:
### Description of changes: 

Bare bones test that runs `--visualize` in CI. Checks that we print out the path to the report (which the new cargo-kani does).

### Call-outs:

Doesn't test that the report directory actually exists, but the cargo-kani prints the report location only after it successfully calls cbmc-viewer, so this should be somewhat robust unless that particular fact changes.

### Testing:

* How is this change tested? new test only

### Checklist


zhassan-aws (approved):
Thanks!

PR #868, opened 2022-02-25
Add CI script to check for bookrunner failures

adpaco-aws:
### Description of changes: 

Adds a CI script to check for bookrunner failures, and another step in the bookrunner workflow to call it.

### Testing:

* How is this change tested? CI.

### Checklist


scripts/ci/detect_bookrunner_failures.sh
```diff
@@ -0,0 +1,35 @@
...
+
+set -eu
+
+# This script checks that the number of failures in a bookrunner run are below
+# the threshold declared below.
+#
+# The threshold is roughly computed as: `1.05 * <number of expected failures>`
+# The extra 5% allows us to account for occasional timeouts. It is reviewed and
+# updated whenever the Rust toolchain version is updated.
+THRESHOLD=652
```
celinval, line 13: nit: can we add the calculation here instead of relying on the person that is updating this?
```bash
EXPECTED=<num>
THRESHOLD=$(expr ${EXPECTED} \* 105 / 100) # Add 5% threshold.
```

celinval (commented):
One tiny comment. Thanks for doing this so quick. :)

scripts/ci/detect_bookrunner_failures.sh
adpaco-aws: Done!

celinval (approved):

PR #869, opened 2022-02-25
Only compile as binary target if verifying main function

celinval:
### Description of changes: 

In order to verify a harness using `kani` command, user either have to
have a `main` function defined in their crate or they have to set the
env variable RUSTFLAGS to include --crate-type lib.

With these changes, Kani will pick the crate type based on the
--function argument.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/src/cargo-kani/src/call_single_file.rs
+++ b/src/cargo-kani/src/call_single_file.rs
@@ -56,6 +56,10 @@ impl KaniSession {
             }
         } else {
             if self.args.function != "main" {
+                // Unless entry function for proof harness is main, compile code as lib.
+                // This ensures that rustc won't prune functions that are not reachable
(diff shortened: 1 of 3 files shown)


src/cargo-kani/src/call_single_file.rs
```diff
@@ -54,6 +54,11 @@ impl KaniSession {
             if !args.contains(&t) {
                 args.push(t);
             }
+        } else {
+            if self.args.function != "main" {
```
zhassan-aws, line 58: Please add a comment that explains why we're compiling the crate as a library if the function is not main.

zhassan-aws (approved):

PR #871, opened 2022-02-28
Turn goto-c backend library into a module.

celinval:
### Description of changes: 

Simplify project structure by changing gotoc codegen code from a library to a module. I also introduced a feature flag to eventually allow disabling cprover backend in case we add another backend.

### Resolved issues:

This is related to #719 

### Testing:

* How is this change tested? No behavior change.

### Checklist


zhassan-aws (approved):
LGTM

tedinski (approved):
Seems fine, but one thing I've seen other Rust projects do as they grow is actively try to break their project up into multiple crates to improve compile times. Is that something that won't be a concern for us or this wouldn't be how we'd want it?

tedinski:
> Publish this lib crate separately from kani-compiler on crates.io.

I think we'll only need to publish a few crates on crates.io: kani, kani_macros, cargo-kani. But we can discuss that later.

I'm unopposed to this change, but I don't think we're forced into it by any constraints. As long as you think it's a structural improvement and not "something we're unfortunately forced to do" then LGTM.

celinval:
> > Publish this lib crate separately from kani-compiler on crates.io.
> 
> I think we'll only need to publish a few crates on crates.io: kani, kani_macros, cargo-kani. But we can discuss that later.
> 
> I'm unopposed to this change, but I don't think we're forced into it by any constraints. As long as you think it's a structural improvement and not "something we're unfortunately forced to do" then LGTM.

Cool. In the long run, I would like to create an engine module and abstract these gotoc APIs. But that is still some ways away. For our current architecture, I think it is cleaner to keep this inside the compiler crate. Cheers!

Issue #855, opened 2022-02-22
Enhance the expected testing framework


zhassan-aws:
Requested feature: With the recent switch to the new output format (#828), an assertion's result now spans multiple lines, e.g.:
```
RESULTS:
Check 1: foo.assertion.1
         - Status: FAILURE
         - Description: "assertion failed: x < 4"

Check 2: foo.assertion.2
         - Status: SUCCESS
         - Description: "assertion failed: x == 2"
```
However, the current `expected` test suite relies on the result being on a single line. So there is currently no way to check if the result of a specific assert was SUCCESS or FAILURE (a test that has an `expected` file with `SUCCESS` would pass if any check's result in that run was `SUCCESS`).

There is also no way to check that certain lines do **not** appear in the output.

We need to improve the `expected` testing framework to support the above requirements. One idea that was discussed is to add a mode to the CBMC result parser script that produces json with a well-defined schema. The `compiletest` tool will parse the json and provide a mechanism to query the parsed json for specific results.

zhassan-aws:
This has been largely implemented in #892. Closing.

PR #882, opened 2022-03-02
Make expected test suite work with the new UI

zhassan-aws:
### Description of changes: 

This is a TEMPORARY change to `compiletest` to allow the expected/cargo-kani test suites to work with the new output format without sacrificing on accuracy of testing, until we do a cleaner rework of our verification result testing (#855). A hack is added in compiletest that parses each result in the output of the form:
```
Check 2: foo.assertion.2
- Status: SUCCESS
- Description: "assertion failed: x == 2"
```
and reformats it to:
```
[foo.assertion.2] line X assertion failed: x == 2: SUCCESS
```
Thus, if the `expected` file has:
```
assertion failed: x == 2: SUCCESS
```
the test will pass.

A new `ui` mode is added to `compiletest` that checks Kani's unmodified output.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Call-outs:

Line numbers cannot be included in the `expected` files since they're missing from the new format (https://github.com/diffblue/cbmc/issues/6651). Thus, I had to remove all line numbers from the `expected` files.

In general, the `expected` files have to be of the form:
```
<assert description>: <result>
```
e.g.
```
assertion failed: x == 2: SUCCESS
```

### Testing:

### Checklist


adpaco-aws (approved):

zhassan-aws:
Based on an offline discussion with the team, we'll keep the expected test suite as is for now, and add a separate suite that uses the same compiletest hack in this PR.

PR #884, opened 2022-03-03
Remove temporary variables from translating MIR assert instructions

### LINKED ISSUE #873: Kani display temporary variables as part of assertion descriptions
Kani result report includes temporary variables which is not very helpful: 

```
arithmetic overflow on unsigned to signed type conversion in (signed long int)(1 * var_6)
```

This issue covers assertions that are included in the MIR or are generated by kani-compiler. I'll create a separate issue for the cbmc checks.

celinval:
### Description of changes: 

Replace debug message formatting to use the standard messages used to describe assertion failures in Rust.

### Resolved issues:

Resolves #873 

### Testing:

* How is this change tested? New tests + fixed old tests

* Is this a refactor change?

### Checklist


src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
```diff
@@ -78,14 +79,24 @@ impl<'tcx> GotocCtx<'tcx> {
...
                 };
 
+                let msg = if let AssertKind::BoundsCheck { .. } = msg {
+                    // For bounds check the following panic message is generated at runtime:
+                    // "index out of bounds: the length is {len} but the index is {index}",
+                    // but CBMC only accepts static messages so we don't add values to the message.
+                    "index out of bounds: the length is less than or equal to the given index"
+                } else {
+                    // For all other assert kind we can get the static message.
+                    msg.description()
```
zhassan-aws, line 89: I suppose this is where the "attempt to add with overflow" message is coming from?

tests/expected/check_operations/operations.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// Check the message printed when a checked operation fails.
+fn main() {
+    let v1: u8 = kani::any();
+    let v2: u8 = kani::any();
+    let _ = v1 + v2;
+    let _ = v1 - v2;
+    let _ = v1 * v2;
```
zhassan-aws, line 10: Add for division as well (with an assumption that v2 is non-zero)?

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
zhassan-aws: I think it would be worth revisiting the wording of those messages at some point (including the assert messages), and provide a more meaningful wording. For instance, for this one, I would phrase it as "No overflow with add" or "Add operation is free from overflow", or even something more verbose as "Checking for the possibility of an overflow in the add operation". IMHO, such wording would be more suitable for attaching "SUCCESS" or "FAILURE" to.

zhassan-aws (approved):
Thanks for fixing that!

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
celinval: I'm not sure I agree. These wordings are compatible with the wordings a user would get in a concrete run of their code.

tests/expected/check_operations/operations.rs
celinval: Sure. I can add a few more cases here.

src/kani-compiler/rustc_codegen_kani/src/codegen/statement.rs
zhassan-aws: Yes, but as you say, these wordings are suited for when an assertion fails in a "concrete run". They indicate that this assertion failed in a particular way. The wordings are not suitable for saying "Passed" or "Success" with them. Just my two cents...

celinval: As discussed offline, we should change how we present the results instead of changing the check message.

PR #889, opened 2022-03-04
Report results as UNDETERMINED instead of SUCCESS if an unwinding assertion fails

### LINKED ISSUE #746: Misleading success when unwinding assertions fail
I tried this code:

```rust
fn main() {
    let mut x = 0;
    for i in 0..5 {
        x += 1;
    }
    assert!(x < 5);
}
```

```
rmc --unwind 5 test.rs
```

with RMC version: `f7c13ea47a8`

I expected to see this happen: since the specified unwinding is insufficient (a minimum unwinding of 6 = 5 + 1 is needed), I would expect RMC to not report "SUCCESS" for the assert since there were unexplored paths due to the insufficient unwinding. 

Instead, this happened: RMC reports SUCCESS for `assert!(x < 5)`:
```
...
[main.unwind.0] line 3 unwinding assertion loop 0: FAILURE
...
[main.assertion.3] line 6 assertion failed: x < 5: SUCCESS
...
** 1 of 32 failed (2 iterations)
VERIFICATION FAILED
[RMC] info: Verification output shows one or more unwinding failures.
[RMC] tip: Consider increasing the unwinding value or disabling `--unwinding-assertions`.
```

zhassan-aws:
### Description of changes: 

When an insufficient unwind value is specified and an unwinding assertion fails, Kani may report some checks as "SUCCESS" even though they may fail with a larger unwinding. To prevent giving this somewhat misleading result to the user, Kani will now report UNDETERMINED instead of SUCCESS if any unwinding assertion fails. The unwind tip that also gets emitted when an unwinding assertion fails:
```
[Kani] info: Verification output shows one or more unwinding failures.
[Kani] tip: Consider increasing the unwinding value or disabling `--unwinding-assertions`.
```
should prompt the user to increase the unwind value.

### Resolved issues:

Resolves #746 

### Testing:

### Checklist


tests/expected/unwind_tip/expected
```diff
@@ -1,2 +1,3 @@
+UNDTERMINED
```
jaisnan, line 1: Typo?

zhassan-aws: Yes, fixed.

jaisnan (approved):
Thanks!

tedinski (approved):

PR #890, opened 2022-03-04
Add a new testing mode for checking results in the new format

zhassan-aws:
### Description of changes: 

Add a new mode to compiletest to allow testing verification results that use the new output format without sacrificing on accuracy of testing, until we do a cleaner rework of our verification result testing (https://github.com/model-checking/kani/issues/855). With the new mode (`results-new-format`), compiletest parses each result in the output of the form:
```
Check 2: foo.assertion.2
- Status: SUCCESS
- Description: "assertion failed: x == 2"
```
and reformats it to:
```
[foo.assertion.2] line X assertion failed: x == 2: SUCCESS
```
This allows writing expected files in the form:
```
assertion failed: x == 2: SUCCESS
```
This PR includes moving all the tests that rely on the new output format from the `expected` suite to the new `results-new-format` suite, and enhancing their expected file.

This new mode will be deprecated by a more robust json-based testing.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested? Moved a subset of the regressions to the new suite.

### Checklist


tools/compiletest/src/runtest.rs
```diff
@@ -353,11 +354,11 @@ impl<'test> TestCx<'test> {
     /// Runs Kani on the test file specified by `self.testpaths.file`. An error
     /// message is printed to stdout if verification output does not contain
     /// the expected output in `expected` file.
-    fn run_expected_test(&self) {
+    fn run_expected_test(&self, new_format: bool) {
```
celinval, line 357: Since this suite is temporary, I would recommend against modifying existing functions and creating a run_new_format_test(&self) with a deprecated tag already. It makes cleaning up later on much easier and less likely that people will rely on this logic.

tools/compiletest/src/runtest.rs
```diff
@@ -376,9 +377,13 @@ impl<'test> TestCx<'test> {
 
     /// Print an error if the verification output does not contain the expected
     /// lines.
-    fn verify_output(&self, proc_res: &ProcRes, expected: &str) {
+    fn verify_output(&self, proc_res: &ProcRes, expected: &str, new_format: bool) {
```
celinval, line 380: The same for this function.

tools/compiletest/src/runtest.rs
```diff
@@ -387,6 +392,99 @@ impl<'test> TestCx<'test> {
         }
     }
 
+    /// Modify the results section of the output from the new format to the old
```
celinval, line 395: Please add a comment saying that this is a temporary hack and that it shouldn't be used for any other purpose.

scripts/kani-regression.sh
```diff
@@ -38,6 +38,7 @@ TESTS=(
     "cargo-kani cargo-kani"
     "kani-docs cargo-kani"
     "kani-fixme kani-fixme"
+    "results-new-format results-new-format"
```
celinval, line 41: Let's rename this to something like `reachability-checks` and add a note saying this is temporary.

tools/compiletest/src/runtest.rs
```diff
@@ -387,6 +392,99 @@ impl<'test> TestCx<'test> {
...
+    /// format so that it works for existing `expected` files that are based on
+    /// the old format
+    fn reformat_output(output: &str) -> String {
+        // Process the output on a line-by-line basis
+        let lines: Vec<String> = output.split('\n').map(|s| s.to_string()).collect();
+        let mut adapted_lines: Vec<String> = Vec::new();
+
+        // 1. Copy all lines until and including the "RESULTS:" line, which
```
celinval, line 403: Can you just drop these lines?

tools/compiletest/src/runtest.rs
```diff
@@ -387,6 +392,99 @@ impl<'test> TestCx<'test> {
...
+        let mut line_iter = lines.iter();
+        while let Some(line) = line_iter.next() {
+            adapted_lines.push(line.clone());
+            if adapted_lines.last().unwrap() == "RESULTS:" {
+                break;
+            }
+        }
+
+        if line_iter.clone().peekable().peek().is_none() {
+            // iterated over all the lines, so this is not the new format
```
celinval, line 414: This should error out saying that the output format did not match the one you are expecting.

tools/compiletest/src/runtest.rs
```diff
@@ -387,6 +392,99 @@ impl<'test> TestCx<'test> {
...
+        //        Check 2: foo.assertion.2
+        //        - Status: SUCCESS
+        //        - Description: "assertion failed: x == 2"
+        //
+        //    to:
+        //
+        //       [foo.assertion.2] line X assertion failed: x == 2: SUCCESS
+        //
+        // (Note that since the line numbers are missing from the new format, it
+        // will be printed as "line X")
```
celinval, line 429: Do you need to add the line? This is not adding any value to our tests, right? None of the expectations in this suite should be referring to the location at this point.

tools/compiletest/src/runtest.rs
```diff
@@ -387,6 +392,99 @@ impl<'test> TestCx<'test> {
...
+        let mut check_index = 1usize;
+        while let Some(line) = line_iter.next() {
+            // Check line is of the form: "Check <n>: foo.assertion.5"
+            if let Some(name) = line.strip_prefix(format!("Check {}: ", check_index).as_str()) {
+                // Result line is of the form: "\t - Status: <Result>"
+                let result = line_iter.next().unwrap().strip_prefix("\t - Status: ").unwrap();
```
celinval, line 435: Can you call String::trim() to get rid of the whitespace before comparing the string result here and everywhere else? This seems quite fragile.

tools/compiletest/src/runtest.rs
```diff
@@ -387,6 +392,99 @@ impl<'test> TestCx<'test> {
...
+                    .strip_prefix("\t - Description: \"")
+                    .unwrap()
+                    .to_string();
+                // Strip the double quotes from the end (last line) of the description
+                *description.last_mut().unwrap() =
+                    description.last().unwrap().strip_suffix("\"").unwrap().to_string();
+
+                // Change the output to the following form:
+                //     [<assertion name>] line X <description>: <result>
+                // where description could have multiple lines
```
celinval, line 463: Just checking. The description itself includes explicit line breaks, right? We don't wrap the lines.

celinval (commented):
I know that you are blocked on this, so I'm OK with adding this new mode. However, this new suite logic is quite fragile. We should minimize the blast radius of this change.

zhassan-aws:
I created another PR (#892) with an alternative approach that was suggested by @celinval. We'll decide which of the two approaches to go with.

jaisnan:
Since we've merged the alternative one, would it be okay to close this PR?

zhassan-aws:
Yes, thanks for the reminder.

Closing this PR as it has been replaced by #892

PR #894, opened 2022-03-08
Move vtable restrictions linker into cargo-kani

### LINKED ISSUE #875: Vtable restrictions code restructure
Currently we have the crate `library/kani_restrictions` and the binary `tools/kani-link-restrictions`.

We should:

1. Eliminate the binary in favor of merging that functionality into `cargo-kani` directly.
2. Remove `kani_restrictions`. Instead, we should have some general `kani_metadata` (crate? package?) as part of `cargo-kani`.
    * This means `kani-compiler` will depend on `cargo-kani` somehow to use these definitions.
    * This avoids having to publicly publish `kani_metadata` since it'd just be part of `cargo-kani` instead.
    * Need to clarify exactly how this would work...

tedinski:
### Description of changes: 

This eliminates `kani-link-restrictions` as a separate binary.

### Resolved issues:

Resolves #875

### Call-outs:

1. I noticed `kani_metadata` uses `InternedString` which is a bit unfortunate. Noted as a TODO to possibly remove that dependency on `cprover_bindings`
2. I wrote a lot more documentation on the file format. (So much so that Github isn't even showing this as a rename+diff, but a whole new file in its diff view!)
3. This is a small step forward towards proof harnesses. My next PR will move the `kani-metadata.json` structures to `kani_metadata` so it's not just restrictions anymore.
4. I chose to keep a `kani_metadata` library, since (with the kani-shim approach we will take with installation) it won't need to be published to crates.io.
5. I did the bare minimum clean up of the code from `kani-link-restrictions`. It could use some more love. Later.

### Testing:

### Checklist


celinval (changes requested):
It looks good. Can you please move the kani_metadata to be inside `src/` instead of `library/`?

celinval (approved):
Thanks

PR #900, opened 2022-03-08
Rewrite loops with multiple back edges to having a single back edge only

tautschnig:
### Description of changes: 

rustc reasonably translates if statements at end of a loop body to
backwards jumps to the loop head. This, however, causes CBMC's symbolic
execution to spuriously treat these as nested loops. This commit now
enables a rewrite step that had previously been built for exactly such
cases.

### Testing:

* How is this change tested? Not yet - @adpaco-aws has an example that exercises this.

### Checklist


src/cargo-kani/src/call_goto_instrument.rs
```diff
@@ -87,6 +89,16 @@ impl KaniSession {
         self.call_goto_instrument(args)
     }
 
+    fn rewrite_back_edges(&self, file: &Path) -> Result<()> {
```
tedinski, line 92: Besides the regression failure, could you maybe add basically the pull request description as a doc comment (`///`) on this function? (I know I'm missing those for the others, but let's not make it worse! 😁)

tedinski, line 92: Besides the regression failure, could you maybe add basically the pull request description as a doc comment (`///`) on this function? (I know I'm missing those for the others, but let's not make it worse! 😁)

PR #892, opened 2022-03-07
Support checking for multiple consecutive lines in the expected test suite

zhassan-aws:
### Description of changes: 

**This is an alternative to #890**

Add support for checking for multiple consecutive lines appearing in the output for the expected/cargo-kani test suites. A backslash at the end of a line in the expected file is a marker that indicates that the next line should appear on a **consecutive** line in the output, e.g. for the following `expected` file:
```
SUCCESS\
Description: "index out of bounds: the length is less than or equal to the given index"
```
the test will only pass if there's a line that contains `SUCCESS` that is immediately followed by a line that contains `Description: "index out of bounds: the length is less than or equal to the given index"`.

This is a workaround until we do a cleaner rework of our verification result testing (https://github.com/model-checking/kani/issues/855).

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested? Enhanced checking done in some of the expected tests

* Is this a refactor change?

### Checklist
- [X] Each commit message has a non-empty body, explaining why the change was made
- [X] Methods or procedures are documented
- [X] Regression or unit tests are included, or existing tests cover the modified code
- [X] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #903, opened 2022-03-08
Enhanced the checking that is done in the expected/cargo-kani tests that use the new output format

zhassan-aws:
### Description of changes: 

Tests in the expected/cargo-kani/kani-docs suites that are using the new output format could easily pass vacuously because their expected files often have the check result only. For example:
```
$ cat tests/expected/reach/bounds/reachable_pass/expected
SUCCESS
```
While the intention in the above test is to check that the assertion with description `assertion failed: x != -10` passes, it may pass due to any other check having a `SUCCESS` result in the output.

With #892, it is now possible to check for multiple consecutive lines in the output. This PR updates many such tests to rely on this support for enhancing the precision of the checking done.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


celinval (approved):
Thanks for cleaning this up

PR #906, opened 2022-03-09
Remove expression from cbmc check descriptions

### LINKED ISSUE #874: CBMC's generated checks include temporary variables in their descriptions
Kani relies on a few checks that are generated by CBMC. Those checks may contain messages that mention temporary variables which is not very helpful:

```
pointer arithmetic: pointer outside object bounds in (const char *)(void *)var_5 + (signed long int)(1 * var_6)
same object violation in (const char *)(void *)var_4 >= (const char *)(void *)var_5 + (signed long int)(1 * var_6)
```

This issue covers auto-generated cbmc checks.

celinval:
### Description of changes: 

We have no control over the expression printed. These checks usually end up printing some temporary variables which are rather confusing. For now, we use the post-processor to modify the message displayed to remove the expressions.

### Resolved issues:

Resolves #874

### Call-outs:

The new tests ensures that the logic doesn't break Kani but they are not really testing the message replacement today. I couldn't find any way to do that using `compiletest`.

### Testing:

* Is this a refactor change?

### Checklist

### DIFF
--- a/scripts/cbmc_json_parser.py
+++ b/scripts/cbmc_json_parser.py
@@ -21,10 +21,10 @@ functions:
 
 import json
 import os
-
-from colorama import Fore, Style
 import re
 import sys
+
+from colorama import Fore, Style
 from enum import Enum
 
 # Enum to store the style of output that is given by the argument flags
@@ -208,7 +208,7 @@ def postprocess_results(properties):
     remove_check_ids_from_description(properties)
 
     for property in properties:
-        property["description"] = get_friendly_description(property)
+        property["description"] = get_readable_description(property)
         if has_reachable_unsupported_constructs or has_failed_unwinding_asserts:
             # Change SUCCESS to UNDETERMINED for all properties
             if property["status"] == "SUCCESS":
@@ -259,7 +259,18 @@ def filter_properties(properties, message):
     return filtered_properties, removed_properties
 
 class CProverCheck:
-    """ Represents a CProverCheck and provides methods to replace the check. """
+    """ Represents a CProverCheck and provides methods to replace the check.
+
+        Objects of this class are used to represent specific types of CBMC's
+        check messages. It allow us to identify and to replace them by a more
+        user friendly message.
+
+        That includes rewriting them and removing information that don't
+        make sense in the ust context. E.g.:
+        - Original CBMC message: "dead object in OBJECT_SIZE(&temp_0)"
+                     Not in the original code -> ^^^^^^^^^^^^^^^^^^^^
+        - New message: "pointer to dead object"
+    """
 
     def __init__(self, msg, new_msg=None):
         self.original = msg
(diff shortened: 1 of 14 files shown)


scripts/cbmc_json_parser.py
```diff
@@ -20,6 +20,8 @@
 """
 
 import json
+import os
+
```
jaisnan, line 24: Is there any reason for this separation of libraries? If we're grouping them, then maybe we should group `json, os, re,` and `sys` together as they're native libraries? 

Or is this newline a typo?

scripts/cbmc_json_parser.py
```diff
@@ -255,6 +258,129 @@ def filter_properties(properties, message):
...
+       For CBMC checks, it will ensure that the failure does not include any
+       temporary variable.
+    """
+    original = prop["description"]
+    # Id is structured as [<function>.]<property_class>.<counter>
+    prop_class = prop["property"].rsplit(".", 3)
+    # Do nothing if prop_class is diff than cbmc's convention
+    class_id = prop_class[-2] if len(prop_class) > 1 else None
+    if class_id in CBMC_DESCRIPTIONS:
+        prop_type = [check.replace(original) for check in CBMC_DESCRIPTIONS[class_id] if check.matches(original)]
```
jaisnan, line 372: Could you add a comment that shows what this list comprehension does (or what it looks like after it's done building the list?)

scripts/cbmc_json_parser.py
```diff
@@ -255,6 +258,129 @@ def filter_properties(properties, message):
...
+        if len(prop_type) != 1:
+            if "KANI_FAIL_ON_UNEXPECTED_DESCRIPTION" in os.environ:
+                print("Unexpected description: {}".format(original))
+                print(f"  - class_id: {class_id}")
+                print(f"  - matches: {[prop_type]}")
```
jaisnan, line 377: The array brackets in `{[prop_type]}` is unnecessary. It would be printed as - `[["apple", "oranges"]]` for example. You can replace it with just `{prop_type}`.

jaisnan (approved):

scripts/cbmc_json_parser.py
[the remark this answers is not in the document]
celinval: Sure. I can do that, but do you mind if I still break then into multiple lines? I think it reads better.

scripts/cbmc_json_parser.py
celinval: Oh. Either my IDE or the autopep8 did that. Let me fix it.

scripts/cbmc_json_parser.py
celinval: Yep

scripts/cbmc_json_parser.py
celinval: My bad. This variable used to be an iterator.

scripts/cbmc_json_parser.py
```diff
@@ -255,6 +258,129 @@ def filter_properties(properties, message):
...
 
+class CProverCheck:
+    """ Represents a CProverCheck and provides methods to replace the check. """
+
+    def __init__(self, msg, new_msg=None):
+        self.original = msg
+        self.kani_msg = new_msg if new_msg else msg
+
+    def matches(self, msg):
+        return self.original in msg
```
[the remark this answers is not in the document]
celinval: I need a substring matching. The purpose of this change is to remove the parts of the message that contains some expressions. I could use regex, but I thought it was a bit too much for now. I'll add an example of replacement we're making in the comments and hopefully it will clarify the intended behaviour.

tests/ui/cbmc_checks/float-overflow/expected
```diff
@@ -0,0 +1,10 @@
+Failed Checks: NaN on +
```
zhassan-aws, line 1: Did you find a way to check that `var_n` does not appear in the output? If not, we can add support for a `^<pattern>` to `compiletest` which checks that `<pattern>` does not appear in the output.

tests/ui/cbmc_checks/float-overflow/expected
```diff
@@ -0,0 +1,10 @@
...
+Failed Checks: arithmetic overflow on floating-point addition
+Failed Checks: NaN on -
+Failed Checks: arithmetic overflow on floating-point subtraction
+Failed Checks: NaN on *
+Failed Checks: arithmetic overflow on floating-point multiplication
+Failed Checks: division by zero
+Failed Checks: NaN on /
+Failed Checks: arithmetic overflow on floating-point division
+Failed Checks: division by zero
```
zhassan-aws, line 10: Repeated lines in the expected file currently doesn't add any checking.

scripts/cbmc_json_parser.py
celinval: I figured out why there's a space. It's because the next import was using the `from <package> import <item>`. I split them to make cleaner.

tests/ui/cbmc_checks/float-overflow/expected
zhassan-aws: > BTW, I was actually thinking we might as well replace "+" by addition and so on to keep the messages consistent and more readable. Any thought?

I like that!

zhassan-aws (approved):
Thanks!

zhassan-aws:
Reviewed the latest commits, and everything LGTM.

PR #912, opened 2022-03-09
Migrate tests to new output format

zhassan-aws:
### Description of changes: 

Some of the recently introduced features (e.g. reachability checks from #798, and removal of temporary variables from #906) depend on the CBMC result parser/post-processor, which doesn't run with the old format (the CBMC native output invoked using `--output-format old`). Thus, we need to move away from the old format, and perhaps obsolete it altogether. This PR migrates all of our regressions to the new format except for a few:
1. tests/expected/assert-location/*: These are checking that the line numbers produced by Kani are correct, which are only available with the old format (until https://github.com/diffblue/cbmc/issues/6651 is integrated)
2. The docs tests (e.g. docs/src/tutorial/kani-first-steps): These will be updated in a follow-up PR, which will include updating the related sections of the docs.

The commands used for updating the expected files for the reference are:
```
sed -i 's/\(.*\): SUCCESS/SUCCESS\\\n\1/' expected
sed -i 's/\(.*\): FAILURE/FAILURE\\\n\1/' expected
```

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested?

### Checklist

### DIFF
--- a/tests/cargo-kani/dependencies/Cargo.toml
+++ b/tests/cargo-kani/dependencies/Cargo.toml
@@ -10,6 +10,3 @@ edition = "2021"
 [dependencies]
 memchr = { version = "2", default-features = false }
 rand = "0.8.4"
-
-[kani.flags]
-output-format = "old"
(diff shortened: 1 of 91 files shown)


celinval (approved):
Looks good to me. Do you mind adding the docs/ test suite as well? I think that is still running using the old format.

zhassan-aws:
Will do so in a follow-up PR that includes updating the docs.

PR #919, opened 2022-03-10
Turn on assertion reachability checks by default

### LINKED ISSUE #637: RMC should distinguish between reachable assertions that hold and unreachable ones

zhassan-aws:
### Description of changes: 

Turn on assertion reachability checks by default, and replace the option `--assertion-reach-checks` with `--no-assertion-reach-checks` that can be used for turning them off.

### Resolved issues:

Resolves #637 

### Call-outs:

Since reachability checks add some annotations to each check/assert's description, e.g.
```
[KANI_CHECK_ID_a.c64ace8a::a_0] attempt to add with overflow: FAILURE
```
the CBMC result parser **must** now be run in order to clean those up. This makes the old format pretty much unusable at this point since it generates output that is not user consumable. Perhaps we should rename it from "old" to "raw" or something similar.

### Testing:

* How is this change tested? Regressions have been updated

### Checklist


tests/expected/reach/assert/reachable_fail/test.rs
```diff
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-// kani-flags: --assertion-reach-checks --output-format regular --no-default-checks
+// kani-flags: --output-format regular --no-default-checks
```
tedinski, line 3: do we still need the output-format flag here? (and several more places below...)

(I'm also mildly surprised by the `--no-default-checks` but that seems unrelated: weren't we switching to "regular" because of assertion reachability and now both are default?)

tests/kani/DynTrait/drop_nested_boxed_dyn.rs
```diff
@@ -5,6 +5,12 @@
 // There is an implicit self-recursive call to drop_in_place, so we
 // need to set an unwind bound.
 
+// Temporarily disabling assertion reachability checks because they trigger a
+// crash in CBMC:
+// https://github.com/diffblue/cbmc/issues/6691
+// https://github.com/model-checking/kani/issues/861
+// kani-flags: --no-assertion-reach-checks
```
tedinski, line 12: This is a lot of these comment blocks. It looks like by maybe Monday the new github macos image will be deployed (usually it is by now, but apparently it's only at 3% of fleet this week), and that would let us upgrade to 5.52 which supposedly has the fix.

Planning on doing that version bump and fix all these early next week?

tests/expected/reach/assert/reachable_fail/test.rs
zhassan-aws: > do we still need the output-format flag here? (and several more places below...)
> 
No, we no longer need it. I'll remove it.

> (I'm also mildly surprised by the `--no-default-checks` but that seems unrelated: weren't we switching to "regular" because of assertion reachability and now both are default?)
>
The reason this was using `--no-default-checks` was to reduce the number of checks to make the expected tests less likely to pass due to some other check having "SUCCESS". Now that we support multi-line checks, `--no-default-checks` should no longer be needed. I'll remove it.

tests/kani/DynTrait/drop_nested_boxed_dyn.rs
zhassan-aws: > Planning on doing that version bump and fix all these early next week?

I have the branch for bumping up the version ready in https://github.com/zhassan-aws/kani/commit/ea2026570beb546afc4b1e447f184da214128a24. I was hoping that 5.52.0 would be available in github today to avoid needing to turn reachability off for those tests. Would you prefer that I hold this PR off until we get the new version of CBMC?

tedinski (approved):

PR #925, opened 2022-03-11
Property class comment expect fail

### LINKED ISSUE #793: Feature - Adding Property Classes Fields to Assertion Metadata

jaisnan:
### Description of changes: 

Adds Property_classes and Comment Fields to the Source Location Metadata for Expect_fail asserts only (for now). With this change, the output for the rust line - 

` kani::expect_fail(i > 20, "Blocked by assumption above.");` changes from

Before the change -
`[main.assertion.1] line 7 EXPECTED FAIL: FAILURE` 

After the change -
`[main.expect_fail.1] line 7 EXPECTED FAIL: FAILURE`

It's now possible to customize the assertion ID strings as well as control descriptions for various types of assertions (which currently default to `assertion`.

### Resolved issues:

Resolves #793

### Call-outs:

NOTE: This PR does not yet replace all the existing asserts with the new assert statements. That change shall be added in a separate PR

### Testing:

* How is this change tested?

### Checklist


cprover_bindings/src/goto_program/location.rs
```diff
@@ -46,18 +55,49 @@ impl Location {
...
+            _ => None,
+        }
+    }
+
+    pub fn get_property_class(&self) -> Option<InternedString> {
+        match self {
+            Location::Property { property_class, .. } => *property_class,
+            _ => None,
+        }
+    }
```
tedinski, line 84: None of these functions appear to be used.

cprover_bindings/src/goto_program/location.rs
```diff
@@ -46,18 +55,49 @@ impl Location {
...
+
     /// Convert a location to a short string suitable for (e.g.) logging.
     /// Goal is to return just "file:line" as clearly as possible.
     pub fn short_string(&self) -> String {
         match self {
             Location::None => "<none>".to_string(),
-            Location::BuiltinFunction { function_name, line: Some(line) } => {
+            Location::BuiltinFunction { function_name, line: Some(line), .. } => {
```
tedinski, line 91: There are multiple locations in the PR where you're adding dots to patterns, and these appear to be remnants of a previous version of the PR. Please fix them all.

cprover_bindings/src/irep/to_irep.rs
```diff
@@ -347,7 +360,15 @@ impl ToIrep for Parameter {
 
 impl ToIrep for Stmt {
     fn to_irep(&self, mm: &MachineModel) -> Irep {
-        self.body().to_irep(mm).with_location(self.location(), mm)
+        // Match Assert in Stmt to add fields to Source Location Irep
```
tedinski, line 363: Needs better comment to explain what's going on here.

tedinski (changes requested):
I'm having a hard time being sure about this just from reviewing on Github, but it seems like your PR has two conflicting strategies for adding property class information.

1. Adds `Property` as a `Location`, and ensures that an `Assert` has such a `Location`. Consequently, things should just work.
2. Special-casing `Assert` in `to_irep` for `Stmt`, to inject the property class info with `with_assert_properties`.

It does not seem to me that both are necessary? Seems like you implemented things both ways but only one is actually being used.

cprover_bindings/src/irep/to_irep.rs
jaisnan: sure

cprover_bindings/src/goto_program/stmt.rs
```diff
@@ -184,6 +221,34 @@ impl Stmt {
...
+            )
+        } else {
+            loc
+        };
+
+        stmt!(Assert { cond, property_class, msg }, property_location)
+    }
+
     /// `__CPROVER_assert(cond, msg);`
     pub fn assert(cond: Expr, msg: &str, loc: Location) -> Self {
```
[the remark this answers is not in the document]
jaisnan: It should be removed, and replaced with the new assert.

cprover_bindings/src/goto_program/stmt.rs
```diff
@@ -184,6 +221,34 @@ impl Stmt {
...
+            Location::property_location(
+                file,
+                function,
+                line,
+                col,
+                Some(message.into()),
+                Some(property_class.as_str().to_string().intern()),
+            )
+        } else {
+            loc
```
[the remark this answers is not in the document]
jaisnan: sure, makes sense

cprover_bindings/src/goto_program/location.rs
```diff
@@ -14,6 +14,15 @@ pub enum Location {
...
     /// `function` is `None` for global, `Some(function_name)` for function local.
     Loc { file: InternedString, function: Option<InternedString>, line: u64, col: Option<u64> },
+    /// Location for Statements that use Property Class and Description - Assert, Assume, Cover etc.
+    Property {
+        file: InternedString,
+        function: Option<InternedString>,
+        line: u64,
+        col: Option<u64>,
+        comment: Option<InternedString>,
+        property_class: Option<InternedString>,
```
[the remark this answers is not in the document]
jaisnan: Yes, that is what makes it a Property and since they're key to the variant, I'll make them mandatory.

tedinski:
Fair enough.

@celinval ? I think I like the adding the `Property` variant to the `Location` enum. It's a little weird, but having to treat locations non-uniformly (`to_irep`) I think is worse. Your thoughts?

(BTW, @jaisnan : have you audited all other places that pattern matches on locations? If we add a new variant like this, we need to make sure we don't break some code elsewhere that might have made an assumption like you did that the `Location` would be a `Loc`.)

cprover_bindings/src/goto_program/location.rs
jaisnan: Yup, thanks for spotting these.

jaisnan:
> Fair enough.
> 
> @celinval ? I think I like the adding the `Property` variant to the `Location` enum. It's a little weird, but having to treat locations non-uniformly (`to_irep`) I think is worse. Your thoughts?
> 
> (BTW, @jaisnan : have you audited all other places that pattern matches on locations? If we add a new variant like this, we need to make sure we don't break some code elsewhere that might have made an assumption like you did that the `Location` would be a `Loc`.)

I agree, it's less intuitive than adding fields onto the irep's at the last step feels more hacky. Since we intent to add properties to cover and assume as well, we would be better off just adding a new variant for such statement locations imo.

I think I did do a basic pass to audit but I don't quite remember at what point of development. I can do another check.

cprover_bindings/src/goto_program/stmt.rs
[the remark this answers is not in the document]
celinval: +1. This should at least work with the new Location type you created.

cprover_bindings/src/goto_program/location.rs
```diff
@@ -81,6 +121,27 @@ impl Location {
         Location::Loc { file, function, line, col }
     }
 
+    pub fn property_location<T, U: Into<InternedString>, V: Into<InternedString>>(
+        file: U,
+        function: Option<V>,
+        line: T,
+        col: Option<T>,
+        comment: Option<U>,
+        property_name: Option<U>,
```
celinval, line 130: IMO, we should either keep this function simple or flexible. This implementation is neither. You should either add a 4th parameter here (like `X: Into<InternedString>`) or just keep them all as one type. I.e.:  `pub fn property_location<T, U: Into<InternedString>>`

cprover_bindings/src/goto_program/stmt.rs
```diff
@@ -21,6 +21,37 @@ use tracing::debug;
 ///     `while (c) {stmt1, stmt2}`
 ///      would translate to `Stmt::while_loop(c, vec![stmt1, stmt2], loc)`
 /// Statements can also be created using the converters in the `Expr` module.
+///
+#[derive(Copy, Debug, Clone)]
+pub enum PropertyClass {
```
celinval, line 26: I like the idea of having an enum, but this code should live inside kani-compiler codegen module.

celinval (commented):
Maybe next time publish 2 PRs to compare the approaches. It's a bit hard to figure out what is needed for each.

Overall, I also prefer the new Location variant.

cprover_bindings/src/goto_program/stmt.rs
tedinski: A bit hard to do that when the `Location` type its used in is in `cprover_bindings`. IMO leave it, it's certainly not the only "Kani thing" in this crate.

celinval: This shouldn't be hard though. The Location should still work over strings since the CProver API accepts any string.

Codegen module should have a method that builds the property correctly from the `enum` value.

tedinski: Ah, right, fair.

cprover_bindings/src/goto_program/location.rs
jaisnan: Will do!

cprover_bindings/src/goto_program/stmt.rs
jaisnan: Will move it to codegen

jaisnan: Tried moving to codegen but it presented problems for the assert_stmt implementation. As the Assert is ultimately a Stmt and the StmtBody needs to be implemented, we need to retain the enum in the stmt.rs file.

Would refactoring this later be acceptable @celinval ?

cprover_bindings/src/goto_program/location.rs
```diff
@@ -14,6 +14,15 @@ pub enum Location {
     /// Location in user code.
     /// `function` is `None` for global, `Some(function_name)` for function local.
     Loc { file: InternedString, function: Option<InternedString>, line: u64, col: Option<u64> },
+    /// Location for Statements that use Property Class and Description - Assert, Assume, Cover etc.
```
danielsn, line 17: What happens if a "property" location is put on a non property statement?

cprover_bindings/src/goto_program/location.rs
```diff
@@ -81,6 +93,68 @@ impl Location {
         Location::Loc { file, function, line, col }
     }
 
+    pub fn property_location<T, U: Into<InternedString>>(
```
danielsn, line 96: Move the constraint on `U` into the `where` clause

jaisnan: Sure

src/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -98,13 +98,20 @@ impl<'tcx> GotocHook<'tcx> for ExpectFail {
         assert_eq!(fargs.len(), 2);
         let target = target.unwrap();
         let cond = fargs.remove(0).cast_to(Type::bool());
-        // add "EXPECTED FAIL" to the message because compiletest relies on it
+        //TODO: actually use the error message passed by the user.
```
tedinski, line 101: slightly confused by this TODO

cprover_bindings/src/irep/to_irep.rs
```diff
@@ -323,12 +323,25 @@ impl ToIrep for Location {
                 (IrepId::Function, Irep::just_string_id(function_name.to_string())),
             ])
             .with_named_sub_option(IrepId::Line, line.map(Irep::just_int_id)),
-            Location::Loc { file, function, line, col } => Irep::just_named_sub(vector_map![
+            Location::Loc { file, function, line, col, .. } => Irep::just_named_sub(vector_map![
```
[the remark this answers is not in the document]
jaisnan: Yeah, my bad. I removed these and reverted my commit and forgot to remove them again.

cprover_bindings/src/goto_program/location.rs
```diff
@@ -81,6 +93,69 @@ impl Location {
...
+        let file = file.unwrap().into();
+        let line = line.try_into().unwrap();
+        let col = col.map(|x| x.try_into().unwrap());
+        let function = function.intern();
+        let property_class = property_name.into();
+        let comment = comment.into();
+        Location::Property { file, function, line, col, comment, property_class }
+    }
+
+    // Create a Property type Location from an already existing Location type
```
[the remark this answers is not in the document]
jaisnan: Will keep that in mind although I didn't intend this to be a doc comment. When do we want to place doc comment vs regular comment, is there any guidelines that we follow for that?

cprover_bindings/src/goto_program/location.rs
```diff
@@ -81,6 +93,69 @@ impl Location {
...
+    pub fn create_location<T: Into<InternedString>>(
+        comment: T,
+        property_name: T,
```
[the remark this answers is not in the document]
jaisnan: Yes. I can reshape all of this to use the enum and remove the above constructor

cprover_bindings/src/goto_program/stmt.rs
```diff
@@ -184,6 +221,24 @@ impl Stmt {
...
+    ) -> Self {
+        assert!(cond.typ().is_bool());
+
+        let property_name = property_class.as_str().to_string().intern();
+        let msg = message.into();
+
+        // Create a Property Location Variant from any given Location type
+        let property_location = Location::create_location(msg, property_name, loc);
+
+        stmt!(Assert { cond, property_class, msg }, property_location)
```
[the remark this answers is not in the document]
jaisnan: I think it makes sense logically although it does seem like replication of data. Not sure what the correct approach is either.

src/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
jaisnan: Yeah, this todo is now redundant

jaisnan:
Revision : 

1. Separated out the PropertyClass Enum to codegen crate.
2. Converted all references to the enum to use `InternedString` instead.

cprover_bindings/src/goto_program/location.rs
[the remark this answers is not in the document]
jaisnan: Since we've removed the Enum, I'm marking this as resolved.

cprover_bindings/src/goto_program/stmt.rs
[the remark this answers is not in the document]
jaisnan: Kept the same structure, imo this is an easy way to mandate the new fields in everyplace where we're going to use the new Assert.

cprover_bindings/src/goto_program/location.rs
```diff
@@ -58,6 +67,9 @@ impl Location {
                 format!("<{}>", function_name)
             }
             Location::Loc { file, line, .. } => format!("{}:{}", file, line),
+            Location::Property { property_class, comment, .. } => {
+                format!("{:?}:{:?}", property_class, comment)
```
tedinski, line 71: this should be file and line too

cprover_bindings/src/goto_program/location.rs
```diff
@@ -81,6 +93,57 @@ impl Location {
...
+        let line = line.try_into().unwrap();
+        let col = col.map(|x| x.try_into().unwrap());
+        let function = function.intern();
+        let property_class = property_name.into();
+        let comment = comment.into();
+        Location::Property { file, function, line, col, comment, property_class }
+    }
+
+    /// Create a Property type Location from an already existing Location type
+    pub fn create_location<T: Into<InternedString>>(
```
tedinski, line 120: This function need a better name. `with_property`? Change the final `location` to an initial `&self`?

src/kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -0,0 +1,58 @@
...
+        match self {
+            PropertyClass::ExpectFail => "expect_fail",
+            PropertyClass::UnsupportedStructs => "unsupported_struct",
+            PropertyClass::DefaultAssertion => "assertion",
+        }
+    }
+}
+
+impl<'tcx> GotocCtx<'tcx> {
+    pub fn codegen_assert(
```
tedinski, line 42: This looks good.

src/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -98,13 +99,19 @@ impl<'tcx> GotocHook<'tcx> for ExpectFail {
         assert_eq!(fargs.len(), 2);
         let target = target.unwrap();
         let cond = fargs.remove(0).cast_to(Type::bool());
-        // add "EXPECTED FAIL" to the message because compiletest relies on it
+
+        // msg is the comment field that's set on source location which is presented as the Description field
```
tedinski, line 103: I'd just remove the comments in this function. I don't think they're adding anything?

But you removed a comment that maybe still is: we're adding "EXPECTED FAIL" because of compiletest. We shouldn't lose that comment until we fix compiletest (or our own output formatter?) to handle the `expect_fail` property class correctly.

tedinski (commented):
almost there, looking pretty good

cprover_bindings/src/goto_program/location.rs
jaisnan: Yup, will do

src/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
jaisnan: cool, i'll add it back

cprover_bindings/src/goto_program/location.rs
```diff
@@ -58,6 +67,9 @@ impl Location {
                 format!("<{}>", function_name)
             }
             Location::Loc { file, line, .. } => format!("{}:{}", file, line),
+            Location::Property { property_class, comment, file, line, .. } => {
+                format!("{:?}:{:?},{:?}:{:?}", property_class, comment, file, line)
```
tedinski, line 71: I'm fairly certain, given this function's comment, this should only be printing the file/line.

tedinski (approved):
One thing to fix then lgtm

PR #926, opened 2022-03-14
Enable test for `bitreverse` intrinsic

### LINKED ISSUE #778: Ensure testing for `bitreverse` when support in CBMC is available

adpaco-aws:
### Description of changes: 

Enables the "fixme" test we prepared for `bitreverse` after some modifications (use nondeterministic index, remove tests for signed integer types).

### Resolved issues:

Resolves #778 

### Testing:

* How is this change tested? Existing regression + this new test.

### Checklist


tests/kani/Intrinsics/bitreverse.rs
```diff
@@ -16,19 +12,15 @@ macro_rules! test_bitreverse_intrinsic {
...
-                if $get_bit_name(a, n) != $get_bit_name(b, len - n) {
-                    return false;
-                }
-            }
-            return true;
+        fn $check_name(a: $ty, b: $ty) {
+            let len: usize = (std::mem::size_of::<$ty>() * BITS_PER_BYTE);
+            let n: usize = kani::any();
+            kani::assume(n < len);
+            assert!($get_bit_name(a, n) == $get_bit_name(b, (len - 1) - n));
```
tedinski, line 19: Nice!

tests/kani/Intrinsics/bitreverse.rs
```diff
@@ -4,10 +4,6 @@
 // Check that we get the expected results for the `bitreverse` intrinsic
 // https://doc.rust-lang.org/std/intrinsics/fn.bitreverse.html
 
-// Note: Support for `__builtin_bitreverse` in CBMC is being added in
-// https://github.com/diffblue/cbmc/pull/6581
-// Tracking issue: https://github.com/model-checking/kani/issues/778
```
tedinski, line 9: I guess this might not belong in this test case if there's nothing to change here, but this work remains to be done, right?

tests/kani/Intrinsics/bitreverse.rs
```diff
@@ -38,10 +30,11 @@ fn main() {
...
-    test_bitreverse_intrinsic!(i8, check_reverse_i8, get_bit_at_i8);
-    test_bitreverse_intrinsic!(i16, check_reverse_i16, get_bit_at_i16);
-    test_bitreverse_intrinsic!(i32, check_reverse_i32, get_bit_at_i32);
-    test_bitreverse_intrinsic!(i64, check_reverse_i64, get_bit_at_i64);
-    test_bitreverse_intrinsic!(isize, check_reverse_isize, get_bit_at_isize);
+    // `reverse_bits` is also defined for signed integer types by casting the
+    // number into the corresponding unsigned type and then casting the result
+    // into the original signed type. This causes overflows unless we restrict
+    // their values considerably, making the signed versions not very
+    // interesting to test here.
```
tedinski, line 39: If bit reverse works on signed types in rust... shouldn't we test that we translate it correctly still?

tests/kani/Intrinsics/bitreverse.rs
adpaco-aws: Not sure what you mean, we are enabling this test because support for `__builtin_bitreverse` is now available in CBMC. The Kani issue will be resolved by this PR, note that most of the work on Kani's end was added in #779

tests/kani/Intrinsics/bitreverse.rs
adpaco-aws: Well, the translation is the same. Just to be clear, this is how it is defined: https://github.com/rust-lang/rust/blob/master/library/core/src/num/int_macros.rs#L277-L279

Would you be happy if we add a test like this for each integer type?
```rust
assert!((0 as i8).reverse_bits() == 0);
```

tests/kani/Intrinsics/bitreverse.rs
tedinski: > we are enabling this test because support for __builtin_bitreverse is now available in CBMC. 

Ah, that's what I was missing. ok.

tedinski:
I think is was ready, and the test taking a long time is bugging me, so merging!

PR #947, opened 2022-03-16
Remove litani submodule

ronakfof:
### Description of changes: 

- Kani currently used `Litani` as a submodule. This PR removes the submodule and installs `Litani` from the `.deb` package as a dependency.
- This PR also installs `cbmc-viewer` using the PyPI package.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested?
   Kani CI

* Is this a refactor change?
   No

### Checklist


scripts/setup/install_bookrunner_deps.sh
```diff
@@ -5,18 +5,17 @@
...
 # The book runner report is generated using [Litani](https://github.com/awslabs/aws-build-accumulator)
+FILE="litani-1.22.0.deb"
+URL="https://github.com/awslabs/aws-build-accumulator/releases/download/1.22.0/$FILE"
 
-# Litani's dependencies:
-DEPS=(
-  gnuplot # Not required but recommended
-  graphviz
-)
+set -x
```
zhassan-aws, line 11: This can be included with line 5 (`set -eux`).

zhassan-aws (approved):

scripts/setup/install_bookrunner_deps.sh
ronakfof: I kept it like this to be consistent with other scripts like [THIS](https://github.com/model-checking/kani/blob/main/scripts/setup/ubuntu/install_cbmc.sh) one.
Should i update?

zhassan-aws: Not sure why the other script is done this way, but feel free to skip making the change then.

Issue #560, opened 2021-10-18
Structure mismatch in "Generics - Phantom type parameters" example


adpaco-aws:
The example in `Rust by Example/Generics/Phantom type parameters/Testcase: unit clarification/24.rs`, which can be accessed from https://doc.rust-lang.org/rust-by-example/generics/phantom/testcase_units.html fails to codegen with

```
thread 'rustc' panicked at 'assertion failed: `(left == right)`
left: `2`,
right: `1`: Error in struct_expr; mismatch in number of fields and values.
StructTag("tag-_14457343961392166270")
[Expr { value: DoubleConstant(12.0), typ: Double, location: None }]', compiler/cbmc/src/goto_program/expr.rs:703:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: internal compiler error: unexpected panic
```

zhassan-aws:
Crash is reproducible with `af11d38d93a`. This is the example for the reference:
```rust
use std::ops::Add;
use std::marker::PhantomData;

/// Create void enumerations to define unit types.
#[derive(Debug, Clone, Copy)]
enum Inch {}
#[derive(Debug, Clone, Copy)]
enum Mm {}

/// `Length` is a type with phantom type parameter `Unit`,
/// and is not generic over the length type (that is `f64`).
///
/// `f64` already implements the `Clone` and `Copy` traits.
#[derive(Debug, Clone, Copy)]
struct Length<Unit>(f64, PhantomData<Unit>);

/// The `Add` trait defines the behavior of the `+` operator.
impl<Unit> Add for Length<Unit> {
     type Output = Length<Unit>;

    // add() returns a new `Length` struct containing the sum.
    fn add(self, rhs: Length<Unit>) -> Length<Unit> {
        // `+` calls the `Add` implementation for `f64`.
        Length(self.0 + rhs.0, PhantomData)
    }
}

fn main() {
    // Specifies `one_foot` to have phantom type parameter `Inch`.
    let one_foot:  Length<Inch> = Length(12.0, PhantomData);
    // `one_meter` has phantom type parameter `Mm`.
    let one_meter: Length<Mm>   = Length(1000.0, PhantomData);

    // `+` calls the `add()` method we implemented for `Length<Unit>`.
    //
    // Since `Length` implements `Copy`, `add()` does not consume
    // `one_foot` and `one_meter` but copies them into `self` and `rhs`.
    let two_feet = one_foot + one_foot;
[...]

Issue #556, opened 2021-10-15
Failures when iterating over results


adpaco-aws:
The "Rust by Example" book contains several examples where iterators are used over result values. Take for instance this code (actual example in that section):

```rust
// compile-flags: --edition 2018
// rmc-flags: --cbmc-args --unwind 4 --object-bits 9
#![allow(unused)]
pub fn main() {
    let strings = vec!["tofu", "93", "18"];
    let numbers: Vec<_> = strings
        .into_iter()
        .filter_map(|s| s.parse::<i32>().ok())
        .collect();
    println!("Results: {:?}", numbers);
}
```

This example does not successfully verify with RMC. Running this command
```
rmc src/test/dashboard/books/Rust\ by\ Example/Error\ handling/Iterating\ over\ Results/22.rs --cbmc-args --unwind 4 --object-bits 9 | grep FAIL
```
gets me these results:
```
[std::ptr::const_ptr::<impl *const T>::offset_from.pointer_primitives.12] line 391 pointer outside object bounds in POINTER_OBJECT(var_11): FAILURE
[std::ptr::const_ptr::<impl *const T>::offset_from.pointer_primitives.16] line 391 pointer outside object bounds in POINTER_OBJECT(var_12): FAILURE
[std::result::Result::<T, E>::ok.assertion.1] line 647 unreachable code: FAILURE
VERIFICATION FAILED
```
But the example runs fine in Rust, producing this output:
```
Results: [93, 18]
```

zhassan-aws:
Kani still reports two spurious failures on this example"
```
Check 46: _ZN4core3num60_$LT$impl$u20$core..str..traits..FromStr$u20$for$u20$i32$GT$8from_str17h7557cdc4c649c89fE.assertion.1
	 - Status: FAILURE
	 - Description: "assertion false"


Check 64: std::result::Result::<T, E>::ok.assertion.1
	 - Status: FAILURE
	 - Description: "unreachable code"
```

celinval:
I think this is a duplicate from #576. These are issues related to the fact that we are not able to link against std functions.

celinval:
This issue has been resolved by #1785. The issue is covered by the test added in the same PR. https://github.com/model-checking/kani/blob/main/tests/kani/Strings/parse.rs

Issue #555, opened 2021-10-15
Fix regression tests that require "rmc-flags: --no-undefined-function-checks"


danielsn:
- [x] Cast/cast_abstract_args_to_concrete.rs (fixed in #2999)
- [ ] FatPointers/boxmuttrait.rs
- [ ] FatPointers/boxslice1.rs
- [ ] Variadic/main.rs

PR #952, opened 2022-03-16
Add fixme tests that reproduce issues #555, #556, and #560

celinval:
### Description of changes: 

Adding a few fixme tests while doing bug triage.

### Resolved issues:
N/A

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/tests/kani/Iterator/into_iter_fixme.rs
+++ b/tests/kani/Iterator/into_iter_fixme.rs
@@ -1,0 +1,11 @@
+// compile-flags: --edition 2018
+// cbmc-flags: --unwind 4 --object-bits 9
+//
+// This reproduces the issue seen in "Failures when iterating over results".
+// See https://github.com/model-checking/kani/issues/556 for more information.
+#[kani::proof]
+pub fn main() {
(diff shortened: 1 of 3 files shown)


tests/kani/Iterator/into_iter_fixme.rs
```diff
@@ -0,0 +1,11 @@
+// compile-flags: --edition 2018
+// cbmc-flags: --unwind 4 --object-bits 9
```
zhassan-aws, line 2: Can we use `kani-flags` instead for this test and the next one?

zhassan-aws (approved):

adpaco-aws (changes requested):
I realized these files do not include the copyright (see #961) - can you please add it? Also, I would remove `--object-bits 9` since I don't think it's needed.

celinval:
> I realized these files do not include the copyright (see #961) - can you please add it? Also, I would remove `--object-bits 9` since I don't think it's needed.

Good catch! I'll update the files. Thanks

Issue #477, opened 2021-09-10
General notes on usability issues revealed by documentation writing


tedinski:
I just wanted to publicly track the things I came across somewhere as I find them, so I don't forget.

Critical (ordered):

* [x] #368 This blocks using visualize (and so blocks getting traces) when unwind is needed
* [x] #493 Blank output is very user-unfriendly
* [x] `cargo-rmc` doesn't really work well (e.g. doesn't handle dependencies), and running rmc and cargo on a crate is currently a magic incantation. #451 
* [x] A release of RMC, so users don't need to build from scratch.
  * Maybe just a brew package for RMC (CBMC has one, but CBMC viewer would also need one) (Or possibly just a cargo crate? `cargo install rmc`?)
* [x] Needing the harness to be unmangled `main` is bothersome, I really think we want something like `#[proof]` #464 
* [x] We need to write these magic declarations for things like `__nondet` that really should be implicitly available already, and really should have more idiomatic names. We should have an RMC prelude that gives us `rmc::assume` and `rmc::nondet` etc.

High (unordered)

* [x] MIR gives us a lot of generated variables that are mostly just copies and obfuscate traces. (e.g. `var_10 = 0;` then `x = var_10;`)
* [ ] There should be a "simpler" version of traces for the common case of "just tell me the nondet values in my proof harness"
* [ ] Possibly some "test case reduction" though this would be a CBMC feature. We get traces about `(534, 535)` that could have been `(0, 1)`

Medium

* [x] "Assertion failed: blah: SUCCESS" is confusing. Failed...Success? #189 
* [x] We should preserve color-coded output of rustc and cbmc. We also need to "stream" output instead of waiting until completion (for instance, to see log lines as they are printed when cbmc doesn't terminate.) I believe this is the same issue. We should connect the 'tty' instead of capturing output as a string and re-printing. 
* [ ] We should be able to annotate individual loops with an `unwind` bound instead of having to resort to a global bound.
* [x] A better summary view on the command line. Needing to `| grep FAIL` is not a good user experience. We should summarize failures at the end of a run. (CBMC task?)
* [x] Can we prioritize (even heuristically) the failures in the output? Helping users understand what they should debug first is a good idea.
* [x] Can we deal with unreachable assertions? Currently these will just report success, but maybe that wasn't intended. (Coverage debugging problem?)
* [x] Visualize should also give good output on the command line, not just dump useless xml. (CBMC/viewer task?)

avanhatt:
[...]

PR #962, opened 2022-03-18
Run all proof harnesses by default

### LINKED ISSUE #464: Add a `#[proof]` attribute that allow us to mark proof harnesses
**Requested feature:** Create an attribute that allow users to mark their proof harness functions.
**Use case:** The same way users can tag their test functions using the `#[test]` attribute, users should be able to tag their proof harness with an attribute such as `#[proof]`. We can then create a cargo subcommand, like `cargo verify` that will execute all proofs in a crate.

**Link to relevant documentation (Rust reference, Nomicon, RFC)**: N/A
**Is this a breaking change?** No

**Test case**:

We could modify the test in src/test/expected/enum/main.rs to be something like:

```rust

/* other stuff */

#[cfg(rmc)]
mod rmc_tests {
    use super::*;
    #[proof]
    fn test_one_plus_two() {
        let p = Pair::new(1, 2);
        assert!(p.sum() == 3);
    }
}
```

### LINKED ISSUE #691: End-to-end tests needed for proof harness attributes
(Partially a note to self here)

We have a test in `src/test/cargo-rmc/simple-proof-annotation` that runs the code for proof harness attributes, but we don't have any tests checking that it produces the correct answer.

This test suite has special merit for ensuring that we're actually recognizing proof annotations is all cases. We had an issue early on where the function weren't being codegen'd at all, and so we never saw them to take note of the proof annotation.

So we should include: public functions, private functions, crate-type bin, crate-type lib, conflicting names, and whatever else we think of.

tedinski:
### Description of changes: 

1. Hide (but technically keep) `--function` in favor of `--harness` which must be `#[kani::proof]`. Updated all documentation accordingly.
2. cargo-kani now parses and merges `kani-metadata.json` files, to support looking up proof harnesses.
    * Quick note: I changed dry-run output to report `*.symtab.json` instead of `dry-run.symtab.json`
4. (Minor: I uh, altered `--gen-c` to print where the file was generated.)
5. `crate-type=lib` is now unconditional and we (by documentation) require `#[kani::proof]`. Technically, `--function` can still specify anything but you need some method of ensuring the function gets code-gen'd.
6. THE MEAT: **`cargo-kani` and `kani` now default to running all proof harnesses found.**
    * Switch to just one with `--harness`
    * Behavior is to run each one one after another until either one fails (and we stop) or all suceed and we print "All harnesses successfully verified."
7. Added a new mode for `cargo-kani` mode tests. We can now add a plain `expected` file instead of `function.expected` and this will run `cargo-kani` without any `--function` flags.

phew

### Resolved issues:

Resolves #464
Resolves #691 

Contributes to #477 and #701

### Testing:

* How is this change tested? 2 new tests cases, fairly basic. Plus existing whole test suite.

### Checklist


tedinski:
oh, doh. I need to adapt bookrunner to work with needing `#[kani::proof]` annotations now. Welp, monday.

src/cargo-kani/src/call_goto_instrument.rs
```diff
@@ -28,6 +24,7 @@ impl KaniSession {
         }
 
         if self.args.gen_c {
+            println!("Wrote {}", output.to_string_lossy());
```
celinval, line 27: Is this meant to be user visible? If so, can we expand the message a bit more? Also, should we check if we're in quiet mode?

src/cargo-kani/src/call_goto_instrument.rs
```diff
@@ -124,77 +121,3 @@ impl KaniSession {
         self.run_suppress(cmd)
     }
 }
-
-/// Collect all vtable restriction metadata together, and write one combined output in CBMC's format
-fn link_function_pointer_restrictions(
```
celinval, line 129: Next time, I would prefer if you leave the refactoring of unrelated code to separate PR. Thanks

src/cargo-kani/src/main.rs
```diff
@@ -49,20 +51,34 @@ fn cargokani_main(mut input_args: Vec<OsString>) -> Result<()> {
...
+        }
+
+        if ctx.args.visualize {
+            ctx.run_visualize(&specialized_obj, "target/report")?;
+        } else {
+            let result = ctx.run_cbmc(&specialized_obj)?;
+            if result == VerificationStatus::Failure {
+                // Failure exit code without additional error message
+                drop(ctx);
+                std::process::exit(1);
```
celinval, line 72: Can we run all harnesses first instead of stopping after the first failure?

src/cargo-kani/src/main.rs
```diff
@@ -83,17 +99,37 @@ fn standalone_main() -> Result<()> {
...
-    if ctx.args.visualize {
-        ctx.run_visualize(&linked_obj, "report")?;
-    } else {
-        let result = ctx.run_cbmc(&linked_obj)?;
-        if result == VerificationStatus::Failure {
-            // Failure exit code without additional error message
-            drop(ctx);
-            std::process::exit(1);
+
+    let metadata = ctx.collect_kani_metadata(&[outputs.metadata])?;
```
celinval, line 103: Can we move the duplicated code to a separate function?

src/cargo-kani/src/metadata.rs
```diff
@@ -0,0 +1,156 @@
...
+                return Ok(vec![harness.clone()]);
+            }
+            bail!("A proof harness named {} was not found", name);
+        }
+        if metadata.proof_harnesses.is_empty() {
+            // This is not the desirable behavior we deliver to our users,
+            // but this is a hack to get kani regression tests back to working quickly.
+            // TODO: This should be removed in favor of a good error message suggesting that proof harnesses be written.
+            // Probably even with a link to Kani documentation!
+            Ok(vec![mock_proof_harness("main")])
```
celinval, line 141: Is this still needed? I thought you had added annotations to all tests already.

tests/expected/multiple-harnesses/expected
```diff
@@ -0,0 +1,3 @@
+Checking harness check_first_harness...
```
celinval, line 1: Can you please move these tests to be under `ui` suite?

tools/compiletest/src/runtest.rs
```diff
@@ -317,10 +317,12 @@ impl<'test> TestCx<'test> {
         let function_name = self.testpaths.file.file_stem().unwrap().to_str().unwrap();
         cargo
             .arg("kani")
-            .args(["--function", function_name])
             .arg("--target-dir")
             .arg(self.output_base_dir().join("target"))
             .current_dir(&parent_dir);
+        if "expected" != self.testpaths.file.file_name().unwrap() {
+            cargo.args(["--function", function_name]);
```
celinval, line 324: Do we have a jira to change these to `--harness`?

celinval (changes requested):
Sweet! Mostly minor comments. I do think we should change the behaviour to run all harnesses before failing though.

src/cargo-kani/src/args.rs
```diff
@@ -72,9 +72,12 @@ pub struct KaniArgs {
...
     pub checks: CheckArgs,
 
-    /// Entry point for verification
-    #[structopt(long, default_value = "main")]
-    pub function: String,
+    /// Entry point for verification (symbol name)
+    #[structopt(long, hidden = true)]
+    pub function: Option<String>,
+    /// Entry point for verification (proof harness)
+    #[structopt(long, conflicts_with = "function", conflicts_with = "dry-run")]
```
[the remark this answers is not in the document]
celinval: But do we need to? We can just print something like dummy in the function name, right?

I wonder if we should make the dry-run a hidden option though.

tools/compiletest/src/runtest.rs
```diff
@@ -280,15 +280,17 @@ impl<'test> TestCx<'test> {
...
+        match self.props.kani_panic_step {
+            Some(KaniFailStep::Check) => {
+                self.check();
+            }
+            Some(KaniFailStep::Codegen) => {
+                self.codegen();
+            }
+            Some(KaniFailStep::Verify) | None => {
+                self.verify();
+            }
```
tedinski, line 292: This change is necessary since we now may run `kani-compiler` from `kani` with different flags that `compiletest` wasn't picking up. This change means we avoid doing redundant work, and these differences only matter for things that expected to fail to check or codegen (which wasn't any existing tests, thankfully)

Just thought I'd note this in the review here, since it got mixed up with a bunch of other changes to get tests working again after switching to using `--harness`

tools/compiletest/src/runtest.rs
tedinski: Changed them to harness now.

src/cargo-kani/src/main.rs
tedinski: I've tried to avoid having any functions that are specific to `kani` or `cargo-kani` (except these two of course) because I think these are generally design errors. You'll notice that there are several small variations between these two functions, and trying to abstract those differences out will generally make the code worse, not better.

A small amount of not-really-duplication is better than trying to create a function with a half dozen parameters that only gets called 2 ways. The old python was full of that, and it's just code that only gets worse as people add to it.

That said, I think there IS a little bit I can factor out here, I just need to fix how we do report generation first since I actually broke that...

celinval: That makes sense. Thanks

tedinski:
This should be ready to review again.

It now runs all proof harnesses even if one fails, and prints a summary at the end:

```
```

Hmm. Possibly I should print which harnesses failed, so you don't have to scroll up to search for them... Let me know your thoughts.

We now also produce all reports with `--visualize` which required creating multiple directories, one for each harness:

```
$ cargo kani --visualize
$ ls -l target/
total 24
-rw-rw-r-- 1 ubuntu ubuntu  177 Mar 23 19:13 CACHEDIR.TAG
drwxrwxr-x 7 ubuntu ubuntu 4096 Mar 23 19:13 debug
drwxrwxr-x 4 ubuntu ubuntu 4096 Mar 23 19:22 report-proofs::check_trivial
drwxrwxr-x 4 ubuntu ubuntu 4096 Mar 23 19:22 report-proofs::failed_check
drwxrwxr-x 4 ubuntu ubuntu 4096 Mar 23 19:22 report-proofs::third_harness
drwxrwxr-x 3 ubuntu ubuntu 4096 Mar 23 19:13 x86_64-unknown-linux-gnu
```

I notice that the "pretty name" of function is fully qualified. I'm thinking about fixing that (advice?) but if we don't use the fully qualified names, there might be name collisions. (At the very least, colon isn't legal in filenames on windows, so we have to do something...)

celinval:
The way proof is implemented today by using `no_mangle` guarantees that you can't have name collision between harnesses. I believe if that happens, the compilation fails.

That said, can you replace the `::` by `.`?

src/cargo-kani/src/metadata.rs
```diff
@@ -0,0 +1,193 @@
...
+    name: &str,
+    harnesses: &'a [HarnessMetadata],
+) -> Option<&'a HarnessMetadata> {
+    for h in harnesses.iter() {
+        if h.pretty_name == *name {
+            return Some(h);
+        } else {
+            // pretty_name will be things like `module::submodule::name_of_function`
+            // and we want people to be able to specify `--harness name_of_function`
+            if let Some(prefix) = h.pretty_name.strip_suffix(name) {
```
zhassan-aws, line 165: We should probably check for the case where there is ambiguity, i.e., multiple functions that have the same name, e.g. `module1::submodule1::name_of_function` and `module2::submodule2::name_of_function`, and emit an error.

tests/smack/basic/arith.rs
```diff
@@ -1,8 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-// @expect verified
```
zhassan-aws, line 3: Is removing the SMACK annotations necessary? They're probably meant for the SMACK testing tool, but they may be useful for us to compare results.

tools/bookrunner/src/util.rs
```diff
@@ -203,7 +203,8 @@ pub fn add_codegen_job(litani: &mut Litani, test_props: &TestProps) {
 pub fn add_verification_job(litani: &mut Litani, test_props: &TestProps) {
     let exit_status = if test_props.fail_step == Some(FailStep::Verification) { 10 } else { 0 };
     let mut kani = Command::new("kani");
-    kani.arg(&test_props.path).args(&test_props.kani_args);
+    // Some of test_props.kani_args will contains `--cbmc-args` so we should always put that last
+    kani.arg(&test_props.path).args(&["--function", "main"]).args(&test_props.kani_args);
```
zhassan-aws, line 207: Are we adding `--function main` to avoid having to annotate the bookrunner examples with `#[kani::proof]`? If so, can we add a comment?

zhassan-aws (approved):
This is a huge step forward! Awesome work, @tedinski!

Just had a few minor comments.

src/cargo-kani/src/main.rs
```diff
@@ -83,20 +94,81 @@ fn standalone_main() -> Result<()> {
...
+    ) -> Result<()> {
+        if !self.args.quiet && !self.args.visualize && harnesses.len() > 1 {
+            if !failed_harnesses.is_empty() {
+                println!("Summary:");
+            }
+            for harness in failed_harnesses.iter() {
+                println!("Verification failed for - {}", harness.pretty_name);
+            }
+
+            println!(
```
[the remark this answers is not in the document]
tedinski: Maybe. In my experience, there is a contingent of "if I say quiet, actually print nothing, just give me a return value" software out there. But... maybe this isn't one that should be that way.

cargo test is a compelling point of comparison. Possibly we should do some output comparisons and decide differently later?

tests/smack/basic/arith.rs
tedinski: No: I needed to quickly audit those files to fix a few things and I removed what I thought was unnecessary. I don't see how these would ever get used...

src/cargo-kani/src/main.rs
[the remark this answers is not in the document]
zhassan-aws: Sure.

Issue #576, opened 2021-10-20
Failing verification for proptest example using format!


nchong-at-aws:
I tried this code, which is a slightly modified example from the proptest book (https://altsysrq.github.io/proptest-book/proptest/getting-started.html):

```rust
#![allow(unused)]

fn __nondet<T>() -> T {
    unimplemented!()
}

fn __VERIFIER_assume(cond: bool) {
    unimplemented!()
}

fn parse_date(s: &str) -> Option<(u32, u32, u32)> {
    if 10 != s.len() {
        return None;
    }

    if "-" != &s[4..5] || "-" != &s[7..8] {
        return None;
    }

    let year = &s[0..4];
    let month = &s[5..7]; //< fixed from example that can be found using proptest
    let day = &s[8..10];

    year.parse::<u32>().ok().and_then(|y| {
        month
            .parse::<u32>()
            .ok()
            .and_then(|m| day.parse::<u32>().ok().map(|d| (y, m, d)))
    })
}

pub fn main() {
    let y: u32 = __nondet();
    let m: u32 = __nondet();
    let d: u32 = __nondet();
    __VERIFIER_assume(0 <= y && y < 10000);
    __VERIFIER_assume(1 <= m && m < 13);
    __VERIFIER_assume(1 <= d && d < 32);
    let (y2, m2, d2) = parse_date(&format!("{:04}-{:02}-{:02}", y, m, d)).unwrap();
    assert!(y == y2);
    assert!(m == m2);
    assert!(d == d2);
}
```

using the following command line invocation:

```
rmc example.rs
```

with RMC version: a728d8d41a15c3c71795351b23eddd4ce8b54cc9

I expected to see this happen: VERIFICATION SUCCESSFUL

Instead, this happened: VERIFICATION FAILED

zhassan-aws:
Experiment with switching to running `goto-instrument` with `--generate-function-body assert-false-assume-false`

celinval:
@zhassan-aws Do you mind adding taking a look at this one while you are looking into adding compiler warnings to codegen_unimplemented? Thanks

zhassan-aws:
Summarizing a discussion from Friday 4/8:
- For the program involving `"2021".parse::<u32>()`, the failure is due to a missing definition for an `std` function, which is caused by Kani not linking the standard library.
- The reported failure is not user friendly as it involves the mangled name of the missing function as well as an unclear description ("assertion false"):
```
[...]

PR #964, opened 2022-03-19
Hide --lib-c and add assume-false to unknown fns

celinval:
### Description of changes: 

 - Support to --lib-c is not well tested. Keep it as experimental.
 - Change how CBMC behaves when a function that is not defined is
   reachable. This mitigates issues #576. We should still add an
   unimplemented assertion so we can flip the results of other checks to
   undetermined.

### Resolved issues:

None

### Call-outs:

I had to change the float test since fround() function is missing. 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/src/cargo-kani/src/call_goto_instrument.rs
+++ b/src/cargo-kani/src/call_goto_instrument.rs
@@ -79,7 +79,7 @@ impl KaniSession {
     fn undefined_functions(&self, file: &Path) -> Result<()> {
         let args: Vec<OsString> = vec![
             "--generate-function-body-options".into(),
-            "assert-false".into(),
+            "assert-false-assume-false".into(),
             "--generate-function-body".into(),
             ".*".into(),
             "--drop-unused-functions".into(),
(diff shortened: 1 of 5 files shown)


src/cargo-kani/src/call_goto_instrument.rs
```diff
@@ -79,7 +79,7 @@ impl KaniSession {
     fn undefined_functions(&self, file: &Path) -> Result<()> {
         let args: Vec<OsString> = vec![
             "--generate-function-body-options".into(),
-            "assert-false".into(),
+            "assert-false-assume-false".into(),
```
zhassan-aws, line 82: Please add a test that checks this behavior, e.g.:
```rust

extern "C" {
    fn missing_function() -> u32;
}

#[kani::proof]
fn main() {
    unsafe {
        let x = missing_function();
        assert!(x == 5);
    }
}
```
The assert should be reported as unreachable. Without your change, it fails.

zhassan-aws (changes requested):

src/cargo-kani/src/call_goto_instrument.rs
celinval: Sure.

tedinski (approved):

zhassan-aws (approved):

Issue #968, opened 2022-03-21
Documentation: Add a "Debugging coverage" section


adpaco-aws:
We should add a "Debugging failures" section where we discuss coverage, how to interpret it in traces and how to increase coverage for proofs.

Issue #967, opened 2022-03-21
Documentation: Add a "Debugging non-termination" section


adpaco-aws:
We should add a "Debugging non-termination" section where we discuss how to identify non-termination in Kani and approaches to solve it.

Issue #966, opened 2022-03-21
Documentation: Add a "Debugging failures" section


adpaco-aws:
We should add a "Debugging failures" section where we discuss how to identify failures, obtain execution traces and resolve them.

Issue #700, opened 2021-12-15
In-depth review of general documentation


adpaco-aws:
Including the following:
 * `README.md` file
 * ~~Source code documentation.~~
 * Kani docs

adpaco-aws:
For source code documentation, we would need to make clear what the scope according to the current layout:
 * Code under `src` (`kani-compiler`, `cargo-kani` and `cargo-metadata`).
 * `cprover_bindings`
 * Kani docs (i.e., files under `docs/src`)
 * Kani libraries (i.e., files under `kani/library`: The kani library, `kani-macros` and `std` overrides)
 * Scripts
 * Tests
 * Tools (`compiletest` and `bookrunner`)

PR #969, opened 2022-03-21
Documentation: Structural changes

adpaco-aws:
### Description of changes: 

Basically re-organizes the main chapters of the Kani documentation by grouping them into a few parts, which I think achieves a cleaner look and makes it easier to find information a particular topics. Reviews a few section names and moves files accordingly. 

The left bar now looks like this:

### Resolved issues:

Part of #700 
Creates #966, #967 and #968
Screen Shot 2022-03-23 at 18 51 29

### Call-outs:

There may be more changes coming to the overall structure but this contribution brings us a step closer to it.

### Testing:

* How is this change tested? Running `mdbook` locally.

### Checklist


docs/src/SUMMARY.md
```diff
@@ -1,25 +1,34 @@
...
+
+# Limitations
+- [Guarantees]()
+- [Rust feature support](./rust-feature-support.md)
+- [Overrides](./overrides.md)
+---
+# Appendices
 
-- [Limitations](./limitations.md)
+- [Comparison with other tools](./tool-comparison.md)
```
tedinski, line 34: IMO this belongs with "getting started," the purpose is mostly to answer what Kani is and why people should be interested.

docs/src/SUMMARY.md
```diff
@@ -1,25 +1,34 @@
...
+  - [Introduction](./introduction.md)
   - [Installation](./install-guide.md)
-  - [Comparison with other tools](./tool-comparison.md)
-  - [Kani on a single file](./kani-single-file.md)
-  - [Kani on a package](./cargo-kani.md)
-  - [Kani result types](./kani-result-types.md)
-  - [Debugging failures]()
-  - [Debugging non-termination]()
-  - [Debugging coverage]()
+  - [Usage](./usage.md)
```
tedinski, line 7: I don't see the need for this page, or putting the below in a sublist.

docs/src/SUMMARY.md
```diff
@@ -1,25 +1,34 @@
...
   - [Where to start on real code](./tutorial-real-code.md)
 
-- [Kani developer documentation](./dev-documentation.md)
-  - [Working with `RUSTC`](./rustc-hacks.md)
-  - [Testing](./kani-testing.md)
-  - [Book runner](./bookrunner.md)
+# Developer documentation
+  - [Command cheat sheets](./cheat-sheets.md)
+  - [Working with `rustc`](./rustc-hacks.md)
+  - [Testing](./testing.md)
```
tedinski, line 23: I similarly don't think this page and nesting is adding anything here

docs/src/SUMMARY.md
adpaco-aws: I think of "getting started" as a section that you can read in a few minutes to get a quick understanding of the project. The comparison with other tools goes a bit deeper somehow and is linked in many sections around (including the introduction). In my opinion, this is an article that will be read by people who want to know more, but right after "Introduction" may cause potential users lose interest.

docs/src/SUMMARY.md
adpaco-aws: I think splitting into subsections here (and in the next case) provides a much cleaner and organized look to our documentation. In my opinion, the previous version was a little "too flat", so adding subsections where we can may help with readability and navigation.

docs/src/usage.md
```diff
@@ -0,0 +1,12 @@
+# Usage
+
+At present, Kani can used in two ways:
+ * [On a single file](./kani-single-file.md) with the `kani` command.
+ * [On a package](./cargo-kani.md) with the `cargo-kani` command.
+
+Running [Kani on a single file](./kani-single-file.md) is quite useful for small examples.
```
celinval, line 7: or projects that don't use cargo... like Rust for Linux. :)

celinval, line 7: or projects that don't use cargo... like Rust for Linux. :)

celinval: That said, we still need to fix linking so users can actually use this with makefile or other build systems.

celinval: That said, we still need to fix linking so users can actually use this with makefile or other build systems.

docs/src/usage.md
```diff
@@ -0,0 +1,12 @@
...
+
+However, if you plan to integrate Kani in your projects, the recommended
+approach is to use [Kani on a package](./cargo-kani.md) because it includes
+utilities for tagging, extracting and running all proof harnesses in your
+package without making changes to its codebase.
```
celinval, line 12: It's not clear to me which changes users would have to make to their codebase. I would say that the main difference today between `kani` and `cargo kani` is the ability to handle external dependencies.

celinval, line 12: It's not clear to me which changes users would have to make to their codebase. I would say that the main difference today between `kani` and `cargo kani` is the ability to handle external dependencies.

tedinski: I mean, the ability to build more than a single `.rs` file at all. I don't think `kani` is even useful for customers, really.

celinval: You are able to verify a full crate with standalone kani, independent on the number of files, as long as it doesn't have any external dependency.

For example, you can verify the rectangle example inside [tests/cargo-kani/rectangle-example](https://github.com/model-checking/kani/tree/main/tests/cargo-kani/rectangle-example) with the following command:
```
kani src/lib.rs --function stretched_rectangle_can_hold_original
```
Note that the crate has two files: `src/lib.rs` and `src/rectangle.rs`.

celinval: You are able to verify a full crate with standalone kani, independent on the number of files, as long as it doesn't have any external dependency.

For example, you can verify the rectangle example inside [tests/cargo-kani/rectangle-example](https://github.com/model-checking/kani/tree/main/tests/cargo-kani/rectangle-example) with the following command:
```
kani src/lib.rs --function stretched_rectangle_can_hold_original
```
Note that the crate has two files: `src/lib.rs` and `src/rectangle.rs`.

tedinski: Oh neat. TIL

adpaco-aws: > Oh neat. TIL

+1

Let's just write that then, no?
```suggestion
However, if you plan to integrate Kani in your projects, the recommended
approach is to use [Kani on a package](./cargo-kani.md) because of its ability
to handle external dependencies.
```

adpaco-aws:
Updated with feedback from PR and discussion.

docs/src/SUMMARY.md
```diff
@@ -1,25 +1,29 @@
...
-  - [Book runner](./bookrunner.md)
+- [Developer documentation](dev-documentation.md)
+  - [Command cheat sheets](./cheat-sheets.md)
+  - [Working with `rustc`](./rustc-hacks.md)
+  - [Testing](./testing.md)
+    - [Regression testing](./kani-testing.md)
+    - [Book runner](./bookrunner.md)
 
-- [Limitations](./limitations.md)
+- [Limitations](./rust-feature-support.md)
```
zhassan-aws, line 24: Should this be called "Rust feature support"?

zhassan-aws (approved):
LGTM

docs/src/SUMMARY.md
adpaco-aws: Renamed it, although it is likely that I write a main "Limitations" section when I review this part later.

PR #975, opened 2022-03-23
Restore copyright check

### LINKED ISSUE #961: Copyright check is not working
The copyright check does not seem to be working - [this run](https://github.com/model-checking/kani/runs/5606143618?check_suite_focus=true) had copyright issues that were not detected.

I think this may be due to a change on the clone depth (from 0 to 1), which was necessary for the `git diff` command to work well.

adpaco-aws:
### Description of changes: 

The copyright check has been broken for some time. This PR restores the check by making it work with shallow clones (the reason it broke) and extends the check itself to detected "modified" copyright headers.

### Resolved issues:

Resolves #961 

### Testing:

* How is this change tested? Running in CI.

### Checklist


scripts/ci/copyright-exclude
```diff
@@ -0,0 +1,19 @@
+.clang-format
+.diff
+.md
```
danielsn, line 3: Should we use gitignore format? There is a parser for it https://pypi.org/project/gitignore-parser/

adpaco-aws: We can consider it as an enhancement, either by switching to other tools to retrieve the list of files (`git ls-files` exlcude options only work with "ignored" or "other" files) or doing that ourselves. Note that this is just a file for the (extended) patterns we had in our workflow: https://github.com/model-checking/kani/blob/main/.github/workflows/format-check.yml#L16

scripts/ci/copyright_check.py
```diff
@@ -27,31 +51,67 @@ def copyright_check(filename):
...
-    if all(checks):
+        if checks[i] == CheckResult.PASS_STANDARD:
+            print('PASS')
+        elif checks[i] == CheckResult.PASS_MODIFIED:
+            print('PASS (MODIFIED)')
+        else:
+            print('FAIL')
+
+    # Convert checks into boolean values to determine the exit code
+    bool_checks = map(result_into_bool, checks)
```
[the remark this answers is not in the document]
adpaco-aws: Right, thanks!

adpaco-aws:
> It looks good to me. Thanks for jumping on this one.
> 
> One improvement we should do later is to detect if the file has a previous copyright statement. I think this would fail today, but it should be OK to have an original copyright statement followed by our modified copyright statement.

Yes, because it expects the headers to be in an (almost) fixed set of lines in the file. But I agree, there are lots of room for improvement here (e.g., using a sliding window to detect headers instead).

PR #976, opened 2022-03-23
A few output improvements

celinval:
### Description of changes: 

1. Disable reach assertions when running --visualize.
2. Fix the file path in the compiler to use local file path.
    - Note that using the local file path is not recommended to generate artifacts that can be shared, but I don't this we have that use case today.
3. Try to be smart when displaying the file path to either do relative to current directory or to home.
    - This makes the output much nicer when the file path is deep into the hierarchy.

### Resolved issues:



### Call-outs:

Here is an example of the new output:

```
Check 16: std::boxed::Box::<T, A>::into_unique.assertion.1                                                                                                                               
         - Status: SUCCESS                                                                                                                                                               
         - Description: "resume instruction"                                                                                                                                             
         - Location: ~/.rustup/toolchains/nightly-2022-03-09-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/boxed.rs:1087:5 in function std::boxed::Box::<T, A>::into_un
ique                                                                                                                                                                                     
                                                                                                                                                                                         
Check 17: same_file.assertion.2                                                                                                                                                          
         - Status: SUCCESS                                                                                                                                                               
         - Description: "assertion failed: cond"                                                             
         - Location: main.rs:13:5 in function same_file  
```

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


tests/ui/code-location/expected
```diff
@@ -0,0 +1,7 @@
+module/mod.rs:10:5 in function module::not_empty
+main.rs:13:5 in function same_file
+~/.rustup/toolchains
```
celinval, line 3: I'm not confident we should include this expectation, but this prevents us from inadvertently use `remapped` path. Open to suggestions.

jaisnan:
Could you add some screenshot or add an example output improvement to the description of this PR?

scripts/cbmc_json_parser.py
```diff
@@ -95,9 +96,37 @@ def __init__(self, source_location={}):
...
+        """ Return a more user friendly path.
+
+        - If the path is inside the current directory, return a relative path.
+        - If not but the path is in the ${HOME} directory, return relative to ${HOME}
+        - Otherwise, return the path as is.
+        """
+        if not self.filename:
+            return None
+
+        if not path.exists(self.filename):
```
zhassan-aws, line 109: Is this necessary for what comes after? Checking if the file exists can be expensive (since it involves an OS/filesystem operation) which can slow things down quite a bit if we're checking the path for every check.

tests/ui/code-location/expected
zhassan-aws: `RUSTUP_HOME` can be overriden (see https://rust-lang.github.io/rustup/environment-variables.html), so in the current form, the test may fail for users who override it.

If you're trying to make sure the path is not of the form: `/rustc/1eb72580d076935a3e590deb6e5813a5aef3eca4/library/core/src/mem/mod.rs`, perhaps just check for `toolchains/nightly-`?

scripts/cbmc_json_parser.py
celinval: I thought that this fixed an issue that I saw before, but I couldn't reproduce it. So I'm removing this. :)

I added a few tests to hopefully ensure that I didn't miss anything.

scripts/test_parser.py
```diff
@@ -0,0 +1,102 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+import unittest
```
zhassan-aws, line 3: Much needed! Thanks for adding this unit test module!

Nit: should we renamed it to `test_cbmc_parser.py` to make it more clear?

zhassan-aws (approved):

scripts/test_parser.py
celinval: Sure

PR #980, opened 2022-03-25
Documentation: Disable play feature for all examples

adpaco-aws:
### Description of changes: 

Disables the "play" feature for all examples. The play feature sends code to the Rust playground and returns the result. Unfortunately, most examples in our documentation fail to compile or produce no output using this feature.

### Call-outs:

There is no way at the moment to disable the feature using a global setting. See https://github.com/rust-lang/mdBook/issues/747 for more details.

### Testing:

### Checklist


tedinski (approved):
Annoying it isn't global. Approved but I wonder if we should care? It seems highly likely people will accidentally introduce more snippets without this tag. Maybe we just let it uselessly exist?

adpaco-aws:
I had considered that, but in general this is bad user experience. I will monitor updates on the draft PR for the global flag, but we should have this in the meantime.

PR #984, opened 2022-03-25
Documentation: Application (new chapter)

adpaco-aws:
### Description of changes: 

Adds a new "Application" chapter that includes the "tool comparison" and "where to start on real code" section. I think there is enough material there and, in the future, the section can host an article on integration or complex proof debugging.

Note that some stuff has been moved around, but other than that I only made minor changes to the already-existing articles.

### Resolved issues:

Part of #700 

### Testing:

### Checklist


docs/src/tutorial-real-code.md
```diff
@@ -36,7 +36,7 @@ Dependencies can sometimes blow up the tractability of proofs.
 This can usually be handled, but requires a lot more investment to make it happen, and so isn't a good place to start.
 
 2. Don't forget to consider starting with your dependencies.
-Sometimes the best place to start won't be your code, but in code you depend on.
+Sometimes the best place to start won't be your code, but the code you depend on.
```
jaisnan, line 39: Imo, "on the code that you depend upon" would be a better to phrase this.

docs/src/tutorial-real-code.md
```diff
@@ -47,14 +47,13 @@ Here are some things to avoid, when starting out:
...
-Much like needing to put bounds on loops, these can be hard to model since you needs to put bounds on their size, too.
+2. Inductive data structures.
+These are data structures with unbounded size (e.g., linked lists or trees.)
+These can be hard to model since you need to set bounds on their size, similar to what happens with loops.
 
-3. I/O code.
-Kani doesn't model I/O, so if you're depending on behavior like reading/writing to a file, you won't be able to prove anything.
-This is one obvious area where testability helps provability: often we separate I/O and "pure" computation into different functions, so we can unit test the latter.
+3. Input/Output code.
+Kani doesn't model I/O, so if you depend on behavior like reading/writing to a file, you won't be able to prove anything.
```
jaisnan, line 55: This could be - "if your code depends on behavior".

docs/src/tutorial-real-code.md
```diff
@@ -81,18 +80,5 @@ Running Kani on this simple starting point will help figure out:
...
-You accomplish this by adding new assertions.
-These are not necessarily assertions just in your proof harness: consider also adding new assertions to the code being run.
-These count too!
+You accomplish this by adding new assertions (not just in your harness, but also to the code being run).
 Even if a proof harness has no post-conditions being asserted directly, the assertions encountered along the way can be meaningful proof results by themselves.
-
-
-## Summary
-
-In this section:
```
jaisnan, line 92: Maybe we should keep a summary?

docs/src/tutorial-real-code.md
adpaco-aws: Updated as "the code that you depend on", because in your suggestion:
 * I think adding "on" changes meaning (the sentence would read "the best place to start will be on the code..." if we remove the "won't be your code, but") because it refers to a more specific location. This is not what we want to express.
 * "on"/"upon" are interchangeable, so I would rather keep "on".

docs/src/tutorial-real-code.md
adpaco-aws: Some sections had it while others did not, so during #700 I removed all of them. I don't think they add too much at the section level, at least with the content we currently have.

jaisnan (approved):
Thanks! Apologies for the late approval, I thought I had approved this already

PR #999, opened 2022-03-30
Pass config from cargo-kani to kani-compiler

tedinski:
### Description of changes: 

This change is a bit of a mash-up of things I need to start making things releasable, mostly centering around how we find certain environment/config info in kani-compiler.

In roughly github diff display order you will see:

1. Addition of the `home` crate to compute `RUSTUP_HOME` the same way rustup does.
2. **The addition of passing `--sysroot` and `--kani-lib` arguments to `kani-compiler` from `cargo-kani`**
3. (To keep things accurate, this required switching that function to accumulating a list of `OsString` now, instead of `String`, now that' we're including Paths.)
4. A smattering of clippy warnings that I fixed.
5. Fixing how we construct a lot of `PathBuf`s to get cleaner code.
6. **The addition of a new RPATH for `kani-compiler`, corresponding to where it should search for the rustc libraries in a release bundle.** (Basically: next to `bin/kani-compiler` will be a symlink `toolchain` to the rustup-installed toolchain we require. Hence, `../toolchain/lib`)
7. **I removed the `/deps` path being provided to the linker with `-L`. This was how it used to work, but this directory didn't exist anymore.**
8. SYSROOT accidentally didn't take an argument!

### Call-outs:

1. There is no clean way currently to figure out where the kani libraries are being built, so I'm unfortunately not passing that down in development builds. The path is currently something like `/home/ubuntu/rmc/target/debug/build/kani-compiler-b8c35844f7600bbf/out/lib` and I wish we could change that outdir to something more predicable, so we could make this unconditional, but I couldn't figure out how.
2. Since we aren't yet using the RPATH for a release build, I manually verified that the executable ends up with the correct config:

```
RUNPATH              /home/ubuntu/.rustup/toolchains/nightly-2022-03-09-x86_64-unknown-linux-gnu/lib:$ORIGIN/../toolchain/lib
```

### Testing:

* Is this a refactor change? not quite

### Checklist


src/cargo-kani/src/session.rs
```diff
@@ -222,4 +223,27 @@ impl InstallType {
             }
         }
     }
+
+    pub fn rust_toolchain(&self) -> Result<PathBuf> {
+        // This is a compile-time constant, not a dynamic lookup at runtime,
+        // so we get the toolchain we're built with (i.e. that from rust_toolchain.toml)
+        let toolchain = env!("RUSTUP_TOOLCHAIN");
```
celinval, line 230: This is a tricky one, eh? We should be careful with custom toolchain override and users setting up this variable in the first place. That said, as long as we automate our releases to not do that, we should be OK.

src/kani-compiler/src/parser.rs
```diff
@@ -90,10 +90,16 @@ pub fn parser<'a, 'b>() -> App<'a, 'b> {
                 .long("--restrict-vtable-fn-ptrs")
                 .help("Restrict the targets of virtual table function pointer calls."),
         )
-        .arg(Arg::with_name(SYSROOT).long("--sysroot").help("Override the system root.").long_help(
-            "The \"sysroot\" is the location where Kani will look for the Rust \
+        .arg(
+            Arg::with_name(SYSROOT)
+                .long("--sysroot")
+                .takes_value(true)
```
celinval, line 96: oops. :)

celinval (changes requested):
It looks good, but I think we should be able to just use relative paths everywhere. When we run `cargo build`, cargo will put all workspace artifacts in the same directory, doesn't it? So for dev builds you should be able to use executable directory.

 I think we are building the libraries twice right now, which isn't ideal.

tedinski:
> It looks good, but I think we should be able to just use relative paths everywhere. When we run `cargo build`, cargo will put all workspace artifacts in the same directory, doesn't it? So for dev builds you should be able to use executable directory.
> 
> I think we are building the libraries twice right now, which isn't ideal.

Are you just talking about `--kani-lib`? (I think I forgot to write this up as part of the callout, oops)

I agree it's not ideal. I avoided changing anything in this PR because I wasn't sure what to do. The issue with using the regular build directory is that it has a lot of other stuff in the directory (other dependencies, etc), whereas using the `lib` outputs we currently use we end up with just the 3 files only: the 2 `rlib` and the 1 `so`. (kani, std, and kani_macros, respectively.)

If we use the regular output directory I was slightly worried we might end up with those other dependencies mattering, somehow. On the other hand, this only matters for development builds, since release will distribute only the relevant 3 files. If you think it's probably fine, I'll switch over to doing that.

tedinski:
Ah, actually the bigger problem is that they aren't built with `--cfg=kani` and so don't behave correctly when we build them "normally"

celinval:
> Ah, actually the bigger problem is that they aren't built with `--cfg=kani` and so don't behave correctly when we build them "normally"

What about adding a cargo config file that passes --cfg=kani to the build? That said, I'm OK doing that in a separate PR.

celinval (approved):
I am OK to push further improvements to a separate PR. Can you please move the comment out of [src/cargo-kani/build.rs](https://github.com/model-checking/kani/pull/999/files/81efec0fe7cbfd3d55dc817e457b6081571378b2#diff-00d3173b180b71c3febe3ea0eb27a7c0d8c745f19cb442814a90323fa7304197) before pushing the change though?

PR #1011, opened 2022-04-04
Use dynamic allocation for non-deterministic slices

### LINKED ISSUE #1009: AnySlice may yield unexpected results
I tried this code:
```rust
#[kani::proof]
fn check_out_of_bounds() {
    let mut bytes = kani::slice::any_slice::<u8, 5>();
    let val = unsafe { *bytes.get_slice().as_ptr().add(4) };
    kani::assume(val != 0);
    assert_ne!(val, 0);
}
```

```
kani check.rs
```

I expected to see this happen: I expected that accessing position `4` of any_slice may fail with out of bounds access.

Instead, this happened: The verification succeeded. Because any_slice always allocates a buffer with size MAX_SIZE (in this case 5), the access to position `4` is always valid.

zhassan-aws:
### Description of changes: 

Kani's API for non-deterministic slices (`NonDetSlice`) has unsoundness as it may hide failures due to out-of-bound accesses. This is due to *statically* allocating an array with the specified maximum length and then returning a slice of a nondet length of that. Accessing an element beyond the slice length should be a failure, but if the element is within the array length, no failure is reported.

### Resolved issues:

Resolves #1009 

### Testing:

### Checklist


celinval (commented):
Sweet! I hope that doesn't impact performance.

One quick thing, since you are changing this struct, what do you think about renaming it too? Maybe `AnySlice`?

celinval:
One more thing, I believe all these methods should either be marked as unsafe or T should implement Arbitrary type.

zhassan-aws:
While trying to check the impact of this change on performance using one of the s2n-quic proofs, I ran into a CBMC invariant violation:
```
Invariant check failed
File: ../src/util/simplify_expr.cpp:2642 function: simplify_rec
Condition: as_const(tmp).type() == expr.type()
Reason: Postcondition
Backtrace:
```
I confirmed that without this change, verification passes on that particular proof. This is a blocker for this PR until we figure out a fix.

library/kani/src/slice.rs
```diff
@@ -43,45 +44,57 @@ fn any_range<const LENGTH: usize>() -> (usize, usize) {
...
-impl<T, const MAX_SLICE_LENGTH: usize> DerefMut for NonDetSlice<T, MAX_SLICE_LENGTH> {
+impl<T, const MAX_SLICE_LENGTH: usize> DerefMut for AnySlice<T, MAX_SLICE_LENGTH> {
     fn deref_mut(&mut self) -> &mut Self::Target {
         self.get_slice_mut()
     }
 }
 
-pub fn any_slice<T, const MAX_SLICE_LENGTH: usize>() -> NonDetSlice<T, MAX_SLICE_LENGTH> {
-    NonDetSlice::<T, MAX_SLICE_LENGTH>::new()
+pub fn any_slice<T, const MAX_SLICE_LENGTH: usize>() -> AnySlice<T, MAX_SLICE_LENGTH> {
```
celinval, line 98: Just wanna make sure you've seen my comments about making this method unsafe or using the Arbitrary trait.

celinval (commented):
It looks much better. Thanks for doing this.

Have you been able to narrow down the problem to see if it's an issue on our side or CBMC?

library/kani/src/slice.rs
zhassan-aws: Yes, I have seen it, but I haven't addressed it yet. I think we have discussed this before, but I don't recall what we concluded. What I'm thinking right now is to perhaps add `Invariant` as a trait bound for `T` in `AnySlice` and set every element in the slice to `kani::any()`. We can also provide an unsafe version, e.g. `AnyRawSlice` that does not require `Invariant` for `T`. What do you think?

camshaft:
Nice catch! Do you know how much slower the dynamic version is?

zhassan-aws:
> Nice catch! Do you know how much slower the dynamic version is?

I'm in the process of evaluating the impact of the change on performance. I don't have numbers yet because I ran into a CBMC crash while doing so.

library/kani/src/slice.rs
zhassan-aws: As discussed offline, changed `T` to require `Arbitrary`, and added an unsafe `any_raw_slice` version that doesn't require `Arbitrary`.

tests/expected/nondet-slice/main.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// This test checks that Kani reports out-of-bound accesses on a non-det slice
+// created using kani::slice::any_slice
+
+#[kani::proof]
+fn check_out_of_bounds() {
```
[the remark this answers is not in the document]
zhassan-aws: Done.

library/kani/src/slice.rs
```diff
@@ -47,24 +51,37 @@ fn any_range<const LENGTH: usize>() -> (usize, usize) {
 /// foo(&slice); // where foo is a function that takes a slice and verifies a property about it
 /// ```
 pub struct NonDetSlice<T, const MAX_SLICE_LENGTH: usize> {
-    arr: [T; MAX_SLICE_LENGTH],
+    layout: Layout,
+    ptr: *mut u8,
```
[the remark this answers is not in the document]
zhassan-aws: Good idea. This removed the need to have `PhantomData`.

zhassan-aws:
Performance does not seem to degrade with the switch to using dynamic allocation. In fact, on one of the s2n-quic proofs (`vectored_copy`), it is slightly faster with dynamic allocation:
Current:
```
SUMMARY:
 ** 0 of 820 failed

VERIFICATION:- SUCCESSFUL


        Command being timed: "cargo kani --function check_vectored_copy_u8 --unwind 5"
        User time (seconds): 1149.01
        Maximum resident set size (kbytes): 3733332
```
New:
```
SUMMARY:
 ** 0 of 910 failed

VERIFICATION:- SUCCESSFUL


        Command being timed: "cargo kani --function check_vectored_copy_u8 --unwind 5"
        User time (seconds): 1023.27
        Maximum resident set size (kbytes): 5487408
```
Memory usage seems to go up quite a bit though (3.7 GB to 5.5 GB).

library/kani/src/slice.rs
```diff
@@ -38,50 +39,105 @@ fn any_range<const LENGTH: usize>() -> (usize, usize) {
...
     pub fn get_slice_mut(&mut self) -> &mut [T] {
-        &mut self.arr[..self.slice_len]
+        unsafe { std::slice::from_raw_parts_mut(self.ptr, self.slice_len) }
     }
 }
 
-impl<T, const MAX_SLICE_LENGTH: usize> Deref for NonDetSlice<T, MAX_SLICE_LENGTH> {
+impl<T, const MAX_SLICE_LENGTH: usize> Drop for AnySlice<T, MAX_SLICE_LENGTH> {
+    fn drop(&mut self) {
+        if self.slice_len > 0 {
```
celinval, line 112: I'm curious to know what happens if we remove this `if`. It's a bit counter intuitive that we are still calling `alloc` when `slice_len = 0` but we don't call `dealloc`. I would expect some symmetry in those calls.

zhassan-aws: Removing the `if` produces a failure of the form:
```
Check 35: std::alloc::dealloc.pointer_primitives.4
         - Status: FAILURE
         - Description: "pointer outside object bounds"
         - Location: ~/.rustup/toolchains/nightly-2022-03-23-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:105:14 in function std::alloc::dealloc
```
You're right about the asymmetry not looking correct: calling `alloc` with a zero size is actually undefined behavior (https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html#safety-1). I'll see if I can fix that. Might need to store a null pointer in the case of a zero slice!

danielsn: Doesn't rust normally prefer to use a non-null but invalid value

zhassan-aws: Indeed! `NonNull::dangling()` is the recommended approach on https://doc.rust-lang.org/core/slice/fn.from_raw_parts.html#safety.

zhassan-aws: I added a commit that explicitly handles the `slice_len == 0` case via a `ptr::null()`.

celinval (approved):
Looks good! Thanks

PR #1017, opened 2022-04-06
A script to create a Kani release bundle.

### LINKED ISSUE #939: Script to build release bundle
We should write a script to create the kani release bundle, this should probably include:

* cargo-kani (prebuilt rust binary)
* kani-compiler (prebuilt rust binary)
* cbmc (binary)
* cbmc-viewer (python 3 wheel)
* kani artifacts (this list is not yet complete?)
    * scripts/cbmc_json_parser.py
        * Includes additional python dependency:
        * colorama
* a license note file
* library/kani
* library/kani_macros
* library/std
* (pre-built versions of all the above?)

Notable dependencies this does not include:

* Our specific Rust toolchain
* Python 3

tedinski:
### Description of changes: 

This is an initial script to create a Kani release bundle. It's written it Rust instead of Bash because why not, it's easier to make certain things more robust that way.

### Resolved issues:

Resolves #939 

### Call-outs:

1. We're not bundling `cbmc-viewer` because there are issues with potential portability between Python version on the customer machine. (We'll `pip install` this during first-time setup.)
2. We're pulling the CBMC binaries from PATH, instead of from a specific pinned CBMC version. This is to avoid putting CBMC versions in more places. There's a big comment about it. We'll invert this in a later refactoring/PR.
3. This isn't currently tested really - I'm making this PR now, and we'll test for real it when I follow up with the next PR which will create a "release" `InstallType` in `cargo-kani` and thus allow everything (except viewer, next step there) to run out of a release bundle instead of out of a dev repo checkout.

### Testing:

* How is this change tested? manually for now, follow-up PR will add to CI

### Checklist


tools/make-kani-release/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+fn main() -> Result<()> {
+    let version_string = parse_args()?;
+    let kani_string = format!("kani-{}", version_string);
+    let bundle_name = format!("{}-{}.tar.gz", kani_string, env!("TARGET"));
+
+    if !Path::new("src/kani-compiler").exists() {
+        bail!("Run from project root directory. Couldn't find 'src/kani-compiler'.");
+    }
+
+    let dir = Path::new(&kani_string);
```
celinval, line 23: Can you make the installation path customizable?

tools/make-kani-release/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+    if dir.exists() {
```
celinval, line 24: Maybe add an `--override` option to make testing easier.

tools/make-kani-release/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+/// Copy Kani files into `dir`
+fn bundle_kani(dir: &Path) -> Result<()> {
+    let bin = dir.join("bin");
+    std::fs::create_dir(&bin)?;
+
+    // 1. Kani binaries
+    Command::new("cargo").args(&["build", "--release"]).run()?;
+
+    let release = Path::new("./target/release");
+    cp(&release.join("cargo-kani"), &bin)?;
```
celinval, line 64: What about standalone kani?

celinval (changes requested):
Can you please add some test coverage in our CI? Ideally we should be able to run our full regression using the bundled `kani` but for now, we can just have some sanity checks to ensure that the bundled `kani` and `cargo-kani` works. 

On a different note, I was also wondering if we should make the an `cbmc-viewer` an optional dependency.

tools/make-kani-release/src/main.rs
tedinski: This is just building the release bundle. The installation path is anything (unpacked anywhere), nothing to configure here, so I'm not sure what you mean?

tools/make-kani-release/src/main.rs
tedinski: That's part of the shim/proxy, not the release bundle. Both invoke this `cargo-kani` binary just in different modes.

tools/make-kani-release/src/main.rs
tedinski: I decided against for now. I've used this sort of release script a lot in the past and testing (running into problems during bundling) is pretty rare actually.

Note that this happily overwrites the `tar.gz` file. It's just the directory that only gets left existing if the bundling process crashes for some reason. So if you're trying to create the bundle over and over again (e.g. repeatedly running `cargo run -p make-kani-release -- latest`, that works fine.

tedinski:
> Can you please add some test coverage in our CI? Ideally we should be able to run our full regression using the bundled `kani` but for now, we can just have some sanity checks to ensure that the bundled `kani` and `cargo-kani` works.

So I'm working on the next step which is a release `InstallType`, and as part of that I've built a Dockerfile to (for now manually) do the first-time setup stuff. I could add that to CI and get it to run the tests.

I was planning on that being a separate PR after this one was merged. Would you prefer that I push it into this PR?

> On a different note, I was also wondering if we should make the an `cbmc-viewer` an optional dependency.

I don't think so, it's such a small thing. There are complications with Python, but we have those to run our script (which needs colorama anyway) too. If we ported the renderer to Rust later, then maybe there'd be a purpose, but only a tiny one: anyone who didn't want a Python installation around could skip it.

tedinski:
> get it to run the tests.

Ugh, compiletest has a lot of baked assumptions about running in a repo, etc. Running our existing tests on a release bundle is going to be a lot of work.

tools/make-kani-release/src/main.rs
celinval: I mean, generate the files somewhere different. For debug purpose, it might be nice to generate the bundle in a directory that is not the root folder of our repository.

tools/make-kani-release/src/main.rs
celinval: I thought you had mentioned the idea of creating two different binaries. Did you give up on that idea?

tedinski: Right, that's the shim/proxy that calls this binary.

celinval:
I think we should add a minimum test coverage to this PR. We should at least do the following:
1. Run `kani-make-release`.
2. Unpack the contents of the generated tar ball.
3. Run a dummy `cargo kani` test.

tedinski:
> 3\. Run a dummy `cargo kani` test.

Sure, to do this, I'll need the `InstallType` stuff, so I'll fold that into this PR then.

tedinski:
I'm not super happy with putting the Dockerfile at the root, but moving it elsewhere comes with issues w.r.t. copying `tests` into it or whatever.

I could maybe rename it to `Dockerfile.internal_release_test` or something if you think that'd be an improvement?

adpaco-aws:
> I'm not super happy with putting the Dockerfile at the root, but moving it elsewhere comes with issues w.r.t. copying `tests` into it or whatever.

Can't you just move/symlink it as part of the workflow?

tedinski:
> Can't you just copy/symlink it as part of the workflow?

A reasonable idea. Keep it under something like `tests/releasebundle/` maybe?

adpaco-aws:
I was thinking `scripts/ci/` because it's CI-related. Maybe create a "release" (or any name) folder there?
Would like to avoid putting anything that is not a test under `tests/` :/

scripts/ci/Dockerfile.release-bundle-test
```diff
@@ -0,0 +1,33 @@
...
+FROM ubuntu:20.04
+ENV DEBIAN_FRONTEND=noninteractive \
+    DEBCONF_NONINTERACTIVE_SEEN=true
+RUN apt-get update && \
+    apt-get install -y python3 python3-pip curl ctags && \
+    curl -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+# This section is roughly (+tests) what "first time setup" will do:
+RUN rustup toolchain install nightly-2022-03-23
```
celinval, line 15: Do we need to hardcode the path here? Can we include the toolchain file in the bundle and use that instead?

src/cargo-kani/src/session.rs
```diff
@@ -172,56 +178,27 @@ impl InstallType {
...
-            Self::DevRepo(repo) => {
-                let path = repo.join("library/kani/kani_lib.c");
-                if path.as_path().exists() {
-                    Ok(path)
-                } else {
-                    bail!("Unable to find kani_lib.c. Looked for {}", path.display());
-                }
-            }
-        }
+        self.base_path_with("library/kani/kani_lib.c")
```
celinval, line 193: Nice

tools/make-kani-release/src/main.rs
celinval: Not a blocker, but we should eventually do this.

For now, can you please ensure that our `.gitignore` will ignore the `tar.gz` file and the folder you are creating here.

scripts/ci/Dockerfile.release-bundle-test
celinval: Or I guess we can just copy the file here. That would work too.

tedinski: The next followup PR (creating the `kani-verifier` crate with proxy binaries) will solve most of the hard-coding in this Dockerfile, because it will become something like `cargo-kani setup --use-local-bundle kani-etc.tar.gz` instead of doing all this manually.

celinval: OK... we can live with this for a couple of days.

celinval (approved):
Please update the `.gitignore` before pushing this change.

tedinski:
I've updated gitignore to include the `tar.gz` files. I've also factored out "prebundle checks" into a separate function that:

1. Ensures we've build Kani in release mode *before* doing anything.
2. Checks for cbmc, again before we ever start bundling.

This should reduce the odds of failing during a bundle creation (and leaving the directory still in existence) to close to nothing, hopefully alleviating your concern?

celinval:
Sure. We can change it later if this becomes a problem. Thanks

PR #1018, opened 2022-04-06
Minor updates in README for readability.

jaisnan:
### Description of changes: 

The current README has some text and code that's difficult to read. This change adds some spacing and some capitalizing.

Before - 

image

After - 

image

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@ The Kani Rust Verifier is a bit-precise model checker for Rust.
 
 Kani verifies:
  * Memory Safety -- even in unsafe Rust code
- * User specified assertions
+ * User-specified assertions
  * Absence of panics
- * Checks for some classes of undefined behavior
+ * Absence of some classes of undefined behavior
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -3,10 +3,10 @@
 The Kani Rust Verifier is a bit-precise model checker for Rust.
 
 Kani verifies:
- * Memory Safety -- even in unsafe rust code.
- * User specified assertions.
- * Absence of panics.
- * Checks for some classes of undefined behavior.
+ * Memory Safety -- even in unsafe Rust code
+ * User specified assertions
+ * Absence of panics
+ * Checks for some classes of undefined behavior
 
 ## Installation
 
@@ -21,7 +21,6 @@ use my_crate::{function_under_test, is_valid, meets_specification};
 
 #[kani::proof]
 fn check_my_property() {
-
    // Create a nondeterministic input
    let input = kani::any();
 
--- a/README.md
+++ b/README.md
@@ -3,10 +3,10 @@
 The Kani Rust Verifier is a bit-precise model checker for Rust.
 
 Kani verifies:
- * memory safety -- even in unsafe Rust code
- * user-specified assertions
- * absence of panics
- * checks for some classes of undefined behavior
+ * Memory Safety -- Even in unsafe Rust code.
+ * User specified Assertions.
+ * Absence of Panics.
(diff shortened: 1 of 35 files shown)


README.md
```diff
@@ -3,10 +3,10 @@
...
 
 Kani verifies:
- * memory safety -- even in unsafe Rust code
- * user-specified assertions
- * absence of panics
- * checks for some classes of undefined behavior
+ * Memory Safety -- Even in unsafe Rust code.
+ * User specified Assertions.
+ * Absence of Panics.
+ * Checks for some classes of Undefined Behavior.
```
tedinski, line 9: Capitalizing first letters seems fine, but the rest of these appear to be random capitalization. (People frequently have a tendency to capitalize "meaningful words" but this isn't usually an improvement and is an impulse it's best to try to get over. I completely get where you're coming from with e.g. capitalizing UB, but it's better just writing the words normally. :P)

jaisnan: haha, sure. I'll undo those.

README.md
```diff
@@ -3,10 +3,10 @@
...
 
 Kani verifies:
- * memory safety -- even in unsafe Rust code
- * user-specified assertions
- * absence of panics
- * checks for some classes of undefined behavior
+ * Memory Safety -- even in unsafe rust code.
+ * User specified assertions.
+ * Absence of panics.
+ * Checks for some classes of undefined behavior.
```
adpaco-aws, line 9: In case it's helpful, the rule I followed to write this is the second case in https://www.dailywritingtips.com/bulleted-lists/

I'm okay with first letters being capitalized, but I don't think we should keep the dots. Also, Rust is a name.

README.md
```diff
@@ -21,12 +21,16 @@ use my_crate::{function_under_test, is_valid, meets_specification};
 
 #[kani::proof]
 fn check_my_property() {
+
```
adpaco-aws, line 24: I agree lines between these help, but I would remove this first one (most of the time, programs have a comment or declaration here).

README.md
jaisnan: Thanks for the link, I'll use it here on. I think keeping the first letters capitalized makes it easier to read. I'll remove the dots add Rust as a name.

README.md
```diff
@@ -3,10 +3,10 @@
 The Kani Rust Verifier is a bit-precise model checker for Rust.
 
 Kani verifies:
- * memory safety -- even in unsafe Rust code
- * user-specified assertions
- * absence of panics
- * checks for some classes of undefined behavior
+ * Memory Safety -- even in unsafe Rust code
+ * User specified assertions
```
adpaco-aws, line 7:
```suggestion
 * User-specified assertions
```

README.md
```diff
@@ -3,10 +3,10 @@
...
+ * Absence of panics
+ * Checks for some classes of undefined behavior
```
zhassan-aws, line 9: We should also change this to "Absence of some classes of undefined behavior", since the sentence is "Kani verifies ..." which does not fit with "Check for some classes ...".

adpaco-aws (approved):
Thanks! Just a minor correction.

README.md
jaisnan: @tedinski @adpaco-aws thoughts?

I personally agree with the suggestion.

adpaco-aws: Actually, the way I understood this sentence is that Kani verifies the checks for those classes (here "checks" could be a noun or verb). But it may be better to rephrase just in case.

jaisnan: Yeah, that could be a bit confusing to first time readers

PR #1021, opened 2022-04-07
Add a test for `discriminant_value`

adpaco-aws:
### Description of changes: 

Adds a test for `discriminant_value`. The implementation works as expected for all covered cases, so this completes the audit for the intrinsic.

### Resolved issues:

Part of #727 

### Testing:

### Checklist

### DIFF
--- a/tests/kani/Intrinsics/discriminant_value.rs
+++ b/tests/kani/Intrinsics/discriminant_value.rs
@@ -1,0 +1,50 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Check that `discriminant_value` returns the expected results
+// for different cases
+#![feature(core_intrinsics)]
+use std::intrinsics::discriminant_value;
+
+// A standard enum with variants containing fields
+enum MyError {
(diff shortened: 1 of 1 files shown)


tests/kani/Intrinsics/discriminant_value.rs
```diff
@@ -0,0 +1,50 @@
...
+
+// An enum that assigns constant values (one of them negative) to all variants
+enum Ordering {
+    Less = -1,
+    Equal = 0,
+    Greater = 1,
+}
+
+#[kani::proof]
+fn main() {
```
celinval, line 31: Can you please split this into 3 different harnesses? I think it will make this much easier to read.

celinval (approved):
Nice coverage! Please address the comment below before pushing.

tests/kani/Intrinsics/discriminant_value.rs
adpaco-aws: Thanks for the suggestion, I didn't think about that! It's now split into 4 harnesses, can you take another look?

celinval (approved):
Looks good to me. Maybe next time use `///` to comment the harness instead of the function body. But that is just a test so no need to bother. Thanks!

PR #1022, opened 2022-04-07
Add a test for `breakpoint`

adpaco-aws:
### Description of changes: 

Adds a test for `breakpoint`. It checks that is supported (`breakpoint` generates a `SKIP` statement)
and that it does not affect the verification result by adding an `assert!(true)` after it.

If you have ideas on how to test "skip" intrinsics, please let me know.

### Resolved issues:

Part of #727 

### Testing:

### Checklist

### DIFF
--- a/tests/kani/Intrinsics/breakpoint.rs
+++ b/tests/kani/Intrinsics/breakpoint.rs
@@ -1,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Check that `breakpoint` is supported (generates a `SKIP` statement)
+// and that it does not affect the verification result
+#![feature(core_intrinsics)]
+
+#[kani::proof]
+fn main() {
+    unsafe { std::intrinsics::breakpoint() };
+    assert!(true);
+}


tests/kani/Intrinsics/breakpoint.rs
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Check that `breakpoint` is supported (generates a `SKIP` statement)
+// and that it does not affect the verification result
+#![feature(core_intrinsics)]
+
+#[kani::proof]
+fn main() {
+    unsafe { std::intrinsics::breakpoint() };
+    assert!(true);
```
celinval, line 11: Can you add something more meaningful and maybe move this to the expected test suite to ensure it is reachable?

If I'm  not wrong, `assert!(true)` doesn't even get codegen. So this test would still succeed even if the breakpoint was translated to a return statement.

celinval (changes requested):
Can you please make this test a bit stronger?

tests/kani/Intrinsics/breakpoint.rs
adpaco-aws: `assert!(true)` does codegen, and in fact produces the only check:
```
RESULTS:
Check 1: main.assertion.1
         - Status: SUCCESS
         - Description: "assertion failed: true"
         - Location: breakpoint.rs:11:5 in function main
```

I'm okay with moving this into `expected`, but I'm not sure how that ensures it is reachable if I remove the assertion.

celinval: Oh, this works now because we override `assert!()`. Before that, this code would be pruned by the compiler. Good to know! :)

In this case, you can just check that this assertion is reachable.

adpaco-aws: Converted to an `expected` test. Both approaches seem equivalent to me (if the assertion were not reachable, verification would fail), but if anything goes wrong I'm sure `expected` will complain before 😄

zhassan-aws: Actually, verification succeeds if there are no failures (even if a check is unreachable).

celinval (approved):
Thanks!

Issue #934, opened 2022-03-15
Signed integer versions of bit-manipulation intrinsics cannot be throughly tested


adpaco-aws:
Signed versions of bit-manipulation intrinsics cannot be tested as thoroughly as the unsigned versions. This is due to the signed to unsigned conversion overflows which we cannot avoid due to their definition (see #778 for more details, and #880 #883 for other examples).

PR #1023, opened 2022-04-07
Add a test for `bswap`

adpaco-aws:
### Description of changes: 

Adds a test for `bswap`, which relies on CBMC's implementation. As in other bit-manipulation intrinsics, it suffers from #934 

### Resolved issues:

Part of #727 

### Testing:

### Checklist


tests/kani/Intrinsics/bswap.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Check that `bswap` returns the expected results
```
celinval, line 4: Can you please state here what that result is? `bswap` is not the most intuitive name. It could be bit / byte swap. I had to google to check exactly what this does.

tests/kani/Intrinsics/bswap.rs
```diff
@@ -0,0 +1,45 @@
...
+            assert!(a_byte == b_byte);
+        }
+
+        let x: $ty = kani::any();
+        $check_name(x, std::intrinsics::bswap(x));
+    }
+}
+
+#[kani::proof]
+fn main() {
```
celinval, line 35: Can you split these into multiple harnesses again?

tests/kani/Intrinsics/bswap.rs
```diff
@@ -0,0 +1,45 @@
...
+    ($ty:ty, $check_name:ident, $get_byte_name:ident) => {
+        // Gets the i-th byte in `x`
+        fn $get_byte_name(x: $ty, i: usize) -> $ty {
+                let mask = 0xFF as $ty << i * BITS_PER_BYTE;
+                let masked_res = x & mask;
+                let bytes_res = masked_res >> i * BITS_PER_BYTE;
+                bytes_res
+        }
+
+        // Checks that the original byte is equal to the swapped byte
```
celinval, line 19: Maybe change this to say checks that the order of the bytes in the original integer is reverse in the swapped one.

tests/kani/Intrinsics/bswap.rs
```diff
@@ -0,0 +1,45 @@
...
+        fn $check_name(a: $ty, b: $ty) {
+            let bytes = std::mem::size_of::<$ty>();
+            let i: usize = kani::any();
```
celinval, line 22: I like how you used any here. Much cleaner than an iterator of some sort.

adpaco-aws:
Extended the description, added an example and link to the intrinsic documentation. Also split into several proofs (one per macro call), so all comments should be addressed now.

tests/kani/Intrinsics/bswap.rs
adpaco-aws: And more efficient too!

celinval (approved):
Thanks!

danielsn (approved):

PR #1024, opened 2022-04-07
Refactor CI action

tedinski:
### Description of changes: 

This separates "setup" from "build", in preparation for a new workflow that will test out the release build/bundle. This should be a pure refactor in the workflow, no behavior changes.

### Call-outs:

1. Oof, only just noticed the python version requirement. Will investigate/ask if this can be fixed...

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


adpaco-aws (approved):
We should check if it's possible to use a lower Python version in the future.

Issue #1025, opened 2022-04-08
Rounding intrinsics are not working


adpaco-aws:
Calling rounding intrinsics (and likely other math intrinsics) results in an "assertion false". For example, the following code:

```rust
fn main() {
    let x: f32 = kani::any();
    let x_ceil = unsafe { std::intrinsics::ceilf32(x) };
}
```
returns this:

```
RESULTS:
Check 1: ceilf.assertion.1
         - Status: FAILURE
         - Description: "assertion false"
         - Location: <builtin-library-ceilf> in function ceilf

Check 2: <T as kani::Arbitrary>::any.assertion.1
         - Status: SUCCESS
         - Description: "resume instruction"
         - Location: ~/kani/library/kani/src/arbitrary.rs:17:5 in function <T as kani::Arbitrary>::any


SUMMARY: 
 ** 1 of 2 failed
Failed Checks: assertion false

VERIFICATION:- FAILED
```

It's still not clear what intrinsics are affected, but at least rounding intrinsics (ceil, floor, round) are not working. This is probably due because the irep expressions we generate for them are not exposed in CBMC, or because they are not the right ones.

tautschnig:
The problem appears to be in how `goto-instrument` invokes the C parser vs `cbmc` as I'm seeing the following error when running the individual steps:
```goto-instrument --add-library ceil.out.for-main ceil.out.for-main.with-library
Reading GOTO program from 'ceil.out.for-main'
Adding CPROVER library (x86_64)
file /usr/include/x86_64-linux-gnu/bits/mathcalls-helper-functions.h line 21: syntax error before '__value'
Writing GOTO program to 'ceil.out.for-main.with-library'
```
I'll try to figure out what is going on here.

tautschnig:
https://github.com/diffblue/cbmc/pull/6789 fixes the problem.

tautschnig:
The CBMC PR is now merged, and will be part of the next release (due 2022-05-26).

adpaco-aws:
@tautschnig can you check the "fixme" test case in #1248 ? `fmaxf` continues to behave as if it was missing, but `fmax` and `fminf` for example aren't. I've checked and we don't do anything different for `fmaxf`, so my guess is that CBMC is failing to pick it up for some reason.

tautschnig:
@adpaco-aws With apologies for this weird bug: https://github.com/diffblue/cbmc/pull/6904 now fixes the issue.

adpaco-aws:
[...]

PR #1026, opened 2022-04-08
Disable rounding intrinsics

adpaco-aws:
### Description of changes: 

Disables rounding intrinsics which appear to be replaced by assert-false statements for some reason (debugging will come later). We could leave them "as is", but in my opinion we should control how they behave for now and only enable them when they are fixed and tested.

### Resolved issues:

Part of #727 

### Testing:

* How is this change tested? Adds 6 negative tests.

### Checklist

### DIFF
--- a/docs/src/rust-feature-support.md
+++ b/docs/src/rust-feature-support.md
@@ -309,8 +309,8 @@ bitreverse | Yes | |
 breakpoint | Yes | |
 bswap | Yes | |
 caller_location | No | |
-ceilf32 | Yes | |
-ceilf64 | Yes | |
+ceilf32 | No | |
+ceilf64 | No | |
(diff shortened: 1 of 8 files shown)


zhassan-aws (approved):
Disabling them until they're fixed sounds like the right thing to me. Should we replace their implementations with:
```
codegen_unimplemented_intrinsic!(
                "https://github.com/model-checking/kani/issues/1025"
```
though?

adpaco-aws:
> Should we replace their implementations with:
> 
> ```
> codegen_unimplemented_intrinsic!(
>                 "https://github.com/model-checking/kani/issues/1025"
> ```
> 
> though?

Yes, thanks! It's done.

PR #1031, opened 2022-04-11
Feature : Unwind attribute

### LINKED ISSUE #1027: Enable Unwind Attribute from Parsing metadata files
Requested feature:

We are writing the unwind values into the metadata files. We should now read form these metadata files to use them as attributes fully.

This is a follow up of - https://github.com/model-checking/kani/pull/846.

Use case:

Enable the unwind attribute.

Test case:

```rust

#[kani::proof]
#[kani::unwind(10)]
fn harness() {
    let mut counter = 0;
    loop {
        counter += 1;
        assert!(counter < 10);
    }
}

```

jaisnan:
### Description of changes: 

Cargo-kani and kani do not have the unwind attribute enabled as a feature, this PR enables kani and cargo-kani to read from the metadata and use the unwind attribute.

Tests explicitly pass the unwind value in the flags. with these changes, we can just use the attribute over the harnesses.

### Resolved issues:

Resolves #1027

### Testing:

* How is this change tested? Tests changed to use the unwind attribute

### Checklist


celinval (commented):
I have a few questions.

1. Will the command line option `--unwind` and the property file `unwind` flag still work?
2. If so, which one takes precedence?

We should make sure we add tests that cover those cases.

src/cargo-kani/src/metadata.rs
```diff
@@ -138,6 +139,14 @@ impl KaniSession {
             Ok(metadata.proof_harnesses.clone())
         }
     }
+
+    /// Deserialize a *.restrictions.json file
+    fn read_kani_metadata(&self, path: &Path) -> Result<KaniMetadata> {
```
celinval, line 144: Why do you need `self` here?

src/cargo-kani/src/call_cbmc_viewer.rs
```diff
@@ -60,7 +60,7 @@ impl KaniSession {
     }
 
     fn cbmc_variant(&self, file: &Path, extra_args: &[&str], output: &Path) -> Result<()> {
-        let mut args = self.cbmc_flags(file)?;
+        let mut args = self.cbmc_flags(file, None)?;
```
tedinski, line 63: `cbmc_flags` shouldn't take an option, and we should propagate the harness to this function so it can be passed it, I'm pretty sure.

jaisnan:
> 

Thanks for the callout, for now, the flags or the property file values for unwind have been disabled (we can enable them if we want). i didn't know which one should take preference over the other so I disabled the flag for now to prevent confusion. 

imo, we should either disable the command line flag or make it have preference over the attribute value. Allowing the command line argument but having it be low preference doesnt make much sense to me.

Would like to know thoughts

src/cargo-kani/src/metadata.rs
jaisnan: I thought this was a function that would be useful in other files as well, so i made it a utililty method for the current Session.

tedinski:
Ah yeah, there are some options for that behavior aren't there?

Since we now run all proof harnesses by default, I propose: (1) We always use the unwind flag on the proof harness if present. (2) if not present, only then do we pay attention to the global command line flag. In short, the command line only sets the global default, a flag on the harness always has precedence.

This might be unexpected behavior if people expect to be able to override the unwind flag from the command line however, but I think that would only make sense to do in some circumstances (like if they've also provided `--harness` so they're only running one harness?). My vote is for ignoring this case for now and let's see what happens or if customers get confused or want the override behavior?

jaisnan:
> Ah yeah, there are some options for that behavior aren't there?
> 
> Since we now run all proof harnesses by default, I propose: (1) We always use the unwind flag on the proof harness if present. (2) if not present, only then do we pay attention to the global command line flag. In short, the command line only sets the global default, a flag on the harness always has precedence.
> 
> This might be unexpected behavior if people expect to be able to override the unwind flag from the command line however, but I think that would only make sense to do in some circumstances (like if they've also provided `--harness` so they're only running one harness?). My vote is for ignoring this case for now and let's see what happens or if customers get confused or want the override behavior?

I agree with this behavior. @celinval, any thoughts?

tests/ui/check_operations/operations.rs
```diff
@@ -2,12 +2,13 @@
...
 //
 // Check the message printed when a checked operation fails.
-// kani-flags: --unwind 3
 extern crate kani;
 
 use kani::any;
 
 #[kani::proof]
+#[kani::unwind(3)]
+
```
adpaco-aws, line 11: Remove newline?

adpaco-aws (dismissed):
Documentation must be updated to describe the feature, and whatever behavior you agree on for precedence should be noted there as well.

In my opinion, we should not be silent about conflicts and emit a user warning in this scenario (regardless of preference). Otherwise users could become really confused.

src/cargo-kani/src/metadata.rs
celinval: 1. I think we should only parse the file once.
2. Can you create a public utility function instead that doesn't rely on self?

celinval:
IMO, the precedence order should be command line, harness attribute, configuration file. That said, I think it will be hard to implement this today given how we merge the configuration file options with the command line one.

I'm OK with @tedinski 's suggestion but I also agree with @adpaco-aws that we shouldn't be silent about any conflict. My suggestion is to do the following:

1. Introduce a "--default-unwind" option if we want to have a weaker behavior that allows harnesses annotations to have higher precedence.
2. Only allow "--unwind" in conjunction with `--harness`. In this case, the command line option would have higher precedence.

jaisnan:
> IMO, the precedence order should be command line, harness attribute, configuration file. That said, I think it will be hard to implement this today given how we merge the configuration file options with the command line one.
> 
> I'm OK with @tedinski 's suggestion but I also agree with @adpaco-aws that we shouldn't be silent about any conflict. My suggestion is to do the following:
> 
>     1. Introduce a `--default-unwind` option if we want to have a weaker behavior that allows harnesses annotations to have higher precedence.
> 
>     2. Only allow `--unwind` in conjunction with `--harness`. In this case, the command line option would have higher precedence.

Alright, this is what I'll implement. Thanks for your inputs everyone.

docs/src/tutorial/loops-unwinding/Cargo.toml
```diff
@@ -10,4 +10,3 @@ edition = "2018"
 [workspace]
 
 [package.metadata.kani]
```
tedinski, line 12: if this section is empty now, I'd remove the heading too

src/kani-driver/src/args.rs
```diff
@@ -321,6 +324,14 @@ impl KaniArgs {
             )
             .exit();
         }
+
+        if self.harness_unwind.is_some() && self.harness.is_none() {
```
tedinski, line 328: Ah, I don't know if there's a good example but you should be able to put `requires` on the `structopt` flags. I think maybe there's an example of a `conflicts`. Then you don't need to write any code here.

(These checks are legacy of trying to avoid changing some tests when I was rewriting the python. If you're looking for another tiny cleanup a separate PR to remove these manual checks in favor of adding the correct annotations, then fixing the expected test, might be nice.)

src/kani-driver/src/call_cbmc.rs
```diff
@@ -127,3 +133,192 @@ impl KaniSession {
...
+
+    #[test]
+    fn check_default_annotation_unwind_resolve() {
+        // Annotation value for unwind takes precedence over unwind value
+        // except when there is no #[kani::unwind()] provided.
+        let args: Vec<OsString> =
+            ["kani", "--no-default-checks", "--unwind", "2"].iter().map(|&s| s.into()).collect();
+        let unwind_args = args::KaniArgs::from_iter(args);
+
+        let harnesses = vec![
```
tedinski, line 168: These should just be separate variables, and there's no need for `find_proof_harness` then.

src/kani-driver/src/call_cbmc.rs
```diff
@@ -127,3 +133,192 @@ impl KaniSession {
...
+
+        assert!(merged_1 == Some(1));
+        assert!(merged_2 == Some(3));
+        assert!(merged_3 == Some(2));
+    }
+
+    #[test]
+    fn check_harness_annotation_unwind_resolve() {
+        // Command line unwind value for specific harnesses take precedence over unwind annotation value
+        let args_1: Vec<OsString> =
```
tedinski, line 195: These tests are over-complex. The behavior of `resolve_unwind_value` does not change with different `--harness` values, so checking multiple doesn't test anything here

src/kani-driver/src/call_cbmc.rs
```diff
@@ -127,3 +133,192 @@ impl KaniSession {
...
+            find_proof_harness("not_check_three", &harnesses).unwrap(),
+        );
+
+        assert!(merged_1 == Some(2));
+        assert!(merged_2 == Some(2));
+        assert!(merged_3 == Some(2));
+    }
+
+    #[test]
+    fn check_default_harness_unwind_resolve() {
```
tedinski, line 246: So there's two unwind flags that can be present or not, and an annotation that can be present or not. 

There should be one `#[test]` that has those 8 asserts. And some of those combinations maybe aren't even possible. This is too complex

src/kani-driver/src/metadata.rs
```diff
@@ -138,22 +131,30 @@ impl KaniSession {
             Ok(metadata.proof_harnesses.clone())
         }
     }
+
+    /// Deserialize a *.restrictions.json file
+    fn read_kani_metadata(&self, path: &Path) -> Result<KaniMetadata> {
```
tedinski, line 136: This function change should be reverted. This doesn't use `self`

tedinski (commented):
There's a couple of stray changes that should be cleaned up, but the main thing is the testing.

src/kani-driver/src/args.rs
```diff
@@ -280,18 +283,18 @@ impl CargoKaniArgs {
 }
 impl KaniArgs {
     pub fn validate(&self) {
-        let extra_unwind = self.cbmc_args.contains(&OsString::from("--unwind"));
+        let extra_unwind = self.cbmc_args.contains(&OsString::from("--default-unwind"));
```
tedinski, line 286: this should still be `--unwind` actually

src/kani-driver/src/args.rs
```diff
@@ -365,6 +368,15 @@ mod tests {
         assert_eq!(err.kind, ErrorKind::ArgumentConflict);
     }
 
+    #[test]
+    fn check_unwind_conflicts() {
+        // harness needs metadata which we don't have with dry-run
```
tedinski, line 373: comment is wrong :)

src/kani-driver/src/metadata.rs
```diff
@@ -140,20 +132,28 @@ impl KaniSession {
...
         mangled_name: name.into(),
         original_file: "<unknown>".into(),
         original_line: "<unknown>".into(),
-        unwind_value: None,
+        unwind_value: unwind_value,
     }
 }
 
+/// Deserialize a *.restrictions.json file
+fn read_kani_metadata(path: &Path) -> Result<KaniMetadata> {
```
tedinski, line 146: this is still showing up as a diff, it moved in the file

src/kani-driver/src/metadata.rs
```diff
@@ -140,20 +132,28 @@ impl KaniSession {
...
+    let reader = BufReader::new(file);
+    let restrictions = serde_json::from_reader(reader)?;
+    Ok(restrictions)
+}
+
 /// Search for a proof harness with a particular name.
 /// At the present time, we use `no_mangle` so collisions shouldn't happen,
 /// but this function is written to be robust against that changing in the future.
-fn find_proof_harness<'a>(
+pub fn find_proof_harness<'a>(
```
tedinski, line 156: i think this doesn't need pub anymore

tedinski (approved):

PR #1039, opened 2022-04-12
Introduce kani-verifier: proxy binaries

### LINKED ISSUE #944: Minimum cargo-kani setup
Supporting #943, we should have a minimum `cargo-kani setup` or first-time run step that at least ensure that the correct Rust toolchain is installed, if nothing else.

This way at least "download release bundle" will work.

tedinski:
### Description of changes: 

Adds the `kani-verifier` crate. This contains "proxy" binaries: `kani` and `cargo-kani`.

These will attempt to download a release from github, but since one of those doesn't exist yet, it can currently be given a special `--use-local-bundle <bundle>` flag instead.

### Resolved issues:

Resolves #944 

### Call-outs:

1. Temporarily, we're keeping `kani-verifier` not part of the main workspace. This is because we have two binaries named `cargo-kani`. I'll follow up with a PR that renames `src/cargo-kani` to `src/kani-driver` and then this conflict will be resolved.
2. The Dockerfile is now simple! Yay!
3. I have hardcoded dependency versions in the proxy binaries. So these versions are scattered around. We need to come up with a solution so there's a single source of truth here. :(

### Testing:

* How is this change tested? The Dockerfile

### Checklist


.gitignore
```diff
@@ -86,3 +86,4 @@ tests/kani-multicrate/type-mismatch/mismatch/target
 /docs/book
 /docs/mdbook*
 /kani-*.tar.gz
+/src/kani-verifier/target
```
celinval, line 89: Why is this building inside src?

Cargo.toml
```diff
@@ -16,10 +16,12 @@ members = [
 ]
 
 exclude = [
+  # temporarily exclude `kani-verifier` until we don't have colliding binary names (cargo-kani)
+  "src/kani-verifier",
```
celinval, line 20: Since we will no longer publish the other crates to crates.io, we probably want to move them out of src and only keep this new crate in.

This will also fix the readme issue.

src/kani-verifier/src/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    // 3. Connect to our toolchain
+    Command::new("ln").arg("-s").arg(toolchain).arg(kani_dir.join("toolchain")).run()?;
+
+    // 4. Install our python dependencies
+    let pyroot = kani_dir.join("pyroot");
+
+    // TODO: this is a repetition of versions from elsewhere
+    Command::new("python3")
+        .args(&["-m", "pip", "install", "cbmc-viewer==2.11", "--target"])
```
celinval, line 118: Have you considered having one source of truth for these dependencies? This looks a bit fragile.

src/kani-verifier/src/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+    // We need a workspace to build them in, otherwise repeated builds generate different hashes and `kani` can't find `kani_macros`
+    let contents = "[workspace]\nmembers = [\"kani\",\"kani_macros\",\"std\"]";
+    std::fs::write(kani_dir.join("library").join("Cargo.toml"), contents)?;
+
+    // A little helper for invoking Cargo repeatedly here
+    let cargo = |crate_name: &str| -> Result<()> {
+        let manifest = format!("library/{}/Cargo.toml", crate_name);
+        Command::new("cargo")
+            .args(&[
+                "+nightly-2022-03-23",
```
celinval, line 136: Why is this hardcoded?

src/kani-verifier/src/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+
+        // when given a bundle, it's often "-latest" but we expect "-1.0" or something. Hack it up.
+        let file = path.file_name().expect("has filename").to_string_lossy();
+        let components: Vec<_> = file.split("-").collect();
+        let expected_dir = format!("{}-{}", components[0], components[1]);
+
+        std::fs::rename(base_dir.join(expected_dir), &kani_dir)?;
+    } else {
+        let bundle = base_dir.join(download_filename());
+        Command::new("curl").args(&["-sSLf", "-o"]).arg(&bundle).arg(download_url()).run()?;
```
celinval, line 97: Can we add some error handling here? We probably want to give the user a nice message if we fail to download the bundle. And a message that help us root cause the issue. :)

celinval (commented):
Sweet! Just a few initial comments. I was also wondering if we should add any sort of hash check against the bundle to make sure we're downloading the correct file.

Cargo.toml
tedinski: I don't understand this comment. We possibly need to chat about this because I'm not sure how best to solve this issue, and this sounds like you understand something here that I don't yet.

src/kani-verifier/src/lib.rs
tedinski: Sure, I will do a whole pass trying to improve the error handling and messaging. I know it's a problem currently.

It's not suppressing the output though, so you do see a `curl` error here saying 404 and giving the URL, fwiw.

src/kani-verifier/src/lib.rs
tedinski: Yeah, we could really use a solution to that problem. I haven't yet figured out how we can change dev environments and CI around to use something more like a release bundle approach (which I think would be the best way to have a single source of truth: right here).

src/kani-verifier/src/lib.rs
tedinski: Ooops! Will fix. Good catch.

tedinski:
>I was also wondering if we should add any sort of hash check against the bundle to make sure we're downloading the correct file.

I agree we should, but I need to nail down the release process to figure out if we can, exactly. Probably, but it will entail making a release bundle and then making a commit to update the hash, so it's a bit awkward ("This is 1.0! Release it! Then add a commit and tag THAT as 1.0 and push to crates.io!")

tedinski:
Here are some examples of the output we get now.

Successful log:

```
[0/6] Running Kani first-time setup...
[1/6] Ensuring the existence of: /root/.kani
[2/6] Installing local Kani bundle: /tmp/kani/kani-latest-x86_64-unknown-linux-gnu.tar.gz
[3/6] Installing rust toolchain version: nightly-2022-03-23-x86_64-unknown-linux-gnu
info: syncing channel updates for 'nightly-2022-03-23-x86_64-unknown-linux-gnu'
info: latest update on 2022-03-23, rust version 1.61.0-nightly (5f3700105 2022-03-22)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'

  nightly-2022-03-23-x86_64-unknown-linux-gnu installed - rustc 1.61.0-nightly (5f3700105 2022-03-22)

info: checking for self-updates
[4/6] Installing Kani python dependencies...
Collecting cbmc-viewer==2.11
  Downloading cbmc_viewer-2.11-py3-none-any.whl (80 kB)
Collecting voluptuous
  Downloading voluptuous-0.13.1-py3-none-any.whl (29 kB)
Collecting jinja2
  Downloading Jinja2-3.1.1-py3-none-any.whl (132 kB)
Collecting setuptools
  Downloading setuptools-62.1.0-py3-none-any.whl (1.1 MB)
Collecting MarkupSafe>=2.0
  Downloading MarkupSafe-2.1.1-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (25 kB)
Installing collected packages: voluptuous, MarkupSafe, jinja2, setuptools, cbmc-viewer
Successfully installed MarkupSafe-2.1.1 cbmc-viewer-2.11 jinja2-3.1.1 setuptools-62.1.0 voluptuous-0.13.1
Collecting colorama==0.4.3
  Downloading colorama-0.4.3-py2.py3-none-any.whl (15 kB)
Installing collected packages: colorama
Successfully installed colorama-0.4.3
[5/6] Building Kani library prelude...
   Compiling kani_macros v0.1.0 (/root/.kani/kani-0.1.0/library/kani_macros)
   Compiling kani v0.1.0 (/root/.kani/kani-0.1.0/library/kani)
    Finished dev [unoptimized + debuginfo] target(s) in 0.82s
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
   Compiling std v0.1.0 (/root/.kani/kani-0.1.0/library/std)
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
[6/6] Successfully completed Kani first-time setup.
```

Failed to download:

```
[0/6] Running Kani first-time setup...
[1/6] Ensuring the existence of: /root/.kani
[2/6] Downloading Kani release bundle: kani-0.1.0-x86_64-unknown-linux-gnu.tar.gz
curl: (22) The requested URL returned error: 404 
Error: Failed command: curl -sSLf -o /root/.kani/kani-0.1.0-x86_64-unknown-linux-gnu.tar.gz https://github.com/model-checking/kani/releases/download/kani-0.1.0/kani-0.1.0-x86_64-unknown-linux-gnu.tar.gz
```

Forgot to create the Kani base directory:

```
[0/6] Running Kani first-time setup...
[1/6] Ensuring the existence of: /root/.kani
[2/6] Installing local Kani bundle: /tmp/kani/kani-latest-x86_64-unknown-linux-gnu.tar.gz
Error: Internal failure before invoking command: tar zxf /tmp/kani/kani-latest-x86_64-unknown-linux-gnu.tar.gz

Caused by:
    No such file or directory (os error 2)
```

(The above is a bit confusing because it's caused by the `current_dir` not existing in a command. But at least this tells us which command failed to run which should point us, if not the customers, to the problem.)

`rust-toolchain` didn't exist in the bundle:

```
[0/6] Running Kani first-time setup...
[1/6] Ensuring the existence of: /root/.kani
[2/6] Installing local Kani bundle: /tmp/kani/kani-latest-x86_64-unknown-linux-gnu.tar.gz
Error: Reading release bundle rust-toolchain

Caused by:
    No such file or directory (os error 2)
```

Python package didn't exist:

```
[0/6] Running Kani first-time setup...
[1/6] Ensuring the existence of: /root/.kani
[2/6] Installing local Kani bundle: /tmp/kani/kani-latest-x86_64-unknown-linux-gnu.tar.gz
[3/6] Installing rust toolchain version: nightly-2022-03-23-x86_64-unknown-linux-gnu
info: syncing channel updates for 'nightly-2022-03-23-x86_64-unknown-linux-gnu'
info: latest update on 2022-03-23, rust version 1.61.0-nightly (5f3700105 2022-03-22)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'

  nightly-2022-03-23-x86_64-unknown-linux-gnu installed - rustc 1.61.0-nightly (5f3700105 2022-03-22)

info: checking for self-updates
[4/6] Installing Kani python dependencies...
ERROR: Could not find a version that satisfies the requirement asdfasdfcbmc-viewer==2.11 (from versions: none)
ERROR: No matching distribution found for asdfasdfcbmc-viewer==2.11
Error: Failed command: python3 -m pip install asdfasdfcbmc-viewer==2.11 --target /root/.kani/kani-0.1.0/pyroot
```

src/kani-verifier/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+        if !appears_setup() {
+            setup(None)?;
+        }
+        exec(bin)
+    }
+}
+
+/// Fast check to see if we look setup already
+fn appears_setup() -> bool {
+    kani_dir().exists()
```
celinval, line 70: I do have a question about versions. What will happen if the user has `kani` installed but they are trying to use a different version. I think we want a way to check the version and allow user to upgrade or even downgrade their current setup.

I'm ok for now to only support one version, but it is also something that we might want to deal with eventually.

src/kani-verifier/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+        Command::new("curl").args(&["-sSLf", "-o"]).arg(&bundle).arg(download_url()).run()?;
+
+        Command::new("tar").arg("zxf").arg(&bundle).current_dir(base_dir).run()?;
+
+        std::fs::remove_file(bundle)?;
+    }
+
+    let toolchain_version = std::fs::read_to_string(kani_dir.join("rust-toolchain"))
+        .context("Reading release bundle rust-toolchain")?;
+    println!("[3/6] Installing rust toolchain version: {}", &toolchain_version);
```
celinval, line 108: These messages are much nicer. Thanks!

For future thought, we might want to provide a way to control verbosity.

src/kani-verifier/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+        println!("[2/6] Downloading Kani release bundle: {}", &filename);
+        let bundle = base_dir.join(filename);
+        Command::new("curl").args(&["-sSLf", "-o"]).arg(&bundle).arg(download_url()).run()?;
+
+        Command::new("tar").arg("zxf").arg(&bundle).current_dir(base_dir).run()?;
+
+        std::fs::remove_file(bundle)?;
+    }
+
+    let toolchain_version = std::fs::read_to_string(kani_dir.join("rust-toolchain"))
```
celinval, line 106: I would prefer if we just upload the rust toolchain file here. If you are trying to avoid any sort of parsing, I would recommend using a different name. The `rust-toolchain` file is a legacy name used by `rustup` for configuration files that have specific file format specifications.

https://rust-lang.github.io/rustup/overrides.html?highlight=rust-toolchain.toml#the-toolchain-file

src/kani-verifier/src/lib.rs
tedinski: The version is currently baked in to the proxy. This one for example is looking for 0.1 (set as the package version in Cargo.toml).

So it's finding `~/.kani/kani-0.1/` and everything is under that.

tedinski:
We should add a license note

src/kani-verifier/src/lib.rs
celinval: Ah, I totally missed that. Thanks!

celinval (approved):
Thanks! Please don' forget to create the issue to fix the hard-coded dependency versions.

PR #1040, opened 2022-04-13
Update VecDeque CVE example

nchong-at-aws:
- Add README instructions for reproducing results
- Add a reproducible example that can be detected using valgrind
- Add a minimal example that Kani can error when using the buggy version of `VecDeque` and verify when using the fixed version
- Add a symbolic example that causes Kani to timeout
- Add an abstract implementation of `VecDeque` for deeper verification results


tests/cargo-kani/vecdeque-cve/Cargo.toml
```diff
@@ -8,8 +8,3 @@ description = "Reproduces CVE-2018-1000657"
 
 [lib]
 path = "src/harness.rs"
-
-[dependencies]
-
-[package.metadata.kani.flags]
-cbmc-args = ["--no-built-in-assertions"]
```
nchong-at-aws, line 15: I found this wasn't needed to verify the harnesses.

tests/cargo-kani/vecdeque-cve/src/harness.rs
```diff
@@ -11,9 +11,16 @@
 #![feature(ptr_internals)]
 #![feature(rustc_allow_const_fn_unstable)]
 
+#[cfg(disable_debug_asserts)]
+macro_rules! debug_assert {
+    ( $( $x:expr ),* ) => {};
+}
```
nchong-at-aws, line 17: I'm sure there's a better way using `RUSTFLAGS` but I couldn't work out a way to disable `debug_asserts` at the `cargo kani` command line.

zhassan-aws: Kani overrides `debug_asserts` to apply the same instrumentation used for asserts: https://github.com/model-checking/kani/blob/4d4a5a7b99121f943b6d726b2c559767209531cd/library/std/src/lib.rs#L82 so what you did is likely the only way to disable them, unless we want to support disabling debug asserts (similar to the behavior of `cargo build --release`) which we can do through defining them as follows:
```
#[macro_export]
macro_rules! debug_assert {
    ($($x:tt)*) => (if $crate::cfg!(debug_assertions) ({ $crate::assert!($($x)*); })
}
```

nchong-at-aws: Ah, I see. That's why my attempts didn't work. I tried `RUSTFLAGS='--cfg "debug_assertions=\"false\""'` and using a profile in `Cargo.toml` (https://doc.rust-lang.org/book/ch14-01-release-profiles.html):

```
[profile.dev]
opt-level = 3
```

tests/cargo-kani/vecdeque-cve/Cargo.toml
nchong-at-aws: @celinval Was there a reason why we removed these checks in the initial commit?

celinval: I was trying to reduce the size of the report to make it more demo friendly. :)

tests/cargo-kani/vecdeque-cve/src/harness.rs
zhassan-aws: Would adding support for turning off `debug_assertions` help? I can push a PR to do that.

tests/cargo-kani/vecdeque-cve/Cargo.toml
```diff
@@ -7,10 +7,4 @@ edition = "2018"
...
 
 [lib]
-path = "src/harness.rs"
-
-[dependencies]
-
-[package.metadata.kani.flags]
-enable-unstable = true
-cbmc-args = ["--no-built-in-assertions"]
+path = "src/harness.rs"
```
zhassan-aws, line 10: Missing newline.

tests/cargo-kani/vecdeque-cve/src/abstract_vecdeque.rs
```diff
@@ -0,0 +1,359 @@
...
+/// We don't model alloc type `A`
+struct AbstractVecDeque {
+    tail: usize,
+    head: usize,
+    buf: AbstractRawVec,
+}
+
+unsafe impl kani::Invariant for AbstractVecDeque {
+    fn is_valid(&self) -> bool {
+        self.tail < self.cap() && self.head < self.cap() && is_nonzero_pow2(self.cap())
```
zhassan-aws, line 29: Nice invariant!

tests/cargo-kani/vecdeque-cve/src/abstract_vecdeque.rs
```diff
@@ -0,0 +1,359 @@
...
+fn handle_reserve(result: Result<(), TryReserveError>) {
+    match result.map_err(|e| e.kind()) {
+        Err(CapacityOverflow) => capacity_overflow(),
+        Err(AllocError) => handle_alloc_error(),
+        Ok(()) => { /* yay */ }
+    }
+}
+
+fn capacity_overflow() {
+    assert!(false);
```
zhassan-aws, line 278: Replace with `panic` or `unreachable`?

tests/cargo-kani/vecdeque-cve/src/abstract_vecdeque.rs
```diff
@@ -0,0 +1,359 @@
...
+}
+
+fn handle_alloc_error() {
+    assert!(false);
```
zhassan-aws, line 282: Same for this one.

zhassan-aws (commented):
Looks good!

PR #1041, opened 2022-04-13
Add tests for `forget`

adpaco-aws:
### Description of changes: 

Adds a positive and a negative test case for `forget` and updates the support table accordingly.

### Resolved issues:

Part of #727 

### Call-outs:

PR was originally "Disable `forget` intrinsic" but was changed after discussion below. Description included below for reference.

> Removes the implementation for `forget`, which was being encoded as a no-op.
> 
> Note that this is not the well-known `forget` which calls `ManuallyDrop`, but an alternative that moves the value out of scope without calling drop functions (used by `mem::forget_unsized`): https://doc.rust-lang.org/std/intrinsics/fn.forget.html

### Testing:

### Checklist


zhassan-aws:
It seems to me that using a no-op for `forget` is sound since the only effect of using the `forget` intrinsic is to produce a compiler error if the value is referenced after "forgetting" it. I cannot think of any other verification-related effects of calling it.

adpaco-aws:
Thanks @zhassan-aws for the detailed explanation!

To be clear, I was under the impression this was not sound because I thought we had to generate code for "forgetting" the value in CBMC. But as you explained we do not need to do anything during code generation, so a no-op would be fine.

Now, instead of removing it, I was thinking of adding two tests similar to the example found [here](https://doc.rust-lang.org/std/mem/fn.forget.html#relationship-with-manuallydrop) but using this `forget`.
```rust
fn main() {
let mut v = vec![65, 122];
let s = unsafe { String::from_raw_parts(v.as_mut_ptr(), v.len(), v.capacity()) };
std::intrinsics::forget(v);
assert!(v[0] == 65); // this causes failure to codegen
assert_eq!(s, "Az");
}
```
This should fail to codegen, and another test not including the first assertion should pass. What do you think?

adpaco-aws:
Just pushed the changes I was proposing because I figured it's something we would like to have anyway. Updated title and description as well.

zhassan-aws (approved):
Makes sense to me (technically it would be a compilation error as opposed to a codegen failure with the first assert, but I guess for the purpose of compiletest, it's the same thing).

adpaco-aws:
Right, switched to `kani-check-fail` which is used to indicate errors before codegen.

PR #1053, opened 2022-04-15
Fix empty assertions with "assertion" description

### LINKED ISSUE #1051: Missing Property String and Description in some Checks
Some checks have missing property name and description fields which causes them to look like this - 

```
RESULTS:
Check 1: 1
         - Status: FAILURE
         - Description: "assertion"
Check 2: 2
         - Status: SUCCESS
         - Description: "assertion"
```

These aren't caught by the regression tests as they mostly seem to be having the right status or they're not checked by the test files.

These checks need to be corrected by placing their appropriate property class and description fields.

jaisnan:
### Description of changes: 

1) Fixes empty assertions by adding property classes to `assert_false`. This fix should affect a lot of test cases wherever `assert_false` was used. 

2) Adds new property class variants - `cover`, `abort`, `unequal-assign`.

3) Refactor creating location with property types from `codegen` to `Stmt::assert.`

4) Add basic assert to `Stmt::assert` to make sure `Property_Class` and `Description` are not empty.

### Resolved issues:

Resolves #1051 

### Call-outs:

This doesn't yet add any unit tests to the Python parser or UI tests. That can be addressed in a future PR. 

### Testing:

* How is this change tested? Codegen assertion.

### Checklist


cprover_bindings/src/goto_program/stmt.rs
```diff
@@ -176,6 +176,7 @@ impl Stmt {
                 rhs.typ()
             );
             let assert_stmt = Stmt::assert_false(
+                "unequal-assign",
```
tedinski, line 179: I think you're probably creating too many property classes here. I think this one should be a generic "sanity check" (do we have an existing one for that?)

cprover_bindings/src/goto_program/stmt.rs
```diff
@@ -193,16 +194,21 @@ impl Stmt {
...
+        // Create a Property Location Variant from any given Location type
+        let loc_with_property =
+            Location::create_location_with_property(message, property_name, loc);
 
         // Chose InternedString to seperate out codegen from the cprover_bindings logic
         let property_class = property_name.to_string().intern();
         let msg = message.into();
 
-        stmt!(Assert { cond, property_class, msg }, loc)
+        stmt!(Assert { cond, property_class, msg }, loc_with_property)
```
tedinski, line 207: is this the root cause of the bug? seems like this went from accidentally ignoring them to using them?

cprover_bindings/src/goto_program/stmt.rs
```diff
@@ -193,16 +194,21 @@ impl Stmt {
...
     }
 
-    pub fn assert_false(msg: &str, loc: Location) -> Self {
-        Stmt::assert(Expr::bool_false(), "assert_false", msg, loc)
+    pub fn assert_false(property_name: &str, message: &str, loc: Location) -> Self {
```
tedinski, line 210: I like this

src/kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -75,24 +75,18 @@ impl<'tcx> GotocCtx<'tcx> {
...
         assert!(cond.typ().is_bool());
-
         let property_name = property_class.as_str();
-
-        // Create a Property Location Variant from any given Location type
-        let property_location =
-            Location::create_location_with_property(message, property_name, loc);
-
-        Stmt::assert(cond, property_name, message, property_location)
+        Stmt::assert(cond, property_name, message, loc)
```
tedinski, line 79: Ah, maybe this answers my previous question about root cause. So we went from applying it early, to applying it in a later choke point (with asserts looking for reasonable values), to ensure we don't forget to do so. Right? Seems like an excellent change. Nice.

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -46,8 +46,16 @@ impl<'tcx> GotocCtx<'tcx> {
...
             }
-            TerminatorKind::Resume => self.codegen_assert_false("resume instruction", loc),
-            TerminatorKind::Abort => self.codegen_assert_false("abort instruction", loc),
+            TerminatorKind::Resume => self.codegen_assert_false(
+                PropertyClass::CustomProperty("resume".to_string()),
+                "resume instruction",
+                loc,
+            ),
+            TerminatorKind::Abort => self.codegen_assert_false(
+                PropertyClass::CustomProperty("abort".to_string()),
```
tedinski, line 55: Unsupported language construct?

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -360,6 +372,7 @@ impl<'tcx> GotocCtx<'tcx> {
                     // No target block means this function doesn't return.
                     // This should have been handled by the Nevers hook.
                     return self.codegen_assert_false(
+                        PropertyClass::CustomProperty("nonterminating".to_string()),
```
tedinski, line 375: I *think* Sanity check?

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -505,7 +518,11 @@ impl<'tcx> GotocCtx<'tcx> {
         let loc = self.codegen_caller_span(&span);
         Stmt::block(
             vec![
-                self.codegen_assert_false(msg, loc),
+                self.codegen_assert_false(
+                    PropertyClass::CustomProperty("fatal-error".to_string()),
```
tedinski, line 522: I'm not actually sure how `codegen_fatal_error` is used, so I'm not sure what property class this one should be.

tedinski (commented):
Nice work finding these bugs

src/kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
jaisnan: Yeah, that's about it. It wasn't consistently applied everywhere which created the issue for a few `assert_false` calls. Now, `assert` all ensure structure and quality of the values that it needs.

cprover_bindings/src/goto_program/stmt.rs
jaisnan: yes we do! "sanity-checks" would get cleaned up by the final UI too so it makes sense to use that here.

jaisnan:
@tedinski do you think we need tests (ui tests) specifically for this bug? as far i can see, the bug is fixed within all the regression tests wherever it was happening but there's no test for the bug itself.

tedinski:
> do you think we need tests (ui tests) specifically for this bug

I think your change to the function signature and addition of asserts should cover things nicely.

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -505,7 +518,7 @@ impl<'tcx> GotocCtx<'tcx> {
         let loc = self.codegen_caller_span(&span);
         Stmt::block(
             vec![
-                self.codegen_assert_false(msg, loc),
+                self.codegen_assert_false(PropertyClass::SanityCheck, msg, loc),
```
tedinski, line 521: I think Celina is right and `codegen_fatal_error` should take a property class parameter.

The first thing I saw using it seemed to be an unsupported construct.

jaisnan: oh yeah, let me fix that, I didn't include it in this commit.

tedinski (approved):

PR #1054, opened 2022-04-15
Assume an x86_64 architecture for the machine model

### LINKED ISSUE #64: Ensure we are building the right machine model from a session
In particular, should char be signed? What should the rounding mode be?

adpaco-aws:
### Description of changes: 

Checks that the target architecture is x86_84, prints a nice error otherwise. Also adds comments to the function building the machine model.

Should we add further requirements for the values coming from the session?

### Resolved issues:

Resolves #64 

### Testing:

* How is this change tested? CI regression

### Checklist


src/kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -184,6 +184,11 @@ impl CodegenBackend for GotocCodegenBackend {
 }
 
 fn check_options(session: &Session, need_metadata_module: bool) {
+    // The requirement below is needed to build a valid CBMC machine model
```
tedinski, line 187: Nit: can we point/link to/reference between these two things.

e.g. mention `machine_model_from_session` here.

You already back-reference to here from the other spot, just completing the circle :)

src/kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
```diff
@@ -345,27 +345,35 @@ impl MetadataLoader for GotocMetadataLoader {
     }
 }
 
+// Builds a machine model which is required by CBMC.
```
tedinski, line 348: Nit: 3 slashes

tedinski (approved):

src/kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
celinval: Maybe add a link to an issue too.

adpaco-aws:
Question: Do we want to also check here the values we get from the session? I am not sure if it's possible to get 32 as `pointer_width` for an x86_64 architecture, or in general a value we don't expect here (big endian, min. alignment > 1).

adpaco-aws:
I have used the commands [here](https://docs.rust-embedded.org/embedonomicon/compiler-support.html) to obtain get the `pointer_width` values for all supported "x86_64-*" architectures:

```
x86_64-apple-darwin
  "target-pointer-width": "64",
x86_64-apple-ios
  "target-pointer-width": "64",
x86_64-apple-ios-macabi
  "target-pointer-width": "64",
x86_64-apple-tvos
  "target-pointer-width": "64",
x86_64-fortanix-unknown-sgx
  "target-pointer-width": "64",
x86_64-fuchsia
  "target-pointer-width": "64"
x86_64-linux-android
  "target-pointer-width": "64"
x86_64-pc-solaris
  "target-pointer-width": "64",
x86_64-pc-windows-gnu
  "target-pointer-width": "64",
x86_64-pc-windows-msvc
  "target-pointer-width": "64",
x86_64-sun-solaris
  "target-pointer-width": "64",
x86_64-unknown-dragonfly
  "target-pointer-width": "64"
x86_64-unknown-freebsd
  "target-pointer-width": "64"
x86_64-unknown-haiku
  "target-pointer-width": "64"
x86_64-unknown-hermit
  "target-pointer-width": "64",
x86_64-unknown-illumos
  "target-pointer-width": "64"
x86_64-unknown-l4re-uclibc
  "target-pointer-width": "64"
x86_64-unknown-linux-gnu
  "target-pointer-width": "64"
x86_64-unknown-linux-gnux32
  "target-pointer-width": "32"
x86_64-unknown-linux-musl
  "target-pointer-width": "64"
x86_64-unknown-netbsd
  "target-pointer-width": "64",
x86_64-unknown-none
  "target-pointer-width": "64"
x86_64-unknown-none-hermitkernel
  "target-pointer-width": "64"
x86_64-unknown-none-linuxkernel
  "target-pointer-width": "64"
x86_64-unknown-openbsd
  "target-pointer-width": "64"
x86_64-unknown-redox
  "target-pointer-width": "64"
x86_64-unknown-uefi
  "target-pointer-width": "64"
x86_64-uwp-windows-gnu
  "target-pointer-width": "64",
x86_64-uwp-windows-msvc
  "target-pointer-width": "64",
x86_64-wrs-vxworks
  "target-pointer-width": "64",
```

So there is at least one platform (`x86_64-unknown-linux-gnux32`) where the analysis would not be sound. Please note that [custom targets](https://docs.rust-embedded.org/embedonomicon/custom-target.html) can also be specified.

I have changed the original proposal so that the values we get from the session are checked against the ones we expect, in  addition to these ones:
* Add a new function `check_target_arch_spec` similar to `check_options` to specifically check target values that aren't really options.
 * Use `///` for comment on `machine_model_from_session`
 * Add comments to point to relevant functions (from `check_*` to `machine_model_*` and vice versa).

I am requesting another review since the PR has changed substantially.

tedinski (approved):

celinval:
I suggest that we restrict this to the following platforms:

- x86_64-unknown-linux-gnu
- x86_64-apple-darwin

A bunch of the platforms you listed are not even Tier 1 for the rustc compiler (https://doc.rust-lang.org/nightly/rustc/platform-support.html)

tedinski:
Yeah, FWIW, the Kani installer will only be working for those two platforms.

(Which reminds me, I should add an explicit check. Right now it's just fail to download a release bundle...)

adpaco-aws:
Agreed, updated the check to allow only these platforms.

adpaco-aws:
The regression for macOS fails because it does not detect `x86_64-apple-darwin` but `x86_64-apple-macosx10.7.0`. Not sure what to do now.

tedinski:
I'm adding the "nice" check to the installer in #1069 

I think your original idea of asserting on "x86_64" and "64" on those original values is fine here.

adpaco-aws:
> I think your original idea of asserting on "x86_64" and "64" on those original values is fine here.

Thanks @tedinski , but I think we can keep this solution if we just check that `session.target.llvm_target.starts_with("x86_64-apple-")`. Please take a look.

celinval (approved):
That's definitely better than checking for just x86_64 and 64 bits. Too bad the MacOs had to be relaxed a bit. Thanks for doing this!

Issue #707, opened 2021-12-20
Add support for binding sub-arrays using slice patterns


zhassan-aws:
Requested feature: Add full support for Rust slice patterns. Currently, only binding sub-slices is supported (see #357). Binding sub-arrays panics.
Use case: Slice patterns
Link to relevant documentation (Rust reference, Nomicon, RFC): https://rust-lang.github.io/rfcs/2359-subslice-pattern-syntax.html
Is this a breaking change? No

Test case:

```rust
let [x, y @ .., z] = [1, 2, 3, 4]; // `y: [i32, 2] = [2, 3]
```

zhassan-aws:
We currently don't handle the head/tail pattern fully. Add an appropriate error message and add to the limitations section of the docs, then we can remove the "MLP - Must Have" and change it from a bug to a feature.

zhassan-aws:
With #1066, this no longer crashes, and instead we codegen assert/assume false that will cause verification to fail if the array subslice pattern is reachable. Removing the "MLP - Must Have" label.

matthiaskrgr:
````rust

fn array() -> [(String, String); 3] {
    Default::default()
}

#[kani::proof]
fn move_out_by_const_subslice_and_index_field() {
    let a = array();
    let [_, _y @ ..] = a;
    let [(ref _x, _), _, _] = a;
}

#[kani::proof]
fn move_out_by_const_subslice_and_end_index_field() {
    let a = array();
    let [_y @ .., _] = a;
    let [.., (ref _x, _)] = a;
}

// Slice + Slice
#[kani::proof]
fn move_out_by_subslice_and_subslice() {
    let a = array();
    let [x @ .., _, _] = a;
    let [_, ref _y @ ..] = a;
}

````
with mir-opt-level=3: `Failed Checks: Sub-array binding is not currently supported by Kani. Please post your example at https://github.com/model-checking/kani/issues/707`

PR #1057, opened 2022-04-15
Wrap returned value from codegen_place in a Result to account for the possibility of an error due to unsupported operations

zhassan-aws:
### Description of changes: 

Currently, `codegen_place` returns a `ProjectedPlace`. For projections that are currently not supported, Kani panics since it's not possible to return a `codegen_unimplemented` expression as a `ProjectedPlace`. This PR changes the return type of `codegen_place` to a `Result<ProjectedPlace, Err(...)>` so that it can return an error for those cases that is then handled by the caller through calling `codegen_unimplemented`.

### Resolved issues:

Part of #707.

### Testing:

* How is this change tested? A new test for sub-array binding was added that checks that Kani doesn't crash, and instead reports that the construct is not supported.

* Is this a refactor change? Part of it is.

### Checklist


src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -550,3 +584,39 @@ impl<'tcx> GotocCtx<'tcx> {
         arr.member("0", &self.symbol_table).index_array(idx)
     }
 }
+
+/// A convenience macro that unwraps a `Result<ProjectPlace<'tcx>,
+/// Err<UnimplementedData>` if it is Ok and returns an `codegen_unimplemented`
+/// expression otherwise
```
tedinski, line 590: quick question: why macro and not function?

tedinski, line 590: quick question: why macro and not function?

zhassan-aws: For this to have minimal changes on the caller side, it has to do a `return`, which cannot be done using a function (since it will return from that function, and not from the caller function). I hope the explanation is not too confusing.

src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -385,25 +399,42 @@ impl<'tcx> GotocCtx<'tcx> {
...
-            // This is known to fail with a CBMC invariant violation
-            // in some cases. Full support to be added in
-            // https://github.com/model-checking/kani/issues/357
+            // Full support to be added in
+            // https://github.com/model-checking/kani/issues/707
             ProjectionElem::Subslice { from, to, from_end } => {
                 // https://rust-lang.github.io/rfcs/2359-subslice-pattern-syntax.html
                 match before.mir_typ().kind() {
-                    ty::Array(..) => unimplemented!(),
+                    ty::Array(ty, len) => {
```
danielsn, line 419: Should this be a separate PR? Or change the title of this PR?

src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -550,3 +584,39 @@ impl<'tcx> GotocCtx<'tcx> {
         arr.member("0", &self.symbol_table).index_array(idx)
     }
 }
+
+/// A convenience macro that unwraps a `Result<ProjectPlace<'tcx>,
+/// Err<UnimplementedData>` if it is Ok and returns an `codegen_unimplemented`
+/// expression otherwise
+#[macro_export]
+macro_rules! unwrap_or_codegen_unimplemented {
```
danielsn, line 592: This macro affects control flow where it is called; we should make sure to call that our clearly in the name.

src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -550,3 +584,39 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                err.bug_url.as_str(),
+            );
+        }
+        $pp_result.unwrap()
+    }};
+}
+
+/// Same as the above macro, but returns a goto program `Stmt` instead
+#[macro_export]
+macro_rules! unwrap_or_codegen_unimplemented_stmt {
```
danielsn, line 608: Same, call out that this returns

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -497,7 +507,9 @@ impl<'tcx> GotocCtx<'tcx> {
         if self.place_ty(p).is_unit() {
             e.as_stmt(Location::none())
         } else {
-            self.codegen_place(&p).goto_expr.assign(e, Location::none())
+            unwrap_or_codegen_unimplemented_stmt!(self, self.codegen_place(&p))
```
danielsn, line 510: Is there a way to avoid repeating `self`, `self.codegen_place`

danielsn (dismissed):
Like the general idea, but we need a better name for the macro

src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
zhassan-aws: Makes sense to do in a separate PR. I took it out (along with the expected test that was added).

src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
zhassan-aws: I added return in the macro name: `unwrap_or_return_codegen_unimplemented`. Let me know if you have other suggestions.

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
zhassan-aws: I had the same thought, but couldn't figure out a way to do it. I initially tried not passing `self` in, and just calling it from inside the macro, but that didn't work due to macro hygiene: https://stackoverflow.com/a/44124477.

celinval:
It looks good to me.

Can you please check what kind of error the user will get when we fail to generate a projection `Expr` that is used to generate other expressions? For example, when the projection is on an assignment like this:
```
let wrong.projection = value;
```
Thanks!

zhassan-aws:
> Can you please check what kind of error the user will get when we fail to generate a projection `Expr` that is used to generate other expressions?

I used the following small program as an example:
```rust
#[kani::proof]
fn main() {
    let mut x = 5;
    let y = &mut x;
    *y = 9;
    assert!(x == 9);
}

```
and replaced the implementation of `ProjectionElem::Deref` in https://github.com/model-checking/kani/blob/446167f8d81a33d0e3fa4de906fb6c9ef3c8a086/src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs#L297 by 
```rust
                Err(UnimplementedData {
                    operation: "deref".to_string(),
                    bug_url: "".to_string(),
                    goto_type: self.codegen_ty(inner_mir_typ).to_pointer(),
                    loc: Location::None,
```
With that, Kani reported a single failure:
```
RESULTS:
Check 1: unsupported_construct.1
         - Status: FAILURE
         - Description: "deref is not currently supported by Kani. Please post your example at "

Check 2: main.assertion.1
         - Status: UNDETERMINED
         - Description: "assertion failed: x == 9"
         - Location: deref.rs:6:5 in function main


SUMMARY: 
 ** 1 of 2 failed (1 undetermined)
Failed Checks: deref is not currently supported by Kani. Please post your example at 

VERIFICATION:- FAILED

** WARNING: A Rust construct that is not currently supported by Kani was found to be reachable. Check the results for more details.
```
and neither Kani nor CBMC produced any other warnings/errors. The `--gen-c` output also indicates that the assignment to `*y` was completed omitted:
```
// file /home/ubuntu/examples/deref.rs line 2 column 1 function main
struct Unit main(void)
{
  struct Unit var_0;
  int x;
  int *y;
  struct Unit var_3;
  _Bool var_4;
  int var_5;

bb0:
  ;
  x = 5;
  y = &x;
  /* deref is not currently supported by Kani. Please post your example at  */
  assert(0);
  __CPROVER_assume(0);
  nondet_0();
  var_5 = x;
  var_4 = var_5 == 9;
```
This is because the callers to `codegen_place` will not proceed with performing the assignment if it returned `Err`.

celinval (approved):
Thanks for checking!

Issue #1045, opened 2022-04-14
kani-verifier tracking issue


tedinski:
Followups to https://github.com/model-checking/kani/pull/1039

**Critical or urgent:**

* [ ] Create a `licenses-note.txt` file in our release bundle, and verify we're good to go on all licensing requirements.
* [ ] Rename `src/cargo-kani` to `src/kani-driver`. Integrate `kani-verifier` into the workspace instead of having to keep it separate because of conflicting binary names.
* [ ] Make `kani-verifier` the top-level crate in our repo, with all other crates being in its workspace underneath it. Resolves problem of having a README in this crate.
* [ ] Consider adding hash-checking as part of the download.
* [ ] #1036 
* [ ] #1037 

**Eventual:**

* [ ] Checking of system dependencies/versions as part of first-time setup. (python3, pip, rustup, curl/tar, ctags)
* [ ] We need a single source of truth for our dependency versions. For python packages, we currently duplicate this information into `kani-verifier` and also in our CI setup scripts. For CBMC, we're pulling from PATH instead of downloading it correctly.
* [ ] Refactor some of the command line helpers I wrote so they're not duplicated. (By having other tools depend on `kani-verifier` where it lives?)
* [ ] Ability to control verbosity of Kani first-time setup output.
* [ ] Handle upgrading/uninstalling Kani. Remove old Kani release bundles and Rustup toolchains.

Kixunil:
I noticed `kani setup` calls `curl` which isn't present on all systems. Would it be too much to depend on `reqwest` to avoid having to deal with that dependency?

tedinski:
It would be too much by default I think: we're really trying to keep the number of dependencies and build time for the `kani-verifier` crate to an absolute minimum.

I _think_ this would be acceptable as an optional feature? (I'm not sure how `cargo install` reacts to features that aren't enabled, but if a normal install wouldn't pull down all the dependencies, but `cargo install kani-verifier --locked --features downloader` did, that would be ok.)

Out of curiosity, what platform is this an issue with? I really expected Python to be the bigger problem, curl seems required by the default way of getting `rustup` even, so I previously thought it wouldn't be an issue! :)

Kixunil:
Yeah, I see the point. Disabled features are not compiled at all, IIRC not even downloaded. So optional feature could be fine.

> what platform is this an issue with?

Actually even something quite popular: Debian stable. It has `wget` by default though.

> curl seems required by the default way of getting `rustup`

[...]

PR #1058, opened 2022-04-15
Rejoin kani-verifer into the overall workspace

tedinski:
### Description of changes: 

1. Rejoin `kani-verifier` into the overall workspace.
2. This exposed a bug where the PATH was being amended in compile test incorrectly and it's had `target/debug` in PATH before `scripts`. I solved this by borrowing "bookrunner's" approach of amending its own path once correctly, versus the existing compiletest approach of amending the path (erroneously, now) with each invocation.
3. I also added a mitigation to prevent that sort of bug causing dev environments to spam hundreds of http requests to github.

### Resolved issues:

towards #1045 

### Testing:

* Is this a refactor change? sort of, but it exposed some bugs

### Checklist


src/kani-verifier/src/lib.rs
```diff
@@ -70,6 +71,24 @@ fn appears_setup() -> bool {
...
+/// In dev environments, this proxy shouldn't be used.
+/// But accidentally using it (with the test suite) can fire off
+/// hundreds of HTTP requests trying to download a non-existent release bundle.
+/// So if we positively detect a dev environment, raise an error early.
+fn fail_if_in_dev_environment() -> Result<()> {
+    if let Ok(exe) = std::env::current_exe() {
+        if let Some(path) = exe.parent() {
+            if path.ends_with("target/debug") || path.ends_with("target/release") {
+                bail!(
+                    "Running the proxies from a development environment? These are just meant for releases."
```
zhassan-aws, line 83: Can we clarify the error message a bit? Example: "Running a release-only executable, \<executable-name\>, from a development environment. Please remove executable from path and rerun."

zhassan-aws (commented):
LGTM. The `releasebundle` check in CI failed though.

.dockerignore
```diff
@@ -1,4 +1,5 @@
 .git
 firecracker
-target
+target/debug
```
celinval, line 3: Why do we need to change this?

src/kani-verifier/src/lib.rs
```diff
@@ -58,6 +58,7 @@ pub fn proxy(bin: &str) -> Result<()> {
             setup(None)
         }
     } else {
+        fail_if_in_dev_environment()?;
```
celinval, line 61: How are we supposed to test this crate locally?

celinval (commented):
Is it possible to use `cargo package` and `cargo install --path` in the CI to verify the installation behavior?

.dockerignore
tedinski: The Dockerfile is now copying `target/release/{kani,cargo-kani}` into the container.

src/kani-verifier/src/lib.rs
tedinski: We don't except by using the Dockerfile.

tedinski:
> Is it possible to use `cargo package` and `cargo install --path` in the CI to verify the installation behavior?

Hmm, possibly. I will try this out and see.

.dockerignore
tedinski: Updated: Now `cargo install`ing the package, so we ignore `target/{release,debug}`, but need `target/package`

celinval (approved):
Nice. Thanks

Issue #674, opened 2021-12-02
Panic due to unwrapping ` None` during codegen of tokio crate


zhassan-aws:
Steps to reproduce:
1. `cargo new foo`
2. `cd foo`
3. Add `tokio` to the dependencies in `Cargo.toml`:
 ```
[dependencies]
tokio = { version = "1.14.0", features = ["full"] }
```
4. Run:
```
RUST_BACKTRACE=1 RUSTFLAGS=$(rmc-rustc --rmc-flags) RUSTC=$(rmc-rustc --rmc-path) cargo build --target x86_64-unknown-linux-gnu -j1
```

with RMC version: `7d0d4cfc6da`

I expected to see this happen: Verification successful

Instead, this happened:
```
thread 'rustc' panicked at 'called `Option::unwrap()` on a `None` value', compiler/rustc_codegen_rmc/src/codegen/typ.rs:1194:91
stack backtrace:
   0: rust_begin_unwind
             at /home/ubuntu/tools/rmc/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /home/ubuntu/tools/rmc/library/core/src/panicking.rs:107:14
   2: core::panicking::panic
             at /home/ubuntu/tools/rmc/library/core/src/panicking.rs:48:5
   3: <rustc_codegen_rmc::context::goto_ctx::GotocCtx>::fn_typ
   4: <rustc_codegen_rmc::context::goto_ctx::GotocCtx>::declare_function::{closure#0}
             at /home/ubuntu/tools/rmc/compiler/rustc_codegen_rmc/src/codegen/function.rs:194:17
   5: <rustc_codegen_rmc::context::goto_ctx::GotocCtx>::ensure::<<rustc_codegen_rmc::context::goto_ctx::GotocCtx>::declare_function::{closure#0}, &alloc::string::String>
             at /home/ubuntu/tools/rmc/compiler/rustc_codegen_rmc/src/context/goto_ctx.rs:135:23
   6: <rustc_codegen_rmc::context::goto_ctx::GotocCtx>::declare_function
             at /home/ubuntu/tools/rmc/compiler/rustc_codegen_rmc/src/codegen/function.rs:190:9
   7: <rustc_codegen_rmc::compiler_interface::GotocCodegenBackend as rustc_codegen_ssa::traits::backend::CodegenBackend>::codegen_crate::{closure#0}
[...]

PR #1061, opened 2022-04-18
Borrow the arguments of the print macros to avoid moving them

### LINKED ISSUE #1052: The overriden print macros take ownership of the arguments
I tried this code:

```rust
#[derive(Debug)]
struct Foo {
    x: i32,
}

#[kani::proof]
fn main() {
    let foo = Foo { x: 5 };
    println!("{:?}", foo);
    let bar = foo;
}
```

```
kani mv.rs
```

with Kani version: c65b3ec7477

I expected to see this happen: Verification successful

Instead, this happened: 
```
error[E0382]: use of moved value: `foo`
 --> mv.rs:9:15
  |
7 |     let foo = Foo { x: 5 };
  |         --- move occurs because `foo` has type `Foo`, which does not implement the `Copy` trait
8 |     println!("{:?}", foo);
  |                      --- value moved here
9 |     let bar = foo;
  |               ^^^ value used here after move
```

zhassan-aws:
### Description of changes: 

Kani's overridden versions of the print macros may result in moving the arguments. This could cause borrow-after-move compiler errors. This PR fixes the print macros by ensuring that the arguments are borrowed (as is the case with the std versions), to avoid this error.

### Resolved issues:

Resolves #1052 and fixes the error in https://github.com/model-checking/kani/issues/674#issuecomment-1069730955

### Testing:

* How is this change tested? One test added

### Checklist

### DIFF
--- a/library/std/src/lib.rs
+++ b/library/std/src/lib.rs
@@ -103,7 +103,7 @@ macro_rules! evaluate_print_args {
         // Evaluate each of the arguments since they may have side effects
         {
             $(
-                $arg;
+                let _ = &$arg;
             )*
         }
     };
--- a/tests/kani/Print/arg_not_moved.rs
+++ b/tests/kani/Print/arg_not_moved.rs
@@ -1,0 +1,20 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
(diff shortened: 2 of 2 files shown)


tedinski (approved):
> (as is the case with the std versions)

Do they do it the same way? If so, LGTM

zhassan-aws:
> Do they do it the same way? If so, LGTM

Effectively yes. A more detailed explanation is given here: https://stackoverflow.com/a/30451360

PR #1062, opened 2022-04-18
Fix codegen of never return functions

### LINKED ISSUE #648: Codegen does not generate calls to functions that return the never type
Example:

```rust
#![feature(never_type)]
pub fn foo() -> ! {
    panic!();
}
```
The associated function hook (`output_of_instance_is_never`) needs to be removed.

### LINKED ISSUE #751: Panic for `assert_inhabited` has no effect because of `Nevers` hook
Related to #648.

Instead of executing the code for `assert_inhabited`, the `Nevers` hook is being applied. This causes the test introduced in #742 to fail with an unexpected property. The `Nevers` hook has to be removed at some point, this issue is for keeping track of that and going back to ensure the test is working as expected.

celinval:
### Description of changes: 

I have removed the Nevers hook and I have also introduced two stubs in our std library to provide more user friendly errors. Functions overriden: std::process::{abort, exit}

### Resolved issues:

Resolves #648 and resolves #751.

### Call-outs:

I did add a tiny fix to a wrong PropertyClass that I noticed while working on this fix.

### Testing:

### Checklist


library/std/src/lib.rs
```diff
@@ -12,6 +12,9 @@
 // re-export all std symbols
 pub use std::*;
 
+// Override process calls with stubs.
+pub mod process;
```
tedinski, line 16: I vaguely recall us having issues with overriding things here that weren't macros. Was that resolved somehow?

src/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -57,9 +57,10 @@ impl<'tcx> GotocCtx<'tcx> {
                 "reached intrinsic::abort",
                 span,
             ),
+            // Transmuting to an uninhabited type is UB.
             "transmute" => self.codegen_fatal_error(
-                PropertyClass::UnsupportedConstruct,
-                "transmuting to uninhabited type",
+                PropertyClass::DefaultAssertion,
+                "transmuting to uninhabited type has undefined behavior",
```
tedinski, line 63: Ah, I made the wrong call helping Jai. Good catch. :)

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -368,22 +382,16 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                    return self.codegen_assert_false(
-                        PropertyClass::SanityCheck,
-                        &format!("reach some nonterminating function: {:?}", func),
-                        loc.clone(),
-                    );
-                }
-
-                let (p, target) = destination.unwrap();
+                let (place, target) =
+                    if let Some((p, t)) = destination { (Some(p), Some(t)) } else { (None, None) };
```
tedinski, line 386: Nit: This boggled me. Can we convert to match instead of if let?

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -481,7 +493,11 @@ impl<'tcx> GotocCtx<'tcx> {
 
         // Virtual function call and corresponding nonnull assertion.
         let call = fn_ptr.dereference().call(fargs.to_vec());
-        let call_stmt = self.codegen_expr_to_place(place, call).with_location(loc.clone());
+        let call_stmt = if let Some(place) = place {
+            self.codegen_expr_to_place(place, call).with_location(loc.clone())
+        } else {
+            call.as_stmt(loc.clone())
+        };
```
tedinski, line 500: Nit: here too

tests/expected/abort/main.rs
```diff
@@ -14,7 +14,7 @@ fn main() {
         }
         if i == 2 {
             // This should never happen.
-            process::abort();
+            process::exit(1);
```
tedinski, line 17: The comment says test `abort` but now we're calling `exit`. Is there a reason for choosing either?

tedinski (approved):
I think just a style thing: I prefer we use `match` instead of an `if let` as an expression.

library/std/src/lib.rs
zhassan-aws: The one I recall failing to override  was `std::mem::swap`, which turned out was because we have a hook for it: https://github.com/model-checking/kani/blob/e20874a616c0e2726726d4e23143ce77077ea7e2/src/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs#L390

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
celinval: I can. But why did it bother you?

tests/expected/abort/main.rs
celinval: I just wanted to increase test coverage

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
tedinski: I think my brain treats `if let` as statement-like and `match` as expression-like. This is despite being happy with `if` as expression-like.

It's possible it's just me needing to get used to it. If you disagree, no problem.

tests/expected/abort/expected
```diff
@@ -1,4 +1,16 @@
...
-a panicking function std::process::abort is invoked
+Status: UNREACHABLE\
+Description: ""This is unreachable""\
+in function main
+
+Status: FAILURE\
+Description: "Function abort() was invoked"\
+in function std::process::abort
+
+Status: SUCCESS\
```
zhassan-aws, line 9: I was expecting this to be UNREACHABLE, but it isn't because `std::exit` calls `panic` which we currently don't override. If we want it to be reported as UNREACHABLE (I think it should, but I'm not sure), then we can change the `panic(msg)` to `assert!(false, msg)`.

src/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -277,30 +252,6 @@ impl<'tcx> GotocHook<'tcx> for Panic {
     }
 }
 
-struct Nevers;
-
-impl<'tcx> GotocHook<'tcx> for Nevers {
```
zhassan-aws, line 282: Nice to get rid of this!

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
zhassan-aws: It took me some time to get used to as well :)

library/std/src/process.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This module introduces stubs for process methods.
```
zhassan-aws, line 4: Nice!

tests/expected/abort/expected
celinval: We can't. It won't compile because of the never return. I think we should instrument panic inside the compiler to have a reach check.

zhassan-aws: I see.

zhassan-aws (approved):

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
celinval: If you don't mind, I prefer to move on with this PR since there's nothing else to change.

I don't have any strong opinion about this. I tend to favor `if let` when I'm looking for one specific match. It's less verbose and clear to me that it's an all or nothing situation. We can discuss this later and if one is preferred over the other, maybe we can have a little convention page.

adpaco-aws (approved):
Thanks for getting rid of the `Nevers` hook!

Can you please check if this fixes #751 too?

src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
tedinski: Yeah, that's fine, I prefer velocity over this nit.

celinval: Thanks

celinval:
> Thanks for getting rid of the `Nevers` hook!
> 
> Can you please check if this fixes #751 too?

Yes it does.

PR #1066, opened 2022-04-19
Codegen unimplemented for the array subslice pattern

zhassan-aws:
### Description of changes: 

Kani currently panics if the program uses the array subslice pattern (see #707). With this PR, Kani will now codegen unimplemented (assert/assume false), which will result in verification failure with an informative message if it's reachable.

### Resolved issues:

Resolves the crash in #707 (but doesn't add support for the array subslice pattern).

### Testing:

* How is this change tested? One test added.

### Checklist

### DIFF
--- a/src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
+++ b/src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
@@ -416,7 +416,25 @@ impl<'tcx> GotocCtx<'tcx> {
             ProjectionElem::Subslice { from, to, from_end } => {
                 // https://rust-lang.github.io/rfcs/2359-subslice-pattern-syntax.html
                 match before.mir_typ().kind() {
-                    ty::Array(..) => unimplemented!(),
+                    ty::Array(ty, len) => {
+                        let len = len.val().try_to_machine_usize(self.tcx).unwrap();
+                        let subarray_len = if from_end {
+                            // `to` counts from the end of the array
+                            len - to - from
+                        } else {
+                            to - from
+                        };
+                        let typ = self.tcx.mk_array(*ty, subarray_len);
+                        let goto_typ = self.codegen_ty(typ);
+                        // unimplemented
+                        Err(UnimplementedData {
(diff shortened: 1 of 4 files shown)


src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -416,7 +416,25 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    ty::Array(ty, len) => {
+                        let len = len.val().try_to_machine_usize(self.tcx).unwrap();
+                        let subarray_len = if from_end {
+                            // `to` counts from the end of the array
+                            len - to - from
+                        } else {
+                            to - from
+                        };
+                        let typ = self.tcx.mk_array(*ty, subarray_len);
+                        let goto_typ = self.codegen_ty(typ);
```
adpaco-aws, line 428: Does it make sense to add this code if we will fail anyway?

adpaco-aws (commented):
I recommend to update the [Rust support table](https://model-checking.github.io/kani/rust-feature-support.html) (row "Patterns") to link this issue. Feel free to remove the "Needs more testing".

src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
zhassan-aws: My thinking is to keep the code that determines the type because:
1. `codegen_unimplemented` takes the goto type as an argument, so we need to pass something in. The actual type that we pass in shouldn't matter though since `codegen_unimplemented` generates an `assume(false)` after the `assert(false)`, but it made sense for me to pass the correct type in to be consistent with the rest of the code base.
2. We should soon start working on adding support for the array subslice pattern, so having the type available would save some effort.

zhassan-aws:
> I recommend to update the [Rust support table](https://model-checking.github.io/kani/rust-feature-support.html) (row "Patterns") to link this issue. Feel free to remove the "Needs more testing".

Done.

adpaco-aws (approved):

tedinski (approved):

PR #1071, opened 2022-04-19
Do not create a ProjectedPlace for unsupported fields

zhassan-aws:
### Description of changes: 

With the mechanism for handling unsupported scenarios in projection code introduced in #1057, it is now possible to avoid creating a projected place for unsupported fields. This PR introduces a similar mechanism for handling unsupported fields in `codegen_field` that propagates an error upstream.

### Testing:

### Checklist

### DIFF
--- a/src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
+++ b/src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
@@ -39,6 +39,17 @@ pub struct UnimplementedData {
     pub loc: Location,
 }
 
+impl UnimplementedData {
+    pub fn new(operation: &str, bug_url: &str, goto_type: Type, loc: Location) -> Self {
+        UnimplementedData {
+            operation: operation.to_string(),
+            bug_url: bug_url.to_string(),
+            goto_type,
+            loc,
+        }
+    }
+}
+
 /// Relevent information about a projected place (i.e. an lvalue).
 #[derive(Debug)]
 pub struct ProjectedPlace<'tcx> {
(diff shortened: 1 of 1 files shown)


celinval (commented):
Is there any test case for this?

zhassan-aws:
This test: https://github.com/model-checking/kani/blob/3c008db98f2559b131f2b10ba355ee73d6dd4943/tests/kani/Generator/main.rs
has a generator. I'm not sure it's hitting this case though. I'll check.

zhassan-aws:
I found out that the `assert(false)` that is generated in https://github.com/model-checking/kani/blob/3c008db98f2559b131f2b10ba355ee73d6dd4943/tests/kani/Generator/main.rs is due to this block: https://github.com/model-checking/kani/blob/3c008db98f2559b131f2b10ba355ee73d6dd4943/src/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs#L611
and not the one in `place.rs`. However, my attempts at creating a test that covers the line of code in `place.rs`, which I think requires having a struct with a generator field, e.g.:
```rust
#![feature(generators, generator_trait)]

use std::ops::{Generator};

struct HasGeneratorField<T> 
where T: Generator {
    _g: T,
}

impl<T> HasGeneratorField<T> 
where T: Generator {
    pub fn new(g: T) -> Self {
        HasGeneratorField { _g: g }
    }
}

//#[kani::proof]
fn main() {
    let _s = HasGeneratorField::new(|| { yield 1; return 2; });
}
```
all resulted in a crash that I filed as #1075.

So I think this may have to stay untested till the other issue is fixed.

celinval (approved):
Thanks for trying to add a test. I think this is still beneficial as is.

PR #1073, opened 2022-04-20
Fix handling of `dyn T`

celinval:
### Description of changes: 

We were incorrectly handling `dyn T` as well as unsized structures with `dyn T`. This change fixes that.

### Resolved issues:

Resolves #82
Resolves #990
Mitigates #1072

### Call-outs:

1. This fixes many of the warnings from #277 but not all.
2. This fixes an issue with Rc<dyn T> and many other issues with Struct<dyn T>. I didn't find any issue related to them.
3. ~~There is a hack that people must be aware. The initial goal was to encode `dyn T` as a `typedef void` so the fat pointer would be in practice `void*`. However, this doesn't work well when the `dyn T`. The reference to this field is usually of the form `&*struct.unsized_field` which generates something like `&void` which is invalid. I am using `bool` for now to encode `dyn T`.~~
4. The `drop` logic still needs further adjustment (see #1072). I added a mitigation to capture the behavior that we don't support yet.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist
- [x] Each commit message has a non-empty body, explaining why the change was made
- [x] Methods or procedures are documented
- [x] Regression or unit tests are included, or existing tests cover the modified code
- [x] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

Issue #277, opened 2021-06-30
Fix casting warnings in `Projection::new`


avanhatt:
There is a commented-out assertion with a TODO in `Projection::new`: the assertion fails on some regression tests; but also on some standard library issues that would be useful to have an actual check for. Once we fix the regressions we should enable it.

```rust
        // TODO: these assertions fail on a few regressions. Figure out why.
        // I think it may have to do with boxed fat pointers.
        // assert!(
        //     Self::check_expr_typ(&goto_expr, &mir_typ_or_variant, ctx),
        //     "\n{:?}\n{:?}",
        //     &goto_expr,
        //     &mir_typ_or_variant
        // );
```

avanhatt:
Example: boxed closures, like this:

```rust
   // Create a boxed once-callable closure
    let f: Box<dyn FnOnce()> = Box::new(|| {});
    f();
```

tautschnig:
Not that I have a solution here as to why "Pointer" and "FlexibleArray" seem to be confused here, but I just wanted to note that CBMC now takes a `#flexible_array_member` annotation on flexible array members.

celinval:
From what I gather so far, it looks like we are incorrectly using Struct { Pointer } when adding the symbol to the symbol table. Trying to pin down where this is happening.

celinval:
Oh, I think I found it. :)

https://github.com/model-checking/kani/blob/e6fb98834a72a30c80875da9d42de0e790fcdcd7/src/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs#L770

It looks like we are assuming in a lot of places in this file that a slice is encoded as:
```rust
struct<T> Slice{
   data: const* T,
   len: usize,
}
 ```
But in the following line we expect a FlexibleArray:
https://github.com/model-checking/kani/blob/e6fb98834a72a30c80875da9d42de0e790fcdcd7/src/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs#L542

PR #1086, opened 2022-04-21
Inject runtime error for type mismatch in projection

celinval:
### Description of changes: 

This is to mitigate potentially wrong code generation detected by our type mismatch logic. Eventually we want to enable this assertion and fail at runtime, but this currently affects the compilation of std.

Note: I had to fix the projection error handling for codegen_place. This wasn't working if the failure was in the middle of the projection.

### Resolved issues:

Mitigates #277 

### Call-outs:

The regression will fail until we merge #1073. I was able to verify that once we merge that, this should pass. I won't push until then but I could use some early feedback. 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -138,13 +138,13 @@ impl<'tcx> ProjectedPlace<'tcx> {
...
         goto_expr: Expr,
         mir_typ_or_variant: TypeOrVariant<'tcx>,
         fat_ptr_goto_expr: Option<Expr>,
         fat_ptr_mir_typ: Option<Ty<'tcx>>,
         ctx: &mut GotocCtx<'tcx>,
-    ) -> Self {
+    ) -> Result<Self, UnimplementedData> {
         let mir_typ_or_variant = mir_typ_or_variant.monomorphize(ctx);
         let fat_ptr_mir_typ = fat_ptr_mir_typ.map(|t| ctx.monomorphize(t));
         if let Some(fat_ptr) = &fat_ptr_goto_expr {
```
danielsn, line 150: Should this also be a return error?

danielsn (approved):

src/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
celinval: Not sure where you are pointing at.

danielsn: We have a mix of `panic` and `Error` in the same function. Was wondering if we should standardize on `Error`

celinval: I don't think so. Panic is stronger and we should keep doing that if something is really unexpected.

That said, I would be OK compromising and turning those things into debug_assert!() for now so they at least fail in our regression but in the user code, it results in a unsupported_construct.

In the long run, I don't like pushing errors downstream and we should provide stronger guarantees that our compiler works. Everything in this function should eventually trigger an internal compiler error.

PR #1088, opened 2022-04-21
Replace the hook for realloc with a C implementation

### LINKED ISSUE #702: Bogus failures in simple vector example
I tried this code:

```rust
fn main() {
    let mut v = Vec::new();
    v.push(0);
    v.push(0);
    v.push(0);
    v.push(0);
    v.push(0);
    assert!(v[0] == 0);
    assert!(v[1] == 0);
    assert!(v[2] == 0);
    assert!(v[3] == 0);
}
```

```
rmc test.rs
```

with RMC version: `8de5cd67609`

I expected to see this happen: Verification successful

Instead, this happened: All four assertions failed:
```
[main.assertion.1] line 8 assertion failed: v[0] == 0: FAILURE
[main.assertion.2] line 9 assertion failed: v[1] == 0: FAILURE
[main.assertion.3] line 10 assertion failed: v[2] == 0: FAILURE
[main.assertion.4] line 11 assertion failed: v[3] == 0: FAILURE
** 4 of 545 failed (2 iterations)
VERIFICATION FAILED
```
If I remove one of the `push` calls, verification passes:
```
[main.assertion.1] line 8 assertion failed: v[0] == 0: SUCCESS
[main.assertion.2] line 9 assertion failed: v[1] == 0: SUCCESS
[main.assertion.3] line 10 assertion failed: v[2] == 0: SUCCESS
[main.assertion.4] line 11 assertion failed: v[3] == 0: SUCCESS

** 0 of 545 failed (1 iterations)
VERIFICATION SUCCESSFUL
```
I suspect this is a bug in reallocation; if I print the vector's capacity after each push, I see that it changes from 0 to 4 after the first push (initial allocation) and changes from 4 to 8 after the last push (reallocation).

zhassan-aws:
### Description of changes: 

Kani currently has a hook for the low-level `__rust_realloc` function (see [here](https://doc.rust-lang.org/src/alloc/alloc.rs.html#34)) that calls CBMC's built-in `realloc`. CBMC's implementation of realloc (see [here](https://github.com/diffblue/cbmc/blob/6253abd1370c98a3afee1350ec8c977bb6e388a4/src/ansi-c/library/stdlib.c#L472)) seems to be causing an issue when it is called via the `RawVec` data structure, which is the underlying data structure for `Vec`. This PR replaces the hook with a C implementation of `__rust_realloc`. Providing an implementation for `__rust_realloc` is desirable since the function has more parameters than the C version (the old size and the alignment) that should enable a more reliable and efficient implementation. 

### Resolved issues:

Resolves #702 

### Call-outs:

It might be better to implement `__rust_realloc` directly in goto as a hook, or in Rust in Kani's `std` overrides library, but we're going with a C implementation for now since it's the easiest.

### Testing:

* How is this change tested? Test that was fixed. **Need to add more tests before turning this into a PR**

### Checklist


library/kani/kani_lib.c
```diff
@@ -1,2 +1,33 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+// This is a C implementation of the __rust_realloc function that has the following signature
+//     fn __rust_realloc(ptr: *mut u8, old_size: usize, align: usize, new_size: usize) -> *mut u8;
+// The implementation pretty much follows CBMC's implementation:
+// https://github.com/diffblue/cbmc/blob/94b81b22799876b52632ba9cdc3f486f87abadb8/src/ansi-c/library/stdlib.c#L472
+uint8_t* __rust_realloc(uint8_t* ptr, size_t old_size, size_t align, size_t new_size) {
+    // if current ptr is NULL, this behaves like malloc
+    if (ptr == 0)
```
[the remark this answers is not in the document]
zhassan-aws: Done

celinval (commented):
I have a conceptual question. Why can't this be implemented in rust instead?

library/kani/kani_lib.c
```diff
@@ -1,2 +1,37 @@
...
+#include <assert.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+// This is a C implementation of the __rust_realloc function that has the following signature
+//     fn __rust_realloc(ptr: *mut u8, old_size: usize, align: usize, new_size: usize) -> *mut u8;
+// The implementation pretty much follows CBMC's implementation:
+// https://github.com/diffblue/cbmc/blob/94b81b22799876b52632ba9cdc3f486f87abadb8/src/ansi-c/library/stdlib.c#L472
```
celinval, line 13: Can you please clarify in the comments why we need our own?

library/kani/kani_lib.c
```diff
@@ -1,2 +1,37 @@
...
+uint8_t *__rust_realloc(uint8_t *ptr, size_t old_size, size_t align, size_t new_size)
+{
+    // if current ptr is NULL, this behaves like malloc
+    if (ptr == 0) {
```
celinval, line 17: should this be an assert?

library/kani/kani_lib.c
```diff
@@ -1,2 +1,37 @@
...
+        assert(old_size == 0);
+        return malloc(new_size);
+    }
+
+    // if malloc-size is 0, free original object and return malloc(0), which in
+    // CBMC returns a pointer to a zero-sized object
+    if (new_size == 0) {
```
celinval, line 24: same here... `assert (new_size > 0)`

library/kani/kani_lib.c
```diff
@@ -1,2 +1,37 @@
...
+        free(ptr);
+        return malloc(0);
+    }
+
+    uint8_t *result = malloc(new_size);
+    if (result) {
```
celinval, line 30: should we special case this if new_size == old_size?

zhassan-aws:
> I have a conceptual question. Why can't this be implemented in rust instead?

I made an attempt in https://github.com/zhassan-aws/kani/blob/realloc-stub/library/std/src/alloc.rs, but it didn't work because `is_null` is a std library function that we're not linking :(

library/kani/kani_lib.c
zhassan-aws: Done

library/kani/kani_lib.c
zhassan-aws: Changed

library/kani/kani_lib.c
zhassan-aws: Done

library/kani/kani_lib.c
zhassan-aws: I haven't seen that done in any of the implementations I looked at. For performance, it seems desirable to check for this special case, but for verification purposes, I'm not sure if it somehow complicates the logic. Something to experiment with...

celinval: Sounds good

celinval (approved):
Thanks

celinval:
> > I have a conceptual question. Why can't this be implemented in rust instead?
> 
> I made an attempt in https://github.com/zhassan-aws/kani/blob/realloc-stub/library/std/src/alloc.rs, but it didn't work because `is_null` is a std library function that we're not linking :(

Let's explore that later. Thanks

PR #1090, opened 2022-04-21
Audit for `ptr_guaranteed_<cmp>`

adpaco-aws:
### Description of changes: 

Completes the audit for `ptr_guaranteed_<cmp>`. These behave as regular pointer comparison at runtime.

### Resolved issues:

Part of #727 

### Testing:

* How is this change tested? Adds 2 tests.

### Checklist


docs/src/rust-feature-support.md
```diff
@@ -371,8 +371,8 @@ prefetch_read_data | No | |
 prefetch_read_instruction | No | |
 prefetch_write_data | No | |
 prefetch_write_instruction | No | |
-ptr_guaranteed_eq | Partial | |
-ptr_guaranteed_ne | Partial | |
+ptr_guaranteed_eq | Yes | Over-approximation: Returns nondet. value even if comparison is true |
+ptr_guaranteed_ne | Yes | Over-approximation: Returns nondet. value even if comparison is true |
```
danielsn, line 375: Should one of these say "false"

tests/kani/Repr/main_fixme.rs
```diff
@@ -17,5 +17,8 @@ fn mmap() -> *mut MyCVoid {
 fn main() {
     let v = mmap();
     assert!(v != MAP_FAILED);
+    // The assertion below fails because it must be using `ptr_guaranteed_eq` or
```
danielsn, line 20: Why must it?

tests/kani/Intrinsics/PtrGuaranteedCmp/ne_ok.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Checks that `ptr_guaranteed_ne` returns false when the pointers are
+// equal, which causes the test to pass in the `else` branch. The `if`
+// branch is unreachable.
```
danielsn, line 6: Should we test that this branch is unreachable, maybe with an assert false?

docs/src/rust-feature-support.md
adpaco-aws: No, these are not symmetric in the sense that if one returns true, the other one returns false. Under some circumstances, both may return false, which is the desired behavior. See https://doc.rust-lang.org/std/primitive.pointer.html#method.guaranteed_eq for more details.

tests/kani/Intrinsics/PtrGuaranteedCmp/ne_ok.rs
adpaco-aws: Because we are assuming `ptr1 == ptr2`, the unreachable assertion is in practice `assert!(false)`.

adpaco-aws:
The PR is ready to review again.

We had a discussion where we agreed that these intrinsics should be implemented as regular comparison instead of being over-approximated, otherwise we get nondet. results for comparison like `ptr::is_null`. This has been noted in both the Kani documentation and code.

tests/kani/Intrinsics/PtrGuaranteedCmp/eq.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-verify-fail
```
zhassan-aws, line 3: Is `kani-verify-fail` required?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -246,9 +246,17 @@ impl<'tcx> GotocCtx<'tcx> {
...
         }
 
-        // Intrinsics which encode a simple binary operation
-        macro_rules! codegen_intrinsic_boolean_binop {
-            ($f:ident) => {{ self.binop(p, fargs, |a, b| a.$f(b).cast_to(Type::c_bool())) }};
+        // Intrinsics which encode a pointer comparison (e.g., `ptr_guaranteed_eq`).
+        //
+        // Note(std): These intrinsics cannot be over-approximated because they
+        // are used for regular comparison in `std` when they shouldn't. An earlier
+        // version that returned nondet. values when the result of the comparison
```
zhassan-aws, line 253: > because they are used for regular comparison in `std` when they shouldn't

This sounds a bit too strong for me. The documentation (https://doc.rust-lang.org/beta/std/primitive.pointer.html#method.guaranteed_eq) says that this may spuriously return `false` in **compile-time evaluation**, which I understood to mean that it may block some optimizations, etc., but it can never impact correctness since it behaves like `==`/`!=` at run-time (which is what we're modeling).

tests/kani/Intrinsics/PtrGuaranteedCmp/ne.rs
```diff
@@ -0,0 +1,15 @@
...
+// Checks that `ptr_guaranteed_ne` returns true when both pointers
+// are different.
+#![feature(core_intrinsics)]
+use std::intrinsics::ptr_guaranteed_ne;
+
+#[kani::proof]
+fn test_ptr_eq(ptr1: *const u8, ptr2: *const u8) {
+    kani::assume(ptr1 != ptr2);
+    assert!(ptr_guaranteed_ne(ptr1, ptr2));
+    kani::expect_fail(ptr1 == ptr2, "Pointers aren't equal");
```
danielsn, line 14: Should this use `!ptr_guaranteed_ne`? And will it even be reachable given how we turn `assert` into `assert; assume`

tests/kani/Intrinsics/PtrGuaranteedCmp/eq.rs
adpaco-aws: It was not needed, but I changed the tests to have two positive harnesses instead.

tests/kani/Intrinsics/PtrGuaranteedCmp/ne.rs
adpaco-aws: I changed the tests to have (each) two positive harnesses instead, please check.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: Thanks for making this clear! Removed the comments here and in Kani docs.

zhassan-aws (approved):
Thanks @adpaco-aws!

adpaco-aws:
Merging this PR now, but if you have any other concerns please let me know and I'll take care of them.

Issue #920, opened 2022-03-10
Reading uninitialized memory gives symbolic value rather than error


nchong-at-aws:
In the following program we read from uninitialized memory so the program is undefined [0]. Using CBMC as our symbolic engine means Kani does not warn of this error. Instead the value of `v[0]` is symbolic/nondet. Adding a "poison" value for freshly allocated objects is likely to be a large change for CBMC. For now, we can document this behavior.

```rust
fn main() {
    let mut v: Vec<u8> = Vec::with_capacity(8);
    unsafe {
        v.set_len(3);
    }
    let _b = v[0]; //< reading uninitialized memory
}
```

[0] https://doc.rust-lang.org/reference/behavior-considered-undefined.html#behavior-considered-undefined
> Uninitialized memory is also implicitly invalid for any type that has a restricted set of valid values. In other words, the only cases in which reading uninitialized memory is permitted are inside unions and in “padding” (the gaps between the fields/elements of a type).

nchong-at-aws:
I think the quoted para isn't appropriate to the example since all values of 0..255 are okay for a u8. Instead, I now think the cause is the line:

> An integer (i*/u*), floating point value (f*), or raw pointer obtained from [uninitialized memory](http://llvm.org/docs/LangRef.html#undefined-values), or uninitialized memory in a str.

zhassan-aws:
Related to https://github.com/diffblue/cbmc/issues/7014

giltho:
It turns out that Kani's own test suite has undefined behaviour because of that, with a mix of that and the `any_raw` issue, pointed out [there](https://github.com/model-checking/kani/issues/1394) and partially fixed [there](https://github.com/model-checking/kani/pull/1415).

Take the test [result3.rs](https://github.com/model-checking/kani/blob/main/tests/kani/Enum/result3.rs).
It's initialized with `kani::any_raw`, which means that both the discriminent field and value field are nondet.

Then `foo` is called, with the any_raw value.
And here's the code generated by kani (in demangled c form):
```c
unsigned int foo(struct std::result::Result<u32, Unit> *input)
{
  unsigned int var_0;
  long int var_2;
  unsigned int *num;

bb0:
  ;
  var_2 = (long int)input->case;
  if(!(var_2 == 0))
    goto bb2;


bb1:
  ;
  num = &input->cases.Ok.0;
  var_0 = *num;
  goto bb3;

bb2:
  ;
  var_0 = 3;

bb3:
  ;
  return var_0;
}
[...]

PR #1091, opened 2022-04-22
Update `raw_eq` documentation and tests

### LINKED ISSUE #347: Extend handling of `raw_eq` intrinsic
#331 introduced code for a new `raw_eq` intrinsic but also future work for it. We should do layout comparison and previous failure check (soundness).

adpaco-aws:
### Description of changes: 

First, I'm not clear on why I wrote that layout comparison was a requirement. `raw_eq` is implemented as `memcmp` and the documentation does not mention layout anywhere:  https://doc.rust-lang.org/core/intrinsics/fn.raw_eq.html
What it does mention is that uninitialized memory in any of the arguments is undefined behavior.

This PR updates the notes on Rust feature support to mention uninitialized memory, which we had to do anyways because of #920. It also updates the comment on `raw_eq` and adds two new tests for checking the behavior with uninitialized data.

I think there is nothing else to be done with `raw_eq` until we can poison memory (see #920).

### Resolved issues:

Resolves #347 
Part of #727 

### Testing:

* How is this change tested? Adds two tests

### Checklist

### DIFF
--- a/docs/src/rust-feature-support.md
+++ b/docs/src/rust-feature-support.md
@@ -192,6 +192,16 @@ ensure there is no resource leak or persistent data inconsistency. Check out
 [this issue](https://github.com/model-checking/kani/issues/692) for updates on
 stack unwinding support.
 
+### Uninitialized memory
+
+Reading uninitialized memory is
+[considered undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#behavior-considered-undefined) in Rust.
+At the moment, Kani cannot detect if memory is uninitialized, but in practice
+this is mitigated by the fact that all memory is initialized with nondet.
+values.
+Therefore, any code that depends on uninitialized data will exhibit nondeterministic behavior.
+See [this issue](https://github.com/model-checking/kani/issues/920) for more details.
+
 ### Destructors
 
 At present, we are aware of some issues with destructors, in particular those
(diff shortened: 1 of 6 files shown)


tests/kani/Intrinsics/RawEq/uninit_ne.rs
```diff
@@ -0,0 +1,18 @@
...
+use std::intrinsics::raw_eq;
+use std::mem::{zeroed, MaybeUninit};
+
+#[kani::proof]
+fn main() {
+    let zeroed_arr: [u8; 8] = unsafe { zeroed() };
+    let uninit_arr: [u8; 8] = unsafe { MaybeUninit::uninit().assume_init() };
+
+    let arr_are_eq = unsafe { raw_eq(&zeroed_arr, &uninit_arr) };
+    assert!(!arr_are_eq);
```
tedinski, line 17: I'm not sure if this second test is necessary?

tedinski (approved):

docs/src/rust-feature-support.md
```diff
@@ -192,6 +192,16 @@ ensure there is no resource leak or persistent data inconsistency. Check out
 [this issue](https://github.com/model-checking/kani/issues/692) for updates on
 stack unwinding support.
 
+### Uninitialized memory
+
+Reading uninitialized memory is
+[considered undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#behavior-considered-undefined) in Rust.
+At the moment, Kani cannot detect if memory is uninitialized, but in practice
+this is mitigated by the fact that all memory is initialized with nondet.
```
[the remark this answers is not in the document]
adpaco-aws: Done.

tests/kani/Intrinsics/RawEq/uninit_ne.rs
adpaco-aws: It's expressing that `raw_eq` is also true in one trace, thus complementing the other test case.

PR #1094, opened 2022-04-22
Audit for `offset` intrinsic

### LINKED ISSUE #1082: Pointer sub() operation is giving spurious overflow check error
I tried this code:
```rust
#[kani::proof]
fn check_thin_ptr() {
    let array = [0, 1, 2, 3, 4, 5, 6];
    let second_ptr: *const i32 = &array[3];
    unsafe {
        let before = second_ptr.sub(1);
        assert_eq!(*before, 2);
    }
}
```

```
kani <test_file>
```

with Kani version:

I expected to see this happen: Verification succeed.

Instead, this happened: It fails with the following message:

```
SUMMARY:
 ** 1 of 9 failed
Failed Checks: attempt to compute offset which would overflow
 File: "rustlib/src/rust/library/core/src/ptr/const_ptr.rs", line 295, in std::ptr::const_ptr::<impl *const T>::offset

VERIFICATION:- FAILED
```

adpaco-aws:
### Description of changes: 

Changes the `offset` intrinsic to use addition with overflow to detect overflow, but computes the actual destination pointer with regular addition. This fixes spurious failures that appeared when offsets were negative or being substracted.

In addition, standard pointer checks cover the case where the pointer is dereferenced outside the object bounds. I've observed that this works as expected: One byte past the object is fine, more than that is not. 5 test cases included.

### Resolved issues:

Resolves #1082 
Part of #727 

### Testing:

* How is this change tested? Adds 5 test cases.

### Checklist


tests/kani/Intrinsics/Offset/offset_u8_fail.rs
```diff
@@ -0,0 +1,18 @@
...
+#![feature(core_intrinsics)]
+use std::intrinsics::offset;
+
+#[kani::proof]
+fn test_offset() {
+    let s: &str = "123";
+    let ptr: *const u8 = s.as_ptr();
+
+    unsafe {
+        let x = *offset(ptr, 4) as char;
```
celinval, line 16: nit: For cases like this, I recommend using the expected test suite. `kani-verify-fail` is very fragile.

Not a blocker though.

tests/kani/PointerOffset/fixme_overflow_ptr.rs
```diff
@@ -1,12 +0,0 @@
-// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-// SPDX-License-Identifier: Apache-2.0 OR MIT
```
celinval, line 2: I would prefer if we don't delete this one in favor of things that call the intrinsic directly. This is actually closer to what the user code would look like. Can you please just rename it?

celinval (commented):
Just small comments. Don't worry about moving the tests to expected, but it is something I recommend for the future.

adpaco-aws:
Thanks, I moved the negative tests to `expected` and restored the "fixme" test (sorry about that).

adpaco-aws:
Thanks, I moved the negative tests to `expected` and restored the "fixme" test (sorry about that).

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -874,6 +874,31 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        &mut self,
+        intrinsic: &str,
+        mut fargs: Vec<Expr>,
+        p: &Place<'tcx>,
+        loc: Location,
+    ) -> Stmt {
+        let src_ptr = fargs.remove(0);
+        let offset = fargs.remove(0);
+        // Check that the computation would not overflow an `isize`
+        let dst_ptr_of = src_ptr.clone().cast_to(Type::ssize_t()).add_overflow(offset.clone());
```
danielsn, line 889: This does integer arithmatic, not pointer arithmatic, and I think will give the wrong results

danielsn (changes requested):

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
celinval: I.e.: You should account for the pointee size.

adpaco-aws: Right, should be getting the right results now.

celinval (commented):
Can you add a test that exercise the new check? Something like:

```rust
/// Check that an offset that overflow an isize::MAX triggers a verification failure.
#[kani::proof]
unsafe fn check_wrap_offset() {
        let v : &[u128] = &[0; 200];
        let v_100 : *const u128 = &v[100];
        let max_offset = usize::MAX / std::mem::size_of::<u128>();
        let v_wrap : *const u128 = v_100.offset((max_offset + 1).try_into().unwrap());
        assert_eq!(v_100, v_wrap);
        assert_eq!(*v_100, *v_wrap);
}
```

danielsn (approved):
Approved modulo @celinval comments

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -868,6 +868,45 @@ impl<'tcx> GotocCtx<'tcx> {
...
+
+        // Check that computing `bytes` would not overflow
+        let ty = self.monomorphize(instance.substs.type_at(0));
+        let layout = self.layout_of(ty);
+        let size = Expr::int_constant(layout.size.bytes(), Type::ssize_t());
+        let bytes = offset.clone().mul_overflow(size);
+        let bytes_overflow_check = self.codegen_assert(
+            bytes.overflowed.not(),
+            PropertyClass::ArithmeticOverflow,
+            format!("{}: attempt to compute `bytes` which would overflow", intrinsic).as_str(),
```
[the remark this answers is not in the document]
adpaco-aws: Thanks, changed both messages to avoid `format!` :)

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -868,6 +868,45 @@ impl<'tcx> GotocCtx<'tcx> {
         Stmt::atomic_block(vec![skip_stmt], loc)
     }
 
+    // Computes the offset from a pointer
```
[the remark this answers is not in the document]
adpaco-aws: Done.

adpaco-aws:
Addressed all @celinval's comments, including the proposed test case.

PR #1099, opened 2022-04-23
Audit for `ptr_offset_from`

### LINKED ISSUE #1098: `ptr_offset_from` doesn't care if one of the pointers is out of bounds
Related to https://github.com/diffblue/cbmc/issues/6822

The assertion in `ptr_offset_from` based on `__CPROVER_same_object` to check that both pointers point to the same object returns true if any or both pointers point to addresses out of the objects. For example:

```rust
fn main() {
    let a = [0; 5];
    let b = [0; 5];
    let ptr1: *const u32 = &a[0];
    let ptr2 = unsafe { ptr1.add(10) };
    unsafe {
        let offset = ptr_offset_from(ptr2, ptr1);
    }
}
```
Passes even though `ptr2` is not pointing to `a` anymore.

adpaco-aws:
### Description of changes: 

Adds an arithmetic overflow check for `ptr_offset_from` and removes the `__CPROVER_same_object` check because CBMC applies it to the arithmetic operation by default.

As with `offset`, I had concerns because there were no checks for the pointers staying within bounds of the object. So I added `__CPROVER_r_ok` irep expression and used it, just to find that a bunch of regressions fail.

### Resolved issues:

Resolves #1098 
Part of #727 

### Testing:

* How is this change tested? Existing tests (I don't think the overflow case can occur in practice).

### Checklist


cprover_bindings/src/goto_program/expr.rs
```diff
@@ -873,6 +874,7 @@ impl Expr {
                 (lhs.typ == rhs.typ && lhs.typ.is_integer())
                     || (lhs.typ.is_pointer() && rhs.typ.is_integer())
             }
+            ROk => lhs.typ.is_pointer() && rhs.typ.is_c_size_t(),
```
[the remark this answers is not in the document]
danielsn: @tautschnig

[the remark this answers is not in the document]
adpaco-aws: In my first attempt I declared it as a unary operator but CBMC complained that two arguments were expected for `__CPROVER_r/w_ok`, so that is why I implemented this. I can remove if it's going to be a blocker.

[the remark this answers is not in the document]
danielsn: My concern had been more whether we should think of it as a binop, or as a CBMC intrinsic.  But we can merge this PR  and have a separate ticket to track that. https://github.com/model-checking/kani/issues/1111

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -878,26 +878,28 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                self.codegen_expr_to_place(p, a.sub(b)),
-            ],
-            loc,
-        )
+        // Check that the computation would not overflow an `isize`
+        let cast_dst_ptr = dst_ptr.clone().cast_to(Type::ssize_t());
+        let cast_src_ptr = src_ptr.clone().cast_to(Type::ssize_t());
+        let offset = cast_dst_ptr.sub_overflow(cast_src_ptr);
+        let overflow_check = self.codegen_assert(
+            offset.overflowed.not(),
```
danielsn, line 894: does this cause performance problems

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -878,26 +878,28 @@ impl<'tcx> GotocCtx<'tcx> {
     /// https://doc.rust-lang.org/std/intrinsics/fn.ptr_offset_from.html
```
danielsn, line 878: There are a bunch of checks, and we're only checking some of them
```
Both the starting and other pointer must be either in bounds or one byte past the end of the same [allocated object](https://doc.rust-lang.org/std/ptr/index.html#allocated-object).

Both pointers must be derived from a pointer to the same object. (See below for an example.)

The distance between the pointers, in bytes, must be an exact multiple of the size of T.

The distance between the pointers, in bytes, cannot overflow an isize.

The distance being in bounds cannot rely on “wrapping around” the address space.
```

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -878,26 +878,28 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                    pointers_to_same_object,
-                    PropertyClass::PointerOffset,
-                    "ptr_offset_from: pointers point to same object",
-                    loc.clone(),
-                ),
-                self.codegen_expr_to_place(p, a.sub(b)),
-            ],
-            loc,
-        )
+        // Check that the computation would not overflow an `isize`
```
danielsn, line 889: Why did you remove the `same_object` check

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: I couldn't come up with a test for this overflow check and then I realized it's wrong. The overflow check has to be done for the distance in bytes, which is what we do now. An `expected` test for the overflow case is included.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: The `same_object` check is being done by CBMC by default. I converted the test we already had for it into an `expected` test so we ensure the "same object violation" stays there.

adpaco-aws:
Please check the most recent version, there are 4 tests now for this intrinsic.

PR #1100, opened 2022-04-23
Remove implementation for the `unreachable` intrinsic

adpaco-aws:
### Description of changes: 

The implementation we had for the `unreachable` intrinsic is unreachable itself. The code being executed is actually the `TerminatorKind::Unreachable`, which generates an assert-false statement with the message "unreachable code".

Adding a test to ensure this keeps happening.

### Resolved issues:

Part of #727 

### Testing:

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
@@ -607,12 +607,6 @@ impl<'tcx> GotocCtx<'tcx> {
             }
             "unchecked_sub" => codegen_op_with_overflow_check!(sub_overflow),
             "unlikely" => self.codegen_expr_to_place(p, fargs.remove(0)),
-            "unreachable" => self.codegen_assert(
-                Expr::bool_false(),
-                PropertyClass::DefaultAssertion,
-                "unreachable",
-                loc,
-            ),
             "volatile_copy_memory" => unstable_codegen!(codegen_intrinsic_copy!(Memmove)),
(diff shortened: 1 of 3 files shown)


tests/expected/intrinsics/unreachable/expected
```diff
@@ -0,0 +1,2 @@
+FAILURE\
+"unreachable code"
```
danielsn, line 2: newline?

danielsn (approved):

tedinski (approved):
Nice catch

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -607,12 +607,6 @@ impl<'tcx> GotocCtx<'tcx> {
             }
             "unchecked_sub" => codegen_op_with_overflow_check!(sub_overflow),
             "unlikely" => self.codegen_expr_to_place(p, fargs.remove(0)),
-            "unreachable" => self.codegen_assert(
```
zhassan-aws, line 610: Since `unreachable` should be unreachable, I suggest making this more explicit:
```rust
"unreachable" => unreachable!("Expecting unreachable intrinsic to be handled using Terminator::Unreachable")
```
because just removing it gives the impression that it is not yet implemented (gets included in the `_` bucket at the end of match statement).

adpaco-aws: Ok, I will add this in a follow-up PR.

PR #1089, opened 2022-04-21
Fix issue with value_of / align_of traits

celinval:
### Description of changes: 

We were trying to figure out the values of those properties during runtime, which is not possible. Fix the code to look for the information in the vtable instead.

### Resolved issues:

Fixes #1074 

### Call-outs:

This is v2. The initial PR was trying to mitigate it but it turned out that the fix was straight forward.

### Testing:

* How is this change tested? New tests

* Is this a refactor change? No

### Checklist
- [ ] Each commit message has a non-empty body, explaining why the change was made
- [ ] Methods or procedures are documented
- [ ] Regression or unit tests are included, or existing tests cover the modified code
- [ ] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #1101, opened 2022-04-23
Basic tests for `size_of_val` and `min_alig_of_val`

adpaco-aws:
### Description of changes: 

Adds two tests for `size_of_val` and `min_align_of_val`.

@celinval added tests for dynamically-sized types in #1089 , these ones just complement them by testing basic types.

### Resolved issues:

Part of #727 

### Testing:

* How is this change tested? Adds two tests.

### Checklist


tests/kani/Intrinsics/AlignOfVal/align_of_basic.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Check that we get the expected results for the `min_align_of_val` intrinsic
+// with common data types
```
danielsn, line 5: Is this architecture specific, of do these min aligns come from the rust spec?

tests/kani/Intrinsics/SizeOfVal/size_of_basic.rs
```diff
@@ -0,0 +1,44 @@
...
+use std::intrinsics::size_of_val;
+
+struct MyStruct {
+    val: u32,
+}
+
+enum MyEnum {
+    Variant,
+}
+
```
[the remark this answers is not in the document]
adpaco-aws: Added a `repr(C)` struct and assertions based on it. As far as I can tell, it's working as expected:
 * Size takes into account padding
 * Alignment is the max alignment for each field
But let me know if you want to test specific cases and I will add them.

tests/kani/Intrinsics/AlignOfVal/align_of_basic.rs
adpaco-aws: My understanding is that `min_align_of_val` depends on `min_global_align`, which is an architecture option.
Please note that in #1054 we added checks to ensure that `min_global_align` is unspecified or equal to 1, otherwise Kani will stop there.

danielsn: Can you add a comment that the tests here assume x86, otherwise good to go.

danielsn (approved):
Approved modulo one comment

PR #1104, opened 2022-04-25
Audit for `transmute`

adpaco-aws:
### Description of changes: 

Updates the `transmute` intrinsic with alignment checks for both the argument type and the result type. However, this does not seem to be working as I expected, and a large number of failures are being triggered on our regression.

In fact, the test added in this PR is failing with "transmute: result type is properly aligned" instead of "transmute: argument type is properly aligned". I don't know if I'm missing something.

### Resolved issues:

Part of #727 

### Testing:

### Checklist


celinval (commented):
Can you please add a bit more test coverage to transmute? Things like transmute pointers,  structures, slices? Thanks

celinval (approved):
Nice! Thanks Adrian

adpaco-aws:
I have restored the original implementation and added 4 tests to check correct transmutations between
 * an array into a `u32`
 * an array into a struct
 * a `&str` into a `&[u8]`
 * a pointer into a function pointer.
This is in addition to another one that is present in the `expected` test suite (transmute between float and integer I think).

Issue #316, opened 2021-07-08
Global ASM is not supported


danielsn:
Following LLVM, Rust allows global ASM in a module. This means that crates can have arbitrary code that we cannot analyze.

## Likelihood
We know of crates that use this feature

## Mitigation
We could fail any crate that uses this feature, but this would block any use of those crates, even if they did not use any of the risky features. Even worse, this would block all consumers of such crates, from even compiling, again even if they do not use any of the dangerous features.

Unlike for inline ASM in a function, where we can insert `assert(0)` as a mechanism to guarantee soundness, there is no obvious way to only fail code paths that use the injected ASM.

- [x] Raise a warning if global asm appears
- [ ] Incorporate that warning into viewer

## Path to soundness

- [ ] Support ASM in CBMC

## Documentation
https://rust-lang.github.io/rfcs/1548-global-asm.html

adpaco-aws:
The "Raise a warning" item can be mark as done, right?

zhassan-aws:
Capturing the current behavior as of c3d41b94896: Kani produces a codegen warning, but does not cause verification to fail. For example, on the following test (adapted from the one [here](https://rust-docs.redwoodjs.cn/unstable-book/library-features/global-asm.html)):
```rust
pub mod sally {
    std::arch::global_asm!(
        ".global foo",
        "foo:",
        "jmp baz",
    );

    #[no_mangle]
    pub unsafe extern "C" fn baz() {}
}

// the symbols `foo` and `bar` are global, no matter where
// `global_asm!` was used.
extern "C" {
    fn foo();
    fn bar();
}

pub mod harry {
    std::arch::global_asm!(
        ".global bar",
        "bar:",
        "jmp quux",
    );

    #[no_mangle]
    pub unsafe extern "C" fn quux() {}
}

#[kani::proof]
fn main() {
    assert!(1 + 1 == 2);
}
```
Kani produces two codegen warnings:
```
 WARN kani_compiler::codegen_cprover_gotoc::compiler_interface Crate global_asm contains global ASM, which is not handled by Kani
 WARN kani_compiler::codegen_cprover_gotoc::compiler_interface Crate global_asm contains global ASM, which is not handled by Kani
warning: 2 warnings emitted
```
but verification passes:
```
RESULTS:
Check 1: main.assertion.1
         - Status: SUCCESS
         - Description: "assertion failed: 1 + 1 == 2"
         - Location: global_asm.rs:32:5 in function main


SUMMARY: 
[...]

PR #1112, opened 2022-04-26
Error out for crates with global ASM

zhassan-aws:
### Description of changes: 

Currently, for crates that contain [module-level assembly](https://doc.rust-lang.org/beta/core/arch/macro.global_asm.html), codegen emits a warning, and skips the codegen of all `GlobalAsm` items. This may impact soundness as the assembly that we don't model could be affecting the state in that crate. To maintain soundness, Kani will now error out if any crate contains global ASM with the following message:
```
error: Crate crate_with_global_asm contains global ASM, which is not supported by Kani. Rerun with `--enable-unstable --ignore-global-asm` to suppress this error (**Verification results may be impacted**).
```
To allow the user to override this behavior (in case they're confident the global ASM does not impact results), an option, `--ignore-global-asm` (requires `--enable-unstable`), is introduced that allows compilation/verification to continue. A warning will be printed when this option is used in the presence of global ASM:
```
Ignoring global ASM in crate crate_with_global_asm. Verification results may be impacted.
```

### Resolved issues:

Mitigates soundness risk in #316 

### Testing:

* How is this change tested? Two tests added

### Checklist

### DIFF
--- a/kani-driver/src/args.rs
+++ b/kani-driver/src/args.rs
@@ -146,7 +146,7 @@ pub struct KaniArgs {
 
     /// Do not error out for crates containing `global_asm!`.
     /// This option may impact the soundness of the analysis and may cause false proofs and/or counterexamples
-    #[structopt(long)]
+    #[structopt(long, hidden_short_help(true), requires("enable-unstable"))]
     pub ignore_global_asm: bool,
     /*
     The below is a "TODO list" of things not yet implemented from the kani_flags.py script.
--- a/kani-driver/src/args.rs
+++ b/kani-driver/src/args.rs
@@ -143,6 +143,11 @@ pub struct KaniArgs {
     /// Turn off assertion reachability checks
     #[structopt(long)]
     pub no_assertion_reach_checks: bool,
+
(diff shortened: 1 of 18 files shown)


kani-driver/src/args.rs
```diff
@@ -143,6 +143,11 @@ pub struct KaniArgs {
     /// Turn off assertion reachability checks
     #[structopt(long)]
     pub no_assertion_reach_checks: bool,
+
+    /// Do not error out for crates containing `global_asm!`.
+    /// This option may impact the soundness of the analysis and may cause false proofs and/or counterexamples
+    #[structopt(long)]
+    pub ignore_global_asm: bool,
```
celinval, line 150: Should we mark this as unstable?

celinval (approved):
I would prefer if we mark this option as unstable since this can affect the proof results in ways that we are not quite sure about. Everything else looks good to me.

Thanks!

kani-driver/src/args.rs
zhassan-aws: Done.

PR #1117, opened 2022-04-27
Rewrite loops with multiple back edges to having a single back edge only

### LINKED ISSUE #1046: Low performance verification caused by conditional jump
While writing a test case for the `ctpop` intrinsic, I wrote something similar to this code:

```rust
fn my_ctpop_u32(x: u32) -> u32 {
    let mut count = 0;
    let num_bits = u32::BITS;
    for i in 0..num_bits {
        // Get value at index `i`
        let bitmask = 1 << i;
        let bit = x & bitmask;
        if bit != 0 {
            count += 1;
        }
        // assert!(count >= 0);
    }
    count
}

#[kani::proof]
fn main() {
    let var: u32 = kani::any();
    assert!(my_ctpop_u32(var) == ctpop(var));
}
```

Running `kani` on this example (without arguments) does not terminate (at least for a few minutes). We can pass `--unwind` up to 8 or so in order to get a result, but it will fail the unwinding assertion unless we pass 33 (which gets us back to non-termination).

What is surprising about this example is that adding back the assertion that is commented out will allow CBMC to successfully verify the code in a few seconds. Apparently, this is because the conditional jump in the goto program following from the `if bit != 0 { ... }` complicates the unwinding and verification of this example.

@tautschnig proposed using the `--ensure-one-backedge-per-target` option from `goto-instrument` in order to fix this, but it is not clear if this transformation preserves soundness in some cases (see the regression failure in #900).

tedinski:
This is a re-creation of #900 after the Kani history rewrite. I can't force-push to that branch to do the rebase, so I'm just recreating it.

### Description of changes: 

rustc reasonably translates if statements at end of a loop body to
backwards jumps to the loop head. This, however, causes CBMC's symbolic
execution to spuriously treat these as nested loops. This commit now
enables a rewrite step that had previously been built for exactly such
cases.

### Resolved issues:

Resolves #1046 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/docs/src/tutorial/arbitrary-variables/Cargo.toml
+++ b/docs/src/tutorial/arbitrary-variables/Cargo.toml
@@ -11,4 +11,4 @@ vector-map = "1.0.1"
 [workspace]
 
 [workspace.metadata.kani]
-flags = { default-checks = false, unwind = "2" }
+flags = { unwind = "3" }
--- a/kani-driver/src/call_goto_instrument.rs
+++ b/kani-driver/src/call_goto_instrument.rs
@@ -23,6 +23,8 @@ impl KaniSession {
             self.just_drop_unused_functions(output)?;
         }
 
+        self.rewrite_back_edges(output)?;
+
         if self.args.gen_c {
             if !self.args.quiet {
                 println!("Generated C code written to {}", output.to_string_lossy());
@@ -93,6 +95,16 @@ impl KaniSession {
     fn just_drop_unused_functions(&self, file: &Path) -> Result<()> {
         let args: Vec<OsString> = vec![
             "--drop-unused-functions".into(),
+            file.to_owned().into_os_string(), // input
+            file.to_owned().into_os_string(), // output
+        ];
+
+        self.call_goto_instrument(args)
+    }
+
+    fn rewrite_back_edges(&self, file: &Path) -> Result<()> {
+        let args: Vec<OsString> = vec![
+            "--ensure-one-backedge-per-target".into(),
             file.to_owned().into_os_string(), // input
             file.to_owned().into_os_string(), // output
         ];
--- a/tests/kani/AssociatedTypes/into_iter.rs
+++ b/tests/kani/AssociatedTypes/into_iter.rs
@@ -19,7 +19,7 @@ impl<'a> MyStruct<'a> {
 }
 
 #[kani::proof]
-#[kani::unwind(2)]
+#[kani::unwind(3)]
 pub fn check_into_iter_type() {
     let original = "h";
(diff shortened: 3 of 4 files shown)


danielsn (approved):
Discussed with @tautschnig. This rewrite has no known soundness issues. The only relevent effect is that it affects how CBMC counts loop unwindings, which is what was causing the test failure.

celinval:
> Discussed with @tautschnig. This rewrite has no known soundness issues. The only relevent effect is that it affects how CBMC counts loop unwindings, which is what was causing the test failure.

Interesting, do we have any idea when that's the case? It doesn't seem to affect all loop unwinding.

BTW, I don't see this option documented anywhere in their public documents or `goto-instrument --help`. Do we know why that's not customer visible? @tautschnig

Is this transformation pass similar to [LoopSimplify in LLVM](https://llvm.org/doxygen/LoopSimplify_8cpp_source.html)?

danielsn:
I believe its cases where the final statement of the loop body is an `if` with no matching `else`: CBMC was previously misidentifying them as nested loops, instead of as a single loop with multiple back-edges

tautschnig:
> > Discussed with @tautschnig. This rewrite has no known soundness issues. The only relevent effect is that it affects how CBMC counts loop unwindings, which is what was causing the test failure.
> 
> Interesting, do we have any idea when that's the case? It doesn't seem to affect all loop unwinding.
> 

It was really only a problem when there were multiple back-edges in a loop. This is entirely reasonable for a compiler to generate, but it catches CBMC's symbolic execution off guard for CBMC's C front-end (and goto conversion) will never generate this. (The choice is for what to do with `continue` instructions: a compiler could either choose to generate a jump to the backwards goto at the end of a loop, or directly jump back to the loop head. CBMC's goto conversion does the former, but apparently `rustc` at least sometimes does the latter. And presumably `javac` also does the latter at times.)

> BTW, I don't see this option documented anywhere in their public documents or `goto-instrument --help`. Do we know why that's not customer visible? @tautschnig
> 

I'm pretty sure it just was forgotten about when it was built. Presumably it was built to deal with code originating from `javac`. We should fix the documentation problem, but there already is an item on my todo list to include automated checks in CI that all options are documented (unless intentionally left undocumented). So I'll get to this eventually...

> Is this transformation pass similar to [LoopSimplify in LLVM](https://llvm.org/doxygen/LoopSimplify_8cpp_source.html)?

It's not quite as elaborate, it does only one part of it, the "This pass also guarantees that loops will have exactly one backedge."

adpaco-aws:
Added #1046 to "Resolved issues".

PR #1119, opened 2022-04-27
Declare `unreachable` intrinsic as unreachable

### LINKED ISSUE #1115: Case for `unreachable` intrinsic is unreachable
In #1100 we removed the `unreachable` intrinsic, but it would be better to add it back with an unreachable statement to make clear that it is handled by the `TerminatorKind::Unreachable`.

adpaco-aws:
### Description of changes: 

Includes the case for `unreachable` in the intrinsics match that is expected to be unreachable. This case was removed in #1100 (includes a test for `unreachable`), but as suggested by @zhassan-aws  this way we avoid confusion.

### Resolved issues:

Resolves #1115 

### Testing:

### Checklist


zhassan-aws (approved):
Thanks!

PR #1120, opened 2022-04-27
Replace vector-map by linear-map dependency

celinval:
### Description of changes: 

The `vector-map` crate indirectly depends on a vulnerable version of rand_core. This crate hasn't been released for the past 2 years, and there hasn't been any activity in their github either. I'm replacing it by `linear-map` for now to fix the security warnings.

### Resolved issues:

This should fix the open dependabot alert.

### Call-outs:

Note that linear-map last release was 6 years ago, so we may have to
remove it eventually. At least this crate only depends on serde.

### Testing:

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,7 +18,7 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
- "getrandom 0.2.6",
+ "getrandom",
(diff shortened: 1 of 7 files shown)


tedinski (approved):

zhassan-aws (commented):
We're also using `VecMap` in `docs/src/tutorial/arbitrary-variables`. Should we replace that one as well?

celinval:
> We're also using `VecMap` in `docs/src/tutorial/arbitrary-variables`. Should we replace that one as well?

I hope we don't have to. I was wondering if that would be the case but it would make sense for us to be able to add tests with crates that have known vulnerabilities to them.

PR #1129, opened 2022-04-28
Add minimal release automation to CI

tedinski:
### Description of changes: 

This should, on the creation of a tag named `kani-*`:

1. Validate the version of the crate and tag match
2. Create a Github release. Marked as "draft" to be manually published later.
3. Build linux/macos release bundles (with proper versions) and attach them to the draft release.

It does basically no testing/validation but that can come later. It will likely require some shuffling in how we do this because it's weird that you create the release THEN upload artifacts...

### Call-outs:

This is basically **completely untested** since it involves creating releases and such. Helping me find typos etc much appreciated, but we basically have to merge, create a tag, and see what happens as a result.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+name: Release
+on:
+  push:
+    tags:
+      - kani-*
+
+jobs:
+  Release:
+    name: Release
+    runs-on: ubuntu-20.04
```
adpaco-aws, line 12: You're bundling in `ubuntu-20.04` and uploading for both `ubuntu-20.04` and `macos-10.15`. Shouldn't you be bundling in `macos-10.15` as well or am I missing something?

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Get version
+        run: |
+          # pkgid is something like file:///home/ubuntu/rmc#kani-verifier:0.1.0
```
adpaco-aws, line 19: Should this be `file:///home/ubuntu/kani#kani-verifier:0.1.0` or is there some GitHub Action limitation that forces the original name?

.github/workflows/release.yml
tedinski: There's two jobs here: Release and Upload. Release does nothing but say "hey github, create a release"

The bundling and upload of a tarball happens in the other job.

.github/workflows/release.yml
tedinski: Just giving an example of what the output of the command is, in CI it'll be different yeah.

PR #1102, opened 2022-04-23
Audit for `write_bytes`

adpaco-aws:
### Description of changes: 

Moves and updates the `write_bytes` intrinsic implementation with an alignment check that it was missing. Adds 3 tests cases (2 of them in `expected`).

### Resolved issues:

Part of #727 

### Testing:

* How is this change tested? Adds 3 tests.

* Is this a refactor change? No.

### Checklist
- [x] Each commit message has a non-empty body, explaining why the change was made
- [x] Methods or procedures are documented
- [x] Regression or unit tests are included, or existing tests cover the modified code
- [x] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #1131, opened 2022-04-28
Refactor overflow checks in intrinsics code

### LINKED ISSUE #1116: Overflows checks in intrinsics should be refactored
Overflow checks (for offsets in bytes, for example) are now a common pattern for intrinsics. Refactoring them into helper functions would make the code simpler and possibly be helpful in other places.

adpaco-aws:
### Description of changes: 

Refactors the overflow checks added in #1094, #1099 and #1102 into a function that returns the result and the assertions associated with it. When working on this I realized the expected failure for `ptr_offset_from` (bytes) was actually a failure coming from `offset`. This was a rabbit hole because it's actually possible to trigger the failure by "wrapping around" a pointer, so I added a comment there to avoid confusion in the future.

### Resolved issues:

Resolves #1116 

### Testing:

* How is this change tested? Updates one test.

* Is this a refactor change? No (but it should have been).

### Checklist


tests/expected/offset-from-bytes-overflow/main.rs
```diff
@@ -3,15 +3,25 @@
...
+    let v: &[u128] = &[0; 10];
+    let v_0: *const u128 = &v[0];
+    // `high_offset` is an offset that is high enough to:
+    //  * Not trigger failures in `v_0.add(...)`
+    //  * Trigger failures in `offset_from` after it
+    let high_offset = usize::MAX / (std::mem::size_of::<u128>() * 4);
     unsafe {
-        let v_wrap: *const u128 = v_100.add((max_offset + 1).try_into().unwrap());
-        let _ = v_wrap.offset_from(v_100) == 2;
+        let v_wrap: *const u128 = v_0.add(high_offset.try_into().unwrap());
```
adpaco-aws, line 24: Add a comment here that explains this is UB, but Kani's default behavior doesn't catch it. We had to disable these checks because `std` does this kind of operation often.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -921,17 +915,10 @@ impl<'tcx> GotocCtx<'tcx> {
...
-        let size = Expr::int_constant(layout.size.bytes(), Type::ssize_t());
-        let offset_bytes = offset.clone().mul_overflow(size);
-        let overflow_check = self.codegen_assert(
-            offset_bytes.overflowed.not(),
-            PropertyClass::ArithmeticOverflow,
-            "attempt to compute offset in bytes which would overflow",
-            loc,
-        );
+        let (_offset_bytes, overflow_check) =
+            self.count_in_bytes(offset.clone(), ty, Type::ssize_t(), "ptr_offset_from", loc);
```
zhassan-aws, line 921: If I understood the documentation correctly (https://doc.rust-lang.org/std/primitive.pointer.html#method.offset_from), I think we need to divide (and not multiply) the offset between the two pointers (`offset`) by the size of `ty`.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1227,18 +1213,41 @@ impl<'tcx> GotocCtx<'tcx> {
...
-        // Check that computing `bytes` would not overflow
+        // Check that computing `count` in bytes would not overflow
+        let (count_bytes, overflow_check) =
+            self.count_in_bytes(count, ty, Type::size_t(), "write_bytes", loc);
+
+        let memset_call = BuiltinFn::Memset.call(vec![dst, val, count_bytes], loc);
+        Stmt::block(vec![align_check, overflow_check, memset_call.as_stmt(loc)], loc)
+    }
+
+    /// Computes (multiplies) the equivalent of a memory-related number (e.g., an offset) in bytes.
```
zhassan-aws, line 1224: Nice documentation!

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: I think there is a typo there and the documentation should say:
> The returned value is in units of T: the distance in bytes divided by `mem::size_of::<T>()`

The offset between two pointers is usually given in units of `T` (compare with the documentation for [`offset`](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset)), which you can convert to bytes by multiplying
`offset * mem::size_of<T>` (what we do here).

I've opened [this PR](https://github.com/rust-lang/rust/pull/96639) to correct the typo.

zhassan-aws: Isn't the offset computes using:
```
  let offset = cast_dst_ptr.sub(cast_src_ptr);
```
in terms of bytes? If so, then to convert it to be in terms of units of `T`, we need to divide by size of `T` since "offset in terms of bytes = offset in terms of T x size of T".

adpaco-aws: No, offset computation is done in terms of T. See the test case in `tests/kani/PointerOffset/offset_from.rs`:

```rust
    let a = [0; 5];
    let b = [0; 5];
    let ptr1: *const i32 = &a[1];
    let ptr2: *const i32 = &a[3];
    let ptr3: *const i64 = &b[1];
    let ptr4: *const i64 = &b[3];
    unsafe {
        assert!(ptr2.offset_from(ptr1) == 2);
        assert!(ptr1.offset_from(ptr2) == -2);
        assert!(ptr4.offset_from(ptr3) == 2);
        assert!(ptr3.offset_from(ptr4) == -2);
    }
```
Here, assertion 1 and 3 (or 2 and 4) return the same offset value. An offset computed in terms of bytes would return 8 in the first one (2 * 4 bytes for each `i32` value) and 16 on assertion 3 (2 * 8 bytes for each `i64` value).

zhassan-aws: One thing I was missing is that the return value is computed separately on line 937/924, and is not computed from `count_in_bytes`. I'm not sure I understand why an overflow check is needed here though. I'm assuming that a pointer subtraction in CBMC would involve a division operation and not a multiplication, i.e. `x-y` translates to `(x-y)/sizeof(*x)`?

adpaco-aws: The overflow check is done because it's one of the safety conditions listed in the documentation:

> The distance between the pointers, in bytes, cannot overflow an isize.

I don't know for sure how CBMC handles pointer arithmetic internally.

zhassan-aws: I see. But is this what the code is doing? I guess here's my source of confusion: what does `offset` represent in the lines below?
```
        let dst_ptr = fargs.remove(0);
        let src_ptr = fargs.remove(0);
        // Compute the offset with standard substraction using `isize`
        let cast_dst_ptr = dst_ptr.clone().cast_to(Type::ssize_t());
        let cast_src_ptr = src_ptr.clone().cast_to(Type::ssize_t());
        let offset = cast_dst_ptr.sub(cast_src_ptr);
```

tests/expected/offset-from-bytes-overflow/main.rs
```diff
@@ -3,15 +3,34 @@
 
 // Check that an offset computed with `offset_from` triggers a verification failure
 // if it overflows an `isize` in bytes.
+
+// This example can be really confusing, but is the only way to reproduce the
+// failure we are looking for. This program works fine in Rust and it's even
+// possible to assert that the value coming from `offset_from` is equal to
+// `high_offset`. But CBMC's memory model is going to cause a "wrapping around"
+// behavior in `v_wrap`, so any values that depend on it are going to show a
+// strange behavior as well.
```
danielsn, line 12: tracking ticket

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1227,18 +1213,41 @@ impl<'tcx> GotocCtx<'tcx> {
             loc,
         );
 
-        // Check that computing `bytes` would not overflow
+        // Check that computing `count` in bytes would not overflow
+        let (count_bytes, overflow_check) =
```
danielsn, line 1217: This catches the `Rust` definition of UB, but will still cause problems in CBMC if we overflow the index bits, correct?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: @zhassan-aws pointed out in a discussion that, because of these conversions, the computed offset would be in bytes and not units of T. So the overflow check should be based on this `sub` operation, and using `count_in_bytes` does not proceed here.

Thanks @zhassan-aws !

adpaco-aws:
This PR is ready for review again:
 * Reviewed the overflow check in `ptr_offset_from` (see discussion above)
 * Changed the `offset-from-bytes-overflow` expected output to look for overflows in both offset operations
 * Added `offset-wraps-around` to test the "wrapping-around" behavior in CBMC.
 * Created #1150 and linked in offset-related overflow checks.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: Yes, and there is now a separate test for CBMC's wrap-around behavior.

tests/expected/offset-from-bytes-overflow/main.rs
adpaco-aws: Created and added: #1150

zhassan-aws (approved):
Thanks!

Issue #416, opened 2021-08-11
Support Rust's `Generator` feature


avanhatt:
Rust has a Generator type: [](https://doc.rust-lang.org/std/ops/trait.Generator.html)

We currently fail with unimplemented for this type:
```rust
            ty::Generator(_, _, _) => unimplemented!(),
```

This is used by the popular `toykio` crate: [](https://github.com/tokio-rs/tokio)

zhassan-aws:
Kani crashes on the following program:
```rust
#![feature(generators, generator_trait)]

use std::ops::{Generator, GeneratorState};
use std::pin::Pin;

fn main() {
    let mut generator = || {
        yield 1;
        return "foo"
    };

    match Pin::new(&mut generator).resume(()) {
        GeneratorState::Yielded(1) => {}
        _ => panic!("unexpected return from resume"),
    }
    match Pin::new(&mut generator).resume(()) {
        GeneratorState::Complete("foo") => {}
        _ => panic!("unexpected return from resume"),
    }
}
```
We should emit an error and exit gracefully until we support this feature.

zhassan-aws:
Emit an error instead of crash, then remove the "must have" label, and switch "bug" to "feature".

PR #1132, opened 2022-04-28
Error out if crate contains a generator call

zhassan-aws:
### Description of changes: 

Kani crashes if a crate contains a call to a generator, e.g. via a `Pin` (see https://github.com/model-checking/kani/issues/416#issuecomment-1069416183). With this PR, Kani will error out with a message instead, e.g.
```
error: Crate `main` uses generators which is currently not supported by Kani
```

### Resolved issues:

Part of #416 

### Call-outs:

Ideally, we should codegen this as unimplemented so that codegen would pass and verification would only fail if the generator call is reachable. However, doing so requires determining the function signature of a generator, which is something that I haven't figured out how to do yet, so in the time being, we'll just error out (instead of crashing).

### Testing:

* How is this change tested? One test added

### Checklist


kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -211,7 +212,14 @@ impl<'tcx> GotocCtx<'tcx> {
...
                 Some(sig)
             }
-            ty::Generator(_def_id, _substs, _movability) => None,
+            ty::Generator(_def_id, _substs, _movability) => {
+                let error_msg = format!(
+                    "Crate `{}` uses generators which is currently not supported by Kani",
+                    self.short_crate_name()
+                );
+                self.tcx.sess.err(&error_msg);
+                FatalError.raise()
```
adpaco-aws, line 221: Haven't we written a standard function for these errors? `unsupported_msg` seems to be building a message similar to this one, but I don't see any function for erroring out.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -891,11 +899,12 @@ impl<'tcx> GotocCtx<'tcx> {
...
             ty::FnPtr(_) => self.codegen_ty(pointee_type).to_pointer(),
 
+            ty::Generator(_, _, _) => self.codegen_ty(pointee_type).to_pointer(),
+
             // These types have no regression tests for them.
             // For soundness, hold off on generating them till we have test-cases.
             ty::Bound(_, _) => todo!("{:?} {:?}", pointee_type, pointee_type.kind()),
             ty::Error(_) => todo!("{:?} {:?}", pointee_type, pointee_type.kind()),
-            ty::Generator(_, _, _) => todo!("{:?} {:?}", pointee_type, pointee_type.kind()),
             ty::GeneratorWitness(_) => todo!("{:?} {:?}", pointee_type, pointee_type.kind()),
```
adpaco-aws, line 908: Can we do the same for `GeneratorWitness`?

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
zhassan-aws: Good point. Done.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
zhassan-aws: I would rather wait till we see a testcase first. The documentation for `GeneratorWitness` https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/relate/struct.GeneratorWitness.html doesn't have an explanation of when it would arise.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -211,7 +212,14 @@ impl<'tcx> GotocCtx<'tcx> {
...
                 Some(sig)
             }
-            ty::Generator(_def_id, _substs, _movability) => None,
+            ty::Generator(_def_id, _substs, _movability) => {
+                let error_msg = GotocCtx::unsupported_msg(
+                    "The `generators` feature",
+                    Some("https://github.com/model-checking/kani/issues/416"),
+                );
+                self.tcx.sess.err(&error_msg);
+                FatalError.raise()
```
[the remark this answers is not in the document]
zhassan-aws: We will probably need it. I'll add a function for it.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -891,11 +899,12 @@ impl<'tcx> GotocCtx<'tcx> {
             // https://github.com/model-checking/kani/issues/214
             ty::FnPtr(_) => self.codegen_ty(pointee_type).to_pointer(),
 
+            ty::Generator(_, _, _) => self.codegen_ty(pointee_type).to_pointer(),
```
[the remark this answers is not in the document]
zhassan-aws: No, I'm not sure. But the point is that if there is a reference to a generator, then when we codegen the generator itself, we will either codegen unimplemented or error out. I'll add a comment that mentions this.

adpaco-aws (approved):

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -899,6 +897,13 @@ impl<'tcx> GotocCtx<'tcx> {
             // https://github.com/model-checking/kani/issues/214
             ty::FnPtr(_) => self.codegen_ty(pointee_type).to_pointer(),
 
+            // Use the default for a reference to a generator. This is a
+            // temporary workaround to allow codegen to continue to a point
+            // where it either:
+            // 1. codegens unimplemented for the generator OR
+            // 2. erros out
```
danielsn, line 904: errors

kani-compiler/src/codegen_cprover_gotoc/utils/utils.rs
```diff
@@ -94,6 +95,11 @@ impl<'tcx> GotocCtx<'tcx> {
         }
         s
     }
+
+    pub fn emit_error_and_exit(&self, error_msg: &str) -> ! {
+        self.tcx.sess.err(&error_msg);
```
danielsn, line 100: Do we need the reborrow?

danielsn (approved):
Two nit⛏️

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
zhassan-aws: Fixed.

PR #1143, opened 2022-05-02
Disable SIMD intrinsics

adpaco-aws:
### Description of changes: 

Disables all SIMD intrinsics, which were not considered during the general intrinsics audit. The main concern is with these intrinsics is that overflow checks do not seem to be introduced for some of them.

### Testing:

* How is this change tested? Disables 4 tests.

### Checklist


nchong-at-aws:
Looks reasonable to me, Adrian. Do we have a ticket to track re-enabling these intrinsics at some point?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -352,8 +352,13 @@ impl<'tcx> GotocCtx<'tcx> {
...
 
         if let Some(stripped) = intrinsic.strip_prefix("simd_shuffle") {
-            let n: u64 = stripped.parse().unwrap();
-            return self.codegen_intrinsic_simd_shuffle(fargs, p, cbmc_ret_ty, n);
+            let _n: u64 = stripped.parse().unwrap();
+            return unstable_codegen!(self.codegen_intrinsic_simd_shuffle(
+                fargs,
+                p,
+                cbmc_ret_ty,
+                n
```
danielsn, line 360: Should this be `_n`

danielsn (approved):

adpaco-aws:
> Looks reasonable to me, Adrian. Do we have a ticket to track re-enabling these intrinsics at some point?

We didn't, but made one: #1148 . Thanks!

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: I had to add `_` to all non-used variables, functions or macros because it would introduce a warning otherwise (the `unstable_codegen` macro causes them to not be used). I want to avoid "using" them because the changes would be untested. Of course, the compiler will let us know what variables have to be corrected when we restore them.

PR #1146, opened 2022-05-02
Filter sanity checks

### LINKED ISSUE #1020: Hide Sanity Checks from Output
Requested feature:

There are a lot of Sanity checks present in the final output that need to be hidden from the final presentation to the user. They add a lot of noise and this change would help the user find relevant information quicker.

Tagging this as a "must have" because with the Property Class changes, this should be relatively easy to add, maybe first behind an experimental flag.

jaisnan:
### Description of changes: 

Properties which are sanity checks needed to be removed from the UI. They are now marked with the property_class "sanity_check" which makes it easy to filter such checks out as the pipeline to add, remove and edit checks based on their property_class has already been setup.

### Resolved issues:

Resolves #1020 

### Call-outs:

Couldn't add tests for this UI change as it's a removal. In the future we should contain negative tests for similar changes.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


scripts/cbmc_json_parser.py
```diff
@@ -373,19 +374,27 @@ def has_check_failure(properties, message):
 def filter_reach_checks(properties):
     return filter_properties(properties, GlobalMessages.REACH_CHECK_DESC)
 
+def filter_sanity_checks(properties):
+    return filter_properties(properties, None, "sanity_check")
 
-def filter_properties(properties, message):
+def filter_properties(properties, message=None, property_class=None):
```
danielsn, line 380: Do we want a single `property_class` or a list?

jaisnan: Maybe a wrapper function over this function which takes in a list of property_classes and messages? For this function itself, I think it's better to keep it simple, and call it over a list somewhere else.

scripts/cbmc_json_parser.py
```diff
@@ -329,6 +329,7 @@ def postprocess_results(properties, extra_ptr_check):
     has_failed_unwinding_asserts = has_check_failure(properties, GlobalMessages.UNWINDING_ASSERT_DESC)
     has_reachable_undefined_functions = modify_undefined_function_checks(properties)
     properties, reach_checks = filter_reach_checks(properties)
+    properties = filter_sanity_checks(properties)[0]
```
celinval, line 332: Is there any chance this is empty?

zhassan-aws (changes requested):
We need to check if any of the sanity checks fail, and if so, report verification failure (and the sanity check that failed).

scripts/cbmc_json_parser.py
celinval: also, you should move this logic to be inside filter_sanity_checks

tests/kani/Invariants/fixme_niche_opt.rs
```diff
@@ -1,38 +0,0 @@
-// Copyright Kani Contributors
-// SPDX-License-Identifier: Apache-2.0 OR MIT
-
-/// This testcase is currently failing with the following error:
-///
-/// [assertion.2] Reached assignment statement with unequal types Pointer { typ:
```
celinval, line 6: Good catch. As we discussed offline, sanity checks that fail should be visible and fail the verification. Thanks!

jaisnan: let me address those cases and clean up some of the code as well

scripts/cbmc_json_parser.py
jaisnan: sure, thanks!

scripts/cbmc_json_parser.py
```diff
@@ -369,23 +370,41 @@ def has_check_failure(properties, message):
...
+    for property in properties:
+        if property_class in extract_property_class(property) and property["status"] == "SUCCESS":
+            pass
+        else:
+            filtered_properties.append(property)
+
+    return filtered_properties
 
-def filter_properties(properties, message):
+def filter_properties(properties, message=None, property_class=None):
```
zhassan-aws, line 390: The changes to this function are no longer needed, right?

jaisnan:
I've simplified this PR to just do the minimum and added a simple test to check for sanity_check. Will try to address comments on the spot to get it merged before tomorrow.

scripts/cbmc_json_parser.py
jaisnan: Yes, redundant changes.

zhassan-aws (approved):
LGTM. In the future, we might want to make a sanity check failure more pronounced, e.g. through erroring out and not displaying any results at all. But we can do that in another PR.

PR #1147, opened 2022-05-03
Documentation: Change `first-steps` examples structure

adpaco-aws:
### Description of changes: 

Splits the `kani-first-steps` example into two: `first-steps-v1` and `first-steps-v2`. `kani-first-steps` source included two files: `lib.rs` (the original version) and `final_form.rs` (the final version), which `lib.rs` imported. To avoid confusion, the source code is a simple file `main.rs` on each project.

After the recent changes in `cargo kani` some of the tutorial information was outdated, which this PR updates as well.

### Testing:

* How is this change tested? Docs: Local server. Code: Regression.

### Checklist


docs/src/tutorial-first-steps.md
```diff
@@ -4,10 +4,10 @@ Kani is unlike the testing tools you may already be familiar with.
 Much of testing is concerned with thinking of new corner cases that need to be covered.
 With Kani, all the corner cases are covered from the start, and the new concern is narrowing down the scope to something manageable for the checker.
 
-Consider this first program (which can be found under [`kani-first-steps`](https://github.com/model-checking/kani/tree/main/docs/src/tutorial/kani-first-steps/)):
+Consider this first program (which can be found under [`first-steps-original`](https://github.com/model-checking/kani/tree/main/docs/src/tutorial/first-steps-original/)):
```
jaisnan, line 7: Is there any other naming convention that we can use over original vs final? Maybe before vs after or initial vs final?

adpaco-aws: Yes, we could use "initial" and "final" or version numbers if you like (e.g., `first-steps-v1` and `first-steps-v2`). I don't like "before" and "after" because they remind me of TV commercials 😆

jaisnan: Sure let's not do before/after 😆 

I personally like v1 and v2, if you agree with it as well, i think it makes sense to rename it to that.

adpaco-aws: Updated to use `first-steps-v1` and `first-steps-v2`!

jaisnan (approved):
Thanks!

PR #1153, opened 2022-05-03
Remove redundant check from README.md

celinval:
### Description of changes: 

In the example, we were calling is_valid after `kani::any()` which is redundant since the type has to implement the Arbitrary type.

### Resolved issues:

N/A

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,15 +25,12 @@ See [the installation guide](https://model-checking.github.io/kani/install-guide
 Similar to testing, you write a harness, but with Kani you can check all possible values using `kani::any()`:
 
 ```rust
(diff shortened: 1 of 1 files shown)


zhassan-aws:
I think the `is_valid` in the example proof harness is not meant to be the same as the one provided by the `Invariant` trait. Perhaps we can rename it to avoid confusion, e.g. `satisfies_some_conditions`? IMO including `kani::assume` in the example is useful.

celinval:
What about `kani::assume(pre_condition(input))`?

zhassan-aws:
Sounds good

PR #1176, opened 2022-05-06
Remove PtrWrite hook

### LINKED ISSUE #1173: Remove `PtrWrite` hook

danielsn:
### Description of changes: 

The PtrWrite hook is no longer necessary given that we support the underlying intrinsics.

### Resolved issues:

Resolves #1173

### Call-outs:

Removing the `PtrRead` hook makes the regression timeout. Only removing PtrWrite for now. #1175

### Testing:

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
@@ -282,40 +282,6 @@ impl<'tcx> GotocHook<'tcx> for PtrRead {
     }
 }
 
-struct PtrWrite;
-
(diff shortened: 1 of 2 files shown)


tedinski (approved):
Was this just legacy from before we had intrinsic handling code?

adpaco-aws (changes requested):
I thought that removing this hook would still require you to add implementations for `write` and `write_unaligned` as intrinsics, but this does not seem to be the case. So the code for handling these "writes" must be somewhere else.

Can you please find out where the code handling these "writes" is? The regular version should have an alignment check as in `codegen_volatile_store`.

danielsn:
Blocked on https://github.com/diffblue/cbmc/pull/6856

PR #1183, opened 2022-05-10
Use a specific version of CBMC for both Ubuntu and macOS

### LINKED ISSUE #933: Ubuntu and macOS regressions do not use the same CBMC version
Our setup script for macOS does not control the CBMC version to be installed. This will cause confusing CI failures if there are bugs in the latest CBMC version, blocking contributions for a while (as in #928).

Instead, we should use a single, fixed CBMC version for both regressions.

zhassan-aws:
### Description of changes: 

Kani's CI currently installs a specific version of CBMC in Ubuntu, but installs CBMC in macos-10.15 using `brew install cbmc`, which gives us whichever version is available on the GitHub servers. The difference in versions can cause CI to suddenly start failing if a new CBMC version becomes available on the GitHub Mac servers, e.g.:
1. If a bug was introduced in the newer CBMC version, this may cause regressions to fail
2. If a bug was fixed in the newer CBMC version, it may cause a "fixme" test to pass, causing regressions to fail

This PR addresses this version discrepancy through:
1. Pining the CBMC version used in macos-10.15 using the newly available [CBMC tap](https://github.com/diffblue/cbmc/blob/develop/doc/ADR/homebrew_tap.md)
2. Refactoring the setup scripts to make sure the same version is installed in Ubuntu and Mac

### Resolved issues:

Resolves #933 

### Testing:

* How is this change tested? Regressions

### Checklist


celinval (commented):
What if instead of passing the cbmc version as an argument, we read it from a file? This file becomes the source of truth and we don't have to update the version in different places including the documentation.

tedinski:
I'm also reminded that adding parameters to these scripts means there's more than one location for versions: the github action and the documentation that describes how to run the setup scripts locally to prepare a development environment.

I like the file idea as a temporary measure.

zhassan-aws:
> What if instead of passing the cbmc version as an argument, we read it from a file? This file becomes the source of truth and we don't have to update the version in different places including the documentation.

> I'm also reminded that adding parameters to these scripts means there's more than one location for versions: the github action and the documentation that describes how to run the setup scripts locally to prepare a development environment.

Created a `cbmc-version` file that is used in CI and referenced in the docs.

zhassan-aws:
Turned this into a draft until https://github.com/diffblue/homebrew-cbmc/pull/2 is merged or https://github.com/diffblue/homebrew-cbmc/issues/3 is resolved.

scripts/setup/macos-10.15/install_cbmc.sh
```diff
@@ -4,5 +4,15 @@
...
 set -eux
 
-# Install CBMC for macOS 10.15
-brew install cbmc
+if [[ $# -ne 1 ]]; then
+  echo "$0: Error: Specify the CBMC version file"
+  exit 1
+fi
+
+CBMC_VERSION_FILE=$1
```
celinval, line 12: We can change this later, but here are my two cents. We should create one file: kani-dependencies. The format should be:
```
CBMC_VERSION="5.6"
CBMC_VIEWER_VERSION="3.2"
...
```
You can now use `source kani-dependencies` inside your script and those variables are available in your script. Note that you can also use the same file inside python scripts and it is also a valid toml format.

zhassan-aws: Done.

kani-dependencies
```diff
@@ -0,0 +1,2 @@
+CBMC_VERSION="5.61.0"
+CBMC_VIEWER_VERSION="3.5"
```
tedinski, line 2: Please keep a comment here pointing to `src/setup.rs`. (And I think there might be a comment there that pointed back to the action? At any rate, that should also have a comment pointing here)

scripts/setup/install_viewer.sh
```diff
@@ -6,11 +6,14 @@ set -eu
 
 # Install cbmc-viewer
 
-if [[ $# -ne 1 ]]; then
-  echo "$0: Error: Specify the version to install"
+# Source kani-dependencies to get the CBMC viewer version
```
tedinski, line 9: nit: I'd sate the variable name, i.e. "to get CBMC_VERSION"

scripts/setup/macos-10.15/install_cbmc.sh
```diff
@@ -4,5 +4,15 @@
 
 set -eux
 
-# Install CBMC for macOS 10.15
-brew install cbmc
+# Source kani-dependencies to get the CBMC version
```
tedinski, line 7: same nit

scripts/setup/macos-10.15/install_deps.sh
```diff
@@ -7,7 +7,7 @@ set -eux
 # Github promises weekly build image updates, so we can skip the update step and
 # worst case we should only be 1-2 weeks behind upstream brew.
 # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-software
-brew update
+#brew update
```
tedinski, line 10: Unfortunately, we might need to keep this :/

Even with the version pin, I think it has to be in the index?

scripts/setup/ubuntu/install_cbmc.sh
```diff
@@ -4,8 +4,15 @@
 
 set -eu
 
+# Source kani-dependencies to get the CBMC version
```
tedinski, line 7: same nit

tedinski (approved):
Small things to maybe fix, but lgtm

kani-dependencies
zhassan-aws: Done. I also updated the comment in `src/setup.rs` to point to `kani-dependencies` instead of the action.

scripts/setup/macos-10.15/install_deps.sh
zhassan-aws: Based on [the documentation](https://docs.brew.sh/Taps#the-brew-tap-command), `brew tap` clones the repository, so a `brew update` should not be needed (at least for cbmc, not sure about other packages).

scripts/setup/install_viewer.sh
zhassan-aws: Done.

scripts/setup/macos-10.15/install_cbmc.sh
zhassan-aws: Done

scripts/setup/ubuntu/install_cbmc.sh
zhassan-aws: Done.

tedinski (approved):

PR #1185, opened 2022-05-10
UI Update : Modify `expect_fail` behavior

jaisnan:
### Description of changes: 

1) Checks with `expect_fail` property class should behave differently to other checks. Semantically, in the UI, it shows the status for the assert inside the `expect_fail`, not the status of the `expect_fail` assert itself. This PR updates the status to show the status of the `expect_fail` check.

For example - for the harness, 

``` rust
#[kani::proof]
fn main() {
    let i: i32 = kani::any();
    kani::assume(i < 10);
    kani::expect_fail(i > 20, "Blocked by assumption above.");
}
```
The output previously was, 

`Check 1: main.expect_fail.1
         - Status: FAILED`

which is incorrect since it is expected that the boolean should be false which it is not. It is now updated to - 

`Check 1: main.expect_fail.1
         - Status: SUCCESS`

2) The final UI flow for `expect_fail` is -  

```
if : 
    sanity_check fails, expect_fail should show UNDETERMINED.
else:
    if expect_fail assert fails, the prop should show SUCCESS
    if expect_fail assert succeeds, and all unwind assertion / undef function check succeed, the prop should show FAIL
    if expect_fail assert succeeds, and an unwind assertion / undef function check fails, the prop should show UNDETERMINED

```

### Call-outs:

1) The UI logic is complicated and the testing does not fully capture all the scenarios possible. 
2) I shall try to organize the parser code in my next PR before rewriting in rust, the code is quite messy right now.
3) We should think about renaming `expect_fail` to `should_fail`, which I think @zhassan-aws suggested.

### Testing:

* How is this change tested? Added tests

### Checklist


scripts/cbmc_json_parser.py
```diff
@@ -370,6 +399,15 @@ def has_check_failure(properties, message):
             return True
     return False
 
+def has_check_failure_property_class(properties, property_class):
+    """
+    Search in properties for a failed property with the given message
```
jaisnan, line 404:
```suggestion
    Search in properties for a failed property with the given property class
```

scripts/cbmc_json_parser.py
```diff
@@ -281,15 +285,42 @@ def modify_undefined_function_checks(properties):
...
+def modify_expect_fail_checks(
+        properties,
+        has_failed_sanity_check,
+        has_reachable_undefined_functions,
+        has_failed_unwinding_asserts):
+    """
+    Invert status for Checks with the property_class "expect_fail" from FAILURE TO SUCCESS and vice versa.
+    Modify further depending on unwind/undefined function check statuses.
+    """
+    for property in properties:
```
zhassan-aws, line 307: It seems to me that if we call this function after the loop that updates `SUCCESS` to `UNDETERMINED` in `postprocess_results`, the logic can be simplified to "swap SUCCESS with FAILURE".

scripts/cbmc_json_parser.py
```diff
@@ -325,11 +356,18 @@ def postprocess_results(properties, extra_ptr_check):
     of the special cases above was hit.
     """
 
-    has_reachable_unsupported_constructs = has_check_failure(properties, GlobalMessages.UNSUPPORTED_CONSTRUCT_DESC)
-    has_failed_unwinding_asserts = has_check_failure(properties, GlobalMessages.UNWINDING_ASSERT_DESC)
+    has_reachable_unsupported_constructs = has_check_failure_message(
+        properties, GlobalMessages.UNSUPPORTED_CONSTRUCT_DESC)
+    has_failed_unwinding_asserts = has_check_failure_message(properties, GlobalMessages.UNWINDING_ASSERT_DESC)
     has_reachable_undefined_functions = modify_undefined_function_checks(properties)
+    has_failed_sanity_check = has_check_failure_property_class(properties, GlobalMessages.SANITY_CHECK)
```
zhassan-aws, line 363: I think sanity checks should be handled differently: if one fails, we should print the one that failed and not print any results at all, i.e. error out. This should be done in a separate PR.

zhassan-aws:
In addition to the comments, I think using "SUCCESS" and "FAILURE" for the `expect_fail` checks might be a bit confusing. Perhaps, we can create new categories for them, e.g. "EXPECTED FAILURE", and "UNEXPECTED SUCCESS"?

scripts/cbmc_json_parser.py
jaisnan: I can address that right away in another PR.

scripts/cbmc_json_parser.py
jaisnan: Will do.

scripts/cbmc_json_parser.py
```diff
@@ -281,15 +286,33 @@ def modify_undefined_function_checks(properties):
...
             if property["status"] == "FAILURE":
-                has_unknown_location_checks = True
-    return has_unknown_location_checks
+                has_undefined_function_checks = True
+    return has_undefined_function_checks
+
+def modify_expect_fail_checks(properties):
+    """
+    Invert status for Checks with the property_class "expect_fail" from FAILURE TO SUCCESS and vice versa.
+    Modify further depending on unwind/undefined function check statuses.
```
zhassan-aws, line 302: The second sentence no longer applies.

scripts/cbmc_json_parser.py
```diff
@@ -281,15 +286,33 @@ def modify_undefined_function_checks(properties):
...
+    """
+    expected_fail_string = "Failed as expected."
+    unexpected_success_string = "Unexpectedly passed."
+    for property in properties:
+        if extract_property_class(property) == GlobalMessages.EXPECT_FAIL:
+            if property["status"] == GlobalMessages.STATUS_SUCCESS:
+                property["status"] = GlobalMessages.STATUS_FAILURE
+                property["description"] += ', ' + unexpected_success_string
```
zhassan-aws, line 310: I don't think we should append it to the property description. This would make the description depend on the result which isn't the case for any other property. We should instead add it to the status.

scripts/cbmc_json_parser.py
zhassan-aws: Can we take out sanity checks from this PR? Sanity checks are different than unwinding asserts and undefined functions in that if one fails, none of the results (whether SUCCESS or FAILURE) can be trusted.

scripts/cbmc_json_parser.py
jaisnan: while we haven't done it anywhere else, would you say it's better to expand the status with an additional phrase as opposed to expanding the description? I assumed that the description would be where we could add on additional information wherever applicable.

jaisnan:
1. This PR

Issue #1178, opened 2022-05-08
`can't find crate for std` error when running with `rust_decimal`


drogus:
I've started playing with Kani today and it's an awesome tool!

I ran into a problem when trying to run it in a project relying on the `rust_decimal` crate. If I try to run `cargo kani` in a `first-steps-v1` repository without any changes, it works properly. After adding `rust_decimal = "*"` to the dependencies I get a bunch of errors starting with:

```rust
error[E0463]: can't find crate for `std`

error[E0463]: can't find crate for `std`
 --> /home/drogus/.cargo/registry/src/github.com-1ecc6299db9ec823/rust_decimal-1.23.1/build.rs:1:5
  |
1 | use std::{fs, path::PathBuf};
  |     ^^^ can't find crate
```

The rest of the errors are also about missing symbols.

kani version:

```
$ cargo kani --version
cargo-kani 0.1.0
```

drogus:
I have just tried to run with a local clone of `rust_decimal` and `cargo kani` works correctly after removing the `build.rs` file from the `rust_decimal` crate.

drogus:
One more note: it looks like it fails if there is any `build.rs` file in the tree. I just tried to also add `chrono` and it fails on the `build.rs` file from its `libc` dependency.

zhassan-aws:
Thanks @drogus for reporting this issue! It seems that our support for crates with a `build.rs` recently broke.

@celinval @tedinski any ideas? I found that the original issue, https://github.com/model-checking/kani/issues/613, is also not working.

zhassan-aws:
This seems to be an issue with the proxy binaries: the following two `cargo-kani` regressions fail when run with the proxy binaries:
```
failures:
    [cargo-kani] cargo-kani/cbmc-unknown-lang-mode/test.expected
    [cargo-kani] cargo-kani/dependencies/check_dummy.expected

test result: FAILED. 29 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 46.30s

Some tests failed in compiletest suite=cargo-kani mode=cargo-kani host=(none) target=(none)
```

tedinski:
Diagnosis: We're not setting `--sysroot` or `LD_LIBRARY_PATH` correctly. We set `--sysroot` in `KANIFLAGS` but when building a `build.rs` that environment variable gets ignored!

With `strace -v -s 10000 -f cargo kani 2> trace.txt` to diff between a working (built locally) and failing (from release):

[...]

PR #1187, opened 2022-05-10
Move logic to find rustup toolchain from -driver to -compiler.

tedinski:
### Description of changes: 

1. We do a bunch of logic to determine our rustup toolchain location in `kani-driver` only to imperfectly pass this information to `kani-compiler`. This causes problems when building `build.rs` files, as this is a case where our `--sysroot` path in `KANIFLAGS` gets ignored by `kani-compiler`. As a result, these fail in release builds as `kani-compiler` only had a compile-time path hardcoded.
2. Just move all this logic to `kani-compiler` instead! Now it can just reliably determine the rustup toolchain location correctly every time.
3. The added test didn't fail before *locally*, but I invoke it inside the docker container during the release-bundle tests. It did fail there, exhibiting the exact problems in #1178. This is because the container no longer had the hard-coded path available.

### Resolved issues:

Resolves most of #1178. I'd still like to take a look at the environment variables getting set by rustup there, but this is a cleanliness issue, not a problem people are encountering.

### Testing:

* How is this change tested? new test

* Is this a refactor change? bug fix

### Checklist


celinval (approved):
Nice to have an explicit build.rs test, but it would pass without your fix, wouldn't it?

We really need to be able to run the regression using the release bundle. Have you checked how much work would it be?

celinval:
> Nice to have an explicit build.rs test, but it would pass without your fix, wouldn't it?
> 
> We really need to be able to run the regression using the release bundle. Have you checked how much work would it be?

Oops. I  should've read the entire PR description. Thanks for adding it. It would still be valuable to run the full regression though.

PR #1192, opened 2022-05-11
Allow cargo-kani to run outside a Cargo project

### LINKED ISSUE #1154: Cargo-kani always assumes it is run inside a cargo project folder
I tried running `cargo kani --help` inside `/tmp` and it failed with the following error:

```
$ cargo-kani --help
Error: error: could not find `Cargo.toml` in `/tmp` or any parent directory


Stack backtrace:
   1: kani_driver::args_toml::join_args
   2: kani_driver::main
   3: std::sys_common::backtrace::__rust_begin_short_backtrace
   4: std::rt::lang_start::{{closure}}
   5: std::rt::lang_start_internal
   6: main
   7: __libc_start_main
   8: _start
```

Note: The same thing happens when running `cargo --version`

tedinski:
### Description of changes: 

Allow `cargo-kani` to run at all without being in a Cargo project.

### Resolved issues:

Resolves #1154

### Call-outs:

A normal run outside a project goes from:

```
$ cargo kani
Error: error: could not find `Cargo.toml` in `/home/ubuntu` or any parent directory
```

to

```
$ cargo kani
error: could not find `Cargo.toml` in `/home/ubuntu` or any parent directory
```

So still pretty good. But now we can run `cargo kani --version`

### Testing:

* How is this change tested? added test in docker

### Checklist

### DIFF
--- a/.github/workflows/kani.yml
+++ b/.github/workflows/kani.yml
@@ -103,6 +103,7 @@ jobs:
       - name: Run installed tests
         if: ${{ matrix.os == 'ubuntu-20.04' }}
         run: |
+          docker run kani-latest  cargo kani --version
           docker run -w /tmp/kani/tests/cargo-kani/simple-lib  kani-latest  cargo kani
           docker run -w /tmp/kani/tests/cargo-kani/simple-visualize  kani-latest  cargo kani
           docker run -w /tmp/kani/tests/cargo-kani/build-rs-works  kani-latest  cargo kani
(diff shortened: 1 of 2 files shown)


tedinski:
We don't really have an available mechanism for such a test.

Besides, I think the erroring is less important than the actual text of the error. It'd probably best be a UI test of some sort. But right now any test would just be like "ok, I'll build kani under kani! 🤪" (hence the docker for the test I did add)

zhassan-aws (approved):
I see.

celinval:
One thing we should do is to use `kani` or `cargo kani` instead of `cargo-kani` in our doc. I.e.: We shouldn't tell users to invoke `cargo-kani` directly.

PR #1196, opened 2022-05-13
Correct types for alignment expressions

adpaco-aws:
### Description of changes: 

Corrects the types used to build alignment expressions.

Previously, the type was obtained using `instance.substs.type_at(0)` which returns the first monomorphized type for `T` in a generic function. This often matches the pointee type for intrinsic arguments (e.g., `T` resolves to both the monomorphized type and the pointee type in `fn volatile_store<T>(dst: *mut T, val: T)`), but it's not what should be used to build these expressions.

The changes in this PR also ensure that alignment expressions are only built for pointer types (renames `is_aligned` to `is_ptr_aligned` and asserts the type passed is a pointer).

### Testing:

* How is this change tested? Existing regression, in particular tests including alignment checks.

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
@@ -1,6 +1,7 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 //! this module handles intrinsics
+use super::typ::pointee_type;
 use super::PropertyClass;
 use crate::codegen_cprover_gotoc::GotocCtx;
 use cbmc::goto_program::{BuiltinFn, Expr, Location, Stmt, Type};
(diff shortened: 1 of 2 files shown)


celinval (commented):
Quick question. Were there any side effects related to this issue?

adpaco-aws:
> Quick question. Were there any side effects related to this issue?

No, because we only include alignment checks for `volatile_store` and `write_bytes`. In both cases, the monomorphized type for `T` and the pointee type are the same, as mentioned in the PR description. On the other hand, this could be the explanation to my failed attempt to include alignment checks for `transmute` in #1104.

I'm currently working on copy intrinsics which also depend on this correction for aligment checks.

adpaco-aws:
Renamed `dst` to `ptr` because `is_ptr_aligned` can be used for any pointer. Thanks @celinval for the suggestion!

celinval (approved):

celinval:
Awesome! Thanks

Issue #1198, opened 2022-05-13
`copy_nonoverlapping` is handled with `StatementKind`, not an intrinsic


adpaco-aws:
We previously believed that `copy_nonoverlapping` was handled as an intrinsic but the code that handles it is actually in the `StatementKind::CopyNonOverlapping`. We should audit this code as if it were an intrinsic (and reuse as much code from `copy` as possible).

Issue #1163, opened 2022-05-04
Restore and audit remaining intrinsics


adpaco-aws:
A significant amount of intrinsics was disabled in #727 because they could not be audited in time. In addition, some intrinsics were unimplemented.

The goal is the same, for each intrinsic here:
1. Identify what is needed to claim full support.
2. Implement support for it if possible.

This issue will keep track of the work done on them (a check will mark it as "done") and link other issues related to the implementation/rework/cleanup of intrinsics.

 - [ ] `arith_offset`:
 - [ ] `caller_location`:
 - [ ] `ceilf32`:
 - [ ] `ceilf64`:
 - [ ] `copy`:
 - [ ] `copy_nonoverlapping`:
 - [ ] `copysignf32`:
 - [ ] `copysignf64`:
 - [ ] `drop_in_place`:
 - [ ] `exp2f32`:
 - [ ] `exp2f64`:
 - [ ] `expf32`:
 - [ ] `expf64`:
 - [ ] `float_to_int_unchecked`:
 - [ ] `floorf32`:
 - [ ] `floorf64`:
 - [ ] `fmaf32`:
 - [ ] `fmaf64`:
 - [ ] `frem_fast`:
 - [ ] `log2f32`:
 - [ ] `log2f64`:
 - [ ] `log10f32`:
 - [ ] `log10f64`:
 - [ ] `logf32`:
 - [ ] `logf64`:
 - [ ] `maxnumf32`:
 - [ ] `maxnumf64`:
 - [ ] `minnumf32`:
 - [ ] `minnumf64`:
 - [ ] `move_val_init`:
 - [ ] `nearbyintf32`:
 - [ ] `nearbyintf64`:
 - [ ] `nontemporal_store`:
 - [ ] `powf32`:
 - [ ] `powf64`:
 - [ ] `powif32`:
 - [ ] `powif64`:
 - [ ] `prefetch_read_data`:
 - [ ] `prefetch_read_instruction`:
 - [ ] `prefetch_write_data`:
 - [ ] `prefetch_write_instruction`:
 - [ ] `rintf32`:
 - [ ] `rintf64`:
 - [ ] `roundf32`:
 - [ ] `roundf64`:
 - [ ] `rustc_peek`:
 - [ ] `sqrtf32`:
 - [ ] `sqrtf64`:
 - [ ] `truncf32`:
 - [ ] `truncf64`:
 - [ ] `try`:
 - [ ] `unaligned_volatile_load`:
[...]

PR #1081, opened 2022-04-20
Disable unaudited intrinsics

adpaco-aws:
### Description of changes: 

Disables a considerable amount of unaudited intrinsics and updates the documentation to reflect the status. Converts 3 tests to "fixme" because they use or depend on intrinsics that are now disabled.

### Resolved issues:

Part of #727 

### Testing:

* How is this change tested? Converts 3 tests to "fixme".

* Is this a refactor change? No.

### Checklist
- [x] Each commit message has a non-empty body, explaining why the change was made
- [x] Methods or procedures are documented
- [x] Regression or unit tests are included, or existing tests cover the modified code
- [x] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #1199, opened 2022-05-13
Audit for `copy` intrinsic

adpaco-aws:
### Description of changes: 

Completes the audit for the `copy` intrinsic, which was disabled in #1081:
 * Adds alignment checks for both `src` and `dst`.
 * Adds overflow check for the count in bytes (this is not mentioned in the documentation).
 * Refactors the code into a function because macros are harder to debug.
 * Adds tests for all possible cases.

Initially, the changes in this PR were for both `copy` and `copy_nonoverlapping`, but as it has been noted in #1176 the `copy_nonoverlapping` case is handled somewhere else.

### Resolved issues:

Creates #1198 
Part of #1163 

### Testing:

* How is this change tested? Existing regression and 7 new/restored tests.

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
@@ -146,39 +146,6 @@ impl<'tcx> GotocCtx<'tcx> {
         let farg_types = sig.inputs();
         let cbmc_ret_ty = self.codegen_ty(ret_ty);
 
-        /// https://doc.rust-lang.org/core/intrinsics/fn.copy.html
-        /// https://doc.rust-lang.org/core/intrinsics/fn.copy_nonoverlapping.html
-        /// An intrinsic that translates directly into either memmove (for copy) or memcpy (copy_nonoverlapping)
-        macro_rules! _codegen_intrinsic_copy {
-            ($f:ident) => {{
-                let src = fargs.remove(0).cast_to(Type::void_pointer());
-                let dst = fargs.remove(0).cast_to(Type::void_pointer());
-                let count = fargs.remove(0);
-                let sz = {
-                    match self.fn_sig_of_instance(instance).unwrap().skip_binder().inputs()[0]
-                        .kind()
-                    {
-                        ty::RawPtr(t) => {
-                            let layout = self.layout_of(t.ty);
-                            Expr::int_constant(layout.size.bytes(), Type::size_t())
-                        }
-                        _ => unreachable!(),
-                    }
-                };
-                let n = sz.mul(count);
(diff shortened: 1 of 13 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -928,6 +897,65 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    /// An intrinsic that translates directly into `memmove`
+    /// https://doc.rust-lang.org/core/intrinsics/fn.copy.html
+    ///
+    /// Undefined behavior if any of these conditions are violated:
+    ///  * Both `src`/`dst` must be properly aligned (done by alignment checks)
+    ///  * Both `src`/`dst` must be valid for reads/writes of `count *
+    ///      size_of::<T>()` bytes (done by calls to `memmove`)
+    /// In addition, we check that computing `count` in bytes (i.e., the third
+    /// argument for the `memmove` call) would not overflow.
+    fn codegen_copy_intrinsic(
```
celinval, line 909: Functions are so much better!

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -928,6 +897,65 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        loc: Location,
+    ) -> Stmt {
+        // The two first arguments are pointers. It's safe to cast them to void
+        // pointers or directly unwrap the `pointee_type` result as seen later.
+        let src = fargs.remove(0).cast_to(Type::void_pointer());
+        let dst = fargs.remove(0).cast_to(Type::void_pointer());
+
+        // Generate alignment checks for both pointers
+        let src_align = self.is_ptr_aligned(farg_types[0], src.clone());
+        let src_align_check = self.codegen_assert(
```
celinval, line 924: Are these checks relevant if length is zero?

celinval (commented):
Just one question about whether we should check if length is zero before adding the checks.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: Yes, please see the last sentence in "Safety" here: https://doc.rust-lang.org/std/intrinsics/fn.copy.html

> Note that even if the effectively copied size `(count * size_of::<T>())` is 0, the pointers must be non-null and properly aligned.

celinval: perfect! thanks

celinval (approved):
Great test coverage! Thanks

PR #1201, opened 2022-05-16
Audit `copy_nonoverlapping` intrinsic

### LINKED ISSUE #1198: `copy_nonoverlapping` is handled with `StatementKind`, not an intrinsic

adpaco-aws:
### Description of changes: 

Completes the audit for the `copy_nonoverlapping` intrinsic by reusing the work done on the `copy` intrinsic (see #1199):
 * Adds alignment checks for both `src` and `dst`.
 * Adds overflow check for the count in bytes (this is not mentioned in the documentation).
 * Adds tests for all possible cases.

### Resolved issues:

Resolves #1198 
Part of #1163 

### Testing:

* How is this change tested? Existing regression and new tests

### Checklist


kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -731,22 +731,25 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                // We pack the operands and their types, and then call
+                // `codegen_copy_intrinsic` as if it was an intrinsic.
+                let fargs = vec![
+                    self.codegen_operand(src),
+                    self.codegen_operand(dst),
+                    self.codegen_operand(count),
+                ];
+                let farg_types =
+                    &[self.operand_ty(src), self.operand_ty(dst), self.operand_ty(count)];
+                self.codegen_copy_intrinsic(
```
celinval, line 745: It is a bit odd to have this code calling a codegen intrinsic. Any chance you can put the common logic in a a separate function?

tests/expected/intrinsics/copy-nonoverlapping/copy-overlapping/expected
```diff
@@ -0,0 +1 @@
+memcpy src/dst overlap
```
celinval, line 1: Is this a CBMC message? I was wondering if we change it to say copy_nonoverlap here. Do we use memcpy anywhere else?

tests/expected/intrinsics/copy-nonoverlapping/copy-unaligned-dst/expected
```diff
@@ -0,0 +1,2 @@
+FAILURE\
+`dst` is properly aligned
```
celinval, line 2: The failure message usually describes an error or an expectation that should hold.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -897,21 +899,28 @@ impl<'tcx> GotocCtx<'tcx> {
         Stmt::atomic_block(vec![skip_stmt], loc)
     }
 
-    /// An intrinsic that translates directly into `memmove`
+    /// An intrinsic that translates directly into `memmove` (for `copy`)
```
celinval, line 902: This is a bit confusing. It will probably benefit from the refactoring I suggested since this function does more than generating code for a specific intrinsic.

celinval (changes requested):
Thanks for reducing the duplicated code. I just have a few comments besides my previous comment that I published too soon. ☺️

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
adpaco-aws: Sorry, I don't know what you mean. The common logic is already in `codegen_copy_intrinsic`, I'm not sure if you want to move it somewhere else or split it into several functions.

danielsn: Maybe just rename it to not have the word "intrinsic"?

tests/expected/intrinsics/copy-nonoverlapping/copy-overlapping/expected
adpaco-aws: Yes, this is a CBMC message so there's not much we can do. We use `memcpy` in our `__rust_realloc` C implementation but we don't generate calls to `memcpy` anywhere else.

tests/expected/intrinsics/copy-nonoverlapping/copy-unaligned-dst/expected
adpaco-aws: The expectation is that the pointers are aligned. We can change all alignment checks to "X must be properly aligned", would that do it for you?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: As earlier, I don't know what is wrong with this. It's not generating code for a specific intrinsic, but for a specific type of intrinsics (i.e., "copy" intrinsics). This is similar to what we already had before the audit, but using a function instead of a macro.

tests/expected/intrinsics/copy-nonoverlapping/copy-unaligned-dst/expected
celinval: Yes. This message I read as there is a failure because `dst` is properly aligned. Thanks!

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
celinval: The problem is that `StatementKind::CopyNonOverlapping` is not an intrinsic.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -899,22 +899,25 @@ impl<'tcx> GotocCtx<'tcx> {
         Stmt::atomic_block(vec![skip_stmt], loc)
     }
 
-    /// An intrinsic that translates directly into `memmove` (for `copy`)
-    /// or `memcpy` (for `copy_nonoverlapping`)
+    /// Copies `count * size_of::<T>()` bytes from `src` to `dst`.
     /// https://doc.rust-lang.org/core/intrinsics/fn.copy.html
```
celinval, line 903: nit: I would move these links to be under 1. and 2. below.

celinval (approved):
Looks good! Thanks

Issue #1208, opened 2022-05-19
Overflow checks should abort the execution


celinval:
I tried this code:

```rust
#[kani::proof]
fn check_sub_overflow() {
    let a = [0; 5];
    let ptr0: *const i32 = &a[0];
    let ptr1: *const i32 = &a[1];
    let _: usize = ptr0 as usize - ptr1 as usize;
    unreachable!("Previous statement should fail");
}
```

using the following command line invocation:

```
kani check.rs
```

with Kani version: *0.2*

I expected to see this happen: There should be an overflow check failure but the unreachable code should not panic since it is unreachable ([playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1cca359f4da63c0b73b228acd0a6396d)).

Instead, this happened: Both checks fail, the overflow check and the unreachable panic.

adpaco-aws:
Yes, there are many places where we assert false but we should be using (assert + assume) false. If I'm not mistaken, this would be covered by the "assertions audit" described in #1056.

PR #1210, opened 2022-05-20
Add "fixme" testcase for #1208

celinval:
### Description of changes: 

Our checks currently don't abort the execution flow. This adds a test case for it.

I also added an override for the `unreachable!()` macro to print nicer error message.

### Resolved issues:

N/A

### Call-outs:

Since this issue is related to how we handle a check failure, I couldn't use the `fixme` regression. So I added the test to the `expected` regression.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/library/std/src/lib.rs
+++ b/library/std/src/lib.rs
@@ -137,3 +137,11 @@ macro_rules! println {
 macro_rules! eprintln {
     ($($x:tt)*) => { evaluate_print_args!($($x)*); };
 }
+
+#[macro_export]
+macro_rules! unreachable {
+    ($($arg:tt)*) => ({
+        panic!(concat!("internal error: entered unreachable code: ",
+        stringify!($($arg)*)));
+    });
+}
--- a/tests/expected/arith_checks/fixme_check_arith_failures.rs
+++ b/tests/expected/arith_checks/fixme_check_arith_failures.rs
@@ -1,0 +1,13 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Check that code after an overflow check failure is unreachable.
+
+#[cfg_attr(kani, kani::proof)]
+fn main() {
+    let a = [0; 5];
+    let ptr0: *const i32 = &a[0];
+    let ptr1: *const i32 = &a[1];
+    let _: usize = ptr0 as usize - ptr1 as usize;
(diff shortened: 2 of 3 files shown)


adpaco-aws (commented):
Please add `unreachable` to the overrides section in our documentation.

library/std/src/lib.rs
```diff
@@ -137,3 +137,11 @@ macro_rules! println {
 macro_rules! eprintln {
     ($($x:tt)*) => { evaluate_print_args!($($x)*); };
 }
+
+#[macro_export]
+macro_rules! unreachable {
+    ($($arg:tt)*) => ({
+        panic!(concat!("internal error: entered unreachable code: ",
```
zhassan-aws, line 144: How does this compare to the std version?

tests/expected/arith_checks/fixme_check_arith_failures.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Check that code after an overflow check failure is unreachable.
```
zhassan-aws, line 4: Perhaps change the comment to say: should be unreachable, and add a link to #1208?

tests/expected/arith_checks/fixme_check_arith_failures.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[cfg_attr(kani, kani::proof)]
```
zhassan-aws, line 6: Is this to allow us to run it with `rustc`? (Neat!)

celinval: Yeah, but I can remove it. I use that a lot when comparing the two but sometimes I forget to remove them.

tests/expected/arith_checks/fixme_check_arith_failures.rs
celinval: Indeed!

library/std/src/lib.rs
celinval: The message is basically the same except for the following:
 - we still inclue ":" in the no arguments case.
 - the std one use format
 - we print extra quotes for the message

zhassan-aws (approved):

Issue #25, opened 2021-04-09
Extend unstable atomic tests with generic types


adpaco-aws:
Currently unstable atomic tests use only `u8` values, but we should extend them to work with all integer values supported in Kani.

PR #1212, opened 2022-05-23
Add support for `atomic_min*` and `atomic_umin*` intrinsics

adpaco-aws:
### Description of changes: 

Defines a `min` expression on top of other expressions. Then uses the `min` expression to add support for `atomic_min*` and `atomic_umin*` intrinsics.

### Resolved issues:

Part of #1163 

### Testing:

* How is this change tested? Removes 10 tests and adds 2.

### Checklist


cprover_bindings/src/goto_program/expr.rs
```diff
@@ -1080,6 +1080,14 @@ impl Expr {
     pub fn r_ok(self, e: Expr) -> Expr {
         self.binop(ROk, e)
     }
+
+    // Expressions defined on top of other expressions
+
+    /// `min(self, e)`
+    pub fn min(self, e: Expr) -> Expr {
```
danielsn, line 1087: If there are side effects, this will evaluate them twice

adpaco-aws: Please take a look at the new changes, which complete the definition for `is_side_effect` and assert that neither of these cause side effects.

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -282,13 +282,44 @@ impl Expr {
...
+        /// Returns whether any of the expressions on a vector
+        /// causes side effects or not
+        fn are_side_effect(exprs: &Vec<Expr>) -> bool {
+            for expr in exprs {
+                if expr.is_side_effect() {
+                    return true;
+                }
+            }
+            false
+        }
```
danielsn, line 297: Style ⛏️ : you could use `exprs.iter().any(|e| e.is_side_effect())`

tests/kani/Intrinsics/Atomic/Unstable/AtomicMin/main.rs
```diff
@@ -0,0 +1,47 @@
...
+// expected result.
+
+#![feature(core_intrinsics)]
+use std::intrinsics::{
+    atomic_min, atomic_min_acq, atomic_min_acqrel, atomic_min_rel, atomic_min_relaxed,
+};
+
+#[kani::proof]
+fn main() {
+    let mut a1 = 1 as u8;
```
danielsn, line 14: Is there a reason we only test for `u8`? Could we make this take a `<T>` and then test multiple bitwidths?

cprover_bindings/src/goto_program/expr.rs
adpaco-aws: Thanks!

tests/kani/Intrinsics/Atomic/Unstable/AtomicMin/main.rs
adpaco-aws: Yes, that's pending in #25. But I'd like to do it once all atomic tests are in.

danielsn (approved):

PR #1215, opened 2022-05-23
Integrate s2n-quic in the regressions

### LINKED ISSUE #958: Integrate s2n-quic as a git submodule
Run some/all of the existing harnesses on every push to main.

This would help in verifying some of our soundness issues on a real code-base.

zhassan-aws:
### Description of changes: 

Integrate s2n-quic as a submodule and run the current Kani proofs on every push.

The s2n-quic proofs are added to a new `perf` test suite that is **not** run by default with `./scripts/kani-regressions`. One has to pass an additional argument, `all`, to run this suite.

### Resolved issues:

Resolves #958 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/.github/workflows/kani.yml
+++ b/.github/workflows/kani.yml
@@ -24,7 +24,12 @@ jobs:
           cargo build --workspace
 
       - name: Execute Kani regression
+        if: ${{ github.event_name == 'pull_request' }}
         run: ./scripts/kani-regression.sh
+
+      - name: Execute all Kani regressions
+        if: ${{ github.event_name == 'push' }}
+        run: ./scripts/kani-regression.sh all
 
   bookrunner:
     runs-on: ubuntu-20.04
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "firecracker"]
 	path = firecracker
 	url = https://github.com/firecracker-microvm/firecracker.git
+[submodule "tests/perf/s2n-quic"]
+	path = tests/perf/s2n-quic
+	url = https://github.com/zhassan-aws/s2n-quic
--- a/scripts/kani-regression.sh
+++ b/scripts/kani-regression.sh
@@ -8,6 +8,17 @@ fi
 set -o pipefail
 set -o nounset
 
+ARG=
+# Check for optional arguments
+if (( "$#" > 0 )); then
+  ARG=$1
+  if [ "$ARG" != "all" ]; then
+    echo "Unexpected argument: $ARG"
+    echo "Usage: $0 [all]"
+    exit 1
+  fi
+fi
+
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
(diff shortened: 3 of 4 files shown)


zhassan-aws:
@camshaft 

This is based on this commit: https://github.com/zhassan-aws/s2n-quic/tree/3017e24a987c6adec34f51f3618530ca65a4f9fc in my fork of s2n-quic

.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "firecracker"]
 	path = firecracker
 	url = https://github.com/firecracker-microvm/firecracker.git
+[submodule "tests/perf/s2n-quic"]
+	path = tests/perf/s2n-quic
+	url = https://github.com/zhassan-aws/s2n-quic
```
celinval, line 15: Why not s2n-quic's repo?

scripts/kani-regression.sh
```diff
@@ -8,6 +8,17 @@ fi
 set -o pipefail
 set -o nounset
 
+ARG=
+# Check for optional arguments
+if (( "$#" > 0 )); then
```
celinval, line 13: Do we expect developers to run these tests locally? I would prefer if this script always run the same set of tests.

.github/workflows/kani.yml
```diff
@@ -24,8 +24,13 @@ jobs:
           cargo build --workspace
 
       - name: Execute Kani regression
+        if: ${{ github.event_name == 'pull_request' }}
         run: ./scripts/kani-regression.sh
 
+      - name: Execute all Kani regressions
+        if: ${{ github.event_name == 'push' }}
+        run: ./scripts/kani-regression.sh all
```
celinval, line 32: What do you think about creating a separate step named Performance Tests instead?

celinval (commented):
How long does it take?

camshaft:
After https://github.com/aws/s2n-quic/pull/1327 is merged we should be able to just use the main repo

.github/workflows/kani.yml
zhassan-aws: Sure. That would work too.

.gitmodules
zhassan-aws: We generally need to make changes to integrate with our regressions:
1. Adding Kani harnesses (this may not be needed if we're able to run the Bolero harnesses directly using Kani)
2. Adding expected files so that they work with compiletest 
Since we can't push changes to s2n-quic, I created a fork that we can push those changes to.

scripts/kani-regression.sh
zhassan-aws: I think so (at least for debugging purposes). I can add this to a separate script, although it fitted nicely within this script since all the setup steps are handled.

zhassan-aws (commented):
> How long does it take?

About 10 minutes.

celinval:
This job should probably run against the release mode too

zhassan-aws:
Closing in favor of #1267. Most of the comments in this PR are addressed in #1267.

PR #1218, opened 2022-05-24
Refactor kani-verifier

tedinski:
### Description of changes: 

This should exhibit no observable behavior changes, but refactors and organizes the code in `kani-verifier` (i.e. the proxy binaries).

I wanted to submit the refactor separately before I started working on the actual behavior changes.

### Call-outs:

1. If we find the right small dependency to use, we could possibly fully eliminate `cmd.rs`
2. All setup code is now under `setup.rs`
3. All steps in setup have been factored out into their own self-contained functions.
4. Very minor changes were then motivated:
    * Ensure `kani_dir` exists initially instead of `base_dir`. This should be fine.
    * I installed all python deps in one command now instead of two.

### Testing:

* How is this change tested? docker tests

### Checklist

### DIFF
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -1,0 +1,60 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This module contains small helper functions for running Commands.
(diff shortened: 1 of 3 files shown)


zhassan-aws (approved):
LGTM

PR #1220, opened 2022-05-25
Fix initial index in CBMC output parser

### LINKED ISSUE #842: Confusing checks generated for an empty program
I tried this code:

```rust
fn main() {}
```

```
kani empty.rs
```

with Kani version: `541a3100d88`

I expected to see this happen: No checks.

Instead, this happened:
```
RESULTS:
Check 1: assertion.1
         - Status: SUCCESS
         - Description: "Code generation sanity check: Correct CBMC vtable size for StructTag("tag-_13892494147292364991") (MIR type Closure(DefId(2:9427 ~ std[3d6e]::rt::lang_start::{closure#0}), [(), i8, extern "rust-call" fn(()) -> i32, (fn(),)])). Please report failures:
https://github.com/model-checking/kani/issues/new?template=bug_report.md"

Check 2: pointer_primitives.1
         - Status: SUCCESS
         - Description: "dead object in OBJECT_SIZE(&temp_0)"


SUMMARY: 
 ** 0 of 2 failed

VERIFICATION:- SUCCESSFUL
```
If these are internal sanity checks, they should perhaps be put under a `--codegen-sanity-checks` flag that is off by default?

zhassan-aws:
### Description of changes: 

For programs without any checks, Kani currently reports `0 of 1 failed` and doesn't display any checks. This PR fixes this issue so that Kani reports `0 of 0 failed` instead.

### Resolved issues:

Resolves #842 

### Testing:

### Checklist
- [X]  Regression or unit tests are included, or existing tests cover the modified code

### DIFF
--- a/scripts/cbmc_json_parser.py
+++ b/scripts/cbmc_json_parser.py
@@ -683,7 +683,7 @@ def construct_terse_property_message(properties):
     number_tests_failed = 0
     output_message = ""
     failed_tests = []
-    index = 0
+    index = -1
     verification_status = GlobalMessages.FAILED
 
     # Parse each property instance in properties
@@ -752,7 +752,7 @@ def construct_property_message(properties):
     number_tests_undetermined = 0
     output_message = ""
     failed_tests = []
-    index = 0
+    index = -1
     verification_status = GlobalMessages.FAILED
 
     output_message = "RESULTS:\n"
--- a/scripts/cbmc_json_parser.py
+++ b/scripts/cbmc_json_parser.py
@@ -683,11 +683,10 @@ def construct_terse_property_message(properties):
     number_tests_failed = 0
     output_message = ""
     failed_tests = []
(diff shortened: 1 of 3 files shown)


scripts/cbmc_json_parser.py
```diff
@@ -683,7 +683,7 @@ def construct_terse_property_message(properties):
     number_tests_failed = 0
     output_message = ""
     failed_tests = []
-    index = 0
+    index = -1
```
adpaco-aws, line 686: We should avoid using `index` as both flag and counter. Please declare another variable to store `len(properties)` and use that in `output_message` instead.

That way, we ensure `index` in the iterator is always used within its scope.

adpaco-aws (changes requested):

adpaco-aws (approved):
Thanks!

PR #1228, opened 2022-05-26
Audit for `arith_offset` intrinsic

adpaco-aws:
### Description of changes: 

This PR adds support for the `arith_offset` intrinsic (used by `wrapping_offset`) by using the already-implemented `codegen_offset` function.

The [documentation for `wrapping_offset`](https://doc.rust-lang.org/std/primitive.pointer.html#method.wrapping_offset) defines it as a more permissive version of `offset` which allows the computed offset to go out-of-bounds. Currently, we don't check in `codegen_offset` that the computed offset is within bounds to avoid spurious failures, so its definition should work for both.

### Resolved issues:

Part of #1163 

### Testing:

* How is this change tested? Adds 6 tests.

### Checklist


tedinski (approved):
There's a note in your PR description that isn't reflected in the comments of `codegen_offset` (that these two instrinsics should have different behavior?)

Should there be a tracking issue for this? If so, can you create that and link to it from that function's comments too?

Approving because with that minor change to comments, LGTM

adpaco-aws:
Thanks @tedinski , you're right that we should be tracking this.

I created #1233 and extended the comment to both link that issue and clarify that it's only concerning `offset`. If we had to add that check at some point, it would be enough to add a flag to check for this property in `codegen_offset`.

danielsn (approved):

PR #1230, opened 2022-05-26
Add a zero_initializer function

danielsn:
### Description of changes: 

Right now, we can only `zero_init` scalar types. This adds a recursive form to allow us to `zero_init` complex types as well.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Call-outs:

Using this may improve performance, at a potential soundness cost. Since the tradeoff is not clear, putting it behind a flag to allow experimentation, but keeping the current default for now.

### Testing:

* How is this change tested? 
    1. Existing regression tests. Turned on the flag, all regressions still pass, except for ones testing maybe_uninit, which we would expect to fail.
    2. Existing regression tests. Flag off, all regressions still pass.
    3. New test which maybe_units a variable and checks that it is all 0.

### Checklist


tautschnig:
I re-enabled the RustRealloc hook that was removed in #1088 and then tried the examples from #702 with this PR. I got:
```
$ kani 702.rs --enable-unstable --zero-init-vars
thread 'rustc' panicked at 'Can't zero init Code { parameters: [Parameter { typ: Pointer { typ: CInteger(SizeT) }, identifier: None, base_name: None }, Parameter { typ: Pointer { typ: CInteger(SizeT) }, identifier: None, base_name: None }], return_type: StructTag("tag-_11659839863328250985") }', cprover_bindings/src/goto_program/typ.rs:1262:38

If you are seeing this message, please file an issue here: https://github.com/model-checking/kani/issues/new?labels=bug&template=bug_report.md

[Kani] current codegen item: codegen_function: std::cmp::max_by
_RINvNtCs8yXqNEgggfL_4core3cmp6max_byjNvYjNtB2_3Ord3cmpECs1cE2uPJmwEu_3_702
[Kani] current codegen location: Loc { file: "/home/ubuntu/.rustup/toolchains/nightly-2022-05-17-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cmp.rs", function: None, line: 1276, col: Some(1) }
Error: [...]/kani-compiler exited with status exit status: 101
```

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -1239,7 +1250,63 @@ impl Type {
...
+                    components.iter().map(|c| c.typ().zero_initializer(st)).collect();
+                Expr::struct_expr_from_padded_values(self.clone(), values, st)
+            }
+            StructTag(tag) => st.lookup(*tag).unwrap().typ.zero_initializer(st),
+            TypeDef { .. } => unreachable!("Should have been normalized away"),
+            Union { components, .. } => {
+                if components.is_empty() {
+                    Expr::empty_union(self.clone(), st)
+                } else {
+                    let largest = components.iter().max_by_key(|c| c.sizeof_in_bits(st)).unwrap();
```
fzaiser, line 1284: Nit: This could be written in a nicer way with `if let Some(largest) = ... { ... } else { /* empty case */ }`, thus avoiding unwrap().

kani-driver/src/args.rs
```diff
@@ -83,6 +83,11 @@ pub struct KaniArgs {
     #[structopt(flatten)]
     pub checks: CheckArgs,
 
+    /// Zero initilize variables.
+    /// This is an unstable feature.
```
fzaiser, line 87: Add a comment mentioning the performance vs soundness tradeoff?

fzaiser:
Looks good to me. Two things:
* Is there a way this could be tested? Specifically test that it does zero initialization? (Not sure it's worth the hassle.)
* Did you observe a performance improvement on any examples?

fzaiser:
One more thing: shouldn't we zero-initialize `StatementKind::DeInit` as well? As far as I can tell, that part of the code still uses `nondet`. In particular, `DeInit` is used to "initialize" generators, so they wouldn't benefit from the PR as is.

kani-driver/src/args.rs
danielsn: Done

celinval (commented):
Can you at least add a few sanity tests? Have you tried measuring the performance gain at all? I am a bit worried about this PR since it is doing something way out of the rust spec.

fzaiser:
Looks good to me, except I think this should have a test. Running Kani with the flag and checking that the value of an uninitialized variable is zero (or something to that effect). Or is there a reason not to test this?

fzaiser (approved):

celinval (changes requested):
Thanks for adding a test, but it does show how this feature is unsound. I don't think we should add this code to Kani.

I would prefer if you publish a separate PR for the code that isn't directly related to this feature and keep the zero_initializer stuff in a branch for now.

If it is too painful to keep the `zero_initializer` logic on a branch, then I would suggest that we remove the kani option and just use a constant to enable / disable this code.

danielsn:
What about an --enable-unsound flag

PR #404, opened 2021-08-06
Enable RMC on Ubuntu 18.04 (#393)

celinval:
### Description of changes: 

We don't currently support Ubuntu 18.04 even though it is still a widely used distro. Ubuntu 18.04 uses python 3.6 by default and it does not have the package python-is-python3. This change removes the dependency on python 3.8 constructs and adds Ubuntu 18.04 to our CI.

### Resolved issues:

Resolves #393 

### Testing:

* How is this change tested? 
  - Ran the new workflow in my RMC fork [here](https://github.com/celinval/rmc-dev/actions/runs/1106588806).

* Is this a refactor change?
  - No

### Checklist
- [x] Each commit message has a non-empty body, explaining why the change was made
- [x] Methods or procedures are documented
- [x] Regression or unit tests are included, or existing tests cover the modified code
- [x] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #1231, opened 2022-05-26
Add support for Ubuntu 18.04 to installer

tedinski:
### Description of changes: 

1. Lower version for "release bundle" tests to build on 18.04 just like normal releases (for glibc compatibility)
2. Test install on both 18.04 and 20.04
3. An unfortunately ugly fix for the bug on 18.04 with python pip

### Resolved issues:

Resolves #404 

### Call-outs:

1. Introduced `os_hacks` to `kani-verifier`, as a place to hopefully confine os-specific logic without making it harder to see the general case. I'll be adding a nixos one next.
2. The `os_info` crate was just the first one I found for this purpose basically...

### Testing:

* How is this change tested? docker test

### Checklist


src/setup.rs
```diff
@@ -103,15 +104,26 @@ fn setup_rust_toolchain(kani_dir: &Path) -> Result<String> {
...
-fn setup_python_deps(kani_dir: &Path) -> Result<()> {
+fn setup_python_deps(kani_dir: &Path, os: &os_info::Info) -> Result<()> {
     println!("[4/6] Installing Kani python dependencies...");
     let pyroot = kani_dir.join("pyroot");
 
     // TODO: this is a repetition of versions from kani/scripts/setup/$OS/install_deps.sh
+    let pkg_versions = &["cbmc-viewer==3.2", "colorama==0.4.3"];
+
+    if os.os_type() == os_info::Type::Ubuntu
+        && *os.version() == os_info::Version::Semantic(18, 4, 0)
```
danielsn, line 115: Is there a risk there will be an 18.4.1?

.github/workflows/kani.yml
```diff
@@ -75,11 +75,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-10.15, ubuntu-20.04]
+        os: [macos-10.15, ubuntu-18.04]
         include:
           - os: macos-10.15
             artifact: kani-latest-x86_64-apple-darwin.tar.gz
-          - os: ubuntu-20.04
+          - os: ubuntu-18.04
```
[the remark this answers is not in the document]
tedinski: "get rid of" in what sense? Are you saying not building the regressions (currently 18, 20, and mac) there anymore?

I would sort of endorse that, in that I think we don't need the **regressions** running on two linux platforms... but what we DO need is "install/smoke testing" running on multiple linux platforms... which is sort of what this is simulating, but it's not good enough yet.

I spent an hour with a whiteboard yesterday trying to sketch out what I thought we'd ideally want to work towards in terms of CI flow and found it a bit frustratingly complicated. I might write up a doc/proposal soon...

src/setup.rs
tedinski: I think there is not a risk of that because I think it's already running on "18.4.4" and not picking that up. I believe what's happening is it's parsing "18.04" as if it were a version number.

I agree this is seems fragile but this unit test seems to confirm this is the expected behavior: https://github.com/stanislav-tkach/os_info/blob/master/os_info/src/linux/lsb_release.rs#L317-L318

.github/workflows/kani.yml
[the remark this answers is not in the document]
zhassan-aws: @danielsn Note that this is for building the release bundle, and we're building a single bundle for both Ubuntu 18.04 and Ubuntu 20.04.

src/os_hacks.rs
```diff
@@ -0,0 +1,44 @@
...
+        .run()?;
+
+    // Step 2: move `pyroot/lib/python3.6/site-packages/*` up to `pyroot`
+    // This seems to successfully replicate the behavior of `--target`
+    // "mv" is not idempotent however so we need to do "cp -r"
+    let mut mv_cmd = OsString::new();
+    mv_cmd.push("cp -r ");
+    mv_cmd.push(pyroot.as_os_str());
+    mv_cmd.push("/lib/python*/site-packages/* ");
+    mv_cmd.push(pyroot.as_os_str());
```
zhassan-aws, line 40: Is there a risk that the target directory is not writeable (or requires `sudo`)?

src/os_hacks.rs
```diff
@@ -0,0 +1,44 @@
...
+    // Step 1: use `--system --prefix pyroot`. This disables the broken behavior, and creates `bin` but...
+    Command::new("python3")
+        .args(&["-m", "pip", "install", "--system", "--prefix"])
+        .arg(&pyroot)
+        .args(pkg_versions)
+        .run()?;
+
+    // Step 2: move `pyroot/lib/python3.6/site-packages/*` up to `pyroot`
+    // This seems to successfully replicate the behavior of `--target`
+    // "mv" is not idempotent however so we need to do "cp -r"
```
zhassan-aws, line 35: If we're doing `cp` instead of `mv`, should we:
1. Rename the variable
2. Delete the source directory after copying it
?
I'm a bit unclear on the issue with `mv`.

src/os_hacks.rs
tedinski: Nope, `pyroot` here is `~/.kani/kani-VERSION/pyroot` and we created it.

src/os_hacks.rs
tedinski: `mv` makes the `setup` procedure non-idempotent because it fails if the directory already exists (from a previous install).

I can rename the variable and delete the extra directory, that seems reasonable.

PR #1235, opened 2022-05-27
Add support for determining which crate the current function came from

### LINKED ISSUE #1234: Detect when we are compiling std
std is allowed to make use of constructs that other crates are not. We should codegen it differently

danielsn:
### Description of changes: 

Currrently, there is no easy way to tell which crate a function came from.
This is a problem, because `std` and `core` are allowed to make implementation choices and use features that would be UB if used in other crates.
Adding a mechanism to detect when code comes from `core` or `std` is the first step in allowing us to specalize checks for these libraries.

Once this is merged, we should be able to go `self.current_fn().is_std()` to guard whether `std` specific checks should be used.

### Resolved issues:

Resolves #1234

### Call-outs:

Had to call things `krate` cause `crate` is a reserved name.

### Testing:

* How is this change tested?  I dumped out debug info, and manually audited it, and it it did the right thing. I'd love to hear suggestions for better testing.

### Checklist


kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
```diff
@@ -81,6 +84,11 @@ impl<'tcx> CurrentFnCtx<'tcx> {
         self.instance
     }
 
+    /// The crate that function came from
+    pub fn krate(&self) -> String {
+        self.krate.to_string()
```
zhassan-aws, line 89:
```suggestion
        self.krate.clone()
```
?

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
```diff
@@ -104,6 +112,11 @@ impl<'tcx> CurrentFnCtx<'tcx> {
 
 /// Utility functions
 impl CurrentFnCtx<'_> {
+    /// Is the current function from the `std` crate?
+    pub fn is_std(&self) -> bool {
```
zhassan-aws, line 116: Will this only be true for generic `std`/`core` functions, i.e. that will be monomorphized? I'm assuming non-generic `std` functions are not codegen unless we're compiling the `std` crate itself?

zhassan-aws (commented):
I'm generally not in favor of doing something specific for `std`/`core`. We could not conclude previously whether `miri` was treating `std` specially (I doubt this is the case).

Is there a need for this given the workarounds we've put in place of turning off the relevant CBMC checks and injecting them ourselves where needed?

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
danielsn: I think this also applies to std functions that get inlined in.

danielsn:
> I'm generally not in favor of doing something specific for `std`/`core`. We could not conclude previously whether `miri` was treating `std` specially (I doubt this is the case).
> 
> Is there a need for this given the workarounds we've put in place of turning off the relevant CBMC checks and injecting them ourselves where needed?

The issue is exactly those workarounds: we've had to turn off valid checks because they cause failures in `std`.  If we know what functions come from `std`, we can selectively turn off those checks only for std functions.

zhassan-aws:
> If we know what functions come from `std`, we can selectively turn off those checks only for std functions.

I understand. What is not clear for me is whether we want to turn off those checks globally for all of std, or whether there are specific instances (e.g. dangling pointer) that we want to turn off those checks for.

The other alternative that was discussed is for rustc code to use private intrinsics for operations that may have UB, where those intrinsics would not have the same preconditions.

zhassan-aws (approved):
Agreed in an offline discussion that this PR will at least allow us to re-enable some checks on non-`std` crates. Ideally, we'd like to also enable those checks for `std` code that is **not** known to rely on undefined behavior (e.g. dangling pointer usage inside `RawVec`), but doing so would require a stubbing/override mechanism that we don't have at the moment.

PR #1247, opened 2022-06-01
Report CBMC verification time

zhassan-aws:
### Description of changes: 

Currently, there is no easy way to determine the time spent in the verification step (CBMC). What I typically do is to run `kani` or `cargo kani` with `/usr/bin/time -p`, but the reported time includes the time spent in compilation, codegen, verification, and processing of CBMC output. This PR adds reporting of the elapsed time for the verification step, which is needed when doing performance comparisons. The output looks like this:
```
SUMMARY: 
 ** 0 of 811 failed (20 unreachable)

VERIFICATION:- SUCCESSFUL


Verification Time: 3.579173s
```

### Testing:

### Checklist


adpaco-aws:
CBMC reports runtimes for different processes (SSA conversion, solver, etc.), including decision procedure. Is that not good for some reason?

kani-driver/src/call_cbmc.rs
```diff
@@ -40,7 +41,9 @@ impl KaniSession {
             // extra argument
             cmd.arg("--json-ui");
 
+            let now = Instant::now();
             let _cbmc_result = self.run_redirect(cmd, &output_filename)?;
+            let elapsed = now.elapsed().as_secs_f32();
```
jaisnan, line 46: Are there any possible scenarios where there might be no time captured in the variable elapsed?

jaisnan, line 46: Are there any possible scenarios where there might be no time captured in the variable elapsed?

kani-driver/src/call_cbmc.rs
```diff
@@ -51,6 +54,7 @@ impl KaniSession {
                 // the best possible fix is port to rust instead of using python, or getting more
                 // feedback than just exit status (or using a particular magic exit code?)
             }
+            println!("Verification Time: {}s", elapsed);
```
jaisnan, line 57: In case of non-termination, this would line would never be reached right?

jaisnan, line 57: In case of non-termination, this would line would never be reached right?

kani-driver/src/call_cbmc.rs
danielsn, line 46: Why `f32`?

tests/expected/report/verification-time/main.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Checks that the time spent in CBMC is reported as "Verification time: <>s"
```
danielsn, line 4: The actual check is just that there is a line that says `Verification Time:`, correct?

zhassan-aws:
> CBMC reports runtimes for different processes (SSA conversion, solver, etc.), including decision procedure. Is that not good for some reason?

Yes, there is no report of the overall CBMC runtime (which is necessary for performance comparisons). If I add up all the runtimes reported by CBMC on one example:
```
$ grep Runtime cbmc.log 
Runtime Symex: 0.289264s
Runtime Postprocess Equation: 0.00119202s
Runtime Convert SSA: 0.0894082s
Runtime Post-process: 0.00223362s
Runtime Solver: 0.0984715s
Runtime decision procedure: 0.188598s
```
they add up to less than 1 second, whereas the total CBMC runtime reported by the new line I added is more than 3 seconds:
```
Verification Time: 3.548905s
```

kani-driver/src/call_cbmc.rs
zhassan-aws: Can't think of a scenario.

zhassan-aws: What's the alternative you have in mind? Is it `f64`? If so, we probably don't need that much precision, but probably won't hurt much either.

kani-driver/src/call_cbmc.rs
zhassan-aws: Yes, and also in case CBMC exited with an error.

tests/expected/report/verification-time/main.rs
zhassan-aws: Yes. I'll clarify the comment a bit.

tedinski:
> they add up to less than 1 second, whereas the total CBMC runtime

Heh. I wonder if we can get CBMC to add a feature that reports start/end timestamps instead of time.

It'd be nice to create the kind of trace diagrams that `cargo build --timings` gives.

zhassan-aws:
> Heh. I wonder if we can get CBMC to add a feature that reports start/end timestamps instead of time.
> 
> It'd be nice to create the kind of trace diagrams that `cargo build --timings` gives.

That would be neat. I didn't know about `--timings`, but it would be a useful addition.

danielsn (approved):

PR #1248, opened 2022-06-02
Audit for `maxnumf*` and `minnumf*` intrinsics

adpaco-aws:
### Description of changes: 

Restores and completes the audit for `maxnumf32`, `maxnumf64`, `minnumf32` and `minnumf64`. These depend on a set of CBMC builtins (`fmaxf`, `fmax`, `fminf` and `fmin`) that handle NaN arguments in a specific way.

### Resolved issues:

Part of #1163 
Part of #1025 

### Call-outs:

`fmaxf` is not working for an unknown reason, but I think we should restore it in order to ease debugging. The "fixme" test I added for it should work once the issue gets fixed in CBMC (if that's the reason) without needing any changes from Kani.

### Testing:

* How is this change tested? Adds 4 tests (1 of them is a "fixme").

### Checklist

### DIFF
--- a/tests/kani/Intrinsics/MaxNum/maxnumf32_fixme.rs
+++ b/tests/kani/Intrinsics/MaxNum/maxnumf32_fixme.rs
@@ -38,7 +38,11 @@ fn test_one_nan() {
     let y: f32 = kani::any();
     kani::assume((x.is_nan() && !y.is_nan()) || (!x.is_nan() && y.is_nan()));
     let res = std::intrinsics::maxnumf32(x, y);
-    assert!(!res.is_nan());
+    if x.is_nan() {
+        assert!(res == y);
+    } else {
+        assert!(res == x);
+    }
 }
 
 #[kani::proof]
--- a/tests/kani/Intrinsics/MaxNum/maxnumf32_fixme.rs
+++ b/tests/kani/Intrinsics/MaxNum/maxnumf32_fixme.rs
@@ -1,0 +1,51 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Check that `maxnumf32` returns the minimum of two values, except in the
+// following cases:
+//  * If one of the arguments is NaN, the other arguments is returned.
+//  * If both arguments are NaN, NaN is returned.
+#![feature(core_intrinsics)]
+
+// Note: All test cases are failing with the following error:
+// ```
+// Check X: fmaxf.assertion.1
+//          - Status: FAILURE
+//          - Description: "Function with missing definition is unreachable"
+//          - Location: <builtin-library-fmaxf> in function fmaxf
+// ```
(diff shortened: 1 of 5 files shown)


tests/kani/Intrinsics/MaxNum/maxnumf32_fixme.rs
```diff
@@ -0,0 +1,51 @@
...
+// Check X: fmaxf.assertion.1
+//          - Status: FAILURE
+//          - Description: "Function with missing definition is unreachable"
+//          - Location: <builtin-library-fmaxf> in function fmaxf
+// ```
+// This is because the `fmaxf` definition is not being found, either because
+// Kani does not produce the right expression (which is strange, because it's
+// doing the same for similar expressions and they work) or CBMC is not picking
+// it for some reason.
+// Tracked in https://github.com/model-checking/kani/issues/1025
```
tedinski, line 21: Michael seemed to believe the issue there was fixed, have you talked to him about this case?

tedinski (approved):

tests/kani/Intrinsics/MaxNum/maxnumf32_fixme.rs
adpaco-aws: I added a comment in #1025 about this. In principle, all these builtins should be fixed, but `fmaxf` is not working for some reason.

tests/kani/Intrinsics/MinNum/minnumf64.rs
```diff
@@ -0,0 +1,39 @@
...
+    }
+}
+
+#[kani::proof]
+fn test_one_nan() {
+    let x: f64 = kani::any();
+    let y: f64 = kani::any();
+    kani::assume((x.is_nan() && !y.is_nan()) || (!x.is_nan() && y.is_nan()));
+    let res = std::intrinsics::minnumf64(x, y);
+    assert!(!res.is_nan());
```
danielsn, line 29: Could you check that the result is the expected value, not just not NAN?

tests/kani/Intrinsics/MaxNum/maxnumf32_fixme.rs
tautschnig: `fmaxf` and `fmaxl` were affected, https://github.com/diffblue/cbmc/pull/6904 fixes this.

tautschnig: ... which is now merged, and will be part of the next release.

adpaco-aws: Since this is being tracked in #1025, I'm going to merge this PR in its current state. This means that the test will work if all goes well when the CBMC version is upgraded, causing the regression to fail unless the test is renamed @zhassan-aws @tedinski

tedinski: Sounds good!

PR #1259, opened 2022-06-07
Add support to compiletest's cargo-kani mode for tests with multi-level directories

zhassan-aws:
### Description of changes: 

Currently, `compiletest`'s `cargo-kani` mode looks for files of the form `<harness-name>.expected` and executes `cargo kani --harness <harness-name>` for each. However, if it finds a `Cargo.toml` in a directory, it does not recurse in its subdirectories. This makes it unable to handle cargo workspaces, where it's common to have a `Cargo.toml` file at the root directory of the package.

This PR fixes this issue by allowing `compiletest` to continue recursing into subdirectories looking for `*.expected` files, even if it finds a `Cargo.toml` in the parent directory.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

### Checklist

### DIFF
--- a/tools/compiletest/src/main.rs
+++ b/tools/compiletest/src/main.rs
@@ -387,16 +387,24 @@ fn collect_tests_from_dir(
     // output directory corresponding to each to avoid race conditions during
     // the testing phase. We immediately return after adding the tests to avoid
     // treating `*.rs` files as tests.
-    if config.mode == Mode::CargoKani && dir.join("Cargo.toml").exists() {
+    if config.mode == Mode::CargoKani {
+        let has_cargo_toml = dir.join("Cargo.toml").exists();
         for file in fs::read_dir(dir)? {
-            let file_path = file?.path();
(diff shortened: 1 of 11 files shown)


tools/compiletest/src/main.rs
```diff
@@ -387,16 +387,24 @@ fn collect_tests_from_dir(
     // output directory corresponding to each to avoid race conditions during
     // the testing phase. We immediately return after adding the tests to avoid
     // treating `*.rs` files as tests.
-    if config.mode == Mode::CargoKani && dir.join("Cargo.toml").exists() {
+    if config.mode == Mode::CargoKani {
```
tedinski, line 390: This function has become rather awkward. I think I'm not opposed to this, but perhaps we should break this function into two? `collect_expected_tests_from_dir` and `collect_rs_tests_from_dir`?

zhassan-aws: Good idea. Done.

tedinski (approved):

PR #1250, opened 2022-06-03
Audit for `truncf*`

adpaco-aws:
### Description of changes: 

Restores and completes the audit for the `truncf32` (`truncf`) and `truncf64` (`trunc`) intrinsics.

### Resolved issues:

Part of #1163 
Part of #1025 

### Testing:

* How is this change tested? Adds 2 tests.

* Is this a refactor change? No.

### Checklist
- [x] Each commit message has a non-empty body, explaining why the change was made
- [x] Methods or procedures are documented
- [x] Regression or unit tests are included, or existing tests cover the modified code
- [x] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #1262, opened 2022-06-08
Audit for `ceilf*`

adpaco-aws:
### Description of changes: 

Restores and completes the audit for the `ceilf32` (`ceilf`) and `ceilf64` (`ceil`) intrinsics.

### Resolved issues:

Part of #1163 
Part of #1025 

### Testing:

* How is this change tested? Adds 2 tests, removes 2 tests.

### Checklist


adpaco-aws:
Updated the tests to follow a similar style as the ones in #1250 (some concrete cases and two properties). In this case, however, the property `0 <= |x - ceil(x)| < 1` does not hold for all cases (very small numbers cause it, and the behavior is reproducible in Rust), so we need to test `0 <= |x - ceil(x)| <= 1`.

tedinski (approved):

PR #1268, opened 2022-06-09
docs: fix typos

sanjit-bhat:
### Description of changes: 

This PR fixes typos in the documentation.

### Call-outs:

Below are the only typos that are more than spelling fixes:
1. The change in `tutorial/arbitrary-variables/src/inventory.rs` to the commented out assertion seems to be what was originally intended. It matches the assert in the `safe_update` function.

### Testing:

* How is this change tested?
No testing required for doc typos.

* Is this a refactor change?
No

### Checklist


docs/src/undefined-behaviour.md
```diff
@@ -44,6 +44,4 @@ Kani makes a best-effort attempt to detect some cases of UB:
     * Kani can detect invalid dereferences, but may not detect them in [place expression context](https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions).
 * Invoking undefined behavior via compiler intrinsics.
     * See [current support for compiler intrinsics](./rust-feature-support/intrinsics.md).
-* Producing an invalid value, even in private fields and locals. 
-    * Kani provides a [mechanism](./tutorial-nondeterministic-variables.md#safe-nondeterministic-variables-for-custom-types) `is_valid()` which users can use to check validity of objects, but it does not currently apply to all types.
```
adpaco-aws, line 48: This item is repeated because it's a "class of UB that Kani does not detect" in general, but at the same time it "makes a best-effort to detect some cases" of it. So it's being repeated in different contexts and we should keep it as is.

docs/src/tutorial/arbitrary-variables/src/inventory.rs
```diff
@@ -54,7 +54,7 @@ mod verification {
         let quantity: NonZeroU32 = unsafe { kani::any_raw() };
 
         // The assert bellow would fail if we comment it out.
-        // assert!(id.get() != 0, "NonZeroU32 is internally a u32 but it should never be 0.");
+        // assert!(quantity.get() != 0, "NonZeroU32 is internally a u32 but it should never be 0.");
```
adpaco-aws, line 57: Nice catch!

adpaco-aws (approved):
Thanks @sanjit-bhat ! Please undo the change in `undefined-behaviour.md` (see reason below), the other changes look good to me :)

docs/src/rust-feature-support/unstable.md
```diff
@@ -3,7 +3,7 @@
 In general, unstable Rust features are out of scope and any support
 for them available in Kani should be considered unstable as well.
 
-The following are exampels of unstable features that are not supported
+The following are examples of unstable features that are not supported
```
jaisnan, line 6: Thanks for catching this!

docs/src/undefined-behaviour.md
sanjit-bhat: Ah, good catch! I overlooked the different contexts.

Issue #493, opened 2021-09-16
Kani produces no output on non-termination


tedinski:
When running RMC on code with loops and without an `--unwind` bound, RMC will often not terminate.

But, it presently produces no output at all during this infinite loop, whereas CBMC produces continuous stream of logging data that helps indicate what's gone wrong.

I believe this is because at some point we're buffering stdout from cbmc and only printing it once it's finished. We should allow these log messages to stream to the terminal, so we can see them during a non-termination problem.

adpaco-aws:
Note that this issue may also appear in examples without loops. Running with `--unwind 0` will allow the example to terminate. The dashboard has had problems with this (see #482 and its fix #492) and I would expect some more.

We could also consider running `--unwind 0` by default. Thoughts?

tedinski:
That's an interesting idea. The current default was presumably for soundness, but now we have unwinding assertions, right?

adpaco-aws:
Yes, unwinding assertions are on by default so in principle passing `--unwind 0` (or higher) will assert if it is not enough.

jaisnan:
We're planning on addressing this issue by - 

1. Piping the output instead of writing it to a file within the rust code.
2. Add functionality to read from stream in the Python Post processing.

PR #1278, opened 2022-06-14
Add live streaming JSON output from CBMC to prevent Kani output hang

jaisnan:
### Description of changes: 

CBMC currently reads the input from CBMC , waits for termination and writes the output to a file which goes to the parsing script. This runs into an issue when CBMC does not seemingly terminate, which makes Kani appear to be hung.

This PR changes that to show the JSON output live from CBMC which will later be consumed by the parsing to show the post processed Kani output.

For example, for the rust file - 

``` rust
fn rev(mut x: [i32; 4]) -> [i32; 4] {
    x.reverse();
    x
}

#[cfg(kani)]
#[kani::proof]
fn example() {
    let var = kani::any();
    let reversed = rev(var);
    let double_reversed = rev(reversed);

    assert_eq!(var, double_reversed);
}
```

Before -

Screen Shot 2022-06-13 at 11 46 42 PM

After - 

Screen Shot 2022-06-13 at 11 49 32 PM

As we can , there is a long unwinding loop that's not being displayed to the user which can cause some confusion.
With this change, the user can be informed about the status of the verification live.

### Resolved issues:

Related #493 

### Call-outs:

1. This Hides the live output behind a flag `--use-piped-output` for now.
2. This PR only streams the JSON output from CBMC. An incoming PR will convert the live JSON into the regular Kani output.

### Testing:

* How is this change tested?
Hidden behind a flag to be as non-invasive as possible.

* Is this a refactor change?
Yes

### Checklist


scripts/cbmc_json_parser.py
```diff
@@ -83,22 +88,47 @@ def main(argv):
...
         usage_error(f"Invalid output format '{argv[2]}'.")
 
+    # Check the mode of the output
+    mode = os.fstat(sys.stdin.fileno()).st_mode
     extra_ptr_check = False
+
+    # The cbmc output can be either file mode (default) or stream mode (--read-cbmc-from-stream)
+    cbmc_output_mode = argv[1].split(".")[-1]
+    if not ((cbmc_output_mode == "cbmc_output") or (cbmc_output_mode == GlobalMessages.READ_FROM_STREAM)):
+        usage_error(f"CBMC output file or mode invalid\n")
```
adpaco-aws, line 98:
```suggestion
        usage_error(f"CBMC output mode invalid\n")
```

scripts/cbmc_json_parser.py
```diff
@@ -83,22 +88,47 @@ def main(argv):
...
+        # Check if the stdin input comes from the pipe
+        if S_ISFIFO(mode):
+            try:
+                for line in sys.stdin:
+                    print(line)
+            except BaseException:
+                usage_error(f"Unable to complete parsing CBMC Output\n")
+        else:
+            usage_error(f"CBMC Output not piped to post-processing\n")
+        return_code = 0
```
adpaco-aws, line 120: I would use an error code here.

scripts/cbmc_json_parser.py
```diff
@@ -83,22 +88,47 @@ def main(argv):
     if not output_style:
         usage_error(f"Invalid output format '{argv[2]}'.")
 
+    # Check the mode of the output
+    mode = os.fstat(sys.stdin.fileno()).st_mode
     extra_ptr_check = False
+
+    # The cbmc output can be either file mode (default) or stream mode (--read-cbmc-from-stream)
+    cbmc_output_mode = argv[1].split(".")[-1]
+    if not ((cbmc_output_mode == "cbmc_output") or (cbmc_output_mode == GlobalMessages.READ_FROM_STREAM)):
```
adpaco-aws, line 97: Maybe define another variable READ_FROM_FILE?

adpaco-aws (commented):
You should be able to define `expected` tests for this, no?

scripts/cbmc_json_parser.py
jaisnan: I've added a custom exit string and an error code of 1 with - 

```
print(f"Unable to complete parsing CBMC Output\n")
sys.exit(1)
```

Would this work?

adpaco-aws: Sorry, my comment here was not very clear.

If I'm not mistaken, this will print all lines from CBMC's output and return 0. If we change this to 1, do the regression tests still pass?

jaisnan:
> You should be able to define `expected` tests for this, no?

I was able to add in a few tests and refactored the code as suggested. Let me know if they're as expected!

scripts/cbmc_json_parser.py
jaisnan: I should have clarified more. The return code is 1 if the stream of output is interrupted , either by a broken line or broken pipe or user interrupt (Ex - keyboard interrupt). If it completes printing all the lines from CBMC, then it returns 0. 

I hope this is an acceptable usage of the error code?

adpaco-aws: A return code with value 0 basically means that verification is always successful.

adpaco-aws:
Closed in favor of #1433

PR #1280, opened 2022-06-14
Bump to 0.4.0

tedinski:
### Description of changes: 

What it says on the tin

### Call-outs:

* Noticed we hadn't been bumping `cprover_bindings`. Think that was an oversight. Not important though, its version isn't published anywhere.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/cprover_bindings/Cargo.toml
+++ b/cprover_bindings/Cargo.toml
@@ -3,7 +3,7 @@
 
 [package]
 name = "cprover_bindings"
-version = "0.0.0"
+version = "0.4.0"
 edition = "2018"
 license = "MIT OR Apache-2.0"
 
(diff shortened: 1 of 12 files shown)


cprover_bindings/Cargo.toml
```diff
@@ -3,7 +3,7 @@
 
 [package]
 name = "cprover_bindings"
-version = "0.0.0"
+version = "0.4.0"
 edition = "2018"
```
adpaco-aws, line 7: Do we want to upgrade to Rust 2021 here? Same goes for library packages.

tedinski: We could look into that, yeah. Not going to do it for this PR though.

PR #1281, opened 2022-06-14
Switch to using cargo-deny

### LINKED ISSUE #1237: Carefully deprecate scancode?
Our release runbook currently includes running `scancode` but this is a slow and manual process.

We should:

1. Carefully review a scancode run to see if it reports anything beyond what we get with `cargo license`.
2. If not, document this fact and go with `cargo license`
3. Remove it from the runbook.
4. Eventually look into using `cargo-deny` to automate the license checking task.

tedinski:
### Description of changes: 

1. Switch to using `cargo-deny` instead of `cargo-audit`. It can do the same role and more.
2. Add an explicit allow-list for licenses we use in our Rust code.
3. Add 2 crates to our license note file.
4. Organize and clean up the license note file a bit.
5. Add `license` field to crates in our repo.

### Resolved issues:

Resolves #1237 

### Call-outs:

* This rather nicely allows us to audit licenses for our whole workspace, so we can remove `cargo-license` from our release runbook.
* I'm going to separately write up a doc about scancode, as I think with this we can remove that from our regular release runbook as well.

### Testing:

### Checklist

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,9 +1,8 @@
 # Copyright Kani Contributors
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
-# A temporary, non-required workflow to notify us when cargo-audit has things to say.
-# TODO: Remove this, and make cargo-audit part of release automation (when we have it) instead.
-# And maybe also figure out why dependabot wasn't notifying us. :(
+# 1. Checks licenses for allowed license.
+# 2. Checks Rust-Sec registry for security advisories.
 
 name: Cargo Audit
 on: [pull_request]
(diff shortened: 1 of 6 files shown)


.github/workflows/audit.yml
```diff
@@ -13,6 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions-rs/audit-check@v1
+      - uses: EmbarkStudios/cargo-deny-action@v1
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
```
danielsn, line 18: Nice to remove this. Why was it needed before?

danielsn (approved):

.github/workflows/audit.yml
tedinski: Not sure. Don't quote me on this, but I think it was just the old style of passing the token down, and github added a new implicit style for actions, that doesn't require this boilerplate. So it's not actually a real improvement. But it's a read-only token anyway here, so it doesn't matter.

PR #1291, opened 2022-06-20
Refactor machine_model to make-it non-oop

### LINKED ISSUE #1289: Refactor machine model
The cprover_bindings machine model is written in a bit of "oop" style, and could be cleaned up by just making all its field pub:

https://github.com/model-checking/kani/blob/main/cprover_bindings/src/machine_model.rs

Discovered because clippy complains about the number of arguments to `new` in #1287

giltho:
### Description of changes: 

No change in behavior, simple refactoring

### Resolved issues:

Resolves #1289 

### Call-outs:

Nothing that I'm aware of

### Testing:

* How is this change tested? It's a refactoring, all previous tests apply

### Checklist


kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
```diff
@@ -390,9 +390,9 @@ fn machine_model_from_session(sess: &Session) -> MachineModel {
     let word_size = 32;
     let rounding_mode = RoundingMode::ToNearest;
 
-    MachineModel::new(
+    MachineModel {
+        architecture: architecture.to_string(),
```
tedinski, line 394: Why the reorder these? (Also, clone instead of to_string? Same thing ofc, but maybe clearer.)

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -439,7 +439,7 @@ impl<'tcx> GotocCtx<'tcx> {
 
     fn codegen_allocation_data(&mut self, alloc: &'tcx Allocation) -> Vec<AllocData<'tcx>> {
         let mut alloc_vals = Vec::with_capacity(alloc.relocations().len() + 1);
-        let pointer_size = self.symbol_table.machine_model().pointer_width() as usize / 8;
+        let pointer_size = self.symbol_table.machine_model().pointer_width as usize / 8;
```
tedinski, line 442: Looks like this isn't used anywhere else, but let's actually add an impl back to MachineModel for `pointer_width_in_bytes` instead of doing that here?

cprover_bindings/src/machine_model.rs
```diff
@@ -8,27 +8,27 @@ pub struct MachineModel {
...
-    null_is_zero: bool,
-    pointer_width: u64,
-    rounding_mode: RoundingMode,
-    short_int_width: u64,
-    single_width: u64,
-    wchar_t_is_unsigned: bool,
-    wchar_t_width: u64,
-    word_size: u64,
+    pub alignment: u64,
+    pub architecture: String,
```
danielsn, line 12: Not that it really matters for memory usage, but should we used `InternedString` here?

tedinski: _Probably_ not, I don't think we see duplicates of this structure in memory.

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
giltho: Old OCaml/JS habit of putting all the non-shorthand fields first, and then the short-hand fields. So when reading the assignment, you see what's not trivial first, in case your eyes naturally stop reading the bunch of shorthand assignments that follow

Also, here, to_string doesn't work, because it's already a string. `mm.architecture()` used to return a reference to the string, but now I'm accessing the string itself directly

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
giltho: doing that!

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
tedinski: Ah, I see. Makes sense.

tedinski (approved):

PR #1294, opened 2022-06-21
Add doc section on debugging Kani in VS code

fzaiser:
### Description of changes: 

This adds documentation on how to use the VS code debugger on Kani.

### Call-outs:

It was not able to get to launch the executable at the same time as the debugger, hence I settled for adding the code to Kani, which calls the VS code debugger.

### Testing:

* How is this change tested? Manual test

### Checklist

### DIFF
--- a/docs/src/rustc-hacks.md
+++ b/docs/src/rustc-hacks.md
@@ -33,6 +33,25 @@ You may also need to install the `rustc-dev` package using rustup
 rustup toolchain install nightly --component rustc-dev
 ```
 
+#### Debugging in VS code
+
+To debug Kani in VS code, first install the [CodeLLDB extension](https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb).
+Then add the following lines at the start of the `main` function (see [the CodeLLDB manual](https://github.com/vadimcn/vscode-lldb/blob/master/MANUAL.md#attaching-debugger-to-the-current-process-rust) for details):
+
+```rust
+{
+    let url = format!(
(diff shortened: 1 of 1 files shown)


docs/src/rustc-hacks.md
```diff
@@ -33,6 +33,25 @@ You may also need to install the `rustc-dev` package using rustup
 rustup toolchain install nightly --component rustc-dev
 ```
 
+#### Debugging in VS code
+
+To debug Kani in VS code, first install the [CodeLLDB extension](https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb).
+Then add the following lines at the top of the function `codegen_crate` in `kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs` (see [the CodeLLDB manual](https://github.com/vadimcn/vscode-lldb/blob/master/MANUAL.md#attaching-debugger-to-the-current-process-rust) for details):
```
celinval, line 39: Does it mean that the debugger only works past this point? What happens if you add to the main function?

fzaiser: Good point, I wasn't sure about the interaction with the driver, but you're right, adding at at the start of the main function is better. (I checked that it works too.)

celinval: we have two binaries, the driver and the compiler. It will depend on which process you want to debug. I suggest leaving the doc generic.

fzaiser: I've changed it in my latest push to just say "start of the main function".

celinval (approved):
Thanks

PR #1295, opened 2022-06-21
Add script to replace tags in GotoC output with type names

fzaiser:
### Description of changes: 

This adds a Python script to replace the tags in the GotoC output file `file.out.c` with the Rust type names from `file.type_map.json` and writes the result to `file.demangled.c`.
The resulting "demangled" file is much easier to read and helps with debugging.
I also added documentation in the command cheat sheets.

### Call-outs:

Do we need to update the docs anywhere else? I think this was the only place that GotoC output was mentioned.

### Testing:

* How is this change tested? Manually. Since this is only a convenience script for developers, no automated testing seems necessary.

### Checklist

### DIFF
--- a/docs/src/cheat-sheets.md
+++ b/docs/src/cheat-sheets.md
@@ -72,7 +72,12 @@ kani --keep-temps file.rs
 # Generate "C code" from CBMC IR (.c)
 kani --gen-c file.rs
 ```
-
+```bash
+# "Demangle" the tags in the "C code" (file.out.c) with the Rust type names (file.type_map.json).
+# The result is written to file.out.demangled.c
+# This makes reading the generated C code much easier.
+demangle.py file
+```
 ## CBMC
 
 ```bash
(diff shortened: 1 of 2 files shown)


tedinski (commented):
I wonder if we should just do this unconditionally for `--gen-c` in `kani-driver`

fzaiser:
I had considered doing the replacement in the Kani code in Rust, but according to @danielsn, we cannot do this directly because it is CBMC output, so he suggested a Python script instead.

I guess we could call this script from kani-driver, but it may also be useful to have access to the original CBMC output. I'm not sure what is best here.

tedinski:
Well, my suggestion would have been writing it in Rust instead of python, too. And yeah, as part of kani-driver.

I'm not objecting to things as they are though. You can decide if it'd be a good use of your time.

fzaiser:
I'll take a look at the `kani-driver` code to see how involved it would be to do this in Kani directly.

fzaiser:
I implemented this in Kani directly (not just for types), which is the better approach: #1303

PR #1298, opened 2022-06-22
Add support for installing on NixOS

### LINKED ISSUE #1179: cargo-kani setup support for NixOs
A customer on NixOS reports the following command will make Kani releases work:

```
patchelf --set-interpreter "$(cat $NIX_CC/nix-support/dynamic-linker)" ~/.kani/kani-0.1.0/bin/*
```

1. Find a reliable/correct way to detect NixOS.
2. Ensure the above command is the best option for doing this.
3. Add this to `cargo-kani setup` so it happens automatically on that platform.

tedinski:
### Description of changes: 

1. Add a setup hack that detects installation on NixOS and applies the `patchelf` hacks necessary to make the downloaded binaries work on that platform.
2. I added 3 new Docker files for testing releases, none of which I've added to CI at this time. These are: NixOS, Ubuntu 22.04, Amazon Linux 2.

### Resolved issues:

Resolves #1179 

### Testing:

* How is this change tested? With the included docker file.

### Checklist


scripts/ci/Dockerfile.bundle-test-22-04
```diff
@@ -0,0 +1,20 @@
+# Copyright Kani Contributors
```
celinval, line 1: nit: Rename file to include ubuntu-22-04

src/setup.rs
```diff
@@ -114,7 +117,7 @@ fn setup_python_deps(kani_dir: &Path, os: &os_info::Info) -> Result<()> {
     if os.os_type() == os_info::Type::Ubuntu
         && *os.version() == os_info::Version::Semantic(18, 4, 0)
     {
-        crate::os_hacks::setup_python_deps_on_ubuntu_18_04(&pyroot, pkg_versions)?;
+        os_hacks::setup_python_deps_on_ubuntu_18_04(&pyroot, pkg_versions)?;
```
celinval, line 120: Can you move this to be called inside `setup_os_hacks()` instead?

src/os_hacks.rs
```diff
@@ -45,3 +46,78 @@ pub fn setup_python_deps_on_ubuntu_18_04(pyroot: &Path, pkg_versions: &[&str]) -
...
+                return setup_nixos_patchelf(kani_dir);
+            }
+            Ok(())
+        }
+        _ => Ok(()),
+    }
+}
+
+/// On NixOS, the dynamic linker does not live at the standard path, and so our downloaded
+/// pre-built binaries need patching.
```
celinval, line 67: Is it possible to fix that while pre-building them to NixOS instead of patching here or do you mean that there is standard path and it is only available at runtime?

src/os_hacks.rs
```diff
@@ -45,3 +46,78 @@ pub fn setup_python_deps_on_ubuntu_18_04(pyroot: &Path, pkg_versions: &[&str]) -
 
     Ok(())
 }
+
+/// This is the final step of setup, where we look for OSes that require additional setup steps
+/// beyond the usual ones that we have done already.
+pub fn setup_os_hacks(kani_dir: &Path, os: &Info) -> Result<()> {
+    match os.os_type() {
```
celinval, line 53: Can this be done via cargo target_os?

src/os_hacks.rs
```diff
@@ -45,3 +46,78 @@ pub fn setup_python_deps_on_ubuntu_18_04(pyroot: &Path, pkg_versions: &[&str]) -
...
+        }
+        _ => Ok(()),
+    }
+}
+
+/// On NixOS, the dynamic linker does not live at the standard path, and so our downloaded
+/// pre-built binaries need patching.
+/// In addition, the C++ standard library (needed by the CBMC binaries we ship) also does not
+/// have a standard path, and so we need to inject an rpath into those binaries to get them
+/// to successfully link at runtime.
```
celinval, line 70: I'm assuming LD_LIBRARY_PATH didn't do the trick, right?

celinval (commented):
Thanks for figuring this out. Are you planning to add tests for these platforms in our regression as a follow up PR?

scripts/ci/Dockerfile.bundle-test-22-04
tedinski: Do you mean adding "ubuntu" to all 3 file names here?

src/os_hacks.rs
tedinski: We don't pre-build anything specifically for NixOS (well, except that the `kani-verifier` is downloaded and built locally), so this is taking the generic linux binaries and making them link properly on NixOS.

There are no standard paths, it has to be discovered at runtime. Nix seems to hack things into sorta working by injecting flags so that the linker would add rpaths to binaries. This basically replicates the effect for downloaded binaries built elsewhere.

scripts/ci/Dockerfile.bundle-test-nixos
```diff
@@ -0,0 +1,34 @@
...
+RUN rustup toolchain add stable
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+WORKDIR /tmp/kani
+COPY ./tests ./tests
+COPY ./kani-latest-x86_64-unknown-linux-gnu.tar.gz ./
+# Very awkward glob (not regex!) to get `kani-verifier-*` and not `kani-verifier-*.crate`
+COPY ./target/package/kani-verifier-*[^e] ./kani-verifier
+RUN cargo install --path ./kani-verifier
+RUN cargo-kani setup --use-local-bundle ./kani-latest-x86_64-unknown-linux-gnu.tar.gz
```
camshaft, line 34: Do you plan on running a quick smoke test to make sure everything was correctly patched as well?

src/os_hacks.rs
camshaft: yeah `patchelf` is nice since it makes it "just work" every time you invoke the command, rather than having to remember to correctly set up the LD_LIBRARY_PATH

scripts/ci/Dockerfile.bundle-test-nixos
camshaft: sounds good

src/setup.rs
celinval: Oh. I would expect an `else` case then. From reading this code it does look like do this extra thing for ubuntu 18.4.0.

src/os_hacks.rs
celinval: If the path never changes, I agree that this is a more permanent solution, although a bit invasive.

src/os_hacks.rs
celinval: Oh, that's great. :(

src/os_hacks.rs
camshaft: cargo-bolero was actually recently added (not even by me which is cool) so it's pretty easy to get packages in there: https://github.com/NixOs/nixpkgs/blob/master/pkgs/development/tools/rust/cargo-bolero/default.nix

scripts/ci/Dockerfile.bundle-test-22-04
celinval: Yes, I mean adding Ubuntu name to this new file as well as the 18-04 and the 20-04. That seems like the naming convention you are following `Dockerfile.bundle-test-<os>`, right?

src/os_hacks.rs
celinval: That's cool!

In the meantime, if the path changes, we should probably use LD_LIBRARY_PATH or re-patch the binaries if the value has changed. We don't expect users to invoke `kani-driver` or `kani-compiler` directly, so we might as well ensure that they always work.

celinval: I could be wrong, but the annoying part is that our users will have to re-install `kani-verifier` crate, but that won't fix the problem since the bundle we downloaded will still be on the disk. So when they run `kani`, they will get another crash now because kani-driver is broken, right? At that point, they will have to run `kani setup` to force everything to be reinstalled.

Not a blocker since nothing works today, but we should at least document that.

camshaft: Yep that's correct. But what's here is definitely an improvement. Like I said, I think for the best user experience is to get into nixpkgs.

celinval (approved):
As we talked offline, please add a CI job to run NixOS at least as part of a push to main. Thanks!

celinval (approved):
Thanks for adding CI coverage!

PR #1299, opened 2022-06-22
Clean up property class and reclassify some checks

celinval:
### Description of changes: 

1. Remove the property classes that weren't being used.
2. Provide a tighter definition for DefaultAssert and reclassify things
   accordingly.
3. Renamed KaniCheck to SafetyCheck to make it a more meaningful name.

### Resolved issues:

N/A

### Call-outs:

This PR adds new dependencies to these two crates (both with MIT license):
- https://crates.io/crates/strum
- https://crates.io/crates/strum_macros

### Testing:

* How is this change tested? Regular tests

### Checklist

*Edit: Added info to "Call-outs"*


kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -10,17 +10,19 @@ use cbmc::goto_program::{Expr, Location, Stmt};
 #[allow(dead_code)]
 pub enum PropertyClass {
     ArithmeticOverflow,
-    AssertFalse,
     Assume,
     Cover,
-    CustomProperty(String),
```
jaisnan, line 16: Do we want to keep one property class that can be used as a "+" set member, to allow for edge case handling? I think `CustomProperty` could be useful for hotfixes which can be refactored later on. Not a blocker for merge though.

jaisnan (approved):
Thank you for the cleanup. Just had one suggestion/question about the `CustomProperty` property class. Not a blocker and the rest looks good to me.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
celinval: I see your point. Do you have any use case in mind? I don't think fromStr() is used today.

My reasoning to remove this was:

1. Things could silently fail if we try to convert a string to a property class that doesn't exist.
2. This can be confusing for people adding properties in the future. If they want a new property class, should they use the `CustomProperty()` or add a new enum variant.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -52,20 +51,18 @@ impl PropertyClass {
...
             "finite_check" => PropertyClass::FiniteCheck,
-            "kani_check" => PropertyClass::KaniCheck,
-            "pointer_offset" => PropertyClass::PointerOffset,
+            "kani_check" => PropertyClass::SafetyCheck,
             "sanity_check" => PropertyClass::SanityCheck,
             "unimplemented" => PropertyClass::Unimplemented,
             "unreachable" => PropertyClass::Unreachable,
             "unsupported_construct" => PropertyClass::UnsupportedConstruct,
-            _ => PropertyClass::CustomProperty(input.to_owned()),
+            _ => unreachable!("Invalid property class {}", input),
```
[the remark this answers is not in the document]
celinval: I've spent a few cycles on this. Serde derive includes an option to choose the string format, but I couldn't find a serializer that would return just the string with the enum's name.

So we would need to extract that from the serialized value, create our own serializer or maybe look a bit further.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -32,16 +34,13 @@ impl PropertyClass {
...
             PropertyClass::Assume => "assume",
             PropertyClass::Cover => "coverage_check",
-            PropertyClass::CustomProperty(property_string) => property_string.as_str(),
             PropertyClass::DefaultAssertion => "assertion",
             PropertyClass::ExactDiv => "exact_div",
             PropertyClass::ExpectFail => "expect_fail",
             PropertyClass::FiniteCheck => "finite_check",
-            PropertyClass::KaniCheck => "kani_check",
-            PropertyClass::PointerOffset => "pointer_offset",
+            PropertyClass::SafetyCheck => "kani_check",
```
celinval, line 43: Note to myself: rename the string constant!

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
jaisnan: Ah, I see your point. Thanks for the clarification!

danielsn (approved):

PR #1300, opened 2022-06-22
Add cache to codegen_ty

celinval:
### Description of changes: 

Avoid codegening the same types over and over again by adding a simple cache.

### Resolved issues:

N/A

### Testing:

### Checklist


danielsn:
Do we need this? Doesn't the `symbol_table` already cache?

celinval:
> Do we need this? Doesn't the `symbol_table` already cache?

This is more like a nice to have. You are correct that the  type codegen doesn't rerun fully because of the `symbol_table` lookup, but for types like pointers and functions, the codegen function still have considerable logic running until it reaches a `symbol_table` lookup.

celinval:
I'm going to close this for now since this there is no performance impact just a small improvement to code debug. We can revisit this later if we feel like it is worth it.

PR #1303, opened 2022-06-24
Demangle symbol names in GotoC (-> .demangled.c)

fzaiser:
### Description of changes: 

If you run Kani with `--gen-c`, it generates a GotoC file that is consumed by CBMC.
The names (function names, type names etc.) in this file are mangled, which makes it hard to read.
This PR also creates a demangled file (with the extension `.demangled.c`) with the `prettyName`s from the symbol table.

### Call-outs:

Sometimes, the `prettyName`s in the symbol table seem to be missing, even if they are available in the `.type_map.json`.
Rather than also parsing the type map, I think it would be better to ensure that the pretty names for types are consistent between the symbol table and the type map in a follow-up PR.

### Testing:

* How is this change tested? I added a single-file and a multi-file test under `tests/output-files`.

### Checklist


kani-driver/src/call_goto_instrument.rs
```diff
@@ -116,19 +131,45 @@ impl KaniSession {
...
+        symtab_files: &[impl AsRef<Path>],
+        c_file: &Path,
+        demangled_file: &Path,
+    ) -> Result<()> {
+        let c_code = std::fs::read_to_string(c_file)?;
+        let mut demangled_code = c_code.clone();
+        for symtab_file in symtab_files {
+            let reader = BufReader::new(File::open(symtab_file.as_ref())?);
+            let symtab: serde_json::Value = serde_json::from_reader(reader)?;
+            if let Some(symtab) = symtab.get("symbolTable").and_then(|s| s.as_object()) {
```
danielsn, line 156: When would this fail?

kani-driver/src/call_goto_instrument.rs
```diff
@@ -116,19 +131,45 @@ impl KaniSession {
...
             "--dump-c".into(),
             file.to_owned().into_os_string(),
-            output_filename.into_os_string(),
+            output_file.to_owned().into_os_string(),
         ];
 
         self.call_goto_instrument(args)
     }
 
+    /// Generate a .demangled.c file from the .c file using the `prettyName`s from the symbol tables
```
danielsn, line 144: As I understand it, this is only partially complete. A comment about what is and isn't demangled would be useful here

kani-driver/src/call_goto_instrument.rs
```diff
@@ -116,19 +131,45 @@ impl KaniSession {
...
+    pub fn demangle_c(
+        &self,
+        symtab_files: &[impl AsRef<Path>],
+        c_file: &Path,
+        demangled_file: &Path,
+    ) -> Result<()> {
+        let c_code = std::fs::read_to_string(c_file)?;
+        let mut demangled_code = c_code.clone();
```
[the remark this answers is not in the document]
fzaiser: Good catch, that was a leftover from a previous iteration.

kani-driver/src/call_goto_instrument.rs
```diff
@@ -116,19 +131,45 @@ impl KaniSession {
...
+        for symtab_file in symtab_files {
+            let reader = BufReader::new(File::open(symtab_file.as_ref())?);
+            let symtab: serde_json::Value = serde_json::from_reader(reader)?;
+            if let Some(symtab) = symtab.get("symbolTable").and_then(|s| s.as_object()) {
+                for (_, symbol) in symtab {
+                    if let Some(serde_json::Value::String(name)) = symbol.get("name") {
+                        if let Some(serde_json::Value::String(pretty)) = symbol.get("prettyName") {
+                            let name = name.strip_prefix("tag-").unwrap_or(name);
```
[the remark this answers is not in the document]
fzaiser: Struct names start with `tag-` but in the GotoC file, the "tag-" prefix is not there, so we want to get rid of it. On the other hand, not all names start with "tag-", so we only strip it if necessary, and otherwise leave it (-> unwrap_or).

kani-driver/src/call_goto_instrument.rs
danielsn: I think that makes sense: if data we generate has an unexpected format, fail early

kani-driver/src/call_goto_instrument.rs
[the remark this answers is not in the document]
danielsn: Comment explaining that

tests/output-files/check-output.sh
```diff
@@ -0,0 +1,81 @@
...
+  TARGET="x86_64-unknown-linux-gnu"
+elif [[ $PLATFORM == "Darwin i386" ]]
+then
+  TARGET="x86_64-apple-darwin"
+elif [[ $PLATFORM == "Darwin arm" ]]
+then
+  TARGET="aarch64-apple-darwin"
+else
+  echo
+  echo "Test only works on Linux or OSX x86 platforms, skipping..."
```
danielsn, line 24: But you check for arm above?

kani-driver/src/call_goto_instrument.rs
fzaiser: I wasn't sure whether the demangling should be able to stop the whole codegen. But I've changed it to `unwrap` now.

danielsn (approved):

PR #1309, opened 2022-06-27
Further prune dead code from our fork of librustdoc

tedinski:
### Description of changes: 

librustdoc can have deep dependencies into rustc and makes it harder to do rust toolchain updates with Kani. We only use this for bookrunner.

This removes a bunch of dead code we don't use, including a lot of references to `rustc_` structures. This should reduce how hard it is to sync this code in the future.

### Call-outs:

* Notably, as a result of removing a lot of dead code, I was also able to remove a lot of cargo dependencies from the crate. Woo!

### Testing:

* How is this change tested? Bookrunner runs

* Is this a refactor change? yes!

### Checklist


celinval (approved):
Awesome! Thanks

PR #1310, opened 2022-06-28
Produce more pretty names

### LINKED ISSUE #1306: Pretty names missing for some types in the symbol table (but present in type map)
While working on #1303, I noticed that the symbol table does not always contain a pretty name for types, but the corresponding pretty name is in the type map. The symbol table should be fixed to contain the information from the type map in all cases. The reason for the current behavior is the use of `NO_PRETTY_NAME` in a lot of the code.

Example: the `Unit` symbol does not have a pretty name in the symbol table, but the type map contains `()`. The symbol table should have `()` as the pretty name for the type as well.

fzaiser:
### Description of changes: 

As described in #1306, the symbol table often contains no pretty names for types, making the GotoC output harder to read than necessary. This PR changes the type of `pretty_name` from an `Option` to an `InternedString` for many functions, thus making it harder to forget to provide a better name.

In addition, the `readable_instance_name` function now includes the monomorphized type arguments. This solves the issue of many functions with the same name (because they're different monomorphizations).

If you run `kani` with `--enable-unstable --gen-c` and look at the `.demangled.c` output file, you can clearly see the difference:

<details>
<summary>Before</summary>

```c
struct _5683571674675523038::Variant1
{
  // 0
  _Bool 0;
};

struct TestEnum
{
  // Variant1
  struct _5683571674675523038::Variant1 Variant1;
};

struct Unit main(void)
{
  struct Unit var_0;
  struct PrettyStruct var_1;
  struct PrettyStruct var_2;
  struct Unit var_3;
  struct Unit var_4;
  struct [_11046744883169582909; 2] x;
  struct refstr *ref_to_str;
  struct TestEnum test_enum;
  struct refstr *var_8;

bb0:
  ;
  var_1=pretty_function(var_2);

bb1:
  ;
  monomorphize();

bb2:
  ;
  monomorphize();

bb3:
  ;
  x=gen-repeat<_11417795057624843338>(1);
  var_8 = byte_extract_little_endian((struct refstr *)((unsigned char *)&singlefile_92ff0dd2$$singlefile$$alloc14 + 0), 0, struct refstr *);
  ref_to_str = var_8;
  test_enum = nondet_0();
  test_enum.Variant1.0 = 1;
  return VoidUnit;
}
```

</details>

<details>
<summary>After</summary>

```c
struct TestEnum::Variant1
{
  // 0
  _Bool 0;
};

struct TestEnum
{
  // Variant1
  struct TestEnum::Variant1 Variant1;
};

struct () main(void)
{
  struct () var_0;
  struct PrettyStruct var_1;
  struct PrettyStruct var_2;
  struct () var_3;
  struct () var_4;
  struct [bool; 2] x;
  struct &str *ref_to_str;
  struct TestEnum test_enum;
  struct &str *var_8;

bb0:
  ;
  var_1=pretty_function(var_2);

bb1:
  ;
  monomorphize::<()>();

bb2:
  ;
  monomorphize::<usize>();

bb3:
  ;
  x=init_array_repeat<[bool; 2]>(1);
  var_8 = byte_extract_little_endian((struct &str *)((unsigned char *)&singlefile_92ff0dd2$$singlefile$$alloc14 + 0), 0, struct &str *);
  ref_to_str = var_8;
  test_enum = nondet_0();
  test_enum.Variant1.0 = 1;
  return Void();
}
```

</details>

### Call-outs:

If you can think of better pretty names, please let me know.

### Testing:

* How is this change tested? Extended regression test in `tests/output-files`.

* Is this a refactor change? No (but no change in functionality either).

### Checklist


cprover_bindings/src/goto_program/symtab_transformer/identity_transformer.rs
```diff
@@ -247,7 +247,7 @@ mod tests {
 
         let struct_type = Symbol::struct_type(
             "s",
-            None,
+            "s".into(),
```
danielsn, line 250: Why is `.into()` needed?

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -317,6 +316,7 @@ impl<'tcx> GotocCtx<'tcx> {
 
     fn codegen_direct_literal(&mut self, ty: Ty<'tcx>, init: Expr) -> Expr {
         let func_name = format!("gen-{}:direct", self.ty_mangled_name(ty));
+        let pretty_name = format!("init_direct::<{}>", self.ty_pretty_name(ty));
```
danielsn, line 319: How did you come up with this name?

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -841,6 +841,8 @@ impl<'tcx> GotocCtx<'tcx> {
             // for it. Build and insert a function that just calls an unimplemented block
             // to maintain soundness.
             let drop_sym_name = format!("{}_unimplemented", self.symbol_name(drop_instance));
+            let pretty_name =
+                format!("drop_unimplemented<{}>", self.readable_instance_name(drop_instance));
```
danielsn, line 845: Wonder if this would be a better name for the sym_name too?

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -618,9 +618,7 @@ impl<'tcx> GotocCtx<'tcx> {
             ty::FnPtr(sig) => self.codegen_function_sig(*sig).to_pointer(),
             ty::Closure(_, subst) => self.codegen_ty_closure(ty, subst),
             ty::Generator(_, subst, _) => self.codegen_ty_generator(subst),
-            ty::Never => {
-                self.ensure_struct(NEVER_TYPE_EMPTY_STRUCT_NAME, NO_PRETTY_NAME, |_, _| vec![])
-            }
+            ty::Never => self.ensure_struct(NEVER_TYPE_EMPTY_STRUCT_NAME, "!", |_, _| vec![]),
```
danielsn, line 621: Do we want "!" or does that make the C harder to read?

kani-compiler/src/codegen_cprover_gotoc/utils/names.rs
```diff
@@ -56,7 +56,9 @@ impl<'tcx> GotocCtx<'tcx> {
 
     /// A human readable name in Rust for reference, should not be used as a key.
     pub fn readable_instance_name(&self, instance: Instance<'tcx>) -> String {
-        with_no_trimmed_paths!(self.tcx.def_path_str(instance.def_id()))
+        with_no_trimmed_paths!(
+            self.tcx.def_path_str_with_substs(instance.def_id(), instance.substs)
```
danielsn, line 60: Whats the change here?

tests/output-files/check-output.sh
```diff
@@ -48,6 +48,37 @@ then
     echo "Error: demangled file singlefile.out.demangled.c did not contain expected demangled struct and function name."
     exit 1
 fi
+
```
danielsn, line 51: nice test

cprover_bindings/src/goto_program/symtab_transformer/identity_transformer.rs
fzaiser: Because the function expects an `InternedString`.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
fzaiser: I guessed that this was an intiialization function.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
fzaiser: I'll change it (but leaving out the angle brackets).

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: As we discussed, let's go with `!` (and `()` etc.)

kani-compiler/src/codegen_cprover_gotoc/utils/names.rs
fzaiser: the `_with_substs` part: it outputs `foo::<ty_args>` instead of just `foo`.

fzaiser:
This is blocked on #1354 (I changed the `compiletest` code to be able to run the tests on CI), but it should not be merged with these changes.

tests/output-files/check-output.sh
```diff
@@ -48,6 +48,37 @@ then
     echo "Error: demangled file singlefile.out.demangled.c did not contain expected demangled struct and function name."
```
celinval, line 48: Can you simplify this test by creating a list of patterns + what is testing? Then iterating over the list and printing the error if it couldn't find the pattern?

I think it will make this less repetitive and easy to debug / maintain.

celinval (approved):
Thanks for doing this! Just one minor comment about the tests.

danielsn (approved):

tests/output-files/check-output.sh
fzaiser: Doing this in bash is a pain. Would it make sense to have a Python script for this? I heard that we want to reduce our dependencies on Python.

celinval: BTW, we want to reduce our runtime dependencies on Python. IMO, for test purpose, I would prefer python over bash.

celinval: BTW, it shouldn't be that painful to do this in bash either. That's what we do in the regression script.

https://github.com/model-checking/kani/blob/e28256140f0057f9a67cc46e546ebfeae77eacf0/scripts/kani-regression.sh#L44-L65

PR #1319, opened 2022-06-28
Implements `any_vec::<LENGTH, T>()`

### LINKED ISSUE #1301: Request for API Ideas: Implement Invariant and Any for Vec<T>
Requested feature: @danielsn and I were discussing implementing
`kani::any()` for `Vec<T>`, which I have been hacking as.

``` rust
#[kani::proof]
fn main() {
let tmp_slice: [u8;0xfff] = kani::any();
let vector: Vec<u8> = tmp_slice.to_vec();
assert!(vector.len() >= 0);
}
```

I want feedback on a few proposals for ways to pass bounds like
`0xfff` in the above the arbitrary generator.
- ~~Keep `kani::any(): () -> T` and automatically derive bounds from~~
   `#[kani::unwind(10)]`. When `kani::any(): () -> Vec<T>`, always
   require unwinding bounds.
- public `kani::any_vec` that has explicit bounds.
  - What about `kani::any_vec<N>` where N is a constant bound?
- ~~A macro to do the above hack but automatically.~~

Use case: Arbitrary vectors are popular in proptests. See [tokio](https://github.com/tokio-rs/tokio/blob/b42f21ec3e212ace25331d0c13889a45769e6006/tokio-stream/tests/stream_stream_map.rs#L330).

Test case: The below snippet should be

```rust
#[kani::unwind(10)]
#[kani::proof]
fn main() {
let vector: Vec<u8> = kani::any_vec<N>();
assert!(vector.len() >= 0);
}
```

As of 0.4.0, it does not compile:
``` bash
error[E0277]: the trait bound `std::vec::Vec<u8>: kani::Invariant` is not satisfied
  --> vec-kani.rs:4:23
   |
4  | let vector: Vec<u8> = kani::any();
   |                       ^^^^^^^^^ the trait `kani::Invariant` is not implemented for `std::vec::Vec<u8>`
   |
   = help: the following other types implement trait `kani::Invariant`:
             ()
             [T; N]
             bool
             char
             f32
             f64
             i128
             i16
           and 24 others
```

YoshikiTakashima:
### Description of changes: 

This PR implements a feature to generate an arbitrary `Vec<T>` where T implements Invariant

### Resolved issues:

Resolves #1301

### Call-outs:
- Unfortunately, `rustc` does not allow us to skip the `T` and only
  pass in the constant length. Since the constant is always required,
  we are forced to pass in `::<LEN,_>` even when `T` known to
  `rustc`. Proposals to mitigate this would be welcome.

### Testing:

* How is this change tested?  Tests under `/tests/kani/vec/any`, also
modified `/tests/kani/vec/push.rs` to use `any_vec`.

### Checklist

### DIFF
--- a/library/kani/src/vec.rs
+++ b/library/kani/src/vec.rs
@@ -1,11 +1,11 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-use crate::{any, Invariant};
+use crate::{any, Arbitrary};
 
 /// Generates an arbitrary vector given a length.
 pub fn any_vec<T, const VEC_LENGTH: usize>() -> Vec<T>
 where
-    T: Invariant,
+    T: Arbitrary,
 {
     let boxed_any_sice: Box<[T; VEC_LENGTH]> = Box::new(any());
     <[T]>::into_vec(boxed_any_sice)
--- a/library/kani/src/vec.rs
+++ b/library/kani/src/vec.rs
@@ -1,12 +1,25 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-use crate::{any, Arbitrary};
+use crate::{any, assume, Arbitrary, Invariant};
 
-/// Generates an arbitrary vector given a length.
-pub fn any_vec<T, const VEC_LENGTH: usize>() -> Vec<T>
+/// Generates an arbitrary vector whose length is at most MAX_LENGTH.
+pub fn any_vec<T, const MAX_LENGTH: usize>() -> Vec<T>
 where
     T: Arbitrary,
 {
-    let boxed_any_slice: Box<[T; VEC_LENGTH]> = Box::new(any());
-    <[T]>::into_vec(boxed_any_slice)
+    let mut v = Vec::with_capacity(MAX_LENGTH);
+    let real_length: usize = any();
+    assume(real_length <= MAX_LENGTH);
+    unsafe { v.set_len(real_length) };
+
+    v
+}
+
+/// Generates an arbitrary vector that is exactly EXACT_LENGTH long.
+pub fn exact_vec<T, const EXACT_LENGTH: usize>() -> Vec<T>
+where
+    T: Invariant,
+{
+    let boxed_array: Box<[T; EXACT_LENGTH]> = Box::new(any());
+    <[T]>::into_vec(boxed_array)
 }
--- a/library/kani/src/vec.rs
+++ b/library/kani/src/vec.rs
@@ -7,6 +7,6 @@ pub fn any_vec<T, const VEC_LENGTH: usize>() -> Vec<T>
 where
     T: Arbitrary,
 {
-    let boxed_any_sice: Box<[T; VEC_LENGTH]> = Box::new(any());
-    <[T]>::into_vec(boxed_any_sice)
+    let boxed_any_slice: Box<[T; VEC_LENGTH]> = Box::new(any());
+    <[T]>::into_vec(boxed_any_slice)
 }
--- a/library/kani/src/vec.rs
+++ b/library/kani/src/vec.rs
@@ -3,7 +3,7 @@
 use crate::{any, Invariant};
 
 /// Generates an arbitrary vector given a length.
-pub fn any_vec<const VEC_LENGTH: usize, T>() -> Vec<T>
+pub fn any_vec<T, const VEC_LENGTH: usize>() -> Vec<T>
 where
     T: Invariant,
 {
--- a/library/kani/src/vec.rs
+++ b/library/kani/src/vec.rs
@@ -1,13 +1,13 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-use crate::{any, assume, Arbitrary, Invariant};
+use crate::{any, assume, Invariant};
 
 /// Generates an arbitrary vector whose length is at most MAX_LENGTH.
 pub fn any_vec<T, const MAX_LENGTH: usize>() -> Vec<T>
 where
-    T: Arbitrary,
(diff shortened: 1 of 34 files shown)


library/kani/src/vec.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+use crate::{any, Invariant};
+
+/// Generates an arbitrary vector given a length.
+pub fn any_vec<const VEC_LENGTH: usize, T>() -> Vec<T>
+where
+    T: Invariant,
+{
+    let boxed_any_sice: Box<[T; VEC_LENGTH]> = Box::new(any());
```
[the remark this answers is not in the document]
celinval: My IDE complains about typos. I use CLion. :)

library/kani/src/vec.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+use crate::{any, Invariant};
+
+/// Generates an arbitrary vector given a length.
+pub fn any_vec<const VEC_LENGTH: usize, T>() -> Vec<T>
```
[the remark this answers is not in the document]
celinval: Not sure I understand the `const` comment, but I do like the idea of using maximum size, like any_slice does.

library/kani/src/vec.rs
```diff
@@ -0,0 +1,12 @@
...
+where
+    T: Invariant,
```
celinval, line 8: `T: Arbitrary`

tests/kani/Vectors/any/sorting.rs
```diff
@@ -0,0 +1,14 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Sort an arbitrary Vec<u32> of length 3. Assert that the sorting worked.
+#[kani::proof]
+#[kani::unwind(10)]
+fn main() {
+    let mut v: Vec<u32> = kani::vec::any_vec::<3, _>();
+    kani::assume(v.len() == 3);
```
celinval, line 9: Why do you need the assume here?

library/kani/src/vec.rs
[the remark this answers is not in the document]
danielsn: I guess the question is do we want vec of exactly that length, or nondet up to a max length

library/kani/src/vec.rs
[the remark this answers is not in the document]
danielsn: You can install spell checkers for vocode too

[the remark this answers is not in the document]
YoshikiTakashima: Thanks. I use emacs, so after some searching, the best option is probably to enable spell checking for "font-lock-variable-name-face" and "font-lock-function-name-face." Snake case seems OK with the spell checker.

tests/kani/Vectors/any/sorting.rs
YoshikiTakashima: P.S. `>= 3` is probably better, fixing to that.

library/kani/src/vec.rs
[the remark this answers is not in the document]
YoshikiTakashima: > But we don't seem to provide any way to do that via slice do we, and that (I assume) is why you wound up with the const...

Yea, not without modifying `slice.rs` so I can see `AnySlice`'s internals. I try to see if I can implement  `From<AnySlice<T, MAX_SLICE_LENGTH>> for Vec<T>`, which converts `AnySlice` to `*[T]`, then `Box<[T]>` then `Vec<T>`. `any_vec` will just call that `From::from`.

library/kani/src/vec.rs
YoshikiTakashima: Yes, I'll change to `T: Arbitrary` when resolving the max size issue. 

I'm curious why any only requires Invariant?

library/kani/src/vec.rs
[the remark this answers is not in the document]
YoshikiTakashima: After some investigation, 2/4 test cases are working, 1 is still not working (#1322) and 1 migrated to the newly forked `exact_vec` API.

After talking with @danielsn, we will expose `exact_vec` that generates exact-length vectors since the performance is significantly better than `any_vec + assume`. See #1322 for details

tests/kani/Vectors/any/sorting.rs
YoshikiTakashima: Due to above performance issue, this test case was shrunk to size 2 to avoid timeout. `==2` is required for indexing, so that is `==` is kept.

library/kani/src/vec.rs
YoshikiTakashima: Update: I'm going to have to go with  `T: Invariant` since I want to be minimal diff between high-performing `exact_vec` and not-so-well `any_vec` and `exact_vec` needs `T: Invariant`.

If the performance issue is fixed (I'm asking @tautschnig tomorrow 6-30), then we can consider ways to loosen it.

tests/kani/Vectors/push.rs
```diff
@@ -4,7 +4,7 @@
 // Push 5 elements to force the vector to resize, then check that the values were correctly copied.
 #[kani::proof]
 fn main() {
-    let mut v = Vec::new();
+    let mut v: Vec<isize> = kani::vec::any_vec::<_, 0>();
```
danielsn, line 7: I'm not sure this makes sense here, given that we're starting with an empty vec. What does it mean to be an unconstrained empty vec?

danielsn (approved):
Looks good, modulo comments

tests/kani/Vectors/push.rs
YoshikiTakashima: Tracking #1329.

PR #1328, opened 2022-07-01
[DRAFT] Support MMIO regions

### LINKED ISSUE #1304: How to verify mmio access?
I've tried with the below code for mocking a mmio access.

```
const BUFFER: *mut u32 = 0xb8000 as *mut u32;
#[cfg(kani)]
#[kani::proof]
fn test_write() {
    let val = 12;
    unsafe {
        core::ptr::write_volatile(BUFFER, val);
    }
}
```

```
cargo kani
```

with Kani version: 0.3.0

It bought out the following failures. 
```
SUMMARY:
 ** 5 of 7 failed
Failed Checks: dereference failure: pointer NULL
 File: "/rustc/c52b9c10bfb5164015eb977ff498e0597ae63eb1/library/core/src/ptr/mod.rs", line 1519, in std::ptr::write_volatile
Failed Checks: dereference failure: deallocated dynamic object
 File: "/rustc/c52b9c10bfb5164015eb977ff498e0597ae63eb1/library/core/src/ptr/mod.rs", line 1519, in std::ptr::write_volatile
Failed Checks: dereference failure: dead object
 File: "/rustc/c52b9c10bfb5164015eb977ff498e0597ae63eb1/library/core/src/ptr/mod.rs", line 1519, in std::ptr::write_volatile
 File: "/rustc/c52b9c10bfb5164015eb977ff498e0597ae63eb1/library/core/src/ptr/mod.rs", line 1519, in std::ptr::write_volatile
Failed Checks: dereference failure: invalid integer address
 File: "/rustc/c52b9c10bfb5164015eb977ff498e0597ae63eb1/library/core/src/ptr/mod.rs", line 1519, in std::ptr::write_volatile

VERIFICATION:- FAILED
```

Why did the verification failures occur and how to pass the verification when an object of absolute address (e.g., mmio) is involved?

danielsn:
### Description of changes: 

Adds support for MMIO regions to kani proof harnesses.

### Resolved issues:

Resolves #1304 

### Call-outs:

- [ ] This is dependent on https://github.com/diffblue/cbmc/pull/6747. It will not work until that is released. I tested against a locally built copy.
- [ ] We need documentation for this feature
- [x] Should probably add some negative tests too
- [ ] Should test that it behaves like MMIO: writes don't need to be the same as is read back, two reads can get different values

### Testing:

* How is this change tested? New regression tests.

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
@@ -405,16 +405,3 @@ fn extract_integer_argument(attr: &Attribute) -> Option<u128> {
         None
     }
 }
-
-// /// Extracts the integer value argument from the attribute provided
-// /// For example, `unwind(8)` return `Some(8)`
-// fn extract_integer_arguments(attr: &Attribute) -> Result<Vec<u128>, String> {
-//     // Vector of meta items , that contain the arguments given the attribute
-//     let attr_args = attr.meta_item_list().unwrap;
-//     Ok(attr_args.iter().map(|a| {
-//         match a.literal()?.kind {
-//             LitKind::Int(y, ..) => Some(y),
-//             _ => None,
-//         }
-//     }).collect())
-// }
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
@@ -13,7 +13,6 @@ use rustc_ast::{Attribute, LitKind};
 use rustc_middle::mir::{HasLocalDecls, Local};
 use rustc_middle::ty::{self, Instance};
 use std::collections::BTreeMap;
-use std::convert::TryInto;
 use std::iter::FromIterator;
 use tracing::{debug, info_span};
 
@@ -272,6 +271,7 @@ impl<'tcx> GotocCtx<'tcx> {
         let mut harness = self.default_kanitool_proof();
         for attr in other_attributes.iter() {
             match attr.0.as_str() {
+                "mmio_region" => self.handle_kanitool_mmio_region(attr.1, &mut harness),
                 "unwind" => self.handle_kanitool_unwind(attr.1, &mut harness),
                 _ => {
                     self.tcx.sess.span_err(
@@ -297,7 +297,29 @@ impl<'tcx> GotocCtx<'tcx> {
             original_file: loc.filename().unwrap(),
             original_line: loc.line().unwrap().to_string(),
             unwind_value: None,
+            mmio_regions: vec![],
         }
+    }
+
+    fn handle_kanitool_mmio_region(&mut self, attr: &Attribute, harness: &mut HarnessMetadata) {
+        let attr_args = attr.meta_item_list();
+        if let Some(attr_args) = attr_args {
+            if attr_args.len() == 2 {
+                if let (Some(start), Some(len)) = (attr_args[0].literal(), attr_args[1].literal()) {
+                    if let (LitKind::Int(start, ..), LitKind::Int(len, ..)) =
+                        (&start.kind, &len.kind)
+                    {
+                        harness.mmio_regions.push((*start, *len));
+                        return;
+                    }
+                }
+            }
+        }
+
+        // If we didn't successfully process, give an error message
+        self.tcx
+            .sess
+            .span_err(attr.span, "Expected exactly two integers are the argument to 'mmio_region'");
     }
 
     /// Updates the proof harness with new unwind value
@@ -307,6 +329,7 @@ impl<'tcx> GotocCtx<'tcx> {
             self.tcx.sess.span_err(attr.span, "Only one '#[kani::unwind]' allowed");
             return;
         }
+
         // Get Attribute value and if it's not none, assign it to the metadata
         match extract_integer_argument(attr) {
             None => {
@@ -382,3 +405,16 @@ fn extract_integer_argument(attr: &Attribute) -> Option<u128> {
         None
     }
 }
+
+// /// Extracts the integer value argument from the attribute provided
+// /// For example, `unwind(8)` return `Some(8)`
+// fn extract_integer_arguments(attr: &Attribute) -> Result<Vec<u128>, String> {
+//     // Vector of meta items , that contain the arguments given the attribute
+//     let attr_args = attr.meta_item_list().unwrap;
+//     Ok(attr_args.iter().map(|a| {
+//         match a.literal()?.kind {
+//             LitKind::Int(y, ..) => Some(y),
+//             _ => None,
+//         }
+//     }).collect())
+// }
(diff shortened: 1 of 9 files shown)


kroening:
Why not simply have `--mmio_region` enabled by default?

kroening:
> Why not simply have `--mmio_region` enabled by default?

(Ignore, it is already)

tests/kani/MMIO/semantics.rs
```diff
@@ -0,0 +1,63 @@
...
+    let val2 = 314;
+    unsafe {
+        let p1 = BUFFER;
+        let p2 = BUFFER.offset(2);
+        *p1 = val1;
+        *p2 = val2;
+        assert_eq!(*p1, val1);
+        assert_eq!(*p2, val2);
+        assert_eq!(*p2, val1);
+        assert_eq!(*p1, val2);
```
zpzigi754, line 61: I guess that these should be changed like 
```
assert_ne(*p2, val1); 
assert_ne(*p1, val2);
```

library/kani_macros/src/lib.rs
```diff
@@ -73,3 +73,26 @@ pub fn unwind(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+#[cfg(not(kani))]
+#[proc_macro_attribute]
+pub fn mmio_region(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    // When the config is not kani, we should leave the function alone
+    item
+}
+
+/// Set Loop unwind limit for proof harnesses
+/// The attribute '#[kani::unwind(arg)]' can only be called alongside '#[kani::proof]'.
+/// arg - Takes in a integer value (u32) that represents the unwind value for the harness.
```
fzaiser, line 86: Update comment

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -297,16 +297,39 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        return;
+                    }
+                }
+            }
+        }
+
+        // If we didn't successfully process, give an error message
+        self.tcx
+            .sess
+            .span_err(attr.span, "Expected exactly two integers are the argument to 'mmio_region'");
```
celinval, line 322: Can you print or at least log which values we got?

kani_metadata/src/harness.rs
```diff
@@ -16,4 +16,6 @@ pub struct HarnessMetadata {
     pub original_line: String,
     /// Optional data to store unwind value
     pub unwind_value: Option<u32>,
+    /// Optional MMIO regions
+    pub mmio_regions: Vec<(u128, u128)>,
```
celinval, line 20: Should we use `usize` instead?

tests/expected/mmio/expected
```diff
@@ -0,0 +1,26 @@
+Status: SUCCESS\
+Description: "offset: attempt to compute number in bytes which would overflow"
```
celinval, line 2: This seems irrelevant for your test. I would just remove it.

tests/expected/mmio/main.rs
```diff
@@ -0,0 +1,18 @@
...
+#[cfg(kani)]
+#[kani::proof]
+#[kani::mmio_region(0x8000, 8)]
+fn test_write() {
+    let val = 12;
+    unsafe {
+        //BUFFER+2 is not in the MMIO region. Expect pointer check failures.
+        *(BUFFER.offset(2)) = val;
+        //BUFFER2 is not in the MMIO region. Expect pointer check failures.
+        *BUFFER2 = val;
```
celinval, line 16: You are not really checking that both fail though. This test will succeed if only one of these fail. You can either add the line in the file to the expected file or break this into two harness and add the function name. I prefer the second since it is always painful to maintain LoC up to date.

tests/expected/mmio/main.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+const BUFFER: *mut u32 = 0x8000 as *mut u32;
+const BUFFER2: *mut u32 = 0x1000 as *mut u32;
+
+#[cfg(kani)]
+#[kani::proof]
+#[kani::mmio_region(0x8000, 8)]
+fn test_write() {
```
celinval, line 10: can you rename this to `check_invalid_write()` or something like that?

tests/kani/MMIO/semantics.rs
```diff
@@ -0,0 +1,63 @@
...
+#[kani::mmio_region(0x8000, 4)]
+fn test_read_is_stable() {
+    unsafe {
+        let val1 = core::ptr::read_volatile(BUFFER);
+        let val2 = core::ptr::read_volatile(BUFFER);
+        assert_eq!(val1, val2);
+    }
+}
+
+//TODO Weirdly, these fail even tho the other tests pass???
```
celinval, line 29: This looks like a blocker to me. Maybe we should investigate this behavior a bit better.

tests/kani/MMIO/semantics.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+const BUFFER: *mut u32 = 0x8000 as *mut u32;
+
+#[cfg(kani)]
+#[kani::proof]
+#[kani::mmio_region(0x8000, 4)]
+fn test_read_after_write_doesnt_havoc() {
+    let val = 12;
```
celinval, line 10: Can you use `any` instead?

tests/kani/MMIO/semantics.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright Kani Contributors
```
celinval, line 1: Can you please document what these tests are testing?

tests/kani/MMIO/test.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright Kani Contributors
```
celinval, line 1: Can you cover 0 sized operations?

tests/kani/MMIO/semantics.rs
```diff
@@ -0,0 +1,63 @@
...
+        core::ptr::write_volatile(BUFFER, val);
+        let new_val = core::ptr::read_volatile(BUFFER);
+        assert_eq!(new_val, val);
+    }
+}
+
+#[cfg(kani)]
+#[kani::proof]
+#[kani::mmio_region(0x8000, 4)]
+fn test_read_is_stable() {
```
celinval, line 21: But should it be stable?

celinval (changes requested):
The semantics here aren't clear to me. Your `test_read_is_stable` contradicts the item in the callout:

> Should test that it behaves like MMIO: writes don't need to be the same as is read back, two reads can get different values.

celinval:
@danielsn any news here?

PR #1334, opened 2022-07-05
feat: POC for kani driver changes

sanjit-bhat:
### Description of changes: 

Adds executable trace flag to the Kani driver. This flag first runs the normal verification flow, except that it also uses CBMC to get an output trace and parses this trace for concrete values. At the end of verification, it also calls Kani compiler with the executable trace flag to generate a concrete executable, and it runs this.

### Testing:

There is a basic parser test that's labeled as a UI test.


kani-driver/src/args.rs
```diff
@@ -43,6 +43,10 @@ pub struct KaniArgs {
     /// Generate visualizer report to <target-dir>/report/html/index.html
     #[structopt(long)]
     pub visualize: bool,
+    /// Generate executable trace
+    // TODO: Need to exclude other options with this
+    #[structopt(long)]
+    pub gen_exec_trace: bool,
```
celinval, line 49: Can you please mark this as unstable? I.e.: require --enable-unstable for now.

kani-driver/src/args.rs
```diff
@@ -166,6 +170,7 @@ impl KaniArgs {
 
     pub fn assertion_reach_checks(&self) -> bool {
         // Turn them off when visualizing an error trace.
+        // TODO: disable this for exec_trace flag?
```
celinval, line 173: You probably want to keep them since you are still printing the verification report. You should however ignore the traces that get printed for them.

kani-driver/src/call_cargo.rs
```diff
@@ -23,13 +23,13 @@ pub struct CargoOutputs {
 
 impl KaniSession {
     /// Calls `cargo_build` to generate `*.symtab.json` files in `target_dir`
-    pub fn cargo_build(&self) -> Result<CargoOutputs> {
+    pub fn cargo_build(&self, has_run_once: bool) -> Result<CargoOutputs> {
```
celinval, line 26: I think instead of `has_run_once` you should either name it `build_gotoc` or even change this to take the the following enum parameter:
```
enum Backend {
  GOTOC,
  LLVM,
}
```

That will scale nicely once we add other backends. That said, I believe @zhassan-aws was suggesting that we use something different than backend. So we might want to ask if he has any other suggestion.

kani-driver/src/get_det_vals.rs
```diff
@@ -0,0 +1,108 @@
...
+use std::io::BufReader;
+use std::path::{Path, PathBuf};
+
+use crate::session::KaniSession;
+use crate::util::alter_extension;
+use anyhow::Result;
+use kani_metadata::HarnessMetadata;
+use serde_json::Value;
+
+impl KaniSession {
```
celinval, line 15: You should separate the parser code from cbmc execution code. The parser shouldn't be part of KaniSession. This is getting over-used.

kani-driver/src/get_det_vals.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Extract deterministic values from a failing harness.
+    pub fn get_det_vals(&self, file: &Path, harness_metadata: &HarnessMetadata) -> Result<()> {
+        let results_filename = alter_extension(file, "results.json");
+        let det_vals_filename = alter_extension(file, "det_vals.txt");
+
+        {
+            let mut temps = self.temporaries.borrow_mut();
```
celinval, line 22: These shouldn't be temporaries since we want to enable users to run them from their IDEs.

kani-driver/src/get_det_vals.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use crate::session::KaniSession;
+use crate::util::alter_extension;
+use anyhow::Result;
+use kani_metadata::HarnessMetadata;
+use serde_json::Value;
+
+impl KaniSession {
+    /// Extract deterministic values from a failing harness.
+    pub fn get_det_vals(&self, file: &Path, harness_metadata: &HarnessMetadata) -> Result<()> {
```
celinval, line 17: Can improve the name? I expect a `get*` method to return something.

kani-driver/src/main.rs
```diff
@@ -73,6 +77,23 @@ fn cargokani_main(input_args: Vec<OsString>) -> Result<()> {
         let result = ctx.check_harness(&specialized_obj, &report_dir, harness)?;
         if result == VerificationStatus::Failure {
             failed_harnesses.push(harness);
+
+            if !has_run_once && ctx.args.gen_exec_trace {
```
celinval, line 81: Can you move this code to be outside of this loop? Instead of having the `has_run_once`, you can just iterate over the list of `failed_harnesses`. For now, the case where you only run this once, you can just get the first element of the list.

kani-driver/src/main.rs
```diff
@@ -145,6 +172,9 @@ impl KaniSession {
             self.run_visualize(binary, report_dir, harness)?;
             // Strictly speaking, we're faking success here. This is more "no error"
             Ok(VerificationStatus::Success)
+        } else if self.args.gen_exec_trace {
+            self.get_det_vals(binary, harness)?;
+            self.run_cbmc(binary, harness)
```
celinval, line 177: why are you invoking cbmc here?

tests/ui/exec-trace/det_vals.rs
```diff
@@ -0,0 +1,83 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// kani-flags: --debug --harness ovoa --gen-exec-trace
```
celinval, line 4: Can you add a comment on what this is testing please? Maybe state that for now we only print the values that were extracted.

kani-driver/src/get_det_vals.rs
```diff
@@ -0,0 +1,108 @@
...
+        let results_filename = alter_extension(file, "results.json");
```
celinval, line 18: Can you please add the name of the harness to the file name?

kani-driver/src/get_det_vals.rs
```diff
@@ -0,0 +1,108 @@
...
+        // Let the user know
+        if !self.args.quiet {
+            println!("Extracting deterministic values from trace");
+        }
+
+        let cbmc_out = get_cbmc_out(results_filename);
+        let det_vals = handle_cbmc_out(&cbmc_out);
+        let joined_det_vals = det_vals.join("\n");
+        if self.args.debug {
+            println!("Det vals:\n{}", joined_det_vals);
```
celinval, line 38: Maybe print how to use the files and their path.

celinval (commented):
Thanks for doing this! I think it would be nice if you restructure this a bit to make it cleaner. The exact flow is not clear to me.

PR #1335, opened 2022-07-05
feat: POC for kani lib changes

sanjit-bhat:
### Description of changes: 

This PR modifies Kani any_raw to use byte arrays as the low-level unit of non-determinism. This enables us to use a different any_raw backend when we compile with LLVM that injects concrete values from a file.

### Testing:

1. None, as of yet.
2. I had to modify some of the existing regression tests to add support for generic constants.

### DIFF
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 #![feature(rustc_attrs)] // Used for rustc_diagnostic_item.
 #![feature(min_specialization)] // Used for default implementation of Arbitrary.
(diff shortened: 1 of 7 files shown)


library/kani/src/lib.rs
```diff
@@ -94,10 +99,36 @@ pub fn any<T: Arbitrary>() -> T {
 /// kani::assume(char::from_u32(c as u32).is_ok());
 /// ```
 ///
+#[inline(never)]
+pub unsafe fn any_raw<T>() -> T
+where
+    [(); std::mem::size_of::<T>()]:,
```
celinval, line 105: maybe add a comment about what this `where` is doing so people don't have to figure out on their own.

Issue #1320, opened 2022-06-28
Tutorial documentation revisions tracking issue


tedinski:
Review of existing tutorial

* [x] Review [2.1. First steps](https://model-checking.github.io/kani/tutorial-first-steps.html)
* [x] Review [2.2. Failures that Kani can spot](https://model-checking.github.io/kani/tutorial-kinds-of-failure.html)
* [x] Review [2.3. Loop unwinding](https://model-checking.github.io/kani/tutorial-loop-unwinding.html)
* [ ] Review [2.4. Nondeterministic variables](https://model-checking.github.io/kani/tutorial-nondeterministic-variables.html)
* [ ] Review [3.1. Comparison with other tools](https://model-checking.github.io/kani/tool-comparison.html)
* [ ] Review [3.2. Where to start on real code](https://model-checking.github.io/kani/tutorial-real-code.html)
* [ ] Review the getting started and installation flow

Notable problems to fix:

* [x] We should recommend `#[kani::unwind]` not `--default-unwind`.
* [x] May need to review tutorial in light of having multiple proof harnesses. Much of it was written before this was implemented.
* [ ] We should collect "real world examples" somewhere which can at least include links to our current blog posts.
* [x] Let's be more specific about debugging traces in the tutorial
* [x] Ditto debugging non-termination
* [x] We may need new content to help cover "debugging coverage" as a concept. Perhaps this can be a part of the new section on spec writing (below)
* [x] Where do we document annotations like proof/unwind/etc?
* [x] Tutorials should be about using `cargo kani` on a project, not `kani` on an individual file.
* [x] As a final polishing step, make sure we don't leave any harnesses named `main` anymore.
* [ ] `is_valid` is currently mentioned on https://model-checking.github.io/kani/undefined-behaviour.html

Possible new section to write:

* Specification writing. Perhaps "how to specify that sort works". Include minimal techniques like "nondet is forall".
* More advanced spec techniques, like object invariants.
* [ ] Add an explanation about how builds work, the need for `cfg(kani)`, the `kani` dependency, etc

PR #1337, opened 2022-07-05
Revise 'Kind of Failure' tutorial

tedinski:
### Description of changes: 

Notably:

1. Re-adds end of section summary
2. Fix things to talk about `cargo kani` not `kani`
3. Updates based on our formatting changes (e.g. we already get end summaries, so don't need to grep)
4. Updates based on fixes in Kani (the traces are a lot easier to read, yay!)
5. various wordsmithing

### Resolved issues:

Towards #1320

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


docs/src/tutorial-kinds-of-failure.md
```diff
@@ -63,54 +76,57 @@ Any unsafe code that manipulates pointers will, as we see here, raise failures i
...
+VERIFICATION:- FAILED
 ```
 
-The first is Rust's implicit assertion for the safe indexing operation.
+The first is Rust's runtime bounds checking for the safe indexing operation.
 The second is Kani's check to ensure the pointer operation is actually safe.
-This pattern (two checks for similar issues in safe Rust code) is common, and we'll see it again in the next section.
+This pattern (two checks for similar issues in safe Rust code) is common to see, and we'll see it again in the next section.
+
+(Not that, while Kani will emit both kinds of checks always, in the future the output here may change.
```
zhassan-aws, line 106:
```suggestion
(Note that, while Kani will emit both kinds of checks always, in the future the output here may change.
```

docs/src/tutorial-kinds-of-failure.md
```diff
@@ -120,13 +136,12 @@ let index: usize = kani::any();
...
 ```
 
-Try not to be fooled by the first assignments: we're seeing zero-initialization there.
-Their values are overwritten by the later assignments.
 You may see different values here, as it depends on the solver's behavior.
 
-2. Try searching for "failure:". This will be near the end of the document.
-Now you can try reverse-searching for assignments to the variables involved.
-For example, search upwards from the failure for `i =`.
+2. Try searching for `failure:`. This will be near the end of the document.
```
zhassan-aws, line 141: Replace "document" with "page" or "trace"?

docs/src/tutorial-kinds-of-failure.md
```diff
@@ -120,13 +136,12 @@ let index: usize = kani::any();
...
+You can now search upwards from a failure to see what value certain variables had.
+Sometime it can be helpful to change the source code to add intermediate variables, so their value is visible in the trace.
```
zhassan-aws, line 143:
```suggestion
Sometimes it can be helpful to change the source code to add intermediate variables, so their value is visible in the trace.
```

docs/src/tutorial-kinds-of-failure.md
```diff
@@ -120,13 +136,12 @@ let index: usize = kani::any();
...
+For instance, you might want to compute the index before indexing into the array, that way you'd see exactly what value is being used.
```
zhassan-aws, line 144:
```suggestion
For instance, you might want to compute the index before indexing into the array. That way you'd see exactly what value is being used.
```

docs/src/tutorial-kinds-of-failure.md
```diff
@@ -63,54 +76,57 @@ Any unsafe code that manipulates pointers will, as we see here, raise failures i
...
 ```
 
-The first is Rust's implicit assertion for the safe indexing operation.
+The first is Rust's runtime bounds checking for the safe indexing operation.
 The second is Kani's check to ensure the pointer operation is actually safe.
-This pattern (two checks for similar issues in safe Rust code) is common, and we'll see it again in the next section.
+This pattern (two checks for similar issues in safe Rust code) is common to see, and we'll see it again in the next section.
+
+> **NOTE**: While Kani will always emit both kinds of checks, in the future the output here may change.
+> Kani is not currently recognizing that the runtime bounds check means it cannot reach the unsafe pointer dereference.
```
zhassan-aws, line 107: Suggest rephrasing to something along the lines of:
Kani does not currently realize that the unsafe pointer dereference check is redundant: if the runtime bounds check passes, the pointer dereference check will also pass, and it the runtime bounds check fails, the pointer dereference check will also fail (or will be unreachable?).

zhassan-aws (approved):
LGTM other than the comment on the redundant checks note.

docs/src/tutorial-kinds-of-failure.md
tedinski: Ok, maybe this is our miscommunication here.

It's not redundant checks: I'm trying to say we'd always emit both.

What I suspect we're getting here is that the "assert" that Rust is inserting for the bounds check is getting translated as "assert" to cbmc, not our usual "assert-assume".

PR #1338, opened 2022-07-05
Represent FnDefs as zero-sized structs instead of Code

### LINKED ISSUE #1243: Field of type `Code` in struct when making closure
When running the `tests/kani/Refs/main.rs` regression, we get the following error if we turn on typechecking for data-fields.

```
thread 'rustc' panicked at 'Illegal field.
        Name: 0
        Type: Code { parameters: [Parameter { typ: CInteger(SizeT), identifier: None, base_name: None }, Parameter { typ: CInteger(SizeT), identifier: None, base_name: None }], return_type: StructTag("tag-_12144896164184170153") }', /Users/dsn/ws/rmc/cprover_bindings/src/goto_program/typ.rs:187:9
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: cprover_bindings::goto_program::typ::DatatypeComponent::field
             at /Users/dsn/ws/rmc/cprover_bindings/src/goto_program/typ.rs:187:9
   3: kani_compiler::codegen_cprover_gotoc::codegen::typ::<impl kani_compiler::codegen_cprover_gotoc::context::goto_ctx::GotocCtx>::codegen_struct_fields
             at /Users/dsn/ws/rmc/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs:732:39
   4: kani_compiler::codegen_cprover_gotoc::codegen::typ::<impl kani_compiler::codegen_cprover_gotoc::context::goto_ctx::GotocCtx>::codegen_ty_tuple_like
             at /Users/dsn/ws/rmc/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs:765:9
   5: kani_compiler::codegen_cprover_gotoc::codegen::typ::<impl kani_compiler::codegen_cprover_gotoc::context::goto_ctx::GotocCtx>::codegen_ty_closure::{{closure}}
```

Looks like we're codegenning `Code` instead of `Code*` when doing a closure

fzaiser:
### Description of changes: 

Before this change, Kani compiled functions (`FnDef`s) to `Code` in GotoC, i.e. C function types. This causes problems because functions in C are not first-class, e.g. they cannot be stored in a struct. In Rust, this is possible: functions are just zero-sized objects of a unique anonymous type. To allow the same behavior in GotoC, we need to actually represent `FnDef`s as zero-sized objects.

For this reason, we create for each function instance `f` in Rust an empty dummy type `f::FnDefStruct` and a global variable `f::FnDefSingleton`. We need to do this for each instance to ensure that function pointer equality behaves as expected. This is done by the `codegen_fun_expr_zst`.

In some contexts, e.g. when calling a function or casting a function to a function pointer, we need to convert the dummy struct back to an actual function. This is done by calling `codegen_func_expr` instead.

This change also allows us to enable type checking of fields (cf. #1243).

### Resolved issues:

Resolves #1243 

### Call-outs:

Also, we no longer ignore `FnDef` parameters of functions. I believe this is desirable because it makes the code more consistent, but I can undo the change if you want.

### Testing:

* How is this change tested? Additional tests in `tests/kani/FunctionSymbols`, plus the existing regression tests exercising this code as well.

### Checklist


tautschnig:
Just curious: do you actually need to put those elements in the struct at all? I can see how that might make sense at the Rust level for perhaps it introduces scopes or convenient names, but maybe this just doesn't need to be carried over?

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -632,6 +632,41 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        // This zero-sized object that a function name refers to in Rust is globally unique, so we create such a global object.
+        let fn_singleton_name = format!("{func}::FnDefSingleton");
+        let fn_singleton = self.ensure_global_var(
+            &fn_singleton_name,
+            false,
+            fn_struct_ty.clone(),
+            Location::none(),
+            |_, _| None, // zero-sized, so no initialization necessary
+        );
+        fn_singleton.with_location(self.codegen_span_option(span.cloned()))
```
danielsn, line 657: Why clone the `span`

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -696,7 +696,21 @@ impl<'tcx> GotocCtx<'tcx> {
         t: Ty<'tcx>,
     ) -> Expr {
         match k {
-            PointerCast::ReifyFnPointer => self.codegen_operand(o).address_of(),
+            PointerCast::ReifyFnPointer => match self.operand_ty(o).kind() {
```
danielsn, line 699: Comment explaining what this does

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -696,7 +696,21 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                ty::FnDef(def_id, substs) => {
+                    // To be safe, run normal codegen to ensure the symbol table is updated as normally.
+                    // Maybe this isn't necessary though?
```
danielsn, line 702: Issue to track this

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -696,7 +696,21 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    let _ = self.codegen_operand(o);
+                    // Resolve the function and create a pointer to it:
```
danielsn, line 704: Does this create a pointer to the function? Or to the zst representing the fn

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -440,7 +440,13 @@ impl<'tcx> GotocCtx<'tcx> {
                     | InstanceDef::ReifyShim(..)
                     | InstanceDef::ClosureOnceShim { .. }
                     | InstanceDef::CloneShim(..) => {
-                        let func_exp = self.codegen_operand(func);
+                        // We need to codegen the operand normally even though we discard the result
+                        // because we need to ensure the symbol table is updated correctly.
+                        // (When deleted, the function `lang_start_internal` is not found anymore when running `cargo kani`.)
+                        let _ = self.codegen_operand(func);
```
danielsn, line 446: Is this the same issue as above

fzaiser:
@tautschnig I think we could filter out struct fields, parameters and local variables of `FnDef` type and not generate code for them. But staying closer to the MIR has the advantage of less special casing in the GotoC generation. Kani has also introduced some type checking (which is where #1243 was discovered in the first place) where the GotoC type is compared with the MIR type. This would also require more special casing if we just erase `FnDef`s. Do you think there's a performance penalty with the current approach?

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -632,6 +632,41 @@ impl<'tcx> GotocCtx<'tcx> {
             )
             .with_is_extern(true)
         });
+        (func, funct)
+    }
+
+    /// For a given function instance, generates an expression for the function symbol of type Code.
+    pub fn codegen_func_expr(&mut self, instance: Instance<'tcx>, span: Option<&Span>) -> Expr {
```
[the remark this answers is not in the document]
fzaiser: Turns out `codegen_func_expr_zst` does not need to be `pub`, but the rest do. I'll fix that.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
fzaiser: Because `codegen_span_option` expects an `Option<Span>`, not `Option<&Span>`.

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -440,7 +440,13 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        // We need to handle FnDef items specially because we compile them to dummy structs.
```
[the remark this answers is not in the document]
fzaiser: Where should I put all these explanations? It would be good if we had one place where we describe how we compile certain language constructs instead of scattered comments.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -618,7 +618,7 @@ impl<'tcx> GotocCtx<'tcx> {
         self.find_function(&fname).unwrap().call(vec![init])
     }
 
-    pub fn codegen_func_expr(&mut self, instance: Instance<'tcx>, span: Option<&Span>) -> Expr {
+    pub fn ensure_func(&mut self, instance: Instance<'tcx>) -> (String, Type) {
```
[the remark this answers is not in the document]
danielsn: That's reasonable then

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
fzaiser: Added a comment.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
fzaiser: I was able to remove this in the other place, so it should be safe here too.

fzaiser: I've removed it in my latest commit.

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
fzaiser: I was actually able to remove this code now. I'm not sure why it wasn't working before.

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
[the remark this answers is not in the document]
fzaiser: I've now added a link to this PR. I think it gives the most context for the changes.

fzaiser:
I just added some explanations and a link to this PR in the comments to provide context. i also updated the PR description.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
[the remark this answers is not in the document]
danielsn: Might be cleaner to just return the symbol, which would have the name and type included

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -440,7 +440,10 @@ impl<'tcx> GotocCtx<'tcx> {
                     | InstanceDef::ReifyShim(..)
                     | InstanceDef::ClosureOnceShim { .. }
                     | InstanceDef::CloneShim(..) => {
-                        let func_exp = self.codegen_operand(func);
+                        // We need to handle FnDef items in a special way because `codegen_operand` compiles them to dummy structs.
```
celinval, line 443: This comment is duplicated

celinval (commented):
That's a clever approach. The function names are a bit confusing to me though. I would suggest moving the "ensure*" functions to `typ.rs` and rename them to `codegen_*_type`. I also suggest renaming `codegen_func_expr_zst` to something like `codegen_fn_item`.

celinval:
You could also add the definition from the rust reference: https://doc.rust-lang.org/reference/types/function-item.html

tests/kani/FunctionSymbols/main.rs
```diff
@@ -0,0 +1,65 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// examples discovered while debugging issue https://github.com/model-checking/kani/issues/1243
```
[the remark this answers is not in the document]
fzaiser: This comment doesn't apply to the whole file, just the first half. Come to think of it, I'll split the test into two. So I can add proper comments.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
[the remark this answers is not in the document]
fzaiser: Good point, done.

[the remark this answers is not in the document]
fzaiser: Actually, that does not work. Because at least for the function `memcmp`, the signature generated by `codegen_function_sig` is different from the type that's already in the symbol table:
```
$ kani tests/kani/Whitespace/main.rs
[...]
```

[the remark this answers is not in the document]
fzaiser: I did not want to spend too much time on this, so I decided to just return the symbol and the type for now. I filed #1350 to track this problem and added a fixme in the code.

danielsn (approved):

PR #1352, opened 2022-07-08
Update kani's std library to use rust 2021

### LINKED ISSUE #1308: Bumping `library/std` edition to 2021 causes regression failures
These appear to be related to something (abort?) now using a format-string panic message instead of a string constant.

However, this raises the possibly interesting question: are we linking an edition-2018 std lib to (possibly) edition-2021 code from customers?

I'm not sure exactly how the edition has this effect here (e.g. there appears to be just one stdlib `rlib` in the rust toolchain, not one per edition), so 🤷

Needs investigation. Nobody's bothered us about it yet...

celinval:
### Description of changes: 

For that, we needed to change how we handle panic messages. In rust 2021, the panic implementation will invoke `panic_fmt()` (except when no argument is given), which we cannot handle really well. Create our own panic function and use that instead.

### Resolved issues:

Resolves #1308 

### Testing:

### Checklist

### DIFF
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -115,6 +115,10 @@ pub fn expect_fail(_cond: bool, _message: &'static str) {}
 
 /// Function used to generate panic with a static message as this is the only one currently
 /// supported by Kani display.
+///
+/// During verification this will get replaced by `assert(false)`. For concrete executions, we just
+/// invoke the regular `std::panic!()` function. This function is used by our standard library
+/// overrides, but not the other way around.
 #[inline(never)]
 #[rustc_diagnostic_item = "KaniPanic"]
 pub fn panic(message: &'static str) -> ! {
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -113,5 +113,13 @@ pub fn nondet<T: Arbitrary>() -> T {
 #[rustc_diagnostic_item = "KaniExpectFail"]
 pub fn expect_fail(_cond: bool, _message: &'static str) {}
 
+/// Function used to generate panic with a static message as this is the only one currently
+/// supported by Kani display.
+#[inline(never)]
+#[rustc_diagnostic_item = "KaniPanic"]
+pub fn panic(message: &'static str) -> ! {
+    panic!("{}", message)
+}
+
 /// Kani proc macros must be in a separate crate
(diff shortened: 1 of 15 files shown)


library/kani/src/lib.rs
```diff
@@ -113,5 +113,13 @@ pub fn nondet<T: Arbitrary>() -> T {
 #[rustc_diagnostic_item = "KaniExpectFail"]
 pub fn expect_fail(_cond: bool, _message: &'static str) {}
 
+/// Function used to generate panic with a static message as this is the only one currently
+/// supported by Kani display.
+#[inline(never)]
+#[rustc_diagnostic_item = "KaniPanic"]
+pub fn panic(message: &'static str) -> ! {
+    panic!("{}", message)
```
zhassan-aws, line 121: Is this line needed?

zhassan-aws (approved):

library/kani/src/lib.rs
celinval: We need something that never returns, so I just figured that it would be nice to have this statement for when we have the executable trace.

zhassan-aws: I see. I got confused initially that there's mutual recursion (`panic` function calls macro and vice versa), but then realized that this is a diagnostic item, so the `panic` macro is not really getting called.

danielsn: `unreachable!` macro might make sense for this use case

tedinski: IMO `panic` makes sense, but maybe add a comment saying why it's not infinite recursion. (Obvious, I think, after pointing it out, but looks like an easy confusion to me.)

library/std/Cargo.toml
```diff
@@ -6,7 +6,8 @@
 # standard library symbols are preserved
 name = "std"
 version = "0.5.0"
-edition = "2018"
+edition = "2021"
```
tedinski, line 9: Bump the edition for the other `library/` crates? I think I left all of them 2018 iirc

tedinski (approved):
Minor tweaks, plus obviously that rustdoc panic?

library/std/Cargo.toml
celinval: Ah. Thanks for pointing it out. I thought you had only left the std one.

PR #1353, opened 2022-07-08
Revise down the nondeterminism tutorial

tedinski:
### Description of changes: 

The existing nondeterminism tutorial consisted of:

1. Mentioning how `any` is safe
2. Using `any_raw` which is just asking for undefined behavior.
3. Using `Invariant` which... would be undefined behavior on most types.

Ugh. I gave up. I just removed the second two sections and added a little bit more to the first. We have some nasty problems to deal with here that go way beyond writing a tutorial section.

### Resolved issues:

Towards #1320 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,64 @@
...
+For `NonZeroU32`, this means we never return a `0` value.
+The assertion we wrote in this harness was just an extra check we added to demonstrate this fact.
 
-This is an interesting issue that emerges from how `rustc` optimizes the memory layout of `Option<NonZeroU32>`.
-The compiler is able to represent `Option<NonZeroU32>` using `32` bits by using the value `0` to represent `None`.
+## Other kinds of nondeterministic variables
 
-You can try it out by running the example under [`arbitrary-variables`](https://github.com/model-checking/kani/tree/main/docs/src/tutorial/arbitrary-variables/):
+There's nothing special about `kani::any()` except that Kani ships with it only implemented for types where we can guarantee safety.
+Notably, however, there is no implementation for `Vec<T>`, for example.
```
celinval, line 47: But we just added `any_vec`.

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,64 @@
...
 
-```
-cargo kani --harness unsafe_update
-```
+The trouble with a nondeterministic vector is that you usually need to _bound_ the size of the vector, as we saw in the last chapter.
+There are no arguments provided to `kani::any()` to indicate an upper bound.
 
-## Safe nondeterministic variables for custom types
+Likewise, `kani::any()` is implemented for `[T; N]` (if implemeneted for `T`).
+But, it is _not_ implemented for `&[T]`.
```
celinval, line 53: This doesn't make a lot of sense to me. This is a fat pointer. Maybe we should say that we can't handle non-deterministic pointers or that non-deterministic objects that involve memory allocation require special handling?

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,64 @@
...
+Again, there's no upper bound information we can make use of.
 
-Now you would like to add a new structure to your library that allow users to represent a review rating, which can go from 0 to 5 stars.
-Let's say you add the following implementation:
+This does not mean you cannot have a nondeterministic vector.
+You just have to construct it.
```
celinval, line 57: But how? I think this is missing. We could document the Arbitrary type here. That is a very safe trait.

docs/src/tutorial/arbitrary-variables/src/inventory.rs
```diff
@@ -42,23 +44,4 @@ mod verification {
...
-
-    // ANCHOR: unsafe_update
-    #[kani::proof]
-    pub fn unsafe_update() {
-        // Create inventory variable.
-        let mut inventory = Inventory { inner: VecMap::new() };
-
-        // Create non-deterministic variables for id and quantity with unsafe kani::any_raw().
-        let id: ProductId = kani::any();
-        let quantity: NonZeroU32 = unsafe { kani::any_raw() };
```
celinval, line 54: Should we change `kani::any_raw()` to be private?

tedinski:
The is ready for review again. I've revised down the content still, but I also opened this tracking issue:

* #1394 

I think eventually we should add more here, but there are things in the issue that I think need fixing first. This at least revises this chapter and makes it usable today.

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,103 @@
...
 ```
 
-## Unsafe nondeterministic variables
+`kani::any()` is safe Rust, and so Kani only implements it for types where type invariants are enforced.
+For `NonZeroU32`, this means we never return a `0` value.
+The assertion we wrote in this harness was just an extra check we added to demonstrate this fact.
 
-Kani also includes an **unsafe** method to generate unconstrained nondeterministic variables which do not take type invariants into consideration.
-As with any unsafe method in Rust, users have to make sure the right guardrails are put in place to avoid undesirable behavior.
+## Other kinds of nondeterministic variables
```
celinval, line 44: What are you trying to say in this section? I'm tempted to just remove this for now.

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,103 @@
...
+This does not mean you cannot have a nondeterministic vector.
+It just means you have to construct one.
+Our example proof harness above constructs a nondeterministic `Inventory` of size `1`, simply by starting with the empty `Inventory` and inserting a nondeterministic entry.
 
-This is an interesting issue that emerges from how `rustc` optimizes the memory layout of `Option<NonZeroU32>`.
-The compiler is able to represent `Option<NonZeroU32>` using `32` bits by using the value `0` to represent `None`.
+## Custom nondeterministic types
 
-You can try it out by running the example under [`arbitrary-variables`](https://github.com/model-checking/kani/tree/main/docs/src/tutorial/arbitrary-variables/):
+When you need nondeterministic variables of types that `kani::any()` cannot construct, our current recommendation is to simply write a function.
```
celinval, line 62: Why not mention the Arbitrary type?

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,103 @@
...
+> We'd overwhelming generate a `Three` because it's matching "all other integers besides 1 and 2."
+> But Kani just see 3 meaningful possibilities, each of which is not treated any differently from each other.
+> The "proportion" of integers does not matter.
 
-```rust
-{{#include tutorial/arbitrary-variables/src/rating.rs:rating_struct}}
-```
 
-The easiest way to allow users to create nondeterministic variables of the Rating type which represents values from 0-5 stars is by implementing the `kani::Invariant` trait.
+## Exercise
```
celinval, line 80: I like this. Should we provide an answer for this? Maybe refer to a file that they can open at the end.

docs/src/tutorial/arbitrary-variables/src/rating.rs
```diff
@@ -3,40 +3,42 @@
...
 
     // ANCHOR: rating_invariant
-    unsafe impl kani::Invariant for Rating {
-        fn is_valid(&self) -> bool {
-            self.value <= 5
+    fn any_rating() -> Rating {
+        match kani::any() {
+            0 => Rating::One,
+            1 => Rating::Two,
+            _ => Rating::Three,
```
celinval, line 32: This is a bit counter intuitive since it doesn't match `as_int()`.

celinval (commented):
It's looking good. Thanks for taking the time.

docs/src/tutorial-nondeterministic-variables.md
tedinski: That was my original plan, and then Nathan tried it and immediately ran into the lack of `Option<MyType>`.

So I've noted that in the associate issue and was just going to leave it as-is for now.

docs/src/tutorial/arbitrary-variables/src/rating.rs
tedinski: Hmm, I agree, but maybe that's a feature: emphasize that there's not reason for it to match.

docs/src/tutorial-nondeterministic-variables.md
tedinski: I think this structure is good bones, but I see what you're saying about it seeming pointless right now.

As soon as we decide things like `any_vec` are stable and we commit to their API, this is where I'd put it. So this section was kinda meant to be about that until I went and audited the docs and felt that maybe we wanted to revise that API first...

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,105 @@
...
 
 As a Rust developer, this sounds a lot like the `mem::transmute` operation, which is highly `unsafe`.
 And that's correct.
+But `kani::any()` is only implemented for a few types: those where we are able to safely enforce the type's invariants.
 
-In this tutorial, we will show how to:
- 1. Safely use nondeterministic assignments to generate valid symbolic variables that respect Rust's type invariants.
- 2. Unsafely use nondeterministic assignments to generate unconstrained symbolic variables that do not respect Rust's type invariants.
- 2. Specify invariants for types that you define, enabling the creation of safe nondeterministic variables for those types.
+In other words, `kani::any()` should not produce values that are invalid for the type.
```
zhassan-aws, line 10: Join this sentence with the previous paragraph?

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,105 @@
 # Nondeterministic variables
 
-Kani is able to reason about programs and their execution paths by allowing users to assign nondeterministic (i.e., symbolic) values to  certain variables.
-Since Kani is a bit-level model checker, this means that Kani considers that an unconstrained nondeterministic value represents all the possible bit-value combinations assigned to the variable's memory contents.
+Kani is able to reason about programs and their execution paths by allowing users to assign nondeterministic (also called symbolic) values to certain variables using `kani::any()`.
+Kani is a "bit-precise" model checker, which means that Kani considers all the possible bit-value combinations that could be assigned to the variable's memory contents.
 
 As a Rust developer, this sounds a lot like the `mem::transmute` operation, which is highly `unsafe`.
 And that's correct.
+But `kani::any()` is only implemented for a few types: those where we are able to safely enforce the type's invariants.
```
zhassan-aws, line 8: I think this will beg the question: "What should I do if I need `kani::any()` for one of the unimplemented types?" So perhaps add a reference to the section that discusses that, e.g:

(See Section X for a discussion on how to assign nondeterministic values to types that `kani::any()` is not implemented for)

docs/src/tutorial-nondeterministic-variables.md
zhassan-aws: I agree this section is a bit confusing. Not sure what it's getting at.

docs/src/tutorial-nondeterministic-variables.md
zhassan-aws: Why does the lack of `Option<MyType>` affect the `Arbitrary` type story though?

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,105 @@
...
+```rust
+{{#include tutorial/arbitrary-variables/src/rating.rs:rating_invariant}}
 ```
 
-## Safe nondeterministic variables for custom types
+All we're doing here is making use of a nondeterministic integer to decide which variant of `Rating` to return.
 
-Now you would like to add a new structure to your library that allow users to represent a review rating, which can go from 0 to 5 stars.
-Let's say you add the following implementation:
+> **NOTE**: If we thought of this code as generating a random value, this function looks sub-optimal.
```
zhassan-aws, line 74:
```suggestion
> **NOTE**: If we thought of this code as generating a random value, this function looks heavily biased.
```

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,105 @@
...
+> We'd overwhelming generate a `Three` because it's matching "all other integers besides 1 and 2."
```
zhassan-aws, line 75:
```suggestion
> We'd overwhelmingly generate a `Three` because it's matching "all other integers besides 1 and 2."
```

docs/src/tutorial/arbitrary-variables/src/rating.rs
```diff
@@ -3,40 +3,42 @@
...
 #[cfg(kani)]
 mod verification {
     use super::*;
 
     // ANCHOR: rating_invariant
-    unsafe impl kani::Invariant for Rating {
-        fn is_valid(&self) -> bool {
-            self.value <= 5
+    fn any_rating() -> Rating {
+        match kani::any() {
```
zhassan-aws, line 29: I think `kani::any::<u8>()` should be easier for the solver.

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,105 @@
...
+> But Kani just see 3 meaningful possibilities, each of which is not treated any differently from each other.
+> The "proportion" of integers does not matter.
```
zhassan-aws, line 77: Nice note!

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,105 @@
...
-```rust
-{{#include tutorial/arbitrary-variables/src/rating.rs:rating_struct}}
-```
 
-The easiest way to allow users to create nondeterministic variables of the Rating type which represents values from 0-5 stars is by implementing the `kani::Invariant` trait.
+## Exercise
 
-The implementation only requires you to define a check to your structure that returns whether its current value is valid or not.
-In our case, we have the following implementation:
+Try writing a function to generate a nondeterministic inventory (from the first example:)
```
zhassan-aws, line 82:
```suggestion
Try writing a function to generate a nondeterministic inventory of a bounded size (from the first example:)
```
?

docs/src/tutorial-nondeterministic-variables.md
celinval: https://github.com/model-checking/kani/pull/1401

docs/src/tutorial/arbitrary-variables/src/rating.rs
celinval: I personally prefer minimizing the things we have to explain in a tutorial but not a blocker.

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,114 @@
 # Nondeterministic variables
 
-Kani is able to reason about programs and their execution paths by allowing users to assign nondeterministic (i.e., symbolic) values to  certain variables.
-Since Kani is a bit-level model checker, this means that Kani considers that an unconstrained nondeterministic value represents all the possible bit-value combinations assigned to the variable's memory contents.
+Kani is able to reason about programs and their execution paths by allowing users to assign nondeterministic (also called symbolic) values to certain variables using `kani::any()`.
+Kani is a "bit-precise" model checker, which means that Kani considers all the possible bit-value combinations that could be assigned to the variable's memory contents.
```
celinval, line 4: what about saying 
> ... possible bit-value combinations that would be valid if assigned to the variable's memory contents. In other words, `kani::any()` should not produce values that are invalid for the type.

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,114 @@
...
-As with any unsafe method in Rust, users have to make sure the right guardrails are put in place to avoid undesirable behavior.
+## Bounding nondeterministic variables
 
-That said, there may be cases where you want to verify your code taking into consideration that some inputs may contain invalid data.
+There is no implementation of `kani::any()` for more complex data structures like `Vec`, however.
+The trouble with a nondeterministic vector is that you usually need to _bound_ the size of the vector, for the reasons we investigated in the [last chapter](./tutorial-loop-unwinding.md).
+The `kani::any()` function does not have any arguments, and so cannot be given an upper bound.
 
-Let's see what happens if we modify our verification harness to use the unsafe method `kani::any_raw()` to generate the updated value.
+You can use `kani::any()` for `[T; N]` (if implemented for `T`).
```
celinval, line 50: This line doesn't make much sense here.

What if you move talk about custom non-deterministic types first then you introduce this section. I think it is easier to first present the solution and discuss its limitation after.

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,114 @@
...
+It just means you have to construct one.
+Our example proof harness above constructs a nondeterministic `Inventory` of size `1`, simply by starting with the empty `Inventory` and inserting a nondeterministic entry.
 
-We commented out the assertion that the underlying `u32` variable cannot be `0`, since this no longer holds.
-The verification will now fail showing that the `inventory.get(&id).unwrap()` method call can panic.
+## Custom nondeterministic types
 
-This is an interesting issue that emerges from how `rustc` optimizes the memory layout of `Option<NonZeroU32>`.
-The compiler is able to represent `Option<NonZeroU32>` using `32` bits by using the value `0` to represent `None`.
+There's nothing special about `kani::any()` except that Kani ships with implementations for a few types where we can guarantee safety.
```
celinval, line 60: maybe rephrase this to something like:
> `kani::any()` is the only method that Kani provides to inject non-determinism in a harness, but Kani ships with implementations for just a few types where it can guarantee safety.

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,114 @@
...
+2. Just write a function.
 
-```
-cargo kani --harness unsafe_update
-```
+The advantage of the first approach is that it's simple and conventional.
+It also means that in addition to being able to use `kani::any()` with your type, you can also use it with `Option<MyType>` (for example).
 
-## Safe nondeterministic variables for custom types
+The advantage of the second approach is that you're able to pass in parameters, like bounds on the size of the data structure.
```
celinval, line 69: Here you mention the parameter, so you could just replace the previous section by giving a bit more details on the second approach. Maybe add an example.

docs/src/tutorial-nondeterministic-variables.md
```diff
@@ -1,95 +1,112 @@
...
 
-That said, there may be cases where you want to verify your code taking into consideration that some inputs may contain invalid data.
+The advantage of the second approach is that you're able to pass in parameters, like bounds on the size of the data structure.
+(Which we'll discuss more in the next section.)
+This approach is also necessary when you are unable to implement a trait (like `Arbitrary`) on a type you're importing from another crate.
 
-Let's see what happens if we modify our verification harness to use the unsafe method `kani::any_raw()` to generate the updated value.
+Either way, inside this function you would simply return an arbitrary value by generating arbitrary values for its components.
+To generate a nondeterministic struct, you would just generate nondeterministic values for each of its fields.
+For complex data structures, you can start with an empty one and add a (bounded) nondeterministic number of entries.
```
celinval, line 59: Do you mean for containers?

celinval (approved):
I like it! I think it reads much better now and it doesn't have any UB stuff. Thanks!

PR #1354, opened 2022-07-08
Fix compiletest: don't abbreviate the output

fzaiser:
### Description of changes: 

This just cost me three hours of my life. Apparently, for reasons that are beyond me, compiletest "abbreviates" the output, i.e. it removes the middle part. This causes problems with big tests (such as `tests/expected/trait-receiver/receiver_types.rs`), which fail nondeterministically, depending on whether the expected part of the output falls into the skipped middle part.

In this PR, I change it such that compiletest always looks at the whole output.

### Testing:

* How is this change tested? I came across this problem in #1310, so it will be tested by that one, once it's merged.

### Checklist


tools/compiletest/src/read2.rs
```diff
@@ -6,77 +6,29 @@
 // FIXME: This is a complete copy of `cargo/src/cargo/util/read2.rs`
```
sanjit-bhat, line 6: I looked at the latest Cargo util directory, https://github.com/rust-lang/cargo/tree/master/src/cargo/util, and there doesn't seem to be any file named `read2.rs` anymore. Is this copied code outdated? Should we consider updating it?

sanjit-bhat (changes requested):
All changes LGTM. I've tested them locally with my PR, #1356, and they cause the regression to pass. I also manually changed one of the expected files to add gibberish, and as expected, that test fails.

tools/compiletest/src/read2.rs
```diff
@@ -6,77 +6,29 @@
...
 // Consider unify the read2() in libstd, cargo and this to prevent further code duplication.
 
-pub use self::imp::read2;
+// Note: read2_abbreviated was replaced by read2 because the original function skipped some of the output,
+// which caused issues with our tests (in particular tests with large outputs).
```
adpaco-aws, line 10: Appreciate adding this comment, but if you remove `read2_abbreviated` it's surely going to confuse whoever works on this file in the future. Three ways to approach this:
 * Remove `read2_abbreviated` and don't add the comment.
 * Keep `read2_abbreviated` and add the comment for comparison.
 * Remove `read2_abbreviated` and add a comment in new line 90 in `tools/compiletest/src/runtest.rs` that refers to it (explaining it shortened the output) and explain that this is not desirable behavior, so you just the full output instead.

adpaco-aws (changes requested):
Looks good but had some notes about the comments.

tools/compiletest/src/read2.rs
fzaiser: I decided to remove `read2_abbreviated` and the comment.

tools/compiletest/src/read2.rs
fzaiser: I removed the comment.

adpaco-aws (approved):

sanjit-bhat (approved):

PR #1356, opened 2022-07-09
feat: make byte arr the lowest level non-det interface

sanjit-bhat:
### Description of changes: 

This PR is a self-contained chunk of the executable trace feature. Since the rest of the executable trace feature requires the full UX to be working, I will save those PRs for later.

This PR modifies `any_raw` to get its non-deterministic bytes from the function `any_raw_inner`. This inner function returns only constant-length (parameterized by a const type) non-det byte arrays, which makes serializing / deserializing non-det variables easier.

#### Increased number of checks

As a result of using `transmute_copy` inside `any_raw`, this change introduces more (useless) checks that Kani has to verify. In particular, the expected output for the `expected/one-assert` test grew from 2 to 16 checks. Only checks 1 and 5 were present before (see output below). I didn't run a performance analysis, but it looks like even with these checks, all the existing regressions are running in time. I tried increasing the number of non-det vars in `expected/one-assert`, and the number of additional checks stays the same, so I think they're only being added once per harness.

If we didn't use this PR, we would only have to use `transmute_copy` in the alternate implementation of `any_raw_inner` that sticks concrete values into a type. Since that implementation isn't codegen'd for verification, it would get rid of these checks. However, this approach would significantly increase the difficulty of parsing  concrete values from the CBMC trace.

```
RESULTS:
Check 1: check_assert.assertion.1
	 - Status: SUCCESS
	 - Description: "assertion failed: x == y"
	 - Location: tests/expected/one-assert/test.rs:8:5 in function check_assert

Check 2: std::mem::transmute_copy.safety_check.1
	 - Status: SUCCESS
	 - Description: "`src` must be properly aligned"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:1107:9 in function std::mem::transmute_copy

Check 3: std::mem::transmute_copy.safety_check.2
	 - Status: SUCCESS
	 - Description: "`dst` must be properly aligned"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:1107:9 in function std::mem::transmute_copy

Check 4: std::mem::transmute_copy.arithmetic_overflow.1
	 - Status: SUCCESS
	 - Description: "copy_nonoverlapping: attempt to compute number in bytes which would overflow"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:1107:9 in function std::mem::transmute_copy

Check 5: <T as kani::Arbitrary>::any.unsupported_construct.1
	 - Status: SUCCESS
	 - Description: "resume instruction"
	 - Location: library/kani/src/arbitrary.rs:18:5 in function <T as kani::Arbitrary>::any

Check 6: std::mem::transmute_copy.pointer_dereference.1
	 - Status: SUCCESS
	 - Description: "dereference failure: pointer NULL"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/mem/mod.rs:1047:18 in function std::mem::transmute_copy

Check 7: std::mem::transmute_copy.pointer_dereference.2
	 - Status: SUCCESS
	 - Description: "dereference failure: pointer invalid"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/mem/mod.rs:1047:18 in function std::mem::transmute_copy

Check 8: std::mem::transmute_copy.pointer_dereference.3
	 - Status: SUCCESS
	 - Description: "dereference failure: deallocated dynamic object"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/mem/mod.rs:1047:18 in function std::mem::transmute_copy

Check 9: std::mem::transmute_copy.pointer_dereference.4
	 - Status: SUCCESS
	 - Description: "dereference failure: dead object"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/mem/mod.rs:1047:18 in function std::mem::transmute_copy

Check 10: std::mem::transmute_copy.pointer_dereference.5
	 - Status: SUCCESS
	 - Description: "dereference failure: pointer outside object bounds"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/mem/mod.rs:1047:18 in function std::mem::transmute_copy

Check 11: std::mem::transmute_copy.pointer_dereference.6
	 - Status: SUCCESS
	 - Description: "dereference failure: invalid integer address"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/mem/mod.rs:1047:18 in function std::mem::transmute_copy

Check 12: memcpy.pointer.1
	 - Status: SUCCESS
	 - Description: "same object violation"
	 - Location: <builtin-library-memcpy>:33 in function memcpy

Check 13: memcpy.pointer.2
	 - Status: SUCCESS
	 - Description: "same object violation"
	 - Location: <builtin-library-memcpy>:34 in function memcpy

Check 14: std::mem::transmute_copy.precondition_instance.1
	 - Status: SUCCESS
	 - Description: "memcpy src/dst overlap"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:1107:9 in function std::mem::transmute_copy

Check 15: std::mem::transmute_copy.precondition_instance.2
	 - Status: SUCCESS
	 - Description: "memcpy source region readable"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:1107:9 in function std::mem::transmute_copy

Check 16: std::mem::transmute_copy.precondition_instance.3
	 - Status: SUCCESS
	 - Description: "memcpy destination region writeable"
	 - Location: ~/.rustup/toolchains/nightly-2022-07-05-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:1107:9 in function std::mem::transmute_copy
```

#### Weird bug with `expected/trait-receiver` test (@celinval)

When running this test, `compiletest` reports that the line `Checking harness check_arc...` is not present in the output. Indeed, it isn't present when `compiletest` runs that test, but when I run it manually, I see that line, along with all the other lines for the `check_arc` harness. I'm confused why I'm seeing this behavior.

Note: see PR comment that addresses this bug.

### Testing:

* How is this change tested?

There aren't any explicit tests here, but we assume that if the existing tests pass, then the change to using non-det byte arrays didn't fundamentally change the semantics of generating non-det types.

### Checklist

### DIFF
--- a/library/kani/src/arbitrary.rs
+++ b/library/kani/src/arbitrary.rs
@@ -13,6 +13,7 @@ pub trait Arbitrary {
 impl<T> Arbitrary for T
 where
     T: Invariant,
+    [(); std::mem::size_of::<T>()]:,
 {
     default fn any() -> Self {
         let value = unsafe { any_raw::<T>() };
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 #![feature(rustc_attrs)] // Used for rustc_diagnostic_item.
 #![feature(min_specialization)] // Used for default implementation of Arbitrary.
+#![feature(generic_const_exprs)] // Used for getting size_of generic types
 
 pub mod arbitrary;
 pub mod invariant;
@@ -95,10 +96,26 @@ pub fn any<T: Arbitrary>() -> T {
 /// kani::assume(char::from_u32(c as u32).is_ok());
 /// ```
 ///
+#[inline(never)]
+pub unsafe fn any_raw<T>() -> T
+where
+    // This where bound guarantees constant `size_of::<T>()` so we can use `size_of::<T>()`
+    // as a const generic argument and as a compile-time constant array length.
+    [(); std::mem::size_of::<T>()]:,
+{
+    let non_det_byte_arr = any_raw_inner::<{ std::mem::size_of::<T>() }>();
+    // We need `transmute_copy` instead of `transmute` because right now, rustc can't guarantee that the
+    // source and destination types are the same size, even though they are.
+    let non_det_var =
+        std::mem::transmute_copy::<[u8; std::mem::size_of::<T>()], T>(&non_det_byte_arr);
+    non_det_var
+}
+
+/// This function creates an unconstrained byte array of length `T`.
 #[rustc_diagnostic_item = "KaniAnyRaw"]
 #[inline(never)]
-pub unsafe fn any_raw<T>() -> T {
-    unimplemented!("Kani any_raw")
+unsafe fn any_raw_inner<const T: usize>() -> [u8; T] {
+    unimplemented!("Kani any_raw_inner")
 }
 
 /// This function has been split into a safe and unsafe functions: `kani::any` and `kani::any_raw`.
--- a/library/kani/src/slice.rs
+++ b/library/kani/src/slice.rs
@@ -79,7 +79,10 @@ impl<T, const MAX_SLICE_LENGTH: usize> AnySlice<T, MAX_SLICE_LENGTH> {
         any_slice
     }
 
-    fn new_raw() -> Self {
+    fn new_raw() -> Self
+    where
+        [(); std::mem::size_of::<T>()]:,
+    {
         let any_slice = AnySlice::<T, MAX_SLICE_LENGTH>::alloc_slice();
         unsafe {
             let mut i = 0;
@@ -149,6 +152,9 @@ where
     AnySlice::<T, MAX_SLICE_LENGTH>::new()
 }
 
-pub unsafe fn any_raw_slice<T, const MAX_SLICE_LENGTH: usize>() -> AnySlice<T, MAX_SLICE_LENGTH> {
+pub unsafe fn any_raw_slice<T, const MAX_SLICE_LENGTH: usize>() -> AnySlice<T, MAX_SLICE_LENGTH>
+where
+    [(); std::mem::size_of::<T>()]:,
+{
     AnySlice::<T, MAX_SLICE_LENGTH>::new_raw()
 }
--- a/library/kani/src/vec.rs
+++ b/library/kani/src/vec.rs
@@ -6,6 +6,7 @@ use crate::{any, assume, Invariant};
 pub fn any_vec<T, const MAX_LENGTH: usize>() -> Vec<T>
 where
     T: Invariant,
+    [(); std::mem::size_of::<[T; MAX_LENGTH]>()]:,
 {
     let mut v = exact_vec::<T, MAX_LENGTH>();
     let real_length: usize = any();
@@ -19,6 +20,7 @@ where
 pub fn exact_vec<T, const EXACT_LENGTH: usize>() -> Vec<T>
 where
     T: Invariant,
+    [(); std::mem::size_of::<[T; EXACT_LENGTH]>()]:,
 {
     let boxed_array: Box<[T; EXACT_LENGTH]> = Box::new(any());
     <[T]>::into_vec(boxed_array)
--- a/tests/cargo-kani/firecracker-block-example/src/main.rs
+++ b/tests/cargo-kani/firecracker-block-example/src/main.rs
@@ -4,6 +4,7 @@
 
 #![allow(dead_code)]
 #![allow(unused_variables)]
+#![feature(generic_const_exprs)]
 
 mod descriptor_permission_checker;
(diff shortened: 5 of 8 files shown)


sanjit-bhat:
Looks like the issue with `expected/trait-receiver` might be related to #1354.

library/kani/src/arbitrary.rs
```diff
@@ -13,6 +13,7 @@ pub trait Arbitrary {
 impl<T> Arbitrary for T
 where
     T: Invariant,
+    [(); std::mem::size_of::<T>()]:,
```
fzaiser, line 16: Add an explanatory comment here as well? SImilar to the comment in `any_raw`.

library/kani/src/lib.rs
```diff
@@ -95,10 +96,26 @@ pub fn any<T: Arbitrary>() -> T {
...
+where
+    // This where bound guarantees constant `size_of::<T>()` so we can use `size_of::<T>()`
+    // as a const generic argument and as a compile-time constant array length.
+    [(); std::mem::size_of::<T>()]:,
+{
+    let non_det_byte_arr = any_raw_inner::<{ std::mem::size_of::<T>() }>();
+    // We need `transmute_copy` instead of `transmute` because right now, rustc can't guarantee that the
+    // source and destination types are the same size, even though they are.
+    let non_det_var =
+        std::mem::transmute_copy::<[u8; std::mem::size_of::<T>()], T>(&non_det_byte_arr);
```
fzaiser, line 110: Could there be issues with alignment here? If I understand correctly, the alignment of an array of `u8` is the same as `u8`. So could there be an issue if we transmute to a type (say `u64`) with a bigger alignment?

library/kani/src/lib.rs
```diff
@@ -95,10 +96,26 @@ pub fn any<T: Arbitrary>() -> T {
 /// kani::assume(char::from_u32(c as u32).is_ok());
 /// ```
 ///
+#[inline(never)]
+pub unsafe fn any_raw<T>() -> T
+where
+    // This where bound guarantees constant `size_of::<T>()` so we can use `size_of::<T>()`
+    // as a const generic argument and as a compile-time constant array length.
+    [(); std::mem::size_of::<T>()]:,
```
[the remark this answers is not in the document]
celinval: The generic constant expressions is still fairly unstable so some of these deductions might've not been implemented yet (or maybe they might skip them all together). I think [this question](https://github.com/rust-lang/rust/issues/76560#issuecomment-778804018 ) in the tracking issue is relevant.

The answer to the question is not clear to me though. We could ask this on the [project's zulip channel](https://rust-lang.zulipchat.com/#narrow/stream/260443-project-const-generics) if the project plans to address cases like that or why this is needed in the first place.

danielsn:
I'm worried this may cause performance regressions, since we're introducing extra byte-extracts for CBMC.

celinval:
> I'm worried this may cause performance regressions, since we're introducing extra byte-extracts for CBMC.

I agree that this could be the case. Any suggestion on how to measure this? We can try to provide different solutions, including a more convoluted parser, but this is the cleanest solution for the problem at hand.

library/kani/src/lib.rs
sanjit-bhat: According to https://doc.rust-lang.org/stable/std/mem/fn.transmute.html: 
> Because transmute is a by-value operation, alignment of the transmuted values themselves is not a concern. As with any other function, the compiler already ensures both T and U are properly aligned. However, when transmuting values that point elsewhere (such as pointers, references, boxes…), the caller has to ensure proper alignment of the pointed-to values.

Since both types don't have any pointers, I think we should be fine.

library/kani/src/arbitrary.rs
sanjit-bhat: Done

PR #1378, opened 2022-07-14
Implement support for generators

### LINKED ISSUE #416: Support Rust's `Generator` feature

fzaiser:
### Description of changes: 

This implements the generator feature for Rust. We translate a generator type similarly to an enum with a variant for each suspend point.

Consider the following generator:
```rust
|| {
    let a = true;
    let b = &a;
    yield;
    assert_eq!(b as *const _, &a as *const _);
    yield;
};
```

Rustc compiles this to something similar to the following enum (but there are differences, see below!):

```rust
enum GeneratorEnum {
    Unresumed,                        // initial state of the generator
    Returned,                         // generator has returned
    Panicked,                         // generator has panicked
    Suspend0 { b: &bool, a: bool },   // state after suspending (`yield`ing) for the first time
    Suspend1,                         // state after suspending (`yield`ing) for the second time
}
```

However, its layout may differ from normal Rust enums in the following ways:
* Contrary to enums, the discriminant may not be at offset 0.
* Contrary to enums, there may be other fields than the discriminant "at the top level" (outside the variants).

This means that we CANNOT use the enum translation, which would be roughly as follows:

```c
struct GeneratorEnum {
    int case; // discriminant
    union GeneratorEnum-union cases; // variant
}

union GeneratorEnum-union {
    struct Unresumed variant0;
    struct Returned variant1;
    // ...
}
```

Instead, we use the following translation:

```c
union GeneratorEnum {
    struct DirectFields direct_fields;
    struct Unresumed generator_variant_Unresumed;
    struct Returned generator_variant_Returned;
    // ...
}

struct DirectFields {
    // padding (for bool *b in Suspend0 below)
    char case;
    // padding (for bool a in Suspend0 below)
}

struct Unresumed {
    // padding (this variant has no fields)
}

// ...

struct Suspend0 {
    bool *generator_field_0; // variable b in the generator code above
    // padding (for char case in DirectFields)
    bool generator_field_1; // variable a in the generator code above
}
```

Of course, if the generator has any other top-level/direct fields, they'd be included in the `DirectFields` struct as well.

### Resolved issues:

Resolves #416 

### Call-outs:

* The function `ignore_var_ty` became redundant, so I removed it.
* There is some code duplication with `codegen_enum`. I'm not sure how to remove it because `codegen_enum` relies on an `AdtDef`, which is not available for generators (because the relevant enum is generated in MIR).
* I added a couple of assertions in `codegen_enum` that document assumptions that we currently make about enum layouts. (I don't know if they're guaranteed for enums, but they don't hold for generators, so I thought it would be good to be explicit about it.)
* One of the new tests (`discriminant.rs`, taken from rustc) takes a bit over a minute on my laptop. Is that too much?
    - Discussed with @danielsn, decision: leave it in, but move slow tests to a separate test suite in a follow-up PR

### Testing:

* How is this change tested? 3 existing regression tests (that now work) + 16 test cases from rustc (`src/test/ui/generator/`)

### Checklist


celinval (dismissed):
Can you please add comments to your code and maybe a high level explanation of how you modeled generators in the PR body?

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -360,9 +360,7 @@ impl<'tcx> GotocCtx<'tcx> {
         args.iter()
             .filter_map(|o| {
```
danielsn, line 361: Is `filter_map` still needed? Looks like both arms return `Some`

fzaiser: Thanks, fixed.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,184 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                {
+                    fields.push(padding);
+                }
+                fields.push(DatatypeComponent::Field {
+                    name: field_name,
+                    typ: ctx.codegen_ty(field_ty),
+                });
+                offset += field_size;
+            }
+            // TODO: is this necessary? I don't see how padding at the end matters. But codegen_struct_fields does it too.
```
fzaiser, line 999: @danielsn @celinval Do you know if/why this part of the code is necessary? I copied it from `codegen_struct_fields`. Intuitively, it seems superfluous to me.

fzaiser: I did some experiments: if i remove it from the generator code, all tests still pass. (But I don't know whether the tests just don't cover the necessary case?) If I remove it from `codegen_struct_fields`, tests fail.

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -653,8 +651,6 @@ impl<'tcx> GotocCtx<'tcx> {
                 let layout = self.layout_of(dst_mir_ty);
                 if layout.is_zst() || dst_type.sizeof_in_bits(&self.symbol_table) == 0 {
                     // We ignore assignment for all zero size types
-                    // Ignore generators too for now:
-                    // https://github.com/model-checking/kani/issues/416
```
fzaiser, line 657: I believe this comment was outdated because the if condition only checks for ZST.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: Just discussed with @danielsn. The code is important to get the right sizeof. I'll remove the TODO.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -197,27 +200,62 @@ impl<'tcx> GotocCtx<'tcx> {
...
-            }
-            ty::Generator(_def_id, _substs, _movability) => {
-                let error_msg = GotocCtx::unsupported_msg(
-                    "The `generators` feature",
-                    Some("https://github.com/model-checking/kani/issues/416"),
-                );
-                self.emit_error_and_exit(&error_msg)
+                sig
             }
+            ty::Generator(_def_id, substs, _movability) => self.generator_sig(fntyp, substs),
```
danielsn, line 258: What is `movability`?

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,181 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    assert!(matches!(tag_encoding, TagEncoding::Direct));
+                    (tag_field, variants)
+                }
+                Variants::Single { .. } => {
+                    unreachable!("generators cannot have a single variant")
+                }
+            };
+            let mut fields = vec![];
+            // generate a struct for the direct fields of the layout (fields that don't occur in the variants)
+            fields.push(DatatypeComponent::Field {
```
danielsn, line 935: Can combine this by putting the first field into the vec

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,181 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                name: "direct_fields".into(),
```
danielsn, line 936: Magic string

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,181 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        pretty_name,
+                        type_and_layout.for_variant(ctx, var_idx),
+                        variant_name,
+                        None,
+                    ),
+                });
+            }
+            fields
+        });
+        generator_enum
```
danielsn, line 960: Why store then immediately return it?

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,181 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        discriminant_field: Option<usize>,
+    ) -> Type {
+        let struct_name = format!("{generator_name}::{variant_name}");
+        let pretty_struct_name = format!("{pretty_generator_name}::{variant_name}");
+        debug!("codegen generator variant struct {}", pretty_struct_name);
+        self.ensure_struct(struct_name, pretty_struct_name, |ctx, _| {
+            let mut offset = Size::ZERO;
+            let mut fields = vec![];
+            for idx in type_and_layout.fields.index_by_increasing_offset() {
+                let field_name = if Some(idx) == discriminant_field {
```
danielsn, line 978: Explain what's going on here

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,181 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    /// || {
+    ///     let a = true;
+    ///     let b = &a;
+    ///     yield;
+    ///     assert_eq!(b as *const _, &a as *const _);
+    ///     yield;
+    /// };
+    /// ```
+    ///
+    /// Rustc compiles this to something similar to the following enum (but there are differences, see below!):
```
celinval, line 858: Can you please add a reference for this?

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,181 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    ///
+    /// struct Suspend0 {
+    ///     bool *generator_field_0; // variable b in the generator code above
+    ///     // padding (for char case in DirectFields)
+    ///     bool generator_field_1; // variable a in the generator code above
+    /// }
+    /// ```
+    ///
+    /// Of course, if the generator has any other top-level/direct fields, they'd be included in the `DirectFields` struct as well.
+    fn codegen_ty_generator(&mut self, ty: Ty<'tcx>) -> Type {
```
celinval, line 919: Please add a debug log in this function.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,181 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        &mut self,
+        generator_name: InternedString,
+        pretty_generator_name: InternedString,
+        type_and_layout: TyAndLayout<'tcx, Ty<'tcx>>,
+        variant_name: InternedString,
+        discriminant_field: Option<usize>,
+    ) -> Type {
+        let struct_name = format!("{generator_name}::{variant_name}");
+        let pretty_struct_name = format!("{pretty_generator_name}::{variant_name}");
+        debug!("codegen generator variant struct {}", pretty_struct_name);
```
celinval, line 973: `debug!(?pretty_struct_name, "codeged_generator_variant_struct")`

celinval (commented):
This is my first pass. I'll keep taking a look at it tomorrow. Thanks!

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -269,12 +276,12 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                    ty::Generator(..) => Err(UnimplementedData::new(
-                        "ty::Generator",
-                        "https://github.com/model-checking/kani/issues/416",
-                        Type::code(vec![], Type::empty()),
-                        *res.location(),
-                    )),
+                    ty::Generator(..) => {
+                        let field_name = self.generator_field_name(f.index().into());
+                        Ok(res
+                            .member("direct_fields", &self.symbol_table)
```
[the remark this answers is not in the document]
fzaiser: My impression was that we don't usually use constants for magic strings. There's a lot of them in the existing code: "case", "cases", "data", "len", "vtable" etc. I'm happy to use a constant instead, but then we should make things consistent IMO. What do you think?

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -358,14 +358,12 @@ impl<'tcx> GotocCtx<'tcx> {
...
     pub fn codegen_funcall_args(&mut self, args: &[Operand<'tcx>]) -> Vec<Expr> {
         args.iter()
-            .filter_map(|o| {
+            .map(|o| {
                 let ot = self.operand_ty(o);
-                if self.ignore_var_ty(ot) {
-                    None
-                } else if ot.is_bool() {
-                    Some(self.codegen_operand(o).cast_to(Type::c_bool()))
+                if ot.is_bool() {
```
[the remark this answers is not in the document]
fzaiser: We don't, good point.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: Done

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: Good point, changed.

tests/expected/generators/as_parameter/main.rs
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 // Test that includes a generator as a parameter to a function
-// Codegen should succeed, but verification should fail (codegen_unimplemented)
```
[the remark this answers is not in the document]
fzaiser: That a generator can be passed as a parameter. It was there before, so I thought I should keep it and just updated the comment.

tests/kani/Generator/main.rs
```diff
@@ -2,14 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 // Check that we can codegen code that has a Generator type present,
-// as long as the path is not dynamically used. Full Generator support
-// tracked in: https://github.com/model-checking/kani/issues/416
-
+// when the path is not dynamically used.
```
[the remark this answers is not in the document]
fzaiser: I don't think this test is really necessary anymore, but I'm hesitant to remove tests. It used to test that a file containing generators worked as long as generators are not actually used in the execution of the program. Should I remove it?

tests/kani/Generator/rustc-generator-tests/conditional-drop.rs
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// Modifications Copyright Kani Contributors
+// See GitHub history for details.
+
+// Copyright rustc Contributors
+// Adapted from rustc: src/test/ui/generator/conditional-drop.rs
```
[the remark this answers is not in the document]
fzaiser: Added links to the github repository.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -498,33 +509,40 @@ impl<'tcx> GotocCtx<'tcx> {
...
                     }
-                    _ => unreachable!("it's a bug to reach here!"),
-                }
+                    ty::Generator(..) => {
+                        (self.generator_variant_name(idx), TypeOrVariant::GeneratorVariant(idx))
+                    }
+                    _ => unreachable!("it's a bug to reach here! {:?}", &t.kind()),
+                };
+                let layout = self.layout_of(t);
+                let expr = match &layout.variants {
```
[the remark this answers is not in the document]
fzaiser: Yes, that is correct. I'll document this as assertions.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: Done.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: Changed.

tests/expected/generators/as_parameter/main.rs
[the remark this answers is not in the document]
celinval: You could just add an assert at the end to make sure it works.

tests/kani/Generator/main.rs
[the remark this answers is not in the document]
celinval: Either remove it or add a check at the end

tests/kani/Generator/rustc-generator-tests/discriminant.rs
```diff
@@ -0,0 +1,147 @@
...
+                assert_eq!(max_discr, expected_max_discr);
+                return;
+            }
+        }
+    }
+}
+
+#[kani::proof]
+#[kani::unwind(260)]
+fn main() {
```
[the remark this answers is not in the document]
celinval: Or just say that. :)

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -129,7 +130,7 @@ impl<'tcx> ProjectedPlace<'tcx> {
                 }
             }
             // TODO: handle Variant https://github.com/model-checking/kani/issues/448
-            TypeOrVariant::Variant(_) => None,
+            TypeOrVariant::Variant(_) | TypeOrVariant::GeneratorVariant(_) => None,
```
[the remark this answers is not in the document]
fzaiser: Done

tests/expected/generators/as_parameter/main.rs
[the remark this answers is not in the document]
fzaiser: Done

tests/kani/Generator/main.rs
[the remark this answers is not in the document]
fzaiser: I've changed the test completely. It now checks whether generators work with resume types that are not `()`. I don't think the original test makes sense anymore, now that generators are properly implemented.

tests/kani/Generator/rustc-generator-tests/discriminant.rs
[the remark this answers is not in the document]
fzaiser: Done

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -498,33 +509,44 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                        )
+                        let variant = def.variant(idx);
+                        (variant.name.as_str().into(), TypeOrVariant::Variant(variant))
                     }
-                    _ => unreachable!("it's a bug to reach here!"),
-                }
+                    ty::Generator(..) => {
+                        (self.generator_variant_name(idx), TypeOrVariant::GeneratorVariant(idx))
+                    }
+                    _ => unreachable!("it's a bug to reach here! {:?}", &t.kind()),
```
danielsn, line 520: ⛏️ error message could be more informative

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -498,33 +509,44 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                };
+                let layout = self.layout_of(t);
+                let expr = match &layout.variants {
+                    Variants::Single { .. } => {
+                        assert!(!t.is_generator());
```
danielsn, line 525: Might be clearer to lift these to a precondition
```rust
if t.is_generator {
    assert!(matches!(layout.variants, Variants::Single { .. });
    etc ...
}
```

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -510,6 +510,11 @@ impl<'tcx> GotocCtx<'tcx> {
             }
             Variants::Multiple { tag, tag_encoding, .. } => match tag_encoding {
                 TagEncoding::Direct => {
+                    let e = if ty.is_generator() {
```
danielsn, line 513: Explain what's going on here

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -699,13 +681,19 @@ impl<'tcx> GotocCtx<'tcx> {
...
                             )
-                            .goto_expr
-                            .member("case", &self.symbol_table)
-                            .assign(discr, location)
+                            .goto_expr;
+                            let place_goto_expr = if pt.is_generator() {
+                                place_goto_expr.member("direct_fields", &self.symbol_table)
+                            } else {
+                                place_goto_expr
+                            };
```
danielsn, line 693: This happens more than once. Utility fn?

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,186 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        debug!(?pretty_struct_name, "codeged_generator_variant_struct");
+        self.ensure_struct(struct_name, pretty_struct_name, |ctx, _| {
+            let mut offset = Size::ZERO;
+            let mut fields = vec![];
+            for idx in type_and_layout.fields.index_by_increasing_offset() {
+                // The discriminant field needs to have a special name to work with the rest of the code:
+                let field_name = if Some(idx) == discriminant_field {
+                    "case".into()
+                } else {
+                    ctx.generator_field_name(idx).into()
```
danielsn, line 986: What does the `None` case mean

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,186 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    ),
+                });
+            }
+            fields
+        })
+    }
+
+    /// Generates a struct for a variant of the generator.
+    ///
+    /// The parameter `discriminant_field` may contain the index of a field that is the discriminant.
```
danielsn, line 966: If it doesn't, what then?

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -804,13 +842,186 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            if overhang != 0 {
+                fields.push(
+                    ctx.codegen_struct_padding(
+                        offset.bits(),
+                        offset.bits() + align - overhang,
+                        type_and_layout.fields.count(),
+                    )
+                    .unwrap(),
+                )
+            }
```
danielsn, line 1014: Merge this with the enum code as a utility fn

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -1114,7 +1303,11 @@ impl<'tcx> GotocCtx<'tcx> {
                         }
                     }
                 }
-                Variants::Multiple { tag_encoding, variants, .. } => {
+                Variants::Multiple { tag_encoding, variants, tag_field, .. } => {
+                    // Contrary to generators, currently enums have only one field (the discriminant), the rest are in the variants:
+                    assert!(layout.fields.count() <= 1);
+                    // Contrary to generators, the discriminant is the first (and only) field for enums:
+                    assert_eq!(*tag_field, 0);
```
danielsn, line 1310: What happens in a Niche encoding

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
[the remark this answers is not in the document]
fzaiser: Opened an issue to refactor this in a follow-up PR: #1397

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -510,6 +510,11 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        e.member("direct_fields", &&self.symbol_table)
```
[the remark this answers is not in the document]
fzaiser: see above #1397

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: see above #1397

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
fzaiser: Improved

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
fzaiser: Created a function `check_generator_layout_assumptions` because this code was repeated in a few places.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
fzaiser: Added a comment (in the factored out function).

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
fzaiser: Factored out into a utility function.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: Added a comment

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: Added an explanation

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: Done

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: It seems that they have a `tag_field` anyway: https://doc.rust-lang.org/stable/nightly-rustc/rustc_target/abi/enum.Variants.html#variant.Multiple.field.tag_field. So this assertion should not be a problem (and the regression tests pass).

fzaiser:
Regarding the assertions checking that the layout of a generator is multiple variants, with direct encoding: I created a function `check_generator_layout_assumptions` because there were many places where you suggested we check these assumptions. But thinking about it more, I think it's enough to check it in once, when the generator is created, i.e. in `codegen_ty_generator`. The later checks seem redundant. Do you agree and can I remove them?

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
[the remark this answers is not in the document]
celinval: We definitely do a poor job at that today (thanks for creating an issue). That said, I would prefer if we avoid adding more of those to the code.

celinval (changes requested):
Regarding the issue #1395, can you please inject a unimplemented failure if the user tried to get the size of a generator? Silently failing might result in an unsound analysis.

fzaiser:
> Regarding the issue https://github.com/model-checking/kani/issues/1395, can you please inject a unimplemented failure if the user tried to get the size of a generator? Silently failing might result in an unsound analysis.

I first thought this was a good idea and I just tried it, but it turns out that 5 of the other tests rely on `size_of`, so we'd have to deactivate those as well, which I think does more harm than good. The WASM backend has similarly decided that it's fine to disable that one test where the size differs. (Why LLVM computes a different size is still a mystery to me. All the code I can find in rustc's codegen uses `layout.size.bytes()`, which is exactly what we're doing.)

tests/kani/Generator/rustc-generator-tests/niche-in-generator-slow_fixme.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[kani::proof]
+fn main() {
+    let x = false;
+    let mut gen1 = || {
+        yield;
+        take(x);
+    };
+
+    // FIXME: for some reason, these asserts are very hard for CBMC to figure out
```
celinval, line 30: Can you please create an issue to take a look at the performance of this test?

celinval (approved):
Thanks for doing this!

tests/kani/Generator/rustc-generator-tests/niche-in-generator-slow_fixme.rs
fzaiser: Opened #1406

Issue #1379, opened 2022-07-14
Fix symbol table generated during the `std` library codegen


celinval:
After fixing #1361, I realized that there are other problems that might still come up during the `symtab2gb` execution. So this issue is to capture the entire work to fix the symbol table generated by Kani.

Here are the steps:

1. `./scripts/std-lib-regression.sh`
2. Go to the build test folder location that is generated by the script above:
```
cd /tmp/std_lib_test/target/x86_64-unknown-linux-gnu/debug/deps
```
3. Convert symbol table json files to goto binaries using
```
for symtab in *.symtab.json;  \
   do echo -n "file: $symtab";  \
   symtab2gb ${symtab} --out ${symtab}.out >& /tmp/${symtab}.log && echo " OK" || echo " FAILED"; \
done
```

The step 3 prints the following result:
```
file: addr2line-3f4cf112b519197b.symtab.json OK                                                      
file: adler-f5149953734bf5e3.symtab.json OK
file: alloc-281073fe438e6cf6.symtab.json OK    
file: cfg_if-0ed47ff0d6bf89ba.symtab.json OK
file: compiler_builtins-606d827214a9b1db.symtab.json OK
file: core-c7fd88564a2d3196.symtab.json FAILED
file: gimli-8b4cebb005ed89ed.symtab.json OK            
file: hashbrown-4796c4a50e166ab2.symtab.json OK
file: kani-1805395fffbdf681.symtab.json OK
file: libc-6cd8b2f54bf2ebfb.symtab.json OK
file: memchr-86e94d5191e5d69f.symtab.json OK
file: miniz_oxide-7ecee372960b75a7.symtab.json OK
file: object-856c2527c19e4e9b.symtab.json FAILED
file: panic_abort-54aa208422f6c037.symtab.json OK
file: panic_unwind-bbdc74bcc211e1f2.symtab.json OK
file: proc_macro-1538872edcf7599f.symtab.json FAILED
file: rustc_demangle-2e0d569655a10787.symtab.json OK
file: rustc_std_workspace_alloc-095c74165043ee43.symtab.json OK
file: rustc_std_workspace_core-47661bab16414e16.symtab.json OK
file: std-cb921bcbb6d9cb8f.symtab.json FAILED
file: std_detect-c027811d7aa4d753.symtab.json FAILED
file: std_lib_test-d2ef1fe261c19014.symtab.json OK
file: unwind-d25f4abf4f22ebda.symtab.json OK
```

[...]

PR #1380, opened 2022-07-14
Fix static variable malformed symbol

### LINKED ISSUE #1361: Symbol malformed while trying to link std library with symtab
After applying the fix from #1360, if I try to link the standard library, I get `Symbol malformed` errors (In contrast to the CBMC invariant violation reported by @zhassan-aws  https://github.com/model-checking/kani/issues/1213#issuecomment-1143033711).

Here are the steps:

1. `./scripts/std-lib-regression.sh`
2. Go to the build test folder location that is generated by the script above:
```
cd /tmp/std_lib_test/target/x86_64-unknown-linux-gnu/debug/deps
```
3. Convert symbol table json files to goto binaries using
```
for symtab in *.symtab.json;  \
   do echo -n "file: $symtab";  \
   symtab2gb ${symtab} --out ${symtab}.out >& /tmp/${symtab}.log && echo " OK" || echo " FAILED"; \
done
```

The step 3 prints the following result:
```
file: addr2line-3f4cf112b519197b.symtab.json OK                                                      
file: adler-f5149953734bf5e3.symtab.json OK
file: alloc-281073fe438e6cf6.symtab.json OK    
file: cfg_if-0ed47ff0d6bf89ba.symtab.json OK
file: compiler_builtins-606d827214a9b1db.symtab.json OK
file: core-c7fd88564a2d3196.symtab.json FAILED
file: gimli-8b4cebb005ed89ed.symtab.json OK            
file: hashbrown-4796c4a50e166ab2.symtab.json OK
file: kani-1805395fffbdf681.symtab.json OK
file: libc-6cd8b2f54bf2ebfb.symtab.json OK
file: memchr-86e94d5191e5d69f.symtab.json OK
file: miniz_oxide-7ecee372960b75a7.symtab.json OK
file: object-856c2527c19e4e9b.symtab.json FAILED
file: panic_abort-54aa208422f6c037.symtab.json OK
file: panic_unwind-bbdc74bcc211e1f2.symtab.json OK
file: proc_macro-1538872edcf7599f.symtab.json FAILED
file: rustc_demangle-2e0d569655a10787.symtab.json OK
file: rustc_std_workspace_alloc-095c74165043ee43.symtab.json OK
file: rustc_std_workspace_core-47661bab16414e16.symtab.json OK
file: std-cb921bcbb6d9cb8f.symtab.json FAILED
file: std_detect-c027811d7aa4d753.symtab.json FAILED
file: std_lib_test-d2ef1fe261c19014.symtab.json OK
file: unwind-d25f4abf4f22ebda.symtab.json OK
```

Looking at the logs of the `symtab2gb` for the failed executions. I see the same error signatures:
```
Symbol is malformed:

<< EXTRA DIAGNOSTICS >>
_ZN4core7unicode12unicode_data9uppercase19BITSET_INDEX_CHUNKS17h6f6fdf24a1340794E
<< END EXTRA DIAGNOSTICS >>
```
The malformed symbols were:
```
_ZN4core7unicode12unicode_data9uppercase19BITSET_INDEX_CHUNKS17h6f6fdf24a1340794E
_ZN6object3elf12ELF_NOTE_GNU17h4cc92826a5db29e2E
_ZN10proc_macro6bridge6client24maybe_install_panic_hook28HIDE_PANICS_DURING_EXPANSION17h549254005c9e063dE
_ZN3std9panicking9HOOK_LOCK17hff49b4883f12eecfE
_ZN10std_detect6detect5cache5CACHE17hd4cc2eaec2c4eb1bE
```

celinval:
### Description of changes: 

A symbol's `base_name` has to be a suffix of its `name` otherwise it is considered a malformed symbol. For static variables, Kani was using the `pretty_name` to generate the `base_name` which was generating invalid iRep as a result. This was caught while trying to run symtab2gb after the std compilation.

### Resolved issues:

Fixes #1361

### Call-outs:

I couldn't figure out a good way of testing this. I also cannot enable `symtab2gb` on the std regression because I am now facing a new problem. I created #1379 to track the entire work of getting `symtab2gb` command working.

### Testing:

* How is this change tested? The assert I added to `Symbol::new()` would fail without the actual fix.

### Checklist

### DIFF
--- a/scripts/std-lib-regression.sh
+++ b/scripts/std-lib-regression.sh
@@ -61,6 +61,7 @@ kani = {path=\"${KANI_DIR}/library/kani\"}
 cp ${KANI_DIR}/rust-toolchain.toml .
 
 echo "Starting cargo build with Kani"
+export RUST_BACKTRACE=1
 export RUSTC_LOG=error
 export KANIFLAGS="--goto-c --ignore-global-asm"
 export RUSTFLAGS="--kani-flags"
--- a/cprover_bindings/src/goto_program/symbol.rs
+++ b/cprover_bindings/src/goto_program/symbol.rs
@@ -71,6 +71,13 @@ impl Symbol {
         let name = name.into();
         let base_name = base_name.intern();
         let pretty_name = pretty_name.intern();
+        // See https://github.com/model-checking/kani/issues/1361#issuecomment-1181499683
+        assert!(
+            name.to_string().ends_with(&base_name.map_or(String::new(), |s| s.to_string())),
+            "Symbol's base_name must be the suffix of its name.\nName: {:?}\nBase name: {:?}",
+            name,
+            base_name
+        );
         Symbol {
             name,
             location,
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/static_var.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/static_var.rs
@@ -10,9 +10,6 @@ use rustc_middle::mir::mono::MonoItem;
 use rustc_middle::ty::{subst::InternalSubsts, Instance};
 use tracing::debug;
 
-/// Separator used to generate function static variable names (<function_name>::<variable_name>).
-const SEPARATOR: &str = "::";
-
 impl<'tcx> GotocCtx<'tcx> {
     pub fn codegen_static(&mut self, def_id: DefId, item: MonoItem<'tcx>) {
         debug!("codegen_static");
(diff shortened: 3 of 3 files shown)


tedinski (approved):
The assert is a great addition.

Before merging: can you give me an example of what the before/after names were here? What were we doing? It's hard to tell from the code.

celinval:
> 

It was something like:
```
name: _ZN3std9panicking9HOOK_LOCK17hff49b4883f12eecfE
friendly name: std::panicking::HOOK_LOCK
base name: HOOK_LOCK
```

scripts/std-lib-regression.sh
```diff
@@ -61,6 +61,7 @@ kani = {path=\"${KANI_DIR}/library/kani\"}
 cp ${KANI_DIR}/rust-toolchain.toml .
 
 echo "Starting cargo build with Kani"
+export RUST_BACKTRACE=1
```
fzaiser, line 64: Is this intentional?

celinval: Yes. I think we should always include that in our scripts to make sure we get a stack trace back when there's a crash.

I believe we do that in the regression script already, but it wouldn't cover when running the script on its own.

celinval:
TBH, I was tempted to just remove `base_name` from all `Symbol` construction method since we just basically duplicate `name` everywhere we call these functions.

fzaiser:
What is the `base_name` used for actually? The docstring is not very helpful.

celinval:
> What is the `base_name` used for actually? The docstring is not very helpful.

Same question I had. See https://github.com/model-checking/kani/issues/1361#issuecomment-1185749783

PR #1383, opened 2022-07-15
Change --dry-run to inject a dummy harness

celinval:
### Description of changes: 

This is a small improvement to --dry-run. Instead of requiring the user to provide "--function dummy_function" we
just use "harness" as the name of the dummy harness.

### Resolved issues:

N/A

### Call-outs:

I added a new test suite (cargo-ui) to test arguments and other UI related to cargo-kani and updated the tests documentation.

### Testing:

### Checklist


docs/src/regression-testing.md
```diff
@@ -56,6 +56,7 @@ In particular, the Kani testing suites are composed of:
                 Arguments can be specified in the `Cargo.toml` configuration file.
                 Similar to the `expected` suite, we look for `*.expected` files
                 for each harness in the package.
+* `cargo-ui`: Similar to `cargo-kani`, but focuses on the user interface like the `ui` test suite.
```
tedinski, line 59: Every additional suite we add loses us parallelism in the test suite. Is this really necessary?

kani-driver/src/metadata.rs
```diff
@@ -110,7 +120,7 @@ impl KaniSession {
     pub fn collect_kani_metadata(&self, files: &[PathBuf]) -> Result<KaniMetadata> {
         if self.args.dry_run {
             // Mock an answer
-            Ok(KaniMetadata { proof_harnesses: vec![] })
+            Ok(KaniMetadata { proof_harnesses: vec![generate_mock_harness()] })
```
tedinski, line 123: @danielsn has previously proposed that `--dry-run` should actually do the build still, so we actually get the real proof harness metadata.

Basically, so you can run the cbmc command manually with some tweaks or whatever during development.

IMO if we're changing dry-run maybe that's what we actually change it to?

celinval: We could add dry-run to the compiler and only collect the metadata.

celinval: The concept of dry does get a bit fuzzy though if we now run some commands but not others. I think I'm that case you probably want to have a not so verbose mode that prints the command before running. Like cargo does

docs/src/regression-testing.md
celinval: I think it's cleaner to have suites that test specific things. I don't think the impact on the time the regression runs is significant.

Issue #957, opened 2022-03-18
Kani should make use of CBMC's internal consistency checks.


tautschnig:
Requested feature: `cbmc` and also `goto-instrument` can be invoked with `--validate-goto-model` to perform consistency checks on the GOTO model. `cbmc` furthermore supports `--validate-ssa-equation` to enable consistency checks on the formula generated during symbolic execution. These bear additional runtime cost, but could help spot bugs creeping in as Kani builds expressions using a completely different code base. CBMC uses these options in its regression test suite. 
Use case: Enabling those in regression tests might well be the right approach for Kani as well.
Link to relevant documentation (Rust reference, Nomicon, RFC):
Is this a breaking change? Yes, because there currently are some inconsistencies reported when manually enabling these options on Kani-generated goto binaries.

Test case: Running CBMC on the goto binary attached to https://github.com/diffblue/cbmc/issues/6731:
```
cbmc test.out --validate-goto-model
CBMC version 5.52.0 (cbmc-5.52.0-61-g686da25177) 64-bit x86_64 linux
Reading GOTO program from file test.out
Generating GOTO Program
--- begin invariant violation report ---
Invariant check failed
../src/goto-programs/goto_function.cpp:43 function: validate
Condition: identifier.empty() || ns.lookup(identifier).is_parameter
Reason: parameter should be marked 'is_parameter' in the symbol table
Backtrace:
build/bin/cbmc(print_backtrace(std::ostream&)+0x50) [0x56243462d600]
build/bin/cbmc(get_backtrace[abi:cxx11]()+0x169) [0x56243462dba9]
build/bin/cbmc(std::enable_if<std::is_base_of<invariant_failedt, invariant_failedt>::value, void>::type invariant_violated_structured<invariant_failedt, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)+0x48) [0x562433c4d618]
build/bin/cbmc(goto_functiont::validate(namespacet const&, validation_modet) const+0x103) [0x562434144543]
[...]

PR #1387, opened 2022-07-18
Add option to run CBMC sanity checks + set is_param for function parameters 

celinval:
### Description of changes: 

1. Add an unstable hidden option to `kani-driver` to run sanity checks that are available in CBMC to perform consistency checks on the GOTO model.
2. Fix consistency checks related to missing `is_parameter` flag. We were not setting this parameter at all. CBMC expects all symbols that are function parameters to have this flag set to `true`.

### Resolved issues:

This is related to #957 

### Call-outs:

We would like to enable these checks (as requested in #957), however the regression is still failing as described in https://github.com/model-checking/kani/issues/957#issuecomment-1188031817.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


kani-driver/src/call_goto_instrument.rs
```diff
@@ -130,6 +134,16 @@ impl KaniSession {
         self.call_goto_instrument(args)
     }
 
+    fn goto_sanity_check(&self, file: &Path) -> Result<()> {
```
tedinski, line 137: This seems to be running it twice, when we're already doing it with the cbmc args?

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -58,7 +67,8 @@ impl<'tcx> GotocCtx<'tcx> {
             let loc = self.codegen_span(&ldata.source_info.span);
             let sym =
                 Symbol::variable(name, base_name, t, self.codegen_span(&ldata.source_info.span))
-                    .with_is_hidden(!ldata.is_user_variable());
+                    .with_is_hidden(!ldata.is_user_variable())
+                    .with_is_parameter(idx > 0 && idx <= num_args);
```
tedinski, line 71: I don't know why this condition is here? Definitely calls for a comment.

kani-driver/src/call_goto_instrument.rs
celinval: Yeah. I figured it's a cheap analysis, so we might as well run before any `goto-instrument` optimization and right before verification.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
celinval: yeah... that's a good point. I'll add some comments.

kani-driver/src/call_goto_instrument.rs
celinval: that said, I totally misplaced it. Let me move it up.

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
```diff
@@ -153,7 +159,8 @@ impl<'tcx> GotocCtx<'tcx> {
         loc: Location,
     ) -> (Expr, Stmt) {
         let c = self.current_fn_mut().get_and_incr_counter();
-        let var = self.gen_stack_variable(c, &self.current_fn().name(), "temp", t, loc).to_expr();
+        let var =
+            self.gen_stack_variable(c, &self.current_fn().name(), "temp", t, loc, false).to_expr();
```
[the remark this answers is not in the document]
tedinski: Oh, a proliferation of boolean arguments isn't good either. Maybe ignore my comment here, or consider a "gen_hidden_local_variable" function or something like that.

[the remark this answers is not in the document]
celinval: Sorry, I thought you were talking about the explicit `Symbol::variable` call in function.rs.

This does name variables "temp" instead of "var". I would rather not change that in this PR if that's OK for you.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
tedinski: What's 0? And the rest are locals I assume?

tedinski (approved):

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
celinval: 0 is the return variable. There is a comment right below that talks about that.

tedinski (approved):

PR #1398, opened 2022-07-20
Comment and Refactor a few files

tedinski:
### Description of changes: 

1. Add a lot of doc comments and other comments, focusing mostly on `statement.rs`
2. Move some code around.
3. Rename some bad variables.

Please note: anything that looks like new/deleted code is actually just moving. Github makes it a bit rough to review these diffs, I think.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
@@ -1,9 +1,12 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! This file contains the code that acts as a wrapper to create the new assert and related statements
+//! This module is the central location for handling assertions and assumptions in Kani.
+
+use crate::codegen_cprover_gotoc::utils;
 use crate::codegen_cprover_gotoc::GotocCtx;
-use cbmc::goto_program::{Expr, Location, Stmt};
+use cbmc::goto_program::{BuiltinFn, Expr, Location, Stmt};
+use rustc_span::Span;
 use std::convert::AsRef;
 use strum_macros::{AsRefStr, EnumString};
 
@@ -39,6 +42,7 @@ impl PropertyClass {
 }
 
 impl<'tcx> GotocCtx<'tcx> {
+    /// Generates a CBMC assertion. Note: Does _NOT_ assume.
     pub fn codegen_assert(
         &self,
         cond: Expr,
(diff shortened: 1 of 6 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -66,6 +71,7 @@ impl<'tcx> GotocCtx<'tcx> {
         )
     }
 
+    /// A shorthand for generating a CBMC assert-false. TODO: This should probably be eliminated!
```
adpaco-aws, line 74: Do we want an issue for this?

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
```diff
@@ -1,12 +1,16 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! This file contains functions related to codegenning MIR blocks into gotoc
```
adpaco-aws, line 4: Why remove this? I feel like having at least one comment is better than not having it at all.

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -24,19 +23,166 @@ use rustc_target::abi::{FieldsShape, Primitive, TagEncoding, Variants};
...
-            FN_RETURN_VOID_VAR_NAME,
-            is_file_local,
-            ty,
-            Location::none(),
-            |_, _| None,
-        );
-        Stmt::ret(Some(var), Location::none())
+    /// Generate Goto-C for MIR [Statement]s.
+    /// This does not cover all possible "statements" because MIR distinguishes between ordinary
+    /// statements and [Terminator]s, which can exclusively appear the the end of a basic block.
```
adpaco-aws, line 28:
```suggestion
    /// statements and [Terminator]s, which can exclusively appear at the end of a basic block.
```

adpaco-aws (approved):
Thanks! Just a few comments.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
tedinski: It's the assertions audit. I will be doing it next. :)

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
tedinski: I basically just moved the comment from the file to the function.

adpaco-aws: Right, I thought this implementation was way longer.

Issue #1394, opened 2022-07-20
Improvements needed for the 'kani' library


tedinski:
# `any`, `Arbitrary` and `Invariant`

* [x] The current implementation of `Invariant` is basically inviting undefined behavior. For instance, an invariant like "this enum must match one of its variants" is something the rust compiler could replace with just `true`. (It sort of already does: if you do this with pattern matching, it recognizes the wildcard branch as dead code, and Kani helpfully codegen's dead code as `assert(false)`!)
* [x] The current implementation of `Arbitrary` offers essentially nothing useful. For instance, implementing `Arbitrary` for `MyType` does not mean you can use `any()` for `Option<MyType>` because the implication instances for `Option` are on `Invariant`
* [x] We could move some of these from `Invariant` to `Arbitrary` but that starts to eliminate the usefulness of `Invariant`, and... (next item)
* [ ] In the long term, we DO want `Invariant`, it's just that we'd want the compiler to synthesize the basic ones for types. (All the invariants that are undefined behavior are the ones we should be able to implement automatically because they're the Rust invariants for types.)
* [x] We should make `any_raw` private, users probably shouldn't be using it, and it's mostly UB. It can be mimic'd by getting `any` bytes and then `transmute`, with all the obvious UB implications.

# any_vec, slices

I don't believe the current API is the one we should consider stabilized.

* [ ] `any_slice` returns a fairly complicated `AnySlice` type, because it needs something to own the data. Thing is, the standard library already has a type for that... `Vec`. I'm not sure why we don't just have `any_vec` alone here.
* [x] We should probably mark all `raw` functions as private.
* [ ] IMO `any_slice_of_array` should be `any_subslice`
* [ ] I'm not sure why `any_slice` and `any_vec` pass in the bound as a const generic. Seems like it'd be best if it were an ordinary parameter. It makes these hard to use. (I didn't look into the implementations though, so could easily be a reason we have to unfortunately)
* [ ] These implementations should be exported at the top level. i.e. `kani::any_vec` not `kani::vec::any_vec`
* [ ] We should add an `any_string` implementation.

# Assertions

* [x] `nondet` should be removed or at least `doc(hidden)`
* [ ] `panic` should probably be at least `doc(hidden)`. Not something we expect customers to know about.
* [ ] `assert` should be renamed `check`
* [x] `expect_fail` should be replaced with `cover` #1197

celinval:
[...]

PR #1401, opened 2022-07-22
Replace struct impl of Invariant by Arbitrary

celinval:
### Description of changes: 

We have recently realized that implementing Invariant of Option, Result and array has a limitation. This is not sufficient to generate arbitrary values of types that only implements Arbitrary.

Thus, implement Arbitrary instead.

### Resolved issues:

N/A. Related to #1394 

### Testing:

* Is this a refactor change?

### Checklist


celinval:
I forgot to mention that I did a few experiments with `check_both` test that I added. I changed the test to create an array with 100 elements instead.

For the arbitrary array, the performance impact with the changes in this PR was small. I tried a different implementation using MaybeUninit (similar to this example: https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#initializing-an-array-element-by-element) since this would be the safest, and there was a significant penalty on the verification time. Here is the data I collected:

| Before Changes  | After Changes | MaybeUninit |
| ------------- | ------------- | --------- |
| 10.0s | 10.7s | 12.7s |

tedinski (approved):
I guess I don't object after sleeping on it. I think we'll reverse course later, but it's probably not important.

Please hold off on merging until after today's release maybe?

Issue #1373, opened 2022-07-12
Clippy Compliance within Kani


YoshikiTakashima:
As noted by @giltho & @celinval, there is a lot of clippy warnings in
Kani. This issue is for tracking compliance and tasks related to
measuring compliance.

- [x] Figure out which parts of Kani have Clippy warnings, and how many.
  - All 32 **top-level** hard errors come from `kani-compiler`.
    - This result does not conflict with people getting 100+ issues
    since these are top-level errors. Some errors may contain
    sub-errors, and we exclude all warnings.
  - When warnings are included, the breakdown is as follows.
![Screen Shot 2022-07-13 at 8 27 17 PM](https://user-images.githubusercontent.com/26704033/178859383-bf50c196-d941-4c95-93a6-e2e071d09867.png)
```js
{'library/kani': 3,
 'src': 55,
 'cprover_bindings': 17,
 'kani-driver': 6,
 'kani-compiler': 23,
 'tools/bookrunner': 5,
 'tools/compiletest': 44}
```

- [x] Add Clippy warnings to `script/kani-regressions.sh`. For now,
      explicitly ignore libraries that warnings, and incrementally
      remove them as they are fixed.

YoshikiTakashima:
@celinval Shall we enable `-D warnings` for clippy as well? I recall they are enabled for `cargo build`.

YoshikiTakashima:
For regression: 

```bash
cargo clippy --all-targets --all-features -- -D warnings
```

celinval:
Yes, please go ahead and add them to the regression!

YoshikiTakashima:
As per discussion on #1389, we permanently will ignore these two.

> @YoshikiTakashima The two lints are called [`expect_fun_call`](https://rust-lang.github.io/rust-clippy/master/#expect_fun_call) and [`or_fun_call`](https://rust-lang.github.io/rust-clippy/master/#or_fun_call).

YoshikiTakashima:
For actual kinds of warnings:
```
  54 "clippy::cargo_common_metadata"
   3 "clippy::collapsible_else_if"
   5 "clippy::derive_partial_eq_without_eq"
   1 "clippy::expect_fun_call"
   2 "clippy::format_push_string"
   1 "clippy::into_iter_on_ref"
   1 "clippy::iter_skip_next"
   1 "clippy::let_and_return"
   1 "clippy::manual_map"
   1 "clippy::manual_range_contains"
   1 "clippy::manual_strip"
[...]

PR #1403, opened 2022-07-23
Clippy in Regressions.

YoshikiTakashima:
### Description of changes: 

No change to kani behavior. Added `clippy` linter to regression suite. 

### Resolved issues:

Lack of clippy lint checking in kani regression script. #1373 will stay open to track actual clippy errors.

### Call-outs:
The current setup addresses the following clippy regressions.
- New clippy error/warning in a library that did not have them (thus not ignored)
- New class of clippy warnings/errors by code change

### Testing:

* How is this change tested? Run the regression script

* Is this a refactor change?

### Checklist


.github/workflows/format-check.yml
```diff
@@ -20,3 +20,20 @@ jobs:
...
+      - name: Setup Kani Dependencies
+        uses: ./.github/actions/setup
+        with:
+            os: ubuntu-20.04
+
+      - name: "Run Clippy"
+        run: |
+          cargo clippy --all -- -D warnings \
+          -A "clippy::expect_fun_call" \
+          -A "clippy::or_fun_call"
```
giltho, line 39: Can this specific configuration be put in a clippy.toml or something similar?
If I run cargo clippy on my machine, I want it to run just like on the CI, since this is a repo-wide decision, not just a CI decision.

And then, they can also be removed from the lib files

scripts/setup/macos-10.15/install_deps.sh
```diff
@@ -10,7 +10,7 @@ set -eux
 #brew update
 
 # Install dependencies via `brew`
-brew install universal-ctags wget
+brew install universal-ctags wget jq
```
giltho, line 13: Why is jq required? I think you used it for the script that gathers data, but you didn't commit that script

giltho (commented):
I'm not sure how clippy gets configured for a whole workspace, so I'm not sure if the first comment has a solution, the second one should probably be "fixed" before merging

.github/workflows/format-check.yml
YoshikiTakashima: @giltho I think that would be good as well. I'm less sure of the exact method to implement it. If you know of a working example, please send me the link.

A quick search suggests the method for doing that seems to be through `.cargo/config.toml`. https://github.com/emilk/egui/blob/e76c919c7e70c208c9a6209b9fe3369e7b6db99d/.cargo/config.toml

Let me prototype a little bit b/c I am not sure if this will have negative effects. I read **unconfirmed** rumors that this negatively effects incremental recompile.

scripts/setup/macos-10.15/install_deps.sh
YoshikiTakashima: Oops. Thanks for catching it. I will add the statistics gathering command

tedinski: If we're clean now, I'm not sure we need it. But at any rate we don't need to add jq to CI if it's not being used

YoshikiTakashima: We are not clean. This PR papers over the actual warning by turning off the lints. 

I believe this data should be tracked, but I agree you don't need to install jq on everything. Clippy check is only running for Ubuntu-20.04.

YoshikiTakashima: Addressed with the following:
stats print script: 37cd8030b4492a07630401e74e0ffdf70eb21abd
remove jq install from OSX: 7803971d1f646310f146c3658f284a552feb9af5

As of now, no stats will print b/c lints are turned off: 

P.S. Wait what, did GH just suggest a fix??? 
```suggestion
brew install universal-ctags wget jq
```

.github/workflows/format-check.yml
YoshikiTakashima: Update:
1. Incremental recompile issue can be mitigated.
2. However, .cargo/config.toml does not play nicely with the custom build command for crates under `library`.
```
error: failed to run custom build command for `kani-compiler v0.6.0 (/Users/ytakashl/Desktop/kani/kani-compiler)`

Caused by:
  process didn't exit successfully: `/Users/ytakashl/Desktop/kani/target/debug/build/kani-compiler-bfa3790a81ae900c/build-script-build` (exit status: 1)
  --- stdout
  cargo:rustc-link-arg-bin=kani-compiler=-Wl,-rpath,/Users/ytakashl/.rustup/toolchains/nightly-2022-07-19-aarch64-apple-darwin/lib
  cargo:rustc-link-arg-bin=kani-compiler=-Wl,-rpath,@loader_path/../toolchain/lib
  cargo:rerun-if-changed=../library/kani

  --- stderr
     Compiling kani_macros v0.6.0 (/Users/ytakashl/Desktop/kani/library/kani_macros)
  error: length comparison to zero
    --> library/kani_macros/src/lib.rs:42:13
     |
  42 |     assert!(attr.to_string().len() == 0, "#[kani::proof] does not take any arguments");
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: using `is_empty` is clearer and more explicit: `attr.to_string().is_empty()`
     |
     = note: `-D clippy::len-zero` implied by `-D warnings`
```

While there are only 4 warnings under `library/`, one of the warnings is a bit thorny b/c it asks us to define a safety doc comment for Invariant.

YoshikiTakashima: Resolved by adding `.cargo/config.toml`.

Warnings under `library/` are fixed since they appear not to be ignored.

celinval (commented):
Thanks for doing this. I have 2 questions:
1. Is there a reason why you preferred `.cargo/config.toml` instead of `clippy.toml`.
2. How long does `cargo clippy` takes? I was wondering if we should enable this in our regression.

YoshikiTakashima:
@celinval 

> 1. Is there a reason why you preferred `.cargo/config.toml` instead of `clippy.toml`.

I don't think that is possible, and keeping it so seems to be the stance from clippy devs: https://github.com/rust-lang/rust-clippy/issues/1313#issuecomment-633991593. The idea behind it seems to be that you would control all lints at once, not just clippy lints.

> 2. How long does `cargo clippy` takes? I was wondering if we should enable this in our regression.

It takes no longer than 2 compiles. This is the first job to finish.

giltho:
LGTM ✅

giltho (approved):

YoshikiTakashima:
Not sure how big the "code owner" set is but it still saids it waiting for review. @tedinski / @celinval

.cargo/config.toml
```diff
@@ -0,0 +1,39 @@
...
+    "-Aclippy::map_entry",
+    "-Aclippy::match_like_matches_macro",
+    "-Aclippy::missing_safety_doc",
+    "-Aclippy::module_inception",
+    "-Aclippy::needless_arbitrary_self_type",
+    "-Aclippy::needless_bool",
+    "-Aclippy::needless_return",
+    "-Aclippy::new_ret_no_self",
+    "-Aclippy::new_without_default",
+    "-Aclippy::or_fun_call",
```
tedinski, line 31: appears above

.gitignore
```diff
@@ -25,6 +25,7 @@ Session.vim
 ## Tool
 .valgrindrc
 .cargo
+!/.cargo
```
tedinski, line 28: Huh?

scripts/setup/ubuntu/install_deps.sh
```diff
@@ -26,6 +26,7 @@ DEPS=(
   wget
   zlib1g
   zlib1g-dev
+  jq
```
tedinski, line 29: ?

celinval: > @celinval
> 
> > 1. Is there a reason why you preferred `.cargo/config.toml` instead of `clippy.toml`.
> 
> I don't think that is possible, and keeping it so seems to be the stance from clippy devs: [rust-lang/rust-clippy#1313 (comment)](https://github.com/rust-lang/rust-clippy/issues/1313#issuecomment-633991593). The idea behind it seems to be that you would control all lints at once, not just clippy lints.
> 
> > 2. How long does `cargo clippy` takes? I was wondering if we should enable this in our regression.
> 
> It takes no longer than 2 compiles with `cargo build --workspace`. This is the first job to finish.

This is so confusing... That's totally what they recommend in their documentation: https://github.com/rust-lang/rust-clippy#configuration

.gitignore
YoshikiTakashima: 27 includes `*/.cargo`, 28 negates only top-level config.

I did that to minimize semantic diff, but I can just get rid of 27 as well.

scripts/setup/ubuntu/install_deps.sh
YoshikiTakashima: @tedinski `jq` is required for generating the warning stats given at the bottom of .github/workflows/format-check.yml

YoshikiTakashima: @celinval Yea, both are under the same section, which is not great.

I could not find a single example of allowing lints with `clippy.toml` so I went looking around and landed on that issue thread.

.cargo/config.toml
YoshikiTakashima: Fixed with 88a98da5fd200043b8bc0090b61c8d028db99ddf

.gitignore
YoshikiTakashima: Closing as this appears to be a question rather than a request.

scripts/setup/ubuntu/install_deps.sh
YoshikiTakashima: Closing as this appears to be a question rather than a request.

.gitignore
tedinski: Does .cargo appear anywhere else? I don't think so, just removing the line seems best.

scripts/setup/ubuntu/install_deps.sh
tedinski: I would put instally jq into that workflow, not as a general dependency.

tedinski (approved):

.gitignore
YoshikiTakashima: No effect on regressions. I will just get rid of line 27.

scripts/setup/ubuntu/install_deps.sh
YoshikiTakashima: Moved in commit 455c42f492c79745e94ceaed92acf8cc0d27ed14

PR #1414, opened 2022-07-27
Add basic support for async/.await

fzaiser:
### Description of changes: 

This implements the language constructs `async` and `.await`. Since they are lowered to generators in the Rust MIR, #1378 did most of the heavy lifting. The only necessary change was to stop ignoring the `std::future::from_generator` function, which turns a generator into a future. Kani can compile it successfully now.

### Call-outs:

To test this, I had to add a function `poll_loop`, which repeatedly polls a future until it completes. We'd probably want to keep such a function in the kani library under `kani::async_await`. But this PR is intentionally minimal and leaves such improvements for later.

### Testing:

* How is this change tested? Added a test for both `async { ... }` blocks and `async fn` functions.

### Checklist


tests/kani/AsyncAwait/main.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// compile-flags: --edition 2018
```
celinval, line 4: Please add a comment saying what you are testing

tests/kani/AsyncAwait/main.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, RawWaker, RawWakerVTable, Waker},
+};
+
+#[kani::proof]
+#[kani::unwind(10)]
+fn main() {
```
celinval, line 14: nit: Can you please use a more informative name instead of `main`.

celinval (approved):
It looks good to me. Just some minor comments.

zhassan-aws (approved):
Awesome!

danielsn (approved):

scripts/kani-fmt.sh
```diff
@@ -32,7 +32,7 @@ for suite in "${TESTS[@]}"; do
     set -f; IFS=$'\n'
     files=($(find "${suite}" -name "*.rs"))
     set +f; unset IFS
-    rustfmt --unstable-features "$@" "${files[@]}" || error=1
+    rustfmt --unstable-features "$@" "${files[@]}" --edition 2021 || error=1
```
adpaco-aws, line 35: Can this be added to `rustfmt.toml` instead? If possible, I'd rather put it there instead of passing as a parameter in  this script.

tedinski (approved):
well I don't want to be left out of the approval chain

tests/kani/AsyncAwait/main.rs
fzaiser: Done

tests/kani/AsyncAwait/main.rs
fzaiser: Done

scripts/kani-fmt.sh
fzaiser: Good point! Done

Issue #726, opened 2022-01-05
Create custom derive macro for Arbitrary type.


celinval:
Requested feature: Create a proc_macro that allows users to annotate their structs and enums with #[derive(Invariant)]
Use case: Allow users to easily generate implementation of the Invariant types for their custom structs and enums whenever the invariant of their types is the combination of their field's invariant.
Link to relevant documentation (Rust reference, Nomicon, RFC):
Is this a breaking change? No

Test case:
```rust
#[derive(Invariant)]
enum Operation {
    Sum,
    Sub,
}

fn main() {
    let op : Operation = rmc::any();
    let v1 : i32 = rmc::any();
    let v2: i32 = rmc::any();
    calculate(v1, v2, op);
}
```

This should generate something like:
```rust
unsafe impl Invariant for Operation {
    fn is_valid(&self) -> bool {
        matches!(*self, Operation::Sum | Operation::Sub);
    }
}
```

celinval:
Since we have deprecated `Invariant`, I'm updating this issue to create a derive macro for Arbitrary.

PR #1415, opened 2022-07-27
Deprecate any_raw and Invariant

celinval:
### Description of changes: 

Basically invariant type was only safe for T that had no invalid value independent on its binary values. All other types had some sort of UB involved. Thus, deprecate this for now.

I also removed `nondet` which has been deprecated for quite some time.

### Resolved issues:

Related to #1394 

### Call-outs:

We will likely want the Invariant type reintroduced one day, but we might need to either change its form or ensure that `any_raw()` at least respect `primitives` and `std` type invariants to avoid users having to explicitly check for any potential UB (which in itself is UB).

### Testing:

* How is this change tested? New performance tests + old tests

### Checklist


kani-driver/src/call_cbmc.rs
```diff
@@ -54,6 +54,7 @@ impl KaniSession {
                 // the best possible fix is port to rust instead of using python, or getting more
                 // feedback than just exit status (or using a particular magic exit code?)
             }
+            // TODO: We should print this even the verification fails but not if it crashes.
```
sanjit-bhat, line 57: If verification crashes, wouldn't the code not reach the `println`? Also, I think it should be "even if the".

library/kani/src/arbitrary.rs
```diff
@@ -1,33 +1,103 @@
...
+trivial_arbitrary!(i64);
+trivial_arbitrary!(i128);
+trivial_arbitrary!(isize);
+
+// We do not constraint floating points values per type spec. Users must add assumptions to their
+// verification code if they want to eliminate NaN, infinite, or subnormal.
+trivial_arbitrary!(f32);
+trivial_arbitrary!(f64);
+
+trivial_arbitrary!(());
```
sanjit-bhat, line 45: What would this show up as in the CBMC output?

library/kani/src/arbitrary.rs
```diff
@@ -1,33 +1,103 @@
...
+
+impl Arbitrary for bool {
+    #[inline(always)]
+    fn any() -> Self {
+        let byte = u8::any();
+        byte == 0
```
sanjit-bhat, line 51: I'm confused by this. Why is `any` returning a bool expression instead of just the `u8::any()`?

library/kani/src/lib.rs
```diff
@@ -74,38 +75,31 @@ pub fn any<T: Arbitrary>() -> T {
...
 
-/// This function has been split into a safe and unsafe functions: `kani::any` and `kani::any_raw`.
-#[deprecated]
+/// This function will replace `any_raw` that has been deprecated and it should only be used
+/// internally when we can guarantee that it will not trigger any undefined behavior.
+#[rustc_diagnostic_item = "KaniAnyRaw"]
 #[inline(never)]
-pub fn nondet<T: Arbitrary>() -> T {
-    any::<T>()
+pub(crate) unsafe fn any_raw_internal<T>() -> T {
```
sanjit-bhat, line 101: Is there a way to add a compiler bound to guarantee that `T` is an int? This would prevent us from accidentally misusing this function.

sanjit-bhat (commented):
Looks good. A much needed change!

General clarification: now, if a user wants to call `kani::any` on a custom type, do they manually have to implement `kani::any`? And their implementation would in-turn call `kani::any` on the sub-fields?

kani-driver/src/call_cbmc.rs
celinval: Definetely "even if". :)

Today, we only print the verification time if the verification finishes and it is successful. I am trying to say that we should print the time whenever the verification finishes, even if there was one or more checks that failed. Does it make sense?

library/kani/src/arbitrary.rs
celinval: Because this is the implementation for `bool::any()`. :)

library/kani/src/arbitrary.rs
celinval: That's a great question! We actually don't generate this value in the compiler. https://github.com/model-checking/kani/blob/4465b9819286bea550f3f62fdf6322a55939500f/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs#L206-L208

library/kani/src/lib.rs
celinval: Yes. That's a great idea! Once we completely remove `any_raw`, we could ensure that T is either an integer, a floating point or unit type.

celinval:
> Looks good. A much needed change!
> 
> General clarification: now, if a user wants to call `kani::any` on a custom type, do they manually have to implement `kani::any`? And their implementation would in-turn call `kani::any` on the sub-fields?

Yes. Users will have to call `any()` for each field of their struct. We provide the `Arbitrary` trait that allows user to implement `any()` for their type once and reuse that for things like array, Option, Result and others.

We are also planning to add a `#[derive(Arbitrary)]` macro to make this easier for the user (https://github.com/model-checking/kani/issues/726).

library/kani/src/lib.rs
```diff
@@ -74,38 +75,31 @@ pub fn any<T: Arbitrary>() -> T {
...
 ///
-#[rustc_diagnostic_item = "KaniAnyRaw"]
+/// We have decided to deprecate this function due to the fact that its result can be the source
+/// of undefined behavior.
 #[inline(never)]
+#[deprecated(
+    since = "0.8.0",
+    note = "This function may return symbolic values that don't respects the language type invariants."
+)]
 pub unsafe fn any_raw<T>() -> T {
```
tedinski, line 93: Should we also `doc_hidden` ?

tedinski (approved):

library/kani/src/arbitrary.rs
```diff
@@ -1,33 +1,103 @@
...
     }
 }
 
+/// Validate that a char is not outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
+/// Ref: <https://doc.rust-lang.org/stable/nomicon/what-unsafe-does.html>
+impl Arbitrary for char {
+    #[inline(always)]
+    fn any() -> Self {
+        // Kani translates char into i32.
```
zhassan-aws, line 60: Should the comment say "into u32"?

library/kani/src/invariant.rs
```diff
@@ -14,84 +15,33 @@ use std::num::*;
...
+///
+/// # Deprecation
+///
+/// We have decided to deprecate this trait in favor of using `Arbitrary` for now. The main
+/// benefit of using Invariant was to avoid calling `kani::any_raw()` followed by `kani::assume()`.
+/// However, `kani::any_raw()` today cannot guarantee the rust type validity invariants, which
+/// could lead to UB in our analysis.
+#[deprecated(
+    since = "0.8.0",
+    note = "Kani cannot guarantee that the type respects the language type invariants which may \
```
zhassan-aws, line 27:
```suggestion
    note = "With `kani::Invariant`, Kani cannot guarantee that the type respects the language type invariants which may trigger UB. Use `kani::Arbitrary` instead. \
```

library/kani/src/invariant.rs
```diff
@@ -14,84 +15,33 @@ use std::num::*;
...
+///
+/// # Deprecation
+///
+/// We have decided to deprecate this trait in favor of using `Arbitrary` for now. The main
+/// benefit of using Invariant was to avoid calling `kani::any_raw()` followed by `kani::assume()`.
+/// However, `kani::any_raw()` today cannot guarantee the rust type validity invariants, which
+/// could lead to UB in our analysis.
+#[deprecated(
+    since = "0.8.0",
+    note = "With `kani::Invariant`, Kani cannot guarantee that the type respects the language type invariants which may trigger UB. Use `kani::Arbitrary` instead. \
```
zhassan-aws, line 27: Whoops, this got messed up due to my comment. You can remove the next line and perhaps split this line into two.

library/kani/src/arbitrary.rs
celinval: This is an old comment actually.

library/kani/src/lib.rs
celinval: Sure. I don't think we publish our crates doc yet, but hopefully soon.

zhassan-aws (approved):
Thanks!

library/kani/src/invariant.rs
```diff
@@ -14,84 +15,33 @@ use std::num::*;
...
-
-/// Validate that a char is not outside the ranges [0x0, 0xD7FF] and [0xE000, 0x10FFFF]
-/// Ref: <https://doc.rust-lang.org/stable/nomicon/what-unsafe-does.html>
-unsafe impl Invariant for char {
-    #[inline(always)]
-    fn is_valid(&self) -> bool {
-        // Kani translates char into i32.
-        let val = *self as i32;
-        val <= 0xD7FF || (0xE000..=0x10FFFF).contains(&val)
+// We cannot apply #[deprecated] too trait impl so add this to ignore the deprecation warnings.
```
sanjit-bhat, line 36: too -> to

kani-driver/src/call_cbmc.rs
sanjit-bhat: Yup, that makes sense!

sanjit-bhat: Do you want to make the typo change?

library/kani/src/arbitrary.rs
```diff
@@ -1,33 +1,103 @@
...
+nonzero_arbitrary!(NonZeroIsize, isize);
+
 impl<T, const N: usize> Arbitrary for [T; N]
 where
     T: Arbitrary,
 {
     fn any() -> Self {
         // The "correct way" would be to MaybeUninit but there is performance penalty.
-        let mut data: [T; N] = unsafe { crate::any_raw() };
+        let mut data: [T; N] = unsafe { crate::any_raw_internal() };
```
sanjit-bhat, line 100: I never fully understood why we call `any_raw_internal` here followed by `any` for every element? Can we instead just initialize the array with zeros and then fill it with nondet data?

Also, from what I understand, this is calling `any_raw_internal` for `[T; N]`. Doesn't this violate the invariant that we only call it on primitives?

celinval: We can't initialize everything to zero. Depending on the type it would be UB. MaybeUnit is the correct way but the performance was rather bad. Let me see what else I can do.

zhassan-aws: Would this be more efficient:
```rust
[(); N].map(|_| kani::any());
```
?
Source: https://stackoverflow.com/a/69488028

celinval: I can give it a shot. Thanks

celinval: Yep, it's slightly faster and much cleaner. Thanks for the suggestion!

PR #1419, opened 2022-07-28
Add assigns clause to Irep generation for function contracts.

monal:
### Description of changes: 

Added `assigns` clause to the Irep generation code. This is a required field in the CBMC Irep for contracts. The Irep generation logic for the `assigns` clause is identical to those of `ensures` and `requires` clauses. 

### Resolved issues:

No issues created for this.

### Call-outs:
This is a fairly straightforward change (mimics what we are doing for `ensures` and `requires` already).

### Testing:

* How is this change tested?
This code is not tested as of yet which is why it will remain in the `features/function-contracts` branch as of now. Once we have macros for function contracts, we can test the code end-to-end and merge into `main`.

* Is this a refactor change?
No.

### Checklist


cprover_bindings/src/goto_program/contract.rs
```diff
@@ -12,7 +12,7 @@ use std::fmt::Debug;
 /// Represents a contract on a function, loop, etc.
 #[derive(Clone, Debug)]
 pub enum Contract {
-    FunctionContract { ensures: Vec<Spec>, requires: Vec<Spec> },
+    FunctionContract { ensures: Vec<Spec>, requires: Vec<Spec>, assigns: Vec<Spec> },
```
zhassan-aws, line 15: Is the same `Spec` type suitable for `assigns` clauses? What would an `assigns` clause look like in Rust? Would it refer to expressions or just symbols?

monal: Yeah, `assigns: Vec<Spec>` is too general. An assigns clause `__CPROVER_assigns(x)` gets wrapped into a lambda expression on the CBMC side. So, it will look something like - (|function_return_type, function_arguments| x). 

On the Rust side, it would be `kani::assigns(x)` where x is an `Expr` with `ExprValue` being a symbol. 

I've updated the code to make sure that the only way we can construct a `FunctionContract` type is through the constructor. In the constructor, in checking to make sure that each `assigns(x)` clause satisfies `x.is_symbol()`. 

Does this sound good?

zhassan-aws: Sounds good.

library/kani_macros/src/contracts.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright Kani Contributors
```
zhassan-aws, line 1: This should be left to a different PR?

monal: Which part? The copyright information?

zhassan-aws: No, I meant this whole file. This seems to be for adding the Kani macros for pre-post conditions? If so, should this be in a separate PR?

zhassan-aws (approved):
Thanks!

Issue #1425, opened 2022-07-29
The `kani` library doesn't get linked to goto binaries


tedinski:
This was previously invisible because we either (1) had codegen hooks, so we replaced calls during codegen, or (2) were generic, and so got monomorphized into dependent crates.

But as soon as we add a normal symbol, no generics and no codegen hook, you'd see missing symbol errors because we never get a `symtab.json` from `kani` and link it into generated goto binaries.

tedinski:
@celinval I believe this is fixed now right? Closing...

PR #1427, opened 2022-07-29
Add a `kani::futures` library containing `block_on`

fzaiser:
### Description of changes: 

Following up on #1414, this adds a `kani::block_on` function to drive futures to completion in Kani. The `block_on` function from [`futures`](https://docs.rs/futures/latest/futures/executor/fn.block_on.html) or [`tokio`](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on) do not work because Kani does not understand the thread primitives these implementations use.

### Callouts:

I had to add `#[inline]` annotations to work around the linking issue #1425. But inlining makes sense for these functions anyway (they're small and do basically nothing), so this is not a problem.

### Testing:

* How is this change tested? Added  a test in `tests/kani/AsyncAwait/main.rs`

### Checklist

### DIFF
--- a/library/kani/src/futures.rs
+++ b/library/kani/src/futures.rs
@@ -1,15 +1,20 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+//! This module contains functions to work with futures (and async/.await) in Kani.
+
 use std::{
     future::Future,
     pin::Pin,
     task::{Context, RawWaker, RawWakerVTable, Waker},
 };
 
-/// Polls the future in a loop until completion (a very simple executor)
+/// A very simple executor: it polls the future in a busy loop until completion
 ///
 /// This is intended as a drop-in replacement for `futures::block_on`, which Kani cannot handle.
+/// Whereas a clever executor like `block_on` in `futures` or `tokio` would interact with the OS scheduler
+/// to be woken up when a resource becomes available, this is not supported by Kani.
+/// As a consequence, this function completely ignores the waker infrastructure and just polls the given future in a busy loop.
 pub fn block_on<T>(mut fut: impl Future<Output = T>) -> T {
     let waker = unsafe { Waker::from_raw(NOOP_RAW_WAKER) };
     let cx = &mut Context::from_waker(&waker);
--- a/library/kani/src/futures.rs
+++ b/library/kani/src/futures.rs
@@ -10,12 +10,14 @@ use std::{
 /// Polls the future in a loop until completion (a very simple executor)
 ///
 /// This is intended as a drop-in replacement for `futures::block_on`, which Kani cannot handle.
-pub fn block_on<F: Future>(mut fut: F) -> <F as Future>::Output {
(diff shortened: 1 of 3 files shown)


library/kani/src/futures.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
danielsn, line 3: File level comment explaining what this is

library/kani/src/futures.rs
```diff
@@ -0,0 +1,38 @@
...
+    loop {
+        match fut.as_mut().poll(cx) {
+            std::task::Poll::Ready(res) => return res,
+            std::task::Poll::Pending => continue,
+        }
+    }
+}
+
+/// A dummy waker, which is needed to call [`Future::poll`]
+const NOOP_RAW_WAKER: RawWaker = {
```
danielsn, line 28: Can this be overriden?

library/kani/src/futures.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, RawWaker, RawWakerVTable, Waker},
+};
+
+/// Polls the future in a loop until completion (a very simple executor)
```
danielsn, line 10: Explain a little more what this does and doesn't do

library/kani/src/futures.rs
fzaiser: I'm not sure what you mean? It's a constant, so it cannot be changed.

library/kani/src/futures.rs
fzaiser: As for what it does, I think the comment explains that already. What do you think is missing?

As for what it doesn't do: clever executors will only poll the future again if the waker that was passed in the previous `poll` call is invoked. But Kani cannot just sleep/be idle for a certain amount of time, so busy-looping is the only option. If we're doing that, wakers don't help, so we don't need to support this. Being able to ignore most of the `waker` stuff also makes things easier for CBMC.

library/kani/src/futures.rs
danielsn: My question was really "should there be a way to override this"

fzaiser: I don't foresee a way to take advantage of the waker for now in the context of Kani. Normal execution makes use of threads that are woken up by the OS (if I understand correctly). But that is not an option for Kani. In any case, if we add another executor, it will need a different name than `block_on` and will be able to use a different waker, if necessary.

library/kani/src/futures.rs
fzaiser: Added

library/kani/src/futures.rs
fzaiser: I've added some explanation to the documentation for that function.

danielsn (approved):

PR #1428, opened 2022-07-29
Audit all assertions that Kani emits and ensure we do not continue execution past a panic

### LINKED ISSUE #1056: Assertions audit
I don't think this is release blocking, but we should do a general audit of all the assertions we generate. I believe we're very inconsistent in how we do this.

Inspired by this PR https://github.com/model-checking/kani/pull/1053

I can see at least:

* `codegen_assert`
* `codegen_assert_false`
* `codegen_panic`
* `codegen_fatal_error`
* (`codegen_cover*` maybe?)
* `codegen_unimplemented`

It's not fully clear to me exactly what each of these should be doing or why they're different.

And that makes me suspect we're not using these consistently or correctly. For example, I think we definitely `fatal_error` some `unimplemented`.

### LINKED ISSUE #1208: Overflow checks should abort the execution

### LINKED ISSUE #1349: index out of bounds exceptions don't assume(false)
This is apparent in the `docs/src/tutorials/loops-unwinding` example. Where we get both a bounds and a pointer failure:

```
SUMMARY: 
 ** 2 of 350 failed
Failed Checks: index out of bounds: the length is less than or equal to the given index
 File: "./src/bounds_check.rs", line 11, in bounds_check::get_wrapped
 File: "./src/bounds_check.rs", line 11, in bounds_check::get_wrapped

VERIFICATION:- FAILED
```

gen-c shows:

```
/* [KANI_REACHABILITY_CHECK] KANI_CHECK_ID_loops_unwinding.6ff84bc3::loops_unwinding_0 */
    assert(!1);
    if(var_18 == 0)
    {
      /* [KANI_CHECK_ID_loops_unwinding.6ff84bc3::loops_unwinding_0] index out of bounds: the length is less than or equal to the given index */
      assert(0);
      goto bb8;
    /* unreachable code */
    }


  bb8:
    ;
```

in contrast a typical assert shows:

```
      assert(0);
      __CPROVER_assume(0);
```

The result is confusing, since it seems to imply the bounds check isn't good enough and could still causing memory unsafety.

tedinski:
### Description of changes: 

Audit all assertions that Kani generates, and ensure we consistently use cbmc's assert-assume pattern.

1. Concentrates everything assertion/assumption/etc fully in `assert.rs`
2. There are a few `PropertyClass`es that I'm not sure if we want necessarily. Out of scope for this PR however.

### Resolved issues:

* Fixes #1056
* Fixes #1208
* Fixes #1349

### Testing:

* How is this change tested? existing tests, some of which required changes due to new behavior

* Is this a refactor change? yes but also no since wrong-behaviors were identified
### Checklist


kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -127,4 +166,93 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    pub fn codegen_unimplemented_expr(
+        &mut self,
+        operation_name: &str,
+        t: Type,
+        loc: Location,
+        url: &str,
+    ) -> Expr {
+        let body = vec![
+            self.codegen_unimplemented_stmt(operation_name, loc, url),
+            t.nondet().as_stmt(loc).with_location(loc),
```
danielsn, line 185: Do we need the `with_location` if we're passing `loc` in already?

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -2,35 +2,80 @@
...
-    /// Assertions and Panic that are not specific to Kani. In a concrete execution, we expect
-    /// these assertions to be available to the user.
-    /// E.g.: User assertions, compiler invariant checks
+    /// Ordinary (Rust) assertions and panics.
+    ///
+    /// SPECIAL BEHAVIOR: None, this is the default type of assertion.
     Assertion,
+    /// Another instrinsic check.
+    ///
+    /// SPECIAL BEHAVIORL None TODO: Why should this exist?
```
adpaco-aws, line 52: Saw this one is being used for all checks in `exact_div`. In principle, if we decided to keep it, I think it should be used only for the cases where an exact division cannot be done. Other cases (division by zero, division does not overflow) are better represented by `ArithmeticOverflow`. At the same time, these are all being introduced by Kani, so we could get replace all of them with `SafetyCheck`.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -2,35 +2,80 @@
...
     ExactDiv,
+    /// The `kani::expect_fail` assertion.
+    ///
+    /// SPECIAL BEHAVIOR: Inverted interpretation of success/failure.
+    /// (Note: Possibly this should be the same as Cover, or perhaps an ExplicitCover?)
     ExpectFail,
+    /// Another instrinsic check.
+    ///
+    /// SPECIAL BEHAVIOR: None TODO: Why should this exist?
```
adpaco-aws, line 61: This is another instance of a check that we include to detect undefined behavior coming from intrinsics. I think that's the purpose of `SafetyCheck` (i.e., checks added by Kani to detect the violation of safety conditions), so we could use that instead.

adpaco-aws (commented):
Wanted to give some feedback before you do the remaining work here.

My impression is that `SafetyCheck` was an early attempt to identify all the checks the Kani compiler introduces during code generation. In the case of intrinsics, you could replace most "special" cases with that. But we also check for arithmetic overflows, so when to use `ArithmeticOverflow` over `SafetyCheck`?

celinval (commented):
Just a few notes on the variants.

tedinski:
Here are my notes on the assertions audit.

1. I believe it's complete. We should now be consistently asserting on things.
2. I've decided the `PropertyClass` cleanup is out of scope for this PR. I have TODO notes in the comments towards that, but disentangling that is a cleanliness issue without significant impact at this point.

I chased assertion/assumption/etc related code up from the base types in `cprover_bindings` to ensure everything in Kani goes through `assert.rs`. Then I examined every use of things in `assert.rs`. Here's the new state of things:

BuiltinFn
- Abort - dead code
- Assert - dead code
- CProverAssume - dead code
- CProverCover - Used by `Stmt::cover`... which is dead code

StmtBody
- Assert - `Stmt::assert`
- Assume - `Stmt::assume`
- ::assert - `assert.rs` `codegen_*` only
- ::assert_false - `assert.rs` `codegen_*` only
- ::assume  -  `assert.rs` `codegen_*` only
- ::cover - dead code


assert.rs
- codegen_assert - outside this file, used once: "expect_fail" hook (we don't yet have `kani::check`)
- codegen_assume - outside this file, used once: "assume" hook
- codegen_assert_assume - used in:
  * "assert" hook
  * TerminatorKind::assert
  * many intrinsics' checks
- codegen_assert_assume_false - used in:
  * SafetyCheck: "unstable vtable comparison"
  * TerminatorKind::Unreachable
- codegen_panic - used in "panic" hook only
- codegen_fatal_error - used in intrinsics:
  * "abort"
  * "transmute uninhabited" SafetyCheck
  * Unsupported intrinsic
  * Safety checks for `assert_inhabited` `_uninit_valid` `_zero_valid`
- codegen_cover (only used in below right now)
- codegen_cover_loc - only used in:
  * "assert" hook
  * TerminatorKind::assert
- codegen_unimplemented_expr - used in operand.rs, rvalue.rs
- codegen_unimplemented_stmt - used in intrinsic.rs, rvalue.rs, statement.rs
- codegen_mimic_unimplemented - used in functions we skip codegenning, and TerminatorKind::Resume and Abort (which we should never actually hit because we don't "unwind" on panics)
- codegen_sanity - used in:
  * codegen_vtable_size
  * codegen_end_call
  * codegen_virtual_funcall

----

I think there's a followup task for some renaming that could be done here. `codegen_assert` should now be `codegen_check` and `codegen_assert_assume` can then take `codegen_assert`. But that can be for the future. I think it'd make this PR worse, as it'd confuse what's happening with some of the changes.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -95,36 +178,94 @@ impl<'tcx> GotocCtx<'tcx> {
...
-        &self,
-        property_class: PropertyClass,
-        msg: &str,
-        span: Option<Span>,
+    /// Kani does not currently support all MIR constructs.
+    ///
+    /// This action will
+    ///
+    /// 1. Fail verification in a machine-detectable manner if reachable
+    /// 2. Warn about unsupported features at compile-time
```
danielsn, line 186: Maybe comment how we return a `nondet` in `Expr` context?

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -95,36 +178,94 @@ impl<'tcx> GotocCtx<'tcx> {
...
-        // unless it is run with '--cover cover' (see
-        // https://github.com/diffblue/cbmc/issues/6613). So for now use
-        // assert(!cond).
-        self.codegen_assert(cond.not(), PropertyClass::Cover, msg, loc)
+    /// There are a handful of location where we want to codegen unimplemented... but also
+    /// not really report these statically to the user on compilation. This does exactly
+    /// the same thing as `codegen_unimplemented_stmt` but doesn't add it to the list
+    /// of `unsupported_constructs`
+    ///
+    /// TODO: Ideally we'd eliminate this. Currently used in two places:
```
danielsn, line 234: Tracking issue

tests/ui/cbmc_checks/signed-overflow/check_message.rs
```diff
@@ -14,12 +14,14 @@ fn dummy(var: i32) {
...
-    dummy(any::<i32>() / any::<i32>()); // This is not emitting CBMC check.
-    dummy(any::<i32>() % any::<i32>()); // This is not emitting CBMC check.
-    dummy(any::<i32>() << any::<i32>());
-    dummy(any::<i32>() >> any::<i32>());
-    dummy(-any::<i32>()); // This is not emitting CBMC check.
+    match kani::any() {
+        0 => dummy(any::<i32>() + any::<i32>()),
+        1 => dummy(any::<i32>() - any::<i32>()),
+        2 => dummy(any::<i32>() * any::<i32>()),
+        3 => dummy(any::<i32>() / any::<i32>()),
```
danielsn, line 21: Is this now emitting a check

tests/expected/pointer-overflow/expected
```diff
@@ -1,2 +1,2 @@
 Status: FAILURE\
-arithmetic overflow on signed multiplication
+offset: attempt to compute number in bytes which would overflow
```
danielsn, line 2: What does "number in bytes" mean?

tests/expected/array/expected
```diff
@@ -4,5 +4,3 @@ SUCCESS\
 assertion failed: y[1] == 2
 FAILURE\
 index out of bounds: the length is less than or equal to the given index
-FAILURE\
-assertion failed: y[z] == 3
```
danielsn, line 8: What happened to this check

tests/expected/arith_checks/expected
```diff
@@ -1,2 +1 @@
 Failed Checks: attempt to subtract with overflow
-Failed Checks: internal error: entered unreachable code: Previous statement should fail
```
danielsn, line 2: What happened to this check

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -149,13 +150,13 @@ impl<'tcx> GotocHook<'tcx> for Assert {
 
         // Since `cond` might have side effects, assign it to a temporary
         // variable so that it's evaluated once, then assert and assume it
+        // TODO: I don't think `cond` can have side effects, this is MIR, it's going to be temps
```
danielsn, line 153: Can we just assert that? https://github.com/model-checking/kani/blob/632fda645364790ada91b5e9980731638c7fc8b7/cprover_bindings/src/goto_program/expr.rs#L316

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -380,16 +374,14 @@ impl<'tcx> GotocCtx<'tcx> {
                         // https://github.com/model-checking/kani/issues/426
                         // Unblocks: https://github.com/model-checking/kani/issues/435
                         if Expr::typecheck_call(&func, &args) {
-                            func.call(args)
+                            func.call(args).as_stmt(Location::none())
```
danielsn, line 377: No location?

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -328,15 +326,11 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                                        format!("Unsupported drop unsized struct: {:?}", typ)
-                                            .as_str(),
-                                        Type::Empty,
-                                        Location::None,
-                                        "https://github.com/model-checking/kani/issues/1072",
-                                    )
-                                    .as_stmt(Location::None);
+                                return self.codegen_unimplemented_stmt(
+                                    format!("Unsupported drop unsized struct: {:?}", typ).as_str(),
+                                    Location::None,
```
danielsn, line 331: Is there no location we can use here?

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -844,14 +841,11 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                    .codegen_unimplemented(
-                        format!("drop_in_place for {}", drop_sym_name).as_str(),
-                        Type::empty(),
-                        Location::none(),
-                        "https://github.com/model-checking/kani/issues/281",
-                    )
-                    .as_stmt(Location::none());
+                let unimplemented = ctx.codegen_unimplemented_stmt(
+                    format!("drop_in_place for {}", drop_sym_name).as_str(),
+                    Location::none(),
```
danielsn, line 846: No location to use here?

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -96,13 +95,14 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                    None,
-                ),
-                Some(self.current_fn().mir().span),
+            let loc = self.codegen_span(&self.current_fn().mir().span);
+            let readable_name = format!("The function {}", self.current_fn().readable_name());
+            // We'll ideally just get rid of this eventually, but use "mimic" to avoid extra compilation warnings
+            let body = self.codegen_mimic_unimplemented(
+                &readable_name,
+                loc,
+                // There actually are links to specific issues in `should_skip_current_fn`...
```
danielsn, line 104: Should we keep a map from fn-name to url?

danielsn: Make `should_skip_current_fn` 

```rust
    fn should_skip_current_fn(&self) -> Option<&str> {
        let current_fn_name =
            with_no_trimmed_paths!(self.tcx.def_path_str(self.current_fn().instance().def_id()));
        // We cannot use self.current_fn().readable_name() because it gives the monomorphized type, which is more difficult to match on.
        // Ideally we should not rely on the pretty-printed name here. Tracked here: https://github.com/model-checking/kani/issues/1374
        match current_fn_name.as_str() {
            // https://github.com/model-checking/kani/issues/202
            "fmt::ArgumentV1::<'a>::as_usize" => Some("https://github.com/model-checking/kani/issues/202"),
           // ...
            _ => None,
        }
    }
```

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -50,11 +103,15 @@ impl<'tcx> GotocCtx<'tcx> {
         message: &str,
         loc: Location,
     ) -> Stmt {
-        assert!(cond.typ().is_bool());
         let property_name = property_class.as_str();
         Stmt::assert(cond, property_name, message, loc)
     }
 
+    /// Generates a CBMC assumption.
+    pub fn codegen_assume(&self, cond: Expr, loc: Location) -> Stmt {
```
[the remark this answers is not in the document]
tedinski: Yes: Several of these don't actually use `self` but some could in the future. I thought about this and decided that it might be best to keep them all regular. (In fact, the weirdest one was introducing `codegen_assume`, but I felt that having NO references to the `cprover_bindings` assert/assume/etc stuff outside of this file was a consistency worth having...)

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -63,24 +120,50 @@ impl<'tcx> GotocCtx<'tcx> {
...
         message: &str,
         loc: Location,
     ) -> Stmt {
-        // Convert Property Class to String
-        let property_name = property_class.as_str();
-        Stmt::assert_false(property_name, message, loc)
+        self.codegen_assert_assume(Expr::bool_false(), property_class, message, loc)
+    }
+
+    /// A shorthand for assert-assume(false) that takes a MIR `Span` instead of a CBMC `Location`.
```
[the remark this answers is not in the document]
tedinski: Indeed, this is a little bit messy. I was also surprised slightly by choices of Expr/Stmt for some of these.

This is one of those little areas where it frustrates me that I can't spend more time _just_ doing code cleanup work...

tests/expected/pointer-overflow/expected
tedinski: Hah, good question. That's the assertion message that's there though, not part of this change.

(AFAICT, the change to this expected file is that it was previously expecting the "wrong" assertion which is no longer reached now that the "right" one stops control flow)

tests/expected/array/expected
tedinski: Control flow stops at the previous assertion failure. The "check" is still there, but now succeeds. (In the case where it would fail, it fails on the earlier assertion, and it's not reported as "unreachable" because paths that would succeed for both checks still exist.)

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
tedinski: This code bothered me, but it's a distraction from the point of this PR, but I couldn't help but shove a comment in there for later...

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tedinski: I'm fixing missing locations in some of the statement stuff, but apparently here is vtable generation we don't really have a location?

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
tedinski: I thought about sneaking this change in here but there are some other un-linked skips in there, and it seemed like too much to figure out now.

tedinski:
I opened a tracking issue for PropertyClass stuff: https://github.com/model-checking/kani/issues/1551

Propagated locations a bit more, fixed some comments.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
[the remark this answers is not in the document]
celinval: You could also just use pure functions without the `self`. I do agree that for consistency sake, we might as well not use `Stmt` directly.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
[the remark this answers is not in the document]
danielsn: Maybe we should have a "code cleanup sprint" one of these days

danielsn (approved):
A couple places where we could still use tracking issues, and some work for future PRs.
Otherwise LGTM.

PR #1429, opened 2022-07-29
Fix some clippy warnings

fzaiser:
### Description of changes: 

Fix some clippy warnings.

### Testing:

* How is this change tested? No tests needed (refactor)

### Checklist


cprover_bindings/src/goto_program/expr.rs
```diff
@@ -345,10 +345,8 @@ impl Expr {
             source.is_numeric()
         } else if target.is_pointer() {
             source.is_integer() || source.is_pointer()
-        } else if target.is_empty() {
-            true
         } else {
-            false
+            target.is_empty()
```
danielsn, line 349: This is the only change I don't love, because it messes with the parallel pattern in the `else if`s .

danielsn (commented):
Looks good, other than one change I'm unsure if its an improvement

cprover_bindings/src/goto_program/expr.rs
fzaiser: Alright, I can revert that one.

danielsn (approved):

PR #1430, opened 2022-08-01
Implement `#[kani::async_proof]` attribute

fzaiser:
### Description of changes: 

This adds an attribute `#[kani::async_proof]` to make model-checking `async` functions more ergonomic. Before, one had to manually call `kani::block_on` on an `async` function.

### Call-outs:

This adds the dependencies `syn` and `quote` to `kani_macros`.

### Testing:

* How is this change tested? Added a regression test.

### Checklist

### DIFF
--- a/library/kani_macros/src/lib.rs
+++ b/library/kani_macros/src/lib.rs
@@ -44,7 +44,7 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
 pub fn proof(attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut result = TokenStream::new();
 
-    assert!(attr.to_string().is_empty(), "#[kani::proof] does not take any arguments");
+    assert!(attr.is_empty(), "#[kani::proof] does not take any arguments");
     result.extend("#[kanitool::proof]".parse::<TokenStream>().unwrap());
     // no_mangle is a temporary hack to make the function "public" so it gets codegen'd
     result.extend("#[no_mangle]".parse::<TokenStream>().unwrap());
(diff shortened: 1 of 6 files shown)


library/kani_macros/src/lib.rs
```diff
@@ -51,6 +56,40 @@ pub fn proof(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+
+#[cfg(kani)]
+#[proc_macro_attribute]
+pub fn async_proof(attr: TokenStream, item: TokenStream) -> TokenStream {
+    assert!(attr.to_string().is_empty(), "#[kani::async_proof] does not take any arguments");
+    let fn_item = parse_macro_input!(item as ItemFn);
+    let attrs = fn_item.attrs;
+    let vis = fn_item.vis;
+    let sig = fn_item.sig;
+    assert!(sig.asyncness.is_some(), "#[kani::async_proof] can only be applied to async functions");
```
danielsn, line 73: Should probably have a test for this

library/kani_macros/src/lib.rs
```diff
@@ -51,6 +56,40 @@ pub fn proof(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+#[cfg(not(kani))]
+#[proc_macro_attribute]
+pub fn async_proof(attr: TokenStream, item: TokenStream) -> TokenStream {
+    proof(attr, item) // Treat like a normal #[kani::proof] if Kani is not active
+}
+
+#[cfg(kani)]
+#[proc_macro_attribute]
+pub fn async_proof(attr: TokenStream, item: TokenStream) -> TokenStream {
+    assert!(attr.to_string().is_empty(), "#[kani::async_proof] does not take any arguments");
```
danielsn, line 68: This is true for now, but will probably not be true in the future if we want to control which executor we use

library/kani_macros/src/lib.rs
```diff
@@ -51,6 +56,40 @@ pub fn proof(attr: TokenStream, item: TokenStream) -> TokenStream {
...
 
+#[cfg(not(kani))]
+#[proc_macro_attribute]
+pub fn async_proof(attr: TokenStream, item: TokenStream) -> TokenStream {
+    proof(attr, item) // Treat like a normal #[kani::proof] if Kani is not active
+}
+
+#[cfg(kani)]
+#[proc_macro_attribute]
+pub fn async_proof(attr: TokenStream, item: TokenStream) -> TokenStream {
```
danielsn, line 67: This macro is more extensive than some of the others, and could probably use a comment.

library/kani_macros/src/lib.rs
fzaiser: Agreed. I'll mention that this may change in the future.

library/kani_macros/src/lib.rs
fzaiser: Added.

library/kani_macros/src/lib.rs
fzaiser: Added.

danielsn (approved):

camshaft:
Do we actually need a separate attribute for async? Couldn't the existing `proof` macro look to see if the function has an `async` and do the right thing if it does?

fzaiser:
> Do we actually need a separate attribute for async? Couldn't the existing proof macro look to see if the function has an async and do the right thing if it does?

Thanks for the suggestion! I had considered that, but thought it would be better to keep them separate: we expect to extend the attribute later on to allow the user to choose different Kani executors: `#[kani::async_proof(executor = "...")`, similarly to tokio's `#[tokio::test(flavor = "...")]`. Such arguments don't make sense for non-`async` functions, so I thought it would be best to keep them separate. Happy to be convinced otherwise though :)

fzaiser:
Opened #1464 to discuss

PR #1433, opened 2022-08-01
Replace CBMC output parser

### LINKED ISSUE #493: Kani produces no output on non-termination

### LINKED ISSUE #1431: New parser doesn't produce colored output
The new parser misses this feature where status as reported with green, red and yellow colors. Fix this!

adpaco-aws:
### Description of changes: 

Replaces the `cbmc_json_parser.py` script with the Rust module `cbmc_json_parser.rs`.

In contrast to the old parser, this one doesn't wait to read from a file. Instead, it pipes CBMC output into Kani, which allows us to perform parsing on the fly. The main advantage of doing this is that users will get feedback even if CBMC is stuck in automatic unwinding (i.e., this solves #493).

While the basic functionality is being thoroughly tested by our regression tests, there are some places where we aren't doing the exact same thing as in the old script. See the "Callouts" section for more details.

Here's the output for a non-terminating example:

Screen Shot 2022-08-01 at 16 37 37

For the test in `tests/expected/assert-eq/main.rs` (`regular` output format):
Screen Shot 2022-08-03 at 14 57 28

For the test in `tests/expected/assert-eq/main.rs` (`terse` output format):
Screen Shot 2022-08-03 at 14 57 56

### Resolved issues:

Resolves #493 
Resolves #1431 

### Call-outs:
 * The new parser relays on `diff_path` in order to print locations, which are relative to the current working directory. The old script had different branches based on the location of the involved files (i.e., it distinguished between files located below the current working directory, the user directory and others).
 * ~~Colored results are gone, but will come back soon: https://github.com/model-checking/kani/issues/1431~~
 * Messages produces after looking a result should be handled as regular messages. I've some idea on how to handle this, but will have to wait until then: https://github.com/model-checking/kani/issues/1432
 * Some Python regexes used to match on Kani reachability IDs were too greedy in Rust, therefore I changed them to NOT match on square brackets, which gives the expected result.
 * Had to change a `dry-run` regression because now we don't produce a CBMC output file.
 * I'm sure there are places where we can use more idiomatic Rust. Please point those out!

### Testing:

* How is this change tested? Existing regression. In addition, I used an ad-hoc script that `diff`s the output between the old and new versions. No "Status" lines appeared there.

* Is this a refactor change? Mostly yes.

### Checklist


kani-driver/src/call_cbmc.rs
```diff
@@ -45,7 +45,7 @@ impl KaniSession {
         } else {
             // Spawn the CBMC process and process its output below
             let cbmc_process = self.run_piped(cmd)?;
-            if cbmc_process.is_some() {
+            if let Some(..) = cbmc_process {
```
fzaiser, line 48: I think the point of the clippy warning is to get rid of the `.unwrap()`. So the code should be `if let Some(cbmc_process) = cbmc_process`. Then you can get rid of the unwrap below.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -287,7 +287,7 @@ fn filepath(file: String) -> String {
     let cur_dir = env::current_dir().unwrap();
 
     let diff_path = diff_paths(file_path, cur_dir);
-    if diff_path.is_some() {
+    if let Some(..) = diff_path {
```
fzaiser, line 290: see above (unwrap)

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -403,7 +403,7 @@ impl<'a, 'b> Parser<'a, 'b> {
     fn parse_item(&self) -> ParserItem {
         let string_without_delimiter = &self.input_so_far.as_str()[0..self.input_so_far.len() - 2];
         let block: Result<ParserItem, _> = serde_json::from_str(string_without_delimiter);
-        if block.is_ok() {
+        if let Ok(..) = block {
```
fzaiser, line 406: see above (unwrap)

kani-driver/src/call_cbmc.rs
adpaco-aws: Right, thanks! These should all be fixed now, please let me know if you spot others :D

adpaco-aws:
> Thank you so much for this! Just so we have a fair idea of what the output looks like now, can you add a few screenshots to the PR description?

Screenshots added!
They correspond to one test (with output format `regular` and `terse`) and the non-terminating example:

```rust
fn rev(mut x: [i32; 4]) -> [i32; 4] {
    x.reverse();
    x
}

#[cfg(kani)]
#[kani::proof]
fn example() {
    let var = kani::any();
    let reversed = rev(var);
    let double_reversed = rev(reversed);

    assert_eq!(var, double_reversed);
}
```

The main differences are highlighted in the call-outs section (e.g., no color). Other than that, I've tried to keep it as faithful as possible.

kani-driver/src/session.rs
```diff
@@ -140,6 +137,26 @@ impl KaniSession {
...
+    /// Returns an error if the process could not be spawned
+    pub fn run_piped(&self, mut cmd: Command) -> Result<Option<Child>> {
+        if self.args.verbose || self.args.dry_run {
+            println!("{}", render_command(&cmd).to_string_lossy());
+            if self.args.dry_run {
+                return Ok(None);
+            }
+        }
+        // Run the process as a child process
+        let process = cmd.stdout(Stdio::piped()).spawn();
```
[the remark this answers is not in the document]
adpaco-aws: Well, I don't think it's risk-free. But I haven't found any issues so far.

For example, I think the scenario you mention would correspond to running a non-terminating example through Kani. I've tried doing that locally and it's still going on after 45mins. Kani's memory % is below 0.1, and CBMC's memory % has slowly increased to 4.3% (it's doing automatic unwinding).

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,974 @@
...
+    ProverStatus {
+        _c_prover_status: String,
+    },
+}
+
+impl ParserItem {
+    /// Determines if an item must be skipped or not.
+    fn must_be_skipped(&self) -> bool {
+        matches!(&self, ParserItem::Message { message_text, .. } if message_text.starts_with("Building error trace"))
+            || matches!(&self, ParserItem::Message { message_text, .. } if message_text.starts_with("VERIFICATION"))
```
zhassan-aws, line 215: Can the two conditions be combined, e.g.:
```
         matches!(&self, ParserItem::Message { message_text, .. } if message_text.starts_with("Building error trace")) || message_text.starts_with("VERIFICATION")
```
?

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,974 @@
...
+
+/// `Display` implement for `SourceLocation`.
+///
+/// This is used to format source locations for individual checks. But source
+/// locations may be printed in a different way in other places (e.g., in the
+/// "Failed Checks" summary at the end).
+impl std::fmt::Display for SourceLocation {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut fmt_str = String::new();
+        if self.file.is_some() {
```
zhassan-aws, line 261: Use `if let Some`

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,974 @@
...
+            let file_str = filepath(self.file.clone().unwrap()).to_string();
+            fmt_str.push_str(file_str.as_str());
+            if self.line.is_some() {
```
zhassan-aws, line 264: Same here

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,974 @@
...
+                let line_str = format!(":{}", self.line.clone().unwrap());
+                fmt_str.push_str(line_str.as_str());
+                if self.column.is_some() {
```
zhassan-aws, line 267: Same here.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,974 @@
...
+                    let column_str = format!(":{}", self.column.clone().unwrap());
+                    fmt_str.push_str(column_str.as_str());
+                }
+            }
+        } else {
+            fmt_str.push_str("Unknown File");
+        }
+        if self.function.is_some() {
```
zhassan-aws, line 275: Same here.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,974 @@
...
+    fn is_missing(&self) -> bool {
+        self.file.is_none() && self.function.is_none()
+    }
+}
+
+/// `Display` implement for `SourceLocation`.
+///
+/// This is used to format source locations for individual checks. But source
+/// locations may be printed in a different way in other places (e.g., in the
+/// "Failed Checks" summary at the end).
```
zhassan-aws, line 257: Can you describe what the formatted location looks like? It's hard to infer from reading the code.

zhassan-aws (commented):
First round of comments. Only half way through `cbmc_output_parser.rs`.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: Yes, done!

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: Added a description, let me know how it looks!

adpaco-aws:
Added colored output. Will update screenshots in a moment.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,991 @@
...
+/// CBMC's JSON output is defined as a JSON array which contains:
+///  1. One program at the beginning (i.e., a message with CBMC's version).
+///  2. Messages, which can appears anywhere, and are either status messages or error messages.
+///  3. Verification results, another JSON array with all individual checks.
+///  4. Prover status, at the end. Because the verification results depends on
+///     our postprocessing, this is not used.
+///
+/// The parser reads the output line by line. A line may trigger one action, and
+/// the action may return a parsed item.
+struct Parser<'a, 'b> {
```
zhassan-aws, line 358: Would using something like the [StreamDeserializer](https://docs.rs/serde_json/latest/serde_json/struct.StreamDeserializer.html#) be possible here?

adpaco-aws: Good point.

I tried using `StreamDeserializer` for parsing from the buffer in one of my earliest attempts, but it wasn't successful. The problem is that it expects self-delineating values, but the output we process is a JSON array with delineated values. In other words, the output we get is of the form:
```
[
    Program,
    Message,
    ...,
    Message,
    Result,
    Message,
    ProverStatus
]
```
But for `StreamDeserializer` to work we would need an output of the form:
```
Program
Message
...
Message
Result
Message
ProverStatus
```

Note that the parsing logic is mostly concerned with this problem: It assumes it's already within a JSON array, ignoring its delimiters. Changing the CBMC output to have the second form described is a no-go, at least according to the discussions I had then.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,991 @@
...
+///    attribute was formatted.
+impl std::fmt::Display for SourceLocation {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut fmt_str = String::new();
+        if let Some(file) = self.file.clone() {
+            let file_str = filepath(file);
+            fmt_str.push_str(file_str.as_str());
+            if let Some(line) = self.line.clone() {
+                let line_str = format!(":{}", line);
+                fmt_str.push_str(line_str.as_str());
```
fzaiser, line 275: I think it would be cleaner to use `write!(&mut fmt_str, ":{line}")?` instead of these two lines. Similar in the rest of the function (and maybe other instances in the file.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,991 @@
...
+            fmt_str.push_str(fun_str.as_str());
+        }
+
+        write! {f, "{}", fmt_str}
+    }
+}
+
+/// Returns a path relative to the current working directory.
+fn filepath(file: String) -> String {
+    let file_path = PathBuf::from(file.clone());
```
fzaiser, line 295: Can't this be an `.as_ref()` or `.into()` instead of `.clone()`?

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,991 @@
...
+                let item = self.parse_item();
+                self.clear_input();
+                Some(item)
+            }
+        }
+    }
+
+    // Adds a string to the input accumulated so far
+    fn add_to_input(&mut self, input: String) {
+        self.input_so_far.push_str(input.as_str());
```
fzaiser, line 405: can't you change `x.as_str()` to `&x` if `x: String`?

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,991 @@
...
+    }
+
+    // Returns a `ParserItem` from the input we have accumulated so far. Since
+    // all items except the last one are delimited (with a comma), we first try
+    // to parse the item without the delimiter (i.e., the last character). If
+    // that fails, then we parse the item using the whole input.
+    fn parse_item(&self) -> ParserItem {
+        let string_without_delimiter = &self.input_so_far.as_str()[0..self.input_so_far.len() - 2];
```
fzaiser, line 413: isn't the `as_str` redundant here?

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,991 @@
...
+        }
+
+        if show_checks {
+            // TODO: Add color to status if printing to terminal.
+            // <https://github.com/model-checking/kani/issues/1431>
+            let check_id = format!("Check {}: {}\n", index, name);
+            let status_msg = format!("\t - Status: {}\n", status);
+            let descrition_msg = format!("\t - Description: \"{}\"\n", description);
+
+            result_str.push_str(check_id.as_str());
```
fzaiser, line 609: change `as_str()` to `&`

scripts/setup/ubuntu/install_deps.sh
```diff
@@ -49,7 +49,6 @@ sudo DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends --ye
 # Add Python package dependencies
 PYTHON_DEPS=(
   autopep8
-  colorama # Used for introducing colors into terminal output
```
fzaiser, line 52: Nice to get rid of dependencies :)

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: Yes, but then I need to add type annotations as follows:
```
help: try using a fully qualified path to specify the expected types
    |
295 |     let file_path = PathBuf::from(<std::string::String as AsRef<T>>::as_ref(&file));
    |                                   +++++++++++++++++++++++++++++++++++++++++++    ~

For more information about this error, try `rustc --explain E0283`.
```

Doesn't look so readable to me. What do you think?

fzaiser: Leave it as is. Apparently the types are more complicated than what I inferred in my head.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: Yes, thanks! Replaced almost all of them.

zhassan-aws: AFAIK, `x.as_str()` is the same as `&x`: https://users.rust-lang.org/t/converting-between-str-and-string/53595

fzaiser: Yes, but `&x` is easier to read and less noisy IMO

adpaco-aws: Yeah, I wasn't aware of this implicit conversion earlier. It's explained in detail here: https://doc.rust-lang.org/std/string/struct.String.html#deref

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,989 @@
...
+                }
+            }
+        } else {
+            write!(&mut fmt_str, "Unknown file")?;
+        }
+        if let Some(function) = self.function.clone() {
+            write!(&mut fmt_str, " in function {function}")?;
+        }
+
+        write! {f, "{}", fmt_str}
```
fzaiser, line 287: Since we're first writing to `fmt_str` and then write `fmt_str` to `f`, can't we just write to `f` directly?

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,989 @@
...
+        // Some items (e.g., messages) are skipped.
+        // We could also process them and decide to skip later.
+        if item.must_be_skipped() {
+            continue;
+        }
+        let processed_item = process_item(item, extra_ptr_checks, &mut result);
+        // Both formatting and printing could be handled by objects which
+        // implement a trait `Printer`.
+        let formatted_item = format_item(&processed_item, output_format);
+        if formatted_item.is_some() {
```
zhassan-aws, line 528: if let Some

kani-driver/src/cbmc_output_parser.rs
zhassan-aws: I see.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,989 @@
...
+        Parser { input_so_far: String::new(), buffer }
+    }
+
+    /// Triggers an action based on the input:
+    ///  * Square brackets ('[' and ']') will trigger the `ClearInput` action
+    ///    because we assume parsing is done on a JSON array.
+    ///  * Curly closing bracket ('}') preceded by two spaces will trigger the
+    ///    `ProcessItem` action.
+    fn triggers_action(&self, input: String) -> Option<Action> {
+        if input.starts_with('[') || input.starts_with(']') {
```
zhassan-aws, line 372: This may ignore any characters that appear after `[` or `]`. If CBMC's current output never includes characters, perhaps we should assert that the line only has `[` or `]`?

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,989 @@
...
+/// temporary variable in their descriptions.
+fn get_readable_description(property: &Property) -> String {
+    let original = property.description.clone();
+    let class_id = extract_property_class(property).unwrap();
+
+    let description_alternatives = CBMC_ALT_DESCRIPTIONS.get(class_id);
+    if let Some(alt_descriptions) = description_alternatives {
+        for (desc_to_match, opt_desc_to_replace) in alt_descriptions {
+            if original.contains(desc_to_match) {
+                if opt_desc_to_replace.is_some() {
```
zhassan-aws, line 811: if let Some

zhassan-aws (approved):
This is a huge effort! Thanks @adpaco-aws!

Only have a few more minor comments.

kani-driver/src/call_cbmc.rs
```diff
@@ -103,6 +109,10 @@ impl KaniSession {
 
         args.push("--slice-formula".into());
 
+        if self.args.output_format != OutputFormat::Old {
+            args.push("--json-ui".into());
+        }
```
tedinski, line 114: Does this impact `--visualize`?

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,989 @@
...
+use std::{
+    collections::HashMap,
+    env,
+    io::{BufRead, BufReader},
+    path::PathBuf,
+    process::{Child, ChildStdout},
+};
+use structopt::lazy_static::lazy_static;
+
+lazy_static! {
```
tedinski, line 37: I do not want this to block this PR, but IIRC `lazy_static` isn't really maintained anymore and I forget what crate is supposed to replace it. `once` or `oncecell` or something like that.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,989 @@
...
+        if show_checks {
+            // TODO: Add color to status if printing to terminal.
+            // <https://github.com/model-checking/kani/issues/1431>
+            let check_id = format!("Check {}: {}\n", index, name);
+            let status_msg = format!("\t - Status: {}\n", status);
+            let description_msg = format!("\t - Description: \"{}\"\n", description);
+
+            result_str.push_str(&check_id);
+            result_str.push_str(&status_msg);
+            result_str.push_str(&description_msg);
```
tedinski, line 609: I think you can use `write!` on a `String`, but I'd not worry about this now, do a follow-up refactoring!

tedinski (approved):
I like your description of how you tested it. Seems like it should be pretty high-confidence?

I think there's some things that could be improved, but nothing I think should block this improvement. Do you want to merge before release tomorrow?

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,989 @@
...
+    /// Triggers an action based on the input:
+    ///  * Square brackets ('[' and ']') will trigger the `ClearInput` action
+    ///    because we assume parsing is done on a JSON array.
+    ///  * Curly closing bracket ('}') preceded by two spaces will trigger the
+    ///    `ProcessItem` action.
+    fn triggers_action(&self, input: String) -> Option<Action> {
+        if input.starts_with('[') || input.starts_with(']') {
+            return Some(Action::ClearInput);
+        }
+        if input.starts_with("  }") {
```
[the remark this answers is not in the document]
adpaco-aws: The spaces are important here: Because we're iterating over a JSON array, matching on this specific string guarantees that we'll always get an item when we attempt to process a parser item. Otherwise, we wouldn't have this guarantee, which would likely result in a slow-down of the parser.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -0,0 +1,989 @@
...
+            return Some(Action::ProcessItem);
+        }
+        None
+    }
+
+    /// Clears the input accumulated so far.
+    fn clear_input(&mut self) {
+        self.input_so_far = String::new();
```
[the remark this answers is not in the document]
adpaco-aws: Right, thanks!

kani-driver/src/call_cbmc.rs
adpaco-aws: Thanks for pointing this out. It wasn't clear to me: `--visualize` still works, but I don't know if it's adding more files. So I reverted this change and added a comment.

kani-driver/src/cbmc_output_parser.rs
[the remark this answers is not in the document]
zhassan-aws: But can we guarantee that CBMC's JSON format will not change in a way that would break the current matching (that uses a specific number of spaces)?

adpaco-aws:
Addressed all comments except "use `write!` on strings directly", which I'm leaving as a refactoring task in #1480

kani-driver/src/cbmc_output_parser.rs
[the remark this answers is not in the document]
tedinski: Since we pin to a specific cbmc version in our releases, we'd at least get notification that it broke in all our tests.

But I do think this explanation of what's going on here should be added as a comment on this line of code.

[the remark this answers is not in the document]
adpaco-aws: > But can we guarantee that CBMC's JSON format will not change in a way that would break the current matching (that uses a specific number of spaces)?

We don't guarantee that. This could be changed to make it more resilient, but in my opinion it's better to break here than in some other places.

Added a comment to explain this.

PR #825, opened 2022-02-11
Override the print macros with a version suitable for verification

zhassan-aws:
### Description of changes: 

Currently any programs containing (reachable) print macros (e.g. `println`) result in verification failure (see #824). This PR overrides std's print macros by providing versions that are suitable for verification.

### Resolved issues:

Resolves #194 and #824 

### Call-outs:

The overriden versions of the macros do not properly parse the arguments, so Kani may accept programs that are rejected by rustc, e.g.
```rust
fn main() {
    println!("Two arguments expected, but only one provided: {} {}", 4+5);
}
```

### Testing:

* How is this change tested? Added two new tests.

* Is this a refactor change? No

### Checklist
- [X] Each commit message has a non-empty body, explaining why the change was made
- [X] Methods or procedures are documented
- [X] Regression or unit tests are included, or existing tests cover the modified code
- [X] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

Issue #803, opened 2022-02-07
Kani's assert macro implementation bypasses some compile-time checks


zhassan-aws:
With PR #797, Kani's implementation of the assert macro's second form that accepts a custom message bypasses several compile-time checks. 

For example, while `rustc` errors out on the following program:

```rust
fn main() {
    assert!(1 + 1 == 3, "Bad arithmetic: {}", foo);
}
```
output:
```
error[E0425]: cannot find value `foo` in this scope
 --> no_display.rs:2:47
  |
2 |     assert!(1 + 1 == 3, "Bad arithmetic: {}", foo);
  |                                               ^^^ not found in this scope
```
kani does not complain about the non-existence of the variable `foo`:
```
CBMC version 5.48.0 (cbmc-5.48.0) 64-bit x86_64 linux
Reading GOTO program from file no_display.goto
Generating GOTO Program
Adding CPROVER library (x86_64)
Removal of function pointers and virtual functions
Generic Property Instrumentation
Running with 16 object bits, 48 offset bits (user-specified)
Starting Bounded Model Checking
Runtime Symex: 0.00102581s
size of program expression: 75 steps
simple slicing removed 3 assignments
Generated 3 VCC(s), 1 remaining after simplification
Runtime Postprocess Equation: 7.25e-06s
Passing problem to propositional reduction
converting SSA
Runtime Convert SSA: 0.000213004s
Running propositional reduction
Post-processing
Runtime Post-process: 2.594e-05s
Solving with MiniSAT 2.2.1 with simplifier
100 variables, 33 clauses
SAT checker: instance is SATISFIABLE
Runtime Solver: 9.1981e-05s
Runtime decision procedure: 0.000343287s

** Results:
[assertion.1] Code generation sanity check: Correct CBMC vtable size for StructTag("tag-_4317141983261250885") (MIR type Closure(DefId(2:9147 ~ std[2f3c]::rt::lang_start::{closure#0}), [(), i8, extern "rust-call" fn(()) -> i32, (fn(),)])). Please report failures:
https://github.com/model-checking/kani/issues/new?template=bug_report.md: SUCCESS
[pointer_primitives.1] dead object in OBJECT_SIZE(&temp_0): SUCCESS
/home/ubuntu/examples/no_display.rs function main
[main.assertion.1] line 2 assertion: "Bad arithmetic: {}", foo: FAILURE

** 1 of 3 failed (2 iterations)
VERIFICATION FAILED
```
[...]

PR #1446, opened 2022-08-02
Fix kani's overridding of print macros

### LINKED ISSUE #1358: Compiling `insta` crate fails
Steps to reproduce:
1. `cargo new --lib insta_dep`
2. Add `insta` to `dependencies` in `Cargo.toml`:
```toml
[dependencies]
insta = "1"
```
3. Run `cargo kani`

with Kani version: ff033adf39c

Produces:
```
   Compiling insta v1.15.0
error[E0425]: cannot find value `title` in this scope
  --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/insta-1.15.0/src/output.rs:94:9
   |
94 |         title = style(" Snapshot Differences ").bold(),
   |         ^^^^^ not found in this scope

error[E0425]: cannot find value `title` in this scope
   --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/insta-1.15.0/src/output.rs:117:9
    |
117 |         title = style(" Snapshot Summary ").bold(),
    |         ^^^^^ not found in this scope

error[E0425]: cannot find value `title` in this scope
   --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/insta-1.15.0/src/output.rs:121:34
    |
121 |     println!("{title:━^width$}", title = "", width = width);
    |                                  ^^^^^ not found in this scope

error[E0425]: cannot find value `hint` in this scope
   --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/insta-1.15.0/src/runtime.rs:386:13
    |
386 |             hint = style("To update snapshots run `cargo insta review`").dim(),
    |             ^^^^ not found in this scope

error[E0425]: cannot find value `hint` in this scope
   --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/insta-1.15.0/src/runtime.rs:394:17
    |
394 |                 hint = style(
    |                 ^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
error: could not compile `insta` due to 5 previous errors
```

### LINKED ISSUE #1436: Missing error checking for print macros
I tried this code:

```rust
fn main() {
    println!("{}, {}", "Hello");
}
```

```
kani test.rs
```

with Kani version: 43cb18c5d21

I expected to see this happen: Compilation fails as is the case with `rustc`:
```
$ rustc test.rs 
error: 2 positional arguments in format string, but there is 1 argument
 --> test.rs:3:15
  |
3 |     println!("{}, {}", "Hello");
  |               ^^  ^^   -------

```

Instead, this happened: Verification passed:
```
$ kani test.rs 
Checking harness main...
CBMC 5.61.0 (cbmc-5.61.0)
CBMC version 5.61.0 (cbmc-5.61.0) 64-bit x86_64 linux
Reading GOTO program from file test.out.for-main
Generating GOTO Program
Runtime Symex: 0.000529311s
size of program expression: 38 steps
slicing removed 18 assignments
Generated 0 VCC(s), 0 remaining after simplification
Runtime Postprocess Equation: 1.9921e-05s
VERIFICATION SUCCESSFUL

RESULTS:

SUMMARY: 
 ** 0 of 0 failed

VERIFICATION:- SUCCESSFUL


Verification Time: 0.01128442s
```

zhassan-aws:
### Description of changes: 

The approach introduced in #825 for overriding print macros does not work in general because there are special cases (e.g #1358) that require proper parsing of the arguments. It also introduced issues like https://github.com/model-checking/kani/issues/1436.

This PR replaces the implementation of the print macros with one that uses std's `format_args!` macro to guarantee that parsing is properly done and that any features supported by print macros are supported.

### Resolved issues:

Resolves #1358 
Resolves #1436 

### Call-outs:

Verification performance will likely be impacted if print macros are involved in the code. Hopefully, CBMC's formula slicer will minimize the performance impact.

### Testing:

* How is this change tested? Existing regressions for print macros. Added two more cases for one of them.

### Checklist


celinval:
How bad is the performance degradation? I am not sure CBMC would slice any of this since it is probably handling memory. If we expect this to be sliced off, can we just implement an empty macro or augment the match conditions?

YoshikiTakashima:
I'm curious if you should be doing the same to panics. When compiling proptests, we are getting cases where panic! and other macros cause unused variables because they are being overwritten with kani versions.

zhassan-aws:
> How bad is the performance degradation? 

On the following program:
```rust
fn main() {
    const N: u8 = 100;
    let mut a = 0;
    for _i in 1..N {
        println!("{}", a);
        a += 1;
    }
    assert_eq!(a, N-1);
}
```
Verification time is 1.26 seconds on `main` and 1.86 seconds on this branch, so roughly a 1.5X slowdown.

> I am not sure CBMC would slice any of this since it is probably handling memory. 

The nice thing about `format_args!` is that it doesn't perform heap allocations  as it is meant to produce arguments that are consumed by other functions, e.g. `print`, `format`, `write` (source: https://github.com/rust-lang/rust/blob/4a8d2e3856c0c75c71998b6c85937203839b946d/library/alloc/src/fmt.rs#L505)

> can we just implement an empty macro or augment the match conditions?

I don't know how to do that in a way that preserves side effects as well as handles special cases like named arguments:
```
    println!("Using a named argument: {foo}", foo="Hello world");
```

The only solution I could think of is to use `format_args!`, which is implemented in the compiler, and its implementation is not trivial as it parses the positional arguments, named arguments, matches them to the `{}`, `{:?}` literals in the format expression, etc : https://github.com/rust-lang/rust/blob/e141246cbbce2a6001f3181d3d0f661bbfd9c7ea/compiler/rustc_builtin_macros/src/format.rs#L1173

zhassan-aws:
> When compiling proptests, we are getting cases where panic! and other macros cause unused variables because they are being overwritten with kani versions.

Can you give an example?

fzaiser:
In case it's relevant: we're [currently ignoring the function `fmt::ArgumentV1::<'a>::as_usize` in Kani](https://github.com/model-checking/kani/blob/5246a6d691f7786c8ff383baed8f73e153e8aa07/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs#L30), which seems related to the `format_args!`. Not sure if this has interactions with this change tho.

zhassan-aws:
Ah, thanks for pointing this out @fzaiser! I'll take a look to check whether it affects things.

zhassan-aws:
Commenting it out doesn't seem to have an impact (at least on a test with `println!`).

celinval:
> > When compiling proptests, we are getting cases where panic! and other macros cause unused variables because they are being overwritten with kani versions.
> 
> Can you give an example?

```rust
panic!("Error: {}", msg);
```

YoshikiTakashima:
@zhassan-aws   example is likely cleaner, but here's my case.

`err` is supposed to be used here, but it is not getting used after expansion.

```
 let err =
                        match strategy.new_tree(&mut runner) {
                            Ok(s) => { state = s; break; }
                            Err(e) => e,
                        };
                    gen_tries += 1;
                    if gen_tries > 100 {
                            {
                                kani::panic("\"Strategy passed to check_strategy_sanity failed \\\n                     to generate a value over 100 times in a row; \\\n                     last failure reason: {}\",\nerr");
                            };
                        }
                }
```

zhassan-aws:
OK, I see. Yes, we can do something similar for panics/asserts. This would resolve the warnings as well as https://github.com/model-checking/kani/issues/803.

celinval:
Are you planning to apply these changes to panic and assert in this PR? I still worry about performance, but we can assess that when we actually bump into some problem like that. Thanks

zhassan-aws:
No, was planning to leave those to a separate PR.

celinval (approved):

PR #1400, opened 2022-07-21
Produce summary for unsupported features in top 100 crates experiment

adpaco-aws:
### Description of changes: 

No changes to Kani itself, just adds/changes scripts related to the top 100 crates experiment. In particular, it adds:
 1. A script to filter all unsupported constructs expressions. Before this change, the top-level script printed only the first 3 lines after a warning, which means the data was not complete.
 2. A script to produce a summary (table) from unsupported features data.

The top-level script is called the same way, but now it produces the following output at the end:
```
--- OVERALL STATS ---
10 crates failed to compile
33 crates had warning(s)

SUMMARY - UNSUPPORTED FEATURES
=========================================================
Unsupported feature | Crates impacted | Instances of use
---------------------------------------------------------
                      Rvalue::ThreadLocalRef  |  20 |  33
                         'simd_or' intrinsic  |  14 |  17
                         'simd_eq' intrinsic  |  11 |  13
                         'simd_gt' intrinsic  |  10 |  10
          Constant slice value with 2+ bytes  |   9 |  74
                   'simd_shuffle4' intrinsic  |   6 |  27
                        'simd_xor' intrinsic  |   6 |  12
                        'simd_and' intrinsic  |   6 |  11
                   'simd_shuffle8' intrinsic  |   5 |  20
                                         try  |   4 |  13
                  'simd_shuffle16' intrinsic  |   4 |  12
                   'simd_shuffle2' intrinsic  |   4 |   9
                        'simd_add' intrinsic  |   4 |   9
                     'simd_insert' intrinsic  |   3 |  11
                    'simd_extract' intrinsic  |   3 |   5
               PointerCast::ClosureFnPointer  |   2 |   6
                  'simd_shuffle32' intrinsic  |   1 |  34
  drop_in_place for drop_unimplemented__RINvN |   1 |   1
  drop_in_place for drop_unimplemented__RINvN |   1 |   1
  drop_in_place for drop_unimplemented__RINvN |   1 |   1
  drop_in_place for drop_unimplemented__RINvN |   1 |   1
  drop_in_place for drop_unimplemented__RINvN |   1 |   1
  Unsupported drop unsized struct: std::cell: |   1 |   1
                         'sqrtf64' intrinsic  |   1 |   1
                         'powif64' intrinsic  |   1 |   1
                         'powif32' intrinsic  |   1 |   1
                          'logf64' intrinsic  |   1 |   1
                          'expf64' intrinsic  |   1 |   1
=========================================================
```

### Testing:

[...]

PR #1452, opened 2022-08-04
Support ThreadLocalRef (by treating it like a static variable)

### LINKED ISSUE #157: Thread local statics is currently unimplemented
Requested feature: Thread local statics
Use case: Firecracker vmm crate
https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/vstate/vcpu/mod.rs#L123
https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/mir/enum.Rvalue.html#variant.ThreadLocalRef
https://doc.rust-lang.org/std/macro.thread_local.html
https://doc.rust-lang.org/std/thread/struct.LocalKey.html
Is this a breaking change?
No

Test case:
```rust
thread_local! {
    static FOO: u32 = 42;
}

fn main() {
    FOO.with(|v| assert!(*v == 42));
}
```

### LINKED ISSUE #541: Rvalue::ThreadLocalRef is unimplemented
While compiling the latest version of std library, RMC fails to compile library/std/src/sys_common/thread_info.rs
```rust
#![allow(dead_code)] // stack_guard isn't used right now on all platforms
#![allow(unused_unsafe)] // thread_local with `const {}` triggers this liny

use crate::cell::RefCell;
use crate::sys::thread::guard::Guard;
use crate::thread::Thread;

struct ThreadInfo {
    stack_guard: Option<Guard>,
    thread: Thread,
}

thread_local! { static THREAD_INFO: RefCell<Option<ThreadInfo>> = const { RefCell::new(None) } }

impl ThreadInfo {
    fn with<R, F>(f: F) -> Option<R>
    where
        F: FnOnce(&mut ThreadInfo) -> R,
    {
        THREAD_INFO
            .try_with(move |thread_info| {
                let mut thread_info = thread_info.borrow_mut();
                let thread_info = thread_info.get_or_insert_with(|| ThreadInfo {
                    stack_guard: None,
                    thread: Thread::new(None),
                });
                f(thread_info)
            })
            .ok()
    }
}

pub fn current_thread() -> Option<Thread> {
    ThreadInfo::with(|info| info.thread.clone())
}

pub fn stack_guard() -> Option<Guard> {
    ThreadInfo::with(|info| info.stack_guard.clone()).and_then(|o| o)
}

pub fn set(stack_guard: Option<Guard>, thread: Thread) {
    THREAD_INFO.with(move |thread_info| {
        let mut thread_info = thread_info.borrow_mut();
        rtassert!(thread_info.is_none());
        *thread_info = Some(ThreadInfo { stack_guard, thread });
    });
}
```

```
RUSTFLAGS="-Z trim-diagnostic-paths=no -Z codegen-backend=gotoc --cfg=rmc" RUSTC=rmc-rustc cargo +nightly build -Z build-std --target x86_64-unknown-linux-gnu
```

with RMC version:

I expected to see this happen: The module should be able to compile without any problem.

Instead, this happened:

```
thread 'rustc' panicked at 'not implemented', compiler/rustc_codegen_rmc/src/codegen/rvalue.rs:433:42
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: rustc_codegen_rmc::codegen::rvalue::<impl rustc_codegen_rmc::context::goto_ctx::GotocCtx>::codegen_rvalue
             at /home/ANT.AMAZON.COM/celinval/workspace/rmc-project/rmc/compiler/rustc_codegen_rmc/src/codegen/rvalue.rs:433:42
   4: rustc_codegen_rmc::codegen::statement::<impl rustc_codegen_rmc::context::goto_ctx::GotocCtx>::codegen_statement
             at /home/ANT.AMAZON.COM/celinval/workspace/rmc-project/rmc/compiler/rustc_codegen_rmc/src/codegen/statement.rs:537:60
   5: rustc_codegen_rmc::codegen::block::<impl rustc_codegen_rmc::context::goto_ctx::GotocCtx>::codegen_block
             at /home/ANT.AMAZON.COM/celinval/workspace/rmc-project/rmc/compiler/rustc_codegen_rmc/src/codegen/block.rs:27:32
   6: rustc_codegen_rmc::codegen::function::<impl rustc_codegen_rmc::context::goto_ctx::GotocCtx>::codegen_function::{{closure}}
             at /home/ANT.AMAZON.COM/celinval/workspace/rmc-project/rmc/compiler/rustc_codegen_rmc/src/codegen/function.rs:99:71
   7: core::iter::traits::iterator::Iterator::for_each::call::{{closure}}
             at /home/ANT.AMAZON.COM/celinval/workspace/rmc-project/rmc/library/core/src/iter/traits/iterator.rs:728:29
   8: core::iter::adapters::map::map_fold::{{closure}}
             at /home/ANT.AMAZON.COM/celinval/workspace/rmc-project/rmc/library/core/src/iter/adapters/map.rs:84:21
   9: <core::iter::adapters::enumerate::Enumerate<I> as core::iter::traits::iterator::Iterator>::fold::enumerate::{{closure}}
             at /home/ANT.AMAZON.COM/celinval/workspace/rmc-project/rmc/library/core/src/iter/adapters/enumerate.rs:106:27
  10: core::iter::traits::iterator::Iterator::fold
             at /home/ANT.AMAZON.COM/celinval/workspace/rmc-project/rmc/library/core/src/iter/traits/iterator.rs:2165:21
  11: <core::iter::adapters::enumerate::Enumerate<I> as core::iter::traits::iterator::Iterator>::fold
             at /home/ANT.AMAZON.COM/celinval/workspace/rmc-project/rmc/library/core/src/iter/adapters/enumerate.rs:112:9
  12: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
             at /home/ANT.AMAZON.COM/celinval/workspace/rmc-project/rmc/library/core/src/iter/adapters/map.rs:124:9
  13: core::iter::traits::iterator::Iterator::for_each
 ...
```

### LINKED ISSUE #1450: Support ThreadLocalRef (needed for tokio)

fzaiser:
### Description of changes: 

Kani did not support thread locals, in particular the MIR construct `ThreadLocalRef`. As discovered #1400, this is the most used feature that is currently unsupported in Kani. In particular it is needed to compile tokio.

Since Kani is single-threaded, thread locals are treated like static variables and translated in the same way.

### Resolved issues:

Resolves #1450
Resolves #541
Resolves #157

### Testing:

* How is this change tested? A previously failing regression test that now passes. It uses the same code as static variables, so the same code path is also tested in other tests using static variables.

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
@@ -368,7 +368,7 @@ impl<'tcx> GotocCtx<'tcx> {
                 // here we have a function pointer
                 self.codegen_func_expr(instance, span).address_of()
             }
-            GlobalAlloc::Static(def_id) => self.codegen_static_pointer(def_id),
+            GlobalAlloc::Static(def_id) => self.codegen_static_pointer(def_id, false),
             GlobalAlloc::Memory(alloc) => {
                 // Full (mangled) crate name added so that allocations from different
                 // crates do not conflict. The name alone is insufficient because Rust
@@ -395,7 +395,7 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
-    pub fn codegen_static_pointer(&mut self, def_id: DefId) -> Expr {
+    pub fn codegen_static_pointer(&mut self, def_id: DefId, is_thread_local: bool) -> Expr {
         // here we have a potentially unevaluated static
         let instance = Instance::mono(self.tcx, def_id);
 
@@ -421,6 +421,7 @@ impl<'tcx> GotocCtx<'tcx> {
                 ctx.codegen_span(&span),
             )
             .with_is_extern(rlinkage.is_none())
+            .with_is_thread_local(is_thread_local)
         });
         sym.clone().to_expr().address_of()
     }
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
@@ -395,6 +395,7 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
+    /// Generates a pointer to a static or thread-local variable.
     pub fn codegen_static_pointer(&mut self, def_id: DefId, is_thread_local: bool) -> Expr {
         // here we have a potentially unevaluated static
         let instance = Instance::mono(self.tcx, def_id);
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
@@ -368,38 +368,10 @@ impl<'tcx> GotocCtx<'tcx> {
                 // here we have a function pointer
                 self.codegen_func_expr(instance, span).address_of()
             }
-            GlobalAlloc::Static(def_id) => {
-                // here we have a potentially unevaluated static
-                let instance = Instance::mono(self.tcx, def_id);
-
-                let sym = self.ensure(&self.symbol_name(instance), |ctx, name| {
-                    // check if this static is extern
-                    let rlinkage = ctx.tcx.codegen_fn_attrs(def_id).linkage;
-
-                    // we believe rlinkage being `Some` means the static not extern
-                    // based on compiler/rustc_codegen_cranelift/src/linkage.rs#L21
-                    // see https://github.com/model-checking/kani/issues/388
-                    //
-                    // Update: The assertion below may fail in similar environments.
-                    // We are disabling it until we find out the root cause, see
-                    // https://github.com/model-checking/kani/issues/400
-                    //
-                    // assert!(rlinkage.is_none());
-
-                    let span = ctx.tcx.def_span(def_id);
-                    Symbol::static_variable(
-                        name.to_string(),
-                        name.to_string(),
-                        ctx.codegen_ty(instance.ty(ctx.tcx, ty::ParamEnv::reveal_all())),
-                        ctx.codegen_span(&span),
-                    )
-                    .with_is_extern(rlinkage.is_none())
-                });
-                sym.clone().to_expr().address_of()
-            }
+            GlobalAlloc::Static(def_id) => self.codegen_static_pointer(def_id),
             GlobalAlloc::Memory(alloc) => {
                 // Full (mangled) crate name added so that allocations from different
-                // crates do not conflict. The name alone is insufficient becase Rust
+                // crates do not conflict. The name alone is insufficient because Rust
                 // allows different versions of the same crate to be used.
                 let name = format!("{}::{:?}", self.full_crate_name(), alloc_id);
                 self.codegen_allocation(alloc.inner(), |_| name.clone(), Some(name.clone()))
@@ -421,6 +393,36 @@ impl<'tcx> GotocCtx<'tcx> {
             assert!(res_t.is_pointer());
             offset_addr.cast_to(res_t)
         }
+    }
+
+    pub fn codegen_static_pointer(&mut self, def_id: DefId) -> Expr {
+        // here we have a potentially unevaluated static
+        let instance = Instance::mono(self.tcx, def_id);
+
+        let sym = self.ensure(&self.symbol_name(instance), |ctx, name| {
+            // check if this static is extern
+            let rlinkage = ctx.tcx.codegen_fn_attrs(def_id).linkage;
+
+            // we believe rlinkage being `Some` means the static not extern
+            // based on compiler/rustc_codegen_cranelift/src/linkage.rs#L21
+            // see https://github.com/model-checking/kani/issues/388
+            //
+            // Update: The assertion below may fail in similar environments.
+            // We are disabling it until we find out the root cause, see
+            // https://github.com/model-checking/kani/issues/400
+            //
+            // assert!(rlinkage.is_none());
+
+            let span = ctx.tcx.def_span(def_id);
+            Symbol::static_variable(
+                name.to_string(),
+                name.to_string(),
+                ctx.codegen_ty(instance.ty(ctx.tcx, ty::ParamEnv::reveal_all())),
+                ctx.codegen_span(&span),
+            )
+            .with_is_extern(rlinkage.is_none())
+        });
+        sym.clone().to_expr().address_of()
     }
 
     pub fn codegen_allocation_auto_imm_name<F: FnOnce(&mut GotocCtx<'tcx>) -> String>(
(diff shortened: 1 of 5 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -485,14 +485,9 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                let typ = self.codegen_ty(res_ty);
-                self.codegen_unimplemented(
-                    "Rvalue::ThreadLocalRef",
-                    typ,
-                    Location::none(),
-                    "https://github.com/model-checking/kani/issues/541",
-                )
+            Rvalue::ThreadLocalRef(def_id) => {
+                // Since Kani is single-threaded, we treat a thread local like a static variable:
+                self.codegen_static_pointer(*def_id)
```
danielsn, line 490: Should probably set `    pub is_thread_local: bool,` on the struct
https://github.com/model-checking/kani/blob/main/cprover_bindings/src/goto_program/symbol.rs#L42

tests/kani/ThreadLocalRef/main.rs
```diff
@@ -9,9 +9,8 @@ thread_local! {
 
 #[kani::proof]
 fn main() {
-    COND.with(|&b|{
+    COND.with(|&b| {
```
danielsn, line 12: How did the formatter miss this?

tests/kani/ThreadLocalRef/main.rs
```diff
@@ -9,9 +9,8 @@ thread_local! {
 
 #[kani::proof]
```
danielsn, line 10: Maybe also add a test where you do a series of operation on a thread local? And maybe one where the type is a struct not just a primitive?

tests/kani/ThreadLocalRef/main.rs
fzaiser: I think this test may have been committed before we had the format checker?

tests/kani/ThreadLocalRef/main.rs
fzaiser: Done. I added a test with `&'static str` which is a fat pointer, so compiles to a struct.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
fzaiser: Good catch. Done.

tests/kani/ThreadLocalRef/main.rs
celinval: The format checker runs in all files, not just the ones that has changed.

celinval (approved):
Looks good to me.

danielsn (approved):
Should probably address Celina's comments, but beyond that LGTM

tests/kani/ThreadLocalRef/main.rs
fzaiser: I don't think `cargo fmt` formats everything in all subfolders. Opened an issue: #1455

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -485,14 +485,9 @@ impl<'tcx> GotocCtx<'tcx> {
...
             }
-            Rvalue::ThreadLocalRef(_) => {
-                let typ = self.codegen_ty(res_ty);
-                self.codegen_unimplemented(
-                    "Rvalue::ThreadLocalRef",
-                    typ,
-                    Location::none(),
-                    "https://github.com/model-checking/kani/issues/541",
-                )
+            Rvalue::ThreadLocalRef(def_id) => {
```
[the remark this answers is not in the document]
fzaiser: What kind of warning do you have in mind? If you mean the single-threaded assumption, I don't think this is the place to warn about it. If anything, we should warn about that if the user tries to spawn threads. But at the moment it is not possible to spawn threads anyway. If we warn for thread locals, the user will get spurious warnings even if they only use one thread (tokio uses thread locals even in single-threaded mode for instance).

tests/kani/ThreadLocalRef/main.rs
```diff
@@ -0,0 +1,44 @@
...
+
+use std::cell::RefCell;
+
+thread_local! {
+    static COND : bool = true;
+    static COUNTER: RefCell<i32> = RefCell::new(0);
+    static COMPLEX_DATA: RefCell<&'static str> = RefCell::new("before");
+}
+
+pub fn main() {}
```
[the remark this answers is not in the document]
fzaiser: Thanks, removed

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -423,6 +395,37 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
+    pub fn codegen_static_pointer(&mut self, def_id: DefId, is_thread_local: bool) -> Expr {
```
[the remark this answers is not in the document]
fzaiser: Done

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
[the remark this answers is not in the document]
celinval: I do understand the concern of giving spurious warnings and that using thread-local in single threaded analysis is a legit use case.

However, my main concern is that we are making assumptions here that might not hold later. This assumption could be easily missed when we enable multi-thread support.

Note that this is already the behavior for all intrinsics that assume single threads. Here is the message we print:

https://github.com/model-checking/kani/blob/f719b565968568335d9be03ef27c5d05bb8fd0b7/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs#L14-L22

tests/kani/ThreadLocalRef/main.rs
celinval: Thanks. Fixed here https://github.com/model-checking/kani/pull/1459

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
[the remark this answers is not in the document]
fzaiser: Yeah, I've seen that message before (actually a whole wall of text made up of just this message) and found it very distracting. If it's not actually a soundness issue in Kani, I don't think we should print warnings. That's just going to lower the user's trust in the tool. But that's a separate issue, for this PR, I'll go with the `emit_concurrency_warning` as you suggested.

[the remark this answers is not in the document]
fzaiser: Issue filed here to discuss: #1460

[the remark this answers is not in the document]
fzaiser: Added a concurrency warning in this PR.

PR #1457, opened 2022-08-05
Cleanup TagEncoding::Direct. Instead of using a function to create a temporary, just directly create the rvalue we need

### LINKED ISSUE #1456: codegen_direct_literal should use statement expression

danielsn:
### Description of changes: 

Currently, we create a function, which creates a struct, assigns to one field of it, and returns. 
We then call that function, and use its return value as an rvalue.
But CBMC has rvalue struct expressions, which let us do the entire thing in one expression.

### Resolved issues: 

Resolves #1456 

### Testing:

* How is this change tested? Existing regression tests. I confirmed that this code is covered by these tests (see issue)

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
@@ -287,6 +287,18 @@ impl<'tcx> GotocCtx<'tcx> {
                                 let discr_ty = self.codegen_enum_discr_typ(ty);
                                 let init = self.codegen_scalar(s, discr_ty, span);
                                 let cgt = self.codegen_ty(ty);
+                                let fields =
+                                    cgt.get_non_empty_components(&self.symbol_table).unwrap();
+                                assert_eq!(
+                                    fields.len(),
+                                    1,
+                                    "TagEncoding::Direct encountered for enum with non-empty variants"
+                                );
+                                assert_eq!(
+                                    fields[0].name().to_string(),
+                                    "case",
+                                    "Unexpected field in enum/generator. Please report your failing case at https://github.com/model-checking/kani/issues/1465"
+                                );
                                 Expr::struct_expr_with_nondet_fields(
                                     cgt,
                                     btree_string_map![("case", init)],
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
@@ -3,6 +3,7 @@
 use crate::codegen_cprover_gotoc::utils::slice_fat_ptr;
 use crate::codegen_cprover_gotoc::GotocCtx;
 use crate::unwrap_or_return_codegen_unimplemented;
+use cbmc::btree_string_map;
 use cbmc::goto_program::{DatatypeComponent, Expr, Location, Stmt, Symbol, Type};
 use rustc_ast::ast::Mutability;
 use rustc_middle::mir::interpret::{
@@ -284,9 +285,13 @@ impl<'tcx> GotocCtx<'tcx> {
                             TagEncoding::Direct => {
                                 // then the scalar field stores the discriminant
                                 let discr_ty = self.codegen_enum_discr_typ(ty);
-
                                 let init = self.codegen_scalar(s, discr_ty, span);
-                                self.codegen_direct_literal(ty, init)
+                                let cgt = self.codegen_ty(ty);
+                                Expr::struct_expr_with_nondet_fields(
+                                    cgt,
+                                    btree_string_map![("case", init)],
+                                    &self.symbol_table,
+                                )
                             }
                         },
                     }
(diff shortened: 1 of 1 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -319,32 +324,6 @@ impl<'tcx> GotocCtx<'tcx> {
...
-        let func_name = format!("gen-{}:direct", self.ty_mangled_name(ty));
-        let pretty_name = format!("init_direct::<{}>", self.ty_pretty_name(ty));
-        let cgt = self.codegen_ty(ty);
-        self.ensure(&func_name, |tcx, _| {
-            let target_ty = init.typ().clone(); // N
-            let param = tcx.gen_function_parameter(1, &func_name, target_ty);
-            let var = tcx.gen_function_local_variable(2, &func_name, cgt.clone()).to_expr();
-            let body = vec![
-                Stmt::decl(var.clone(), None, Location::none()),
-                tcx.codegen_discriminant_field(var.clone(), ty)
```
fzaiser, line 332: `codegen_discriminant_field` has special handling for generators, which your code does not seem to cover. I don't think we should go back to hard-coding a "case" field everywhere.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -284,9 +285,13 @@ impl<'tcx> GotocCtx<'tcx> {
                             TagEncoding::Direct => {
                                 // then the scalar field stores the discriminant
                                 let discr_ty = self.codegen_enum_discr_typ(ty);
-
                                 let init = self.codegen_scalar(s, discr_ty, span);
-                                self.codegen_direct_literal(ty, init)
+                                let cgt = self.codegen_ty(ty);
+                                Expr::struct_expr_with_nondet_fields(
```
fzaiser, line 290: How does this interact with your zero_initializer PR? Is this function covered there?

fzaiser (changes requested):

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
fzaiser: Also, this code only runs for enums without additional data, just the discriminant, correct? Because only the discriminant is initialized. If that's the case, then using `srtuct_expr_with_nondet_fields` is misleading because there are no other fields to initialize with nondet.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
danielsn: Added a check for this

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
danielsn: Added a check for this

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -284,9 +285,25 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                                    fields.len(),
+                                    1,
+                                    "TagEncoding::Direct encountered for enum with non-empty variants"
+                                );
+                                assert_eq!(
+                                    fields[0].name().to_string(),
+                                    "case",
+                                    "Unexpected field in enum/generator. Please report your failing case at https://github.com/model-checking/kani/issues/1465"
+                                );
+                                Expr::struct_expr_with_nondet_fields(
```
fzaiser, line 302: I think this should just use Expr::struct_expr since there are no other fields to nondeterministically initialize.

fzaiser (approved):
Looks good to me, thanks! Please just fix that one little issue I mentioned.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -284,9 +285,25 @@ impl<'tcx> GotocCtx<'tcx> {
                             TagEncoding::Direct => {
                                 // then the scalar field stores the discriminant
                                 let discr_ty = self.codegen_enum_discr_typ(ty);
-
                                 let init = self.codegen_scalar(s, discr_ty, span);
-                                self.codegen_direct_literal(ty, init)
+                                let cgt = self.codegen_ty(ty);
+                                let fields =
+                                    cgt.get_non_empty_components(&self.symbol_table).unwrap();
+                                assert_eq!(
```
tedinski, line 292: Please add a comment explaining "why" this assert is here.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
danielsn: There are empty fields

fzaiser: But are these fields actually codegen'ed? I thought we skip fields of size 0.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
danielsn: added

tedinski (approved):

PR #1458, opened 2022-08-05
Cleanup rvalue_repeat: Use the array_constant expression rather than an explicit loop

### LINKED ISSUE #1451: Replace function call in codegen_rvalue_repeat with statement expr
The function call is unnecessary now that we have statement expr, and makes debugging harder.

danielsn:
### Description of changes: 

Currently, we generate a function to allow us to have a temporary variable in Expr context.
This has an explicit loop.
But there is an expression that does exactly what we want in a single step.

### Resolved issues:

Resolves #1451 

### Call-outs:

I had to change the check on `ArrayOf`. But CBMC seems to handle the non integer types just fine.

### Testing:

* How is this change tested? New regression test, plus confirmed that the existing regression has tests that cover this (see the issue)

### Checklist


fzaiser:
I like this change, assuming the CBMC behavior can be relied on. Two things:

* Can we confirm with the CBMC people if what we're doing is ok?
* Could you add more tests, with more complex and bigger types, just to make sure that's fine with CBMC, for example:
  - non-`Copy` types (then the length needs to be `<= 1`).
  - nested arrays

danielsn:
> I like this change, assuming the CBMC behavior can be relied on. Two things:
> 
> * Can we confirm with the CBMC people if what we're doing is ok?
> * Could you add more tests, with more complex and bigger types, just to make sure that's fine with CBMC, for example:
>   
>   * non-`Copy` types (then the length needs to be `<= 1`).
>   * nested arrays

All done :)

scripts/kani-regression.sh
```diff
@@ -38,7 +38,7 @@ cargo test -p kani-driver
 
 # Check output files (--gen-c option)
 echo "Check GotoC output file generation"
-time "$KANI_DIR"/tests/output-files/check-output.sh
+#time "$KANI_DIR"/tests/output-files/check-output.sh
```
adpaco-aws, line 41: Guess this was not intentional.

tests/kani/Array/array_init.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#[kani::proof]
+fn init_int() {
+    let a = [4u8; 6];
```
adpaco-aws, line 6: Can you add a test similar to this one?
```rust
fn main() {
    let mut a = [4u8; 6];
    a[2] = 1;
    assert_eq!(a[2], 1);
}
```

tests/kani/Array/array_init.rs
```diff
@@ -0,0 +1,67 @@
...
+    let a = [Some(4u8); 6];
+    let b = [a; 3];
+    let i: usize = kani::any();
+    kani::assume(i < 3);
+    let j: usize = kani::any();
+    kani::assume(j < 6);
+    assert_eq!(b[i][j], Some(4));
+}
+
+#[repr(C)]
```
fzaiser, line 31: Why #[repr(C)]? (also below)

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -203,49 +203,22 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
+    /// Codegens expressions of the type `let a  = [4u8; 6];`
     fn codegen_rvalue_repeat(
         &mut self,
         op: &Operand<'tcx>,
         sz: &ty::Const<'tcx>,
         res_ty: Ty<'tcx>,
+        _loc: Location,
```
fzaiser, line 212: Add the location to the returned expr?

fzaiser (approved):
Nice simplification!

PR #1422, opened 2022-07-29
Create Coding Conventions page

celinval:
### Description of changes: 

This commit creates a coding guidelines page. I used [rustc coding conventions](https://rustc-dev-guide.rust-lang.org/conventions.html) as a good model to follow.

### Resolved issues:

Resolves #1416

### Testing:

* How is this change tested? build docs script

* Is this a refactor change? No

### Checklist
- [ ] Each commit message has a non-empty body, explaining why the change was made
- [ ] Methods or procedures are documented
- [ ] Regression or unit tests are included, or existing tests cover the modified code
- [ ] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #1459, opened 2022-08-05
Add missing test folders to fmt script

### LINKED ISSUE #1455: Rust formatting check does not check all files
The Rust formatting check does not seem to check all Rust files in the repository, as discovered in https://github.com/model-checking/kani/pull/1452#discussion_r938130758.

celinval:
### Description of changes: 

The format script didn't include all suites that require formatting. I changed the logic to exclude specific folders via `rustfmt.toml` instead of keeping a list of folders to format.

### Resolved issues:

Resolves #1455

### Call-outs:

I couldn't find any easy way to make `rustfmt` ignore a folder that contains their own configuration file (like `s2n-quic`) submodule inside tests. We should probably not have submodules scattered around our repo, but that's a problem for another day. For now, I am forcing `rustfmt` to always use the root configuration file when checking tests.

### Testing:

### Checklist


fzaiser:
Wouldn't it make more sense to remove the test files from the ignore list here https://github.com/model-checking/kani/blob/main/rustfmt.toml#L12 instead of formatting them one by one in this script?

celinval:
> Wouldn't it make more sense to remove the test files from the ignore list here https://github.com/model-checking/kani/blob/main/rustfmt.toml#L12 instead of formatting them one by one in this script?

I believe we still have to format them one by one (AFAIK, you can't run `rustfmt` on a folder and `cargo fmt` will not format things out of our workspace). That said, you have a great point when it comes to the right mechanism to implement the ignore logic.

I had forgotten about that file. We haven't really used that. If you look closely, all paths there are related to the `rust` repo, not ours.

celinval:
Note that this was one of the things I was trying to fix here: https://github.com/model-checking/kani/pull/1422

tedinski (approved):

adpaco-aws (approved):

Issue #450, opened 2021-08-25
Rethink naming for external C libraries


avanhatt:
Right now, linking external C libraries depends on the Rust name pretty printer matching the C type name. This is fragile and definitely won't work if two different versions of a dependent crate use different versions of C structs. Rethink this.

PR #1461, opened 2022-08-05
Skip name mangling only for #[repr(C)] types without generic parameters

### LINKED ISSUE #1438: Type mismatch error when compiling tokio
I tried this code: https://github.com/tokio-rs/tokio at commit b2ada60e701d5c9e6644cf8fc42a100774f8e23f

```
git clone https://github.com/tokio-rs/tokio.git
cd tokio
git checkout b2ada60e701d5c9e6644cf8fc42a100774f8e23f
cargo kani
```

with Kani version: 16f10e9a0f40e4e50a1869b09361c280ae840d21

I expected to see this happen: Kani compiles the crate successfully

Instead, this happened:
```
thread 'rustc' panicked at 'Unexpected type mismatch in projection:
Expr { value: Member { lhs: Expr { value: Dereference(Expr { value: Symbol { identifier: "_RINvNtCs4YygHDVoDdI_4core3ptr13drop_in_placeINtNtNtNtCs43oGvCTZBYr_5tokio7runtime4task4core4CellINtNtNtBN_8blocking4task12BlockingTaskINtNtCsbutnkbugfxH_5alloc5boxed3BoxNCNvXs_NtNtBP_2io8blockingINtB2O_8BlockingNtNtNtCsfcET2xiHyR1_3std2io5stdio6StdoutENtNtB2Q_11async_write10AsyncWrite10poll_flush0EENtNtB1B_8schedule12NoopScheduleEEBP_::1::var_1" }, typ: Pointer { typ: StructTag("tag-runtime::task::core::Cell<runtime::blocking::task::BlockingTask<std::boxed::Box<[closure@tokio/src/io/blocking.rs:152:52: 152:59]>>, runtime::blocking::schedule::NoopSchedule>") }, location: None }), typ: StructTag("tag-runtime::task::core::Cell<runtime::blocking::task::BlockingTask<std::boxed::Box<[closure@tokio/src/io/blocking.rs:152:52: 152:59]>>, runtime::blocking::schedule::NoopSchedule>"), location: None }, field: "core" }, typ: StructTag("tag-_16865545542732683769"), location: None }
Expr type
StructTag("tag-_16865545542732683769")
Type from MIR
StructTag("tag-_11444338364738518381")', kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs:184:13
```

---

Strangely, the pretty names of the two struct tags are the same:

```
runtime::task::core::Core<runtime::blocking::task::BlockingTask<std::boxed::Box<[closure@tokio/src/io/blocking.rs:152:52: 152:59]>>, runtime::blocking::schedule::NoopSchedule>
```

fzaiser:
### Description of changes: 

Currently, Kani does not mangle all `#[repr(C)]` types to make accessing them using C FFI easier. Instead, it uses the pretty name of the type in the symbol table. This caused problems in #1438 because the pretty name does not include generic parameters of closures, leading to two different types accidentally being given the same name in the symbol table. We fix this by mangling a type with generic parameters, even if it is `#[repr(C)]`.

### Resolved issues:

Resolves #1438 

### Call-outs:

This solution is still not ideal. As the comment in the code mentions: skipping name mangling is likely insufficient if a dependent crate has two versions of linked C libraries (see #450).

I also added a function to debug print a GotoC type recursively. This will make debugging type mismatches easier in the future.

### Testing:

* How is this change tested? Adde a regression test

### Checklist


kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -127,6 +127,129 @@ impl ExprExt for Expr {
 
 /// Function signatures
 impl<'tcx> GotocCtx<'tcx> {
+    /// This method prints the details of a GotoC type, for debugging purposes.
+    #[allow(unused)]
+    pub(crate) fn debug_print_type_recursively(&self, ty: &Type) -> String {
```
danielsn, line 132: Not that it matters for this particular use case, but we should really just have a `Type` visitor

danielsn (approved):

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
fzaiser: Agreed.

PR #1466, opened 2022-08-05
Fix all remaining clippy lints

### LINKED ISSUE #1373: Clippy Compliance within Kani

fzaiser:
### Description of changes: 

This fixes all the remaining clippy lints.

### Resolved issues:

Resolves #1373 

### Call-outs:

This PR adds local `#[allow(clippy::___)]` annotations in a lot of places if I think a lint is generally good to have, but does not apply in that particular instance.

This PR also globally ignores `clippy::new_without_default` because in my opinion, just because a type has a `Self::new` function, it does not always warrant a `Default` implementation.

Agreeing on this will probably involve some discussion. For reference, descriptions of each clippy lints can be found here: https://rust-lang.github.io/rust-clippy/master/

### Testing:

* How is this change tested? Clippy is run in CI.

### Checklist


kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -37,8 +37,8 @@ pub struct GotocCodegenBackend {
 }
 
 impl GotocCodegenBackend {
-    pub fn new(queries: &Rc<QueryDb>) -> Box<dyn CodegenBackend> {
-        Box::new(GotocCodegenBackend { queries: Rc::clone(queries) })
+    pub fn new(queries: &Rc<QueryDb>) -> Self {
```
danielsn, line 40: Nice

danielsn (approved):

fzaiser:
I'll keep this open for a bit, so other people can comment too, e.g. if they think some lints should not be locally or globally disabled or dealt with in a different way.

tedinski (approved):

Issue #1363, opened 2022-07-12
Support Primtive Proptests


YoshikiTakashima:
Requested feature: Support primitive proptests.

Use case: Automatically converting proptests to Kani harnesses allow
us to serve large number of customers like `tokio` that have proptests
implemented but no Kani harnesses.

Test case:
https://github.com/YoshikiTakashima/kani/tree/yoshi-proptest-primitive/tests/proptest/primitive

Feature Branch on Fork: https://github.com/YoshikiTakashima/kani/tree/yoshi-proptest-primitive/

YoshikiTakashima:
### Replicating linker version.
First, replicate the kani version.
```
git clone https://github.com/YoshikiTakashima/kani.git
cd kani
git checkout yoshi-proptest-primitive
git checkout eac76a7e -- library/proptest
cargo clean; cargo build --workspace
```

Then run it on the case that replicates this.
```
git clone https://github.com/YoshikiTakashima/simplified-prost-rs-timestamp.git
cd simplified-prost-rs-timestamp
cargo kani
```

YoshikiTakashima:
Info from deep dive: The following is required for kani to work.
- `*.rlib`
- `symtab.json`
Currently, we only provide the `rlib`

Possible 
`--enable-unstable --c-lib <path to gotobinary>` where `gotobinary` is `*.goto`

YoshikiTakashima:
Code changes implemented to link proptest. After cleanup, we can merge.

YoshikiTakashima:
See the following comment for way to integrate proptest w/o significant changes. https://github.com/model-checking/kani/pull/1435#issuecomment-1215117585

PR #1435, opened 2022-08-02
Support for Primitive Proptests

YoshikiTakashima:
### Description of changes: 

Adds support for running proptests with primitive inputs using kani.

### Resolved issues:

Resolves #1363 

### Call-outs:

- Feature only runs with `cargo kani`.
- Since we are only merging a small  portion of proptest, we cannot support proptests like `tests/cargo-kani/rectangle-example`. This means we disable the proptest import in that test case.
### Testing:

* How is this change tested? `tests/cargo-kani/proptest`

* Is this a refactor change? No.

### Checklist
- [x] Each commit message has a non-empty body, explaining why the change was made
- [x] Methods or procedures are documented
- [x] Regression or unit tests are included, or existing tests cover the modified code
- [x] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #1467, opened 2022-08-05
[Blocked on #1435] Support proptest arbitrary generation for bool and tuple

YoshikiTakashima:
### Description of changes: 

Added support for arbitrary generation trait, and implemented it for booleans and tuples.

### Resolved issues: 

Resolves: #1363, PR 2/3

### Call-outs:

- This requires #1435 to merge first. This is PR 2/3 to implement support for primitive types in proptest.
- As of now, the diff looks large. However, once the above PR merges, the size of the PR should be around 10 files.
- Since only primitives and tuples of them are supported, the test cases is only 1 file. Tthis test case is designed to test both tuples and booleans at once.

### Testing:

* How is this change tested? tests/cargo-kani/proptest/src/arbitrary_boolean.rs

### Checklist


kani-driver/src/call_cargo.rs
```diff
@@ -80,10 +82,18 @@ impl KaniSession {
             });
         }
 
+        let mut symtabs = glob(&outdir.join("*.symtab.json"))?;
+        let mut metadata = glob(&outdir.join("*.kani-metadata.json"))?;
+        if std::env::var("IS_KANI_PROPTEST").is_ok() {
```
celinval, line 87: Shouldn't this be a runtime option?

celinval (changes requested):
Copying my comment from https://github.com/model-checking/kani/pull/1435

> I don't think we should add the proptest code in our repository. We should probably do one of the following:
>
>  1. Refactor the changes so we can extend proptest instead of forking from it.
>  2. If we do need to fork proptest, we should create a separate repository with the fork. We should then add a submodule here to this fork.
>
> Feel free to ping me offline if you want to discuss this further or if you have any suggestions.

YoshikiTakashima:
Closing pull request in favor of merging to `features/proptest`.

Issue #1072, opened 2022-04-20
Kani implementation of drop for Struct<dyn T> is broken


celinval:
We were handling `drop_in_place` incorrectly when the object to be dropped had type Struct<dyn T> like in the example below.

For now, we are going to codegen unimplemented, so if this code is reachable, the verification will fail.

I tried this code:
```rust
use std::rc::Rc;

pub trait DummyTrait {}

pub struct Wrapper<T: ?Sized> {
    pub w_id: u128,
    pub inner: T,
}

impl<T: ?Sized> Drop for Wrapper<T> {
    fn drop(&mut self) {
        assert_eq!(self.w_id, 0);
    }
}

struct DummyImpl {
    pub id: u128,
}

impl DummyTrait for DummyImpl {}

impl Drop for DummyImpl {
    fn drop(&mut self) {
        assert_eq!(self.id, 1);
    }
}

#[kani::proof]
fn check_drop_dyn() {
    let original = Rc::new(Wrapper { w_id: 0, inner: DummyImpl { id: 1 } });
    let _wrapper =
        unsafe { Rc::from_raw(Rc::into_raw(original) as *const Wrapper<dyn DummyTrait>) };
}
```

using the following command line invocation:

```
kani check.rs
```

with Kani version:

I expected to see this happen: Harness should succeed.

Instead, this happened:
 - With codegen unimplemented check, this harness fails with:
 ```
Failed Checks: drop unsized struct: Wrapper<dyn DummyTrait>
```
 
- Without codegen unimplemented check, this harness fails with:
```
Failed Checks: assertion failed: self.id == 1
```

tedinski:
I think the fix for this MIGHT be to refactor the "virtual table drop" code to actually call the normal virtual table invocation code, instead of sort of duplicating it, apparently incorrectly. I'm not sure if that's the only issue here however.

Also, this refactoring is not quite as straightforward as it sounds for uninteresting reasons: drop returns unit/void, but our existing virtual table function invocation code expects a `Place` to store the return value. So THAT in turn needs refactoring to separate out those two steps, which in turn suggests some refactorings to `codegen_function`...

Issue #426, opened 2021-08-16
Fix drop type checking for nested `Arc`/`Mutex` objects


avanhatt:
The two `SizeAndAlignOfDst` tests fail with:

```rust
"func":"Expr"{
   "value":"Symbol"{
      "identifier":"_RINvNtCsgWci0eQkB8o_4core3ptr13drop_in_placeNtNtNtCs1HAdiQHUxxm_3std10sys_common5mutex12MovableMutexECsb7rQPrKk64Y_4main"
   },
   "typ":"Code"{
      "parameters":[
         "Parameter"{
            "typ":"Pointer"{
               "typ":"StructTag(""tag-std::sys_common::mutex::MovableMutex"")"
            },
            "identifier":"None",
            "base_name":"None"
         }
      ],
      "return_type":"StructTag(""tag-Unit"")"
   },
   "location":"None"
}"args":[
   "Expr"{
      "value":"Symbol"{
         "identifier":"_RINvNtCsgWci0eQkB8o_4core3ptr13drop_in_placeINtNtNtCs1HAdiQHUxxm_3std4sync5mutex5MutexDNtCsb7rQPrKk64Y_4main10SubscriberEL_EEB1p_::1::var_1"
      },
      "typ":"StructTag(""tag-dyn Subscriber"")",
      "location":"None"
   }
]"', compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs:532:9"

stack backtrace:
   0: rust_begin_unwind
             at ./library/std/src/panicking.rs:517:5
   1: std::panicking::begin_panic_fmt
             at ./library/std/src/panicking.rs:460:5
   2: rustc_codegen_llvm::gotoc::cbmc::goto_program::expr::Expr::call
             at ./compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs:532:9
   3: rustc_codegen_llvm::gotoc::statement::<impl rustc_codegen_llvm::gotoc::metadata::GotocCtx>::codegen_drop
             at ./compiler/rustc_codegen_llvm/src/gotoc/statement.rs:148:29
   4: rustc_codegen_llvm::gotoc::statement::<impl rustc_codegen_llvm::gotoc::metadata::GotocCtx>::codegen_terminator
             at ./compiler/rustc_codegen_llvm/src/gotoc/statement.rs:58:66
[...]

PR #1468, opened 2022-08-05
Simplify drop logic

### LINKED ISSUE #1434: Support drop_in_place (needed for tokio)
If you clone the [tokio repository](https://github.com/tokio-rs) and run `cargo kani`, you get a lot of warnings about unsupported features, one of them is `drop_in_place`, example:

```
             - Rvalue::ThreadLocalRef (7)
             - drop_in_place for drop_unimplemented__RINvNtCs4YygHDVoDdI_4core3ptr13drop_in_placeNtNtNtCsfcET2xiHyR1_3std2io5error5ErrorECs9FYAfFSAVYQ_3mio (1)
             - try (1)
```

The lack of ThreadLocalRef and try [can be worked around by disabling certain functions](https://github.com/model-checking/kani/commit/7704538bd5d38bce36eada763d3b8fcb53711e00) in `functions.rs`:
```
            name if name.contains("metadata") => true,
            name if name.contains("std::sync::atomic::AtomicPtr") => true,
            name if name.contains("runtime::task::harness::Harness") => true,
            name if name.contains("runtime::task::core::Cell") => true,
```
But even if we do this, we get an error codegen'ing `std::ptr::drop_in_place::<runtime::task::core::Cell<runtime::blocking::task::BlockingTask<std::boxed::Box<[closure@tokio/src/io/blocking.rs:152:52: 152:59]>>, runtime::blocking::schedule::NoopSchedule>>`. The assertion failure happens in https://github.com/model-checking/kani/blob/16f10e9a0f40e4e50a1869b09361c280ae840d21/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs#L178.

It is important to fix this bug to support tokio

fzaiser:
### Description of changes: 

This removes a check from the `drop_in_place` code that has become unnecessary.

### Resolved issues:

Resolves #1434 

### Testing:

* How is this change tested? Regression tests and I manually tried compiling tokio with it.

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
@@ -851,7 +851,7 @@ impl<'tcx> GotocCtx<'tcx> {
                 // Function body
                 let unimplemented = ctx
                     .codegen_unimplemented(
-                        format!("drop_in_place for {}", drop_sym_name).as_str(),
+                        &format!("drop_in_place for missing type {}", ty),
                         Type::empty(),
                         Location::none(),
                         "https://github.com/model-checking/kani/issues/281",
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
@@ -372,24 +372,7 @@ impl<'tcx> GotocCtx<'tcx> {
                         // The only argument should be a self reference
                         let args = vec![arg];
 
-                        // We have a known issue where nested Arc and Mutex objects result in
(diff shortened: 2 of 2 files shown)


celinval:
Does this mean that https://github.com/model-checking/kani/issues/426 is no longer an issue?

fzaiser:
Huh, this is interesting. The std lib test passes on my M1 Mac, but fails on Ubuntu in CI... Will investigate next week.

fzaiser:
So this is actually the Firecracker test, not std lib. The firecracker test is only run on Linux, which explains why it did not fail on my computer. If I run the firecracker test without the patch, I get: `Unsupported drop unsized struct: std::cell::UnsafeCell<dyn event_manager::Subscriber>` (which is tracked in #1072). Thus I'll close this pull request.

PR #1469, opened 2022-08-06
Add Poison to the Kani-GotoC AST

giltho:
### Description of changes: 

Right now, because of CBMC, Kani produces a `x = nondet()` for a `Deinit(x)`.
Progressing on #920, I'm making this small PR to mark these statements using a comment.

### Call-outs:

I don't think this is the optimal way of doing it, so I'm happy with any suggestion.
Right now, I added a field in one of the Location:: kinds, the one that corresponds to my use case, and then I set it when required, but the way it's done isn't very systematic

### Testing:

I'm not sure how I should be testing this inside Kani.
I've parsed these with Kanillian and successfully compiled them to the right poison statement instead.

### Checklist


danielsn:
Why not create a poison expression, that translates to nondet plus comment when you lower to irep?

giltho:
I'm ok with that, but it means actively taking the decision to get stmt.rs and cbmc out of sync and so I went with a softer solution first
But happy to update!

giltho:
Replacing the Nondet expression with Deinit does raise one question: what happens with the [NondetTransformer](https://github.com/model-checking/kani/blob/main/cprover_bindings/src/goto_program/symtab_transformer/gen_c_transformer/nondet_transformer.rs#L56-L66) ?

1) I don't understand if and when it's called. If it was, I would expect no Nondet expressions to ever be sent to me by Kani, but it's not the case
2) If I add a poison value but want to maintain the invariant that this tranformer currently has, I need to also change every call to poison to a call to some poison function that would have the same behaviour. I will implement that function, but do note that it probably means creating 2 nondet functions for each type that both needs Poison and Nondet

cprover_bindings/src/irep/to_irep.rs
```diff
@@ -253,6 +253,11 @@ impl ToIrep for ExprValue {
                 ],
             },
             ExprValue::Nondet => side_effect_irep(IrepId::Nondet, vec![]),
+            ExprValue::Poison => {
+                // For  now, Poison is lowered to Nondet with a comment.
+                // In the future, CBMC might handle poison expressions directly
+                side_effect_irep(IrepId::Nondet, vec![]).with_comment("deinit")
```
danielsn, line 259: Is "deinit" always the comment we want? What about cases where Kani might introduce poison itself? Why not just use "poison" as the comment?

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -111,6 +111,9 @@ pub enum ExprValue {
     },
     /// `__nondet()`
     Nondet,
+    /// Poison comes from the Deinit statement of Rust
```
danielsn, line 114: That's not the only reason we might want to use Poison

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -111,6 +111,9 @@ pub enum ExprValue {
...
+    /// CBMC doesn't model it (as of now)
```
danielsn, line 115: Instead of just explaining what CBMC doesn't do, explain what CBMC will do.

cprover_bindings/src/irep/irep.rs
```diff
@@ -44,6 +44,11 @@ impl Irep {
         }
     }
 
+    pub fn with_comment<T: Into<InternedString>>(self, c: T) -> Self {
+        // Using Irep::
```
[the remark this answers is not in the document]
giltho: Oh I don't know how that got in, removing that

cprover_bindings/src/irep/to_irep.rs
giltho: Yes actually, I moved that code from the previous version of the pr and should have changed the name

giltho:
Ok fixed the various comments from @danielsn 

That being said, after rereading my PR with a fresh post-weekend mind, I'm a bit unhappy with a Poison expression.

## The argument against Poison as an expression

My opinion is that Poison should not be an expression, because it is not a valid value. Values are mathematical that result from evaluating an expression. A [good definition](https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#value) is given by the UCG WG.

Poison is precisely not a value because it cannot result from an expression evaluation, and it cannot be read from memory. It can only exist in memory.
That's why MIR has a deinit statement and not a uninit value that is assigned to a place.

Another good description is given by [Robbert Krebbers in the description of the CH2O memory model](https://arxiv.org/pdf/1509.03339.pdf). Poison bits are part of a value representation in memory, but not part of the value itself.

## The argument in favor of Poison as an expression

CBMC has this surprising behaviour of 1) explicitely having paddings as part of the type definitions and 2) explicitely writing nondet in the padding as part of Struct expressions. This conflicts with what I said above for C, and in that sense, CBMC is not C. I'm not sure of the theoretical and practical ramifications of being allowed to produce poison expressions, since the whole point of poison is that it cannot be read in the first place, but it seems that, to work properly in CBMC, you'd need a poison expression. That probably means that you need to add, every read, a check that the produce value isn't poison or something similar

giltho:
Note that this behaviour of expliciting the value of padding might create discrepancies between C and GotoC.
Take the following example

```c
struct S {
   u8 x;
   [24 bit padding]
   u32 y;
}
```

```c
struct S s = { 0; Poison; 4 }
u8* y = ((u8*) s) + 1;
*y = 2;
y++;
*y = 2;
y++
*y = 3;
```
Variable `s` now contains the equivalent of the u8 array `[0, 1, 2, 3, 4]`.

However, if you then write
```c
struct S t = s
```

then actually, you shouldn't be allowed to read the padding bits and you'll only be writing poison in there. That means, `t` should be the equivalent of the "u8" array `[0, P, P, P, 4]`

The natural semantics for a language that explicits the value of padding would be to simply copy the value of `s` in `t` and therefore not have the poison in the right place.

cprover_bindings/src/goto_program/symtab_transformer/gen_c_transformer/nondet_transformer.rs
```diff
@@ -9,21 +9,30 @@ use std::collections::HashMap;
...
+        .transform_symbol_table(original_symbol_table)
     }
 
     /// Extract `nondet_types` map for final processing.
     pub fn nondet_types_owned(&mut self) -> HashMap<String, Type> {
         std::mem::take(&mut self.nondet_types)
     }
+
+    pub fn poison_types_owned(&mut self) -> HashMap<String, Type> {
+        std::mem::take(&mut self.nondet_types)
```
tedinski, line 34: copy/paste mistake?

giltho: Yes!

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -111,6 +111,10 @@ pub enum ExprValue {
     },
     /// `__nondet()`
     Nondet,
+    /// Poison corresponds to a region of memory that
```
celinval, line 114: Poison is more than uninitialized, right? Or are we narrowing down the concept for now?

I do wonder if we should state here that CBMC doesn't currently support poison and this gets translated to nondet.

It might be helpful to add a link to https://github.com/diffblue/cbmc/issues/7014.

celinval:
> Ok fixed the various comments from @danielsn
> 
> That being said, after rereading my PR with a fresh post-weekend mind, I'm a bit unhappy with a Poison expression.
> ## The argument against Poison as an expression
> 
> My opinion is that Poison should not be an expression, because it is not a valid value. Values are mathematical entities that result from evaluating an expression. A [good definition](https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#value) is given by the UCG WG.
> 
> Poison is precisely not a value because it cannot result from an expression evaluation, and it cannot be read from memory. It can only exist in memory. That's why MIR has a deinit statement and not an uninit value that is assigned to a place.
> 
> Another good description is given by [Robbert Krebbers in the description of the CH2O memory model](https://arxiv.org/pdf/1509.03339.pdf). Poison bits are part of a value representation in memory, but not part of the value itself.
> ## The argument in favor of Poison as an expression
> 
> CBMC has this surprising behaviour of 1) explicitely having paddings as part of the type definitions and 2) explicitely writing nondet in the padding as part of Struct expressions. This conflicts with what I said above for C, and in that sense, CBMC is not C. I'm not sure of the theoretical and practical ramifications of being allowed to produce poison expressions, since the whole point of poison is that it cannot be read in the first place, but it seems that, to work properly in CBMC, you'd need a poison expression. That probably means that you need to add, every read, a check that the produce value isn't poison or something similar

I totally agree with your point of view. What if you change poison to be a statement instead of an expression? `Poison(place)` would take an address as its only argument.

This matches the DeInit API as well as the initial API suggested here: https://github.com/diffblue/cbmc/issues/7014.

giltho:
Updated to a deinit statement.
This is not ready to be merged for two reasons:
- First I need to sync my Kanillian branch with it to make sure it works as expected
- I need to understand when the NondetTransformer is used and why. Right now, even when the NondetTransformer is used, the final code will contain Nondet expressions because the lowering is done at the very last moment.

cprover_bindings/src/goto_program/symtab_transformer/transformer.rs
```diff
@@ -592,6 +592,11 @@ pub trait Transformer: Sized {
         Stmt::decl(transformed_lhs, transformed_value, Location::none())
     }
 
+    fn transform_stmt_deinit(&mut self, place: &Expr) -> Stmt {
+        let transformed_place = self.transform_expr(place);
+        Stmt::deinit(transformed_place, Location::none())
```
danielsn, line 603: Is it supposed to be location none?

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -1401,6 +1401,10 @@ impl Expr {
         Stmt::assign(self, rhs, loc)
     }
 
+    pub fn deinit(self, loc: Location) -> Stmt {
+        Stmt::deinit(self, loc)
```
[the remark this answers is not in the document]
giltho: Yes, I'll add the condition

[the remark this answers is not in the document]
giltho: Yes, but I don't know how to check for that.
Assignment checks for type equality but that's it, even though the left-hand side has to be an lvalue
image

cprover_bindings/src/goto_program/symtab_transformer/transformer.rs
giltho: Yes, because the location is modified added-later on. Every `transform_stmt_*` method returns a none location, and the location is propagated by the parent function `transform_stmt`

giltho:
I have one broken test in the regression test suite, I'll try to figure out what's wrong, but I'm not sure how to approach that

cprover_bindings/src/goto_program/expr.rs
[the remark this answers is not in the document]
tedinski: It might be nice to add an "is l-value" check to cprover bindings, but obviously not in this pr...

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -1493,6 +1493,10 @@ impl Expr {
         Stmt::assign(self, rhs, loc)
     }
 
+    pub fn deinit(self, loc: Location) -> Stmt {
```
tedinski, line 1496: We definitely want a comment here, especially as "deinit" is not a c notion really, and this is "cprover_bindings"

That's kind of why I thought this should be called "poison" but I don't remember if that was decided against for some reason?

Also, I don't know if this is me reading it weird, but I have the comment note that this is a shorthand/helper (perhaps by example? Like: "Construct a deinit statement from an expression by writing `expr.deinit()`"

cprover_bindings/src/goto_program/stmt.rs
```diff
@@ -256,6 +258,10 @@ impl Stmt {
         stmt!(Decl { lhs, value }, loc)
     }
 
+    pub fn deinit(place: Expr, loc: Location) -> Self {
```
tedinski, line 261: doc comment here

cprover_bindings/src/irep/irep.rs
```diff
@@ -44,6 +44,10 @@ impl Irep {
         }
     }
 
+    pub fn with_comment<T: Into<InternedString>>(self, c: T) -> Self {
```
tedinski, line 47: cool. doc comment? I know I'm the one who vaguely waved you in the direction of "putting a comment on an irep" but did you dig into exactly how this works? It'd be good to explain I think...

cprover_bindings/src/irep/to_irep.rs
```diff
@@ -430,6 +430,10 @@ impl ToIrep for StmtBody {
                     code_irep(IrepId::Decl, vec![lhs.to_irep(mm)])
                 }
             }
+            StmtBody::Deinit(place) => {
+                code_irep(IrepId::Assign, vec![place.to_irep(mm), place.typ().nondet().to_irep(mm)])
```
tedinski, line 434: In this spot I would leave a comment explaining "why" this is the right translation.

```
// CBMC doesn't yet have a notion of poison (link?)
// So we translate identically to `nondet` here, but add a comment noting we wish it were poison
// potentially for other backends to pick up and treat specially.
```

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -59,7 +59,9 @@ impl<'tcx> GotocCtx<'tcx> {
             }
             StatementKind::Deinit(place) => {
                 // From rustc doc: "This writes `uninit` bytes to the entire place."
-                // Thus, we assign nondet() value to the entire place.
+                // Our model of GotoC has a similar statement, which is later lowered
+                // to assigning a Nondet in CBMC, with a comment specifying that it
+                // corresponds to a Deinit.
```
tedinski, line 64: good!

tedinski:
> I need to understand when the NondetTransformer is used and why. Right now, even when the NondetTransformer is used, the final code will contain Nondet expressions because the lowering is done at the very last moment. Is it ok? Is it not?

NondetTransformer isn't actively used, and was part of an earlier effort to make `--gen-c` emit more runnable/parsable C code. If it just rewrites `deinit` away, I think it'd be fine (just stick a comment in there). We might remove it later...

cprover_bindings/src/goto_program/expr.rs
giltho: I think it's because @celinval (correct me if I'm wrong) pointed out that "poison" in this form is only intended to model uninitialized memory. So calling it poison is too vague and deinit makes more sense. I tend to agree with that

cprover_bindings/src/irep/irep.rs
giltho: That one is a bit blurry actually. It will add a comment directly to the irep.
As of now, comments are usually in the "location" sub, and here I'm making the decision to put it outside.
Not sure that's the best option, but the location is handled by Expr::to_irep not ExprValue::to_irep, so I can't propagate it up without a big refactoring that would clutter the entire thing for one use case.

I do think we need a more standard way of writing comments like that for alternative back-ends

cprover_bindings/src/irep/to_irep.rs
giltho: Yes

cprover_bindings/src/goto_program/stmt.rs
giltho: done!

cprover_bindings/src/goto_program/stmt.rs
```diff
@@ -62,6 +62,8 @@ pub enum StmtBody {
         lhs: Expr, // SymbolExpr
         value: Option<Expr>,
     },
+    /// Marks the target place as uninitialized.
```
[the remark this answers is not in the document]
giltho: done!

cprover_bindings/src/goto_program/expr.rs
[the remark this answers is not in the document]
danielsn: https://github.com/model-checking/kani/blob/main/cprover_bindings/src/goto_program/expr.rs#L385 
It's called `can_take_address_of()`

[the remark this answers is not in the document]
danielsn: Annoyingly, enabling it on assign causes the stdlib test to crash.  Investigating.

tedinski (approved):
Ok, LGTM, let's merge.

1. It looks like Daniel is looking into adding an l-value assertion, but I think that addition can go in that PR when a possible problem there is resolved?
2. Let's open a quick follow-up issue about how we're attaching comments, to go double check next week that our approach seems sensible.

PR #1471, opened 2022-08-08
Use #[kani::proof] for both sync and async functions (instead of #[kani::async_proof] for async)

### LINKED ISSUE #1464: #[kani::async_proof] should be merged into #[kani:proof]
In https://github.com/model-checking/kani/pull/1430#issuecomment-1204585500, @camshaft suggested that instead of having two different attributes for `async fn` and `fn`, we should just have one attribute `#[kani::proof]`, which would do the right thing depending on whether the function is `async` or not.

Pros: more uniform API

Cons: kani::proof does not take arguments, but kani::async_proof probably will in the future (to select an executor)

I'm leaning towards a separate attribute. What do people think? @model-checking/kani-devs

fzaiser:
### Description of changes: 

Previously, we had `#[kani::proof]` for synchronous and `#[kani::async_proof]` for asynchroonus harnesses. This PR unifies the two attributes, allowing `#[kani::proof]` to be used for both kinds of functions.

### Resolved issues:

Resolves #1464 

### Call-outs:

This now uses the syn and quote crates also for normal `#[kani::proof]` annotations for consistency. Previously this attributes modified `TokenStream`s manually.

### Testing:

### Checklist

### DIFF
--- a/library/kani_macros/src/lib.rs
+++ b/library/kani_macros/src/lib.rs
@@ -19,7 +19,7 @@ use {
 #[cfg(all(not(kani), not(test)))]
 #[proc_macro_attribute]
 pub fn proof(_attr: TokenStream, _item: TokenStream) -> TokenStream {
-    // Not-Kani, Not-Test means this code shouldn't exist, return nothing.
+    // Not-Kani, not-test means this code shouldn't exist, return nothing.
     TokenStream::new()
 }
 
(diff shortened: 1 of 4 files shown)


library/kani_macros/src/lib.rs
```diff
@@ -82,30 +59,44 @@ pub fn async_proof(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    } else {
+        assert!(
+            sig.inputs.is_empty(),
+            "#[kani::proof] cannot be applied to async functions that take inputs for now"
+        );
+        let mut modified_sig = sig.clone();
+        modified_sig.asyncness = None;
+        let fn_name = &sig.ident;
+        quote!(
+            #[kani::proof]
```
camshaft, line 91: Why not invoke `kanitool::proof` directly? You've already invoked the proc macro and putting it here will cause another level of macro expansion to occur.

library/kani_macros/src/lib.rs
```diff
@@ -39,37 +39,14 @@ pub fn proof(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
-pub fn async_proof(attr: TokenStream, item: TokenStream) -> TokenStream {
-    proof(attr, item)
-}
-
-#[cfg(kani)]
-#[proc_macro_attribute]
-/// Translates #[kani::async_proof] to a #[kani::proof] harness that calls `kani::block_on`, if Kani is active
+/// Handles #[kani::proof] if #[cfg(kani)] is active
+///
+/// For synchronous functions, it becomes a Kani-internal attribute.
```
camshaft, line 44: Usually proc macros don't describe their implementation details. It's better to spend the time to talk about how to use it and what it's for.

library/kani_macros/src/lib.rs
fzaiser: Because if we change the translation of #[kani::proof] for sync, we may forget to update async, if the code is in different places. This approach seems cleaner to me because it separates two steps: getting rid of async, and then translating to Kani internals. What's the disadvantage of doing it this way? Will performance suffer a lot from this?

library/kani_macros/src/lib.rs
fzaiser: Good point. I will update the comment.

library/kani_macros/src/lib.rs
camshaft: Well it doesn't have to be defined in multiple places. You can call a subfunction inside of the proc macro.

But, yes, it makes compilation slower since it'll have to invoke this macro multiple times. This includes serializing and deserializing the token tree multiple times.

fzaiser: I don't think this matters compared to how long CBMC takes to prove a harness, but I've changed the code to avoid it.

library/kani_macros/src/lib.rs
fzaiser: I've updated the doc comment accordingly. If there's anything else you think it should include from a user's perspective, please let me know.

danielsn (approved):

camshaft (approved):

PR #1479, opened 2022-08-09
Disable calling `--slice-formula` by default with conditional flags.

### LINKED ISSUE #1288: slice-formula is enabled unconditionally, but should not be used when getting traces
CBMC's `--slice-formula` can result in bizarre traces with weird values being assigned to variables CBMC no longer considers important.

We should not enable it when we're generating traces.

We should also have a flag that allows disabling it (`--no-slice-formula`?)

jaisnan:
### Description of changes: 

Hot fix for disabling CBMC's `--slice-formula`. `--slice-formula` was being called by default which led to some unnecessary values being presented in the trace. This PR disables the CBMC flag when calling `--visualize` or when we run Kani with `--no-slice-formula`.

### Resolved issues:

Resolves #1288 

### Call-outs:

Since the fix is a just conditionally enabling a flag, no tests are necessary imo. Calling Cargo Kani with `--dry-run` reveals that `--slice-formula` is disabled when calling `--visualize` or `--no-slice-formula`. 

A test would be more expensive than is needed.

### Testing:

* How is this change tested? 
Checked with dry-run if the extra flag is still being called.

* Is this a refactor change?
Yes

### Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -7,16 +7,5 @@ rustflags = [ # Global lints/warnings. Need to use underscore instead of -.
     # Purposefully disabled lints
     "-Aclippy::expect_fun_call",
     "-Aclippy::or_fun_call",
-
-    # todo address the following lints.
-    "-Aclippy::derive_partial_eq_without_eq",
-    "-Aclippy::manual_map",
-    "-Aclippy::manual_range_contains",
-    "-Aclippy::map_entry",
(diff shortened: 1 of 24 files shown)


tedinski (approved):

tedinski:
If you fix the minor formatting error quickly, we can _probably_ get this in before release today.

PR #1486, opened 2022-08-10
Clean up codegen_niche_literal

### LINKED ISSUE #1485: Cleanup codegen_niche_literal
Similarly to #1456

fzaiser:
### Description of changes: 

Clean up `codegen_niche_literal`: this does not require generating a separate function, similarly to #1457. I believe (and added assertions to that effect) that this code is only called when the whole type is the same size as the discriminant type and the offset of the discriminant is 0. Hence we can simply transmute the given scalar to the result type.

### Resolved issues:

Resolves #1485

### Call-outs

This adds a byte extract instruction. I vaguely remember CBMC struggling with this sometimes. But it seems better than casting to a pointer, adding an offset and dereferencing, which is what we used to do.

### Testing:

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
@@ -274,20 +274,21 @@ impl<'tcx> GotocCtx<'tcx> {
                         Variants::Multiple { tag_encoding, tag_field, .. } => match tag_encoding {
                             TagEncoding::Niche { .. } => {
                                 let niche_offset = layout.fields.offset(*tag_field);
+                                assert_eq!(
+                                    niche_offset,
+                                    Size::ZERO,
+                                    "nonzero offset for niche in scalar"
+                                );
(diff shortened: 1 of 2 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -271,15 +271,24 @@ impl<'tcx> GotocCtx<'tcx> {
...
                             }
                         }
-                        Variants::Multiple { tag_encoding, .. } => match tag_encoding {
+                        Variants::Multiple { tag_encoding, tag_field, .. } => match tag_encoding {
                             TagEncoding::Niche { .. } => {
-                                let offset = match &layout.fields {
-                                    FieldsShape::Arbitrary { offsets, .. } => offsets[0],
-                                    _ => unreachable!("niche encoding must have arbitrary fields"),
-                                };
+                                let niche_offset = layout.fields.offset(*tag_field);
```
danielsn, line 276: Nice instead of hardcoding zero

danielsn (approved):
Nice cleanup

PR #1488, opened 2022-08-10
tests: add tests for deterministic playback of the remaining primitive types

sanjit-bhat:
### Description of changes:

The current tests for the deterministic unit test gen feature (`tests/ui/exe-trace`) don't include tests for `char, bool, NonZero, Result, Option`. This PR adds those tests in. I have manually run the generated unit tests to make sure that they properly trigger assertion failures.

Also, github was telling me I should have newlines at the end of my files, so I added those to all the tests I've created.

### Checklist


tests/ui/exe-trace/bool/main.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// kani-flags: --harness harness --enable-unstable --gen-exe-trace
+
+#[kani::proof]
+pub fn harness() {
+    let bool_1: bool = kani::any();
+    assert!(bool_1 != true);
+    // Note: We can't test a false value yet because any::<bool>() could be any non-zero number.
```
celinval, line 10: Oh, if you want `any::<bool>()` to be pretty in the trace, you can change it to the following:

```rust
impl Arbitrary for bool {
  fn any() -> Self {
    let byte : u8 = crate::any();
    crate::assume(byte < 2);
    byte != 0
  }
}
```

celinval (commented):
Can you please move the comments you are adding at the end of the tests to be a docstring of the harness?

tests/ui/exe-trace/bool/main.rs
sanjit-bhat: This change would be nice, but I think we have to run some performance tests before making it. I created #1490 to track this.

celinval (approved):

PR #1492, opened 2022-08-11
Codegen function contracts from MIR to gotoc.

monal:
### Description of changes: 

Added support to codegen function contracts from MIR to gotoc. 

While handling Kani attributes during codegen, if a `#[kani::assigns(...)]` clause is present (macro yet to be implemented), a new symbol is created for the function contract (using the `Symbol::contract(..)` constructor) and added to the symbol table. 

### Call-outs:

We are only checking the write-set with CBMC. The `[kani::ensures(...)]` and `[kani::requires(...)]` clauses will be inlined as `assert` or `assume` in the future, depending on whether the contract is being checked or replaced. However, CBMC expects all three fields - `#[spec_ensures]`, `#[spec_requires]` and `[spec_assigns]` to be populated while creating a contract symbol. Hence the `ensures` and `requires` are set to be `true` in the newly create contract symbol.

### Testing:

* How is this change tested?
By running Kani regression tests. Need to have end-to-end pipeline before adding additional test cases. Will not merge into main until we have these tests.

* Is this a refactor change?
No

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
@@ -85,7 +85,7 @@ impl<'tcx> GotocCtx<'tcx> {
                         let expr = Expr::symbol_expression(sym.name.clone(), sym.typ.clone());
                         Some(Spec::new(bv, expr, loc))
                     }
-                    None => None, // ignore globals for now.
+                    None => panic!("Global variables are not supported inside modifies clauses."),
                 }
             })
             .collect();
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
@@ -86,9 +86,10 @@ impl<'tcx> GotocCtx<'tcx> {
                         Some(Spec::new(bv, expr, loc))
                     }
                     None => {
-                        self.tcx
-                            .sess
-                            .span_err(mir.span, "Symbol not supported inside modifies clauses.");
+                        self.tcx.sess.span_err(
+                            mir.span,
+                            format!("Symbol {} not supported inside modifies clauses.", basename),
+                        );
                         None
                     }
                 }
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
@@ -1,0 +1,87 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This file contains functions related to codegenning code contracts from MIR into gotoc
+use crate::codegen_cprover_gotoc::GotocCtx;
+use cbmc::goto_program::{Contract, Expr, Spec, Symbol};
+use rustc_ast::NestedMetaItem;
+use rustc_middle::mir::Local;
+use rustc_middle::ty::{self, FnSig};
+
+impl<'tcx> GotocCtx<'tcx> {
+    /// Transforms current function's arguments into expressions
+    fn extract_function_arguments(&mut self, sig: FnSig<'tcx>) -> Vec<Expr> {
+        let args = sig.inputs();
+        args.iter()
+            .enumerate()
+            .map(|(i, t)| {
+                let lc = Local::from_usize(i + 1); // The zeroth index (local variable) is reserved for the return value of the function. Hence, function argument index starts from one.
+                let ident = self.codegen_var_name(&lc);
+                let typ = self.codegen_ty(*t);
+                Expr::symbol_expression(ident, typ)
+            })
+            .collect()
+    }
+
+    /// Transforms current function's return value into an expression
+    fn extract_function_return_value(&mut self, sig: FnSig<'tcx>) -> Expr {
+        let rt = sig.output();
+        let lc = Local::from_usize(0); // The zeroth index (local variable) is reserved for the return value of the function.
+        let ident = self.codegen_var_name(&lc);
+        let typ = self.codegen_ty(rt);
+        Expr::symbol_expression(ident, typ)
+    }
+
+    /// In order to keep the contract symbol self-contained, every function contract clause
+    /// such as `requires`, `ensures`, etc. is wrapped into a lambda expression.
+    /// The binding variables of these lambda expressions are -
+    ///  1) the return value of the function as the first variable, followed by,
+    ///  2) the list of function arguments (in order).
+    fn get_lambda_binding_variables(&mut self, sig: FnSig<'tcx>) -> Vec<Expr> {
+        let mut vars = vec![];
+        let ret = self.extract_function_return_value(sig);
+        vars.push(ret);
+        let args = self.extract_function_arguments(sig);
+        vars.extend(args);
+        return vars;
+    }
+
+    /// Generates a specification containing a lambda expression with "true" as its body
+    fn spec_true(&mut self, sig: FnSig<'tcx>) -> Spec {
+        let bv = self.get_lambda_binding_variables(sig);
+        let mir = self.current_fn().mir();
+        let loc = self.codegen_span(&mir.span);
+        Spec::new(bv, Expr::bool_true(), loc)
+    }
+
+    /// Generates a symbol for the function contract and adds it to the symbol table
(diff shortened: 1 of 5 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Generates a specification containing a lambda expression with "true" as its body
+    fn spec_true(&mut self, sig: FnSig<'tcx>) -> Spec {
+        let bv = self.get_lambda_binding_variables(sig);
+        let mir = self.current_fn().mir();
+        let loc = self.codegen_span(&mir.span);
+        Spec::new(bv, Expr::bool_true(), loc)
+    }
+
+    /// Generates a symbol for the function contract and adds it to the symbol table
+    pub fn codegen_assigns_clause(&mut self, args: Vec<NestedMetaItem>) {
```
celinval, line 58: Can you please expand this documentation? What kind of symbol it creates? Which assumptions do we make?

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -0,0 +1,87 @@
...
+        let mir = self.current_fn().mir();
+        let loc = self.codegen_span(&mir.span);
+        let sig = self.current_fn().sig();
+        let sig =
+            self.tcx.normalize_erasing_late_bound_regions(ty::ParamEnv::reveal_all(), sig.unwrap());
+        let typ = self.fn_typ();
+
+        let name = format!("contract::{}", self.current_fn().name()); // name of the contract symbol
```
celinval, line 66: Can a function be annotated with multiple assigns annotations? If it can, you should probably consider adding an id here. Otherwise, please check that's not the case and emit an error if the user tries to use 2 or more annotations.

celinval (commented):
Where can I find the specification for the annotations you are adding? What are valid inputs? Can a user use multiple annotations in one function?

It might be nice if you populate the original issue (https://github.com/model-checking/kani/issues/1503) with those details.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -0,0 +1,87 @@
...
+
+        // Transform comma-separated "targets" (variables, references, etc.) from the assigns clause into specifications containing lambda expressions
+        let spec_args = args
+            .iter()
+            .map(|a| {
+                let bv = self.get_lambda_binding_variables(sig);
+                let ident = a.meta_item().unwrap().path.segments[0].ident;
+                let sym = self.lookup_local_decl_by_name(&ident.to_string()).unwrap(); // lookup the symbol for the argument in the symbol table
```
celinval, line 74: Can this be a global variable too?

celinval (commented):
I jumped the gun. Sorry! I'm adding a few more comments.

celinval:
Actually, I think it would be great if you document how contracts work in the module documentation (contract.rs). Which functions can be annotated? Which annotations can exist? How do they interact with each other? What is the content allowed in each annotation.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
monal: Good point! I've added a match statement to check if `sym` is a local declaration and I'm ignoring global variables for now.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
monal: Yes, a function can only have one contract. Each contract can have multiple `requires`, `ensures` and `assigns` annotations. They will all be treated as one contract specification collectively. 

If multiple `requires` or `ensures` statements are present, then they will get inlined as separate `assert` or `assume` statements during macro expansion (will send in separate PR). 

Multiple `#[kani::assigns]` statements get wrapped into one `#[kanitool::assigns(..)]` during macro expansion. Hence, we can assume that there is one assigns clause here.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
monal: Done.

monal:
@celinval I've added more documentation to the issue. I also changed "assigns" to "modifies" based on the feedback I received from the team.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+//! This file contains functions related to codegenning code contracts from MIR into gotoc
+use crate::codegen_cprover_gotoc::GotocCtx;
+use cbmc::goto_program::{Contract, Expr, Spec, Symbol};
+use rustc_ast::NestedMetaItem;
+use rustc_middle::mir::Local;
+use rustc_middle::ty::{self, FnSig};
+
+impl<'tcx> GotocCtx<'tcx> {
+    /// Transforms current function's arguments into expressions
+    fn extract_function_arguments(&mut self, sig: FnSig<'tcx>) -> Vec<Expr> {
```
celinval, line 13: This function and the one bellow should be inside functions.rs, but since this is going into a branch for now, it might make sense to keep them here.

Another thing you could do is to create a PR to main that creates these functions inside function.rs and you just use them here. That will reduce the amount of divergence between the feature branch and main.

That said, not a blocker to this PR.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+                let bv = self.get_lambda_binding_variables(sig);
+                let ident = a.meta_item().unwrap().path.segments[0].ident;
+                let basename = &ident.to_string();
+                match self.lookup_local_decl_by_name(basename) {
+                    // lookup the symbol for the argument in the symbol table
+                    Some(sym) => {
+                        let expr = Expr::symbol_expression(sym.name.clone(), sym.typ.clone());
+                        Some(Spec::new(bv, expr, loc))
+                    }
+                    None => None, // ignore globals for now.
```
celinval, line 88: throw an error please saying that this is not supported.

celinval (commented):
Please add an user friendly error for the global variable case. Everything else looks good to me.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
monal: Done. I'm throwing an explicit panic now saying globals aren't supported inside modifies clauses.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+                let bv = self.get_lambda_binding_variables(sig);
+                let ident = a.meta_item().unwrap().path.segments[0].ident;
+                let basename = &ident.to_string();
+                match self.lookup_local_decl_by_name(basename) {
+                    // lookup the symbol for the argument in the symbol table
+                    Some(sym) => {
+                        let expr = Expr::symbol_expression(sym.name.clone(), sym.typ.clone());
+                        Some(Spec::new(bv, expr, loc))
+                    }
+                    None => panic!("Global variables are not supported inside modifies clauses."),
```
celinval, line 88: Can you please replace the `panic!()` with
 ```rust
self.tcx.sess.span_err(mir.span, "Global variables are not supported inside modifies clauses.");
return;
```

Comment (monal) [reply kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs]: Done.

File: kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs [hunk 7]
```diff
@@ -0,0 +1,105 @@
...
+                match self.lookup_local_decl_by_name(basename) {
+                    // lookup the symbol for the argument in the symbol table
+                    Some(sym) => {
+                        let expr = Expr::symbol_expression(sym.name.clone(), sym.typ.clone());
+                        Some(Spec::new(bv, expr, loc))
+                    }
+                    None => {
+                        self.tcx
+                            .sess
+                            .span_err(mir.span, "Symbol not supported inside modifies clauses.");
```
zhassan-aws, line 91: Do we want to print the name of the symbol in the error message?

celinval (approved):
Thank you

Issue #1501, opened 2022-08-11
Add hooks for pre and post conditions in function contracts.


monal:
Add hooks to replace `precondition(x)` and `postcondition(x)` as `assume(x)` or `assert(x)` statements. If a function contract is being enforced, `precondition(x)` gets converted to `assume(x)` and `postcondition(x)` gets converted to `assert(x)`.  If a function is being replaced by its contract, `precondition(x)` gets converted to `assert(x)` and `postcondition(x)` to `assume(x)`.

celinval:
This is no longer in the scope of the contracts work.

PR #1511, opened 2022-08-11
Add hooks for handling pre and post conditions in function contracts.

monal:
### Description of changes: 

Background: During macro expansion (yet to be implemented), `#[kani::requires(x)]` will get inlined as `kani::precondition(x)` in the function body before the actual function body begins. Similarly, `#[kani::ensures(x)]` will get inlined as `kani::postcondition(x)` at the end of the function body. The actual function body will get wrapped in an inner function that gets called and the corresponding return value is returned from the wrapper function. 

This change includes - 
1) Hooks for replacing `precondition(x)` calls into `assert` or `assume` statements depending on whether a function contract is being checked or if a function is being replaced by its contract.  
2) Similarly, add hooks for replacing `postcondition(x)` calls into `assert` or `assume` statements.
2) Add Kani flag `--enforce-contracts` to checks all function contracts.
3) Add Kani flag `--replace-with-contracts` for replacing functions with their contracts, if a contract exists.

### Resolved issues:

Resolves https://github.com/model-checking/kani/issues/1501

### Testing:

* How is this change tested?
By running Kani regression tests. Need to have end-to-end pipeline before adding additional test cases. Will not merge into main until we have these tests.

* Is this a refactor change?
No

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
@@ -256,42 +256,25 @@ impl<'tcx> GotocHook<'tcx> for Postcondition {
     fn handle(
         &self,
         tcx: &mut GotocCtx<'tcx>,
-        _instance: Instance<'tcx>,
+        instance: Instance<'tcx>,
         mut fargs: Vec<Expr>,
-        _assign_to: Place<'tcx>,
+        assign_to: Place<'tcx>,
         target: Option<BasicBlock>,
         span: Option<Span>,
     ) -> Stmt {
         assert_eq!(fargs.len(), 1);
-        let cond = fargs.remove(0).cast_to(Type::bool());
-        let target = target.unwrap();
+        // let cond = fargs.remove(0).cast_to(Type::bool());
+        // let target = target.unwrap();
         let loc = tcx.codegen_span_option(span);
 
         if tcx.queries.get_enforce_contracts() {
-            // If function contract is being checked,
-            // convert "postcondition(x)" into "assert!(x)"
-            let tmp = tcx.gen_temp_variable(cond.typ().clone(), loc).to_expr();
-            Stmt::block(
-                vec![
-                    Stmt::decl(tmp.clone(), Some(cond), loc),
-                    tcx.codegen_assert(tmp.clone(), PropertyClass::Assertion, "postcondition", loc),
-                    Stmt::goto(tcx.current_fn().find_label(&target), loc),
-                ],
-                loc,
-            )
+            fargs.push(Expr::string_constant("Post-condition failed"));
+            Assert.handle(tcx, instance, fargs, assign_to, target, span)
         } else if tcx.queries.get_replace_with_contracts() {
-            // If function is being replace with contract,
-            // convert "postcondition(x)" into "assume!(x)"
-            Stmt::block(
-                vec![
-                    Stmt::assume(cond, loc),
-                    Stmt::goto(tcx.current_fn().find_label(&target), loc),
-                ],
-                loc,
-            )
+            Assume.handle(tcx, instance, fargs, assign_to, target, span)
         } else {
             // The function contract is being ignored in this case.
-            Stmt::block(vec![Stmt::goto(tcx.current_fn().find_label(&target), loc)], loc)
+            Stmt::block(vec![Stmt::goto(tcx.current_fn().find_label(&target.unwrap()), loc)], loc)
         }
     }
 }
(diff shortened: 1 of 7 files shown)


kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -245,6 +245,107 @@ impl<'tcx> GotocHook<'tcx> for Panic {
...
+        &self,
+        tcx: &mut GotocCtx<'tcx>,
+        _instance: Instance<'tcx>,
+        mut fargs: Vec<Expr>,
+        _assign_to: Place<'tcx>,
+        target: Option<BasicBlock>,
+        span: Option<Span>,
+    ) -> Stmt {
+        assert_eq!(fargs.len(), 1);
+        let cond = fargs.remove(0).cast_to(Type::bool());
```
celinval, line 266: I would suggest you to implement something like:

```rust
struct PostCondition {
  // Should we apply this hook (i.e.: either enforce or check precondition).
  enabled: bool,
}

impl<'tcx> GotocHook<'tcx> for Postcondition {
  fn hook_applies(&self, tcx: TyCtxt<'tcx>, instance: Instance<'tcx>) -> bool {
    self.enabled && matches_function(tcx, instance, "KaniPostcondition")
  }

  fn handle (/*args*/) {
    if tcx.queries.get_enforce_contracts() {
      fargs.add(Expr::string_constant("Post-condition failed");
      Assert::handle(tcx, instance, fargs, assign_to, target, span);
    } else {
      Assume::handle(tcx, instance, fargs, assign_to, target, span);
    }
  }
}
```

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -245,6 +245,107 @@ impl<'tcx> GotocHook<'tcx> for Panic {
...
+            )
+        } else {
+            // The function contract is being ignored in this case.
+            Stmt::block(vec![Stmt::goto(tcx.current_fn().find_label(&target), loc)], loc)
+        }
+    }
+}
+
+/// Hook for handling preconditions in function contracts.
+struct Precondition;
```
celinval, line 300: The same idea for this one.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -316,6 +417,49 @@ impl<'tcx> GotocHook<'tcx> for PtrWrite {
     }
 }
 
+/// Hook to set flag to nondet the function body if a function is being replaced by its contract.
+struct ReplaceFunctionBody;
+
+impl<'tcx> GotocHook<'tcx> for ReplaceFunctionBody {
+    fn hook_applies(&self, tcx: TyCtxt<'tcx>, instance: Instance<'tcx>) -> bool {
+        matches_function(tcx, instance, "KaniReplaceFunctionBody")
```
celinval, line 425: I am not sure I understand how this hook would work. Who is calling this function? Note that the hooks get applied at a function call level not function body code generation.

celinval (commented):
I have a question about usability. What will happen by default? Are we going to check the contracts or ignore them?

Do you have any documentation on how all the pieces fit together? What exactly happens when the user annotates their function with a contract annotation? A bit more context will help figuring out what you are trying to accomplish.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
monal: We want to convert the postcondition as follows - 
1) to an `assert` when a function is being checked against its contract (calling Kani with the `--enforce-contracts` flag)
2) to an `assume` if the function is being replaced with its contract (calling Kani with the `--replace-with-contracts` flag)
3) To an empty statement, otherwise (calling Kani without any flags)

In this snippet, enabled must be set to true if `tcx.queries.get_enforce_contracts() == true` or `tcx.queries.get_replace_with_contracts() == true`. Where can I set this value? Also, how do I ensure that in the last case (that is call Kani normally without any contract checking/replacing flags) the inlined `kani::postcondition` gets converted to an empty line?

monal: I've refactored my code to use `Assert.handle(..)` and `Assume.handle(..)` within the same if-else statement.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
monal: The macro expansion stage would call this function. The code would look something like this - 
```
if kani::replace_function_body() {
  let ret = kani::any();
} else {
  let ret = .... //call body closure.
}
```

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -307,42 +290,52 @@ impl<'tcx> GotocHook<'tcx> for Precondition {
...
+        //         vec![
+        //             Stmt::decl(tmp.clone(), Some(cond), loc),
+        //             tcx.codegen_assert(tmp.clone(), PropertyClass::Assertion, "precondition", loc),
+        //         ],
+        //         loc,
+        //     )
+        // } else {
+        //     // The function contract is being ignored in this case.
+        //     Stmt::block(vec![Stmt::goto(tcx.current_fn().find_label(&target), loc)], loc)
+        // }
```
fzaiser, line 338: Did you commit commented code intentionally?

monal: Oops, didn't mean to do that. I've removed them now, thanks! 👍

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -316,6 +380,49 @@ impl<'tcx> GotocHook<'tcx> for PtrWrite {
...
+        assign_to: Place<'tcx>,
+        target: Option<BasicBlock>,
+        span: Option<Span>,
+    ) -> Stmt {
+        assert_eq!(fargs.len(), 0);
+        let loc = tcx.codegen_span_option(span);
+        let target = target.unwrap();
+        let pe = unwrap_or_return_codegen_unimplemented_stmt!(tcx, tcx.codegen_place(&assign_to))
+            .goto_expr;
+        if tcx.queries.get_replace_with_contracts() {
```
celinval, line 405: maybe replace this by:

```rust
let value = if tcx.queries.get_replace_with_contracts() { Expr::c_true() } else { Expr::c_false() };
Stmt::block(
    vec![
       pe.assign(value, loc),
       Stmt::goto(tcx.current_fn().find_label(&target), loc),
    ],
    loc,
)
        
```

celinval (approved):
Looks good! Just one comment. Please address before pushing.

monal:
Thanks for reviewing, @celinval! ❤️  I've made the last change you suggested.

PR #1512, opened 2022-08-11
feat: enable deterministic playback for variables optimized out due to `--slice-formula`

### LINKED ISSUE #1500: Det playback: Enable running without `--slice-formula`
Deterministic playback currently runs with `--slice-formula`, which means that users will have incorrect det test cases if their verification conditions don't include all variables. This issue investigates the behavior of removing `--slice-formula` and still having accurate det test cases.

sanjit-bhat:
### Description of changes: 

Before, deterministic playback ran with the `--slice-formula` CBMC flag. This meant that programs that had symbolic variables not used in the verification condition would have incomplete deterministic test cases.

To fix this, I had to disable `--slice-formula` when running with `--gen-exe-trace`. But it's not as easy as doing that, since our parser picks up more deterministic values as a result. See #1500 for a discussion on how I fixed this issue by changing what the det vals extractor matches on.

### Resolved issues:

Closes #1500.

### Testing:

I added another deterministic playback test. This test contains symbolic variables that aren't included in the verification condition.

### Checklist


library/kani/src/exe_trace.rs
```diff
@@ -24,6 +24,16 @@ pub fn exe_trace_run<F: Fn()>(mut local_det_vals: Vec<Vec<u8>>, proof_harness: F
...
     // Since F is a type argument, there should be a direct, static call to proof_harness().
     proof_harness();
+    // This code will not run if a user assertion fails on deterministic playback.
+    // But if you comment out the failing assertion during playback,
+    // this can be used to know if too many det vals were loaded into the deterministic test case.
+    DET_VALS.with(|glob_det_vals| {
+        let ref_glob_det_vals = &*glob_det_vals.borrow();
+        assert!(
+            ref_glob_det_vals.is_empty(),
+            "At the end of deterministic playback, there were still these deterministic values left over `{:?}`", ref_glob_det_vals
```
celinval, line 34: Maybe say that this can happen because:
1. The code being verified has changed and the user should update their generated tests.
2. There is a bug on kani. Please report the issue here: https://github.com/model-checking/kani/issues/new?assignees=&labels=bug&template=bug_report.md

Once you create the documentation, you could replace this by a "for more details, see `<link-to-doc>`".

celinval (approved):
It looks good. I definitely prefer the solution you implemented better! Thanks

library/kani/src/exe_trace.rs
sanjit-bhat: Good suggestion. Done

Issue #1271, opened 2022-06-09
Missing functions produce non-informative property descriptions


adpaco-aws:
Upgrading the CBMC version to 5.59.0 caused regression failures related to missing functions. These are now failing with the description "assertion", apparently because they do not include source locations.

adpaco-aws:
According to @tautschnig , this is due to https://github.com/diffblue/cbmc/pull/6902 but the intended behavior was to get the following message instead:

```
	 - Status: FAILURE
	 - Description: "undefined function should be unreachable"
```

This is what should be expected when we upgrade to the next version of CBMC. In the meantime, I've submitted changes to the expected results according to the current output in #1270

tautschnig:
I believe the problem is that the symbols you create for missing functions do not have a source location. Looking at `_ZN60_$LT$alloc..string..String$u20$as$u20$core..clone..Clone$GT$5clone17hd1d44edec193123bE` as is generated from the following piece of code (due to @jaisnan :
````rust
#[kani::proof]
fn main() {
    let x = String::from("foo");
    let y = x.clone();
    assert_eq!("foo", y);
}
````
I see that the source location is `nil`, which is unexpected. The CBMC code base would expect that there at least be a source location associated with the declaration, even when the definition is unavailable. This might be a C legacy, but in Kani you'd have a source location.

adpaco-aws:
Fix was done in #1513 but didn't close this issue automatically.

PR #1513, opened 2022-08-11
Fix description updates for missing function properties

### LINKED ISSUE #1424: Missing function assertions aren't generating useful failure messages
```
Failed check: assertion
```

Should state what function and that it is missing.

adpaco-aws:
### Description of changes: 

Changes the conditions to determine whether a property corresponds to a missing function definition. Updates the related UI tests.

### Resolved issues:

Resolves #1424 #1271 

### Testing:

* How is this change tested? Existing regression. The tests have been updated with stronger requirements.

### Checklist


kani-driver/src/cbmc_output_parser.rs
```diff
@@ -816,9 +815,9 @@ fn has_check_failure(properties: &Vec<Property>, description: &str) -> bool {
 fn modify_undefined_function_checks(mut properties: Vec<Property>) -> (Vec<Property>, bool) {
     let mut has_unknown_location_checks = false;
     for mut prop in &mut properties {
-        if prop.description.contains(ASSERTION_FALSE)
-            && extract_property_class(prop).unwrap() == DEFAULT_ASSERTION
+        if prop.description == DEFAULT_ASSERTION
             && prop.source_location.file.is_none()
+            && prop.source_location.function.is_some()
         {
             prop.description = "Function with missing definition is unreachable".to_string();
```
celinval, line 822: It looks like the propery_id as well as the location include the name of the missing function. Can we use that to print a nicer error message? The only issue is that the name is mangled. :(

celinval (commented):
Just one comment. Thanks

kani-driver/src/cbmc_output_parser.rs
tedinski: It's rust code now. Demangle? e.g. https://crates.io/crates/rustc-demangle

adpaco-aws: Thanks for the suggestion! AFAIK we can use `rustc-demangle` to demangle the name. For some reason, demangling works only partially for missing function definitions, but still provides a nice result:

Screen Shot 2022-08-12 at 13 09 09

I could also rebuild the property name with the demangled function name in #1514. Can you review this PR again?

kani-driver/Cargo.toml
```diff
@@ -24,6 +24,7 @@ clap = "2.34"
 glob = "0.3"
 toml = "0.5"
 regex = "1.6"
+rustc-demangle = { version = "0.1.21" }
```
fzaiser, line 27: Why not just
```suggestion
rustc-demangle = "0.1.21"
```

tests/ui/missing-function/rust-by-example-description/expected
```diff
@@ -1,2 +1,4 @@
 Status: FAILURE\
-Description: "assertion"
+Description: "Function
```
fzaiser, line 2: Why not add the name of the function as well, to ensure that the demangling works corectly?

adpaco-aws: Maybe I'm mistaken, but I was under the impression that mangled names may change across different versions of the compiler. But you're right, let's add it and we'll remove if that's the case in future toolchain updates.

adpaco-aws: CI failed because the last (mangled) part is not consistent across different platforms. Went back to two lines but the first one includes the demangled part of the function name.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -811,16 +812,24 @@ fn has_check_failure(properties: &Vec<Property>, description: &str) -> bool {
...
+            // Missing functions come with mangled names. `demangle` produces
+            // the demangled version if it's a mangled name, but for some reason
+            // the name is not being fully demangled. For example:
+            //
+            // function: _ZN60_$LT$alloc..string..String$u20$as$u20$core..clone..Clone$GT$5clone17h830b64d2ac32f613E
+            // demangle(function): <alloc::string::String as core::clone::Clone>::clone::h830b64d2ac32f613
+            //
+            // Note the last part, which doesn't appear to be demangled.
+            // Still, it's more readable than the mangled name.
+            let modified_description = format!("Function {} with missing definition is unreachable", demangle(function));
```
fzaiser, line 831: Can you try `{:#}` instead of `{}`? I think the alternative formatting of `Demangle` should omit the hash at the end, which avoids the platform differences. I believe with that change, you can add the whole line to the `expected` files.

fzaiser: `rustfilt` does something similar: https://github.com/luser/rustfilt/blob/3c81f107b73fdcf7bbb98b426c1214f15946ec90/src/main.rs#L47

tests/ui/missing-function/rust-by-example-description/expected
fzaiser: Thanks! I think you can get rid of the hash at the end tho, see below.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: Thanks @fzaiser, that did the trick!

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -811,16 +812,24 @@ fn has_check_failure(properties: &Vec<Property>, description: &str) -> bool {
...
+            // Missing functions come with mangled names. `demangle` produces
+            // the demangled version if it's a mangled name, but for some reason
+            // the name is not being fully demangled. For example:
+            //
+            // function: _ZN60_$LT$alloc..string..String$u20$as$u20$core..clone..Clone$GT$5clone17h830b64d2ac32f613E
+            // demangle(function): <alloc::string::String as core::clone::Clone>::clone::h830b64d2ac32f613
+            //
+            // Note the last part, which doesn't appear to be demangled.
+            // Still, it's more readable than the mangled name.
+            let modified_description = format!("Function {:#} with missing definition is unreachable", demangle(function));
```
celinval, line 831: nit: can you add quotes around the name of the function?

celinval (approved):
That looks much nicer! Just one little comment.

tedinski (approved):
Woo!

PR #1514, opened 2022-08-11
More reliable property attributes extraction

adpaco-aws:
### Description of changes: 

Improves how we extract property attributes, such as property classes. This work was done while working on #1513 but it turns we don't have to use the property class there, so I opened this PR. Also, opened https://github.com/diffblue/cbmc/issues/7069 to discuss the possibility of adding these attributes separately in the JSON output.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

### Checklist

### DIFF
--- a/kani-driver/src/cbmc_output_parser.rs
+++ b/kani-driver/src/cbmc_output_parser.rs
@@ -250,6 +250,10 @@ impl Property {
         write!(&mut fmt_str, "{id}")?;
         Ok(fmt_str)
     }
+
+    pub fn has_property_class_format(string: &str) -> bool {
+        string == "NaN" || string.chars().all(|c| c.is_ascii_lowercase() || c == '_' || c == '-')
+    }
 }
 
 impl<'de> serde::Deserialize<'de> for PropertyId {
@@ -284,11 +288,8 @@ impl<'de> serde::Deserialize<'de> for PropertyId {
         let attributes_tuple = match property_attributes.len() {
             // The general case, where we get all the attributes
             3 => {
-                if property_attributes[1]
-                    .chars()
-                    .all(|c| c.is_ascii_lowercase() || c == '_' || c == '-')
-                    || property_attributes[1] == "NaN"
-                {
+                // Since mangled function names may contain `.`, we check if
+                if Property::has_property_class_format(property_attributes[1]) {
                     let name = format!("{:#}", demangle(property_attributes[2]));
                     (Some(name), Some(property_attributes[1]), property_attributes[0])
                 } else {
@@ -299,15 +300,11 @@ impl<'de> serde::Deserialize<'de> for PropertyId {
                 }
             }
             2 => {
-                // The case where `property_attributes[1]` could be a function or a
-                // class. If `class_re` matches it, it's likely a class (functions
-                // are usually mangled names which contain many other symbols).
-                if property_attributes[1]
-                    .chars()
-                    .all(|c| c.is_ascii_lowercase() || c == '_' || c == '-')
-                    || property_attributes[1] == "NaN"
-                {
-                    // assert!(false);
+                // The case where `property_attributes[1]` could be a function
+                // or a class. If it has the class format, then it's likely a
+                // class (functions are usually mangled names which contain many
+                // other symbols).
+                if Property::has_property_class_format(property_attributes[1]) {
                     (None, Some(property_attributes[1]), property_attributes[0])
                 } else {
                     let name = format!("{:#}", demangle(property_attributes[1]));
--- a/kani-driver/src/cbmc_output_parser.rs
+++ b/kani-driver/src/cbmc_output_parser.rs
@@ -324,7 +324,11 @@ impl<'de> serde::Deserialize<'de> for PropertyId {
         // Return tuple after converting counter from string into number.
         // Safe to do because we've checked the format earlier.
         let class = String::from(attributes_tuple.1);
-        Ok(PropertyId { fn_name: attributes_tuple.0, class, id: attributes_tuple.2.parse().unwrap() })
+        Ok(PropertyId {
+            fn_name: attributes_tuple.0,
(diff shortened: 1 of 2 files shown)


kani-driver/src/cbmc_output_parser.rs
```diff
@@ -910,12 +910,78 @@ fn remove_check_ids_from_description(mut properties: Vec<Property>) -> Vec<Prope
...
+///
+/// However, there are cases where we only get two attributes:
+///  * `<class>.<counter>` (the function is a CBMC builtin)
+///  * `<function>.<counter>` (missing function definition)
+///
+/// In these cases, we try to determine if the attribute is a function or not
+/// based on regular expressions (we assume property classes are a combination
+/// of lowercase letters and the character `_`). But this is not completely
+/// reliable. CBMC should be able to provide these attributes as separate fields
+/// in the JSON output: <https://github.com/diffblue/cbmc/issues/7069>
```
danielsn, line 925: We should probably have a tracking issue on the kani side too

danielsn (dismissed):

danielsn:
Is there any way to test this other than end-to-end tests?

celinval (commented):
What if instead of having these util functions, you implement a custom deserializer for property id? Something like:

```rust
#[derive(Clone, Debug, Deserialize)]
pub struct Property {
    pub description: String,
    #[serde(rename="property")]
    pub property_id: PropertyId,
    #[serde(rename = "sourceLocation")]
    pub source_location: SourceLocation,
    pub status: CheckStatus,
    pub reach: Option<CheckStatus>,
    pub trace: Option<Vec<TraceItem>>,
}

pub struct PropertyId {
   pub fn_name: Option<String>,
   pub class: Option<String>,
   pub id: u32,
}

impl<'de> serde::Deserialize<'de> for PropertyId {
    fn deserialize<D: Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
        let id_str = String::deserialize(d)?;
        // TODO: Add your code here
    }
}

```

That way, we solve our side of the problem.

adpaco-aws:
> What if instead of having these util functions, you implement a custom deserializer for property id?

That's a great idea, thanks!

> Is there any way to test this other than end-to-end tests?

Yes, I will implement @celinval's suggestion and add unit tests as well. Thanks!

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -914,12 +987,23 @@ fn remove_check_ids_from_description(mut properties: Vec<Property>) -> Vec<Prope
     properties
 }
 
-/// Extracts the property class from the property string.
-///
-/// Property strings have the format `([<function>.]<property_class_id>.<counter>)`
-pub fn extract_property_class(property: &Property) -> Option<&str> {
-    let property_class: Vec<&str> = property.property.rsplitn(3, '.').collect();
-    if property_class.len() > 1 { Some(property_class[1]) } else { None }
+fn get_property_class(property: &Property) -> Option<String> {
```
celinval, line 990: Can you please make these functions methods of Property? Also, can you remove the "get_" per Rust API guidelines: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -1032,3 +1116,106 @@ fn determine_verification_result(properties: &[Property]) -> bool {
         properties.iter().filter(|prop| prop.status == CheckStatus::Failure).count();
     number_failed_properties == 0
 }
+
+#[test]
```
celinval, line 1120: That's awesome! Can you please wrap these with a test module?
```
#[cfg(test)]
mod tests {
}

**Review by celinval — COMMENTED** [review]
It looks good. Just some minor comments. And can you please fix the regression?

File: kani-driver/src/cbmc_output_parser.rs [hunk 4]
```diff
@@ -215,14 +215,87 @@ impl ParserItem {
...
+                (Some(property_attributes[2]), Some(property_attributes[1]), property_attributes[0])
+            }
+            2 => {
+                // The case where `property_attributes[1]` could be a function or a
+                // class. If `class_re` matches it, it's likely a class (functions
+                // are usually mangled names which contain many other symbols).
+                if property_attributes[1].chars().all(|c| c.is_ascii_lowercase() || c == '_') {
+                    (None, Some(property_attributes[1]), property_attributes[0])
+                } else {
+                    (Some(property_attributes[1]), None, property_attributes[0])
```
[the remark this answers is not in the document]
adpaco-aws: That's not a bad idea. I haven't checked if all these cases correspond to missing definitions though, so I'll do that before.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: Done.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: Yes, done.

adpaco-aws:
Sorry for the delay.

The previous version still had some issues due to mangled function names containing dots, so now we also check the format in the case where we get 3 attributes from the property string.

I will fix the `clippy` warnings tomorrow. In addition, I will check if all cases where `class.is_none()` correspond to missing function definitions (I suspect that's true) and implement @celinval suggestion.

adpaco-aws:
> I will fix the `clippy` warnings tomorrow. In addition, I will check if all cases where `class.is_none()` correspond to missing function definitions (I suspect that's true) and implement @celinval suggestion.

This is done. The PR is ready for review.

tedinski (approved):
Happy to see more tests here. LGTM once it's updated?

tedinski (approved):

celinval (approved):
Thanks

PR #1520, opened 2022-08-15
Zero init for common types

danielsn:
### Description of changes: 

Currently, we have support for zero-initilizing bit-vector types in cprover, but not aggregate types.
This PR adds support for aggregate types also.

### Resolved issues:

Resolves #1494 

### Call-outs:
This is phase 1 of a two phase process to add support for zero-init to Kani.
This was split from #1230 to allow for a second PR to address the need to guard init with a config.

### Testing:

* How is this change tested? Using the tests in #1230

* Is this a refactor change? No

### Checklist
- [ ] Each commit message has a non-empty body, explaining why the change was made
- [ ] Methods or procedures are documented
- [ ] Regression or unit tests are included, or existing tests cover the modified code
- [ ] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #1521, opened 2022-08-15
Add zero-init, and put it behind an unsound feature flag

### LINKED ISSUE #1493: Mechanism to safely enable unsound experimental features

danielsn:
### Description of changes: 

1. Adds support for zero initializing otherwise uninit variables during codegen.  This may enable CBMC to do more constant propegation, but at a potential soundness cost.
2. Adds a new flag for potentially unsound features.  This requires users to explicitly opt-in to using potentially unsound features.
3. Use that flag to guard the 0 init code.

### Resolved issues:

Resolves #1493 

### Call-outs:

Using this may improve performance, at a potential soundness cost. Since the tradeoff is not clear, putting it behind a flag to allow experimentation, but keeping the current default for now.

Depends on #1520 

### Testing:

* How is this change tested?
  1. Existing regression tests. Turned on the flag, all regressions still pass, except for ones testing maybe_uninit, which we would expect to fail.
  2. Existing regression tests. Flag off, all regressions still pass.
  3. New test which maybe_units a variable and checks that it is all 0.
  
### Checklist


celinval (changes requested):
If I recall this correctly, we had agreed to put unsound stuff behind a compilation flag instead of a switch.

celinval (commented):
I have mixed feelings about this one. It might be nice to get a second opinion... (@zhassan-aws?)

Honestly, I would like us to have a better defined story for how we should handle under-approximations. For example, how can we reflect the possibly unsound results in the verification report.

I was also hoping that an experimental change would be more contained too.

What if instead of adding `--enable-unsound-experiments --zero-init-vars` you add something like `--enable-unsound-experiments=[zero-init-vars | <other_opt>]+`. This will allow a more generic parser code, and the next experimental feature would only need to add a new value instead of adding new switches everywhere.

celinval:
I forgot to mention, but it might be nice to add an expected test that ensures that this feature is not available in the release mode.

.github/workflows/kani.yml
```diff
@@ -33,6 +33,25 @@ jobs:
...
+      - name: Setup Kani Dependencies
+        uses: ./.github/actions/setup
+        with:
+            os: ubuntu-20.04
+
+      - name: Build Kani
+        run: cargo build
+
+      - name: Execute Kani regression
+        run: ./scripts/kani-regression.sh
```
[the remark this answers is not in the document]
danielsn: Yes, to ensure that "unsound experiments" don't break the regressions

kani-driver/src/args.rs
```diff
@@ -71,6 +71,10 @@ pub struct KaniArgs {
     /// Enable usage of unstable options
     #[structopt(long, hidden_short_help(true))]
     pub enable_unstable: bool,
+    /// Enable usage of potentially unsound experimental features
+    #[cfg(feature = "unsound_experiments")]
+    #[structopt(long, hidden = true, requires("enable-unstable"))]
+    pub enable_unsound_experiments: bool,
```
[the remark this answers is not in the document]
danielsn: Fair, I'll remove the extra flag

kani-compiler/kani_queries/src/unsound_experiments.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#![cfg(feature = "unsound_experiments")]
+
+#[derive(Debug, Default)]
+pub struct UnsoundExperiments {
+    pub zero_init_vars: bool,
```
celinval, line 8: Can you add a comment here please?

celinval (approved):

kani-driver/src/unsound_experiments.rs
```diff
@@ -0,0 +1,35 @@
...
+    /// performance by allowing CBMC to do more constant propegation.
+    /// Unfortunatly, it is unsafe to use for production code, since it may unsoundly hide bugs.
+    /// Marked as `unsound` to prevent use outside of experimental contexts.
+    #[structopt(long, hidden_short_help(true), requires("enable-unstable"))]
+    pub unsound_experiment_zero_init_vars: bool,
+}
+
+impl UnsoundExperimentArgs {
+    pub fn process_args(&self) -> Vec<OsString> {
+        self.print_warnings();
```
zhassan-aws, line 20: Will this get printed even without the option enabled?

danielsn: `print_warnings` only prints for enabled options

zhassan-aws (approved):

zhassan-aws:
The refactoring to move all unsound experimental code to a separate module made things much cleaner. Thanks!

kani-compiler/kani_queries/src/unsound_experiments.rs
danielsn: Done

PR #1526, opened 2022-08-16
Mitigate issue with dangling pointers and `memcmp`

### LINKED ISSUE #1489: Investigate failed check in memcmp
This was discovered in #1332, but also occurs in a synchronous context:

```rust
#[kani::proof]
pub fn main() {
    let coll: Result<Vec<u32>, ()> = std::iter::empty().collect();
    assert_eq!(Ok(vec![]), coll);
}
```

Running this with Kani gives the result:
```
SUMMARY:
 ** 2 of 436 failed
Failed Checks: memcmp region 1 readable
 File: "<builtin-library-memcmp>", line 19, in memcmp
Failed Checks: memcpy region 2 readable
 File: "<builtin-library-memcmp>", line 21, in memcmp

VERIFICATION:- FAILED
```

fzaiser:
### Description of changes: 

The following assertion currently fails in Kani: `assert_eq!(Vec::<u8>::new(), Vec::<u8>::new())`. The reason is that an empty vector allocates a dangling pointer, which is then passed to `memcmp`, but zero bytes will be compared. CBMC does not allow dangling pointers there even if they are not read (see #1489).

Therefore this PR adds a hook to intercept calls to `memcmp` and if the number of bytes to compare is zero it does not call `memcmp` but just checks that the pointers are not null. We should also check that they are aligned, but I'm not sure how to find the alignment at this stage of the translation.

### Resolved issues:

Resolves #1489

### Call-outs:

I'm not happy with this special casing. It would be good to have a more general solution for other C functions as well. An example is `memcpy`, but I wasn't able to produce a test case with safe Rust code. Suggestions welcome.

We should also check that the pointers are aligned, but I'm not sure how to find the alignment of a type at this stage.

### Testing:

* How is this change tested? Added a regression test.

### Checklist


kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -317,6 +317,64 @@ impl<'tcx> GotocHook<'tcx> for SliceFromRawPart {
...
+        mut fargs: Vec<Expr>,
+        assign_to: Place<'tcx>,
+        target: Option<BasicBlock>,
+        span: Option<Span>,
+    ) -> Stmt {
+        let loc = tcx.codegen_span_option(span);
+        let target = target.unwrap();
+        let first = fargs.remove(0);
+        let second = fargs.remove(0);
+        let count = fargs.remove(0);
```
YoshikiTakashima, line 343: `fargs` might be big for macro-generated code. You may consider doing something like the below link

https://stackoverflow.com/questions/43026118/what-is-the-idiomatic-way-to-remove-the-first-n-elements-in-a-mutable-vec

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -317,6 +317,64 @@ impl<'tcx> GotocHook<'tcx> for SliceFromRawPart {
...
+        let (count_var, count_decl) =
+            tcx.decl_temp_variable(count.typ().clone(), Some(count), Location::none());
+        let (first_var, first_decl) =
+            tcx.decl_temp_variable(first.typ().clone(), Some(first), Location::none());
+        let (second_var, second_decl) =
+            tcx.decl_temp_variable(second.typ().clone(), Some(second), Location::none());
+        let is_count_zero = count_var.clone().is_zero();
+        // TODO: we should also check that the pointers have correct alignment.
```
YoshikiTakashima, line 351: Would this effect soundness? If so, tracking issue.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
YoshikiTakashima: Could also use `into_iter()` followed by 3 nexts.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -317,6 +317,64 @@ impl<'tcx> GotocHook<'tcx> for SliceFromRawPart {
     }
 }
 
+/// This hook intercepts calls to `memcmp` and skips CBMC's pointer checks if the number of bytes to be compared is zero.
```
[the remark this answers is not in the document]
danielsn: `C` pseudocode is often a good way to do this

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
danielsn: `fargs` should have exactly that set of arguments, no more no less

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -317,6 +317,64 @@ impl<'tcx> GotocHook<'tcx> for SliceFromRawPart {
...
+/// See issue https://github.com/model-checking/kani/issues/1489
+pub struct MemCmp;
+
+impl<'tcx> GotocHook<'tcx> for MemCmp {
+    fn hook_applies(&self, tcx: TyCtxt<'tcx>, instance: Instance<'tcx>) -> bool {
+        let name = with_no_trimmed_paths!(tcx.def_path_str(instance.def_id()));
+        name.contains("memcmp")
```
celinval, line 327: I'm wondering if this check is too open...

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
fzaiser: Yes,`memcmp` takes exactly 3 arguments, so this shouldn't be an issue.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
fzaiser: Fixed in a different way, see other comment below

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -317,6 +317,64 @@ impl<'tcx> GotocHook<'tcx> for SliceFromRawPart {
...
+        let count = fargs.remove(0);
+        let (count_var, count_decl) =
+            tcx.decl_temp_variable(count.typ().clone(), Some(count), Location::none());
+        let (first_var, first_decl) =
+            tcx.decl_temp_variable(first.typ().clone(), Some(first), Location::none());
+        let (second_var, second_decl) =
+            tcx.decl_temp_variable(second.typ().clone(), Some(second), Location::none());
+        let is_count_zero = count_var.clone().is_zero();
+        // TODO: we should also check that the pointers have correct alignment.
+        // At this point, we don't have the Rust types available anymore, however, so this is more difficult.
```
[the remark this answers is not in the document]
fzaiser: Turns out that the pointers that `memcmp` accepts are `*const u8`, so their alignment is 1, so there is nothing to check here.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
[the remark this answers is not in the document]
fzaiser: Done.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -317,6 +317,64 @@ impl<'tcx> GotocHook<'tcx> for SliceFromRawPart {
...
+        target: Option<BasicBlock>,
+        span: Option<Span>,
+    ) -> Stmt {
+        let loc = tcx.codegen_span_option(span);
+        let target = target.unwrap();
+        let first = fargs.remove(0);
+        let second = fargs.remove(0);
+        let count = fargs.remove(0);
+        let (count_var, count_decl) =
+            tcx.decl_temp_variable(count.typ().clone(), Some(count), Location::none());
```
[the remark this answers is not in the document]
fzaiser: Good point, done.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
fzaiser: Yes, you were right :D It made the tests fail. I fixed it to match the exact name (`core::slice::cmp::memcmp`).

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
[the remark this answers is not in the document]
celinval: That's a good point! :)

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
celinval: We used to have hooks for a bunch of memory operations. I do remember they had to also take into consideration the top module being named `std::`. @danielsn might have better insight why that was the case.

danielsn: The modules get re-exported. So we often see it either as "std" or "core", depending on whether it was re-exported.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -317,6 +317,72 @@ impl<'tcx> GotocHook<'tcx> for SliceFromRawPart {
...
+/// first_var = first;
+/// second_var = second;
+/// count_var == 0 && first_var != NULL && second_var != NULL ? 0 : memcmp(first_var, second_var, count_var)
+/// ```
+pub struct MemCmp;
+
+impl<'tcx> GotocHook<'tcx> for MemCmp {
+    fn hook_applies(&self, tcx: TyCtxt<'tcx>, instance: Instance<'tcx>) -> bool {
+        let name = with_no_trimmed_paths!(tcx.def_path_str(instance.def_id()));
+        name == "core::slice::cmp::memcmp"
```
celinval, line 335: Given the outdated discussion, can you also compare it against `"std::slice::cmp::memcmp"`.

fzaiser: Done

celinval (approved):
Thanks

YoshikiTakashima (approved):

RalfJung:
> We should also check that the pointers are aligned, but I'm not sure how to find the alignment of a type at this stage.

memcmp works on u8, so the alignment requirement is trivial.

However this will accept some code that Miri rejects (like the 2nd example [here](https://github.com/model-checking/kani/issues/1489#issuecomment-1212553924)), when the pointers have provenance that makes this a use-after-free or an out-of-bounds access. I don't know if and how Kani tracks provenance, so I am not sure if Kani can check these things.

zhassan-aws:
It might be possible to strengthen the check:

https://github.com/model-checking/kani/blob/d2546b3f239b5de52f255ccc9437df47babeaee0/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs#L342

to exclude pointers that were deallocated (and thus have provenance), e.g.
```rust
let is_first_ok = first_var.clone().is_nonnull() && first_var.clone().is_not_deallocated();
```
Perhaps, [CBMC's memory primitives](http://cprover.diffblue.com/memory-primitives.html) can be used for determining if the pointer was deallocated?

RalfJung:
For completeness' sake, here is another example Miri considers invalid, using an out-of-bounds pointer rather than a deallocated pointer:
```rust
fn make_oob() -> *const u8 { unsafe {
    let b = Box::into_raw(Box::new(0u8));
    (&*b as *const u8).wrapping_offset(42)
} }

fn main() { unsafe {
    // out-of-bounds is UB
    memcmp(make_oob(), make_oob(), 0) ;
} }
```

fzaiser:
> memcmp works on u8, so the alignment requirement is trivial.

@RalfJung Yeah, I realized this too and changed the code, but forgot to update the PR description.

> However this will accept some code that Miri rejects (like the 2nd example https://github.com/model-checking/kani/issues/1489#issuecomment-1212553924), when the pointers have provenance that makes this a use-after-free or an out-of-bounds access. I don't know if and how Kani tracks provenance, so I am not sure if Kani can check these things.

Thanks for the examples, @RalfJung! I checked the [CBMC documentation](http://cprover.diffblue.com/memory-primitives.html): CBMC does track an object ID and offset for each *valid* pointer, which I believe would correspond to provenance. (I'm not sure what happens with pointer-integer casts though.) However, it does not consider dangling pointers to be valid, so it does not follow Rust's stance of there being an implicit zero-sized allocation.

So as I understand it, we would have to convince the CBMC people to change their semantics of validity of pointers to support this in Kani.

fzaiser:
Since this particular issue is mitigated, but we should find a better solution, I opened a new issue for it: #1574

Issue #1533, opened 2022-08-17
Niche optimization encoding is broken


celinval:
I tried this code:

```rust
#[kani::proof]
fn check_i8() {
    let v1: i8 = -10;
    let n1 = NonZeroI8::new(v1);
    assert!(n1.is_some());
}
```

using the following command line invocation:

```
kani test.rs
```

with Kani version:

I expected to see this happen: Harness should succeed.

Instead, this happened: Verification failed.

```
SUMMARY:
 ** 1 of 8 failed
Failed Checks: assertion failed: n1.is_some()
File: "/kani/tests/kani/Invariants/check_nonzero.rs", line 25, in check_i8

VERIFICATION:- FAILED

```

Note that we do have a test that should be checking this (tests/kani/Invariants/check_nonzero.rs), and if you run it manually, you can see that the test is failing. However, the test uses `kani::expect_fail()` which is also broken. So the tests actually succeeds in our CI.

celinval:
It looks like we are failing when the value is negative. I wonder if this regression was introduced by: https://github.com/model-checking/kani/commit/d0b74ca1fa07b9d163981d2bc1c759c53356f031

celinval:
That said, the following harness only fails in the second assert:
```rust
#[kani::proof]
fn check_i8() {
    let v1: i8 = -10;
    let n1 = unsafe { NonZeroI8::new_unchecked(v1) };

    assert_eq!(n1.get(), v1);
    assert!(Some(n1).is_some());
}
```

celinval:
> d0b74ca

It predates this commit.

danielsn:
I figured out what's wrong: we're doing a `signed` comparison against 0, when we need unsigned.

PR #1534, opened 2022-08-17
Remove deprecated function and adapt tests

### LINKED ISSUE #599: Remove deprecated proof harness utility functions
In RMC, users can use special functions to generate a symbolic variable, create assumptions and expect failures. The naming schema that was initially used was C like and it didn't conform to rust conventions (#589). We are now replacing them but we will keep support to the old for now. This issue is to track the work of removing them completely.

celinval:
### Description of changes: 

 - Removed kani::any_raw() and Invariants.
 - Removed handling of old __VERIFIER_* functions.
 - Adapted tests that were using old functions. Note that I moved most tests  from `tests/kani/Invariants` to be under `tests/kani/Arbitrary`

### Resolved issues:

Resolves #599
Is related to #1394 

### Call-outs:

I found an issue while updating the tests that is orthogonal to this change. I created #1533 and moved the `nonzero.rs` test to be a fix-me test.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


tedinski (approved):
I do wonder if we should be "nicer" and leave things in place with the deprecations, but I suspect there are not a lot of users of this, so we can get away with it this time.

celinval:
> I do wonder if we should be "nicer" and leave things in place with the deprecations, but I suspect there are not a lot of users of this, so we can get away with it this time.

I think we should at least create a proper deprecation plan. For example, we can say that we remove deprecated APIs only after at least 3 releases after the deprecation or something like that. Thoughts?

PR #1537, opened 2022-08-17
docs: add conc playback docs to getting started tutorial

### LINKED ISSUE #1439: Write concrete playback tutorial page

sanjit-bhat:
### Description of changes: 

See the title. This PR only adds doc sections.

Note that there are other sections where we could potentially add info about this feature (e.g., 5 sections name the `--visualize` feature). However, as per @tedinski's suggestion, we should save those changes for later once we figure out more details about how the overall tutorial flow would be impacted.

### Callouts:

Some of the names used in this PR rely on the renames in #1548.

### Resolved issues:

Resolves #1439, adding a doc section. 


docs/src/SUMMARY.md
```diff
@@ -5,6 +5,7 @@
     - [Building from source](./build-from-source.md)
   - [Using Kani](./usage.md)
   - [Verification results](./verification-results.md)
+  - [Debugging harnesses](./concrete-playback.md)
```
celinval, line 8: Just a suggestion: Debugging verification result?

docs/src/concrete-playback.md
```diff
@@ -0,0 +1,59 @@
...
+To do this, run `git clone https://github.com/model-checking/kani.git`.
+Then, the user needs to add the following lines to their `Cargo.toml` file:
+```toml
+[dev-dependencies]
+kani = { path = "{path_to_kani}/library/kani", features = ["exe_trace"] }
+```
+
+## Usage
+
+Run Kani with the `--gen-exe-trace` flag.
```
celinval: I think we should rename these flags before pushing this doc. You can either make this change in this PR or in a separate PR that shall be merged before this one.

docs/src/concrete-playback.md
```diff
@@ -0,0 +1,59 @@
+# Concretely playing back proof harnesses
```
celinval, line 1: I expect the title here to match the section name: "Debugging harness"

docs/src/concrete-playback.md
```diff
@@ -0,0 +1,59 @@
...
+
+When the result of a certain check comes back as a `FAILURE`,
+this _experimental_ feature allows users to debug their proof harness and understand what
```
celinval, line 4: you should probably say that Kani offers two different paths today.
 - Visualize: Kani provide an HTML text based trace that enumerate the execution steps that led to the property failure.
 - Concrete playback: This is a new _experimental_ feature that generates a unit test that when executed will trigger the failure seen during the verification.

docs/src/concrete-playback.md
```diff
@@ -0,0 +1,59 @@
...
+caused the failing check.
+
```
celinval, line 6: Now add:
```md
## Concrete playback
### Setup
`

File: docs/src/concrete-playback.md [hunk 6]
```diff
@@ -0,0 +1,59 @@
...
+## Setup
+
+The user needs to have a fairly recent version of the Kani source code somewhere on their computer.
```
celinval, line 9: Be specific. This will be added on version 0.9.

docs/src/concrete-playback.md
```diff
@@ -0,0 +1,59 @@
...
+To do this, run `git clone https://github.com/model-checking/kani.git`.
```
celinval, line 10: Maybe say, this feature require users to declare dependency on the kani crate. BTW, users don't need to do manually clone the repository. Cargo does that for them.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

docs/src/concrete-playback.md
celinval: In order to enable this feature, run Kani with `--enable-unstable --gen-exe-trace`.

docs/src/concrete-playback.md
```diff
@@ -0,0 +1,59 @@
...
+Then, the user needs to add the following lines to their `Cargo.toml` file:
+```toml
+[dev-dependencies]
+kani = { path = "{path_to_kani}/library/kani", features = ["exe_trace"] }
+```
+
+## Usage
+
+Run Kani with the `--gen-exe-trace` flag.
+After verifying the proof harness checks, Kani will extract concrete values for the `kani::any()` variables
```
celinval, line 20: Maybe describe what the user will see and not what the tool is doing under the hood. Say something like if there is a failure, Kani will print a test case to the console that includes concrete values that will trigger the failure concretely.

I would suggest that you add an example here of a failure and how the test will look like.

docs/src/concrete-playback.md
```diff
@@ -0,0 +1,59 @@
...
+and generate a Rust unit test case.
+This unit test initializes the concrete values and calls the proof harness.
+A user can then either 1) copy this unit test into the same module as their proof harness or
+2) run Kani with the `--add-exe-trace-to-src` flag to have Kani automatically do this.
```
celinval: After you talk about the testcase structure, I would introduce the new flag. Something like:

> Kani also offer an option to add the test directly to your source code. For that, invoke kani using the `--in-place` flag....

celinval (commented):
Great job. I added a few comments. Let me know what you think.

docs/src/SUMMARY.md
sanjit-bhat: done

docs/src/concrete-playback.md
sanjit-bhat: done

docs/src/concrete-playback.md
sanjit-bhat: done

docs/src/usage.md
```diff
@@ -23,6 +23,8 @@ This works like `cargo test` except that it will analyze all proof harnesses ins
 
 Common to both `kani` and `cargo kani` are many command-line flags:
 
+ * `--concrete-playback=JustPrint`: Generates a Rust unit test case that plays back a failing proof harness using a concrete counterexample. The additional `--concrete-playback=InPlace` flag will automatically add this unit test to the source code next to the proof harness.
```
celinval, line 26: `--concrete-playback=[print | inplace]`: Generates a Rust unit test that represents a concrete counter-example found by Kani. Use `print` argument to print the test in the terminal or `inplace` to automatically add this unit test to the source code next to the target proof harness.

docs/src/concrete-playback.md
sanjit-bhat: Done

docs/src/concrete-playback.md
sanjit-bhat: done

docs/src/concrete-playback.md
sanjit-bhat: Done

docs/src/concrete-playback.md
sanjit-bhat: done

docs/src/concrete-playback.md
sanjit-bhat: Done

docs/src/concrete-playback.md
sanjit-bhat: done

docs/src/usage.md
sanjit-bhat: done

sanjit-bhat:
@celinval I added an example, but I decided to only explain a couple parts of the unit test that I thought would be most relevant to the user. Here, I define relevance as "info that's helpful if the user wants to try a different concrete value without regenerating the unit test".

docs/src/debugging-verification-failures.md
```diff
@@ -0,0 +1,105 @@
...
+```
+Here, `133` and `35207` are the concrete values that, when substituted for `a` and `b`,
+cause an assertion failure.
+`vec![135, 137]` is the byte array representation of `35207`.
+
+### Common issues
+
+* `error[E0425]: cannot find function x in this scope`:
+this is usually caused by having `#[cfg(kani)]` somewhere in the control flow path of the user's proof harness.
+To fix this, remove `#[cfg(kani)]` from those paths.
```
celinval, line 86: I think it's better to recommend adding `--cfg=kani` to your rustflags.

docs/src/debugging-verification-failures.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Request for comments
+
+This feature is experimental and is therefore subject to change.
+If you have ideas for improving the user experience of this feature,
+please add them to [this github issue](https://github.com/model-checking/kani/issues/1536).
+
+### Limitations
```
celinval, line 94: Can you link the existing issues here?

docs/src/debugging-verification-failures.md
```diff
@@ -0,0 +1,105 @@
...
+This section describes the concrete playback feature in more detail.
+
+### Setup
+
+The Kani library needs to be linked as a dev dependency to the crate you're trying to debug.
+This requires adding the following lines to your `Cargo.toml` file,
+which differ depending on what version of the Kani library you would like to use:
+* The latest version:
+```toml
+[dev-dependencies]
```
celinval, line 21: Maybe change this to [target.'cfg(Kani)'.dependencies]

So this only gets pulled when user sets `--cfg kani`.

celinval (approved):
That's great! Minor comments but not blockers.

docs/src/concrete-playback.md
sanjit-bhat: Done

docs/src/debugging-verification-failures.md
sanjit-bhat: I'm unclear how to do this? I tried adding this to `Cargo.toml`, but it didn't work:
```toml
[build]
rustflags = ["--cfg=kani"]
```

docs/src/debugging-verification-failures.md
sanjit-bhat: Which existing issue (apart from the one I already linked)?

sanjit-bhat: ah, I see. I created a tracking milestone.

docs/src/debugging-verification-failures.md
celinval: There are ways around that, but we can tune that later. Thanks

PR #1539, opened 2022-08-17
feat: make `kani::any::<bool>()` non-det bit pattern more intuitive

### LINKED ISSUE #1490: Change `kani::any::<bool>()` to make the non-det bit pattern more intuitive
## TLDR

This issue proposes changing our implementation of `kani::any::<bool>()` from:

https://github.com/model-checking/kani/blob/de71c097f266053d6b10c1c4d762b47766a210d9/library/kani/src/arbitrary.rs#L48-L54

to something like

```rust
impl Arbitrary for bool {
    #[inline(always)]
    fn any() -> Self {
        let byte = u8::any();
        crate::assume(byte < 2);
        byte == 1
    }
}
```

## Reasoning

Our current implementation of `kani::any::<bool>()` has the correct API semantics of a `bool`. I.e., that a `bool` can be either `true/false`. However, it doesn't map to the [underlying bit pattern of Rust bools](https://doc.rust-lang.org/reference/types/boolean.html):

> An object with the boolean type has a [size and alignment](https://doc.rust-lang.org/reference/type-layout.html#size-and-alignment) of 1 each. The value false has the bit pattern 0x00 and the value true has the bit pattern 0x01. It is [undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) for an object with the boolean type to have any other bit pattern.

The Rust spec has the following differences:
1. A `true` value is represented by `0x01`, not `0x00`.
2. A `false` value is represented by only `0x00`, not `[0x01, 0xFF]`.

This difference has an impact when playing back deterministic test cases. In our tests for playing back `bool`'s, we can never set an expected deterministic bit pattern for `false`. There are multiple possible bit patterns, and we don't know which one the solver will choose.

## Callouts

The proposed change would add an extra `kani::assume()` for every `bool` generated. This might slow down verification time. We should performance benchmark this change before implementing it.

sanjit-bhat:
### Description of changes: 

See #1490 for a description of the changes and their use case. The bottom-most comment in the thread also has a performance evaluation. Based on that evaluation, I don't think it adds that much overhead, and I recommend that we implement it given that it will make the concrete value more clear when using the `--gen-exe-trace` and `--visualize` flags.

### Resolved issues:

Resolves #1490.

### Testing:

* How is this change tested?

This change clearly still results in `bool::any()` returning a `bool`. Also, I modified the concrete playback tests given that they now have reliable concrete values for `bool`.

### Checklist

### DIFF
--- a/library/kani/src/arbitrary.rs
+++ b/library/kani/src/arbitrary.rs
@@ -49,7 +49,8 @@ impl Arbitrary for bool {
     #[inline(always)]
     fn any() -> Self {
         let byte = u8::any();
-        byte == 0
+        crate::assume(byte < 2);
+        byte == 1
     }
 }
 
--- a/tests/ui/exe-trace/bool/expected
+++ b/tests/ui/exe-trace/bool/expected
@@ -6,7 +6,9 @@ Executable trace
 fn kani_exe_trace_harness
     let det_vals: Vec<Vec<u8>> = vec![
         // 0
-        vec![0
+        vec![0],
+        // 1
+        vec![1]
     ];
     kani::exe_trace_run(det_vals, harness);
 }
--- a/tests/ui/exe-trace/bool/main.rs
+++ b/tests/ui/exe-trace/bool/main.rs
@@ -3,9 +3,9 @@
 
 // kani-flags: --harness harness --enable-unstable --gen-exe-trace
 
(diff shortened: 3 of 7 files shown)


celinval (approved):

zhassan-aws (approved):
Looks good. Thanks!

Issue #1540, opened 2022-08-17
Cleanup TagEncoding::Niche


danielsn:
Its kinda confusing right now

PR #1541, opened 2022-08-17
Fix issue with niches on signed values

### LINKED ISSUE #1533: Niche optimization encoding is broken

danielsn:
### Description of changes: 

Currently, Kani fails to correctly handle niche optimizations when the underlying data-type is signed.
Use unsigned comparisons (as the rustc backend does) to fix this issue.
### Resolved issues:

Resolves #1533 

### Call-outs:

The niche code is confusing, filed #1540 to track

### Testing:

* How is this change tested? Additional test.

### Checklist


celinval (approved):
Awesome! Thanks for doing this.

I just merged #1534 that adds `nonzero-fixme.rs`. When you merge, you will probably need to remove that. Thanks again!

PR #1543, opened 2022-08-18
An RFC for creating an RFC process. :)

celinval:
### Description of changes: 

I am creating this PR to gather feedback on the following:

1. Should we start an RFC process? If we do:
2. How would that process look like?
3. What do we expect to be included in an RFC?

I would like to suggest that we do start a light-weight RFC process. In this PR, I included a suggestion of a process and a suggestion of a template. Finally, this PR also creates a book of RFCs under `kani/rfcs`.

My reasoning here is that the RFC process will:
  - Improve collaboration by requesting comments at an earlier stage of development.
  - Provide better user experience  by writing the RFCs starting from the user impact.
  - improve the quality of our software by thinking in terms of architecture.
  - improve our documentation

The main drawback is the time overhead to write the RFC and iterating over the feedback. So we should aim to make this a lightweight process. The RFC doesn't have to answer all questions and it can be updated as we go.

### Resolved issues:

N/A

### Call-outs:

I am suggesting that RFCs require at least 2 approvals, but I haven't looked into how we could enforce it yet.

### Testing:

* How is this change tested?  N/A

### Checklist


rfcs/src/intro.md
```diff
@@ -0,0 +1,53 @@
+# Introduction
+
+Kani is an open-source verification tool that uses automated reasoning to analyze Rust programs. In order to 
+integate feedback from developers and users on future changes to Kani, we decided to follow a light-weight 
+"RFC" (request for comments) process.
+
+## When to create an RFC
+
+As a rule of thumb you should create an RFC if you are making changes that impact the user experience. New 
+APIs, new features, as well as deprecation, should follow an RFC process. We also ask for substantial architectural
```
jaisnan, line 10:
```suggestion
APIs, new features, as well as depreciation, should follow an RFC process. We also ask for substantial architectural 
```

jaisnan:
Thank you! I have a few questions as someone unfamiliar with the RFC process - 
1. Is the RFC process supposed to formalize and replace the internal document writing process? If not, when do we decide to make it an RFC and when do we make it a design document?
2. Do all RFC's need to be public facing or can the person proposing the RFC choose to make it not available to the public?
3. What is the reason behind closing the PR for an RFC once it's not approved? Is the person expected to come up with a new RFC with the feedback given?

rfcs/src/intro.md
```diff
@@ -0,0 +1,53 @@
+# Introduction
+
+Kani is an open-source verification tool that uses automated reasoning to analyze Rust programs. In order to 
+integrate feedback from developers and users on future changes to Kani, we decided to follow a light-weight 
+"RFC" (request for comments) process.
+
+## When to create an RFC
```
celinval, line 7: TBH, I think we need to improve this criteria. I think it's too open right now.

rfcs/src/template.md
```diff
@@ -0,0 +1,58 @@
+- Feature Name: *Fill me with pretty name and a unique ident. E.g: New Feature (`new_feature`)*
+- Feature Request Issue: *Link to issue*
+- RFC PR: *Link to original PR*
+- Status: *One of the following: [Under Review | Unstable | Stable | Cancelled]*
```
celinval, line 4: Per @danielsn suggestion, we should add an RFC version here and an optional link to a proof-of-concept.

I believe the RFC version should be bumped with subsequent PRs that update the RFC not during the PR feedback.

danielsn:
> Thank you! I have a few questions as someone unfamiliar with the RFC process -
> 
> 1. Is the RFC process supposed to formalize and replace the internal document writing process? If not, when do we decide to make it an RFC and when do we make it a design document?

Yes, the RFC mechanism will be the Kani process going forward.

> 2. Do all RFC's need to be public facing or can the person proposing the RFC choose to make it not available to the public?

 Yes. We expect every RFC to be public since the changes would be made public too.  For changes that are still under development, it might make sense to open a "Draft" RFC

> 3. What is the reason behind closing the PR for an RFC once it's not approved? Is the person expected to come up with a new RFC with the feedback given?

If an RFC is not accepted, the next steps are up to the author (much like when a paper is not accepted by a peer review process).  Depending on their judgement, they can abandon the RFC, start a new RFC with a different approach, or revise their current approach to address the reviewer feedback.

rfcs/src/intro.md
```diff
@@ -0,0 +1,53 @@
...
+
+As a rule of thumb you should create an RFC if you are making changes that impact the user experience. New 
+APIs, new features, as well as depreciation, should follow an RFC process. We also ask for substantial architectural 
+changes to be made via RFC.
+
+Bugs and smaller improvements to existing features do not require an RFC. If you are in doubt, feel free to create 
+a [feature request](https://github.com/model-checking/kani/issues/new?assignees=&labels=&template=feature_request.md)
+and discuss the next steps in the new issue.
```
danielsn, line 15:
```suggestion
You should create an RFC in one of two cases:

1. The change you are proposing would be a "one way door": e.g. a change to the public API, a new feature that would be difficult to modify once released, deprecating a feature, etc.
2. The change you are making has a significant design component, and would benefit from a design review.

Bugs and smaller improvements to existing features do not require an RFC.
If you are in doubt, feel free to create  a [feature request](https://github.com/model-checking/kani/issues/new?assignees=&labels=&template=feature_request.md) and discuss the next steps in the new issue.
Your PR reviewer may also request an RFC if your change appears to fall into category 1 or 2.

You do not necessarily need to create an RFC immediately. It is our experience that it is often best to write some "proof of concept" code to test out possible ideas before writing the formal RFC.```

File: rfcs/src/intro.md [hunk 5]
```diff
@@ -0,0 +1,53 @@
...
+
+## The RFC process
+
+This is the overall workflow for the RFC process:
+
+```toml
+    Create RFC ──> Receive Feedback  ──> Accepted?
```
danielsn, line 22: You might want to note the `recieve feedback`/`revise` loop.

Also, the `Implement` -> `Revise RFC` feedback edge

rfcs/src/intro.md
```diff
@@ -0,0 +1,53 @@
...
+                                            ├───> Implement ───> Stabilize?
+                                            │ N                      │ Y
+                                            └───> Close PR           ├───> RFC Stable
+                                                                     │ N
+                                                                     └───> Remove feature
+```
+
+1. Create an RFC
+   1. Start from a fork of the Kani repository.
+   2. Copy the template file ([`rfcs/src/template.md`](./template.md)) to `rfcs/src/<my-feature>.md`.
```
danielsn, line 33: I like having numbered RFCs.  Perhaps say copy to `rfcs/src/<ID_NUMBER><my-feature>.md`

rfcs/src/intro.md
```diff
@@ -0,0 +1,53 @@
...
+   3. Fill in the details according to the template intructions.
+   4. Submit a pull request.
+2. Build consensus and integrate feedback.
+   1. RFCs should get approved by at least 2 members of the `kani-developers` team.
+   2. Once the RFC has been approved, update the RFC status and merge the PR.
+   3. If the RFC is not approved, close the PR without merging.
```
danielsn, line 39: If the RFC is not approved, replace status of the RFC with a template saying "Closed as per PR blah"

rfcs/src/intro.md
```diff
@@ -0,0 +1,53 @@
...
+   1. Start implementing the new feature in your fork.
+   2. It is OK to implement it incrementally over multiple PRs. Just ensure that every pull request has a testable 
+      end-to-end flow and that it is properly tested.
+   3. In the implementation stage, the feature should only be accessible if the user explicitly passes 
+      `--enable-unstable` as an argument to Kani.
+   4. Document how to use the feature.
+4. Stabilization.
+   1. After the feature has been implemented, start the stabilization process.
+   2. Gather user feedback and make necessary adjustments.
+   3. Create a new PR that removes the `--enable-unstable` guard and that marks the RFC status as "STABLE". Also
```
danielsn, line 50: Call this out. This is super important to stabilize the RFC along with the issue.

danielsn (commented):
We should add the RFC to the PR template

rfcs/src/intro.md
```diff
@@ -0,0 +1,53 @@
+# Introduction
+
+Kani is an open-source verification tool that uses automated reasoning to analyze Rust programs. In order to 
+integrate feedback from developers and users on future changes to Kani, we decided to follow a light-weight 
+"RFC" (request for comments) process.
+
+## When to create an RFC
+
+As a rule of thumb you should create an RFC if you are making changes that impact the user experience. New 
+APIs, new features, as well as depreciation, should follow an RFC process. We also ask for substantial architectural
```
tedinski: IMO, we should be explicit about a lighter-weight process option here. Changes are usually going to be:

1. Simple enough to not need an RFC
2. Complex enough to merit discussion and get consensus on design choices, but perhaps as a github issue rather than an RFC proper.
3. Important enough to merit an RFC-like process, especially when it comes to soliciting customer feedback.

celinval: I would prefer start with 2 modes only. For the case where it merit discussion and get consensus, I would prefer using an RFC PR instead of an issue. It's OK if the RFC document is fairly small.

We can adjust as we go.

rfcs/src/intro.md
```diff
@@ -0,0 +1,56 @@
...
+You should create an RFC in one of two cases:
+
+1. The change you are proposing would be a "one way door": e.g. a change to the public API, a new feature that would be difficult to modify once released, deprecating a feature, etc.
+2. The change you are making has a significant design component, and would benefit from a design review.
+
+Bugs and smaller improvements to existing features do not require an RFC.
+If you are in doubt, feel free to create  a [feature request](https://github.com/model-checking/kani/issues/new?assignees=&labels=&template=feature_request.md) and discuss the next steps in the new issue.
+Your PR reviewer may also request an RFC if your change appears to fall into category 1 or 2.
+
+You do not necessarily need to create an RFC immediately. It is our experience that it is often best to write some "proof of concept" code to test out possible ideas before writing the formal RFC.```
```
adpaco-aws, line 18: Remove back-ticks at the end of the sentence.

rfcs/src/intro.md
```diff
@@ -0,0 +1,56 @@
...
+                                                                     └───> Remove feature
+```
+
+1. Create an RFC
+   1. Start from a fork of the Kani repository.
+   2. Copy the template file ([`rfcs/src/template.md`](./template.md)) to `rfcs/src/<my-feature>.md`.
+   3. Fill in the details according to the template intructions.
+   4. Submit a pull request.
+2. Build consensus and integrate feedback.
+   1. RFCs should get approved by at least 2 members of the `kani-developers` team.
```
adpaco-aws, line 40: I think it's `kani-devs`.

rfcs/src/template.md
```diff
@@ -0,0 +1,58 @@
...
+# Detailed Design
+[detailed-design]: #detailed-design
+
+This is the technical portion of the RFC. Please provide high level details of the implementation you have in mind:
+
+- What are the main components that will be modified? (E.g.: changes to kani-compiler, kani-driver, metadata, 
+  installation...)
+- How will they be modified? Any changes to how these components communicate?
+- Will this require any new dependency?
+- What are the corner cases you antecipate?
```
adpaco-aws, line 39:
```suggestion
- What corner cases do you anticipate?
```

rfcs/src/template.md
```diff
@@ -0,0 +1,58 @@
...
+
+- How would you teach this feature to users? What changes will be required to the user documentation?
+- If the RFC is related to architectural changes and there are no visible changes to UX, please state so.
+
+# Detailed Design
+[detailed-design]: #detailed-design
+
+This is the technical portion of the RFC. Please provide high level details of the implementation you have in mind:
+
+- What are the main components that will be modified? (E.g.: changes to kani-compiler, kani-driver, metadata,
```
adpaco-aws, line 35: Please use back-ticks when referring to Kani components.

rfcs/src/template.md
```diff
@@ -0,0 +1,58 @@
...
+
+This is the technical portion of the RFC. Please provide high level details of the implementation you have in mind:
+
+- What are the main components that will be modified? (E.g.: changes to kani-compiler, kani-driver, metadata, 
+  installation...)
+- How will they be modified? Any changes to how these components communicate?
+- Will this require any new dependency?
+- What are the corner cases you antecipate?
+
+# Rationale and alternatives
```
adpaco-aws, line 41: This could just be part of "Detailed design".

rfcs/src/template.md
```diff
@@ -0,0 +1,58 @@
...
+- What other designs have you considered? Why didn't you choose them?
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Is there any part of the design that you expect to resolve through the RFC process?
+- What kind of user feedback do you expect to gather before stabilization? How will this impact your design?
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
```
adpaco-aws, line 55: What is this notation for?

scripts/build-docs.sh
```diff
@@ -2,6 +2,10 @@
 # Copyright Kani Contributors
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
+# Build all our documentation and place them under book/ directory.
+# The user facing doc is built into book/
+# Our RFCs is placed under book/rfcs/
```
adpaco-aws, line 7:
```suggestion
# RFCs are placed under book/rfcs/
```

rfcs/src/intro.md
```diff
@@ -0,0 +1,56 @@
...
+3. Feature implementation.
+   1. Start implementing the new feature in your fork.
+   2. It is OK to implement it incrementally over multiple PRs. Just ensure that every pull request has a testable 
+      end-to-end flow and that it is properly tested.
+   3. In the implementation stage, the feature should only be accessible if the user explicitly passes 
+      `--enable-unstable` as an argument to Kani.
+   4. Document how to use the feature.
+4. Stabilization.
+   1. After the feature has been implemented, start the stabilization process.
+   2. Gather user feedback and make necessary adjustments.
```
adpaco-aws, line 52: We should define some criteria to stabilize a feature. This reads like "Stabilize right away after implementing it", when in practice we will wait for a feature to be stabilized after a period. Recent bug reports or short periods after implementation should all delay the stabilization of a feature.

rfcs/src/template.md
celinval: Hummm... That's an interesting point. I can see that in some cases that would make sense.

The reason I believe this is not inside "Detailed Design" because the "Rationale and alternatives" may go beyond the implementation details. In some cases, the alternative could be providing a completely different user experience to achieve the same purpose. Like [this rust RFC](https://rust-lang.github.io/rfcs/1861-extern-types.html).

Does it make sense?

rfcs/src/template.md
celinval: I should remove these. I believe they have to do with the link to this section.

rfcs/src/intro.md
celinval: Yeah, I totally agree. Any suggestions?

I would prefer keeping it lightweight for now. Something like creating a PR that marks the RFC as stable and remove the `--enable-unstable` requirement. In the description of the PR, the developer states why they believe the feature can be marked as stable and we go from there. I guess we could add a minimum "bake" time like 2 releases.

 I think for the rust project, they have two different process which is more heavy weight. This includes writing a report and starting an FCP (final comment period).

rfcs/src/intro.md
celinval: I added the first. The second I added a note to the item below (I'm worried that the diagram is getting too complicated). Is that OK?

rfcs/src/intro.md
celinval: I added the ID_NUMBER. Any suggestion on how this will be assigned? We can either try to increment based on existing RFCs and open RFCs or we can use the tracking issue number.

rfcs/src/intro.md
celinval: I actually removed the group name here. I still need to figure out a way to enforce this without changing all our PRs.

rfcs/src/intro.md
celinval: What if we add the reason in the tracking issue itself? Would that be OK?

rfcs/src/intro.md
celinval: I added a stage between "Implementation" and "Stabilization". Let me know what you think.

celinval:
@danielsn I added the "proof-of-concept" field, but I'm having second thoughts on this one. Developers will likely link this against some branch in their fork and this branch will likely be deleted. I think it would make more sense to add that link to the PR description itself. What do you think?

rfcs/src/intro.md
adpaco-aws: Thanks for adding another stage. I think that makes clear there is not an immediate path to stabilization, while at the same time keeping the process lightweight as we want.

rfcs/src/template.md
adpaco-aws: Seems like the last one is still there?

adpaco-aws (approved):
Thanks for doing this!

tedinski (approved):
Ship it, let's try it and see how it goes.

One last question: this creates the `/rfcs/` directory (should it be just `/rfc/` perhaps?) but nothing links to it. Should we add a link from our regular docs to the "rfc book" ?

celinval:
> Ship it, let's try it and see how it goes.
> 
> One last question: this creates the `/rfcs/` directory (should it be just `/rfc/` perhaps?) but nothing links to it. Should we add a link from our regular docs to the "rfc book" ?

Great question! I am happy to rename it to `/rfc/` or `/rfc-book/`. Any preference?

I like the idea of adding a link to it from our regular docs. Initially I wanted to make sure they would play nice to each other, but I was able to test that in my fork, and I don't see why not. Should I just add a link in the Summary table?

tedinski:
I like `/rfc/`.

A link from the summary works for me.

celinval:
Unfortunately, a link from the summary won't work. `mdbook` doesn't support external links inside `SUMMARY.md` (https://github.com/rust-lang/mdBook/issues/919). There is a very hacky way to do it as described in https://github.com/rust-lang/mdBook/issues/1354, but I don't think it's worth it.

I'll add a link under [Project Status](https://model-checking.github.io/kani/getting-started.html#project-status) for now. Let me know if you have any other suggestion.

PR #1555, opened 2022-08-18
Bugfix: Message in assert statement needs to wrapped in a Struct in function contract hooks. 

monal:
### Description of changes: 

Resolves bugfix introduced in [#1511](https://github.com/model-checking/kani/pull/1511). 

The following change was introduced in the earlier PR. However, the `Expr::string_constant` needs to be wrapped in a `Struct`.  Else the `extract_const_message` function returns a None while handling the assert statement and the compiler panics.

```
fn handle (/*args*/) {
    if tcx.queries.get_enforce_contracts() {
      fargs.add(Expr::string_constant("Post-condition failed");
      Assert::handle(tcx, instance, fargs, assign_to, target, span);
    } else {
      Assume::handle(tcx, instance, fargs, assign_to, target, span);
    }
  }
```

### Resolved issues:

Resolves bug introduced by #1511

### Testing:

* How is this change tested?
Locally tested. 

* Is this a refactor change?
No,

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
@@ -12,7 +12,7 @@ use crate::codegen_cprover_gotoc::codegen::PropertyClass;
 use crate::codegen_cprover_gotoc::utils;
 use crate::codegen_cprover_gotoc::GotocCtx;
 use crate::unwrap_or_return_codegen_unimplemented_stmt;
-use cbmc::goto_program::{BuiltinFn, Expr, Location, Stmt, Type};
+use cbmc::goto_program::{BuiltinFn, DatatypeComponent, Expr, Location, Stmt, Type};
 use kani_queries::UserInput;
 use rustc_middle::mir::{BasicBlock, Place};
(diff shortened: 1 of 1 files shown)


kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -266,7 +266,11 @@ impl<'tcx> GotocHook<'tcx> for Postcondition {
         let loc = tcx.codegen_span_option(span);
 
         if tcx.queries.get_enforce_contracts() {
-            fargs.push(Expr::string_constant("Post-condition failed"));
+            let msg = Expr::string_constant("Post-condition failed");
```
zhassan-aws, line 269: Can we wrap this in a helper function?

zhassan-aws (commented):
Can we

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
monal: Done.

zhassan-aws (approved):

PR #1559, opened 2022-08-19
Implement `-p` flag for running kani on specific package

### LINKED ISSUE #1557: Support `-p` flag in Cargo Kani
Requested feature: Enable `-p` with Cargo Kani

Use case: There are a lot of customer code where `--workspace` is not
appropriate. In particular, `--workspace` will pull in dependencies
that trigger ICE, preventing us from using Kani in that crate.

Link to relevant documentation (Rust reference, Nomicon, RFC): https://doc.rust-lang.org/cargo/commands/cargo-build.html#options

Test case: https://github.com/tokio-rs/prost

YoshikiTakashima:
### Description of changes: 

Adds `-p` / `--package` flag for running kani on a specific package within a workspace

### Resolved issues:

Resolves #1557

### Call-outs:

The test cases involve some duplicate code from the workspace commit.

### Testing:
* How is this change tested? tests/cargo-kani/ws-specified/

* Is this a refactor change?

### Checklist


kani-driver/src/args.rs
```diff
@@ -121,6 +121,9 @@ pub struct KaniArgs {
     /// Run Kani on all packages in the workspace.
     #[structopt(long)]
     pub workspace: bool,
+    /// Run Kani on the specified package.
+    #[structopt(long, short)]
+    pub package: Option<String>,
```
zhassan-aws, line 126: I think this should be `Vec<String>` to allow specifying the flag multiple times (as `cargo build` does).

tests/cargo-kani/ws-specified/bincrate/src/main.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This test's purpose is to ensure that we successfully build
```
zhassan-aws, line 4: This comment is from the other test, and does not reflect what the test is checking. Can you please update it?

tests/cargo-kani/ws-specified/expected
```diff
@@ -0,0 +1 @@
+ ** 1 of 1 failed
```
zhassan-aws, line 1: With this expected file, it seems to me that this test would pass even if Kani included the harness in the `bincrate` crate as well? If I understood correctly, the purpose of this test is to check that the `--package` option is processed correctly, and that Kani only runs the specified package. If my understanding is correct, can we check that only one harness ran, by checking the multi-harness summary line that Kani emits, e.g.
```
```

tests/cargo-kani/ws-specified/bincrate/src/main.rs
YoshikiTakashima: Fixed with 64225f0ecb2ffce2c4176c307d9905abb0f16fc7

kani-driver/src/args.rs
YoshikiTakashima: 422f2978c2e86343876997535067b39e5c824f01

tests/cargo-kani/ws-specified/expected
YoshikiTakashima: 96a256db64ff6bfdc7ae72bd1adc0af773ed86a7

tests/cargo-kani/ws-specified/subcrate3/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! One of 3 sub packages used to test specifying packages with -p
+//! flag.
+
+#[kani::proof]
+fn check_libcrate_proof() {
```
zhassan-aws, line 8: Should this be `check_subcrate3_proof`? Also, do you want to make this test failing to make sure the `expected` test excludes it?

zhassan-aws (approved):
Thanks!

PR #1580, opened 2022-08-23
Remove the need to build Kani with --workspace

tedinski:
### Description of changes: 

1. This depends on a feature?/bug fix? that will land in stable Rust in 1.64 (released in a month on Sept 22). It works for us now because we're pinned against a nightly Rust that now includes the patch.
2. This might slightly reduce the over all "load" in CI, but I doubt it will decrease CI times by very much. The current critical path is macos-11 regression, and I believe that there we only save the build time of "bookrunner" and "make-kani-release" and nothing else. Hmm, maybe saving a "rebuild" of `library/kani` as well maybe.

### Testing:

* Is this a refactor change? sorta

### Checklist

### DIFF
--- a/docs/src/cheat-sheets.md
+++ b/docs/src/cheat-sheets.md
@@ -9,11 +9,6 @@ development purposes.
 ### Build
 
 ```bash
-# Build all packages in the repository
-cargo build --workspace
-```
-
-```bash
 # Error "'rustc' panicked at 'failed to lookup `SourceFile` in new context'"
 # or similar error?
 # Clean `kani-compiler` and re-build:
--- a/.github/workflows/kani.yml
+++ b/.github/workflows/kani.yml
@@ -28,7 +28,7 @@ jobs:
             os: ${{ matrix.os }}
 
       - name: Build Kani
-        run: cargo build --workspace
+        run: cargo build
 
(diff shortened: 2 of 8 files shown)


docs/src/cheat-sheets.md
```diff
@@ -8,11 +8,6 @@ development purposes.
 
 ### Build
 
-```bash
-# Build all packages in the repository
-cargo build --workspace
```
zhassan-aws, line 13: Should this be kept with `--workspace` removed?

YoshikiTakashima:
If that feature lands, shall we also deprecate `--workspace` from `cargo kani`?

zhassan-aws:
> If that feature lands, shall we also deprecate --workspace from cargo kani?

I don't think we should, because users may still want to compile their entire workspace (and not just the default members) with Kani. Note that this PR is only concerned with how Kani itself is built, and not how it is used to compile packages.

docs/src/cheat-sheets.md
tedinski: I'm open to other ideas but IMO it was only ever in there in the first place to note that `--workspace` was needed.

zhassan-aws (approved):

Issue #1587, opened 2022-08-25
Release Kani with GH packages.


YoshikiTakashima:
Requested feature: Release Kani Docker images with github packages.
Use case: Smoother integration to customer CI.
Link to relevant documentation (Rust reference, Nomicon, RFC):
- https://docs.github.com/en/actions/publishing-packages/publishing-docker-images

Test case:

``` bash
docker pull  ghcr.io/model-checking/kani:latest
```

@tedinski: moving the discussion here. I think this would be the best
way to release, but open to other plans. I can own this if this plan
sounds good to you.

YoshikiTakashima:
Potential customer running proptest (https://github.com/tokio-rs/prost.git) suggested this will make integration easier for them.

Adding under "Integration with existing testing frameworks (e.g. Proptest, Bolero)"

PR #1589, opened 2022-08-26
Implement GH action script.

### LINKED ISSUE #1312: Create Kani Github Action
It might make things easier for customers if they can simply make use of a github action to run Kani on their code. Ideally this would be as simple as (for example) how we publish our ghpages:

```
- name: Run proofs
  uses: model-checking/kani-github-action@v1
```

YoshikiTakashima:
### Description of changes: 

Adds a github actions script that runs Kani on user code

### Resolved issues:

Resolves #1312

### Call-outs:

-  ~~Blocked on #1587 This action script relies on docker to work. Kani needs to publish Docker images as packages for this action to work. PR #1595!~~

- Testing is a bit odd on this one. We expect our customers to integrate via marketplace, but we can't test that integration until this script is merged. See testing sections for details
- Integration method is through docker, which will require a package release. I think this is cleanest since it requires no build on customer end. Risk of breakage is less, we don't slow their CI pipeline with our install and build.
-  Kani options is intended to be passed in via `Cargo.toml` section `[package.metadata.kani.flags]`. Also configuring this action is only possible through `Cargo.toml` I think maintaining 2 APIs will be a mess and YAML is not very good for wrangling certain parameters (e.g. `cargo kani -p packageA -p packageb`). It would be much easier to maintain if we only allow configuration trhough `Cargol.toml`.

### Testing:

* How is this change tested? You can test this by pasting the action.yml into a cargo project that already uses Kani, then running the kani harnesses with it. Add a workflow referencing this action, and trigger it. Results should match that of running `cargo kani` on that project.

* Is this a refactor change? NO

### Checklist
- [x] Methods or procedures are documented. ~~TBD Adding docs in the Kani book.~~ 3d5b2afdb736d881b5baa583fb003d2b4a107d7c


YoshikiTakashima:
@tedinski Would you mind reviewing this. It's still blocked, but demand (as far as I can tell from Rust users around me) is very strong on this one.

docs/src/install-guide.md
```diff
@@ -5,7 +5,9 @@ Kani offers an easy installation option on two platforms:
 * `x86_64-unknown-linux-gnu` (Most Linux distributions)
 * `x86_64-apple-darwin` (Intel Mac OS)
 
-Other platforms are either not yet supported or require instead that you [build from source](build-from-source.md).
+Other platforms are either not yet supported or require instead that
+you [build from source](build-from-source.md). To use Kani in in your
+GitHub CI workflows, see [GitHub CI Action](./install-github-ci.md).
```
celinval, line 10: Can you explain how users can use this action? Maybe add a new section to this page.

docs/src/SUMMARY.md
```diff
@@ -3,6 +3,7 @@
 - [Getting started](./getting-started.md)
   - [Installation](./install-guide.md)
     - [Building from source](./build-from-source.md)
+    - [GitHub CI Action](./install-github-ci.md)
```
celinval, line 6: Did you forget to add this file to this PR?

YoshikiTakashima: Fixed with c5a50dc841578cb47019b4711712e36b4a551fce

docs/src/install-guide.md
YoshikiTakashima: Snippet in docs. c5a50dc841578cb47019b4711712e36b4a551fce

YoshikiTakashima:
Once this is merged, somebody with sufficient permissions will need to enable actions in the GH marketplace.

YoshikiTakashima:
This issue is no longer blocked. Please do a second pass.

Also, to pin the image to the respective version, I had to make a hard-coded value in action.yml with a "update version" check every release. If anyone knows an auto-upgrading solution, please let me know.

action.yml
```diff
@@ -0,0 +1,19 @@
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+name: 'Kani Model Checker'
+description: 'Run the Kani Model Checker on a Cargo Crate'
```
adpaco-aws, line 5: Can you replace "Kani Model Checker" with "Kani Rust Verifier"? We don't use "Kani Model Checker" anywhere.

docs/src/install-github-ci.md
```diff
@@ -0,0 +1,37 @@
+# GitHub CI Action
+
+Kani offers a GitHub action for running Kani in the CI. As of now,
+only Ubuntu 20.04 wtih `x86_64-unknown-linux-gnu` is supported for
+Kani in the CI.
```
adpaco-aws, line 5: 1. I would say "in CI" or "the CI <something>" (where `<something> := system | workflow | pipelines | ...`, but not "in the CI".
2. "wtih" -> "with"
3. "action" -> "Action"

docs/src/install-github-ci.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Using Kani in Your GitHub Workflow.
```
adpaco-aws, line 7:
```suggestion
## Using Kani in your GitHub workflow
```

docs/src/install-github-ci.md
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - name: 'Checkout your code.'
+        uses: actions/checkout@v2
+
+      - name: 'Run Kani on your code.'
+        uses: model-checking/kani@vVER.SION
+```
+
+
+## Configuring Kani with Flags
```
adpaco-aws, line 32:
```suggestion
## Configuring Kani with flags
```

docs/src/install-github-ci.md
```diff
@@ -0,0 +1,37 @@
...
+
+The github action itself does not take any flags that `cargo kani`
```
adpaco-aws, line 34:
```suggestion
The GitHub Action itself does not take any flags that `cargo kani`
```

docs/src/install-guide.md
```diff
@@ -5,7 +5,9 @@ Kani offers an easy installation option on two platforms:
 * `x86_64-unknown-linux-gnu` (Most Linux distributions)
 * `x86_64-apple-darwin` (Intel Mac OS)
 
-Other platforms are either not yet supported or require instead that you [build from source](build-from-source.md).
+Other platforms are either not yet supported or require instead that
+you [build from source](build-from-source.md). To use Kani in in your
```
adpaco-aws, line 9: Duplicated "in".

adpaco-aws (commented):
Thanks @YoshikiTakashima !

One question: How hard would it be to extend this action with an include/exclude list of proofs to run?

In my experience, users may want to exclude a proof from being run in CI, either because it's too time-consuming or it no longer passes after changing the function under verification. That said, I think the right way to do that would be to add another attribute like "ignore" to the proofs themselves, so `cargo kani` can handle that instead.

action.yml
YoshikiTakashima: 06755ef4508250152023d99bed7632600f29bfb8

docs/src/install-github-ci.md
YoshikiTakashima: 9113c49ae54e848bf702b0ab31822de95f5ee3cd

docs/src/install-guide.md
YoshikiTakashima: ad37719b6c3de3d82fb3500531af3de6ccb68e62

YoshikiTakashima:
@adpaco-aws As of now, any flag that works with rust can be used with `[package.metadata.kani.flags]`. So `--harness` will work, but I think it does not allow multiple harnesses to be specified at this point.

.github/workflows/release.yml
```diff
@@ -138,3 +143,9 @@ jobs:
             org.opencontainers.image.source=${{ github.repositoryUrl }}
             org.opencontainers.image.version=${{ needs.Release.outputs.version }}
             org.opencontainers.image.licenses=Apache-2.0 OR MIT
+
+      - name: Check action and image is updated.
+          uses: ${{ env.OWNER_LC }}/kani@${{ needs.Release.outputs.version }})
+          with:
+            command: |
+              (cargo kani --version | grep -F ${{ needs.Release.outputs.version }})
```
danielsn, line 151: Does it need to be grep?

README.md
```diff
@@ -9,6 +9,12 @@ Kani verifies:
  * The absence of panics (e.g., `unwrap()` on `None` values)
  * The absence of some types of unexpected behavior (e.g., arithmetic overflows)
 
+## GitHub Action
+
+Use Kani in your CI with `model-checking/kani@vVERSION`. See [GitHub
+Action
+section](https://model-checking.github.io/kani/install-github-ci.html).
```
danielsn, line 16: section?

danielsn (approved):
LGTM, a few comments

README.md
YoshikiTakashima: Haven't completely eliminated grep, but improved.

cb8ec7ab51ce78255da9e64ccf0e2c32b47bd884

PR #1590, opened 2022-08-26
Make panic and assert const functions

### LINKED ISSUE #1586: "cannot call non-const fn `kani::assert` in constant functions"

zhassan-aws:
### Description of changes: 

Rust allows `panic!`, `assert!`, and `assert_eq!` to be used in `const` functions. Kani's override of the `panic` and `assert` macros is implemented via a `panic` and an `assert` function that are not marked as constant, which causes an error when those macros are used in a const function. This PR fixes the issue by adding `const` to the `panic` and `assert` functions.

### Resolved issues:

Resolves #1586 

### Testing:

* How is this change tested? Added two tests

### Checklist

### DIFF
--- a/tests/kani/Panic/const.rs
+++ b/tests/kani/Panic/const.rs
@@ -1,0 +1,16 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This test checks that `panic!` can be used in a const fn
+
+const fn my_const_fn() {
+    panic!()
+}
+
+#[kani::proof]
+pub fn check_something() {
+    let x: u8 = if kani::any() { 3 } else { 95 };
+    if x > 100 {
+        my_const_fn();
+    }
+}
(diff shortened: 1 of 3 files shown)


tests/kani/Panic/const.rs
```diff
@@ -0,0 +1,16 @@
...
+
+const fn my_const_fn() {
+    panic!()
+}
+
+#[kani::proof]
+pub fn check_something() {
+    let x: u8 = if kani::any() { 3 } else { 95 };
+    if x > 100 {
+        my_const_fn();
```
adpaco-aws, line 14: This call isn't reachable (`x > 100` is always false). The test shows you can use `panic!` in a `const fn` but I think you want to also call it, right?

tedinski: The impetus for this was definitely a compile-time error, so maybe not

zhassan-aws: Correct, I'm intentionally not calling it so that the test passes. As Ted pointed out, compilation currently fails. The purpose of the `kani::any()` is to make sure the compiler doesn't optimize the function out (so we still have to codegen it).

zhassan-aws:
I updated the PR to make `assert` const as well since it's also allowed to be used in a `const` function: https://rust-lang.github.io/rfcs/2345-const-panic.html

tedinski (approved):

PR #1592, opened 2022-08-26
Update kani.yml

celinval:
### Description of changes: 

Change kani workflow to still run on development and feature branches. Only Ignore tags (per original comment). The main motivation is that developers can have actions running in their development branches. Feature branches also benefit from this.

### Resolved issues:

N/A

### Testing:

### Checklist

### DIFF
--- a/.github/workflows/kani.yml
+++ b/.github/workflows/kani.yml
@@ -6,8 +6,8 @@ on:
   push:
     # Not just any push, as that includes tags.
     # We don't want to re-trigger this workflow when tagging an existing commit.
-    branches:
-      - main
+    tags-ignore:
+      - '**'
 
 env:
   RUST_BACKTRACE: 1


.github/workflows/kani.yml
```diff
@@ -6,8 +6,8 @@ on:
   push:
     # Not just any push, as that includes tags.
     # We don't want to re-trigger this workflow when tagging an existing commit.
-    branches:
-      - main
```
tedinski, line 10: I think `branches: *` might also have worked, but as long as this still ignores the tags successfully, lgtm.

tedinski (approved):

.github/workflows/kani.yml
celinval: indeed.

celinval: Actually, I broke everything. :( I'm implementing your solution instead.

Issue #1588, opened 2022-08-25
Tracking issue for MIR linker


celinval:
## Summary

This is a tracking issue for the RFC `mir_linker` which will enable users to use non-generic + non-inlined methods of the standard library.

During development, this feature will be gated by the unstable option `--mir-linker`.

## Major Steps

Suggestion of major steps:

- [x] Get feedback on the RFC PR.
- [x] Implement compiler options `--reachability=[legacy | none | harnesses]`
- [x] Create a mechanism to generate and ship Kani sysroot.
- [x] Introduce unstable user option `--mir-linker`.
- [x] Implement compiler option `--reachability=pub_fns`.
- [ ] Stabilization PR.

* Broken down issues will be tracked by https://github.com/model-checking/kani/milestone/12

## Open Questions

- Should we build or download the sysroot during `kani setup`?

PR #1600, opened 2022-08-26
Create RFC for MIR linker

celinval:
### Description of changes: 

Create an RFC for the implementation of a MIR linker. The MIR linker would provide a performant solution to solve issues for missing `std` definitions.

I created two prototypes, one for the MIR Linker and one without the linker. They both compile the `std` library as part of `cargo kani`. Here are the links for the [MIR Linker prototype](https://github.com/celinval/kani-dev/tree/issue-1588-proto-linker) and the [Alternative prototype](https://github.com/celinval/kani-dev/tree/issue-1588-alt-proto).

You can also find a rendered version of this document [here](https://github.com/celinval/kani-dev/blob/issue-1588-rfc/rfcs/src/rfcs/0001-mir-linker.md).

### Resolved issues:

Related to #1588 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


giltho (commented):
That sounds great!
I added two tiny comments, one of them is just a confirmation that the MIR-linking approach seems the best. I don’t know if the information should be added to the RFC itself.

 Regarding the code of the mir-linker itself, [reachability.rs](http://reachability.rs/) (http://reachability.rs/) is typically something I’d like to have in a tools-lib crate. I noticed that the only dependency on Kani is the call to GotocCtx::is_proof_harness. Could ctx parameter of collect_reachable_items be changed to a select : Fn(&Instance) -> bool parameter instead and that could actually be shared with the rest of the Rust verification community.
 
Should someone internal to Rust and external to Kani be asked to quickly read this RFC and provide feedback?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+This sysroot will be pre-compiled and included in our release bundle.
+
+We will compile `kani`'s libraries (`kani` and `std`) also with `-Z always-encode-mir`
+and with the new sysroot.
+
+
+### Cross-Crate Reachability Analysis
+
+`kani-compiler` will include a new `reachability` module to traverse over the local and external `MIR` items.
+This module will `monomorphize` all generic code as it's performing the traversal.
```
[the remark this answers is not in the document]
celinval: That is a great question! Conceptually, I don't see any reason why the reachability code couldn't work with polymorphic definitions. If we decide to contribute this code to upstream or put in separate library, it would be a great thing to do.

For our current goal, it does make sense to monomorphize here since Kani's compiler just wants a list of items to generate code for.

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+
+Finally, this implementation relies on a few unstable options from `cargo` and `rustc`.
+These APIs are used by other tools such as MIRI, so we don't see a high risk that they would be removed.
+
+### Alternatives
+
+The other options explored were:
+1. Pre-compile the standard library and ship the generated `*symtab.json` files.
+2. Pre-compile the standard library, convert them to `goto-programs` (via `symtab2gb`) and link them into
+   one single `goto-program` model. Ship the generated model.
```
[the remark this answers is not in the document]
celinval: Huum... I think the process we would follow in these alternatives  would still be sound. We would still use the *rlib / *rmeta files that are included as part of the rustup toolchain sysroot which should have the layout information.

I guess I wasn't very clear here, and that's a great point. I didn't specify this here, but we would also need to include the metadata files for the Kani libraries in the sysroot, no matter what to keep it consistent.

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+3. For every harness, Kani prunes the linked model to only include items reachable from the given harness.
+4. Finally, Kani instruments and verify each harness model via `goto-program` and `cbmc` calls.
+
+After this RFC, the system flow would be slightly different:
+
+1. Kani compiles the user crate dependencies up to the `MIR` translation.
+   I.e., for every crate compiled, `kani-compiler` will generate an artifact that includes the `MIR` representation
+  of all items in the crate.
+2. Kani will generate the `goto-program` only while compiling the target user crate.
+  It will generate one `goto-program` model that includes all items reachable from every harness in the target crate.
```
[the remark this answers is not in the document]
celinval: It's a very important detail. Thanks!

rfcs/src/rfcs/0001-mir-linker.md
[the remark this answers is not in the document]
giltho: Yes totally agree

giltho:
Just noticed the `https://reachability.rs` strangeness, that happened because I wrote that comment in apple notes first and copy pasted in gh, sorry if that was confusing

giltho:
> One of my thoughts was to use this reachability module as my starting point to perform further stable MIR experiments. :)

Very excited for your future work 🤩

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+
+The current system flow for a crate verification is the following (Kani here represents either `kani | cargo-kani`
+executable):
+
+1. Kani compiles the user crate as well as all its dependencies.
+   For every crate compiled, `kani-compiler` will generate a `goto-program` model.
+   This model includes everything reachable from the crate's public functions.
+2. After that, Kani links all models together by invoking `goto-cc`.
+   This step will also link against Kani's `C` library.
+3. For every harness, Kani prunes the linked model to only include items reachable from the given harness.
```
zhassan-aws, line 55:
```suggestion
3. For every harness, Kani invokes `goto-instrument` to prune the linked model to only include items reachable from the given harness.
```

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+4. Finally, Kani instruments and verify each harness model via `goto-program` and `cbmc` calls.
```
zhassan-aws, line 56: `goto-program` -> `goto-instrument`?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+2. Kani will generate the `goto-program` only while compiling the target user crate.
+  It will generate one `goto-program` model that includes all items reachable from any harness in the target crate.
+3. `goto-cc` will still be invoked to link the generated model against Kani's `C` library.
+4. Steps #3 and #4 above will be performed without any change.
+
+This feature will require three main changes to Kani which are detailed in the sub-sections below.
+
+### Kani's Sysroot
+
+Kani currently uses `rustup` sysroot to gather information from the standard library constructs.
```
zhassan-aws, line 72: I'm a bit unclear on this section. What standard library constructs are looked up in `sysroot`? Does `sysroot` have MIR lying around?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+  cargo build <FLAGS>
+  ```
+
+  with:
+
+  ```
+  cargo rustc <FLAGS> -- --reachability=harnesses
+  ```
+
+  to build everything.
```
zhassan-aws, line 137: As far as I understand, we've been primarily using the `KANIFLAGS` environment variable to pass things to the `kani-compiler`? If this is correct, any reason to switch to using `cargo rustc`?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+the harness verification:
+
+
+| Stage                     | MIR Linker | Alternative |
+----------------------------|------------|-------------|
+| compilation               |   22.2s    |    64.7s    |
+| `goto-program` generation |   2.4s     |    90.7s    |
+| `goto-program` linking    |   0.8s     |    33.2s    |
+| code instrumentation      |   0.8s     |    33.1     |
+| verification              |   0.5s     |    8.5s     |
```
zhassan-aws, line 208: The results are amazing! I'd expect memory usage to be significantly lower with the MIR linker as well.

If you still have the log files, can you compare the size of the SAT instance (clauses/literals) between the two alternatives? I'm curious to know how much CBMC's slicer is able to reduce the problem size, and how close it can get to the one generated by the MIR linker.

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+- Should we build or download the sysroot during `kani setup`?
+- What's the best way to enable support to run Kani in the entire `workspace`?
+  - One possibility is to run `cargo rustc` per package.
+- Should we codegen all static items no matter what? Static objects can only be initialized via constant function.
+  Thus, it shouldn't have any side effect.
+  That relies on all constant initializers being evaluated during compilation.
+
+## Future possibilities
+
+- Split `goto-program` models into two or more items to optimize compilation result caching.
```
zhassan-aws, line 237: A related question: are we perhaps able to skip re-generating the MIR for a dependent crate that hasn't changed, and reuse it in the following steps? `cargo build` (and hence `cargo kani`) sort of already does that when compiling a package, but we always run `symtab2gb` even if the symtab file hasn't changed.

zhassan-aws (commented):
This is very exciting! Thanks @celinval. Just had a few comments/questions.

rfcs/src/rfcs/0001-mir-linker.md
celinval: Yes. The rustup sysroot does have MIR lying around. They are inside the *rlib files shipped with the toolchain (see more details here: https://rustc-dev-guide.rust-lang.org/backend/libs-and-metadata.html#metadata).

rfcs/src/rfcs/0001-mir-linker.md
celinval: The main reason is that we want to pass an extra flag to the target crate which we wouldn't pass to the dependencies.

rfcs/src/rfcs/0001-mir-linker.md
celinval: Yeah, I totally forgot to include the memory usage here.

Both prototypes run the test described in CI. So you can see the log files for the CI run for [MIR linker](https://pipelines.actions.githubusercontent.com/serviceHosts/14dcd0bf-cf10-416a-af82-d1e50ea29e0a/_apis/pipelines/1/runs/620/signedlogcontent/10?urlExpires=2022-08-31T00%3A11%3A05.1747440Z&urlSigningMethod=HMACV1&urlSignature=5K%2B8bYQk7nidag1e9Ie3uz2aTF7j2OKVXm6oRVQ5C1Y%3D) and the [alternative](https://pipelines.actions.githubusercontent.com/serviceHosts/14dcd0bf-cf10-416a-af82-d1e50ea29e0a/_apis/pipelines/1/runs/622/signedlogcontent/5?urlExpires=2022-08-31T00%3A11%3A10.9310206Z&urlSigningMethod=HMACV1&urlSignature=S7i6ijOKpOYInWy6HTfBT3joP%2B3wQpluhdnDXFBdC1E%3D).

The specific data I believe you are asking for is:
```
14089 variables, 180 clauses
```
and 
```
21610 variables, 7514 clauses
```
for MIR Linker and the alternative, respectively.

It just occurred to me that I should try to check the difference under visualize. It might make it more clear to understand the difference.

rfcs/src/rfcs/0001-mir-linker.md
celinval: We would be able to skip re-generating the MIR for the dependencies but we would still generate the goto-program to everything linked at every compilation.

Being able to cache the goto-program generation for things that haven't changed would require more work and this is what I'm trying to capture here.

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+This sysroot will be pre-compiled and included in our release bundle.
+
+We will compile `kani`'s libraries (`kani` and `std`) also with `-Z always-encode-mir`
+and with the new sysroot.
+
+
+### Cross-Crate Reachability Analysis
+
+`kani-compiler` will include a new `reachability` module to traverse over the local and external `MIR` items.
+This module will `monomorphize` all generic code as it's performing the traversal.
```
oli-obk, line 91: rustc already has logic for this in https://github.com/rust-lang/rust/blob/master/compiler/rustc_monomorphize/src/collector.rs

Maybe it could be generalized to allow you to plug into it and make it work cross crate (rustc only needs cross crate for generic functions and will assume nongeneric functions got codegened upstream)

We have for now given up on mir-only rlibs, but if we want to pick up that work, we'd need exactly what you are describing anyway, so it doesn't seem unreasonable to change the collector to support it optionally.

celinval: That would be great! Most of the code I've written so far was based on the collector logic.

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+- It won't require any changes to our dependencies.
+- This will fix the harnesses' dependency on the`#[no_mangle]` annotation
+([Issue-661](https://github.com/model-checking/kani/issues/661)).
+
+### Risks
+
+Failures in the linking stage would not impact the tool soundness. I anticipate the following failure scenarios:
+- ICE (Internal compiler error): Some logic is incorrectly implemented and the linking stage crashes.
+  Although this is a bad experience for the user, this will not impact the verification result.
+- Missing items: This would either result in ICE during code generation or a verification failure if the missing
```
danielsn, line 159: What about if its used as a potential function pointer?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+|----------------|----------|-----------------|
+| `symtab.json`  |   950M   |     26M         |
+| `symtab.out`   |   84M    |     24M         |
+| `*.rmeta`      |   92M    |     25M         |
+
+These results were obtained by looking at the artifacts generated during the same experiment.
+
+## Open questions
+
+- Should we build or download the sysroot during `kani setup`?
```
danielsn, line 228: How do we handle cases during development when we modify something the sysroot depends on

rfcs/src/rfcs/0001-mir-linker.md
celinval: That's a great point. CBMC should catch that once the following PR gets merged: https://github.com/diffblue/cbmc/pull/6987

Another possibility is to change how we invoke `--goto-instrument` as pointed out here: https://github.com/diffblue/cbmc/issues/6983#issuecomment-1172020654

I created https://github.com/model-checking/kani/issues/1621 to address that issue and added to https://github.com/model-checking/kani/milestone/12

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+and the generation of the `goto-program` model only when compiling the target crate.
+
+Kani setup will likely take longer and more disk space as mentioned in the section above.
+This change will not be guarded by `--mir-linker` option above.
+
+## Detailed Design
+
+In a nutshell, we will no longer generate a `goto-program` model for every crate we compile.
+Instead, we will generate the `MIR` for every crate, and we will generate only one `goto-program` model.
+This model will only include items reachable from the target crate's harnesses.
```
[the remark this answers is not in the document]
celinval: Totally. I put that as a future work for now. Let me know if you think this should be in the scope of the initial work.

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+
+The approach introduced in this RFC will have the following secondary benefits.
+- Reduce spurious warnings about unsupported features for cases where the feature is not reachable from any harness.
+- In the verification mode, we will likely see a reduction on the compilation time and memory consumption
+ by pruning the inputs of symtab2gb and goto-instrument.
+  - Linking against the standard library goto-models takes more than 5 GB which is not feasible with the current
+  GitHub workers.
+- In a potential assessment mode, only analyze code that is reachable from all public items in the target crate.
+
+One downside is that our release bundle will be bigger.
```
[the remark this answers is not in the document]
celinval: Approx 26MB bigger for linux 64bits. Kani 0.9 bundle size was 22MB. So this will more than double the size. That said, we also download the entire rustup toolchain as part of `kani setup` and I believe it has more than 50MB.

rfcs/src/rfcs/0001-mir-linker.md
celinval: During development, we shall build the sysroot together with all our binaries. I haven't figured out the best mechanism yet though. Building the sysroot as part of the `kani-compiler` build isn't quite working for me. The sysroot will require different cargo configuration, so I am considering splitting our workspace and creating a new workspace for the `library` folder.

celinval:
> Is it possible to have more than one sysroot?

Yes. The `kani-compiler` already includes a `--sysroot` option to configure which sysroot to use. So it would be a matter of building them separately and selecting the correct one.

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+- **RFC Tracking Issue**: <https://github.com/model-checking/kani/issues/1588>
+- **RFC PR:** <https://github.com/model-checking/kani/pull/1600>
+- **Status:** Under Review
+- **Version:** 0
+
+-------------------
+
+## Summary
+
+Fix linking issues with the rust standard library in a scalable manner by only generating `goto-program` for
```
adpaco-aws, line 11: Another nit: I'd use "goto-program(s)" (see [CBMC docs](http://cprover.diffblue.com/group__goto-programs.html)) instead since `goto-program` is not a tool or module. You frequently say "`goto-program` model" but I think "goto-program" and "goto model" are the same thing.

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+
+The main goal of this RFC is to enable Kani users to link against all supported constructs from the `std` library.
+Currently, Kani will only link to items that are either generic or have an inline annotation.
+
+The approach introduced in this RFC will have the following secondary benefits.
+- Reduce spurious warnings about unsupported features for cases where the feature is not reachable from any harness.
+- In the verification mode, we will likely see a reduction on the compilation time and memory consumption
+ by pruning the inputs of symtab2gb and goto-instrument.
+  - Linking against the standard library goto-models takes more than 5 GB which is not feasible with the current
+  GitHub workers.
```
tedinski, line 24: Just confirming this sentence is talking about the alternative approach, not the "mir linker" approach you're recommending?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+
+## User Impact
+
+The main goal of this RFC is to enable Kani users to link against all supported constructs from the `std` library.
+Currently, Kani will only link to items that are either generic or have an inline annotation.
+
+The approach introduced in this RFC will have the following secondary benefits.
+- Reduce spurious warnings about unsupported features for cases where the feature is not reachable from any harness.
+- In the verification mode, we will likely see a reduction on the compilation time and memory consumption
+ by pruning the inputs of symtab2gb and goto-instrument.
```
adpaco-aws, line 22: You mean in the verification stage? Are `symtab2gb` and `goto-instrument` called then?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+   This model includes everything reachable from the crate's public functions.
+2. After that, Kani links all models together by invoking `goto-cc`.
+   This step will also link against Kani's `C` library.
+3. For every harness, Kani invokes `goto-instrument` to prune the linked model to only include items reachable from the given harness.
+4. Finally, Kani instruments and verify each harness model via `goto-program` and `cbmc` calls.
+
+After this RFC, the system flow would be slightly different:
+
+1. Kani compiles the user crate dependencies up to the `MIR` translation.
+   I.e., for every crate compiled, `kani-compiler` will generate an artifact that includes the `MIR` representation
```
tedinski, line 61: "an artifact" is a bit vague. What? Something we put in the `rlib`?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+This feature will require three main changes to Kani which are detailed in the sub-sections below.
+
+### Kani's Sysroot
+
+Kani currently uses `rustup` sysroot to gather information from the standard library constructs.
+The artifacts from this `sysroot` include the `MIR` for generic items as well as for items that may be included in
+a crate compilation (e.g.: functions marked with `#[inline]` annotation).
+The artifacts do not include the `MIR` for items that have already been compiled to the `std` shared library.
+This leaves a gap that cannot be filled by the `kani-compiler`;
+thus, we are unable to translate these items into `goto-program`.
```
tedinski, line 77: `goto-program` doesn't seem like something that should be in code formatting here. "a goto program" I assume you mean?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+
+In order to fulfill this gap, we must compile the standard library from scratch.
+This RFC proposes a similar method to what [`MIRI`](https://github.com/rust-lang/miri) implements.
```
tedinski, line 80: Can... we just use what MIRI ships?

We install a rust toolchain on install. I'd been thinking about whether we could reduce this to a minimal profile as an optimization, but we should also be able to go the other way and add the miri component?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+
+
+### Cross-Crate Reachability Analysis
+
+`kani-compiler` will include a new `reachability` module to traverse over the local and external `MIR` items.
+This module will `monomorphize` all generic code as it's performing the traversal.
+
+The traversal logic will be customizable allowing different starting points to be used.
+The two options to be included in this RFC is starting from all local harnesses
+(tagged with `#[kani::proof]`) and all public functions in the local crate.
```
tedinski, line 95: I assume that under `--test` tests are marked public and that keep them here?

I also wonder if we should just... include both by default and not have options here? This avoids the complication of setting/flags/configuration, doesn't seem like a major downside with slightly more code compiled, and makes everything more flexible.

E.g. a few minutes of thought comes up with the case of Rust calling C code and that C calling Rust code that's "extern C" and which we can't inspect the reachability of, but including both harnesses and public items both would handle seamlessly.

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+  --reachability=[ harnesses | pub_fns |  none | legacy* ]
+  ```
+
+where:
+
+ - `harnesses`: Use the local harnesses as the starting points for the reachability analysis.
+ - `pub_fns`: Use the public local functions as the starting points for the reachability analysis.
+ - `none`: This will be the default value if `--reachability` flag is not provided. It will basically skip
+ reachability analysis (and `goto-program` generation).
+   `kani-compiler` will still generate artifacts with the crate's `MIR`.
```
tedinski, line 109: What is the use of none?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+- For standalone `kani`, we will pass the option `--reachability=harnesses` to `kani-compiler`.
+- For `cargo-kani`, we will replace
+  ```
+  cargo build <FLAGS>
+  ```
+
+  with:
+
+  ```
+  cargo rustc <FLAGS> -- --reachability=harnesses
```
tedinski, line 134: What about when we run `cargo kani --tests`

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+The `kani-compiler` behavior will be customizable via a new flag:
+
+  ```
+  --reachability=[ harnesses | pub_fns |  none | legacy* ]
+  ```
+
+where:
+
+ - `harnesses`: Use the local harnesses as the starting points for the reachability analysis.
+ - `pub_fns`: Use the public local functions as the starting points for the reachability analysis.
```
adpaco-aws, line 106: Ok, I think the "modes" you referred to earlier are either "verification" or "assesment". And `pub_fns` corresponds to the "assesment" mode, while "harnesses" corresponds to "verification". Is that right?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+ - `none`: This will be the default value if `--reachability` flag is not provided. It will basically skip
+ reachability analysis (and `goto-program` generation).
```
adpaco-aws, line 108: Well, goto-programs will still be generated, no?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Risks
+
+Failures in the linking stage would not impact the tool soundness. I anticipate the following failure scenarios:
+- ICE (Internal compiler error): Some logic is incorrectly implemented and the linking stage crashes.
+  Although this is a bad experience for the user, this will not impact the verification result.
+- Missing items: This would either result in ICE during code generation or a verification failure if the missing
+  item is reachable.
+- Extra items: This shouldn't impact the verification results, and they should be pruned by CBMC's reachability
+  analysis.
```
adpaco-aws, line 162: CBMC could prune some stuff, but I don't think it's going to be more aggressive than the MIR linker.

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+  This is already the case today. In extreme cases, this could include a symbol that we cannot compile and cause an ICE.
+
+The new reachability code would be highly dependent on the `rustc` unstable APIs, which could increase
+the cost of the upstream synchronization.
+That said, the APIs that would be required are already used today.
```
adpaco-aws, line 167: Nice!

tedinski, line 167: Any chance of upstreaming a clean API we could use?

rfcs/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,244 @@
...
+| `*.rmeta`      |   92M    |     25M         |
+
+These results were obtained by looking at the artifacts generated during the same experiment.
+
+## Open questions
+
+- Should we build or download the sysroot during `kani setup`?
+- What's the best way to enable support to run Kani in the entire `workspace`?
+  - One possibility is to run `cargo rustc` per package.
+- Should we codegen all static items no matter what? Static objects can only be initialized via constant function.
```
tedinski, line 231: I think I said this out loud at one point, but writing it down:

YOLO. Let's try pruning "unused" statics and see what happens! It should be easy to mitigate (stop pruning statics) if we ever find a real issue, right?

rfcs/src/rfcs/0001-mir-linker.md
tedinski: Splitting is one possibility, but IMO before we pursue any of that, we should take seriously optimizing the MIR -> Goto binary process. This would benefit initial as well as subsequent, and I think there's probably a LOT of low-hanging fruit.

Both in our compiler, symtab2gb, and the vague future possibility that we could emit goto binaries directly without going through json and symtab2gb.

rfcs/src/rfcs/0001-mir-linker.md
celinval: I was trying to make a distinction with the `assess mode` which might not invoke CBMC.

rfcs/src/rfcs/0001-mir-linker.md
celinval: No. With `none` we won't generate any `goto-program`. This will be used while compiling the dependencies.

rfcs/src/rfcs/0001-mir-linker.md
celinval: Indeed. CBMC formula slicer might prune more stuff than the MIR linker though.

rfcs/src/rfcs/0001-mir-linker.md
celinval: Yes

rfcs/src/rfcs/0001-mir-linker.md
celinval: Yes. Either `rlib` or `rmeta`.

rfcs/src/rfcs/0001-mir-linker.md
celinval: Yeah. I clearly did a poor job at code formatting. I'll fix that. :)

rfcs/src/rfcs/0001-mir-linker.md
celinval: That's a great point. I'm assuming the overhead will be the same, but it might save us from maintaining yet another build script.

Let me investigate that.

rfcs/src/rfcs/0001-mir-linker.md
celinval: While compiling the crate dependencies, we want to skip codegen completely. That's what "none" is for.

rfcs/src/rfcs/0001-mir-linker.md
celinval: Yes. That's exactly what I'm doing. :)

rfcs/src/rfcs/0001-mir-linker.md
celinval: Great question. At some point I thought about adding this to the "Open Questions", but I totally forgot. I'll do that now.

I don't have a great answer yet. The two possibilities I thought so far are on the hacky side, but I believe they would work:

1. Add a new value to `--reachability`, maybe "on-tests" or something like that. This would behave as either "harness" or "none" depending whether RUSTFLAGS has `--tests`.
2. Change how we handle not having `--reachability` argument to also take into account the existence of `--tests`.

Both would ensure that we only run the reachability and codegen at the last compilation step. Any other suggestions?

rfcs/src/rfcs/0001-mir-linker.md
celinval: That's the hope. I added that to future work section, and it is the same feedback that @oli-obk provided. Fingers crossed!

rfcs/src/rfcs/0001-mir-linker.md
celinval: I fixed the text. I can see why it was confusing. :)

rfcs/src/rfcs/0001-mir-linker.md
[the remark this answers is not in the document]
celinval: Changed the text a bit. Let me know if it's clearer now.

rfcs/src/rfcs/0001-mir-linker.md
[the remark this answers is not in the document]
celinval: @danielsn is it OK if I leave it like this for the first version of the MIR Linker?

rfcs/src/rfcs/0001-mir-linker.md
celinval: @zhassan-aws what kind of information would you like me to add here to make this more clear?

rfcs/src/rfcs/0001-mir-linker.md
celinval: @tedinski please let me know your thoughts on this one.

rfcs/src/rfcs/0001-mir-linker.md
celinval: @tedinski does that make sense? Let me know if you want me to change anything or make the text more clear about the customization points.

rfcs/src/rfcs/0001-mir-linker.md
celinval: @tedinski and @adpaco-aws I added a bit more information here. Let me know if that makes sense.

rfcs/src/rfcs/0001-mir-linker.md
celinval: @tedinski any suggestions? I added this to the open questions too.

rfcs/src/rfcs/0001-mir-linker.md
celinval: Unfortunately, this will still keep KANIFLAGS working as is today.

rfcs/src/rfcs/0001-mir-linker.md
tedinski: Shucks, I was hoping MIRI would ship it. I'm ok with adding it as part of `setup` or pre-building it. Though I'm guessing MIRI doesn't pre-build it for a good reason?

rfcs/src/rfcs/0001-mir-linker.md
tedinski: I'm fine with whichever course you choose, but my preference is that we should error on the side of simplicity here. IMO it should just use "public + harnesses" as roots and then we don't need to worry about other options. It seems like the worse case scenario is just codegen'ing a bit more than we need, while the advantage is that everything just works and nobody has to think about this again.

rfcs/src/rfcs/0001-mir-linker.md
tedinski: My question was less on the reachability side here, and more on the side of the actual pragmatics of doing the build.

`cargo build` becomes `cargo rustc` but what does `cargo test --no-run` become?

rfcs/src/rfcs/0001-mir-linker.md
celinval: @danielsn Would you like me to add this information here? I tried to focus on the end user point of view, but it might make sense to clarify this.

tedinski (approved):

rfcs/src/rfcs/0001-mir-linker.md
celinval: @oli-obk what's the motivation to build MIRI sysroot as part of the first execution of MIRI opposed to pre-building it?

rfcs/src/rfcs/0001-mir-linker.md
zhassan-aws: I think it's fine as is.

rfcs/src/rfcs/0001-mir-linker.md
celinval: It will likely stay as is and we might add a `--reachability=on-tests` or something like that.

celinval: I have some great news, we can replace
```
cargo test --no-run
```
by
```
cargo rustc --tests -- --reachability=harnesses
```

I'll add some details here and remove that from the open questions.

tedinski: Fantastic!

rfcs/src/rfcs/0001-mir-linker.md
oli-obk: miri can cross interpret, so we'd need to deploy all targets somewhere once (instead of once per host). Also sometimes you'll want to change how the sysroot is built I guess.

Also we need the ability to build the sysroot from scratch anyway for when we modify miri out of tree.

rfc/src/rfcs/0001-mir-linker.md
```diff
@@ -0,0 +1,256 @@
...
+ - `none`: This will be the default value if `--reachability` flag is not provided. It will skip
+ reachability analysis. No goto-program will be generated.
+  This will be used to compile dependencies up to the MIR level.
+   `kani-compiler` will still generate artifacts with the crate's MIR.
+ - `legacy`: Keep `kani-compiler` current behavior by using
+   `rustc_monomorphizer::collect_and_partition_mono_items()` which respects the crate boundary.
+   This will generate a goto-program for each crate compiled by `kani-compiler`, and it will still have the same
+   `std` linking issues.
+
+*This option will be removed as part of the `rfc` stabilization.
```
adpaco-aws, line 116: Since `*` doesn't link or anything, I suggest removing it in:
 ```
  --reachability=[ harnesses | pub_fns |  none | legacy* ]
  ```
and include this text in the item above directly. If you like, you can highlight as "**Note:** This option will be removed as part of the `rfc` stabilization."

adpaco-aws (approved):

rfc/src/SUMMARY.md
```diff
@@ -6,4 +6,5 @@
 
 # Kani RFCs
 
+- [0001-mir-linker](rfcs/0001-mir-linker.md)
```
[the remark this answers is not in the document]
tedinski: I think it's meant to mimic the rust rfc book, which does use the file names, probably so they're more easily googable/greppable? https://rust-lang.github.io/rfcs/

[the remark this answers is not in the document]
adpaco-aws: Wasn't aware of that. I've no clue why they'd do it this way, but feel free to leave as is.

rfc/src/rfcs/0001-mir-linker.md
celinval: sure

PR #1601, opened 2022-08-28
refactor: make concrete values parser use serde trace structs

### LINKED ISSUE #1477: Update concrete values parser to use serde structs
The current deterministic values parser in `kani-driver/src/exe_trace.rs` works on arbitrary `serde::Value` types. This means that if we expect certain `Value` blobs to be arrays, we have to test for that and handle errors. With this change, we would complete the serde struct for traces in `kani-driver/src/cbmc_output_parser.rs` and use those objects directly. We would still have to walk down this structure to extract deterministic values, but we wouldn't need as much error handling since the blob would be automatically serialized into the right type.

sanjit-bhat:
### Description of changes: 

This PR does the following:
1. It adds more trace properties to the CBMC output parser.
2. It removes the previous logic in the CBMC output parser to stream the output to a file.
3. It refactors the concrete values parser. Before, it read the CBMC output file, parsed it, and extracted concrete values. Now, it takes in the parsed CBMC output and extracts concrete values.

### Resolved issues:

Resolves #1477.

### Testing:

* How is this change tested?

No tests were added. The CBMC output parser updates are tested via none of the existing regressions failing. The concrete values extractor is tested via all the concrete playback tests.

* Is this a refactor change?

Yes, in some ways; no, in other ways. It is not a refactor change because it adds new fields to the CBMC output parser. It is a refactor change because the higher-level functionality of extracting concrete values still stays the same.

### Checklist


sanjit-bhat:
This should probably be reviewed by at least @celinval and @adpaco-aws.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -549,31 +550,30 @@ pub fn process_cbmc_output(
...
-    output_filename_opt: Option<&Path>,
-) -> VerificationStatus {
+) -> VerificationOutput {
     let stdout = process.stdout.as_mut().unwrap();
     let mut stdout_reader = BufReader::new(stdout);
-    let parser = Parser::new(&mut stdout_reader, output_filename_opt);
+    let parser = Parser::new(&mut stdout_reader);
     let mut result = false;
-
-    for item in parser {
```
adpaco-aws, line 559: Splitting this loop into two (one to process, another to print) means we would not print anything in non-terminating cases (as in #493), unless I'm missing something. We need to keep the original loop here.

adpaco-aws (changes requested):

kani-driver/src/cbmc_output_parser.rs
sanjit-bhat: Ah, thanks for pointing that out. I fixed it.

kani-driver/src/call_cbmc.rs
```diff
@@ -70,11 +60,8 @@ impl KaniSession {
             println!("Verification Time: {}s", elapsed);
         }
 
-        if let Ok(VerificationStatus::Failure) = verification_result {
-            self.gen_and_add_concrete_playback(&output_filename, harness);
-        }
-
-        verification_result
+        self.gen_and_add_concrete_playback(harness, &verification_output)?;
```
celinval, line 63: This code looks much cleaner. Thanks

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -307,7 +307,39 @@ pub struct TraceItem {
...
+#[serde(untagged)]
+pub enum TraceData {
+    NonBool(String),
+    Bool(bool),
+}
+
+impl std::fmt::Display for TraceData {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let trace_data_str = match self {
+            Self::NonBool(trace_data) => trace_data.to_string(),
```
celinval, line 338: nit:
```rust
let trace_data_str = match self {
            Self::NonBool(trace_data) |  Self::Bool(trace_data) => trace_data.to_string(),
};
```

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -549,12 +550,12 @@ pub fn process_cbmc_output(
...
-) -> VerificationStatus {
+) -> VerificationOutput {
     let stdout = process.stdout.as_mut().unwrap();
     let mut stdout_reader = BufReader::new(stdout);
-    let parser = Parser::new(&mut stdout_reader, output_filename_opt);
+    let parser = Parser::new(&mut stdout_reader);
     let mut result = false;
+    let mut processed_items: Vec<ParserItem> = Vec::new();
 
     for item in parser {
```
celinval, line 560: nit: you could change this to be something like 
```rust
let processed_items = parser.filter_map(|item| {
  if item.must_be_skipped() {
    None;
  }
  // ...
}
```

kani-driver/src/concrete_playback.rs
```diff
@@ -18,42 +21,64 @@ use std::process::Command;
...
         );
 
-        let concrete_vals = parser::extract_concrete_vals(output_filename).expect(
-            "Something went wrong when trying to get concrete values from the CBMC output file",
-        );
-        let concrete_playback = format_unit_test(&harness.mangled_name, &concrete_vals);
+        if verification_output.status == VerificationStatus::Success {
+            if !self.args.quiet {
+                println!(
+                    "INFO: The concrete playback feature does not generate unit tests when verification succeeds."
```
celinval, line 41: Since we have removed the `--harness [h]` requirement, I think this info here might be too verbose.

kani-driver/src/concrete_playback.rs
```diff
@@ -18,42 +21,64 @@ use std::process::Command;
...
 
-        if let Some(playback_mode) = &self.args.concrete_playback && *playback_mode == ConcretePlaybackMode::Print && !self.args.quiet {
-            println!(
-                "Concrete playback unit test for `{}`:\n```\n{}\n```",
-                &harness.mangled_name, &concrete_playback.unit_test_str
-            );
-            println!(
-                "To automatically add the concrete playback unit test `{}` to the src code, run Kani with `--concrete-playback=InPlace`.",
-                &concrete_playback.unit_test_name
+        if let (Some(processed_items), Some(playback_mode)) =
```
celinval, line 47: I think this is a bit hard to read. Can you extract the `playback_mode` on line 29 instead?

kani-driver/src/concrete_playback.rs
```diff
@@ -18,42 +21,64 @@ use std::process::Command;
...
+        {
+            let concrete_vals = extract_from_processed_items(processed_items).expect(
+                "Something went wrong when trying to get concrete values from the CBMC output",
             );
-        }
+            let concrete_playback = format_unit_test(&harness.mangled_name, &concrete_vals);
 
-        if let Some(playback_mode) = &self.args.concrete_playback && *playback_mode == ConcretePlaybackMode::InPlace {
-            if !self.args.quiet {
+            if *playback_mode == ConcretePlaybackMode::Print && !self.args.quiet {
```
celinval, line 55: Not for this PR, but shouldn't this be an error? Users shouldn't be able to select `print` mode and `--quiet` at the same time.

celinval (commented):
This looks much cleaner. Thanks

One thing I noticed is that we are printing a lot of INFO messages on a per harness basis. I don't expect you to change that in this PR, but maybe avoid adding more for now.

kani-driver/src/cbmc_output_parser.rs
sanjit-bhat: Great suggestion, thanks!

kani-driver/src/concrete_playback.rs
sanjit-bhat: I'm not sure I understand why `--harness` impacts the verbosity here.

Regardless of whether that's required or not, I see how this could be a bit too verbose. I'm unclear what the correct behavior should be here: 1) print INFO, 2) quietly skip generating a unit test, 3) print ERROR and panic. I think 3) is too much. I assume that if the user inputted this flag, they expect a test case to be generated. So if its not, 2) is too passive, and we should at least tell them with 1) (the current approach). Do you have thoughts on this?

adpaco-aws: Well, if I run `cargo kani` in a project where all proofs succeed (the common case if Kani is integrated in CI), this message would be printed once per proof, so I agree it's a little too verbose. What about printing if `--verbose`? Another option is to change `cargo kani` to print a message (only one, at the end) if the user passed the concrete playback flag but all proofs passed.

kani-driver/src/concrete_playback.rs
```diff
@@ -256,156 +281,127 @@ struct UnitTest {
...
-            let result_msg = &general_msg["result"];
-            if !result_msg.is_null() {
-                let result_arr = result_msg.as_array().with_context(|| {
-                    format!("Expected this CBMC result object to be an array: {}", result_msg)
-                })?;
-                for result_val in result_arr {
-                    parse_result(result_val, &mut concrete_vals, &mut have_parsed_assert_fail)?;
+        let mut extracted_assert_fail = false;
+        for processed_item in processed_items {
+            if let ParserItem::Result { result } = processed_item {
```
adpaco-aws, line 300: You could use a helper function function here to retrieve the result first and avoid a nested loop. If there is no result in `processed_items`, that's an error.

kani-driver/src/concrete_playback.rs
```diff
@@ -256,156 +281,127 @@ struct UnitTest {
...
-
-        if prop_is_assert && status_is_failure {
-            if *have_parsed_assert_fail {
+    /// The second-level extractor. Traverses properties to find trace items.
+    fn extract_from_property(
+        property: &Property,
+        extracted_assert_fail: &mut bool,
+    ) -> Result<Vec<ConcreteVal>> {
+        let mut concrete_vals: Vec<ConcreteVal> = Vec::new();
+        let property_is_assert = property.property.contains("assertion");
```
adpaco-aws, line 322: Wouldn't it be better to check the property class (part of `property`)?

kani-driver/src/concrete_playback.rs
```diff
@@ -256,156 +281,127 @@ struct UnitTest {
...
-            Some(interp_concrete_val),
-            Some(width_u64),
-        ) = (
-            trace_entry["stepType"].as_str(),
-            trace_entry["lhs"].as_str(),
-            trace_entry["sourceLocation"]["function"].as_str(),
-            trace_entry["value"]["binary"].as_str(),
-            trace_entry["value"]["data"].as_str(),
-            trace_entry["value"]["width"].as_u64(),
-        ) {
```
adpaco-aws, line 372: Nice! You don't need `step_type` anymore?

kani-driver/src/concrete_playback.rs
sanjit-bhat: Nice catch! I added an `ensure` for this.

kani-driver/src/concrete_playback.rs
sanjit-bhat: Good point! Done

kani-driver/src/concrete_playback.rs
sanjit-bhat: Do you mean the `extract_property_class` function? I updated the code to use that function.

kani-driver/src/concrete_playback.rs
sanjit-bhat: Nope, we're still using it later in that function. However, because `step_type` is not an option, we don't need to extract it in the `if let`.

kani-driver/src/concrete_playback.rs
sanjit-bhat: Ah, that makes sense. It would be much less verbose when you have a bunch of successful harnesses and only one failing harness.

I updated the code to add a `inform_if_no_failed` function. This informs the user at the very end if there are no failed harnesses.

sanjit-bhat:
@adpaco-aws do you have any additional feedback? If not, GitHub shows that you're still requesting changes.

sanjit-bhat:
I'm getting a weird error on just the `ubuntu-20.04` regression. No clue what's causing this.

```
2022-08-31T23:36:17.6305038Z  Documenting kani v0.9.0 (/home/runner/work/kani/kani/library/kani)
2022-08-31T23:36:18.1704240Z error: couldn't generate documentation: No such file or directory (os error 2)
2022-08-31T23:36:18.1707362Z   |
2022-08-31T23:36:18.1708384Z   = note: failed to create or modify "/home/runner/work/kani/kani/target/doc/kani/slice/struct.AnySlice.html"
2022-08-31T23:36:18.1709077Z 
2022-08-31T23:36:18.1908332Z error: could not document `kani`
2022-08-31T23:36:18.1908970Z 
2022-08-31T23:36:18.1909470Z Caused by:
2022-08-31T23:36:18.1911731Z   process didn't exit successfully: `rustdoc --edition=2021 --crate-type lib --crate-name kani library/kani/src/lib.rs -o /home/runner/work/kani/kani/target/doc --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat -C metadata=6a4d7a21ecff31cd -L dependency=/home/runner/work/kani/kani/target/debug/deps --extern kani_macros=/home/runner/work/kani/kani/target/debug/deps/libkani_macros-2f09c8ff5b9d853c.so '-Aclippy::expect_fun_call' '-Aclippy::or_fun_call' '-Aclippy::new_without_default' --crate-version 0.9.0` (exit status: 1)
2022-08-31T23:36:18.1913297Z warning: build failed, waiting for other jobs to finish...
2022-08-31T23:36:18.7002243Z ##[error]Process completed with exit code 101.
```

adpaco-aws (approved):
Thanks @sanjit-bhat ! Please wait for @celinval to take another look.

celinval (approved):

sanjit-bhat:
@celinval can you merge once CI runs? I don't have write access anymore.

PR #1623, opened 2022-09-01
Add option to enable layout randomization

### LINKED ISSUE #1622: Should kani run the the compiler with -Zrandomize-layout by default?
I just learned that the `-Zrandomize-layout` feature is already out.
Even though kani is bit-precise, the analysis shouldn't be depending on specific layouts. Some bugs might be caught by running the compiler with this flag

giltho:
### Description of changes: 

Randomize layouts by default, but adds a flag to disable that feature.

### Resolved issues:

Resolves #1622 (if accepted)

### Call-outs:

None that I know of

### Testing:

Regression test suite seems to still be working
The feature itself is kind of untestable. I could write a test that depends on the current rustc layout procedure and check that it passes with the flag but doesn't without it (and that's how I tested the change locally). However, that test could break on any sync with rustc

### Checklist


tedinski (commented):
On the one hand, I want a test, on the other hand, its randomization and so would fail sometimes.

Anyone got a suggestion?

danielsn:
```rust
// Or use macros to make this cleaner
struct Foo1 {
  a: u16;
  b: u32;
}

struct Foo2 {
  a: u16;
  b: u32;
}

struct Foo3 {
  a: u16;
  b: u32;
}

pub fn check() {
  a: u32 = kani::any();
  b: u32 = kani::any();
  foo : struct Foo1 {a,b};
  bar : struct Foo2 = transmute(foo);
  assert_eq!(bar.a, a);
  assert_eq!(bar.b, b);
```

kani-driver/src/args.rs
```diff
@@ -179,6 +179,10 @@ pub struct KaniArgs {
     /// Disable CBMC's slice formula which prevents values from being assigned to redundant variables in traces.
     #[structopt(long, hidden_short_help(true), requires("enable-unstable"))]
     pub no_slice_formula: bool,
+
+    /// Randomizes the layout, potentially helping to catch code that depends on a chosen layout and shouldn't.
```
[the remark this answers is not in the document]
giltho: Ah yeah, I initially had a `randomize-layout` flag, and then changed the default 🤦 
I'll fix it

celinval (changes requested):
Can you please explain why that should be the default? I'm not convinced that's the case.

celinval:
The test above would still fail in 50% of the cases. I don't think we should add a test that might fail randomly, even if the test fails in 1% of the cases. Is there a way to control the randomization seed so we can guarantee that the test always fail?

danielsn:
If we do n steps, it will fail in `2^(-n)` times. if we set `n == 64`, we get an erroneous failure probability less than that of the computer being hit by a cosmic ray and giving the wrong result

giltho:
> Can you please explain why that should be the default? I'm not convinced that's the case.

Writing code that depends on the memory layout should be considered a developer error. The Rust documentation repeatedly says that the code developer should never rely on the structure layout (except if specified using `repr(C)`).
If Kani fails with `randomize_layout` but not without, it means that the code depends on something it shouldn't depend on. It could break on the next release of Rustc without any warnings.

celinval:
> > Can you please explain why that should be the default? I'm not convinced that's the case.
> 
> Writing code that depends on the memory layout should be considered a developer error. The Rust documentation repeatedly says that the code developer should never rely on the structure layout (except if specified using `repr(C)`). If Kani fails with `randomize_layout` but not without, it means that the code depends on something it shouldn't depend on. It could break on the next release of Rustc without any warnings.
> 
> Another example would be the layout of pointers. Although it is defined internally, the [stdlib explicitely mentions](https://github.com/rust-lang/rust/blob/9af618b62e91b45d61f41548f24ba60b834e362b/library/core/src/ptr/metadata.rs#L114-L116) that nobody except std should rely on its layout as it could change without warning at any moment:

I totally agree that we should provide an option to validate that the code depends on the structure layout, but this change is not necessarily doing this.

This change is randomizing whether the verification might fail or not if the user depends on the structure layout. This change could cause inconsistent results that are hard to reproduce and hard to debug. I think that we should figure out a way to reproduce the failure before turning this on by default.

celinval:
Here is an example of a bad user experience in my mind.

User runs cargo kani and gets an assertion failure. Now the user runs `cargo kani --visualize` or `cargo kani --enable-unstable --concrete-playback` to figure out why there was a failure in the first place. The second run doesn't fail and leave the user without any clue what went wrong.

zhassan-aws:
For the record, there is a `layout_seed` debug option that allows one to fix the random seed for reproducibility purposes: https://github.com/rust-lang/rust/blob/2e35f954ada0f0c777844dc4fa66684efe90a035/compiler/rustc_session/src/options.rs#L1497

giltho:
> Here is an example of a bad user experience in my mind

Very good point, it's very dangerous for tests to be non-reproducible.

>  For the record, there is a `layout_seed`debug option that allows one to fix the random seed for reproducibility purposes

That's great news! It does mean that we either have to choose a fix one, or keep track of the seed we chose on every run if we want to rerun with the same seed for concrete playback for example. Second option seems unmaintainable. Should I just set `layout_seed = 42`?

celinval:
Wouldn't always running with a fixed seed defeat the purpose of randomizing though?

My suggestion is that we have the randomization off by default. We add an option to randomize the layout with an optional argument that is the seed value. If no seed value is provided, you can use a fixed one

giltho:
There's an argument for a fixed seed which is "yes it's fixed, but it's possibly different from the standard compiler choice".
Also, note that in general, even without any randomization, there's no guarantees that between two runs the compiler is going to choose the same layout.

So here are the two choices I believe:
- 1) What you proposed: randomization is off by default + an option to randomize the layout with an optional seed argument.
- 2) Randomization with a fix seed by default + an option to set the seed or make it random + an option to disable the randomization

I do agree that on paper, the first one makes more sense, but I also believe that making it optional makes it quite probable that it will never be used, so I still think there's an argument to be made for the second one.

I entirely trust your judgement on that so tell me what option you prefer and I'll implement it

celinval:
One more question about this one. What about the std library? I'm assuming this option wouldn't randomize the std library structures, right? Or are we planning to rebuild std?

I rather keep it simple and coherent with the compiler by default. This will also ensure that the concrete playback also works well by default.

When user turns randomization on, we should probably emit a warning or even a comment as part of the test that tells the user which seed was used.

giltho:
> One more question about this one. What about the std library? I'm assuming this option wouldn't randomize the std library structures, right? Or are we planning to rebuild std?

That is not an issue at all, because metadata files keep track of all the information necessary to recompute the layout. So if std was already compiled, its layout shouldn't be randomized at all. If std was compiled with randomized layouts, the information would also be kept and that wouldn't change anything.

> I rather keep it simple and consistent with the compiler by default. This will also ensure that the concrete playback also works well by default.

I think I agree with you. Although in theory the second solution would work with concrete playback in theory, it also adds up to the list of things the Kani team has to document and remember when adding new features, and that's a pain for maintainability.

> For the concrete playback, when user turns randomization on, we should probably emit a warning or even a comment as part of the test that tells the user which seed was used.

Agreed!

giltho:
I looked into how to get the seed, and actually... It looks like the randomization is already deterministic 😆 

image

`-Zrandomize-layout` isn't really random at all, but allows you to change the seed.
That means it's not necessary to emit a warning with the layout seed because if the layout seed is set, the user will have set it themselves

celinval:
1. For the concrete playback, we should still mention that the user have to compile their code with `-Zrandomize-layout`, right?

2- Sorry if I wasn't clear. My point about the std is that the randomization will not be applied to structures that came from the std, unless if we recompile that too.

giltho:
1. True! It's a bit annoying that the user will not be able to use the LSP interface to rerun the tests, since it wouldn't set that param correctly :(
2. Ah yes true. I guess that's ok

giltho:
Fixed according to discussion and rebased on main

kani-driver/src/args.rs
```diff
@@ -179,6 +179,16 @@ pub struct KaniArgs {
...
+
+    /// Randomize the layout of structures. This option can help catching code that relies on
+    /// a specific layout chosen by the compiler that is not guaranteed to be stable in the future.
+    /// See the `-Zrandomize-layout` argument of the rust compiler.
+    #[structopt(long)]
+    pub randomize_layout: bool,
+
+    /// Specify the layout seed for the `--randomize-layout` option.
+    #[structopt(long, requires("randomize-layout"))]
+    pub layout_seed: Option<u64>,
```
celinval, line 191: Can you join these arguments together?

kani-driver/src/args.rs
```diff
@@ -335,6 +345,22 @@ impl KaniArgs {
             self.cbmc_args.iter().any(|s| s.to_str().unwrap().starts_with("--unwind"));
         let natives_unwind = self.default_unwind.is_some() || self.unwind.is_some();
 
+        if self.randomize_layout && self.concrete_playback.is_some() {
+            let random_seed = if let Some(seed) = self.layout_seed {
+                format!(" -Zlayout-seed={}", seed)
```
celinval, line 350: nit: add space here after `-Z` and below: `-Z layout-seed={}`.

tests/kani/LayoutRandomization/should_fail.rs
```diff
@@ -0,0 +1,73 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Kani now has -Zrandomize-layout activated by default.
+// This test only succeeds on the layout of a structure being
+// decided in a deterministic way.
+// With randomization, there is a very very high probability that it fails (1 - (1/2^64))
+
+// kani-flags: --randomize-layout
```
celinval, line 9: Can you simplify this test and just use a specific seed that ensures the test works as expected?

tests/kani/LayoutRandomization/should_fail.rs
```diff
@@ -0,0 +1,73 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Kani now has -Zrandomize-layout activated by default.
```
celinval, line 4: Please update this comment.

celinval (commented):
Thank you for the changes!

I'm debating about one thing, should we bother adding the `--enable-unstable` for now? @danielsn?

tests/kani/LayoutRandomization/should_fail.rs
giltho: I'm not 100% sure that would work
The seed depends on `tcx.def_path_hash`: is that hash going to be the same for every version of the compiler and on every machine? If not, choosing a seed that works right now on my machine is not enough..

kani-driver/src/args.rs
giltho: Do you mean make 1 argument out of them or just remove the space in between?

celinval: Make 1 argument out of them.

tests/kani/LayoutRandomization/should_fail.rs
celinval: That's fair. Too bad though.

kani-driver/src/args.rs
giltho: Not sure what type it should take then
`Option<Option<u64>>`?

giltho: Or even how the parser would behave
`--randomize-layout <seed>`, with no seed for default?
but then if one writes `--randomize-layout --other-option` the other option will be parsed as the seed and probably fail.
Otherwise `--randomize-layout <default | seed>`, but the default version is a bit verbose

danielsn:
I'm not sure we need `--enable-unstable`. What would our criteria be for marking it more stable than it is now?

> Thank you for the changes!
> 
> I'm debating about one thing, should we bother adding the `--enable-unstable` for now? @danielsn?

kani-driver/src/args.rs
celinval: I would expect the parser to be smart enough to correctly parse `--randomize-layout --other-option`.

I think you are correct that with today's implementation, we cannot omit `default` though. @sanjit-bhat bumped into a similar problem with the concrete playback. Maybe we should consider moving to clap completely.

tedinski: Indeed it should be `Option<Option<T>>` and you can find that documented here: https://docs.rs/structopt/latest/structopt/#type-magic

I'm not sure if it requires the `=`  or nor, but it's at least documented that way. (i.e. `--randomize-layout=seed`)

giltho: Cool that also makes it nicer because that's the interface I used to pass a single argument to `format_unit_test` 
image
So I'll be able to simply pass the config directly

giltho:
I think this is ready to be merged!

kani-driver/src/concrete_playback.rs
```diff
@@ -251,11 +259,23 @@ fn format_unit_test(harness_name: &str, concrete_vals: &[ConcreteVal]) -> UnitTe
...
+            seed,
+        ),
+    };
+
     #[rustfmt::skip]
     let concrete_playback = format!(
 "#[test]
 fn {concrete_playback_func_name}() {{
-    let concrete_vals: Vec<Vec<u8>> = vec![
+    {randomize_layout_message}let concrete_vals: Vec<Vec<u8>> = vec![
```
sanjit-bhat, line 278: For code readability, this should probably be split into:
```rust
{randomize_layout_message}
let concrete_vals:
```

giltho: The code would indeed be more readable but it would result in an empty blank line when the option isn't used
The goal was to leave the current output entirely unchanged in the most common case - no randomisation - but I don't mind adding the blank line either

sanjit-bhat: If you kept the match arms the exact same and added an escape after `randomize_layout_message`, I think it would work:
```rust
#[rustfmt::skip]
    let concrete_playback = format!(
"#[test]
fn {concrete_playback_func_name}() {{
    {randomize_layout_message}\
    let concrete_vals: Vec<Vec<u8>> = vec![
    ...
```

giltho: Oh fair enough!

giltho:
I think this is ready to be merged

giltho:
Ah, this went out of sync :/

tedinski (approved):
LGTM. I'm not sure we can do better than the comment for concrete playback, and I guess with it off by default it won't usually affect that.

I do sort of wish we could turn it on by default, but I understand why not...

PR #1648, opened 2022-09-07
Enable `--list-harnesses` flag

### LINKED ISSUE #1612: Create `--list-harnesses`API
The VSCode extension needs a `--list-harnesses` flag from Kani which gives a list of the metadata on all present harnesses in a crate. While this API is intended to be consumed by the VSCode extension, there could be other uses for it too.

jaisnan:
### Description of changes: 

Kani contains the harness metadata in a JSON file but it is only read temporarily before being deleted. The only way to retain this information was through the `--keep-temps` flag which stores these temporary files but is a bulky operation. The flag now however, does not start the verification process. The `--list-harnesses` flag returns a JSON string which can be consumed as an API by any receiver. 

The output with the flag looks like this - 
$cargo kani --enable-unstable --list-harnesses

``` JSON
[
  {
    "pretty_name": "check_estimate_size_2",
    "mangled_name": "check_estimate_size_2",
    "original_file": "../src/lib.rs",
    "original_start_line": 39,
    "original_end_line": 42,
    "unwind_value": null
  },
  {
    "pretty_name": "check_estimate_size",
    "mangled_name": "check_estimate_size",
    "original_file": "..src/lib.rs",
    "original_start_line": 30,
    "original_end_line": 33,
    "unwind_value": null
  }
]

```
### Resolved issues:

Resolves #1612 

### Call-outs:

The flag only runs with cargo kani and not the single script kani because the cargo kani output would always be a superset of the kani output. i.e running the flag with just kani - 

`kani --list-harnesses test.rs`

would return the regular verification output as if `kani test.rs` were run.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/kani-driver/src/args.rs
+++ b/kani-driver/src/args.rs
@@ -61,7 +61,9 @@ pub struct KaniArgs {
     /// Keep temporary files generated throughout Kani process
     #[structopt(long, hidden_short_help(true))]
     pub keep_temps: bool,
-
+    /// Return list of harnesses as a JSON String
+    #[structopt(long, hidden(true), requires("enable-unstable"))]
+    pub list_harnesses: bool,
     /// Produce full debug information
     #[structopt(long)]
     pub debug: bool,
--- a/kani-driver/src/main.rs
+++ b/kani-driver/src/main.rs
@@ -66,6 +66,11 @@ fn cargokani_main(input_args: Vec<OsString>) -> Result<()> {
 
     let mut failed_harnesses: Vec<&HarnessMetadata> = Vec::new();
 
+    if ctx.args.list_harnesses {
+        ctx.return_list_harnesses(&sorted_harnesses).unwrap();
+        return Ok(());
+    }
+
     for harness in &sorted_harnesses {
         let harness_filename = harness.pretty_name.replace("::", "-");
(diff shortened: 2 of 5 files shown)


celinval:
I would expect `--list-harnesses` to be formatted as a user friendly output though. I would suggest to rename it to be `--print-metadata` or something like that if you want to keep that output.

celinval:
> ### Call-outs:
> 
> The flag only runs with cargo kani and not the single script kani because the cargo kani output would always be a superset of the kani output.

I don't understand what you mean here? What will happen if the user invokes:

```
kani --list-harnesses test.rs
```

You should also clarify that this will not run the verification.

jaisnan:
> > ### Call-outs:
> > The flag only runs with cargo kani and not the single script kani because the cargo kani output would always be a superset of the kani output.
> 
> I don't understand what you mean here? What will happen if the user invokes:
> 
> ```
> kani --list-harnesses test.rs
> ```
> 
> You should also clarify that this will not run the verification.

Makes sense. Will clarify these in the PR.

jaisnan:
PR came directly from a local branch by mistake.

PR #1650, opened 2022-09-07
Enable `--print-harness-metadata` flag

### LINKED ISSUE #1612: Create `--list-harnesses`API

jaisnan:
### Description of changes: 

Kani contains the harness metadata in a JSON file but it is only read temporarily before being deleted. The only way to retain this information was through the `--keep-temps` flag which stores these temporary files but is a bulky operation. The flag now however, does not start the verification process. The `--list-harnesses` flag returns a JSON string which can be consumed as an API by any receiver. 

The output with the flag looks like this - 
$cargo kani --enable-unstable --print-harness-metadata

``` JSON
[
  {
    "pretty_name": "check_estimate_size_2",
    "mangled_name": "check_estimate_size_2",
    "original_file": "../src/lib.rs",
    "original_start_line": 39,
    "original_end_line": 42,
    "unwind_value": null
  },
  {
    "pretty_name": "check_estimate_size",
    "mangled_name": "check_estimate_size",
    "original_file": "..src/lib.rs",
    "original_start_line": 30,
    "original_end_line": 33,
    "unwind_value": null
  }
]

```
### Resolved issues:

Resolves #1612 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/kani-driver/src/args.rs
+++ b/kani-driver/src/args.rs
@@ -63,7 +63,7 @@ pub struct KaniArgs {
     pub keep_temps: bool,
     /// Return list of harnesses as a JSON String
     #[structopt(long, hidden(true), requires("enable-unstable"))]
-    pub list_harnesses: bool,
+    pub print_harness_metadata: bool,
     /// Produce full debug information
     #[structopt(long)]
     pub debug: bool,
--- a/kani-driver/src/args.rs
+++ b/kani-driver/src/args.rs
@@ -61,7 +61,9 @@ pub struct KaniArgs {
     /// Keep temporary files generated throughout Kani process
     #[structopt(long, hidden_short_help(true))]
     pub keep_temps: bool,
-
+    /// Return list of harnesses as a JSON String
+    #[structopt(long, hidden(true), requires("enable-unstable"))]
+    pub list_harnesses: bool,
     /// Produce full debug information
     #[structopt(long)]
     pub debug: bool,
--- a/kani-driver/src/main.rs
+++ b/kani-driver/src/main.rs
@@ -66,6 +66,11 @@ fn cargokani_main(input_args: Vec<OsString>) -> Result<()> {
 
     let mut failed_harnesses: Vec<&HarnessMetadata> = Vec::new();
 
+    if ctx.args.list_harnesses {
+        ctx.return_list_harnesses(&sorted_harnesses).unwrap();
+        return Ok(());
+    }
+
     for harness in &sorted_harnesses {
         let harness_filename = harness.pretty_name.replace("::", "-");
         let report_dir = report_base.join(format!("report-{}", harness_filename));
@@ -192,6 +197,14 @@ impl KaniSession {
             std::process::exit(1);
         }
 
+        Ok(())
+    }
+
+    fn return_list_harnesses(self, sorted_harnesses: &[HarnessMetadata]) -> Result<()> {
+        if !self.args.quiet && !self.args.visualize && sorted_harnesses.len() > 1 {
+            let v: String = serde_json::to_string_pretty(&sorted_harnesses)?;
+            println!("{}", v);
+        }
         Ok(())
     }
 }
--- a/kani-driver/src/main.rs
+++ b/kani-driver/src/main.rs
@@ -66,7 +66,7 @@ fn cargokani_main(input_args: Vec<OsString>) -> Result<()> {
 
     let mut failed_harnesses: Vec<&HarnessMetadata> = Vec::new();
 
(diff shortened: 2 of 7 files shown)


kani-driver/src/main.rs
```diff
@@ -194,6 +204,14 @@ impl KaniSession {
 
         Ok(())
     }
+
+    fn return_list_harnesses(self, sorted_harnesses: &[HarnessMetadata]) -> Result<()> {
```
celinval, line 208: Can you please rename this function and remove the return type. Maybe call it something like `print_metadata`.

kani-driver/src/main.rs
```diff
@@ -66,6 +66,11 @@ fn cargokani_main(input_args: Vec<OsString>) -> Result<()> {
 
     let mut failed_harnesses: Vec<&HarnessMetadata> = Vec::new();
 
+    if ctx.args.print_harness_metadata {
```
celinval, line 69: I would suggest that you move this code to be inside `collect_kani_metadata`. You can print the metadata there.

kani-driver/src/main.rs
```diff
@@ -66,6 +66,11 @@ fn cargokani_main(input_args: Vec<OsString>) -> Result<()> {
...
+        ctx.return_list_harnesses(&sorted_harnesses).unwrap();
```
celinval, line 70: BTW, can you print the entire metadata here?

tests/expected/harness_metadata/expected
```diff
@@ -0,0 +1,18 @@
+[
+  {
+    "pretty_name": "check_estimate_size_2",
```
celinval, line 3: I would suggest that you to remove the spaces to make this less likely to fail.

tests/expected/harness_metadata/expected
```diff
@@ -0,0 +1,18 @@
...
+    "mangled_name": "check_estimate_size_2",
+    "original_file":
+    "original_start_line":
+    "original_end_line":
+    "unwind_value": null
+  },
+  {
+    "pretty_name": "check_estimate_size",
```
celinval, line 11: also, can you add '\' at the end to ensure sequence?

kani-driver/src/args.rs
```diff
@@ -61,7 +61,9 @@ pub struct KaniArgs {
     /// Keep temporary files generated throughout Kani process
     #[structopt(long, hidden_short_help(true))]
     pub keep_temps: bool,
-
+    /// Return list of harnesses as a JSON String
+    #[structopt(long, hidden(true), requires("enable-unstable"))]
```
celinval, line 65: `conflicts_with(quiet)`

celinval (commented):
I would suggest that you make this option to print the metadata but without interrupting the verification flow. You can use --only-codegen if you want to interrupt the flow.

kani-driver/src/main.rs
tedinski: I wouldn't print the whole metadata yet.

What we currently do to "merge" all metadata files into one if not a good idea, it just happens to work right now. (For instance, "assess" will add metadata that, when merged, should not turn into a flat vector, but a hash map of "crate -> unsupported features".)

I recommend starting with just the actually needed information here, because otherwise it'll create complications as we evolve the metadata's structure.

tests/cargo-ui/list-harnesses/expected
```diff
@@ -0,0 +1,18 @@
+[
```
tedinski, line 1: I guess these kinds of tests are fine for now, but please put eliminating them on your todo list.

We should end-to-end test this, not look for specific output in the middle. What we eventually want is not these tests, but tests written against the IDE extension (that also runs as part of Kani's CI).

jaisnan: will open an issue once this is closed. Agreed about testing the pipeline in parts.

celinval: This is a command option in Kani, so it might not be a bad idea to keep it either.

celinval:
> > I would suggest that you make this option to print the metadata but without interrupting the verification flow. You can use --only-codegen if you want to interrupt the flow.
> 
> Now `--only-codegen` does `symtab2gb` which is an unnecessary step for this operation that'd add latency though...

With https://github.com/model-checking/kani/pull/1686 that will be the case no matter what though.

PR #1659, opened 2022-09-08
Add kani::spawn and an executor to the Kani library

### LINKED ISSUE #1504: Implement `kani::spawn`
For asynchronous code, it's important to be able to spawn tasks. Since `tokio::spawn` or something similar will be difficult to support, we should provide our own version that is optimized for Kani. Part of #1393

fzaiser:
### Description of changes: 

This adds an executor (scheduler for async futures) to the Kani library, thus supporting `kani::spawn` as a replacement for `tokio::spawn`.

It also includes `kani::yield_now` which is similar to `tokio::yield_now`.

### Resolved issues:

Resolves #1504 

### Call-outs:

This cannot currently be merged because linking the Kani library is broken (just like the standard library).

### Testing:

* How is this change tested? **Not yet.** Could use a similar test as #1658

### Checklist

### DIFF
--- a/library/kani/src/futures.rs
+++ b/library/kani/src/futures.rs
@@ -15,6 +15,9 @@ use std::{
 /// Whereas a clever executor like `block_on` in `futures` or `tokio` would interact with the OS scheduler
 /// to be woken up when a resource becomes available, this is not supported by Kani.
 /// As a consequence, this function completely ignores the waker infrastructure and just polls the given future in a busy loop.
+///
+/// Note that spawn is not supported with this function. Use [`spawnable_block_on`] if you need it.
+// TODO: Give an error if spawn is used in the future passed to this function.
 pub fn block_on<T>(mut fut: impl Future<Output = T>) -> T {
     let waker = unsafe { Waker::from_raw(NOOP_RAW_WAKER) };
     let cx = &mut Context::from_waker(&waker);
@@ -41,3 +44,212 @@ const NOOP_RAW_WAKER: RawWaker = {
 
     RawWaker::new(std::ptr::null(), &RawWakerVTable::new(clone_waker, noop, noop, noop))
 };
+
+static mut EXECUTOR: Scheduler = Scheduler::new();
+const MAX_TASKS: usize = 16;
+
+type BoxFuture = Pin<Box<dyn Future<Output = ()> + Sync + 'static>>;
+
+/// Allows to parameterize how the scheduler picks the next task to poll in `spawnable_block_on`
+pub trait SchedulingStrategy {
+    /// Picks the next task to be scheduled whenever the scheduler needs to pick a task to run next, and whether it can be assumed that the picked task is still running
+    ///
+    /// Tasks are numbered `0..num_tasks`.
+    /// For example, if pick_task(4) returns (2, true) than it picked the task with index 2 and allows Kani to `assume` that this task is still running.
+    /// This is useful if the task is chosen nondeterministicall (`kani::any()`) and allows the verifier to discard useless execution branches (such as polling a completed task again).
+    fn pick_task(&mut self, num_tasks: usize) -> (usize, bool);
+}
+
+impl<F: FnMut(usize) -> usize> SchedulingStrategy for F {
+    fn pick_task(&mut self, num_tasks: usize) -> (usize, bool) {
+        (self(num_tasks), false)
+    }
+}
+
+/// Keeps cycling through the tasks in a deterministic order
+#[derive(Default)]
+pub struct RoundRobin {
+    index: usize,
+}
+
+impl SchedulingStrategy for RoundRobin {
+    fn pick_task(&mut self, num_tasks: usize) -> (usize, bool) {
+        self.index = (self.index + 1) % num_tasks;
+        (self.index, false)
+    }
+}
+
+/// Picks the next task nondeterministically
+#[derive(Default)]
+pub struct NondeterministicScheduling;
+
+impl SchedulingStrategy for NondeterministicScheduling {
+    #[cfg(kani)]
+    fn pick_task(&mut self, num_tasks: usize) -> (usize, bool) {
+        let index = crate::any();
+        crate::assume(index < num_tasks);
+        (index, true)
+    }
+
+    #[cfg(not(kani))]
+    fn pick_task(&mut self, _num_tasks: usize) -> (usize, bool) {
+        panic!("Nondeterministic scheduling is only available when running Kani.")
+    }
+}
+
+/// A restricted form of nondeterministic scheduling to have some fairness.
+///
+/// Each task has a counter that is increased when it is scheduled.
+/// If a task has reached a provided limit, it cannot be scheduled anymore until all other tasks have reached the limit too,
+/// at which point all the counters are reset to zero.
+pub struct NondetFairScheduling {
+    counters: [u8; MAX_TASKS],
+    limit: u8,
+}
+
+impl NondetFairScheduling {
+    pub fn new(limit: u8) -> Self {
+        Self { counters: [limit; MAX_TASKS], limit }
+    }
+}
+
+impl SchedulingPlan for NondetFairScheduling {
+    #[cfg(kani)]
+    fn pick_task(&mut self, num_tasks: usize) -> (usize, bool) {
+        if self.counters[0..num_tasks] == [0; MAX_TASKS][0..num_tasks] {
+            self.counters = [self.limit; MAX_TASKS];
+        }
+        let index = kani::any();
+        kani::assume(index < num_tasks);
+        kani::assume(self.counters[index] > 0);
+        self.counters[index] -= 1;
+        (index, true)
+    }
+
+    #[cfg(not(kani))]
+    fn pick_task(&mut self, _num_tasks: usize) -> (usize, bool) {
+        panic!("Nondeterministic scheduling is only available when running Kani.")
+    }
+}
+
+pub(crate) struct Scheduler {
+    /// Using a Vec instead of an array makes the runtime jump from 40s to almost 10min if using Vec::with_capacity and leads to out of memory with Vec::new (even with 64 GB RAM).
+    tasks: [Option<BoxFuture>; MAX_TASKS],
+    num_tasks: usize,
+    num_running: usize,
+}
+
+impl Scheduler {
+    /// Creates a scheduler with an empty task list
+    pub(crate) const fn new() -> Scheduler {
+        const INIT: Option<BoxFuture> = None;
+        Scheduler { tasks: [INIT; MAX_TASKS], num_tasks: 0, num_running: 0 }
+    }
+
+    /// Adds a future to the scheduler's task list, returning a JoinHandle
+    #[inline] // to work around linking issue
+    pub(crate) fn spawn<F: Future<Output = ()> + Sync + 'static>(&mut self, fut: F) -> JoinHandle {
+        let index = self.num_tasks;
+        self.tasks[index] = Some(Box::pin(fut));
+        self.num_tasks += 1;
+        self.num_running += 1;
+        JoinHandle { index }
+    }
+
+    /// Runs the scheduler with the given scheduling plan until all tasks have completed
+    #[inline] // to work around linking issue
+    fn run(&mut self, mut scheduling_plan: impl SchedulingStrategy) {
+        let waker = unsafe { Waker::from_raw(NOOP_RAW_WAKER) };
+        let cx = &mut Context::from_waker(&waker);
+        while self.num_running > 0 {
+            let (index, can_assume_running) = scheduling_plan.pick_task(self.num_tasks);
+            let task = &mut self.tasks[index];
+            if let Some(fut) = task.as_mut() {
+                match fut.as_mut().poll(cx) {
+                    std::task::Poll::Ready(()) => {
+                        self.num_running -= 1;
+                        let _prev = std::mem::replace(task, None);
+                    }
+                    std::task::Poll::Pending => (),
+                }
+            } else if can_assume_running {
+                crate::assume(false); // useful so that we can assume that a nondeterministically picked task is still running
+            }
+        }
+    }
+
+    /// Polls the given future and the tasks it may spawn until all of them complete
+    #[inline] // to work around linking issue
+    fn block_on<F: Future<Output = ()> + Sync + 'static>(
+        &mut self,
+        fut: F,
+        scheduling_plan: impl SchedulingStrategy,
+    ) {
+        self.spawn(fut);
+        self.run(scheduling_plan);
+    }
+}
+
+/// Result of spawning a task.
+///
+/// If you `.await` a JoinHandle, this will wait for the spawned task to complete.
+pub struct JoinHandle {
+    index: usize,
+}
+
+impl Future for JoinHandle {
+    type Output = ();
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> std::task::Poll<Self::Output> {
+        if unsafe { EXECUTOR.tasks[self.index].is_some() } {
+            std::task::Poll::Pending
+        } else {
+            cx.waker().wake_by_ref(); // For completeness. But Kani currently ignores wakers.
+            std::task::Poll::Ready(())
+        }
+    }
+}
+
+#[inline] // to work around linking issue
+pub fn spawn<F: Future<Output = ()> + Sync + 'static>(fut: F) -> JoinHandle {
+    unsafe { EXECUTOR.spawn(fut) }
+}
+
+/// Polls the given future and the tasks it may spawn until all of them complete
+///
+/// Contrary to block_on, this allows `spawn`ing other futures
+#[inline] // to work around linking issue
+pub fn spawnable_block_on<F: Future<Output = ()> + Sync + 'static>(
+    fut: F,
+    scheduling_plan: impl SchedulingStrategy,
+) {
+    unsafe {
+        EXECUTOR.block_on(fut, scheduling_plan);
+    }
+}
+
+/// Suspends execution of the current future, to allow the scheduler to poll another future
+///
+/// Specifically, it returns a future that
+#[inline] // to work around linking issue
+pub fn yield_now() -> impl Future<Output = ()> {
+    struct YieldNow {
+        yielded: bool,
+    }
+
+    impl Future for YieldNow {
+        type Output = ();
+
+        fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> std::task::Poll<Self::Output> {
+            if self.yielded {
+                cx.waker().wake_by_ref(); // For completeness. But Kani currently ignores wakers.
+                std::task::Poll::Ready(())
+            } else {
+                self.yielded = true;
+                std::task::Poll::Pending
+            }
+        }
+    }
+
+    YieldNow { yielded: false }
+}
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -14,7 +14,9 @@ pub mod vec;
 pub use arbitrary::Arbitrary;
 #[cfg(feature = "concrete_playback")]
 pub use concrete_playback::concrete_playback_run;
-pub use futures::block_on;
+pub use futures::{
+    block_on, spawn, spawnable_block_on, yield_now, NondeterministicScheduling, RoundRobin,
+};
 
 /// Creates an assumption that will be valid after this statement run. Note that the assumption
 /// will only be applied for paths that follow the assumption. If the assumption doesn't hold, the


library/kani/src/futures.rs
```diff
@@ -41,3 +44,216 @@ const NOOP_RAW_WAKER: RawWaker = {
...
+        (self.index, false)
+    }
+}
+
+/// Picks the next task nondeterministically
+#[derive(Default)]
+pub struct NondeterministicScheduling;
+
+impl SchedulingStrategy for NondeterministicScheduling {
+    #[cfg(kani)]
```
celinval, line 89: I don't think there is any need for this. We only distribute the crate with `kani` enabled. Additionally, `crate::any()` should fail already if this is used outside of Kani.

library/kani/src/futures.rs
```diff
@@ -41,3 +44,216 @@ const NOOP_RAW_WAKER: RawWaker = {
...
+    pub(crate) fn spawn<F: Future<Output = ()> + Sync + 'static>(&mut self, fut: F) -> JoinHandle {
+        let index = self.num_tasks;
+        self.tasks[index] = Some(Box::pin(fut));
+        self.num_tasks += 1;
+        self.num_running += 1;
+        JoinHandle { index }
+    }
+
+    /// Runs the scheduler with the given scheduling plan until all tasks have completed
+    #[inline] // to work around linking issue
```
celinval, line 164: Can we get rid of most of these `[inline]` since you are blocked by the linking issue no matter what?

tests/kani/AsyncAwait/spawn.rs
```diff
@@ -0,0 +1,49 @@
...
+    spawnable_block_on(
+        async move {
+            let x3 = x2.clone();
+            spawn(async move {
+                x3.fetch_add(1, Ordering::Relaxed);
+            });
+            yield_now().await;
+            x2.fetch_add(1, Ordering::Relaxed);
+        },
+        RoundRobin::default(),
```
celinval, line 27: don't you need import this struct?

library/kani/src/futures.rs
```diff
@@ -41,3 +44,216 @@ const NOOP_RAW_WAKER: RawWaker = {
...
     RawWaker::new(std::ptr::null(), &RawWakerVTable::new(clone_waker, noop, noop, noop))
 };
+
+static mut EXECUTOR: Scheduler = Scheduler::new();
+const MAX_TASKS: usize = 16;
+
+type BoxFuture = Pin<Box<dyn Future<Output = ()> + Sync + 'static>>;
+
+/// Allows to parameterize how the scheduler picks the next task to poll in `spawnable_block_on`
+pub trait SchedulingStrategy {
```
celinval, line 54: Can you please improve the documentation? How would a user implement this?

library/kani/src/futures.rs
```diff
@@ -41,3 +44,216 @@ const NOOP_RAW_WAKER: RawWaker = {
...
+    /// Picks the next task to be scheduled whenever the scheduler needs to pick a task to run next, and whether it can be assumed that the picked task is still running
```
celinval, line 55: It's not clear to me what the second parameter means here. What kind of information the scheduler has to make that decision in the first place?

library/kani/src/futures.rs
```diff
@@ -41,3 +44,216 @@ const NOOP_RAW_WAKER: RawWaker = {
 
     RawWaker::new(std::ptr::null(), &RawWakerVTable::new(clone_waker, noop, noop, noop))
 };
+
+static mut EXECUTOR: Scheduler = Scheduler::new();
+const MAX_TASKS: usize = 16;
```
celinval, line 49: Can we make this a parameter of the Scheduler?

library/kani/src/futures.rs
```diff
@@ -41,3 +44,216 @@ const NOOP_RAW_WAKER: RawWaker = {
...
+        self.tasks[index] = Some(Box::pin(fut));
+        self.num_tasks += 1;
+        self.num_running += 1;
+        JoinHandle { index }
+    }
+
+    /// Runs the scheduler with the given scheduling plan until all tasks have completed
+    #[inline] // to work around linking issue
+    fn run(&mut self, mut scheduling_plan: impl SchedulingStrategy) {
+        let waker = unsafe { Waker::from_raw(NOOP_RAW_WAKER) };
```
celinval, line 166: what does this mean? No waker?

celinval (commented):
More tests please! :)

fzaiser:
TODO: add check that the number of tasks is less than `MAX_TASKS`.

tests/kani/AsyncAwait/spawn.rs
fzaiser: You're right, done

library/kani/src/futures.rs
fzaiser: OK, I'll remove the `#[cfg]`s

library/kani/src/futures.rs
fzaiser: I guess we could, but then it would have to be an argument of `spawnable_block_on` as well. Given that anything with more nondeterminism than 16 tasks is unlikely to terminate in a reasonable time frame anyway, I don't think I see the point. Could you elaborate?

library/kani/src/futures.rs
fzaiser: Good point, done

library/kani/src/futures.rs
fzaiser: It is a waker, but it does nothing. That's OK because the scheduler does not rely on wakers. If we implemented wakers as in tokio, that would mean having another queue that slows things down.

library/kani/src/futures.rs
fzaiser: Could you be more specific as to what information is missing? Only one method has to be implemented and its documentation should include all the user needs to know?

library/kani/src/futures.rs
fzaiser: I'm sorry, I'm not quite sure what you're asking here. The second parameter is the number of tasks. Are you talking about the second part of the return value? This is just a flag indicating whether the scheduler can `kani::assume` that the returned task is still running, usually because the task was chosen nondeterministically. This cuts off "useless" paths where a completed task is polled again, and is even necessary to make things terminate, if nondeterminism is involved: if we pick the task nondeterministically, and don't have the restriction to still running tasks, we could poll the same task over and over again. However, for the round robin scheduling strategy, assuming that the picked task is still running is generally not possible, and amounts to `assume(false)`, which is undesirable.

fzaiser:
With the MIR linker, this is no longer blocked! Ready for review @danielsn @celinval

library/kani/src/futures.rs
celinval: Sorry, I was talking about the return value. What does the `bool` part of the return value mean?

library/kani/src/futures.rs
celinval: Sure. This can actually affect the result of the verification, and none of our APIs are reflecting that.

Quick questions to help me understand this:
1- What will happen if by any chance we hit the MAX?
2- Does using a lower MAX help with the verification time?

library/kani/src/futures.rs
celinval: Also, I think the Scheduler is `pub(crate)`, so we shouldn't expect users to know about it in the first place. You should document this trait as to what this does, not how. Something like:

```rust
/// Trait that determines the possible sequence of tasks scheduling for a  harness.
/// 
/// <Expand what this means for the verification task at hand. Avoid giving implementation details>
///
/// <Optional Example>
``

File: library/kani/src/futures.rs (hunk 7 above)
Comment (celinval) [reply library/kani/src/futures.rs:159]: Can we use https://docs.rs/futures/latest/futures/task/fn.noop_waker.html instead?

File: library/kani/src/futures.rs (hunk 5 above)
Comment (fzaiser) [reply library/kani/src/futures.rs:55]: This is just a flag indicating whether the scheduler can `kani::assume` that the returned task is still running. As a rule of thumb: if the scheduling strategy picks the next task nondeterministically (using `kani::any()`), return `true`, otherwise false. If true, the scheduler will then assume that the picked task is still running, which cuts off "useless" paths where a completed task is polled again, and is even necessary to make things terminate, if nondeterminism is involved: if we pick the task nondeterministically, and don't have the restriction to still running tasks, we could poll the same task over and over again. However, for most deterministic scheduling strategies, e.g. the round robin scheduling strategy, assuming that the picked task is still running is generally not possible because if that task has ended, we are saying `assume(false)` and the verification effectively stops (which is undesirable, of course).

File: library/kani/src/futures.rs (hunk 7 above)
Comment (fzaiser) [reply library/kani/src/futures.rs:159]: We could, but it would mean pulling in a dependency just for this one type, which isn't very hard to define ourselves. Also, by defining it ourselves, we retain the option of optimizing it for Kani in the future.

**Comment by fzaiser** [comment]
The latest CI run failed for `spawn.rs`: `CBMC failed with status 137`. Exit code 137 means that it ran out of memory, I think.

@celinval: could this be related to the new linker? Before this PR, all the scheduling code was in the file to be verified and it finished quickly. Now that it is in the library, it takes a long time.

File: library/kani/src/futures.rs (hunk 5 above)
Comment (celinval) [reply library/kani/src/futures.rs:71]: I think you should document that. :)

File: library/kani/src/futures.rs (hunk 6 above)
Comment (celinval) [reply library/kani/src/futures.rs:49]: If lowering may improve the verification time, wouldn't it be helpful for users to be able to set this value?

File: library/kani/src/futures.rs (hunk 7 above)
Comment (celinval) [reply library/kani/src/futures.rs:162]: fair. Thanks

**Comment by celinval** [comment]
Yes,

> The latest CI run failed for `spawn.rs`: `CBMC failed with status 137`. Exit code 137 means that it ran out of memory, I think.
> 
> @celinval: could this be related to the new linker? If the scheduling code is in the same file, like in `manual_spawn.rs`, everything finishes quickly. (#1818 seems to be related?)

Yes, the main purpose of the new linker was to get rid of the undefined symbols from the standard library. Fixing this issue meant that some models got bigger and harder to process.

**Comment by fzaiser** [comment]
> Yes, the main purpose of the new linker was to get rid of the undefined symbols from the standard library. Fixing this issue meant that some models got bigger and harder to process.

@celinval That makes sense. What I don't get is that it works fine if all the scheduling code is in the same file, see [`manual_spawn.rs`](https://github.com/model-checking/kani/blob/main/tests/kani/AsyncAwait/manual_spawn.rs). This file uses the same code that this PR puts into the Kani library, but puts that code directly into the file with the harness. That seems to verify much more quickly, which does not make sense to me. Do you know why?

**Comment by celinval** [comment]
> > Yes, the main purpose of the new linker was to get rid of the undefined symbols from the standard library. Fixing this issue meant that some models got bigger and harder to process.
> 
> @celinval That makes sense. What I don't get is that it works fine if all the scheduling code is in the same file, see [`manual_spawn.rs`](https://github.com/model-checking/kani/blob/main/tests/kani/AsyncAwait/manual_spawn.rs). This file uses the same code that this PR puts into the Kani library, but puts that code directly into the file with the harness. That seems to verify much more quickly, which does not make sense to me. Do you know why?

I see. Let me take a look, but it is possible that moving the file into a different crate changed how rustc optimizes the code.

**Comment by celinval** [comment]
I looked at the issue you mentioned, and I think the difference is due to the second harness in `spawn.rs` (`nondeterministic_schedule`). If you remove it, the time and problem size is basically the same independent on the location of the scheduling code.

The interesting thing is that the time is pretty bad even if you run `--harness deterministic_schedule`. That's probably because we perform reachability analysis for all harnesses at the same time, instead of doing them separately. My guess is that CBMC slicer is not smart enough to prune the closures out of the model due to a more conservative pointer analysis.

You can see a similar effect with the `--legacy-linker` by just moving the scheduling code to `spawn.rs`. If you run `kani --harness deterministic_schedule --legacy-linker` with and without the second harness co-located in `spawn.rs`, the times are drastically different. In my machine, Kani verification succeeds after 70ish seconds if you remove the second harness, but it doesn't finish under 5min if you add the second one.

File: library/kani/src/futures.rs (hunk 5 above)
Comment (fzaiser) [reply library/kani/src/futures.rs:80]: Updated the docs.

**Comment by fzaiser** [comment]
@celinval Thanks for that investigation. I've removed the nondeterministic test for now to make CI (hopefully) pass. I've also converted a `bool` to an enum and improved the documentation.

**Comment by fzaiser** [comment]
Hm, seems like CI still runs out of memory. I'm not sure what to do about that. Any ideas?

**Comment by celinval** [comment]
Can you run it locally and measure the memory consumption so we have an idea on how bad the problem is?

File: tests/kani/AsyncAwait/spawn.rs [hunk 8]
```diff
@@ -0,0 +1,50 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// compile-flags: --edition 2018
+// kani-flags: --enable-unstable --mir-linker
```
celinval, line 5: btw, you don't need these flags any more. The MIR Linker is now the default one.

library/kani/src/futures.rs
fzaiser: Well, as soon as the `Vec` performance improvements land, I'd like to replace this by a vector anyway. This is more flexible and does not require a constant upper bound. Therefore, I'd rather not expose the maximum number of tasks in the API.

tests/kani/AsyncAwait/spawn.rs
fzaiser: A lot of tests still seem to have it. But I'll remove it.

celinval: We haven't had time to remove them yet, but I think we should avoid adding them. We will probably fix the old tests once we remove the `--mir-linker` option which today is a no-op.

fzaiser:
> Can you run it locally and measure the memory consumption so we have an idea on how bad the problem is?

@celinval, I just tried that with the second harness (the nondeterministic one) completely removed, so #1855 does not apply. It OOMed after 1 hour, maximum resident set size was about 15 GB. (My computer has 16 GB RAM, so it took up almost all the available memory.)

Should I open a bug about this? It makes no sense to me that it would take so much longer just because the code is in the Kani library instead of the same file. What do you think? cc @danielsn

tests/kani/AsyncAwait/spawn.rs
fzaiser: A lot of tests still seem to have it. But I'll remove it.

celinval:
> > Can you run it locally and measure the memory consumption so we have an idea on how bad the problem is?
> 
> @celinval, I just tried that with the second harness (the nondeterministic one) completely removed, so #1855 does not apply. It OOMed after 1 hour, maximum resident set size was about 15 GB. (My computer has 16 GB RAM, so it took up almost all the available memory.)
> 
> Should I open a bug about this? It makes no sense to me that it would take so much longer just because the code is in the Kani library instead of the same file. What do you think? cc @danielsn

Sure. You can open a bug.

fzaiser:
Submitted #1876. This PR is **blocked** on this issue.

fzaiser:
The test is passing! Now that vectors are faster, I want to change this to use vectors instead of arrays before merging.

library/kani/src/futures.rs
fzaiser: Now using `Vec` instead of an array, so this is resolved.

fzaiser:
@danielsn @celinval This should be ready for another round of review. Note that we now use vectors instead of arrays because they have gotten fast enough. Two questions to decide before merging:

1. The scheduling strategies `NondeterministicScheduling` and `NondetFairScheduling` are currently untested (commented out) because they take too long. (I stopped them after half an hour and taking 15 GB of RAM.) Should we make them private given that they're untested? I would prefer to leave them in (and maybe hide their documentation), so users can still experiment with them if they absolutely want to.
2. Should we remove `manual_spawn.rs`? Since the two versions now take almost no time (but see #1876), it may be redundant to have both.

celinval:
Our libraries don't usually have much logic into them, so there's not much to test. In these news APIs, I think we should test the scheduler logic. We should either have proofs for them or just concrete tests. An error in the scheduler could result in a soundness issue, right?

fzaiser:
@celinval: There is a proof harness using the scheduler (with the `RoundRobin` strategy) in `spawn.rs`. What additional tests or properties to prove do you have in mind?

The two nondeterministic scheduling strategies can only be tested in Kani proof harnesses, not in unit tests (because they use `kani::any()`). But even simple proof harnesses using them take a long time and lots of RAM. Altogether, this means they should not be included if you want them to be tested. Did I understand this correctly? (Another option to consider would be to include them, but hide their documentation. So users could still experiment with them, but we'd consider them "unstable".)

celinval:
Can you take advantage of the concrete playback infrastructure for injecting concrete values for testing purpose?

I do worry about the scalability. If we can't write simple harnesses with them, what value do we get from adding them to the Kani library at this point?

fzaiser:
That's a good point. I have removed the nondeterministic scheduling strategies. No only the `RoundRobin` strategy is included. (And there's a test for it.)

tautschnig:
> That's a good point. I have removed the nondeterministic scheduling strategies. No only the `RoundRobin` strategy is included. (And there's a test for it.)

But wouldn't the non-deterministic schedules be required for soundness? They might be impractical right now, but perhaps there is a way to make them tractable in future?

celinval:
> > That's a good point. I have removed the nondeterministic scheduling strategies. No only the `RoundRobin` strategy is included. (And there's a test for it.)
> 
> But wouldn't the non-deterministic schedules be required for soundness? They might be impractical right now, but perhaps there is a way to make them tractable in future?

Users can still create non-deterministic schedules and I think we should still keep an implementation in our benchmarks to know when they become practical, so we know it's time to add them.

I just don't think they should be in the Kani library until then. Specially since it cannot be tested at all today.

tests/ui/missing-function/replaced-description/expected
```diff
@@ -1,3 +1,3 @@
 Status: FAILURE\
-Description: "Function `<alloc::string::String as core::clone::Clone>::clone` with missing definition is unreachable"
```
celinval, line 2: Why did this change?

tests/kani/AsyncAwait/spawn.rs
```diff
@@ -0,0 +1,30 @@
...
+//! This file tests the executor and spawn infrastructure from the Kani library.
+
+use std::sync::{
+    atomic::{AtomicI64, Ordering},
+    Arc,
+};
+
+#[kani::proof]
+#[kani::unwind(4)]
+fn deterministic_schedule() {
```
celinval, line 15: This is not necessarily testing that the schedule is deterministic though.

tests/kani/AsyncAwait/spawn.rs
```diff
@@ -0,0 +1,30 @@
...
+    let x = Arc::new(AtomicI64::new(0)); // Surprisingly, Arc verified faster than Rc
+    let x2 = x.clone();
+    kani::spawnable_block_on(
+        async move {
+            let x3 = x2.clone();
+            kani::spawn(async move {
+                x3.fetch_add(1, Ordering::Relaxed);
```
celinval, line 22: If I understood this correctly, this should always runs first. In which case, you could assert that x3 is 0, right?

tests/kani/AsyncAwait/spawn.rs
```diff
@@ -0,0 +1,30 @@
...
+            });
+            kani::yield_now().await;
+            x2.fetch_add(1, Ordering::Relaxed);
```
celinval, line 25: and here you can assert that x2 is 1.

tests/kani/AsyncAwait/manual_spawn.rs
```diff
@@ -60,60 +110,98 @@ pub struct RoundRobin {
 }
 
 impl SchedulingStrategy for RoundRobin {
```
celinval, line 112: can you use the `kani::RoundRobin` instead?

tests/kani/AsyncAwait/manual_spawn.rs
```diff
@@ -2,22 +2,45 @@
...
+/// A very simple executor: it polls the future in a busy loop until completion
+///
+/// This is intended as a drop-in replacement for `futures::block_on`, which Kani cannot handle.
+/// Whereas a clever executor like `block_on` in `futures` or `tokio` would interact with the OS scheduler
+/// to be woken up when a resource becomes available, this is not supported by Kani.
+/// As a consequence, this function completely ignores the waker infrastructure and just polls the given future in a busy loop.
+///
+/// Note that spawn is not supported with this function. Use [`spawnable_block_on`] if you need it.
+// TODO: Give an error if spawn is used in the future passed to this function.
+pub fn block_on<T>(mut fut: impl Future<Output = T>) -> T {
```
celinval, line 30: why not `kani::block_on`?

tests/kani/AsyncAwait/manual_spawn.rs
```diff
@@ -32,25 +55,52 @@ const NOOP_RAW_WAKER: RawWaker = {
...
+/// To ensure some "fairness" in how the tasks are picked, there is [`NondetFairScheduling`].
+/// This is probably what you want when verifying a harness under nondeterministic schedulings.
+///
+/// Nondeterministic scheduling strategies can be very slow to verify because they require Kani to check a large number of permutations of tasks.
+/// So if you want to verify a harness that uses `spawn`, but don't care about concurrency issues, you can simply use a deterministic scheduling strategy,
+/// such as [`RoundRobin`], which polls each task in turn.
+///
+/// Finally, you have the option of providing your own scheduling strategy by implementing this trait.
+/// This can be useful, for example, if you want to verify that things work correctly for a very specific task ordering.
 pub trait SchedulingStrategy {
```
celinval, line 87: `kani::SchedulingStrategy`?

tests/kani/AsyncAwait/manual_spawn.rs
```diff
@@ -201,3 +290,22 @@ fn arc_spawn_deterministic_test() {
     );
     assert_eq!(x.load(Ordering::Relaxed), 2);
 }
+
+// #[kani::proof]
+// #[kani::unwind(4)]
+// fn nondeterministic_schedule() {
```
celinval, line 296: Can you uncomment this and select the harness via kani-flags?

tests/kani/AsyncAwait/manual_spawn.rs
```diff
@@ -160,32 +247,34 @@ pub fn spawnable_block_on<F: Future<Output = ()> + Sync + 'static>(
...
     YieldNow { yielded: false }
 }
 
 #[kani::proof]
 #[kani::unwind(4)]
-fn arc_spawn_deterministic_test() {
+fn deterministic_schedule() {
     let x = Arc::new(AtomicI64::new(0)); // Surprisingly, Arc verified faster than Rc
     let x2 = x.clone();
     spawnable_block_on(
```
celinval, line 280: isn't this duplicating the harness from `spawn.rs`?

tests/kani/AsyncAwait/manual_spawn.rs
celinval, line 296: Can you uncomment this and don't select the harness for now?

Note for self: This would be a great candidate for a regression that tracks intractable problems.

tests/kani/AsyncAwait/manual_spawn.rs
```diff
@@ -201,3 +290,22 @@ fn arc_spawn_deterministic_test() {
...
+// #[kani::proof]
+// #[kani::unwind(4)]
+// fn nondeterministic_schedule() {
+//     let x = Arc::new(AtomicI64::new(0)); // Surprisingly, Arc verified faster than Rc
+//     let x2 = x.clone();
+//     kani::spawnable_block_on(
+//         async move {
+//             let x3 = x2.clone();
+//             kani::spawn(async move {
+//                 x3.fetch_add(1, Ordering::Relaxed);
```
celinval, line 303: Maybe augment this to cover the scenario where this is run first and run second.

tests/kani/AsyncAwait/manual_spawn.rs
```diff
@@ -201,3 +290,22 @@ fn arc_spawn_deterministic_test() {
...
+//             });
+//             kani::yield_now().await;
+//             x2.fetch_add(1, Ordering::Relaxed);
```
celinval, line 306: ditto

rahulku:
Hi folks, what is the status of this PR?

fzaiser:
@rahulku I'll address the comments in the next few days.

tests/ui/missing-function/replaced-description/expected
fzaiser: I believe because the Kani library now uses a vector (for storing the task list) and it just so happens to be checked before the String's clone method.

tests/kani/AsyncAwait/spawn.rs
fzaiser: Now, but it tests the behavior under a deterministic schedule. I'm not sure what you think should be improved.

tests/kani/AsyncAwait/spawn.rs
fzaiser: Well, it depends how exactly the round robin is implemented, i.e. in what order tasks are added to the task list. I'm not sure we want to make any guarantees other than that it's deterministic

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: I thought it would be nice to have a manual implementation (not from the library) in `manual_spawn.rs` because there is a speed difference (#1876). But I can remove the file (since it duplicates code from the Kani library) if you want.

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: see https://github.com/model-checking/kani/pull/1659#discussion_r1104463102

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: see https://github.com/model-checking/kani/pull/1659#discussion_r1104463102

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: see https://github.com/model-checking/kani/pull/1659#discussion_r1104463102

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: I'm not sure what you mean here.

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: Not sure what you mean here.

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: done

fzaiser:
@celinval Thanks for your patience. Please take another look when you have time.

tests/kani/AsyncAwait/spawn.rs
celinval: I want to make sure what we are shipping in the Kani crate follows its spec. There is no test in this PR that actually ensures that the RoundRobin scheduler is in fact RoundRobin.

These tests are internal to the tool, and adding extra checks here doesn't mean we have to always keep the same order. But we ensure that our implementation is correct.

Also, if we do pick one specific order, that should be clear to the user.

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: I did this as well.

tests/kani/AsyncAwait/spawn.rs
fzaiser: Alright, I'll add asserts to check this and change the name of the harness

tests/kani/AsyncAwait/spawn.rs
fzaiser: Done

tests/kani/AsyncAwait/spawn.rs
fzaiser: As per discussion above, added the assert.

tests/kani/AsyncAwait/manual_spawn.rs
celinval: You can add manual tests, but I would avoid a complete duplicate.

fzaiser: I'm not sure what else to test manually, so I'll remove it.

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: removed as discussed above

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: removed as discussed above

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: removed as discussed above

fzaiser:
@celinval I renamed the function `spawnable_block_on` to `block_on_with_spawn` as I think this name is clearer. ("spawnable" sounds like "can be spawned" instead of "can spawn")

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: File is deleted now anyway, so resolving this.

tests/kani/AsyncAwait/manual_spawn.rs
fzaiser: File is deleted now anyway, so resolving this.

fzaiser:
@dsn @celinval friendly reminder :)

celinval:
> @dsn @celinval friendly reminder :)

Sorry about that. Let me take a look

celinval:
Hi @fzaiser, I think this is good to go, except that I think we should add a mechanism to guard this code under a `unstable` flag. Let me just sketch something quick and we can merge this. Thanks!

celinval:
@fzaiser I created https://github.com/model-checking/kani/pull/2281 to get some feedback. Feel free to add comments there, but it should be a very quick change (famous last words).

celinval:
I created https://github.com/model-checking/kani/pull/2373 to add the unstable attribute and I'm planning to create another one to enforce the opt-in model.

Once https://github.com/model-checking/kani/pull/2373 is merged, I believe you can just tag the public APIs from this PR as unstable and we can merge these changes.

celinval:
Hi @fzaiser, I just wanted to let you know that we have merged the unstable API code, so you should be good to go.

fzaiser:
@celinval This should be good to go :) Could you please merge this if you're happy with it?

zhassan-aws:
> I'm not sure what the perf-benchcomp failure is about. Could it be spurious? Any ideas?

Yes, this is most likely spurious due to variance in performance. The `perf-benchcomp` workflow is not blocking for merges.

library/kani/src/futures.rs
celinval: Sounds good. Shouldn't this trait be `pub(crate)` though?

celinval (approved):
Thanks! Can you please add a reason to the unstable attribute? It could be something "recently added" or something like "experimental API". @adpaco-aws any suggestion?

adpaco-aws:
> Thanks! Can you please add a reason to the unstable attribute? It could be something "recently added" or something like "experimental API". @adpaco-aws any suggestion?

I'd suggest "experimental async support/library" or something along those lines.

Also, thanks @fzaiser for persistence on this one!

fzaiser:
@celinval Thanks for reviewing and making the final changes for the merge! 🎉

PR #1660, opened 2022-09-09
Enabled constant prooftest.

YoshikiTakashima:
### Description of changes: 

Changed behavior of the `proptest!` macro to enable proofs. Symbolics are not injected, so only constant proptests will work with kani.

### Resolved issues:

Works towards #1608 

### Call-outs:

- BTreeMap was removed from TestRunner b/c `BTreeMap::new()` required unwinding bounds that were not needed if that was removed. 

### Testing:

* How is this change tested? Trivial proptest under 

### Checklist

### DIFF
--- a/proptest/src/sugar.rs
+++ b/proptest/src/sugar.rs
@@ -262,6 +262,9 @@ macro_rules! proptest {
 /// message includes the point of invocation and the format message. `format`
 /// and `args` may be omitted to simply use the condition itself as the
 /// message.
+///
+/// Kani Adjustment: Use Kani::assume instead of rejecting randomly
+/// generated values.
 #[macro_export]
 macro_rules! prop_assume {
     ($expr:expr) => {
@@ -269,13 +272,7 @@ macro_rules! prop_assume {
     };
 
     ($expr:expr, $fmt:tt $(, $fmt_arg:expr),* $(,)?) => {
-        if !$expr {
-            return ::core::result::Result::Err(
-                $crate::test_runner::TestCaseError::reject(
-                    format!(concat!("{}:{}:{}: ", $fmt),
-                            file!(), line!(), column!()
-                            $(, $fmt_arg)*)));
-        }
+        kani::assume($expr)
     };
 }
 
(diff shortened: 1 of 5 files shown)


proptest/src/sugar.rs
```diff
@@ -734,18 +728,15 @@ macro_rules! prop_compose {
 /// ```
 #[macro_export]
 macro_rules! prop_assert {
-    ($cond:expr) => {
+    ($cond:expr) => { {
```
danielsn, line 731: Why do we need the extra scope?

proptest/src/sugar.rs
```diff
@@ -734,18 +728,15 @@ macro_rules! prop_compose {
...
-            let message = format!($($fmt)*);
-            let message = format!("{} at {}:{}", message, file!(), line!());
-            return ::core::result::Result::Err(
-                $crate::test_runner::TestCaseError::fail(message));
-        }
-    };
+    ($cond:expr, $($fmt:tt)*) => { {
+        let message = format!($($fmt)*);
+        let message = format!("{} at {}:{}", message, file!(), line!());
+        assert!($cond, "{}", message);
```
danielsn, line 738: Should we just put the actual message string here?

proptest/src/sugar.rs
```diff
@@ -952,41 +943,29 @@ macro_rules! proptest_helper {
...
-        let names = $crate::proptest_helper!(@_WRAPSTR ($($parm),*));
-        match runner.run(
-            &$crate::strategy::Strategy::prop_map(
-                $crate::proptest_helper!(@_WRAP ($($strategy)*)),
-                |values| $crate::sugar::NamedArguments(names, values)),
-            $($mod)* |$crate::sugar::NamedArguments(
-                _, $crate::proptest_helper!(@_WRAPPAT ($($parm),*)))|
-            {
-                let _: () = $body;
+        runner.run(
```
danielsn, line 946: Maybe comment each of these changes

danielsn (commented):
Would it make sense to split this into two PRs

proptest/src/sugar.rs
YoshikiTakashima: Going with the actual message. std::fmt is not linked.

proptest/src/sugar.rs
YoshikiTakashima: Done: e89d5bf93c1e030aa700523e2140a2b06f69d598

danielsn (approved):

PR #1661, opened 2022-09-09
Allow specifying the scheduling strategy in #[kani_proof] for async functions

fzaiser:
### Description of changes: 

Instead of having to manually wrap the code in `kani::spawnable_block_on` as in #1659, this PR allows `#[kani::proof]` to be applied to harnesses that use `spawn` as well. For this to happen, the user has to specify a scheduling strategy.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Testing:

* How is this change tested? Additional regression test.

### Checklist


fzaiser:
With #1659 merged, this is no longer blocked! I've rebased and updated this PR, but there seem to be issues with concrete playback that I don't understand (the `script-based-pre exec` tests are failing). Before I try to figure that out, I wanted to check that the changes of this PR are still desired. What do you think, @celinval @adpaco-aws?

fzaiser:
Huh, seems that the tests pass in CI (not sure why they failed locally). In that case, this PR is ready for review.

library/kani_macros/src/lib.rs
```diff
@@ -126,7 +127,29 @@ mod sysroot {
...
+        schedule: Option<syn::Expr>,
+    }
+
+    impl Parse for ProofOptions {
+        fn parse(input: ParseStream) -> syn::Result<Self> {
+            if input.is_empty() {
+                Ok(ProofOptions { schedule: None })
+            } else {
+                let ident = input.parse::<syn::Ident>()?;
+                assert_eq!(
```
celinval, line 140: Can you please turn this into a user friendly error?

You can try using `proc_macro_error`. We use it here:
https://github.com/model-checking/kani/blob/bb53ec2e4eb6937a5d45774d73307d74c74aa82d/library/kani_macros/src/derive.rs#L70

library/kani_macros/src/lib.rs
```diff
@@ -138,9 +161,11 @@ mod sysroot {
             #[kanitool::proof]
         );
 
-        assert!(attr.is_empty(), "#[kani::proof] does not take any arguments currently");
-
         if sig.asyncness.is_none() {
+            assert!(
```
celinval, line 165: Can you please move this check to the parser? Also, please make it a user friendly error.

celinval (commented):
Thanks @fzaiser.

Does verifying an async harnesses require the unstable flag to be passed? Is there any documentation to how an async harness is verified? What happens if no argument is passed?

fzaiser:
> Does verifying an async harnesses require the unstable flag to be passed?

Yes, it requires `-Z async-lib`.

> Is there any documentation to how an async harness is verified?

There is some documentation here: https://model-checking.github.io/kani/crates/doc/kani/attr.proof.html. I just added additional information regarding `spawn` to the documentation of the `proof` macro in this PR.

> What happens if no argument is passed?

If no argument is passed, it uses `block_on` (instead of `block_on_with_spawn`). If the harness code tries to call `kani::spawn`, it will hit this [panic here](https://github.com/model-checking/kani/blob/06c0fbe60d793f8b47fadc5fc5a83932e96f2389/library/kani/src/futures.rs#L187). I changed this from an `.expect()` call to an explicit `panic!()` because Kani provides better location information that way. So verification will fail with the following output:

```
Check 53: kani::spawn::<[async block@tests/kani/AsyncAwait/spawn.rs:20:30: 22:6]>.assertion.1
         - Status: FAILURE
         - Description: "This is a placeholder message; Kani doesn't support message formatted at runtime"
         - Location: library/kani/src/futures.rs:189:13 in function kani::spawn::<[async block@tests/kani/AsyncAwait/spawn.rs:20:30: 22:6]>
```

library/kani_macros/src/lib.rs
fzaiser: The options parser does not know what function `#[proof]` is applied to, so I cannot move it there, but made it more user-friendly as requested.

celinval (approved):
Can you please add the following ui tests please?
  - Wrong attribute argument
  - No unstable flag provided

 Thanks!

PR #1666, opened 2022-09-12
add an uninitialized memory access fixme test

giltho:
### Description of changes: 

Added `tests/kani/UninitializedMemory/fixme_main.rs`

### DIFF
--- a/tests/kani/UninitializedMemory/fixme_main.rs
+++ b/tests/kani/UninitializedMemory/fixme_main.rs
@@ -1,10 +1,11 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-// At the moment, Kani is unable to detect the usage of uninitialized memory
-// See https://github.com/model-checking/kani/issues/920
-
 // kani-verify-fail
+
+// This tests demonstrates that Kani is unable to detect reads from
+// uninitialized memory, which is considered undefined behavior.
+// More details in https://github.com/model-checking/kani/issues/920
 
 #[kani::proof]
 fn main() {
--- a/tests/kani/UninitializedMemory/fixme_main.rs
+++ b/tests/kani/UninitializedMemory/fixme_main.rs
@@ -1,0 +1,14 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// At the moment, Kani is unable to detect the usage of uninitialized memory
+// See https://github.com/model-checking/kani/issues/920
+
+#[kani::proof]
+fn main() {
+    let mut v: Vec<u8> = Vec::with_capacity(8);
+    unsafe {
+        v.set_len(3);
+    }
+    let _b = v[0]; //< reading uninitialized memory
+}
--- a/tests/kani/UninitializedMemory/fixme_main.rs
+++ b/tests/kani/UninitializedMemory/fixme_main.rs
@@ -11,4 +11,5 @@ fn main() {
         v.set_len(3);
     }
     let _b = v[0]; //< reading uninitialized memory
+    assert(false);
 }
--- a/tests/kani/UninitializedMemory/fixme_main.rs
+++ b/tests/kani/UninitializedMemory/fixme_main.rs
@@ -4,6 +4,8 @@
 // At the moment, Kani is unable to detect the usage of uninitialized memory
 // See https://github.com/model-checking/kani/issues/920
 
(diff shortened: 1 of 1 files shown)


giltho:
Not sure why the CI fails while it passes on my machine :(

zhassan-aws:
Since the test has `fixme` in its name, `compiletest` would expect that this test fails, and if it doesn't, regressions will fail (which is why the `kani-fixme` suite failed).

adpaco-aws:
This should fail to verify, right? Then, you can `// kani-verify-fail` after the copyright to indicate that.

giltho:
It's a fixme because this test passes but shouldn't
On my local machine, I first added `// kani-verify-fail` and the suite would fail saying "Expected to fail but it didn't"
So I remove the `// kani-verify-fail` and now it works locally, but it fails in the CI 🤔

giltho:
Aaaaaah, my bad...
It fails with or without `kani-verify-fail`. It's just that the second time I ran it, the test was ignored. It fails in both cases when I use the `--force-rerun` flag.
I'll add an `assert(false)` to force it to fail

tests/kani/UninitializedMemory/fixme_main.rs
```diff
@@ -0,0 +1,15 @@
...
+// See https://github.com/model-checking/kani/issues/920
+
+#[kani::proof]
+fn main() {
+    let mut v: Vec<u8> = Vec::with_capacity(8);
+    unsafe {
+        v.set_len(3);
+    }
+    let _b = v[0]; //< reading uninitialized memory
+    assert(false);
```
zhassan-aws, line 14: Please add a comment to explain the point of this `assert(false)`.

celinval:
Is this really a fixme test though? This feels more like a test for a feature request.

tests/kani/UninitializedMemory/fixme_main.rs
danielsn: The `assert(false)`, makes no sense to me. This test will fail whether or not we fix the issue. Which seems to me to break the point of a "fixme" test

giltho: that's true
I don't know how to make that test work then

giltho:
Is there a way to write such a test? It is a UB that is not being detected by Kani, and that is in the work.
Maybe you're right this PR should be closed

zhassan-aws:
The problem is that `compiletest` provides a mechanism for marking tests that are failing but should pass, but AFAIK, it doesn't provide a mechanism for marking tests that are passing, but should fail. The question is, what should we do for those tests? If the file is named with a "fixme", the regressions will fail since the test is passing. I think that's why @giltho added the assert false so that regressions pass.

It seems most intuitive to name those tests with "fixme" so that it's clear that they're not working as expected. Perhaps we can add a new designation other than "fixme" for those tests?

danielsn:
fixme_fail?

adpaco-aws:
> The problem is that `compiletest` provides a mechanism for marking tests that are failing but should pass, but AFAIK, it doesn't provide a mechanism for marking tests that are passing, but should fail.

I think `compiletest` can do the latter if:
 1. the assert is removed.
 2. `// kani-verify-fail` is added
 3. `fixme` is added to the name

Then run the **regression script**, not `kani` or `cargo-kani`. `// kani-verify-fail` only has effect in `compiletest`.

giltho:
@adpaco-aws that's exactly what I did at first and the test appeared as "FAILED"...

giltho:
Oh god... I just realised what happened.
There was a interleaved alternation of modifying the `kani-fixme/...` file and the `kani/...` file, only the second one should be modified... I ended up getting inconsistent results and didn't understand why 🤦 
My bad, it's fixed now

giltho:
Doesn't solve @celinval's question: should this be added to the repo?

tests/kani/UninitializedMemory/fixme_main.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// At the moment, Kani is unable to detect the usage of uninitialized memory
```
adpaco-aws, line 4: A small rewording suggestion.

> This tests demonstrates that Kani is unable to detect reads from uninitialized memory, which is considered undefined behavior. More details in <link>.

tedinski:
Decision came down against fixme for this, but recording as part of the feature request issue: #1666

PR #1684, opened 2022-09-14
Refactor harness running

tedinski:
### Description of changes: 

This refactoring prepares the code for adding a parallel harness runner by creating a single piece of code that runs the harnesses.

Along the way, I discovered a horrible bug where plain `kani` would pick the same filename (`foo.c`) for every proof harness, overwriting each other. It's questionable whether we even want per-harness `--gen-c` or if we should apply that to the general linked binary instead. But this is the minimal fix for our inconsistent (`cargo kani` was fine) and buggy (which harness wins? whichever writes last!) behavior. That we were apparently "testing" for with `check-outputs.sh`. Hmm

### Testing:

* Is this a refactor change?

### Checklist


kani-driver/src/cbmc_output_parser.rs
```diff
@@ -3,7 +3,7 @@
 
 //! Module for parsing CBMC's JSON output. In general, this output follows
 //! the structure:
-//! ```
+//! ```text
```
celinval, line 6: json?

kani-driver/src/harness_runner.rs
```diff
@@ -0,0 +1,124 @@
...
+                results.successes.push(wrapped_result);
+            }
+        }
+
+        Ok(results)
+    }
+}
+
+impl KaniSession {
+    pub(crate) fn check_harness(
```
celinval, line 76: Shouldn't this be part of the new HarnessRunner?

celinval (commented):
Can you please clarify what you plans are? I can't tell from this PR how you are planning to parallelize each execution.

kani-driver/src/cbmc_output_parser.rs
tedinski: Not valid json.

Only important thing here is making rustdoc/rust-analyzer not thing this is a rust doctest

kani-driver/src/harness_runner.rs
tedinski: Probably, but there's also going to be some other changes coming here eventually.

I'm not sure I'll get to this as part of the initial parallel runner, but I will eventually "lift" all the output code outwards, so that we don't have e.g. interleaving problems, and we can more deeply inspect the actual results. (For example, I intend `VerificationStatus::Failure` to start carrying information about why it failed. Both for "assess" and to lift the "printing" out of the "running" to eventually better solve the interleaving output problem, which I might just initially solve with an annoying mutex.)

celinval:
I like the idea of breaking down KaniSession object, and we will likely need that to run things in parallel, but HarnessRunnerSession seems a bit random here. HarnessRunnerSession still invokes KaniSession. I am assuming you want one runner per harness, don't you?

kani-driver/src/cbmc_output_parser.rs
celinval: fair

kani-driver/src/harness_runner.rs
celinval: Interesting... how are you planning to display the verification status for the user? I'm thinking here about the case where there are 2 harnesses stuck due to unlimited unwinding. We won't be able to print CBMC messages on the fly anymore, right?

tedinski:
> I am assuming you want one runner per harness, don't you?

Nope, it's about running them all.

Really the only major need for this object is that the runner function starts to take way too many arguments.

kani-driver/src/harness_runner.rs
tedinski: Right. When running in parallel we wouldn't be able to use that approach to display unwinding status anymore.

Parallel running will initially be an unstable feature that will require `--enable-unstable`. Making it work well will take a lot more work (that's frankly already necessary: see issues complaining about cbmc's output, and the need for better communication to users about unbounded unwinding.)

danielsn:
Ideally, we'd probably want an `async run_harness()` function to allow easy parallelization.  We shouldn't even need threads. What if instead of putting this data inside a big object, we factored it out into a function that could be eventually called async?

tedinski:
> Ideally, we'd probably want an `async run_harness()` function to allow easy parallelization.

Indeed! Threads are a simpler starting point, however.

(Besides dependencies, there's also a problem of _limiting concurrency_. This is actually pretty complex for async code, but trivial (to do in a simple/dumb way) for threads. That said, in the longer run we'll absolutely want more complex concurrency controls than the dumb "one blocking thread" approach.)

tedinski:
> instead of putting this data inside a big object

I'm a bit curious about this comment, though. It's basically just an object containing things you'd have to pass as parameters. Splitting it out to a struct as "background information" versus the harness list ("the thing to actually operate on") is just a cleaner way of writing this code compared to passing in an extra 5 parameters. There's nothing about this that seems big or complex to me.

tedinski:
Also, just want to emphasize from a code review perspective here... I've had no requests for changes here. If this isn't mergeable, I have no idea why.

kani-driver/src/harness_runner.rs
```diff
@@ -0,0 +1,124 @@
...
+
+use anyhow::Result;
+use kani_metadata::HarnessMetadata;
+use std::path::{Path, PathBuf};
+
+use crate::call_cbmc::VerificationStatus;
+use crate::session::KaniSession;
+use crate::util::specialized_harness_name;
+
+pub(crate) struct HarnessRunnerSession<'sess> {
```
celinval, line 12: please add comments to the new structs and methods in this file.

kani-driver/src/harness_runner.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub sess: &'sess KaniSession,
+    /// The build CBMC goto binary for the "whole program" (will be specialized to each proof harness)
+    pub linked_obj: &'sess Path,
+    /// The directory we should output cbmc-viewer reports to
+    pub report_base: &'sess Path,
+    /// A behavior difference between `kani` and `cargo kani`: `cargo kani` never deleted the specialized goto binaries
+    pub retain_specialized_harnesses: bool,
+
+    /// The collection of symtabs that went into the goto binary
+    /// (TODO: this is only for --gen-c, which possibly should not be done here?)
```
celinval, line 23: issue for this TODO.

kani-driver/src/harness_runner.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub result: VerificationStatus,
+}
+
+#[derive(Default)]
+pub(crate) struct HarnessResults<'sess> {
+    pub successes: Vec<HarnessResult<'sess>>,
+    pub failures: Vec<HarnessResult<'sess>>,
+}
+
+impl<'sess> HarnessRunnerSession<'sess> {
```
celinval, line 38: nit: Can we call this `HarnessesRunner` instead?

kani-driver/src/harness_runner.rs
```diff
@@ -0,0 +1,124 @@
...
+            }
+            self.sess.run_goto_instrument(
+                self.linked_obj,
+                &specialized_obj,
+                self.symtabs,
+                &harness.mangled_name,
+            )?;
+
+            let result = self.sess.check_harness(&specialized_obj, &report_dir, harness)?;
+            let wrapped_result = HarnessResult { harness, result };
```
celinval, line 63: Just a suggestion, I think it would be much cleaner to just return a vector of results here. You could replace the HarnessResult structure and all this code here by one line inside `print_final_summary` that filters the failed results.

kani-driver/src/util.rs
```diff
@@ -154,4 +142,21 @@ mod tests {
         c1.env("PARAM", "VALUE");
         assert_eq!(render_command(&c1), OsString::from("PARAM=\"VALUE\" a b \"/c d/\""));
     }
+
+    #[test]
+    fn check_specialized_harness_name() {
```
celinval, line 147: nice!

kani-driver/src/main.rs
```diff
@@ -61,30 +59,20 @@ fn cargokani_main(input_args: Vec<OsString>) -> Result<()> {
...
-        let result = ctx.check_harness(&specialized_obj, &report_dir, harness)?;
-        if result == VerificationStatus::Failure {
-            failed_harnesses.push(harness);
-        }
-    }
+    let results = runner.check_all_harnesses(&harnesses)?;
 
-    ctx.inform_if_no_failed(&failed_harnesses);
-    ctx.print_final_summary(&sorted_harnesses, &failed_harnesses)
+    ctx.inform_if_no_failed(&results);
```
celinval, line 74: Still not clear to me why this is part of `KaniSession` and not `HarnessRunner`.

celinval: Actually, is there a reason to call these two functions here instead of putting them inside `check_all_harnesses`?

celinval (commented):
I'm not a big fan on how this code is broken down right now. I asked about your future plans to understand the decisions made in this PR.

It's hard for me to tell what the `HarnessRunnerSession` responsibility is. What should go here in the future? Or are adding this just for the purpose of not providing more arguments to the `check_all_harnesses` function?

Another question I have is what should be an argument to `check_all_harnesses`, and what should be a member of `HarnessRunnerSession`. Why `symtabs` is a member and `harnesses` is an argument?

I guess we can live with this as is for now, but I was just trying to figure out if there was a better way to move forward before adding more comments.

kani-driver/src/main.rs
tedinski: `print_final_results` consumes the kani session (if there were failures it needs to drop it, and then exit the process with appropriate error code.)

So that's why that needs to be a separate function called by main. `assess` for instance, will do other things after a `HarnessRunner` completes.

As for `check_harness`, it does not use any of the data in `HarnessRunner`, and that's why it's not a function under that structure. Every argument it gets, for instance, is specialized per-harness, not something that's run-wide. (`report_dir` for instance is not `report_base`).

kani-driver/src/harness_runner.rs
tedinski: I pursued this and it also lead to removing the `inform_if_no_failed` function (by merging it into `print_final_summary`) as well.

celinval (approved):

PR #1691, opened 2022-09-16
Fix performance regressions script

### LINKED ISSUE #1688: Performance regressions broken in CI
Performance regressions have been failing for quite some time, at least since https://github.com/model-checking/kani/commit/e151d2232b33f799953a9585c1fc39cf77942d09:

https://github.com/model-checking/kani/actions/runs/2786241652

But they're not causing CI to fail!

### LINKED ISSUE #1689: Performance regression failure does not cause CI to fail
They've been failing since at least https://github.com/model-checking/kani/commit/e151d2232b33f799953a9585c1fc39cf77942d09. Need to make sure their failure causes CI to fail.

zhassan-aws:
### Description of changes: 

Performance regressions are currently broken in two ways:
1. They build Kani using `cargo build --release`, which causes two `kani-driver` binaries to exist under the target directory:
```
target/debug/kani-driver
target/release/kani-driver
```
which in turn causes the kani/cargo-kani script to fail, since they explicitly check that only binary is found:
```bash
[ ${#KANI_CANDIDATES[@]} -ne 1 ]]
```
2. The `kani-perf.sh` script always exits with 0 status even if the regressions failed!

This PR fixes both issues.

### Resolved issues:

Resolves #1688 
Resolves #1689 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/scripts/kani-perf.sh
+++ b/scripts/kani-perf.sh
@@ -8,8 +8,8 @@ set -o nounset
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 KANI_DIR=$SCRIPT_DIR/..
 
-# Build Kani using release mode.
-cargo build --release
+# Build Kani
+cargo build
 
 PERF_DIR="${KANI_DIR}/tests/perf"
 
@@ -40,3 +40,4 @@ else
   echo "***Kani perf tests failed."
 fi
 echo
+exit $exit_code
--- a/scripts/kani-perf.sh
+++ b/scripts/kani-perf.sh
@@ -8,8 +8,8 @@ set -o nounset
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 KANI_DIR=$SCRIPT_DIR/..
 
-# Build Kani
-cargo build
+# Build Kani using release mode
+cargo build --release
 
 PERF_DIR="${KANI_DIR}/tests/perf"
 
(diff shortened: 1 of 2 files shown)


scripts/kani-perf.sh
```diff
@@ -8,8 +8,8 @@ set -o nounset
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 KANI_DIR=$SCRIPT_DIR/..
 
-# Build Kani using release mode.
-cargo build --release
+# Build Kani
+cargo build
```
celinval, line 12: Shouldn't we use release though?

zhassan-aws: Yes, we should, but our `kani` and `cargo-kani` dev scripts currently do not support having two binaries in the build directory. 

I could add a `cargo clean` before the `cargo build --release` line to make sure other binaries are removed, but this might be unexpected by the script user.

Another alternative is to build Kani with `--release` in the GH action:
https://github.com/model-checking/kani/blob/61c8d1180fc49cce8d730f6ebdfb6d6198fd610b/.github/workflows/kani.yml#L67

But this may fail locally if someone tries to run the `kani-perf.sh` script directly.

Thoughts?

celinval: Maybe set the CI to build with release for now. I might fix this issue in my next change.

zhassan-aws: Sounds good. Done.

celinval (approved):
I usually prefer failing fast:
```bash
set -o errexit
```
But not a blocker

zhassan-aws:
>I usually prefer failing fast:
> set -o errexit

I deliberately didn't put that there because I want to make sure all the overlay files that are copied onto the s2n-quic repo before running are deleted even if the `cargo run` fails, so that there are no leftover files.

PR #1708, opened 2022-09-21
Use `#kani_comment` instead of `#comment` for deinit

### LINKED ISSUE #1575: Clean GotoC AST annotations
Following #1469, it looks like a viable solution for supporting alternative backends is to comment the produced Irep to allow for a different interpretation than CBMC.

At the moment, the choice has been made to add a "#comment" named sub to the Irep itself. For example, in the case of `Deinit`, the assignment statement corresponding to `lvalue = Nondet` has this name sub. However, comments are usually attached to the `location` member of the irep.

The reason the comment was attached to the irep itself is simplicity: `Stmt::to_irep`'s job is to call `Location::to_irep` and `StmtBody::to_irep`, and then aggregating the results. Therefore, if we want `StmtBody::to_irep` to produce a comment that is then added to the Location's irep, we need `StmtBody::to_irep` to return two things, instead of one. That would mean refactoring the function to change every return value, for 1 use case.

This issue aims at discussing the best and cleanest way to handle AST comments in the future.

giltho:
### Description of changes: 

Instead of inserting a `comment` field on the Irep for `deinit` and potential future annotations, inserts a `#kani_comment`.
It seems that having a field that doesn't that with a `#` might be dangerous otherwise.

### Resolved issues:

Resolves #1575

### Call-outs:

Kani still outputs some `comment` fields, as it used to before #1469, but these `comment` fields are inside a `#c_location` field, so they will be ignored by transitivity.

### Testing:

This is just a slight change, it should not change anything in kani.

### Checklist


cprover_bindings/src/irep/irep.rs
```diff
@@ -47,8 +47,8 @@ impl Irep {
     /// Adds a `comment` sub to the irep.
     /// Note that there might be comments both on the irep itself and
     /// inside the location sub of the irep.
-    pub fn with_comment<T: Into<InternedString>>(self, c: T) -> Self {
-        self.with_named_sub(IrepId::Comment, Irep::just_string_id(c))
+    pub fn with_kani_comment<T: Into<InternedString>>(self, c: T) -> Self {
```
danielsn, line 50: This ends up with "kani" stuff inside cprover_bindings. Is there a better way?

giltho: Good point.
The only other way is to use a `IrepId::FreeformString`.

tedinski:
Was "comment" ever used? I wonder if we could just change that to `#comment` and call this done...

celinval:
@giltho @tedinski What's left to be done here?

giltho:
Check with the cbmc team that it's ok to just use `#comment` and if so, simply do that

tautschnig:
There is nothing special about '#comment' in CBMC. It's the '#' prefix itself that is special, but beyond that there are no conflicts for this.

giltho:
Actually... `comment`s are used to propagate the `assert` messages to CBMC, and then are given in the CBMC output which is then used by the CBMC_output_parser.
Using `#comment` instead of `comment` for `Kani::Comment.to_string()` entirely breaks the regression test suite.

giltho:
I renamed `KaniComment` into `HiddenComment`, so now
```rust
IrepId::HiddenComment => "#comment",
IremId::Comment => "comment"
```
it's not much better but it does remove mentions of `Kani` from the `cprover_bindings`

tedinski:
Maybe it's just fine as-is then?

giltho:
Yep, any change was breaking the CI anyway 🤔

celinval:
So I'm closing this based on the comments. Please feel free to reopen if you think it is still relevant (and I'm sorry if I closed it prematurely). :)

PR #1715, opened 2022-09-23
Stabilized Performance: improved harnessing

YoshikiTakashima:
### Description of changes: 

This change improves performance by doing the following by removing sources of performance instability in PropProof's harnessing (`crate::test_runner` and macros)

### Resolved issues:

Works towards #1608 

### Call-outs:

This change alters the public API of `crate::test_runner::Config`, and henceforth has the risk of breaking potential customer code. However, crawled data suggests that interaction with this API is mostly done through the use of `$config` in the macros. (`src/sugar.rs`). We turn this off with appropriate adjustments in the macros (Kani was already not using this data) so any customer code that is using `#![proptest_config(...)]` still works. However, if the customer code directly interacts with the 2 deleted parameters of `Config`, then we will fail to type check
- Tracking: #1716
   The mitigation will be to put unwind bounds per function or other more fine-grained units of code so that we can stop the unrolling problem.

Separately, some commits were reverted (see enums). Reviewers should ignore https://github.com/model-checking/kani/pull/1715/commits/129be91649f0218cc83c91a6d588c7ac37b4313a and https://github.com/model-checking/kani/pull/1715/commits/987953c2c4c666e0866b9b53e49da07f36ea9135

### Testing:

* How is this change tested?
  Performance: External test with enum (implementation coming soon)
  Still works with `proptest_config` attributes: 

### Checklist

### DIFF
--- a/proptest/src/arbitrary/_std/env.rs
+++ b/proptest/src/arbitrary/_std/env.rs
@@ -44,6 +44,7 @@ fn jpe() -> JoinPathsError {
 
 // Algorithm from: https://stackoverflow.com/questions/47749164
 #[cfg(any(target_os = "windows", test))]
+#[allow(dead_code)] // Unused b/c of merge: see #1608.
 fn make_utf16_invalid(buf: &mut [u16], p: usize) {
     // Verify that length is non-empty.
     // An empty string is always valid UTF-16.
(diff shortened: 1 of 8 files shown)


YoshikiTakashima:
@danielsn Would you mind taking a look?

danielsn (approved):
Looks good.

PR #1717, opened 2022-09-23
Build custom sysroot with `--always-encode-mir` on

### LINKED ISSUE #241: Failure to cast back to `&str` from `&[u8]`
#235 adds the test `src/test/cbmc/Strings/copy_empty_string_by_intrinsic_fixme.rs`, which has this chunk:

```rust 
        let dest_slice : &[u8] = from_raw_parts(dest, l);
        let dest_as_str : &str= str::from_utf8(dest_slice).unwrap();
        assert!(dest_as_str.len() == l);
```

With failures verification with:
```rust
[...copy_empty_string_by_intrinsic.assertion.2] line 1035 unreachable code: FAILURE
[...copy_empty_string_by_intrinsic.assertion.1] line 1037 a panicking function std::result::unwrap_failed is invoked: FAILURE
[...copy_string.assertion.2] line 28 assertion failed: dest_as_str.len() == l: FAILURE
```

### LINKED ISSUE #1605: Build custom sysroot
Kani currently uses `rustup` sysroot to gather information from the standard library constructs. The artifacts from this `sysroot` do not include the `MIR` for items that have already been compiled to the `std` shared library. We also use a similar method to build `kani`'s custom libraries.

This leaves a gap that cannot be filled by the `kani-compiler`; thus, we are unable to translate these items into `goto-program`.

In order to fulfill this gap, we must build a `sysroot` from scratch. For more details, see https://github.com/model-checking/kani/pull/1600

celinval:
### Description of changes: 

This change builds a custom sysroot for Kani. This new sysroot will contain a "lib/" folder with Kani libraries as well as the standard libraries compiled with `--always-encode-mir`. This enable us to fully traverse the std MIR and fix the missing functions errors.

Other changes to the build were described in the issue here: https://github.com/model-checking/kani/issues/1605#issuecomment-1252972301

### Resolved issues:

Resolves #1605
Resolves #241 

### Related RFC:

Tracking issue #1588 .

### Call-outs:

I'm still fixing all the workflows, but I could use some early feedback.

### Testing:

### Checklist


tools/build-kani/Cargo.toml
```diff
@@ -2,13 +2,15 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
 [package]
-name = "make-kani-release"
-version = "0.1.0"
+name = "build-kani"
+version = "0.10.0"
```
zhassan-aws, line 6: 0.11.0?

tools/build-kani/src/sysroot.rs
```diff
@@ -0,0 +1,184 @@
...
+fn filter_kani_std(cargo_cmd: &mut Child) {
+    let reader = BufReader::new(cargo_cmd.stdout.take().unwrap());
+    for message in Message::parse_stream(reader) {
+        match message.unwrap() {
+            Message::CompilerMessage(msg) => {
+                // Print message as cargo would.
+                println!("{:?}", msg)
+            }
+            Message::CompilerArtifact(artifact) => {
+                // Remote the `rlib` and `rmeta` kept in the deps folder.
```
zhassan-aws, line 148:
```suggestion
                // Remove the `rlib` and `rmeta` kept in the deps folder.
```

zhassan-aws (approved):
The PR changes are quite big! Might help to break this up into multiple PRs (e.g. may want to have a first PR that updates the `make-kani-release` tool).

tools/build-kani/Cargo.toml
celinval: Yeah, I need to update my branch.

celinval:
> The PR changes are quite big! Might help to break this up into multiple PRs (e.g. may want to have a first PR that updates the `make-kani-release` tool).

I would really like to break this PR down, but I'm afraid the regression won't pass if I do.

kani-compiler/src/main.rs
```diff
@@ -79,6 +79,17 @@ fn rustc_gotoc_flags(lib_path: &str) -> Vec<String> {
     args.iter().map(|s| s.to_string()).collect()
 }
 
+/// This function generates all rustc configurations required by regular codegen.
+/// This is because we may use a custom sysroot.
+fn rustc_llvm_flags() -> Vec<String> {
+    // The option below provides a mechanism by which definitions in the
```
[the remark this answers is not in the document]
celinval: Oops. Yeah, that's garbage. :)

kani-compiler/src/main.rs
```diff
@@ -128,13 +137,37 @@ struct KaniCallbacks {}
 /// Use default function implementations.
 impl Callbacks for KaniCallbacks {}
 
+/// The Kani root folder has all binaries inside bin/ and libraries inside lib/.
+/// This folder can also be used as a rustc sysroot.
+fn kani_root() -> PathBuf {
+    match env::current_exe() {
+        Ok(mut exe_path) => {
+            // Current folder (bin/)
```
adpaco-aws, line 145: Isn't there a better way to move to the root folder? This looks like it's only good to go from `bin/` to the root folder.

kani-compiler/src/main.rs
```diff
@@ -170,34 +203,53 @@ fn convert_arg(arg: &OsStr) -> String {
...
     // This is a *compile-time* constant, not a dynamic lookup at runtime, so this is reliable.
     let toolchain = env!("RUSTUP_TOOLCHAIN");
 
+    // We use the home crate to do a *runtime* determination of where rustup toolchains live
+    let rustup = home::rustup_home().expect("Couldn't find RUSTUP_HOME");
+    let path = rustup.join("toolchains").join(toolchain);
+
+    if !path.exists() {
+        panic!("Couldn't find Kani Rust toolchain {}. Tried: {}", toolchain, path.display());
+    }
```
adpaco-aws, line 228: Nice!

tests/kani/Strings/copy_empty_string_by_intrinsic.rs
```diff
@@ -21,11 +29,18 @@ fn copy_string(s: &str, l: usize) {
...
+        // Back to str, check length
+        let dest_slice: &[u8] = from_raw_parts(dest, l);
+        let dest_as_str: &str = str::from_utf8(dest_slice).unwrap();
+        assert!(dest_as_str.len() == l);
     }
 }
 
 #[kani::proof]
 fn main() {
+    // Verification fails for both of these cases.
```
adpaco-aws, line 43: Looks like these are passing now. Is this comment needed?

tools/build-kani/src/sysroot.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    let kani_rlib_folder = path_buf!(target_folder, "debug");
+    cp_files(&kani_rlib_folder, &legacy_lib, &is_rlib).unwrap();
+}
+
+fn is_rlib(path: &Path) -> bool {
+    path.is_file() && String::from(path.file_name().unwrap().to_string_lossy()).ends_with(".rlib")
+}
+
+/// Extra arguments to be given to "cargo build" while building Kani's binaries.
```
adpaco-aws, line 203: "cargo build" -> \`cargo build\`?

adpaco-aws (approved):

tests/kani/Strings/copy_empty_string_by_intrinsic.rs
celinval: Indeed. I can update the comments.

kani-compiler/src/main.rs
celinval: Unfortunately, this does rely on the structure of the root folder and the location of `kani-compiler`. This is the logic we already used for bundled Kani and it's very similar to `rustc` logic. One thing I just noticed that `rustc` does is to canonicalize the path to follow symbolic links. I'll add this here.

As a fallback plan, I kept the command argument to set sysroot and the library path in case we need to override those values.

Do you have any other suggestion?

adpaco-aws: Not really, thanks for the detailed explanation!

kani-compiler/src/main.rs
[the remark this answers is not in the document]
celinval: Also, I just noticed that I don't need this function. Long story short, I changed the code to use the rustup toolchain for now to compile things like the build script.

PR #1723, opened 2022-09-27
RFC for Function and Method Stubbing

### LINKED ISSUE #1695: RFC: Function stubbing
This is a tracking issue for the RFC `function_stubbing`.

The goal is to allow users to specify that some functions should be replaced by mock functions (stubs) during verification. Although the current focus is on functions, the hope is that this mechanism might in the future be extended to other features (like methods and types).

The need for stubs is twofold:

1. Users might need to stub functions containing features that Kani does not support, such as inline assembly.
2. Users might need to stub functions containing code that Kani supports in principle, but which in practice leads to bad verification performance.

There are questions to resolve:

1. What should be the user experience? How and where should users specify which stubs to use?
2. Should stubs be applied consistently across all harnesses or should they be specific to a harness (i.e., different harnesses can use different stubs)?

aaronbembenek-aws:
### Description of changes: 

Create an RFC for the UX and backend implementation of function and method stubbing. This would allow users to specify that, during verification, some functions/methods should be replaced with other functions/methods. These substitutions would help Kani handle code that has unsupported features (like inline assembly) or leads to bad verification performance.

I have created a [proof-of-concept](https://github.com/aaronbembenek-aws/kani/tree/mir_transform) of the backend MIR-to-MIR transformation; the stub mapping is read from a file.

You can also find a rendered version of this document [here](https://github.com/aaronbembenek-aws/kani/blob/issue-1695-rfc/rfc/src/rfcs/0002-function-stubbing.md).

### Related issues:

Resolves #1695

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/rfc/src/rfcs/0002-function-stubbing.md
+++ b/rfc/src/rfcs/0002-function-stubbing.md
@@ -83,10 +83,6 @@ mod my_mod {
 }
 ```
 
-### Accessing Private Fields in Method Stubs
-
-**TODO**
-
 ### Stub Sets
 
 As a convenience, users will also be able to specify sets of stubs that can be applied to multiple harnesses.
@@ -119,9 +115,44 @@ The same mechanism can be used to union together stub sets:
 fn all_my_stubs() {}
 ```
 
+### Accessing Private Fields in Method Stubs
+
+When stubbing a method, users might need to access private fields within the struct.
+Because users are writing stubs in Rust and are subject to Rust's restrictions, they cannot access private fields directly.
+Instead, they have to create a mock struct that has the same layout as the original struct, but with public fields, and then use `std::mem::transmute` to cast between them.
+
+```rust
+struct Foo {
+    x: u32,
+}
+
+impl Foo {
+    pub fn m(&self) -> u32 {
+        0
+    }
+}
+
+struct MockFoo {
+    pub x: u32,
+}
+
+fn mock_m(foo: &Foo) {
+    let mock: &MockFoo = unsafe { std::mem::transmute(foo) };
+    return mock.x;
+}
+
+#[cfg(kani)]
+#[kani::proof]
+#[kani::stub_by(Foo::m, mock_m)]
+fn my_harness() { ... }
+```
+
+We acknowledge that this approach is both slightly clunky and brittle, since it requires knowing the layout of the original struct (which could change) and depends on `rustc` producing the same layout for both structs (which is not guaranteed unless both are annotated with `repr(C)`).
+It is an open question whether we can provide a mechanism to hide some of this ugliness from the user, or at least error if, say, the struct layouts differ.
+
 ### Error Conditions
 
-Given a set of (`original`, `replacement`) pairs, Kani will exit with an error if
+Given a set of `original`-`replacement` pairs, Kani will exit with an error if
 
 1. a specified `replacement` stub does not exist;
 2. the user specifies conflicting stubs for the same harness (i.e., if the same `original` function is mapped to multiple `replacement` functions); or
@@ -129,7 +160,7 @@ Given a set of (`original`, `replacement`) pairs, Kani will exit with an error i
 
 ### Pedagogy
 
-To teach this feature, we will update the documentation with a section on function stubbing, including simple examples showing how stubbing can help Kani handle code that currently cannot be verified.
+To teach this feature, we will update the documentation with a section on function and method stubbing, including simple examples showing how stubbing can help Kani handle code that currently cannot be verified.
 
 ## Detailed Design
 
@@ -157,7 +188,6 @@ Since there is only a single stub set in this situation, `kani-driver` needs to 
 ## Rationale and alternatives: user experience
 
 Stubbing is a *de facto* necessity for verification tools, and the lack of stubbing has a negative impact on the usability of Kani.
-**TODO**: Argue why stubbing functions/methods is enough/a good start
 
 ### Benefits
 
@@ -239,7 +269,9 @@ At this stage of the compiler, names have been fully resolved, and there is no p
 
 The major downside with the MIR-to-MIR transformation is that it does not appear to be possible to stub types at that stage (there is no way to change the definition of a type through the MIR).
 Thus, our proposed approach will not be a fully general stubbing solution.
-However, it can be used as part of a portfolio of stubbing approaches, where users stub local types using conditional compilation (see Alternative #1), and Kani provides a modified version of the standard library with verification-friendly versions of types like `std::vec::Vec`.
+However, it is technically feasible and relatively clean, and provides benefits over having no stubbing at all (as can be seen in the examples in the first part of this document).
+
+Furthermore, it can be used as part of a portfolio of stubbing approaches, where users stub local types using conditional compilation (see Alternative #1), and Kani provides a modified version of the standard library with verification-friendly versions of types like `std::vec::Vec`.
 
 ### Alternative #1: Conditional compilation
 
@@ -279,9 +311,10 @@ Furthermore, it would require that we have access to the AST/HIR for all externa
 Requiring the replacement's type to be a subtype of the original type is likely stronger than what we want.
 For example, if the original function is polymorphic but monomorphized to only a single type, then it seems okay to replace it with a function that matches the monomorphized type.
 - When a user stubs a method and wants access to private fields, is there some way we can hide the `std::mem::transmute` ugliness?
+Can we error if the mock struct's layout differs from the original struct's?
 
 ## Future possibilities
 
 - It would increase the utility of stubbing if we supported stubs for types.
-The source code annotations could likely stay the same, although the underlying technical mechanisms in `kani-compiler` performing these substitutions might be significantly more complex.
+The source code annotations could likely stay the same, although the underlying technical approach performing these substitutions might be significantly more complex.
 - It would probably make sense to provide a library of common stubs for users, since many applications might want to stub the same functions and mock the same behaviors (e.g., `rand::random` can be replaced with a function returning `kani::any`).
--- a/rfc/src/rfcs/0002-function-stubbing.md
+++ b/rfc/src/rfcs/0002-function-stubbing.md
@@ -1,0 +1,99 @@
+- **Feature Name:** Function stubbing (`function_stubbing`)
+- **Feature Request Issue:** [model-checking#1695](https://github.com/model-checking/kani/issues/1695)
+- **RFC PR:** *Link to original PR*
+- **Status:** Under Review
+- **Version:** 0
+- **Proof-of-concept:** *Optional field. If you have implemented a proof of concept, add a link here*
+
+## Summary
+
+This will feature will allow users to specify that certain functions should be replaced with mock functions (stubs) during verification.
+
+## User Impact
+
+We anticipate that stubbing will have a substantial positive impact on the usability of Kani. There are two main motivations for stubbing:
+
+
+1. Users might need to stub functions containing features that Kani does not support, such as inline assembly.
+2. Users might need to stub functions containing code that Kani supports in principle, but which in practice leads to bad verification performance.
+
+In both cases, stubbing would enable users to verify code that cannot currently be verified by Kani (or at least not within a reasonable resource bound).
+
+As an example, consider verifying the following assertion (which can fail):
+
+```rust
+assert!(rand::random::<u32>() != 0);
+```
+
+Currently, running Kani on this leads to 1503 checks: one is a failure because of a missing definition, and the other 1502 are undetermined. That is, Kani is currently unable to prove that this assertion can fail. Verification time is 0.52 seconds.
+
+Using stubbing, we can specify that the function `rand::random` should be replaced with the function `mock_random`, which we can define as
+
+```rust
+#[cfg(kani)]
+fn mock_random<T: kani::Arbitrary>() -> T {
+    kani::any()
+}
+```
+
+Under this substitution, Kani has a single check, which proves that the assertion can fail. Verification time is 0.03 seconds.
+
+## User Experience
+
+This feature is currently limited to stubbing functions; however, the hope is that the basic ideas and mechanisms will carry over to stubbing other features in the future (such as methods and types).
+
+Stubs will be specified per harness; that is, different harnesses can use different stubs (the reasoning being that users might want to mock different behavior for different harnesses).
+Users will specify stubs by attaching the `#[kani::stub_by(original, replacement)]` attribute to each harness function.
+The attribute may be specified multiple times per harness, so that multiple (non-conflicting) stub pairings are supported.
+The arguments `original` and `replacement` give the names of functions, relative to the crate of the harness (*not* relative to the module of the harness).
+
+For example, this code specifies that the function `mock_random` should be used in place of the function `rand::random` for the harness `check_random`:
+
+```rust
+#[cfg(kani)]
+fn mock_random<T: kani::Arbitrary>() -> T {
+    kani::any()
+}
+
+mod my_mod {
+
+    #[cfg(kani)]
+    #[kani::proof]
+    #[kani::stub_by(rand::random, mock_random)]
+    fn check_random() {
+        assert!(rand::random::<u32>() != 0);
+    }
+
+}
+```
+
+Kani will exit with an error if a specified `replacement` stub does not exist or if the user specifies conflicting stubs for the same harness (i.e., if the same `original` function is mapped to multiple `replacement` functions).
+
+To teach this feature, we will update the documentation with a section on function stubbing, including simple examples showing how stubbing can help Kani handle code that currently cannot be verified.
+
+## Detailed Design
+
+This is the technical portion of the RFC. Please provide high level details of the implementation you have in mind:
+
+- What are the main components that will be modified? (E.g.: changes to `kani-compiler`, `kani-driver`, metadata,
+  installation...)
+- How will they be modified? Any changes to how these components communicate?
+- Will this require any new dependency?
+- What corner cases do you anticipate?
+
+## Rationale and alternatives
+
+- What are the pros and cons of this design?
+- What is the impact of not doing this?
+- What other designs have you considered? Why didn't you choose them?
+
+## Open questions
+
+- Is there any part of the design that you expect to resolve through the RFC process?
+- What kind of user feedback do you expect to gather before stabilization? How will this impact your design?
+- Would there ever be the need to stub particular monomorphizations of functions, as opposed to the generic function?
+
+## Future possibilities
+
+What are natural extensions and possible improvements that you predict for this feature that is out of the
+scope of this RFC? Feel free to brainstorm here.
--- a/rfc/src/rfcs/0002-function-stubbing.md
+++ b/rfc/src/rfcs/0002-function-stubbing.md
@@ -11,15 +11,17 @@ This will feature will allow users to specify that certain functions and methods
 
 ## User Impact
 
-We anticipate that stubbing will have a substantial positive impact on the usability of Kani. There are two main motivations for stubbing:
+We anticipate that stubbing will have a substantial positive impact on the usability of Kani:
 
 1. Users might need to stub functions/methods containing features that Kani does not support, such as inline assembly.
-2. Users might need to stub functions/methods containing code that Kani supports in principle, but which in practice leads to bad verification performance.
+2. Users might need to stub functions/methods containing code that Kani supports in principle, but which in practice leads to bad verification performance (for example, if it contains loops).
+3. Users could use stubbing to perform compositional reasoning: prove the behavior of a function/method `f`, and then in other proofs use a stub of `f` that mocks that behavior but is less complex.
 
-In both cases, stubbing would enable users to verify code that cannot currently be verified by Kani (or at least not within a reasonable resource bound).
+In all cases, stubbing would enable users to verify code that cannot currently be verified by Kani (or at least not within a reasonable resource bound).
+Stubbing might also be helpful in the development of Kani, as it would make it possible to run experiments like "Will Kani get better performance if we replace all instances of `Vec::new` in the codebase with `Vec::with_capacity?`" (following [Issue 1673](https://github.com/model-checking/kani/issues/1673)).
 
-In what follows, we give three examples of stubbing being put to work.
-Each of these examples runs on a prototype version of the stubbing mechanism we propose (except that the prototype does not support stub-related annotations; instead, it reads the stub mapping from a file).
+In what follows, we give XXX examples of stubbing being put to work, using the annotations we propose in this RFC.
+We are able to run each of these examples on a modified version of Kani using a proof-of-concept MIR-to-MIR transformation implementing stubbing (the prototype does not support stub-related annotations; instead, it reads the stub mapping from a file).
 
 ### Mocking Randomization
 
@@ -249,7 +251,8 @@ Stubbing is a *de facto* necessity for verification tools, and the lack of stubb
 - Because stubs are specified by annotating the harness, the user is able to specify stubs for functions they do not have source access to (like library functions).
 This contrasts with annotating the function to be replaced (such as with function contracts).
 - The current design provides the user with flexibility, as they can specify different sets of stubs to use for different harnesses.
-- The stub mappings are all located right by the harness, which makes it easy to understand which replacements are going to happen for each harness.
+This is important if users are trying to perform compositional reasoning using stubbing, since in some harnesses a function/method should be verified, in in other harnesses its behavior should be assumed.
+- The stub selections are located adjacent to the harness, which makes it easy to understand which replacements are going to happen for each harness.
 
 ### Risks
 
@@ -262,6 +265,7 @@ This is similar to other specification bugs.
 
 - In many cases, stubs are more user-friendly than contracts. With contracts, it is necessary to explicitly provide information that is automatically captured in Rust (such as which memory is written).
 - The [currently proposed function contract mechanism](https://github.com/model-checking/kani/tree/features/function-contracts) does not provide a way to put contracts on external functions.
+This is one of the key motivations for stubbing.
 
 ### Alternative #1: Annotate stubbed functions
 
--- a/rfc/src/rfcs/0002-function-stubbing.md
+++ b/rfc/src/rfcs/0002-function-stubbing.md
@@ -20,13 +20,14 @@ We anticipate that stubbing will have a substantial positive impact on the usabi
 In all cases, stubbing would enable users to verify code that cannot currently be verified by Kani (or at least not within a reasonable resource bound).
 Stubbing might also be helpful in the development of Kani, as it would make it possible to run experiments like "Will Kani get better performance if we replace all instances of `Vec::new` in the codebase with `Vec::with_capacity?`" (following [Issue 1673](https://github.com/model-checking/kani/issues/1673)).
 
-In what follows, we give XXX examples of stubbing external code, using the annotations we propose in this RFC.
+In what follows, we give two examples of stubbing external code, using the annotations we propose in this RFC.
 We are able to run each of these examples on a modified version of Kani using a proof-of-concept MIR-to-MIR transformation implementing stubbing (the prototype does not support stub-related annotations; instead, it reads the stub mapping from a file).
-These examples are the types of functions/methods that are commonly stubbed in other verification and analysis projects.
+These examples -- involving randomization and deserialization -- are the types of functions/methods that are commonly stubbed in other verification and analysis projects.
+Other common examples that we should be able to handle include system calls and timer functions.
 
 ### Mocking Randomization
 
-The crate [`rand`](https://crates.io/crates/rand) has been downloaded 150M times.
+The crate [`rand`](https://crates.io/crates/rand) is widely used (150M downloads).
 However, Kani cannot currently handle code that uses it (Kani users have run into this; see <https://github.com/model-checking/kani/issues/1727>).
 Consider this example:
 
@@ -58,11 +59,9 @@ fn random_cannot_be_zero() {
 
 Under this substitution, Kani has a single check, which proves that the assertion can fail. Verification time is 0.02 seconds.
 
-### Mocking System Calls
-
-**TODO**
-
 ### Mocking Deserialization
+
+In this example, we mock a [serde_json](https://crates.io/crates/serde_json) (96M downloads) deserialization method so that we can prove a property about the following [Firecracker function](https://github.com/firecracker-microvm/firecracker/blob/01eba51ded2f5439da91a2d73280f579651b067c/src/api_server/src/request/vsock.rs#L11) that parses a configuration from some raw data:
 
 ```rust
 fn parse_put_vsock(body: &Body) -> Result<ParsedRequest, Error> {
@@ -76,7 +75,8 @@ fn parse_put_vsock(body: &Body) -> Result<ParsedRequest, Error> {
     let mut deprecation_message = None;
     if vsock_cfg.vsock_id.is_some() {
         // vsock_id field in request is deprecated.
-        METRICS.deprecated_api.deprecated_http_api_calls.inc();
+        // XXX We needed to comment this out to get it complete in a reasonable time
+        // METRICS.deprecated_api.deprecated_http_api_calls.inc();
         deprecation_message = Some("PUT /vsock: vsock_id field is deprecated.");
     }
 
@@ -91,7 +91,42 @@ fn parse_put_vsock(body: &Body) -> Result<ParsedRequest, Error> {
 }
 ```
 
+In our proof, we commented out one line in the function that updates statistics (and that seems to lead to bad solver performance), and manually mocked some of the Firecracker types with simpler versions to reduce the number of dependencies we had to pull in (e.g., we removed some enum variants, unused struct fields).
+With these changes, we were able to prove that the configuration data has a vsock ID if and only if the parsing metadata includes a deprecation message: 
+
+```rust
+#[cfg(kani)]
+fn get_vsock_device_config(action: RequestAction) -> Option<VsockDeviceConfig> {
+    match action {
+        RequestAction::Sync(vmm_action) => match *vmm_action {
+            VmmAction::SetVsockDevice(dev) => Some(dev),
+            _ => None,
+        },
+        _ => None,
+    }
+}
+
+#[cfg(kani)]
+#[kani::proof]
+#[kani::unwind(5)]
+#[kani::stub_py("serde_json::deserialize_slice", "mock_deserialize")]
+fn test_deprecation_vsock_id_consistent() {
+    // We are going to mock the parsing of this body, so might as well use an empty one.
+    let body: Vec<u8> = Vec::new();
+    if let Ok(res) = parse_put_vsock(&Body::new(body)) {
+        let (action, mut parsing_info) = res.into_parts();
+        let config = get_vsock_device_config(action).unwrap();
+        assert_eq!(
+            config.vsock_id.is_some(),
+            parsing_info.take_deprecation_message().is_some()
+        );
+    }
+}
 ```
+
+Crucially, we did this by stubbing out `serde_json::from_slice` and replacing it with our mock version below, which ignores its input and creates a "symbolic" configuration struct:
+
+```rust
 #[cfg(kani)]
 fn symbolic_string(len: usize) -> String {
     let mut v: Vec<u8> = Vec::with_capacity(len);
@@ -103,7 +138,7 @@ fn symbolic_string(len: usize) -> String {
 
 #[cfg(kani)]
 fn mock_deserialize(_data: &[u8]) -> serde_json::Result<VsockDeviceConfig> {
-    let str_len = 0;
+    let str_len = 4;
     let vsock_id = if kani::any() {
         None
     } else {
@@ -118,33 +153,9 @@ fn mock_deserialize(_data: &[u8]) -> serde_json::Result<VsockDeviceConfig> {
     };
     Ok(dev)
 }
-
-#[cfg(kani)]
-fn get_vsock_device_config(action: RequestAction) -> Option<VsockDeviceConfig> {
-    match action {
-        RequestAction::Sync(vmm_action) => match *vmm_action {
-            VmmAction::SetVsockDevice(dev) => Some(dev),
-            _ => None,
-        },
-        _ => None,
-    }
-}
-
-#[cfg(kani)]
-#[kani::proof]
-fn test_deprecation_vsock_id_consistent() {
-    // We are going to mock the parsing of this body, so might as well use an empty one.
-    let body: Vec<u8> = Vec::new();
-    if let Ok(res) = parse_put_vsock(&Body::new(body)) {
-        let (action, mut parsing_info) = res.into_parts();
-        let config = get_vsock_device_config(action).unwrap();
-        assert_eq!(
-            config.vsock_id.is_some(),
-            parsing_info.take_deprecation_message().is_some()
-        );
-    }
-}
 ```
+
+The proof takes 226 seconds to complete.
 
 ## User Experience
 
--- a/rfc/src/rfcs/0002-function-stubbing.md
+++ b/rfc/src/rfcs/0002-function-stubbing.md
@@ -9,7 +9,7 @@
 
 Allow users to specify that certain functions and methods should be replaced with mock functions (stubs) during verification.
 
-## User Impact
+## User impact
 
 We anticipate that stubbing will have a substantial positive impact on the usability of Kani:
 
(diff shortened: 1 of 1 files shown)


rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,340 @@
...
+
+```rust
+#[cfg(kani)]
+#[kani::stub_set]
+#[kani::use_stub_set("my_io_stubs")]
+#[kani::use_stub_set("other_stub_set")]
+fn all_my_stubs() {}
+```
+
+### Accessing Private Fields in Method Stubs
```
celinval, line 134: Part of me would prefer if we leave this out of the initial RFC and just leave it as part of "Open Questions".

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,340 @@
+- **Feature Name:** Function and method stubbing (`function_stubbing`)
+- **Feature Request Issue:** <https://github.com/model-checking/kani/issues/1695>
+- **RFC PR:** <https://github.com/model-checking/kani/pull/1723> 
+- **Status:** Under Review
+- **Version:** 0
+- **Proof-of-concept:** <https://github.com/aaronbembenek-aws/kani/tree/mir_transform>
+
+## Summary
+
+This will feature will allow users to specify that certain functions and methods should be replaced with mock functions (stubs) during verification.
```
celinval, line 10:
```suggestion
Allow users to specify that certain functions and methods should be replaced with mock functions (stubs) during verification.
```

celinval (commented):
Awesome! I haven't reviewed everything, but I left a few comments. Thanks

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,340 @@
...
+#[kani::proof]
+#[kani::stub_by("rand::random", "mock_random")]
+fn random_cannot_be_zero() {
+    assert_ne!(rand::random::<u32>(), 0);
+}
+```
+
+Under this substitution, Kani has a single check, which proves that the assertion can fail. Verification time is 0.02 seconds.
+
+### Mocking System Calls
```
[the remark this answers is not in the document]
aaronbembenek-aws: Okay, I'm adding a longer example based on Firecracker and then just listing other stubbing possibilities.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,340 @@
...
+    #[kani::stub_by("my_mod::foo", "my_mod::bar")]
+    fn my_harness() { ... }
+
+}
+```
+
+### Stub Sets
+
+As a convenience, users will also be able to specify sets of stubs that can be applied to multiple harnesses.
+First, users write a "dummy" function with the name of the stub set, annotate it with the `#[kani::stub_set]` attribute, and add the desired stub pairings as further attributes:
```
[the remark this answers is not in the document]
aaronbembenek-aws: I added a note saying that we can likely wrap up the stub set creation in a more ergonomic macro.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: That'd be fine with me. @adpaco-aws what do you think?

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,340 @@
...
+We anticipate that this design will evolve and be iterated upon.
+
+### Full form
+
+In its full form, we expect that this feature will require substantial changes both to `kani-driver` and `kani-compiler`.
+
+`kani-driver` will need to be responsible for determining which stubs are required for each harness, which will require moving the code for identifying harnesses from `kani-compiler` to `kani-driver`.
+After `kani-driver` has determined the set of stubs to use for each harness, it will invoke `kani-compiler` once for each set of stubs, passing the relevant stub pairings as command line arguments.
+The output of `kani-compiler` will be specialized to that stub set.
+After `kani-compiler` has finished compiling under a given stub set, `kani-driver` will run the harnesses that use that stub set.
```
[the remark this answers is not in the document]
aaronbembenek-aws: Perfect - that'll make life easier :)

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,340 @@
...
+## Detailed Design
+
+We discuss both the design in its full form and a simplified version appropriate for a first step.
+We anticipate that this design will evolve and be iterated upon.
+
+### Full form
+
+In its full form, we expect that this feature will require substantial changes both to `kani-driver` and `kani-compiler`.
+
+`kani-driver` will need to be responsible for determining which stubs are required for each harness, which will require moving the code for identifying harnesses from `kani-compiler` to `kani-driver`.
```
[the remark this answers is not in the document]
aaronbembenek-aws: No longer necessary, given what you say below. Sounds like most changes can happen just in `kani_compiler` (which is great).

rfc/src/rfcs/0002-function-stubbing.md
adpaco-aws: Agree, this is a subtle detail that appears way too early in the RFC.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+fn symbolic_string(len: usize) -> String {
+    let mut v: Vec<u8> = Vec::with_capacity(len);
+    for _ in 0..len {
+        v.push(kani::any());
+    }
+    unsafe { String::from_utf8_unchecked(v) }
+}
+
+#[cfg(kani)]
+fn mock_deserialize(_data: &[u8]) -> serde_json::Result<VsockDeviceConfig> {
```
adpaco-aws, line 140: The `Error` case isn't covered here so the coverage in `parse_put_vsock` isn't complete. You're likely to run into performance issues again, feel free to comment out the `METRICS.put_api_requests.vsock_fails.inc()` call as well.

Basically, you need to (at the beginning) nondeterministically:
 1. Throw an `Error` or,
 2. Do what you're already doing.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+    let str_len = 4;
```
adpaco-aws, line 141: Make this a constant.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+```
+
+The proof takes 226 seconds to complete.
+
+## User Experience
+
+This feature is currently limited to stubbing functions and methods.
+We anticipate that the annotations we propose here could also be used when stubbing types, although the underlying technical approach might have to change.
+
+Stubs will be specified per harness; that is, different harnesses can use different stubs (the reasoning being that users might want to mock different behavior for different harnesses).
```
adpaco-aws, line 165: Or maybe they don't want to mock its behavior. In general, you don't want to force the same stubs for a complete code base. Needs more explaining... this is the main design point.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+Users will specify stubs by attaching the `#[kani::stub_by("<original>", "<replacement>")]` attribute to each harness function.
```
adpaco-aws, line 166: I would remove the `"` and say what you expect for `original` and `replacement` immediately after (i.e., swap the next two sentences).

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+## Rationale and alternatives: user experience
+
+Stubbing is a *de facto* necessity for verification tools, and the lack of stubbing has a negative impact on the usability of Kani.
+
+### Benefits
+
+- Because stubs are specified by annotating the harness, the user is able to specify stubs for functions they do not have source access to (like library functions).
+This contrasts with annotating the function to be replaced (such as with function contracts).
+- The current design provides the user with flexibility, as they can specify different sets of stubs to use for different harnesses.
+This is important if users are trying to perform compositional reasoning using stubbing, since in some harnesses a function/method should be verified, in in other harnesses its behavior should be assumed.
```
adpaco-aws, line 293: I would say "fully verified"/"mocked" instead.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+- The stub selections are located adjacent to the harness, which makes it easy to understand which replacements are going to happen for each harness.
+
+### Risks
+
+- Users can always write stubs that do not correctly correspond to program behavior, and so a successful verification does not actually mean the program is bug-free.
```
adpaco-aws, line 298: Let's note that the stubbing code will be available, so the assumptions it makes are available to consult.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+This is similar to other specification bugs. 
+
+### Comparison to function contracts
+
+- In many cases, stubs are more user-friendly than contracts. With contracts, it is necessary to explicitly provide information that is automatically captured in Rust (such as which memory is written).
+- The [currently proposed function contract mechanism](https://github.com/model-checking/kani/tree/features/function-contracts) does not provide a way to put contracts on external functions.
```
adpaco-aws, line 304: "put" -> "specify".

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+- The stub selections are located adjacent to the harness, which makes it easy to understand which replacements are going to happen for each harness.
+
+### Risks
+
+- Users can always write stubs that do not correctly correspond to program behavior, and so a successful verification does not actually mean the program is bug-free.
+This is similar to other specification bugs. 
+
+### Comparison to function contracts
+
+- In many cases, stubs are more user-friendly than contracts. With contracts, it is necessary to explicitly provide information that is automatically captured in Rust (such as which memory is written).
```
adpaco-aws, line 303: Are `assign` clauses really necessary? A good point is that contract predicates are a DSL on its own... Not just Rust code.

rfc/src/rfcs/0002-function-stubbing.md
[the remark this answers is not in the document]
celinval: `DummyStubs` is not necessarily a struct. We can pick whatever syntax we want. I just used `{}` to make it clear that this is a group.

I think we will have to eventually generate a dummy item (either a function, a module, a struct or a constant) to add the attributes to. I don't think it matters what we pick. The only real difference between them is the name convention. If we want the name to be PascalCase, we should use an empty struct, if we want snake_case it should be a module or a function.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+}
+
+#(cfg[kani])
+#[kani::proof]
+#[kani::use_stubs("my_stubs")]
+fn my_harness() { ... }
+```
+
+The benefit is that stubs are specified per harness, and (using modules) it might be possible to group stubs together.
+The downside is that multiple annotations are required and the stub mappings themselves are remote from the harness.
```
adpaco-aws, line 352: Conflicts wouldn't be easy to deal with, right? For example, if I had to stubs replacing the same function, I need to change one stub set so only one stub gets used for the replacement.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+
+The downside with this approach is that it requires all source code to be available.
+It also might be difficult to inject code in a way that names are correctly resolved (e.g., if the replacement code comes from a different crate).
+Also, source code is difficult to work with, and includes things like unexpanded macros.
+
+On the last two points, we might be able to take advantage of an existing source analysis platform like `rust-analyzer` (which has facilities like structural search replace), but this would add more (potentially fragile) dependencies to Kani.
+
+### Alternative #3: AST-to-AST or HIR-to-HIR transformation
+
+In this alternative, we implement stubbing by rewriting the AST or [High-Level IR (HIR)](https://rustc-dev-guide.rust-lang.org/hir.html) of the program.
```
adpaco-aws, line 395: Is there documentation for AST as there is for HIR? If so, I suggest linking it here.

adpaco-aws (commented):
Thanks, this looks great overall! Just have a couple more comments.

rfc/src/rfcs/0002-function-stubbing.md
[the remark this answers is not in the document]
aaronbembenek-aws: I updated this section to use the macro syntax.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: I removed this section.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+## User Impact
+
+We anticipate that stubbing will have a substantial positive impact on the usability of Kani:
+
+1. Users might need to stub functions/methods containing features that Kani does not support, such as inline assembly.
+2. Users might need to stub functions/methods containing code that Kani supports in principle, but which in practice leads to bad verification performance (for example, if it contains loops).
+3. Users could use stubbing to perform compositional reasoning: prove the behavior of a function/method `f`, and then in other proofs use a stub of `f` that mocks that behavior but is less complex.
+
+In all cases, stubbing would enable users to verify code that cannot currently be verified by Kani (or at least not within a reasonable resource bound).
+Stubbing might also be helpful in the development of Kani, as it would make it possible to run experiments like "Will Kani get better performance if we replace all instances of `Vec::new` in the codebase with `Vec::with_capacity?`" (following [Issue 1673](https://github.com/model-checking/kani/issues/1673)).
```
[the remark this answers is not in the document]
aaronbembenek-aws: I clarified this.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: Declared inside the function or outside?

adpaco-aws: I'd prefer inside, since all uses are local and it'd be within the proof itself (not usable outside of it).

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: Wouldn't this problem occur whenever you are using stub sets (e.g., also in the stub set approach we suggest)? Like our proposed approach, this alternative doesn't require using stub sets - it's just an option

rfc/src/rfcs/0002-function-stubbing.md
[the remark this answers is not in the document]
aaronbembenek-aws: @adpaco-aws What is the particular syntax you propose?

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,416 @@
...
+        METRICS.put_api_requests.vsock_fails.inc();
+        err
+    })?;
+
+    // Check for the presence of deprecated `vsock_id` field.
+    let mut deprecation_message = None;
+    if vsock_cfg.vsock_id.is_some() {
+        // vsock_id field in request is deprecated.
+        // XXX We needed to comment this out to get it complete in a reasonable time
+        // METRICS.deprecated_api.deprecated_http_api_calls.inc();
```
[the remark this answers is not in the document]
aaronbembenek-aws: I actually tried stubbing these calls with a NOP, and it still leads the SAT solver to run out of memory. Is there a way to increase the amount of memory allocated to the SAT solver?

rfc/src/rfcs/0002-function-stubbing.md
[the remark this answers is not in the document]
adpaco-aws: The `stub_by` part isn't needed (we can generate it), we could include it if you think that'll make things clearer. But in any case we just need the name for the set and the pairs.

rfc/src/rfcs/0002-function-stubbing.md
adpaco-aws: I see, I've had this problem in the past and it's really really hard to work around it. It's OK to leave as is.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+## User impact
+
+We anticipate that stubbing will have a substantial positive impact on the usability of Kani:
+
+1. Users might need to stub functions/methods containing features that Kani does not support, such as inline assembly.
+2. Users might need to stub functions/methods containing code that Kani supports in principle, but which in practice leads to bad verification performance (for example, if it contains deserialization code).
+3. Users could use stubbing to perform compositional reasoning: prove the behavior of a function/method `f`, and then in other proofs---that call `f` indirectly---use a stub of `f` that mocks that behavior but is less complex.
+
+In all cases, stubbing would enable users to verify code that cannot currently be verified by Kani (or at least not within a reasonable resource bound).
+Even without stubbing types, the ability to stub functions/methods can help provide verification-friendly abstractions for standard data structures.
```
tedinski, line 21: > Even without stubbing types

purely as a readability / flow nit: this starts talking about a case before you've really defined what these different types of stubbing are.

As I read on, I think this RFC is missing a paragraph or section discussing exactly this. We mention "stubbing types" here and "stubbing function/methods" later, but what else might there be?

One pattern I like in design docs like this is to have a section clearly articulating what's in-scope and out-of-scope. This might be a way to do this. For example, in-scope are function/methods and stub sets, out-of-scope are replacing types.

And while I'm thinking about this, I wonder if (as part of this work) we should develop a library of safe stubs for standard library functions that users can opt-into. e.g. `#[kani::stubs(kani::std_stubs)]` or something. Another idea besides `Vec::with_capacity` might be how `HashMap::new` works (e.g. removing/mitigating the randomness and thread local?)

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+#[kani::proof]
+#[kani::stub_by("rand::random", "mock_random")]
+fn random_cannot_be_zero() {
+    assert_ne!(rand::random::<u32>(), 0);
+}
+```
+
+Under this substitution, Kani has a single check, which proves that the assertion can fail. Verification time is 0.02 seconds.
+
+### Mocking deserialization
```
celinval, line 63: This is pretty cool.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+
+1. a specified `replacement` stub does not exist;
+2. the user specifies conflicting stubs for the same harness (i.e., if the same `original` function is mapped to multiple `replacement` functions); or
+3. the signature of the `replacement` function is not compatible with the signature of the `original` function.
+
+### Pedagogy
+
+To teach this feature, we will update the documentation with a section on function and method stubbing, including simple examples showing how stubbing can help Kani handle code that currently cannot be verified.
+
+## Detailed design
```
celinval, line 241: We will likely need some changes to the metadata so we can tell `kani-driver` which files were generated.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+
+We expect that this feature will require changes primarily to `kani-compiler`.
```
tedinski, line 243: This should be understood as a change to `kani-driver` as well.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+When declaring a harness, users can use the `#[kani::use_stub_set("<stub_set_name>")]` attribute to apply the stub set:
+
+```rust
+#[cfg(kani)]
+#[kani::proof]
+#[kani::use_stub_set("my_io_stubs")]
+fn my_harness() { ... }
+```
+
+A similar mechanism can be used to aggregate stub sets:
```
adpaco-aws, line 219: Nice! Makes sense to explicitly write `stub_by(...)` because of this.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+
+One alternative would be to specify stubs in a file that is passed to `kani-driver` via a command line option.
+Users would specify per-harness stub pairings in the file; JSON would be a possible format.
+Using a file would eliminate the need for `kani-driver` to extract harness information from the Rust source code; the rest of the implementation would stay the same.
+It would also allow the same harness to be run with different stub selections (by supplying a different file).
+The disadvantage is that the stub selection is remote from the harness itself.
+
+## Rationale and alternatives: stubbing mechanism
+
+Our approach is based on a MIR-to-MIR transformation.
```
celinval, line 335: I would highlight that another benefit is that we can extend our compiler to integrate `--concrete-playback` with the abstractions. But that this is out of the scope for this work.

celinval (commented):
Do we want to make any change to the results report to highlight that a stub was used?

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+Furthermore, it would require that we have access to the AST/HIR for all external code, and---provided we supported stubbing across crate boundaries---we would need to figure out how to inject the AST/HIR from one crate into another (the AST/HIR is usually just constructed for the crate currently being compiled).
+
+## Open questions
+
+- Would there ever be the need to stub a particular monomorphization of a function, as opposed to the polymorphic function?
+- What does it mean for the replacement function/method to be "compatible" with the original one?
+Requiring the replacement's type to be a subtype of the original type is likely stronger than what we want.
+For example, if the original function is polymorphic but monomorphized to only a single type, then it seems okay to replace it with a function that matches the monomorphized type.
+- How can we allow a user to stub a method and access private fields of `self`?
+This should be possible using `std::mem::transmute`, but can we hide this from the user and make it less brittle?
```
adpaco-aws, line 381: Not sure if it's good for all cases, but one possible approach here is instrumentation.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+}
+
+#(cfg[kani])
+#[kani::proof]
+#[kani::use_stubs("my_stubs")]
+fn my_harness() { ... }
+```
+
+The benefit is that stubs are specified per harness, and (using modules) it might be possible to group stubs together.
+The downside is that multiple annotations are required and the stub mappings themselves are remote from the harness.
```
tedinski, line 323: I don't understand the downside. We don't lose the ability to still "cherry-pick" things from a module right? Stub-sets are just as "remote" as a module like this, no?

It seems like this might be a "nicer" and more Rust-friendly way of writing stub-sets and not actually have any real downsides that I can see...

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+The HIR is a more compiler-friendly version of the AST; it is what is used for type checking.
+To swap out a function, method, or type at this level, it looks like it would be necessary to add another pass to `rustc` that takes the initial AST/HIR and produces a new AST/HIR with the appropriate replacements.
+
+The advantage with this approach is, like source transformations, it would be very flexible.
+The downside is that it would require modifying `rustc` (as far as we know, there is not an API for plugging in a new AST/HIR pass), and would also require performing the transformations at a very syntactic level: although the AST/HIR would likely be easier to work with than source code directly, it is still very close to the source code and not very abstract.
+Furthermore, it would require that we have access to the AST/HIR for all external code, and---provided we supported stubbing across crate boundaries---we would need to figure out how to inject the AST/HIR from one crate into another (the AST/HIR is usually just constructed for the crate currently being compiled).
```
tedinski, line 372: > we would need to figure out how to inject the AST/HIR from one crate into another

I'm not sure this is true?

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+
+## Open questions
+
```
zhassan-aws, line 375: How can the user verify that the stub is an abstraction of the original function/method?

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+- Would there ever be the need to stub a particular monomorphization of a function, as opposed to the polymorphic function?
+- What does it mean for the replacement function/method to be "compatible" with the original one?
+Requiring the replacement's type to be a subtype of the original type is likely stronger than what we want.
+For example, if the original function is polymorphic but monomorphized to only a single type, then it seems okay to replace it with a function that matches the monomorphized type.
+- How can we allow a user to stub a method and access private fields of `self`?
+This should be possible using `std::mem::transmute`, but can we hide this from the user and make it less brittle?
+
+## Future possibilities
```
celinval, line 383: Should we add a limitations section? E.g.:
- This won't work with `--concrete-playback` for now.
- We are only able to apply abstractions to dependencies if user enables MIR Linker.

rfc/src/rfcs/0002-function-stubbing.md
zhassan-aws: The RFC is focused on function stubbing. Can we add examples of method stubbing?

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: We will clarify this.

In scope:
- Replacing function bodies
- Replacing method bodies (which means that the new method body will be executed, whether the method is invoked directly or through a vtable)

Out of scope:
- Replacing type definitions
- Replacing macro definitions
- Mocking traits

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: Yes, that makes sense. I will add this.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: Yes! I'll clarify that.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: Some stronger downsides with this approach:

1. How do you mock multiple functions with the same stub? (Especially if, say, harness A wants to use `stub1` to mock `foo`, and harness B wants to use `stub1` to mock `bar`.)
2. How do you combine stub sets defined via modules? (This is probably possible, but would need to be fleshed out.)
3. Is it a regular module or not? Modules can contain other things than functions - how should we interpret them?

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: @tedinski We don't know until we compile later crates that earlier crates would need to have their HIR modified. Relatedly, a local abstraction would not be available while compiling a dependency.

@zhassan-aws We still need to iterate on exactly what the user experience for stubbing methods will be like. However, we can add an example of one approach to the Open Questions section.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: I'll add this - a good open question.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+We anticipate that the annotations we propose here could also be used when stubbing types, although the underlying technical approach might have to change.
+
+Stubs will be specified per harness; that is, different harnesses can use different stubs.
+This is one of the main design points.
+Users might want to mock the behavior of a function within one proof harness, and then mock it a different way for another harness, or even use the original function definition.
+It would be overly restrictive to impose the same stub definitions across all proof harnesses.
+A good example of this is compositional reasoning: in some harnesses, we want to prove properties of a particular function (and so want to use its actual implementation), and in other harnesses we want to assume that that function has those properties.
+
+Users will specify stubs by attaching the `#[kani::stub_by(<original>, <replacement>)]` attribute to each harness function.
+The arguments `original` and `replacement` give the names of functions (as string literals), relative to the crate of the harness (*not* relative to the module of the harness).
```
camshaft, line 172: If I understand this correctly, something like this would not be supported:

Say I had both the `rand` crate dependency and a local module:

```rust
mod rand {
    pub fn random() ...
}

#[kani::stub_by("rand::random", "mock_random")] // resolves to `self::rand::random`, not `::rand::random`
```

In the same vein, we won't handle imports either?

```rust
use path::to::rand_thing as rand;

#[kani::stub_by("rand::random", "mock_random")] // resolves to `path::to::rand_thing::random`, not `::rand::random`
```

To me, using the existing Rust scoping rules would be the most natural way to reference things.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+    };
+    Ok(config)
+}
+```
+
+The proof takes 170 seconds to complete (using Kissat as the backend SAT solver for CBMC).
+
+## User experience
+
+This feature is currently limited to stubbing functions and methods.
```
camshaft, line 162: Are you planning on enforcing visibility rules? In other words, are we going to allow stubbing a private function?

```rust
mod thing {
    pub fn foo() {
        bar();
    }

    fn bar() {
        ...
    }
}

#[kani::stub_by("thing::bar", "my_bar")]
```

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+}
+```
+
+### Stub sets
+
+As a convenience, we will provide a macro `kani::stub_set` that allows users to specify sets of stubs that can be applied to multiple harnesses:
+
+```rust
+kani::stub_set! {
+    my_io_stubs,
```
camshaft, line 204: Are these global symbols or resolved through the module path in which they're defined? I would prefer the latter.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,387 @@
...
+}
+```
+
+### Error conditions
+
+Given a set of `original`-`replacement` pairs, Kani will exit with an error if
+
+1. a specified `replacement` stub does not exist;
+2. the user specifies conflicting stubs for the same harness (i.e., if the same `original` function is mapped to multiple `replacement` functions); or
+3. the signature of the `replacement` function is not compatible with the signature of the `original` function.
```
camshaft, line 235: I see it's in the open questions section, but we definitely need to know what "compatible" means. Arity is an obvious aspect. But things start getting tricky when we're talking about generic constraints. For example are these compatible?

```rust
pub fn random<T>() -> T
    where Standard: Distribution<T>
```

```rust
pub fn random<T>() -> T
    where T: kani::Arbitrary
```

rfc/src/rfcs/0002-function-stubbing.md
celinval: Oh, that's a good point. I don't know why we have this restriction here. @aaronbembenek-aws can please clarify?

aaronbembenek-aws: The restriction was there so that the provided names would match the identifiers as they appear in the MIR. However, I agree that it would be a better UX to use the Rust scoping rules. So, I think we should target that approach, and try to resolve names on our end.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: Yes, you will be able to stub a private function.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: Thanks for the making this point - I agree that the latter is preferable, and that's what we'll target.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: We need to iterate on this and see what is technically feasible.  From my perspective, the optimal outcome would be that a stub is compatible if the monomorphized program is well typed. Our current prototype supports the example you give.

rfc/src/rfcs/0002-function-stubbing.md
camshaft: This is definitely a tradeoff of flexibility and maintainability. On the one hand, stubbing private functions makes it really easy for writing a new proof.

However, private functions don't have any guarantees on stability or behavior. Having to maintain a proof that stubs private functions could become really tedious every time you want to refactor or upgrade a dependency.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,455 @@
...
+### Risks
+
+- Users can always write stubs that do not correctly correspond to program behavior, and so a successful verification does not actually mean the program is bug-free.
+This is similar to other specification bugs.
+All the stubbing code will be available, so it is possible to inspect the assumptions it makes.
+
+### Comparison to function contracts
+
+- The [currently proposed function contract mechanism](https://github.com/model-checking/kani/tree/features/function-contracts) does not provide a way to specify contracts on external functions.
+This is one of the key motivations for stubbing.
```
zhassan-aws, line 300: I wouldn't say this is the key motivation because conceptually, it should be possible to allow contracts to be specified on external functions (which AFAIK, CBMC allows), but it hasn't been prioritized in the current implementation. I would say the key benefit of stubbing is the simplicity of replacing a function's implementation with another that is written in Rust. The replacement can be really be anything: an over-approximation, an under-approximation, or neither. On the other hand, the disadvantage of stubbing is that there's no easy mechanism to verify that a proof is sound with the specified stubs.

zhassan-aws (commented):
Very-well written! Great work, @aaronbembenek-aws!

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,455 @@
...
+
+```rust
+#[cfg(kani)]
+fn mock_random<T: kani::Arbitrary>() -> T {
+    kani::any()
+}
+
+#[cfg(kani)]
+#[kani::proof]
+#[kani::stub_by(rand::random, mock_random)]
```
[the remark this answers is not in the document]
aaronbembenek-aws: Sure, I think that's a good idea.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,455 @@
...
+fn random_cannot_be_zero() {
+    assert_ne!(rand::random::<u32>(), 0);
+}
+```
+
+Under this substitution, Kani has a single check, which proves that the assertion can fail. Verification time is 0.02 seconds.
+
+### Mocking deserialization
+
+In this example, we mock a [serde_json](https://crates.io/crates/serde_json) (96M downloads) deserialization method so that we can prove a property about the following [Firecracker function](https://github.com/firecracker-microvm/firecracker/blob/01eba51ded2f5439da91a2d73280f579651b067c/src/api_server/src/request/vsock.rs#L11) that parses a configuration from some raw data:
```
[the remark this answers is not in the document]
aaronbembenek-aws: We've decided to move this to the end of the RFC, as an appendix.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,455 @@
...
+It would be overly restrictive to impose the same stub definitions across all proof harnesses.
+A good example of this is compositional reasoning: in some harnesses, we want to prove properties of a particular function (and so want to use its actual implementation), and in other harnesses we want to assume that that function has those properties.
+
+Users will specify stubs by attaching the `#[kani::stub_by(<original>, <replacement>)]` attribute to each harness function.
+The arguments `original` and `replacement` give the names of functions/methods.
+They will be resolved using Rust's standard name resolution rules; this includes supporting imports like `use foo::bar as baz`.
+The attribute may be specified multiple times per harness, so that multiple (non-conflicting) stub pairings are supported.
+
+For example, this code specifies that the function `mock_random` should be used in place of the function `rand::random` and the function `my_mod::foo` should be used in place of the function `my_mod::bar` for the harness `my_mod::my_harness`:
```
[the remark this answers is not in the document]
aaronbembenek-aws: Nice catch!

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,455 @@
...
+#[kani::use_stub_set(my_io_stubs)]
+fn my_harness() { ... }
+```
+
+The name of the stub set will be resolved through the module path (i.e., they are not global symbols), using Rust's standard name resolution rules.
+
+A similar mechanism can be used to aggregate stub sets:
+
+```rust
+kani::stub_set!() {
```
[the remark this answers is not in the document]
aaronbembenek-aws: Another good catch! We don't need parentheses here.

zhassan-aws (approved):

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,461 @@
...
+We will support the stubbing of private functions and methods.
+While this provides flexibility that we believe will be necessary in practice, it can also lead to brittle proofs: private functions/methods can change or disappear in even minor version upgrades (thanks to refactoring), and so proofs that depend on them might have a high maintenance burden.
+In the documentation, we will discourage stubbing private functions/methods except if absolutely necessary.
+
+### Stub sets
+
+As a convenience, we will provide a macro `kani::stub_set` that allows users to specify sets of stubs that can be applied to multiple harnesses:
+
+```rust
+kani::stub_set! {
```
tedinski, line 126: nit/bike-shed: I think this syntax isn't ideal. How much control over it do you have? If I'm not mistaken you're suggesting that the first item in the list is the name of the stub set, right?

Can we do something like `kani::stub_set!(my_io_stubs) { ... the contents ... }` or similar?

tedinski (approved):
Nice doc!

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: Exactly, the first item would be the name of the stub set. How about `kani::stub_set!(my_io_stubs, { ... the contents ... });`? We should be able to support that. I believe there needs to be a single outer set of parentheses or braces enclosing everything (the name, the contents, etc.).

fzaiser: How about: `kani::stub_set! { my_io_stubs: ... }`? This would avoid having two types of delimiters (`{}` and `()`).

aaronbembenek-aws: Thanks @fzaiser! I think this is a pretty good solution. I'm going to go with it for now; we can always iterate on it.

celinval: I am actually not a big fan of using `:` as a separator. It looks like a typo for me. I would prefer regular rust separators,  `,`, or an aggregator: `{}` or `()`. E.g.: like:

```rust
kani::stub_set!(my_io_stubs {
 ... the contents ... 
});
```

camshaft: Ideally we use something that rustfmt will accept too 🙂. It basically needs to look like an existing construct for it to work.

celinval (approved):
I think the RFC overall looks good. I'm OK iterating over some of the details (such as the stub set syntax) as we move along as long.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: @camshaft That's a good point about rustfmt. Apparently it [doesn't format code in macro invocations between braces](https://users.rust-lang.org/t/rustfmt-skips-macro-arguments/74807/17). This syntax (based on @celinval's) works well with rustfmt:

```
kani::stub_set!(my_io_stubs(
    stub(std::fs::read, my_read),
    stub(std::fs::write, my_write),
));
```

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -0,0 +1,461 @@
...
+- We are only able to apply abstractions to some dependencies if the user enables the MIR linker.
+
+## Future possibilities
+
+- It would increase the utility of stubbing if we supported stubs for types.
+The source code annotations could likely stay the same, although the underlying technical approach performing these substitutions might be significantly more complex.
+- It would probably make sense to provide a library of common stubs for users, since many applications might want to stub the same functions and mock the same behaviors (e.g., `rand::random` can be replaced with a function returning `kani::any`).
+- How can we provide a good user experience for accessing private fields of `self` in methods?
+It is possible to do so using `std::mem::transmute` (see below); this is clunky and error-prone, and it would be good to provide better support for users.
+
```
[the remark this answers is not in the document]
aaronbembenek-aws: Another one is `uninterpreted`: treat function as an uninterpreted function (suggested by @weaversa [here](https://github.com/model-checking/kani/issues/1695#issuecomment-1289422800)).

PR #1724, opened 2022-09-27
Install Kani into the action, instead of using a docker

danielsn:
### Description of changes: 

Currently, the action requires a docker, which is not yet published. Instead, as a temporary solution, just install the dependencies every time. 

### Resolved issues:

Resolves #1589 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

- Installing the dependencies takes about 1-1.5 minutes according to some initial testing.

### Testing:

* How is this change tested? Made a repo with Kani proofs, ran this action on it.

### Checklist

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -26,7 +26,7 @@ runs:
         curl "${KANI_SCRIPT_FOLDER}/ubuntu/install_deps.sh" | bash ; \
         curl "${KANI_SCRIPT_FOLDER}/install_rustup.sh" | bash  ; \
         source  ~/.cargo/env; \
-        cargo install --locked kani-verifier; \
+        cargo install --version $KANI_VERSION --locked kani-verifier; \
         cargo-kani setup;
 
     - name: Run Kani
--- a/action.yml
+++ b/action.yml
@@ -13,12 +13,25 @@ inputs:
   command:
     description: 'Command to run.'
     required: false
-    default: 'cargo kani --workspace'
+    default: 'cargo-kani'
 
 runs:
-  using: 'docker'
-  image: docker://ghcr.io/model-checking/kani-ubuntu-20.04:0.11.0
-  args:
-    - bash
-    - '-c'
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -13,12 +13,20 @@ inputs:
   command:
     description: 'Command to run.'
     required: false
-    default: 'cargo kani --workspace'
+    default: 'cargo-kani'
```
tedinski, line 16: Is this default change desireable?

celinval (commented):
I do have one concern regarding mixing the repositories. I was puzzled about having the `action.yml` in the root of the repository. So I decided to read a bit more about actions. This is basically the first thing in their documentation:

> When you plan to publish your action to GitHub Marketplace, you'll need ensure that the repository only includes the metadata file, code, and files necessary for the action.

tedinski:
That's a pretty good point. We should have a `kani-action` repo or something?

action.yml
```diff
@@ -13,12 +13,20 @@ inputs:
...
-  args:
-    - bash
-    - '-c'
-    - 'set -e; export HOME=/root USER=$(id -nu) PATH=/root/.cargo/bin:$PATH; source $HOME/.bashrc; ${{ inputs.command }}'
+  using: "composite"
+  steps:
+    - name: Install Kani
+      shell: bash
+      run: |
+        export KANI_VERSION="0.11.0"; \
```
celinval, line 24: Should we retrieve the version from the repository instead? I was also wondering if we should allow users to choose the version they want to use so they can control when to update Kani.

danielsn: My thought was action version matches kani-version. So we have kani-action@0.11 matches kani 0.11

action.yml
danielsn: Fixed

celinval:
@danielsn Can we close this PR since we have published the Action in a different repo?

Issue #1711, opened 2022-09-22
UI unclear: reported results need to be clearer about the harness


tedinski:
Output concludes with something like this:

```

SUMMARY:
 ** 0 of 355 failed

VERIFICATION:- SUCCESSFUL

Verification Time: 1.0033967s
```

It's not clear (especially with a lot of properties above this) what harness had this result. It's too far away (e.g. 800+ properties between).

PR #1726, opened 2022-09-27
Introduce (experimental, unstable) parallel runner for proof harnesses

tedinski:
### Description of changes: 

What it says on the tin. :) Takes a 24s artificial benchmark down to under 4s to run.

1. Mimics the same options as `cargo build`: `--jobs` or `-j`. Default is 1. Default with just `-j` is num_cpus.
2. Conflicts with concrete-playback due to its dependence on order of harness running
3. Currently requires `--output-format=terse`. And `--enable-unstable` of course.

There is the problem of output. We actually don't have a huge problem with interleaving output because Adrian already wrote the renderer to print one big blob instead of multiple little ones. (Nice!) However, we do still have #1711 that this exacerbates, and the "verification time" isn't really paired with an output anymore. I think this is merge-able as-is, but I do plan on an independent PR to fix up that issue tomorrow, which it might be reasonable to want merged first.

### Call-outs:

Changes include:

1. Adding a new `--jobs` option to `kani-driver`
2. Replacing `KaniSession`'s `RefCell` with a `Mutex` (and adding `record_temporary_files`)
3. Swap a `for` to `rayon` `par_iter`

and that's it! It's actually a pretty simple change now. Sweet.

### Testing:

* How is this change tested? existing ci: all lines of code actually run by default, just without setting `j` higher than 1

### Checklist

### DIFF
--- a/kani-driver/src/args.rs
+++ b/kani-driver/src/args.rs
@@ -148,6 +148,10 @@ pub struct KaniArgs {
     // consumes everything
     pub cbmc_args: Vec<OsString>,
 
+    #[structopt(short, long, requires("enable-unstable"))]
+    // consumes everything
+    pub jobs: Option<Option<usize>>,
+
     // Hide option till https://github.com/model-checking/kani/issues/697 is
     // fixed.
     /// Use abstractions for the standard library.
@@ -224,6 +228,15 @@ impl KaniArgs {
             None
         } else {
             Some(DEFAULT_OBJECT_BITS)
+        }
+    }
+
+    /// Computes how many threads should be used to verify harnesses.
+    pub fn jobs(&self) -> Option<usize> {
+        match self.jobs {
+            None => Some(1),          // no argument, default 1
+            Some(None) => None,       // -j
+            Some(Some(x)) => Some(x), // -j=x
         }
     }
 }
@@ -365,7 +378,6 @@ impl KaniArgs {
                 String::new()
             };
 
-            // `tracing` is not a dependency of `kani-driver`, so we println! here instead.
             println!(
                 "Using concrete playback with --randomize-layout.\n\
                 The produced tests will have to be played with the same rustc arguments:\n\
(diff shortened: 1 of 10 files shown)


kani-driver/src/args.rs
```diff
@@ -377,29 +389,47 @@ impl KaniArgs {
...
+            ));
+        }
+        if self.concrete_playback.is_some() && self.jobs() != Some(1) {
+            // Concrete playback currently embeds a lot of assumptions about the order in which harnesses get called.
+            return Err(Error::with_description(
+                "Conflicting options: --concrete-playback isn't compatible with --jobs.",
+                ErrorKind::ArgumentConflict,
+            ));
         }
+        if self.jobs.is_some() && self.output_format != OutputFormat::Terse {
```
zhassan-aws, line 423: It would be more convenient to automatically switch to terse mode (which we want to switch to anyway) when jobs is specified.

zhassan-aws (commented):
Nice! Just have a couple of comments.

kani-driver/src/args.rs
tedinski: I thought about it, but:

1. The real problem is there's nowhere to put that code. e.g. this function doesn't take `&mut` and shouldn't.
2. But I also felt like being annoying for the moment is fine until we stabilize, and I'm planning on fixing this issue soon as well.

zhassan-aws: OK.

zhassan-aws (approved):

kani-driver/src/args.rs
```diff
@@ -226,6 +230,15 @@ impl KaniArgs {
             Some(DEFAULT_OBJECT_BITS)
         }
     }
+
+    /// Computes how many threads should be used to verify harnesses.
+    pub fn jobs(&self) -> Option<usize> {
+        match self.jobs {
+            None => Some(1),          // no argument, default 1
+            Some(None) => None,       // -j
+            Some(Some(x)) => Some(x), // -j=x
```
adpaco-aws, line 239: Isn't it worth to create an enum for all these options? Something like:
```
enum JobParalellism {
    Single,
    Multiple(Option<u32>),
}
```

kani-driver/src/args.rs
```diff
@@ -377,29 +389,47 @@ impl KaniArgs {
...
-        } else {
-            Ok(())
+            ));
+        }
+        if self.concrete_playback.is_some() && self.jobs() != Some(1) {
+            // Concrete playback currently embeds a lot of assumptions about the order in which harnesses get called.
+            return Err(Error::with_description(
+                "Conflicting options: --concrete-playback isn't compatible with --jobs.",
+                ErrorKind::ArgumentConflict,
+            ));
```
adpaco-aws, line 421: Can't these conflicts be specified as an annotation in `KaniArgs` fields? Not clear to me why we handle them here.

adpaco-aws (commented):
Can you consider this couple of comments in your next PR?

kani-driver/src/args.rs
tedinski: The double option in the args is a bit confusing imo, but is idiomatic there.

The option it returns exactly corresponds to calling `num_threads` on the thread pool (none means no, some means yes with this value), so I thought that was clear.

kani-driver/src/args.rs
tedinski: Only partially. This allows `-j1` to work fine for instance, and is more durable if we end up wanting to change the default to `num_cpus` later, but still need to require single threaded for concrete-playback. (Though hopefully we'll fix that first...)

(Basically this is a semantic check: note the `()` on `self.jobs()`, rather than just a syntactic check that `conflicts` would perform.)

Issue #1480, opened 2022-08-09
Refactor string formatting in new output parser


adpaco-aws:
The new parser contains a lot of code which has to do with formatting verification results, locations, etc. There are many improvements that could be made there in terms of code quality.

PR #1742, opened 2022-10-03
Report failure even when CBMC crashes, refactor CBMC parser/renderer

### LINKED ISSUE #1632: `process_cbmc_output` doesn't seem to check for `Child` process failure
As a consequence, CBMC crashing is not noticed properly.

The visible behavior looks like normal CBMC output, followed by the `Verification Time: 123s` line that Kani follows up with.

tedinski:
### Description of changes: 

This PR was a refactoring that also took on a bug fix.

The primary aim of this PR was two-fold:

1. Separate out `cbmc_property_renderer.rs` from `cbmc_output_parser.rs`. This creates a logical separation from "parsing results" and "rendering results" where the former is (mostly) concerned with CBMC, and that latter with how Kani should interpret those results. This isn't perfectly done (we have a field like `Property::reach` that's Kani-specific) but it's pretty close!
    * The key change here is the output parser now has this signature:
```
pub fn process_cbmc_output(
    mut process: Child,
    eager_filter: impl FnMut(ParserItem) -> Option<ParserItem>,
) -> Result<VerificationOutput>
```
This means we pass in the "filter/map" function from the outside, instead of it being a part of the parser code.

2. We "bubble upwards" the task of printing results, so that the results aren't printed from the parsing code, but instead results are returned up the function call stack to `harness_runner` and they're printed here. Consequently, `run_cbmc` returns a new `VerificationResult` type.

3. Secondarily, this means we now print results in one `println`, so thread interleaving will no longer separate the `Verification time: 2.34s` line from the prior results. (`println` takes a lock for output so it's atomic)
### Resolved issues:

* Fixes #1632 It turns out we were returning "verification status" of "failure", but we weren't PRINTING anything because that code was in the parser. Migrating the printing out necessitated fixing this. I initially also thought we might want to track the exit status of CBMC better, too, but our post-processing actually made this largely irrelevant except when printing a better error message when CBMC crashes.
* Partial progress on #1480 The `format_result` function still needs this, but I did some of it elsewhere because I was there.

### Testing:

* How is this change tested? exiting ci

* Is this a refactor change? mostly

### Checklist


kani-driver/src/call_cbmc.rs
```diff
@@ -164,6 +184,94 @@ impl KaniSession {
...
+                runtime: start_time.elapsed(),
+            }
+        }
+    }
+    pub fn mock_success() -> VerificationResult {
+        VerificationResult {
+            status: VerificationStatus::Success,
+            messages: None,
+            results: None,
+            exit_status: 42,
```
adpaco-aws, line 245: 😂

Make it a constant (e.g., `MOCK_EXIT_STATUS`) and use it here?

kani-driver/src/call_cbmc.rs
```diff
@@ -164,6 +184,94 @@ impl KaniSession {
...
+            // We don't have ANY messages to report - shouldn't really happen unless something went wrong
+            VerificationResult {
+                status: VerificationStatus::Failure,
+                messages: None,
+                results: None,
+                exit_status: output.process_status,
+                runtime: start_time.elapsed(),
+            }
+        }
+    }
```
adpaco-aws, line 239: Add newline after this.

kani-driver/src/call_cbmc.rs
```diff
@@ -164,6 +184,94 @@ impl KaniSession {
...
+    pub fn mock_success() -> VerificationResult {
+        VerificationResult {
+            status: VerificationStatus::Success,
+            messages: None,
+            results: None,
+            exit_status: 42,
+            runtime: Duration::from_secs(0),
+        }
+    }
```
adpaco-aws, line 248: Add newline after this.

kani-driver/src/call_cbmc.rs
```diff
@@ -164,6 +184,94 @@ impl KaniSession {
...
+    ///   1. Examining the actual results of CBMC properties.
+    ///       (CBMC will regularly report "failure" but that's just our cover checks.)
+    ///   2. Positively checking for the presence of results.
+    ///       (Do not mistake lack of results for success: report it as failure.)
+    fn from(output: VerificationOutput, start_time: Instant) -> VerificationResult {
+        if let Some(mut items) = output.processed_items {
+            let result_idx = items.iter().position(|x| matches!(x, ParserItem::Result { .. }));
+            if let Some(result_idx) = result_idx {
+                let result = items.remove(result_idx);
+                if let ParserItem::Result { result } = result {
```
adpaco-aws, line 201: This is quite hard to read. Can we refactor into small functions (find, extract) so it's easier to read?

kani-driver/src/call_cbmc.rs
```diff
@@ -164,6 +184,94 @@ impl KaniSession {
...
+                        results: Some(result),
+                        exit_status: output.process_status,
+                        runtime: start_time.elapsed(),
+                    }
+                } else {
+                    unreachable!() // We filtered for this to be true
+                }
+            } else {
+                // We never got results from CBMC - also shouldn't really happen unless something went wrong
+                VerificationResult {
```
adpaco-aws, line 221: Do we want a constructor?

kani-driver/src/call_cbmc.rs
```diff
@@ -164,6 +184,94 @@ impl KaniSession {
...
+                        VerificationStatus::Success
+                    } else {
+                        VerificationStatus::Failure
+                    };
+                    VerificationResult {
+                        status,
+                        messages: Some(items),
+                        results: Some(result),
+                        exit_status: output.process_status,
+                        runtime: start_time.elapsed(),
```
adpaco-aws, line 214: Store this in a variable at the beginning and use that.

kani-driver/src/call_cbmc.rs
```diff
@@ -164,6 +184,94 @@ impl KaniSession {
...
+    ///   2. Positively checking for the presence of results.
+    ///       (Do not mistake lack of results for success: report it as failure.)
+    fn from(output: VerificationOutput, start_time: Instant) -> VerificationResult {
+        if let Some(mut items) = output.processed_items {
+            let result_idx = items.iter().position(|x| matches!(x, ParserItem::Result { .. }));
+            if let Some(result_idx) = result_idx {
+                let result = items.remove(result_idx);
+                if let ParserItem::Result { result } = result {
+                    let number_failed_properties =
+                        result.iter().filter(|prop| prop.status == CheckStatus::Failure).count();
```
adpaco-aws, line 203: I think a function for this already existed.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -603,534 +445,44 @@ impl<'a, 'b> Iterator for Parser<'a, 'b> {
...
+    let processed_items: Vec<_> = parser.filter_map(eager_filter).collect();
+    // This will get us the process's exit code
+    let status = process.wait()?;
+
+    let process_status = match (status.code(), status.signal()) {
+        // normal unix exit codes (cbmc uses currently 0-10)
+        // https://github.com/diffblue/cbmc/blob/develop/src/util/exit_codes.h
+        (Some(x), _) => x,
+        // process exited with signal (e.g. OOM-killed)
+        // bash/zsh have a convention for translating signal number to exit code:
```
adpaco-aws, line 479: Can you link this convention?

kani-driver/src/call_cbmc.rs
```diff
@@ -164,6 +184,94 @@ impl KaniSession {
...
+    pub fn render(&self, output_format: &OutputFormat) -> String {
+        if let Some(results) = &self.results {
+            let show_checks = matches!(output_format, OutputFormat::Regular);
+            let mut result = format_result(results, show_checks);
+            writeln!(result, "Verification Time: {}s", self.runtime.as_secs_f32()).unwrap();
+            result
+        } else {
+            let verification_result = console::style("FAILED").red();
+            format!(
+                "\nCBMC failed with status {}\nVERIFICATION:- {}\n",
```
adpaco-aws, line 268: I don't know if it makes sense to provide a verification result in this case... We could say something like "verification couldn't be performed".

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -0,0 +1,643 @@
...
+const REACH_CHECK_DESC: &str = "[KANI_REACHABILITY_CHECK]";
+
+impl ParserItem {
+    /// Determines if an item must be skipped or not.
+    fn must_be_skipped(&self) -> bool {
+        matches!(&self, ParserItem::Message { message_text, .. } if message_text.starts_with("Building error trace") || message_text.starts_with("VERIFICATION"))
+    }
+}
+
+pub fn kani_property_filter(
```
adpaco-aws, line 163: This name is not very descriptive. It's filtering more than just properties, right?

kani-driver/src/call_cbmc.rs
tedinski: I didn't think a constant made sense, but I did add comments and audit how they end up used.

This did make me notice that the new results printing should have more exceptions: in addition to `quiet` it also suppresses on `dry-run` and `output-format=old`.

kani-driver/src/call_cbmc.rs
tedinski: I don't think it'd help. After looking through this code more carefully this morning, I was able to simplify this function a lot though.

kani-driver/src/call_cbmc.rs
tedinski: I thought it best to keep a very simple/consistent response for this. The verification did fail, the reasons why will be above that line, just like normal output.

PR #1754, opened 2022-10-05
Use CustomCoerce to get info about the unsized cast

### LINKED ISSUE #1694: Refactor how we are collecting unsized coercion
1. The code is rather complicated.
2. This might not work for custom smart pointers.

We should implement some logic similar to `CustomCoerceUnsized`. See https://doc.rust-lang.org/std/ops/trait.CoerceUnsized.html for more details.

celinval:
### Description of changes: 

Implement logic to extract unsized cast info that takes CustomCoerce trait into account. That seems to be the standard way the compiler defines structure that can be coerced. This also allow us to handle custom smart pointers.

### Resolved issues:

Fixes #1694.

### Related RFC:

Optional #1588 

### Call-outs:

I'm forcing this one test to run under the MIR linker since it had an issue before this change.

### Testing:

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/reachability.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/reachability.rs
@@ -453,19 +453,30 @@ fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) ->
     }
 }
 
-/// Extract the pair (from_ty, to_ty) for a unsized cast.
+/// Extract the pair (`T`, `U`) for a unsized coercion where type `T` implements `Unsize<U>`.
+/// I.e., `U` is either a trait or a slice.
+/// For more details, please refer to:
+/// <https://doc.rust-lang.org/reference/type-coercions.html#unsized-coercions>
+///
+/// This is used to determine the vtable implementation that must be tracked by the fat pointer.
 ///
 /// For example, if `&u8` is being converted to `&dyn Debug`, this method would return:
 /// `(u8, dyn Debug)`.
 ///
-/// This method also handles nested cases and `std` smart pointers. E.g.:
-///
-/// Conversion between `Rc<Wrapper<String>>` into `Rc<Wrapper<dyn Debug>>` should return:
-/// `(String, dyn Debug)`
-///
-/// TODO: Do we need to handle &Wrapper<dyn T1> to &dyn T2 or is that taken care of with super
-/// trait handling?
-/// <https://github.com/model-checking/kani/issues/1692>
+/// There are a few interesting cases (references / pointers are handled the same way):
+/// 1. Coercion between `&T` to `&U`.
+///    - This is the base case.
+///    - In this case, we extract the type that is pointed to.
+/// 2. Coercion between smart pointers like `Rc<T>` to `Rc<U>`.
+///    - Smart pointers implement the `CoerceUnsize` trait.
+///    - Use CustomCoerceUnsized information to traverse the smart pointer structure and find the
+///      underlying pointer.
+///    - Use base case to extract `T` and `U`.
+/// 3. Coercion between `&Wrapper<T>` to `&Wrapper<U>`.
+///    - Apply base case to extract the pair `(Wrapper<T>, Wrapper<U>)`.
+///    - Extract the tail element of the struct which are of type `T` and `U`, respectively.
+/// 4. Coercion between smart pointers of wrapper structs.
+///    - Apply the logic from item 2 then item 3.
 fn extract_trait_casting<'tcx>(
     tcx: TyCtxt<'tcx>,
     src_ty: Ty<'tcx>,
@@ -475,24 +486,27 @@ fn extract_trait_casting<'tcx>(
     let mut src_inner_ty = src_ty;
     let mut dst_inner_ty = dst_ty;
     (src_inner_ty, dst_inner_ty) = loop {
-        // Extract pointee types that form the base of the conversion.
+        // Extract the pointee types from pointers (including smart pointers) that form the base of
+        // the conversion.
         match (&src_inner_ty.kind(), &dst_inner_ty.kind()) {
             (&ty::Adt(src_def, src_substs), &ty::Adt(dst_def, dst_substs))
                 if !src_def.is_box() || !dst_def.is_box() =>
             {
+                // Handle smart pointers by using CustomCoerceUnsized to find the field being
+                // coerced.
                 assert_eq!(src_def, dst_def);
+                let src_fields = &src_def.non_enum_variant().fields;
+                let dst_fields = &dst_def.non_enum_variant().fields;
+                assert_eq!(src_fields.len(), dst_fields.len());
 
                 let CustomCoerceUnsized::Struct(coerce_index) =
                     custom_coerce_unsize_info(tcx, src_inner_ty, dst_inner_ty);
-
-                let src_fields = &src_def.non_enum_variant().fields;
-                let dst_fields = &dst_def.non_enum_variant().fields;
-
-                assert!(coerce_index < src_fields.len() && src_fields.len() == dst_fields.len());
+                assert!(coerce_index < src_fields.len());
 
                 src_inner_ty = src_fields[coerce_index].ty(tcx, src_substs);
                 dst_inner_ty = dst_fields[coerce_index].ty(tcx, dst_substs);
             }
+            // Base case is always a pointer (Box, raw_pointer or reference).
             _ => break (extract_pointer(src_inner_ty), extract_pointer(dst_inner_ty)),
         }
     };
--- a/kani-compiler/src/codegen_cprover_gotoc/reachability.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/reachability.rs
@@ -16,6 +16,7 @@
 use rustc_data_structures::fingerprint::Fingerprint;
 use rustc_data_structures::fx::FxHashSet;
 use rustc_data_structures::stable_hasher::{HashStable, StableHasher};
+use rustc_hir::lang_items::LangItem;
 use rustc_middle::mir::interpret::{AllocId, ConstValue, ErrorHandled, GlobalAlloc, Scalar};
 use rustc_middle::mir::mono::MonoItem;
 use rustc_middle::mir::visit::Visitor as MirVisitor;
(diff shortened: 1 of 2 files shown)


zhassan-aws (approved):
LGTM. Would be useful if someone else can take a look as well.

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -447,7 +446,7 @@ fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) ->
             // Need to compile std with --always-encode-mir first though.
             // https://github.com/model-checking/kani/issues/1605
             // assert!(tcx.is_mir_available(def_id), "no MIR available for {:?}", def_id);
-            //(!tcx.is_mir_available(def_id)).then(|| warn!(?def_id, "Missing MIR"));
+            (!tcx.is_mir_available(def_id)).then(|| warn!(?def_id, "Missing MIR"));
```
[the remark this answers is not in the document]
celinval: Great question. I'm actually removing this code in my other PR: https://github.com/model-checking/kani/pull/1761

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -457,67 +456,97 @@ fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) ->
...
+            {
+                assert_eq!(src_def, dst_def);
+
+                let CustomCoerceUnsized::Struct(coerce_index) =
+                    custom_coerce_unsize_info(tcx, src_inner_ty, dst_inner_ty);
+
+                let src_fields = &src_def.non_enum_variant().fields;
+                let dst_fields = &dst_def.non_enum_variant().fields;
+
+                assert!(coerce_index < src_fields.len() && src_fields.len() == dst_fields.len());
```
tedinski, line 494: The order of operations here confuses me. Can you split this assert into two, then move the part (and lets) that don't reference `coerce_index` above its let?

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -457,67 +456,97 @@ fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) ->
...
-    };
-
+) -> (Ty<'tcx>, Ty<'tcx>) {
     trace!(?dst_ty, ?src_ty, "find_trait_conversion");
-    let dst_ty_inner = extract_pointer(tcx, dst_ty);
-    let src_ty_inner = extract_pointer(tcx, src_ty);
+    let mut src_inner_ty = src_ty;
+    let mut dst_inner_ty = dst_ty;
+    (src_inner_ty, dst_inner_ty) = loop {
+        // Extract pointee types that form the base of the conversion.
```
tedinski, line 481: This comment need improving. Is this right?

> Descend through wrapper types like `Rc` to get to the types (which are always a pointer type??) that are actually being converted

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -457,67 +456,97 @@ fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) ->
...
+            _ => break (extract_pointer(src_inner_ty), extract_pointer(dst_inner_ty)),
+        }
+    };
 
-    trace!(?dst_ty_inner, ?src_ty_inner, "find_trait_conversion");
-    (has_vtable_metadata(dst_ty_inner) && !has_vtable_metadata(src_ty_inner)).then(|| {
-        let param_env = ParamEnv::reveal_all();
-        tcx.struct_lockstep_tails_erasing_lifetimes(src_ty_inner, dst_ty_inner, param_env)
-    })
+    if has_vtable_metadata(tcx, src_inner_ty) {
```
tedinski, line 503: This needs a comment. Also is `has_vtble_metadata` the right name here? That might be what the function literally checks, but are you using that instead to answer a different question maybe?

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -457,67 +456,97 @@ fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) ->
...
-/// This function will return None if this is not a unsizing coercion from concrete to trait.
+/// Extract the pair (from_ty, to_ty) for a unsized cast.
 ///
 /// For example, if `&u8` is being converted to `&dyn Debug`, this method would return:
-/// `Some(u8, dyn Debug)`.
+/// `(u8, dyn Debug)`.
 ///
 /// This method also handles nested cases and `std` smart pointers. E.g.:
 ///
 /// Conversion between `Rc<Wrapper<String>>` into `Rc<Wrapper<dyn Debug>>` should return:
```
tedinski, line 466: Some more context about how this can happen would be appreciated here. I don't think Rust lets you write this kind of conversion, so is this only something that can happen in the context of a dyn-ptr to dyn-ptr cast?

Basically, when does a "coerce unsized" even happen?

celinval: Not sure I understand what you mean by "I don't think Rust let's you write this type of conversion."

But I will try to improve the doc here.

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
celinval: It's always pointer (or reference) type because this is an unsized coercion, i.e., the destination place includes an unsized pointer / reference. Unsized types can only be accessed via pointers / references.

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
celinval: yeah, makes sense.

celinval: I tried but I'm not thrilled by the new name. Let me know if you have any suggestions.

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -454,67 +453,111 @@ fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) ->
...
+        match (&src_inner_ty.kind(), &dst_inner_ty.kind()) {
+            (&ty::Adt(src_def, src_substs), &ty::Adt(dst_def, dst_substs))
+                if !src_def.is_box() || !dst_def.is_box() =>
+            {
+                // Handle smart pointers by using CustomCoerceUnsized to find the field being
+                // coerced.
+                assert_eq!(src_def, dst_def);
+                let src_fields = &src_def.non_enum_variant().fields;
+                let dst_fields = &dst_def.non_enum_variant().fields;
+                assert_eq!(src_fields.len(), dst_fields.len());
```
danielsn, line 500: Is there a way to check that the fields also align in some way?

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -454,67 +453,111 @@ fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) ->
...
+
+                let CustomCoerceUnsized::Struct(coerce_index) =
+                    custom_coerce_unsize_info(tcx, src_inner_ty, dst_inner_ty);
+                assert!(coerce_index < src_fields.len());
+
+                src_inner_ty = src_fields[coerce_index].ty(tcx, src_substs);
+                dst_inner_ty = dst_fields[coerce_index].ty(tcx, dst_substs);
+            }
+            // Base case is always a pointer (Box, raw_pointer or reference).
```
danielsn, line 509: Should we explicitly assert that?

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -454,67 +453,111 @@ fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) ->
...
     } else {
-        ReceiverIter { tcx, curr: typ }.last().unwrap().1
+        unreachable!("Expected pointer type, found: {:?}", typ);
+    }
+}
+
+/// Check if the type has metadata.
+fn has_vtable_metadata<'tcx>(tcx: TyCtxt<'tcx>, mir_type: Ty<'tcx>) -> bool {
+    let (metadata, _) = mir_type.ptr_metadata_ty(tcx, |ty| ty);
+    metadata != tcx.types.unit && metadata != tcx.types.usize
```
danielsn, line 537: I know these are currently the only three kinds of metadata, but is it possible to future proof this in case new types are invented?

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -454,67 +453,111 @@ fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) ->
...
+}
+
+/// Get information about an unsized coercion.
+/// This code was extracted from `rustc_monomorphize` crate.
```
danielsn, line 541: Link to where, and which commit?

danielsn: Also, why can't we use their implementation? Is it private?

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -16,26 +16,61 @@
...
+    sorted_items
+}
+
+/// Collect all items in the crate that matches the given predicate.
+pub fn filter_crate_items<F>(tcx: TyCtxt, predicate: F) -> Vec<MonoItem>
+where
+    F: FnMut(TyCtxt, DefId) -> bool,
+{
+    // Filter proof harnesses.
+    let mut filter = predicate;
```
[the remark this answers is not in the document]
celinval: It moves the predicate into a mutable variable. I thought that was cleaner than making the argument itself `mut`.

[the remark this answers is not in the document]
celinval: BTW, I haven't changed this code. I just moved the public functions to the top of the file.

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -16,26 +16,61 @@
...
+    tcx.hir_crate_items(())
```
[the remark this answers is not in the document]
celinval: Because that's how you get all the elements from a crate.

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
celinval: No, I don't think you can do it here. These are non-monomorphized fields. I also feel like the first assert is probably enough, i.e., the definitions are equal. I kept the second one to be similar to the rustc code.

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
celinval: sure

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
celinval: what do you have in mind?

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
celinval: Yes, it is private. Since the redundant code here is very small, I didn't think it was worth bothering the compiler team to make it public.

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
celinval: I changed the code a bit so the check is more explicit.

kani-compiler/src/codegen_cprover_gotoc/reachability.rs
celinval: Looking back, I think this check might be redundant. Let me see if I can remove this code for now. FYI, we do have the exact same check here:

https://github.com/model-checking/kani/blob/19add43fd2f2f3cbc2f6ef25fef4d12d1b0dc39d/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs#L1968-L1971

tedinski (approved):
I have questions, but they're more over some confusions I think I might have about when this case happens in Rust code. I think the code itself looks good now.

Is there a spot in the test that you can point to saying "here's where this code makes a difference"? That might help me.

celinval:
> I have questions, but they're more over some confusions I think I might have about when this case happens in Rust code. I think the code itself looks good now.
> 
> Is there a spot in the test that you can point to saying "here's where this code makes a difference"? That might help me.

I'm creating some tests for this that I think might help. I'll create a PR shortly for those.

Issue #1448, opened 2022-08-03
cargo kani --tests fails on crates with integration tests


zhassan-aws:
Steps to reproduce:
1. cargo new foo
2. cd foo
3. Append the following to `src/main.rs`:
```rust
#[cfg(test)]
mod tests {
    #[test]
    #[kani::proof]
    fn check() {
        assert!(1 + 1 == 2);
    }
}
```
4. Run `cargo kani --tests`. Verification is successful:
```

SUMMARY: 
 ** 0 of 1 failed

VERIFICATION:- SUCCESSFUL
```
5. Under the package root directory, create a `tests` directory, and place a `my_test.rs` file inside with the following:
```rust
#[test]
fn main() {
    assert!(2 + 2 == 4);
}
```
6. Run `cargo clean`
7. Run `cargo kani --tests` again

This fails with the following:
```
$ cargo kani --tests
   Compiling foo v0.1.0 (/home/ubuntu/examples/integ_tests/foo)
    Finished test [unoptimized + debuginfo] target(s) in 0.26s
  Executable unittests src/main.rs (target/x86_64-unknown-linux-gnu/debug/deps/foo-61c8a3392540768f)
  Executable tests/my_test.rs (target/x86_64-unknown-linux-gnu/debug/deps/my_test-3ee8fa6d631c7ebc)
/home/ubuntu/git/kani/library/std/src/lib.rs: In function 'main':
/home/ubuntu/git/kani/library/std/src/lib.rs:41:9: error: 
reason for conflict at #this: number of members is different (2/0)

struct _15120766920117065333 {
  struct _15678888388531287603 ** data;
  unsigned long int len;
}
struct Unit {
}
error: conflicting types for variable 'main::1::var_1'
old definition in module '' file /home/ubuntu/examples/integ_tests/foo/src/main.rs line 1 column 1 function main
struct _15120766920117065333 {
  struct _15678888388531287603 ** data;
  unsigned long int len;
}
new definition in module '' file /home/ubuntu/git/kani/library/std/src/lib.rs line 41 column 9 function main
struct Unit {
}
         kani::assert($cond, concat!("assertion failed: ", stringify!($cond)));
/home/ubuntu/examples/integ_tests/foo/tests/my_test.rs: In function 'main':
/home/ubuntu/examples/integ_tests/foo/tests/my_test.rs:3:13: error: error: conflicting types for variable 'main::1::var_2'
[...]

PR #1756, opened 2022-10-06
Introduce prototype `cargo-kani assess` feature to help find places to start writing proofs

### LINKED ISSUE #1497: cargo kani assess: unimplemented features report
In https://github.com/model-checking/kani/pull/1400 we got a nice look at what features are unimplemented in the top-100 crates.

I'd like to make this less hacky and integrate it as a part of `kani-driver`.

Running `cargo kani assess` should give a report like the one there for the crate it's being run on and all its dependencies.

### LINKED ISSUE #1498: cargo kani assess: experimental test case runner
#1497 will give a basic report on features Kani does not support in a crate and its dependencies.

We should be able to try running a crate's existing test suite to measure how many of those features are actually hit.

Further, we should be able to track "failure reasons" for other "test case fails to run under Kani" situations. For instance, if it clearly fails because of the stdlib linking issue.

tedinski:
### Description of changes: 

This feature is far from something we can recommend to customers yet, but it should be quite useful for us to both (1) start helping customers find proofs to write and (2) start finding the other bugs that are standing in the way of applying Kani to various crates.

This PR includes several changes:

1. We now track unsupported features in `kani-metadata.json`
2. We now track test harnesses (when build with `--tests`) in `kani-metadata.json`
3. The assess feature itself, which prints three tables of summary information:
   1. Unsupported features table
   2. Reason for test harness failure
   3. Candidate for proof harness
4. (Temporarily) a `--assess` flag, purely for use with `compiletest` in CI, along with a simple test to exercise the code.

I tried assess on the `rand` crate and got summary information like the following:

```
===================================================
 Unsupported feature        |   Crates | Instances 
                            | impacted |    of use 
----------------------------+----------+-----------
 'simd_or' intrinsic        |        4 |         5 
 'simd_gt' intrinsic        |        3 |         3 
 'simd_eq' intrinsic        |        3 |         3 
 try                        |        2 |        17 
 drop_in_place              |        2 |         2 
 'simd_and' intrinsic       |        1 |         1 
 'simd_xor' intrinsic       |        1 |         3 
 'simd_shuffle16' intrinsic |        1 |         2 
 'simd_insert' intrinsic    |        1 |         5 
 'simd_shuffle8' intrinsic  |        1 |         6 
 'simd_shuffle4' intrinsic  |        1 |         8 
 'simd_shuffle2' intrinsic  |        1 |         2 
 'simd_extract' intrinsic   |        1 |         2 
 'powf64' intrinsic         |        1 |         1 
 'simd_add' intrinsic       |        1 |         3 
===================================================
================================================
 Reason for failure           | Number of tests 
------------------------------+-----------------
 unwind                       |              61 
 success                      |               6 
 assertion                    |               4 
 pointer_dereference + unwind |               2 
 assertion + overflow         |               2 
================================================
===============================================================================================================
 Candidate for proof harness                                          | Location                                            
----------------------------------------------------------------------+----------------------------------------
 distributions::float::tests::f64_edge_cases                          | rand/src/distributions/float.rs:226    
 distributions::float::tests::f32_edge_cases                          | rand/src/distributions/float.rs:184    
 distributions::integer::tests::test_integers                         | rand/src/distributions/integer.rs:171  
 distributions::other::tests::test_misc                               | rand/src/distributions/other.rs:284    
 distributions::uniform::tests::test_uniform_from_std_range_inclusive | rand/src/distributions/uniform.rs:1563 
 distributions::uniform::tests::test_uniform_from_std_range           | rand/src/distributions/uniform.rs:1553 
===============================================================================================================
real    7m51.665s
```

A manual inspection of those tests suggests at least half of them are definitely candidates for turning into proofs. (The other half required more investigation, as they're generated by macros. Only one was in the "probably not" camp.)

Trying on a few other crates reveals several issues we might want to prioritize:

* Several fail to link likely due to https://github.com/model-checking/kani/issues/1448 or general "multiple binaries linked together" problems
* I saw some different goto type checking errors as well, but these may have just been a different manifestation of the above.
* A remarkable number of crates failed to build properly and I haven't spent time yet trying to understand why. While we have examined normal builds (e.g. `cargo build --workspace --all-features`) we haven't tried this with `--tests`. Perhaps we should!

Of the crates in the Rust Playground that we successfully build for, we get these results (just unsupported features):

```
Analyzed 222 crates
===========================================================
 Unsupported feature                |   Crates | Instances 
                                    | impacted |    of use 
------------------------------------+----------+-----------
 'simd_eq' intrinsic                |       19 |        21 
 'simd_or' intrinsic                |       19 |        21 
 Constant slice value with 2+ bytes |       16 |        78 
 'simd_gt' intrinsic                |       15 |        15 
 try                                |        9 |       159 
 drop_in_place                      |        7 |         7 
 'simd_shuffle2' intrinsic          |        5 |        11 
 'simd_shuffle4' intrinsic          |        5 |        26 
 drop_in_place call                 |        4 |         7 
 'sqrtf64' intrinsic                |        4 |         4 
 'powif64' intrinsic                |        4 |         4 
 'simd_add' intrinsic               |        4 |         8 
 'simd_shuffle16' intrinsic         |        4 |         8 
 'simd_shuffle8' intrinsic          |        4 |        13 
 PointerCast::ClosureFnPointer      |        3 |         7 
 'simd_xor' intrinsic               |        3 |         7 
 TerminatorKind::InlineAsm          |        3 |         4 
 'expf64' intrinsic                 |        3 |         3 
 drop unsized struct                |        3 |         3 
 'simd_and' intrinsic               |        3 |         7 
 'simd_extract' intrinsic           |        3 |         6 
 'simd_shuffle32' intrinsic         |        2 |        50 
 'simd_mul' intrinsic               |        2 |         3 
 float_to_int_unchecked             |        2 |        24 
 'logf64' intrinsic                 |        2 |         2 
 'sqrtf32' intrinsic                |        1 |         1 
 'powif32' intrinsic                |        1 |         1 
 'simd_insert' intrinsic            |        1 |         5 
 simd_fma                           |        1 |         2 
 'log2f64' intrinsic                |        1 |         1 
 simd_saturating_sub                |        1 |         1 
 simd_saturating_add                |        1 |         1 
===========================================================
````

### Next steps

This is the very start of this feature.

1. I think the biggest priority is to switch away from `unwind` to force termination, and add real timeouts. This should eliminate `unwind` as a failure reasons, and probably not by entirely turning them into `timeout` either. The results should become more interesting. :)
2. Our ability to build under `--tests` needs to become more reliable. Fixing that bug linked above, for instance.
3. The feature needs improved documentation. The _meaning_ of these tables isn't fully clear to a curious user. While some of them have clear meaning to us (and this is, for now, a feature for us to experiment with), they won't be clear to customers. (For instance, "unwind" means what? Similarly "assertion" or "+" in that table.) This should be prioritized before we consider stabilization or making this customer-facing.)
4. I think we should add more support for detecting kinds of failures. If things fail for goto-cc linking problem, currently it's up to you to read that output. Making it more machine-readable will help for:
5. I think I should add a json (machine-readable) output for this, then work on a tool to automatically apply assess to a large number of crates, aggregating the results. This will help with some of the difficulty in finding appropriate target crates to analyze.
6. [Future] We should consider a more in-depth HTML report instead of just console. Users may want to explore some of this data in more depth.

More experimentation with more crates will be helpful in prioritizing things.

### Resolved issues:

Resolves #1497 
Resolves #1498 

### Call-outs:

* Interaction with the MIR linker hasn't been addressed yet.

### Testing:

* How is this change tested? new test

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
@@ -123,7 +123,7 @@ impl<'tcx> GotocCtx<'tcx> {
             self.symbol_table.update_fn_declaration_with_definition(&name, body);
 
             self.handle_kanitool_attributes();
-            self.record_test_metadata();
+            self.record_test_harness_metadata();
         }
         self.reset_current_fn();
     }
@@ -317,8 +317,14 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
-    /// Records test harness closures for KaniMetadata
-    fn record_test_metadata(&mut self) {
+    /// We record test harness information in kani-metadata, just like we record
+    /// proof harness information. This is used to support e.g. cargo-kani assess.
+    ///
+    /// Note that we do not actually spot the function that was annotated by `#[test]`
+    /// but instead the closure that gets put into the "test description" that macro
+    /// expands into. (See comment below) This ends up being preferrable, actually,
+    /// as it add asserts for tests that return `Result` types.
+    fn record_test_harness_metadata(&mut self) {
         let def_id = self.current_fn().instance().def_id();
         if def_id.is_local() {
             let local_def_id = def_id.expect_local();
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
@@ -321,7 +321,38 @@ impl<'tcx> GotocCtx<'tcx> {
     ///
     /// Handle all attributes i.e. `#[kani::x]` (which kani_macros translates to `#[kanitool::x]` for us to handle here)
     fn handle_kanitool_attributes(&mut self) {
-        let all_attributes = self.tcx.get_attrs_unchecked(self.current_fn().instance().def_id());
+        let def_id = self.current_fn().instance().def_id();
+        let all_attributes = self.tcx.get_attrs_unchecked(def_id);
+
+        if def_id.is_local() {
+            let local_def_id = def_id.expect_local();
+            let hir_id = self.tcx.hir().local_def_id_to_hir_id(local_def_id);
+
+            // We want to detect the case where we're codegen'ing the closure inside what test "descriptions"
+            // are macro-expanded to:
+            //
+            // #[rustc_test_marker]
+            // pub const check_2: test::TestDescAndFn = test::TestDescAndFn {
+            //     desc: ...,
+            //     testfn: test::StaticTestFn(|| test::assert_test_result(check_2())),
+            // };
+
+            // The parent item of the closure appears to reliably be the `const` declaration item.
+            let parent_id = self.tcx.hir().get_parent_item(hir_id);
+            let attrs = self.tcx.get_attrs_unchecked(parent_id.to_def_id());
+
+            if self.tcx.sess.contains_name(attrs, rustc_span::symbol::sym::rustc_test_marker) {
+                let loc = self.codegen_span(&self.current_fn().mir().span);
+                self.test_harnesses.push(HarnessMetadata {
+                    pretty_name: self.current_fn().readable_name().to_owned(),
+                    mangled_name: self.current_fn().name(),
+                    original_file: loc.filename().unwrap(),
+                    original_start_line: loc.start_line().unwrap() as usize,
+                    original_end_line: loc.end_line().unwrap() as usize,
+                    unwind_value: None,
+                })
+            }
+        }
         let (proof_attributes, other_attributes) = partition_kanitool_attributes(all_attributes);
         if !proof_attributes.is_empty() {
             self.create_proof_harness(other_attributes);
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
@@ -123,6 +123,7 @@ impl<'tcx> GotocCtx<'tcx> {
             self.symbol_table.update_fn_declaration_with_definition(&name, body);
 
             self.handle_kanitool_attributes();
+            self.record_test_metadata();
         }
         self.reset_current_fn();
     }
@@ -316,14 +317,9 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
-    /// This updates the goto context with any information that should be accumulated from a function's
-    /// attributes.
-    ///
-    /// Handle all attributes i.e. `#[kani::x]` (which kani_macros translates to `#[kanitool::x]` for us to handle here)
-    fn handle_kanitool_attributes(&mut self) {
+    /// Records test harness closures for KaniMetadata
+    fn record_test_metadata(&mut self) {
         let def_id = self.current_fn().instance().def_id();
-        let all_attributes = self.tcx.get_attrs_unchecked(def_id);
-
         if def_id.is_local() {
             let local_def_id = def_id.expect_local();
             let hir_id = self.tcx.hir().local_def_id_to_hir_id(local_def_id);
@@ -353,6 +349,15 @@ impl<'tcx> GotocCtx<'tcx> {
                 })
             }
         }
+    }
+
+    /// This updates the goto context with any information that should be accumulated from a function's
+    /// attributes.
+    ///
+    /// Handle all attributes i.e. `#[kani::x]` (which kani_macros translates to `#[kanitool::x]` for us to handle here)
+    fn handle_kanitool_attributes(&mut self) {
+        let def_id = self.current_fn().instance().def_id();
+        let all_attributes = self.tcx.get_attrs_unchecked(def_id);
         let (proof_attributes, other_attributes) = partition_kanitool_attributes(all_attributes);
         if !proof_attributes.is_empty() {
             self.create_proof_harness(other_attributes);
--- a/kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
@@ -124,6 +124,8 @@ impl CodegenBackend for GotocCodegenBackend {
         // Print compilation report.
         print_report(&gcx, tcx);
 
+        let unsupported_features = gcx.unsupported_metadata();
+
         // perform post-processing symbol table passes
         let passes = self.queries.get_symbol_table_passes();
         let symtab = symtab_transformer::do_passes(gcx.symbol_table, &passes);
@@ -139,7 +141,11 @@ impl CodegenBackend for GotocCodegenBackend {
             None
         };
 
-        let metadata = KaniMetadata { proof_harnesses: gcx.proof_harnesses };
+        let metadata = KaniMetadata {
+            proof_harnesses: gcx.proof_harnesses,
+            unsupported_features,
+            test_harnesses: gcx.test_harnesses,
+        };
 
         // No output should be generated if user selected no_codegen.
         if !tcx.sess.opts.unstable_opts.no_codegen && tcx.sess.opts.output_types.should_codegen() {
--- a/kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
@@ -102,6 +102,11 @@ impl<'tcx> GotocCtx<'tcx> {
         self.current_fn.as_mut().unwrap()
     }
 
+    /// Maps the goto-context "unsupported features" data into the
+    /// KaniMetadata "unsupported features" format.
+    ///
+    /// These are different because the KaniMetadata is a flat serializable list,
+    /// while we need a more richly structured HashMap in the goto context.
     pub(crate) fn unsupported_metadata(&self) -> Vec<UnsupportedFeature> {
         self.unsupported_constructs
             .iter()
--- a/kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
@@ -21,7 +21,7 @@ use cbmc::goto_program::{DatatypeComponent, Expr, Location, Stmt, Symbol, Symbol
 use cbmc::utils::aggr_tag;
 use cbmc::InternedString;
 use cbmc::{MachineModel, RoundingMode};
-use kani_metadata::HarnessMetadata;
+use kani_metadata::{HarnessMetadata, UnsupportedFeature};
 use kani_queries::{QueryDb, UserInput};
 use rustc_data_structures::fx::FxHashMap;
 use rustc_data_structures::owning_ref::OwningRef;
@@ -59,6 +59,7 @@ pub struct GotocCtx<'tcx> {
     pub current_fn: Option<CurrentFnCtx<'tcx>>,
     pub type_map: FxHashMap<InternedString, Ty<'tcx>>,
     pub proof_harnesses: Vec<HarnessMetadata>,
+    pub test_harnesses: Vec<HarnessMetadata>,
     /// a global counter for generating unique IDs for checks
     pub global_checks_count: u64,
     /// A map of unsupported constructs that were found while codegen
@@ -84,6 +85,7 @@ impl<'tcx> GotocCtx<'tcx> {
             current_fn: None,
             type_map: FxHashMap::default(),
             proof_harnesses: vec![],
+            test_harnesses: vec![],
             global_checks_count: 0,
             unsupported_constructs: FxHashMap::default(),
         }
@@ -98,6 +100,24 @@ impl<'tcx> GotocCtx<'tcx> {
 
     pub fn current_fn_mut(&mut self) -> &mut CurrentFnCtx<'tcx> {
         self.current_fn.as_mut().unwrap()
+    }
+
+    pub(crate) fn unsupported_metadata(&self) -> Vec<UnsupportedFeature> {
+        self.unsupported_constructs
+            .iter()
+            .map(|(construct, location)| UnsupportedFeature {
+                feature: construct.to_string(),
+                locations: location
+                    .iter()
+                    .map(|l| {
+                        (
+                            l.filename().unwrap_or_default(),
+                            l.start_line().map(|x| x.to_string()).unwrap_or_default(),
+                        )
+                    })
+                    .collect(),
+            })
+            .collect()
     }
 }
 
--- a/kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
@@ -115,6 +115,9 @@ impl<'tcx> GotocCtx<'tcx> {
                 locations: location
                     .iter()
                     .map(|l| {
+                        // We likely (and should) have no instances of
+                        // calling `codegen_unimplemented` without file/line.
+                        // So while we map out of `Option` here, we expect them to always be `Some`
                         (
                             l.filename().unwrap_or_default(),
                             l.start_line().map(|x| x.to_string()).unwrap_or_default(),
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

**Comment by celinval** [comment]
Quick questions about usability:
1. What does "success" and "assertion" means in the "Reason for failure" table?
2. Can we enforce `--enable-unstable` for the new subcommand?
3. What happens if the user given other `kani` commands to the assess subcommand?

File: kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs [hunk 1]
```diff
@@ -321,7 +321,38 @@ impl<'tcx> GotocCtx<'tcx> {
     ///
     /// Handle all attributes i.e. `#[kani::x]` (which kani_macros translates to `#[kanitool::x]` for us to handle here)
     fn handle_kanitool_attributes(&mut self) {
-        let all_attributes = self.tcx.get_attrs_unchecked(self.current_fn().instance().def_id());
+        let def_id = self.current_fn().instance().def_id();
+        let all_attributes = self.tcx.get_attrs_unchecked(def_id);
+
+        if def_id.is_local() {
```
celinval, line 327: This code seems out of place. You are looking for things specific to `#[test]` attributes inside a function that looks for `kanitool` attributes.

Can you also move this code to its own function?

kani-driver/src/metadata.rs
```diff
@@ -108,10 +108,17 @@ fn read_kani_metadata(path: &Path) -> Result<KaniMetadata> {
...
-    let mut result = KaniMetadata { proof_harnesses: Vec::new() };
+    let mut result = KaniMetadata {
+        proof_harnesses: vec![],
+        unsupported_features: vec![],
+        test_harnesses: vec![],
+    };
     for md in files {
         // Note that we're taking ownership of the original vec, and so we can move the data into the new data structure.
         result.proof_harnesses.extend(md.proof_harnesses);
+        // TODO: these should be merged via a map to aggregate them all
```
celinval, line 119: issue

celinval (commented):
Quick feedback. I haven't reviewed `assess.rs` yet.

tedinski:
>     1. What does "success" and "assertion" means in the "Reason for failure" table?

Hm. I decided on including success in the table (so things could be tallied more easily), but then didn't really think about the "reasons for failure" title.

I could remove it (I disfavor this option), retitle the table, or change that to "no failure", or something else...

>     2. Can we enforce `--enable-unstable` for the new subcommand?

Possibly, I'm not sure it can be done in the usual way, but I can add a check to require it.

>     3. What happens if the user given other `kani` commands to the assess subcommand?

Other kani commands? Flags you mean? The subcommand itself will ignore them, but I believe the order can be swapped in clap.

e.g. `cargo kani assess --harness foo` will fail. But you might be able to cheese things with `cargo kani --harness foo assess`. To fix this we'd have to move the other flags under a subcommand, and there wasn't a "default subcommand" last I checked unfortunately...

kani-driver/src/call_cbmc.rs
```diff
@@ -254,9 +254,18 @@ impl VerificationResult {
             )
         }
     }
+
+    /// Find the failed properties from this verification run
+    pub fn failed_properties(&self) -> Vec<&Property> {
+        if let Some(properties) = &self.results {
```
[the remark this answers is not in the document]
tedinski: An empty vec is a completely valid return type.

kani-driver/src/metadata.rs
```diff
@@ -108,10 +108,17 @@ fn read_kani_metadata(path: &Path) -> Result<KaniMetadata> {
 
 /// Consumes a vector of parsed metadata, and produces a combined structure
 fn merge_kani_metadata(files: Vec<KaniMetadata>) -> KaniMetadata {
-    let mut result = KaniMetadata { proof_harnesses: Vec::new() };
+    let mut result = KaniMetadata {
```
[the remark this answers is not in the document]
tedinski: I think it's inappropriate to have `Default` for this type. This is related to the refactoring needed here. I created the issue (from your other feedback about the TODO) here: https://github.com/model-checking/kani/issues/1758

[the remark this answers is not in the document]
celinval: That's fair. This type shouldn't really have a default.

That said, I'm not sure the metadata needs refactoring. I think `kani-driver` needs to be refactored. We should probably refactor how we are merging the metadata and the harnesses. I believe the current behavior is incorrect (#1762).

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -123,6 +123,7 @@ impl<'tcx> GotocCtx<'tcx> {
             self.symbol_table.update_fn_declaration_with_definition(&name, body);
 
             self.handle_kanitool_attributes();
+            self.record_test_metadata();
```
danielsn, line 126: Is "test" the right word here? It's hard to tell from this function name what its doing: my first thought was that it was recording some metadata related to testing kani.

danielsn: Also, does this really belong as part of codegen? It might make more sense to do this either in its own pass, or when we declare the function, rather than as part of codegenning it.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -316,12 +317,47 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
+    /// Records test harness closures for KaniMetadata
```
danielsn, line 320: Can you expand this comment? What are "test harness closures"?

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -144,7 +146,11 @@ impl CodegenBackend for GotocCodegenBackend {
             None
         };
 
-        let metadata = KaniMetadata { proof_harnesses: gcx.proof_harnesses };
+        let metadata = KaniMetadata {
```
danielsn, line 149: Do we want a constructor for `KaniMetadata`?  Maybe not given that its a POD class.

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
```diff
@@ -59,6 +59,7 @@ pub struct GotocCtx<'tcx> {
     pub current_fn: Option<CurrentFnCtx<'tcx>>,
     pub type_map: FxHashMap<InternedString, Ty<'tcx>>,
     pub proof_harnesses: Vec<HarnessMetadata>,
+    pub test_harnesses: Vec<HarnessMetadata>,
```
danielsn, line 62: What is a test-harness? Is that anything with the `test` attribute?

tests/cargo-kani/assess-works/src/main.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use anyhow::*;
+
+#[test]
```
danielsn, line 6: comment this test?

kani_metadata/src/lib.rs
```diff
@@ -10,7 +10,21 @@ pub use vtable::*;
...
+    /// (These general translate to `assert(false)` so we can still attempt verification.)
+    pub unsupported_features: Vec<UnsupportedFeature>,
+    /// If crates are built in test-mode, then test harnesses will be recorded here.
+    pub test_harnesses: Vec<HarnessMetadata>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct UnsupportedFeature {
+    /// A string identifying the feature.
+    pub feature: String,
```
danielsn, line 27: Can this be an enum?

kani-driver/src/main.rs
```diff
@@ -41,6 +42,11 @@ fn cargokani_main(input_args: Vec<OsString>) -> Result<()> {
     args.validate();
     let ctx = session::KaniSession::new(args.common_opts)?;
 
+    if matches!(args.command, Some(CargoKaniSubcommand::Assess)) || ctx.args.assess {
```
danielsn, line 45: Is this the only alternative command we have?

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
```diff
@@ -99,6 +101,29 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    pub(crate) fn unsupported_metadata(&self) -> Vec<UnsupportedFeature> {
+        self.unsupported_constructs
+            .iter()
+            .map(|(construct, location)| UnsupportedFeature {
+                feature: construct.to_string(),
+                locations: location
+                    .iter()
+                    .map(|l| {
+                        (
+                            l.filename().unwrap_or_default(),
```
danielsn, line 119: This would be cleaner if `UnsupportedFeature` took an `Option` for filename and start line

kani-driver/src/assess.rs
```diff
@@ -0,0 +1,215 @@
...
+fn unsupported_features_table(metadata: &KaniMetadata) -> Table {
+    // Map "unsupported feature name" -> (crates impacted, instance of use)
+    let mut counts: HashMap<String, (usize, usize)> = HashMap::new();
+
+    for item in &metadata.unsupported_features {
+        // key is unsupported feature name
+        let mut key = item.feature.clone();
+        // There are several "feature for <instance of use>" unsupported features.
+        // We aggregate those here by reducing it to just "feature"
+        if let Some((prefix, _)) = key.split_once(" for ") {
```
danielsn, line 40: Can we just use an enum in `unsupported_features` rather than relying on parsing strings which might change?

kani-driver/src/assess.rs
```diff
@@ -0,0 +1,215 @@
...
+            key = prefix.to_string();
+        }
+        let entry = counts.entry(key).or_default();
+        // crates impacted:
+        entry.0 += 1;
```
danielsn, line 45: Make a small struct so we have named fields here. Names > comments

kani-driver/src/assess.rs
```diff
@@ -0,0 +1,215 @@
...
+        // instances of use:
+        entry.1 += item.locations.len();
+    }
+
+    // Sort descending by number of crates impacted by this missing feature
+    let mut sorted_counts: Vec<(String, (usize, usize))> = counts.into_iter().collect();
+    sorted_counts.sort_by_key(|i| usize::MAX - i.1.0);
```
danielsn, line 52: Why not just `|i| - i.1.0`

kani-driver/src/assess.rs
```diff
@@ -0,0 +1,215 @@
...
+        table.column_mut(0).unwrap().set_cell_alignment(CellAlignment::Left);
+        table
+            .column_mut(0)
+            .unwrap()
+            .set_constraint(ColumnConstraint::UpperBoundary(Width::Fixed(80)));
+        table.column_mut(1).unwrap().set_cell_alignment(CellAlignment::Right);
+        table.column_mut(2).unwrap().set_cell_alignment(CellAlignment::Right);
+
+        for item in sorted_counts {
+            table.add_row(vec![item.0, item.1.0.to_string(), item.1.1.to_string()]);
```
danielsn, line 68: named struct fields would make this easier to read and maintain

kani-driver/src/assess.rs
```diff
@@ -0,0 +1,215 @@
...
+        table.column_mut(1).unwrap().set_cell_alignment(CellAlignment::Left);
+
+        for r in results {
+            // For now we're just reporting "successful" harnesses as candidates.
+            // In the future this heuristic should be expanded. More data is required to do this, however.
+            if r.result.failed_properties().is_empty() {
+                // The functions we extract are actually the closures inside the test harness macro expansion
+                // Strip that closure suffix, so we have better names:
+                let name = r.harness.pretty_name.trim_end_matches("::{closure#0}").to_string();
+                // Location in a format "clickable" in e.g. IDE terminals
```
danielsn, line 137: nice

kani-driver/src/assess.rs
```diff
@@ -0,0 +1,215 @@
...
+    ctx.args.jobs = Some(None); // -j, num_cpu
+
+    let outputs = ctx.cargo_build()?;
+    let metadata = ctx.collect_kani_metadata(&outputs.metadata)?;
+
+    let crate_count = outputs.metadata.len();
+
+    // An interesting thing to print here would be "number of crates without any warnings"
+    // however this will have to wait until a refactoring of how we aggregate metadata
+    // from multiple crates together here.
```
danielsn, line 163: tracking ticket

kani-driver/src/assess.rs
```diff
@@ -0,0 +1,215 @@
...
+    println!("Analyzed {} crates", crate_count);
+
+    if !metadata.unsupported_features.is_empty() {
+        println!("{}", unsupported_features_table(&metadata));
+    } else {
+        println!("No crates contained Rust features unsupported by Kani");
+    }
+
+    // The section is a "copy and paste" from cargo kani.
+    // We could start thinking about abtracting this stuff out into a shared function...
```
danielsn, line 173: 👍

danielsn (commented):
Longer term proposal: should we create an HTML report which we could click through to see which crates/tests/etc failed?

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
tedinski: Perhaps not? But this is also where we spot proof harnesses, so these are in the same piece of code here.

kani_metadata/src/lib.rs
tedinski: In principle, maybe. But it's a string throughout `kani-compiler`, so that'd be a more extensive refactoring.

kani-driver/src/main.rs
tedinski: I think we will end up with more later, but it's the first.

For instance, I want us to have `cargo kani self-test` and `cargo kani versions`.

We do sort of have `cargo kani setup` but that's intercepted by the proxy binaries, not passed down to us.

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
tedinski: Only sort of? I think we actually have 0 instances of NOT having a location for our codegen_unimplemented calls. So in some sense, we're going from the too-general type (Location) to the more specific type.

I'll add a comment here.

kani-driver/src/assess.rs
tedinski: It's an unsigned type. I could cast it or something, but this was easy. I did think it was a bit weird there wasn't a better way to reverse the sort order...

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
tedinski: Yep

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
tedinski: Right, a constructor doesn't make sense here.

kani-driver/src/assess.rs
```diff
@@ -0,0 +1,226 @@
...
+    table
+}
+
+/// Internal data type for constructing the unsupported features table
+#[derive(Default)]
+struct UnsupportedFeaturesTableData {
+    crates_impacted: usize,
+    instances_of_use: usize,
+}
+fn unsupported_features_table(metadata: &KaniMetadata) -> Table {
```
adpaco-aws, line 37: I'd prefer we call this `generate_unsupported_features_table` or `build_unsupported_features_table`. Same goes for all other functions in this module. They'd also benefit from some function-level documentation, adding an example of what the table looks like and what the values in there mean.

adpaco-aws (approved):
Adding also these comments to be considered as follow-ups:
 * The feature needs to be documented in our docs.
 * Let's make a comparison between these results and the ones we get with the scripts.

PR #1757, opened 2022-10-06
Remove duplicate symtab calls from stdlib test

tedinski:
### Description of changes: 

Does the same thing we usually do with `symtab2gb` output, but for that test shell script that explicitly calls it.

### Testing:

* Is this a refactor change?

### Checklist

### DIFF
--- a/scripts/std-lib-regression.sh
+++ b/scripts/std-lib-regression.sh
@@ -73,11 +73,16 @@ export RUSTC="$KANI_DIR/target/kani/bin/kani-compiler"
 # Compile rust to iRep
 $WRAPPER cargo build --verbose -Z build-std --lib --target $TARGET
 
+# Runs a command, but suppresses output unless it fails
+function go {
+  OUTPUT=$("$@" 2>&1) || (echo "$OUTPUT" && exit 1)
+}
+
 # Generate goto-program. This will make sure the representation is well formed.
 cd target/${TARGET}/debug/deps
 for symtab in *.symtab.json; do
     echo "======== File: $symtab"
-    symtab2gb ${symtab} --out ${symtab}.out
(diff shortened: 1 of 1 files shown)


scripts/std-lib-regression.sh
```diff
@@ -73,11 +73,16 @@ export RUSTC="$KANI_DIR/target/kani/bin/kani-compiler"
...
+function go {
+  OUTPUT=$("$@" 2>&1) || (echo "$OUTPUT" && exit 1)
+}
+
 # Generate goto-program. This will make sure the representation is well formed.
 cd target/${TARGET}/debug/deps
 for symtab in *.symtab.json; do
     echo "======== File: $symtab"
-    symtab2gb ${symtab} --out ${symtab}.out
+    go symtab2gb ${symtab} --out ${symtab}.out
```
celinval, line 85: wait... can't we just remove these calls now?

celinval (dismissed):

scripts/std-lib-regression.sh
tedinski: Well you tell me :P

I thought we chatted and this test was still necessary because it tested the gotoc generation from stdlib, and the new sysroot stuff only generated MIR.

celinval: The test is still required. But we no longer need to explicit invoke symtab in this script since kani-compiler already does.

celinval: Sorry for the confusion

tedinski:
LGTM, I went and double checked that indeed kani-compiler is invoking symtab here, so this is redundant.

jaisnan:
LGTM, it seems like the concerns were addressed, so I shall approve.

jaisnan (approved):
LGTM!

PR #1761, opened 2022-10-06
Fix issue with mutable static from upstream crate

### LINKED ISSUE #1760: MIR Linker is not handling upstream static items properly.
I tried this code:

https://github.com/model-checking/kani/blob/9bfaacff235758d709918046d2b610c510c86ee7/tests/cargo-kani/asm/global/src/lib.rs#L13-L19

```
cargo kani --enable-unstable --mir-linker --harness reads_static_var_in_crate_with_global_asm
```

with Kani version: 0.12

I expected to see this happen: Harness should succeed

Instead, this happened: Harness failed.

It looks like we are not pulling static from upstream crates while linking.

celinval:
### Description of changes: 

The static allocation function in this case was initializing with a constant. However, because the static is marked as `mut` the compiler doesn't inline the value. We need to codegen the static and its initialization too. The check that the MIR was available returned `false` for some strange reason.

Since we expect that everything that is not foreign to rust to be available, we just skip that check.

### Resolved issues:

Resolves #1760 

### Related RFC:

N/A

### Call-outs:

I changed the test that was the source of this issue to just use the MIR linker.

### Testing:

* How is this change tested? Existing test

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/reachability.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/reachability.rs
@@ -436,7 +436,13 @@ pub fn collect_reachable_items<'tcx>(
 }
 
 /// Return whether we should include the item into codegen.
-/// We don't include foreign items and items that don't have MIR.
+/// - We only skip foreign items.
+///
+/// Note: Ideally, we should be able to assert that the MIR for non-foreign items are available via
+/// call to `tcx.is_mir_available (def_id)`.
+/// However, we found an issue where this function was returning `false` for a mutable static
+/// item with constant initializer from an upstream crate.
+/// See <https://github.com/model-checking/kani/issues/1760> for an example.
 fn should_codegen_locally<'tcx>(tcx: TyCtxt<'tcx>, instance: &Instance<'tcx>) -> bool {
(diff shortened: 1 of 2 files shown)


kani-compiler/src/codegen_cprover_gotoc/reachability.rs
```diff
@@ -439,17 +439,8 @@ pub fn collect_reachable_items<'tcx>(
...
-        } else {
-            // TODO: This should either be an assert or a warning.
-            // Need to compile std with --always-encode-mir first though.
-            // https://github.com/model-checking/kani/issues/1605
-            // assert!(tcx.is_mir_available(def_id), "no MIR available for {:?}", def_id);
-            //(!tcx.is_mir_available(def_id)).then(|| warn!(?def_id, "Missing MIR"));
-            tcx.is_mir_available(def_id)
-        }
+        // We cannot codegen foreign items.
+        !tcx.is_foreign_item(def_id)
```
tedinski, line 443: I think I'd prefer we left a comment about `is_mir_available` and why we're NOT using it here.

tedinski (approved):
With a fix adding a comment, LGTM

tedinski:
> for some strange reason

I think this should be part of the comment 😆

tedinski (approved):

PR #1769, opened 2022-10-10
Remove unused compiletest args from parsing & Config

### LINKED ISSUE #754: Cleanup after `compiletest` move out of `x.py`
Some places where a cleanup would be great:
 1. Either make a wrapper in `compiletest` (and other changes) to have a single call (i.e., passing the suite names as we did before) or add a script that gets called from `rmc-regression.sh`
 2. Remove references to our suites in bootstrap
 3. Remove optional arguments from `compiletest`

JanSvejda:
### Description of changes: 

Removes unused CLI arguments from parsing logic and Config class as per issue #754 , task 3.

### Resolved issues:

Resolves #754 

### Call-outs:

This is a breaking change for the compiletest CLI - is that ok? I am unfamiliar with how compiletest is used here other than what I saw in the docs and scripts.

If the args used in the docs+scripts in this repo should be the only ones available, let me know. I will remove the rest.

### Testing:

* How is this change tested? With existing tests

### Checklist
(no new procs/methods added)

### DIFF
--- a/tools/compiletest/src/common.rs
+++ b/tools/compiletest/src/common.rs
@@ -79,9 +79,6 @@ pub enum PanicStrategy {
 /// Configuration for compiletest
 #[derive(Debug, Clone)]
 pub struct Config {
-    /// The path to the directory where the Kani executable is located
-    pub kani_dir_path: PathBuf,
-
     /// The directory containing the tests to run
     pub src_base: PathBuf,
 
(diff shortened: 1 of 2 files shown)


adpaco-aws (approved):
Thanks @JanSvejda !

> This is a breaking change for the compiletest CLI - is that ok?

Yes, the options removed here aren't used or wouldn't do anything even if they were.

PR #1771, opened 2022-10-10
Fix undefined symbol errors when rustup defaults to nightly rust

### LINKED ISSUE #1709: cargo-kani fails with undefined symbol when using nightly toolchain as default
I tried this command:

```bash
$ cargo kani
error: process didn't exit successfully: `/home/ubuntu/.kani/kani-0.10.0/bin/kani-compiler -vV` (exit status: 127)
--- stderr
/home/ubuntu/.kani/kani-0.10.0/bin/kani-compiler: symbol lookup error: /home/ubuntu/.kani/kani-0.10.0/bin/../toolchain/lib/librustc_driver-e70f935d0865b673.so: undefined symbol: _ZN9hashbrown3raw11Fallibility17capacity_overflow17h52b24f50adb3934dE

```

with Kani version: 0.10.0

I expected to see a check happen.

Instead, this happened.

Looks like cargo sets `LD_LIBRARY_PATH=/home/ubuntu/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib` which causes kani to link against the wrong library with a differently mangled symbol name. 

Worked through with @tedinski.  Who says:
>  I think normally something (rustup or cargo) doesn't set LD_LIBRARY_PATH. But sometimes (apparently) it does, and I think it's changing the search order and gets a slightly wrong dynamic library (e.g. the one from nightly instead of the specific nightly we're built against)

And about this PR:
> I'll be filling in my notes

tedinski:
### Description of changes: 

I very much want this change to be self-documenting, so here's the description from the code:

> `rustup` sets dynamic linker paths when it proxies to the target Rust toolchain. It's not fully clear why. `rustup run` exists, which may aid in running Rust binaries that dynamically link to the Rust standard library with `-C prefer-dynamic`. This might be why. All toolchain binaries have `RUNPATH` set, so it's not needed by e.g. rustc. (Same for Kani)

> However, this causes problems for us when the default Rust toolchain is nightly. Then `LD_LIBRARY_PATH` is set to a nightly `lib` that may contain a different version of `librustc_driver-*.so` that might have the same name. This takes priority over the `RUNPATH` of `kani-compiler` and causes the linker to use a slightly different version of rustc than Kani was built against. This manifests in errors like: `kani-compiler: symbol lookup error: ... undefined symbol`

> Consequently, let's remove from our linking environment anything that looks like a toolchain path that rustup set. Then we can safely invoke our binaries. Note also that we update `PATH` in `exec` to include our favored Rust toolchain, so we won't re-drive `rustup` when `kani-driver` later invokes `cargo`.

### Resolved issues:

* Resolves #1709 

### Call-outs:

This is exceptionally difficult to test. I haven't yet added a test case to this PR, and should before merging.

I think it'll have to be part of the "smoke tests" we run with with docker in CI, since we need to actually be running the `cargo-kani` proxy binary, which we don't run in dev.

### Testing:

* How is this change tested? manually for now, with unit tests for the filter

### Checklist


celinval:
Please don't be vague in the description.

tedinski:
Updated description with the same text as the code, so as not to be vague on the pull request.

tedinski:
A test is now included.

Unfortunately this issue is near-impossible to test reliably: We move our rust toolchain every couple of weeks, and the issue only manifests with certain combinations of toolchain versions (where symbols are different, but library hashes are not).

Consequently, if this code/workaround is broken, we may be unable to observe that breaking immediately through this test failing, and it may not be until a few days or pull requests later when a new nightly version that does manifest the problem appears.

I can't think of any better way to test it, however.

celinval (changes requested):
Wouldn't it make sense to set RUSTUP_TOOLCHAIN env variable to ensure all calls are directed to the correct toolchain instead? https://rust-lang.github.io/rustup/overrides.html This has precedence to all other mechanisms except to selecting the toolchain as a command argument.

This should fix the LD_LIBRARY_PATH issue as well as having to change PATH.

tedinski:
> Wouldn't it make sense to set RUSTUP_TOOLCHAIN env variable to ensure all calls are directed to the correct toolchain instead?

Hmm. Part of the trouble is that the `cargo-kani` proxy doesn't locally know what the correct toolchain is. But perhaps `kani-driver` can do that.

tedinski:
Ah but I'm slightly worried about the future when we invoke `kani-compiler` directly from `kani-driver`. There wouldn't be an intervening `rustup` invocation to reset environment variables again.

celinval:
Sorry if I wasn't clear. I think you should set that variable inside the proxy before invoking `kani-driver`. I.e., in the exact same place that you are currently editing LD_LIBRARY_PATH. The proxy is already responsible for setting up the execution environment.

BTW, how did you manually test the fix?

tedinski:
I can look into adding `RUST_TOOLCHAIN`, it might be an improvement in its own right. But I still think we need to directly do something ourselves about `LD_LIBRARY_PATH`.

Setting `RUST_TOOLCHAIN` does not directly fix the problem. We'd be depending on a subsequent invocation of the `rustup` proxy to then prepend the correct paths to `LD_LIBRARY_PATH`.

So, if we just updated `RUST_TOOLCHAIN` as-is, we'd have to remove the code adding the toolchain to `PATH` (which, sure, maybe we should) because currently we're directly invoking `cargo` (not via `rustup`), so `LD_LIBRARY_PATH` would not be fixed. Removing the toolchain from `PATH` would mean we invoke a `rustup` proxy (for `cargo`) on the way to running `kani-compiler`, and so things would then work... for now.

But again, if we don't directly fix `LD_LIBRARY_PATH`, it leaves us (e.g. `kani-driver`) in a broken environmental state where we're no longer able to directly run `kani-compiler`, and that doesn't seem acceptable to me.

tedinski:
New commit:

1. Fix errant `pub`
2. Set `RUSTUP_TOOLCHAIN`
3. Remove the toolchain from `PATH` (so we're now relying on `RUSTUP_TOOLCHAIN` to select the right one instead)

celinval (approved):

celinval:
Btw, I think adding the correct path to the beginning of LD_LIBRARY_PATH seems like a much less hacky and reliable way to fix this.

tedinski:
Do you want that change made or is merging fine? In some sense, we're sort of also doing that, since we set the environment variable and call the rustup proxy now for `cargo` now.

FWIW, while I agree trying to recognize them and remove this is a bit of a hack, I personally think rustup adding them in the first place is a bigger hack, and trying to clean up and remove them, well... it's a hack trying to undo damage from another hack? Oof.

celinval:
Not a blocker that's why I approved it. Thanks Ted

Issue #1720, opened 2022-09-26
Enable MIR linker in the bookrunner


celinval:
The bookrunner executes Kani with `--function main` instead of using harnesses. For that, we will need to enable the MIR linker to start from specific functions. #1674 will enable the reachability from public functions. We should either ensure that `main` function gets included here or we might want to add a new reachability algorithm that gets the name of the function to look for.

For simplicity, I'm leaning towards re-using the public function approach.

PR #1775, opened 2022-10-12
Fix MIR Linker handling of `--function main`

celinval:
### Description of changes: 

For the specific case where we run Kani with `--function main`, we compile the crate as a binary. The logic for public functions in the MIR linker didn't account for the entry_fn(()) which is set in those cases.

Thus, we explicitly check for the entry function when collecting starting point and the reachability option is set to public functions.

### Resolved issues:

This is required for #1720

### Related RFC:

N/A

### Testing:

### Checklist


adpaco-aws (approved):
Thanks @celinval !!

Issue #1764, opened 2022-10-07
MIR Linker doesn't work with integration tests


celinval:
I tried running the test [cargo-tests-dir](https://github.com/model-checking/kani/tree/main/tests/cargo-kani/cargo-tests-dir) using the MIR linker using the following command line invocation:

```
cargo kani --mir-linker --enable-unstable
```

with Kani version: 0.12

I expected to see this happen: Verification should succeed

Instead, this happened: Compilation failed due to invoking `cargo rustc` with multiple targets.

For the `--tests` option we need to compile each test file separately.

Issue #1740, opened 2022-10-03
Intrinsics safety failures are not as friendly with `--mir-linker`


celinval:
The custom sysroot build for the MIR Linker introduced by #1717 builds the standard library in debug mode. This enables a bunch of safety checks and debug assertions from the standard library. Some of these checks are for intrinsics safety checks. The issues here are:
1. A lot of these checks make our intrinsics checks redundant.
2. The `std` checks are added before actually calling the intrinsic and they are not super user friendly.

For example, the following testcase from the `expected` suite: https://github.com/model-checking/kani/blob/78761bdac057a910a8f23f7622c0a3a9df6953fb/tests/expected/intrinsics/copy-nonoverlapping/copy-overflow/main.rs#L6-L18

Without the MIR Linker, it fails with the following error:
```
Failed Checks: copy_nonoverlapping: attempt to compute number in bytes which would overflow
```

With the MIR Linker, the error is:
```
Failed Checks: called `Option::unwrap()` on a `None` value 
```

This happens because of the std check `is_nonoverlapping` which has a call to `checked_mul().unwrap()`. The `checked_mul()` is None in the overflow scenario.

celinval:
Another interesting one is that the checks are not encoded as normal `debug_assert!()` which `panic!()`, they call `abort` instead to optimize the compilation size. So other failures will look like:
```
Failed Checks: reached intrinsic::abort
```

celinval:
The current behavior implemented for MIR Linker keeps the same behavior as before the MIR Linker. Enabling these checks are desirable but they are orthogonal to the MIR Linker work. Hence, I'm removing the link with the MIR Linker milestone.

cameron1024:
Hi, I've just run into this issue trying to verify a program that depends on `unicode-bidi`, which has this snippet:
```rust
#[cfg(feature = "std")]
std::debug_assert!(
    false,
    "Found broken indices in level run: found indices {}..{} for string of length {}",
    level_run.start,
    level_run.end,
    text.len()
);
```
This fails with an error:
```
error: cannot find macro `__kani__workaround_core_assert` in this scope
```
I'm very new to kani, so I'm not exactly sure if this is actually the same issue, or just user error. I just stumbled upon this issue when googling.

Are there steps a user can take to work around this error? (Apologies if this is the wrong place to ask these sorts of questions :sweat_smile: )

zhassan-aws:
[...]

PR #1785, opened 2022-10-14
Enable MIR Linker by default and adjust tests

### LINKED ISSUE #1677: Switch the regression to run with the MIR Linker
Once we are close to stabilizing the MIR Linker, we should enable the regression to always run with the MIR Linker.

We should also remove the conditional compilation option that allow us to enable the MIR Linker in the regression.

### LINKED ISSUE #1738: Make `--mir-linker` the default behavior
Given the good performance impact from enabling `--mir-linker`, we should probably make it as default. That said, we haven't gone through the full stabilization path. There are a few features missing, and bugs to be uncovered. So the proposal right now is to create a `--legacy-linker` flag that will enable users to switch back to the old behavior in case they encounter any issue.

The `--function` argument will also only work with the `--legacy-linker` for now until we are able to implement #1674.

celinval:
### Description of changes: 

Enable `--mir-linker` by default. This change doesn't remove the option yet though to avoid breaking users harnesses. Introduce a `--legacy-linker` to allow users to switch to the older linker to overcome performance issues as well as any possible issue with the MIR Linker.

### Resolved issues:

Resolves #1677 
Resolves #1738 

### Related RFC:

#1588 

### Call-outs:

I had to adjust a bunch of tests for the following reasons:
  - intrinsics: Use legacy linker for some of them because of not so friendly messages. #1740 
  - performance: A few test are taking too long because of the new symbols being included in the analysis. Keep legacy linker for now.
  - integ-test: Using --tests with integration tests fail with MIR Linker. #1764
  - fixme: Enable a few fixme tests and tests that were marked as verify-fail due to missing symbols. These run fine now with the linker.

### Testing:

* How is this change tested? The regression

### Checklist


tedinski:
> Using --tests with integration tests fail with MIR Linker

I'm worried this might be a blocker. How much of a bother is this to fix? Do we need a fully solution to:

* #1448 

i.e. Understanding the build graph better in `kani-driver`?

celinval:
Yeah, that's in the top of my list. I'll see if I can get a fix in today. Now that I think about it it would break whenever integration tests exists even without any proof harness.

celinval:
I created https://github.com/model-checking/kani/pull/1789 to fix the tests issue. Once that gets merged, I'll update the test in this PR to also use the MIR Linker

tests/expected/intrinsics/copy-nonoverlapping/copy-overflow/main.rs
```diff
@@ -1,8 +1,14 @@
...
-
-// Checks that `copy_nonoverlapping` triggers an overflow failure if the `count`
-// argument can overflow a `usize`
+// kani-flags: --legacy-linker
+//! Checks that `copy_nonoverlapping` triggers an overflow failure if the `count`
+//! argument can overflow a `usize`
+//!
+//! When enabling "--mir-linker", the error we currently get is:
+//! > Failed checks: Called `Option::unwrap()` on a `None` value
+//! This happens because of std debug checks. The `is_nonoverlapping` check has a
```
tedinski, line 9: Honestly I wonder if we should just disable `debug_assert` in our build of `std` for now?

tests/kani/Refs/main.rs
```diff
@@ -1,66 +1,69 @@
...
+//!                   * width: 64
+//!                   * #typedef: __CPROVER_size_t
+//!                   * #c_type: unsigned_long_int
+//!               * #source_location:
+//!                 * file: <built-in-additions>
+//!                 * line: 16
+//!                 * working_directory: /Users/dsn/ws/RustToCBMC/src/RustToCBMC/rust-tests/cbmc-reg/Refs
+//!               * value: 0
+//!
+//! Note: This test takes too long with all the std symbols. Use --legacy-linker for now
```
tedinski, line 66: Please move the comment closer to the flag.

tests/kani/SizeAndAlignOfDst/main_assert_fixme.rs
```diff
@@ -1,35 +1,11 @@
...
-//        },
-//        "typ":"StructTag(""tag-dyn Subscriber"")",
-//     }
-//  ]"', compiler/rustc_codegen_llvm/src/gotoc/cbmc/goto_program/expr.rs:532:9"
+// kani-flags: --legacy-linker
+//! This is a regression test for size_and_align_of_dst computing the
+//! size and alignment of a dynamically-sized type like
+//! Arc<Mutex<dyn Subscriber>>.
+//! https://github.com/model-checking/kani/issues/426
+//! Note: This test takes too long with all the std symbols. Use --legacy-linker for now.
```
tedinski, line 8: Also move comment near flag.

tests/kani/Slice/pathbuf.rs
```diff
@@ -1,11 +1,11 @@
...
-// kani-verify-fail
 
 use std::fs;
 use std::path::PathBuf;
 #[kani::proof]
+#[kani::unwind(3)]
 fn main() {
     let buf = PathBuf::new();
-    let _x = fs::remove_dir_all(buf);
+    let _x = fs::remove_file(buf);
```
tedinski, line 10: Why the change? Also... this verifies? I figured it'd hit syscalls or something at some point.

tests/kani/Strings/parse.rs
```diff
@@ -0,0 +1,14 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This test is to check how Kani handle some std functions, such as parse.
+//! This test used to trigger a missing function before the MIR Linker.
```
tedinski, line 5: 🥳

tests/ui/missing-function/rust-by-example-description/main.rs
```diff
@@ -1,13 +0,0 @@
-// Copyright Kani Contributors
-// SPDX-License-Identifier: Apache-2.0 OR MIT
-
-// kani-flags: --enable-unstable --cbmc-args --unwind 4 --object-bits 9
-// This test is to check if the description for undefined functions has been updated to "Function with missing definition is unreachable"
```
tedinski, line 5: While this one doesn't work, shouldn't we keep a test for the property description? IIRC we had difficulty detecting it in the past. Or did this get replaced with a different test and I didn't notice

scripts/codegen-firecracker.sh
```diff
@@ -26,8 +26,7 @@ cd $KANI_DIR/firecracker/src/devices/src/virtio/
 # Disable warnings until https://github.com/model-checking/kani/issues/573 is fixed
 export RUSTC_LOG=error
 export RUST_BACKTRACE=1
-# Kani cannot locate Cargo.toml correctly: https://github.com/model-checking/kani/issues/717
-cargo kani --only-codegen
+cargo kani --only-codegen --legacy-linker
```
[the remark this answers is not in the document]
celinval: sure

kani-driver/src/args.rs
```diff
@@ -136,11 +136,20 @@ pub struct KaniArgs {
...
+    /// `std` library. See <https://github.com/model-checking/kani/issues/1213> for more details.
+    #[structopt(long)]
+    pub legacy_linker: bool,
+
+    /// Enable the new MIR Linker. This is already the default option and it will be removed once
+    /// the linker is stable.
+    /// The MIR Linker affects how Kani prunes the code to be analyzed. It also fixes previous
+    /// issues with missing `std` function definitions.
+    /// See <https://model-checking.github.io/kani/rfc/rfcs/0001-mir-linker.html> for more details.
+    #[structopt(long, conflicts_with("legacy_linker"))]
```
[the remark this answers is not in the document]
celinval: Sure

tests/expected/intrinsics/copy-nonoverlapping/copy-overflow/main.rs
celinval: That's an interesting idea. It would help with this for now. It would be interesting to see how much more UB checks the std encode as `debug_assert()`.

tests/kani/Refs/main.rs
celinval: will do. I am actually tempted to remove this random log from the comments. I can make sure the original ticket has it and just point to the ticket for more details. I didn't bother doing this here because I was worried about the size of this PR.

tests/kani/SizeAndAlignOfDst/main_assert_fixme.rs
celinval: ok

tests/kani/Slice/pathbuf.rs
celinval: The remove file is a no-op if the `buf` is empty.

celinval: The reason I changed is that I wanted to make sure this wouldn't remove some random folders if someone tried to run this test with regular `rustc`. :)

tests/ui/missing-function/rust-by-example-description/main.rs
celinval: I did keep two test for that. There is one test for the case of an `extern` function and one for the `--legacy-linker`

tests/kani/Refs/main.rs
adpaco-aws: Please do, I don't think the log is that helpful.

tests/ui/missing-function/rust-by-example-description/main.rs
celinval: - https://github.com/model-checking/kani/tree/main/tests/ui/missing-function/replaced-description
- https://github.com/model-checking/kani/tree/main/tests/ui/missing-function/extern_c

tedinski (approved):
LGTM. I would vote for temporarily disabling std debug_assert since the UX is bad, but I don't think that needs to be folded into this PR, and maybe is something we can slip to the next release.

PR #1791, opened 2022-10-17
Disable debug_asserts inside `std` due to poor UX

celinval:
### Description of changes: 

This is a possible workaround for now to improve the UX around UB detection in intrinsics when using the MIR Linker. The issue is tracked by https://github.com/model-checking/kani/issues/1740.

Basically, the rust toolchain uses a release build that removes all debug assertions from the standard library. When we switched to using our custom build of the `std` we decided to enable the debug assertions in order to catch more potential UBs. However, the UX is not great. Many assertions and validation logic don't have clear descriptions and they may fail in unexpected places. E.g.: The violation of an intrinsic safety condition triggers the following failures:
```
> Failed checks: Called `Option::unwrap()` on a `None` value
```

### Resolved issues:

Mitigates #1740

### Related RFC:

#1588 

### Call-outs:

I'm creating this PR to get some feedback on a possible mitigation proposed by @tedinski. I think in the long run, we should see if it's possible to make these errors more user friendly inside the standard library.

The downside of this PR is that other UBs that could be detected via `debug_assert` inside the `std` might be missed or can cause a side effect resulting in another type of failure.

### Testing:

* How is this change tested? Modified tests.

### Checklist

### DIFF
--- a/tests/expected/intrinsics/copy-nonoverlapping/copy-overflow/main.rs
+++ b/tests/expected/intrinsics/copy-nonoverlapping/copy-overflow/main.rs
@@ -1,14 +1,7 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-// kani-flags: --legacy-linker
 //! Checks that `copy_nonoverlapping` triggers an overflow failure if the `count`
 //! argument can overflow a `usize`
-//!
-//! When enabling "--mir-linker", the error we currently get is:
-//! > Failed checks: Called `Option::unwrap()` on a `None` value
-//! This happens because of std debug checks. The `is_nonoverlapping` check has a
-//! `checked_mul().unwrap()` that fails in the overflow scenario.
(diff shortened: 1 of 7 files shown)


tedinski (approved):
Obviously I'm in favor, but we should get consensus on this before merging.

adpaco-aws:
I've got one question: Why were these debug assertions not appearing earlier?

I thought we were [overriding them](https://model-checking.github.io/kani/overrides.html) to work as regular assertions, but then we disabled debug assertions altogether?

celinval:
> I've got one question: Why were these debug assertions not appearing earlier?
> 
> I thought we were [overriding them](https://model-checking.github.io/kani/overrides.html) to work as regular assertions, but then we disabled debug assertions altogether?

Our overrides don't apply to the standard library only to user code. We used to use the toolchain standard library which is pre-compiled in release mode.

adpaco-aws (approved):

PR #1800, opened 2022-10-25
Bump deps

tedinski:
### Description of changes: 

Bump crate deps versions. Still skipping clap. Fixes cargo audit failure: https://github.com/model-checking/kani/actions/runs/3321514596/jobs/5489265216#step:4:52

### Testing:

* Is this a refactor change? effectively

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,9 +25,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.19"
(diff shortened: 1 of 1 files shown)


tedinski:
Long story short, no.

Long story: cargo leans into semver hard. So the `toml` version are basically a minimum without going to the next "major version" and the `lock` version is what's actually used to build. 

If we were distributing a library, we might care what the stated minimum versions were (e.g. "we need the new feature from X.2.0 that wasn't in X.1.0") but we're not, so.

zhassan-aws (approved):

Issue #1810, opened 2022-10-27
Not using `--slice-formula` causes invariant violation in CBMC


adpaco-aws:
When upgrading to CBMC v5.68.0, CBMC crashes in scenarios where we don't push `--slice-formula` with the following error:
```
--- begin invariant violation report ---
Invariant check failed
File: ../src/solvers/flattening/boolbv_width.cpp:225 function: get_entry
Condition: false
Reason: Unimplemented
Backtrace:
cbmc(+0xa8c6d0) [0x55b32942f6d0]
cbmc(+0xa8cc79) [0x55b32942fc79]
cbmc(+0x1e7ea8) [0x55b328b8aea8]
cbmc(+0x828ebd) [0x55b3291cbebd]
cbmc(+0x7fbcf9) [0x55b32919ecf9]
cbmc(+0x7e441e) [0x55b32918741e]
cbmc(+0x82f026) [0x55b3291d2026]
cbmc(+0x7e7a70) [0x55b32918aa70]
cbmc(+0x7e39ca) [0x55b3291869ca]
cbmc(+0x7e3a98) [0x55b329186a98]
cbmc(+0x537d81) [0x55b328edad81]
cbmc(+0x540546) [0x55b328ee3546]
cbmc(+0x5410f2) [0x55b328ee40f2]
cbmc(+0x3379a4) [0x55b328cda9a4]
cbmc(+0x337e25) [0x55b328cdae25]
cbmc(+0x3498b5) [0x55b328cec8b5]
cbmc(+0x1f21c9) [0x55b328b951c9]
cbmc(+0x1ee6fe) [0x55b328b916fe]
cbmc(+0x1e5f3f) [0x55b328b88f3f]
cbmc(+0x1d1f49) [0x55b328b74f49]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3) [0x7f45380190b3]
cbmc(+0x1e78be) [0x55b328b8a8be]
```
https://github.com/diffblue/cbmc/pull/7276 tries to improve the error message we get in this case.

celinval:
The following test is enough to trigger the failure without `--slice-formula`:
```rust
// Copyright Kani Contributors
// SPDX-License-Identifier: Apache-2.0 OR MIT

#[kani::proof]
fn check() {
    kani::assert(true, "");
}

```

celinval:
If I run the same test as above without `--slice-formula` and I disable the reachability checks, I get a different error:
```
[...]

PR #1811, opened 2022-10-27
Upgrade to CBMC 5.68.0 (with fixes)

### LINKED ISSUE #1779: Updating to latest CBMC version requires update to architecture strings
With https://github.com/diffblue/cbmc/pull/7215 Kani will need to update various `int_constant` (and `int_constant_from_bool`) entries in `cprover_bindings/src/env.rs` to produce constants of type "integer" instead of the 32-bit bitvector type int.

adpaco-aws:
### Description of changes: 

Upgrades CBMC to v5.68.0 and changes Kani so that it conforms with the latest changes in CBMC:
 * Changes the type of Kani-generated CProver primitives to the `integer` type where appropriate (also adds required support to generate `integer` ireps). This change was forced by CBMC (see https://github.com/model-checking/kani/issues/1779)
 * Redefines C functions with `void` return types so they use type `struct Unit` instead. This is necessary because of stricter type-checking in CBMC (likely caused by https://github.com/diffblue/cbmc/pull/6547)
 * Always pushes `--slice-formula`, even in the cases where we didn't (opened #1810 for debugging).
 * Moves the diamond dependency test into the `cargo-kani` suite. This was the last test failing, but it run its own compilation process so it was dubious. Moving into `cargo-kani` showed no issues.

### Resolved issues:

Resolves #1779 

### Testing:

* How is this change tested? Existing egression.

### Checklist


cprover_bindings/src/env.rs
```diff
@@ -58,7 +65,7 @@ pub fn machine_model_symbols(mm: &MachineModel) -> Vec<Symbol> {
         ),
         int_constant("__CPROVER_architecture_wchar_t_width", mm.wchar_t_width),
         int_constant("__CPROVER_architecture_word_size", mm.word_size),
-        int_constant("__CPROVER_rounding_mode", mm.rounding_mode),
+        int_constant_c_int("__CPROVER_rounding_mode", mm.rounding_mode),
```
danielsn, line 68: Seriously? There is one remaining outlier? Is there an issue open to CBMC about that?

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -343,6 +347,7 @@ impl Type {
             IncompleteStruct { .. } => unreachable!("IncompleteStruct doesn't have a sizeof"),
             IncompleteUnion { .. } => unreachable!("IncompleteUnion doesn't have a sizeof"),
             InfiniteArray { .. } => unreachable!("InfiniteArray doesn't have a sizeof"),
+            Integer => st.machine_model().int_width,
```
tautschnig, line 350: This doesn’t have a sizeof, should be handled like the line before.

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -286,6 +289,7 @@ impl Type {
             CInteger(CIntType::Bool) => Some(mm.bool_width),
             CInteger(CIntType::Char) => Some(mm.char_width),
             CInteger(CIntType::Int) => Some(mm.int_width),
+            Integer => Some(mm.int_width),
```
tautschnig, line 292: This one is unbounded, it doesn’t have a bitwidth.

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -46,6 +46,8 @@ pub enum Type {
     IncompleteStruct { tag: InternedString },
     /// `union x {}`
     IncompleteUnion { tag: InternedString },
+    /// `integer`
```
danielsn, line 49: What is the difference between this and CInteger?

cprover_bindings/src/goto_program/typ.rs
danielsn, line 350: Are we sure that's true? Is there documentation for this?

kani-driver/src/call_cbmc.rs
```diff
@@ -124,12 +124,17 @@ impl KaniSession {
...
+        // Previously, this would happen if the condition below was satisfied:
+        // ```rust
+        // if !self.args.visualize
+        //   && self.args.concrete_playback.is_none()
+        //   && !self.args.no_slice_formula
+        // ```
+        // But for some reason, not pushing it causes a CBMC invariant violation
+        // since version 5.68.0.
+        // <https://github.com/model-checking/kani/issues/1810>
+        args.push("--slice-formula".into());
```
danielsn, line 137: Does this mess with concrete playback?

library/kani/kani_lib.c
```diff
@@ -8,6 +8,12 @@
 #include <stdlib.h>
 #include <string.h>
 
+// Note: the Kani compiler changes `void` return types to the empty tuple `()`
+// AKA `VoidUnit`. Because of that, we need to declare its type `struct Unit`
+// and the extern instance `VoidUnit` here.
```
danielsn, line 13: I'm not sure this is an accurate description.  Rust `()` is not the same thing as void, and `__rust_dealloc` is extern Rust not extern C https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/alloc/alloc.rs.html#37

kani-driver/src/call_cbmc.rs
adpaco-aws: Needs more investigation.

My understanding was that using `--slice-formula` with concrete playback may remove calls to `kani::any()` from the trace, which doesn't allow us to provide a counter-example (in that case, we just report a test couldn't be generated). But all tests we have for concrete playback and `--visualize` pass if we use `--slice-formula`.

library/kani/kani_lib.c
adpaco-aws: OK, I've changed it to the same comment we use in the Kani compiler's `typ.rs` module:
```
/// Mapping unit to `void` works for functions with no return type but not for
/// variables with type unit. We treat both uniformly by declaring an empty
/// struct type: `struct Unit {}` and a global variable `struct Unit VoidUnit`
/// returned by all void functions (both declared by the Kani compiler).
```

cprover_bindings/src/goto_program/typ.rs
adpaco-aws: This one's a machine independent integer. Added that in the comment.

cprover_bindings/src/env.rs
adpaco-aws: Couldn't find one, so I opened https://github.com/diffblue/cbmc/issues/7282 (and pasted the issue there)

cprover_bindings/src/goto_program/typ.rs
adpaco-aws: Done.

cprover_bindings/src/goto_program/typ.rs
adpaco-aws: Done. Thanks!

cprover_bindings/src/goto_program/typ.rs
adpaco-aws: Looks like it's not (see @tautschnig comments).

adpaco-aws: Done.

cprover_bindings/src/goto_program/typ.rs
adpaco-aws: Actually, do you think we should return `None` or just `Some(x)` where x is a large number?

tautschnig:
High-level note: you might consider upgrading to 5.69.1 right away, which will be out later on today (https://github.com/diffblue/cbmc/pull/7283).

cprover_bindings/src/env.rs
tautschnig: I'll comment on the CBMC issue, but indeed rounding mode is different.

adpaco-aws:
> High-level note: you might consider upgrading to 5.69.1 right away, which will be out later on today ([diffblue/cbmc#7283](https://github.com/diffblue/cbmc/pull/7283)).

Tried now, but `brew` couldn't find it in CI. I'm rolling back to v5.68.0 for now.

tautschnig:
> > High-level note: you might consider upgrading to 5.69.1 right away, which will be out later on today ([diffblue/cbmc#7283](https://github.com/diffblue/cbmc/pull/7283)).
> 
> Tried now, but `brew` couldn't find it in CI. I'm rolling back to v5.68.0 for now.

Please retry, should just be a matter of timing.

cprover_bindings/src/goto_program/typ.rs
danielsn: Maybe we can have an "infinite" token? But `Some(x)` is semantically wrong for all `x`

tests/kani/ForeignItems/lib.c
```diff
@@ -48,7 +55,11 @@ struct Foo2 {
 
 uint32_t S = 12;
 
-void update_static() { S++; }
+struct Unit update_static()
```
danielsn, line 58: This is something that ought to work once C-FFI is in place.  We should have comments and a tracking ticket explaining that this is a workaround we want to remove

danielsn (approved):

PR #1812, opened 2022-10-28
Fix conflicting C function definitions

### LINKED ISSUE #1774: Spurious verification error on s2n-quic harness on NixOS
If I run that on my x86 macbook it's now super fast! That seems to have addressed the issue. However on my Linux machine I'm seeing a verification error, even after nuking `/target`:

```
...
 518   │ Check 95: std::option::Option::<usize>::unwrap.unreachable.1
 519   │      - Status: SUCCESS
 520   │      - Description: "unreachable code"
 521   │      - Location: ../../../../../../runner/.rustup/toolchains/nightly-2022-09-13-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:773:15 in function std::option::Option::<usize>::unwrap
 522   │
 523   │ Check 96: memcpy$link1.assertion.1
 524   │      - Status: FAILURE
 525   │      - Description: "undefined function should be unreachable"
 526   │      - Location: <builtin-library-memcpy> in function memcpy$link1
 527   │
 528   │ Check 97: std::vec::Vec::<std::option::Option<()>>::as_mut_ptr.assume.1
 529   │      - Status: SUCCESS
 530   │      - Description: "assumption failed"
 531   │      - Location: ../../../../../../runner/.rustup/toolchains/nightly-2022-09-13-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:1210:13 in function std::vec::Vec::<std::option::Option<()>>::as_mut_ptr
 532   │
...

SUMMARY:
 ** 1 of 1104 failed (24 unreachable)
Failed Checks: undefined function should be unreachable

VERIFICATION:- FAILED

Verification Time: 0.4641839s
```

This may be worth opening another issue for, though.

_Originally posted by @camshaft in https://github.com/model-checking/kani/issues/1714#issuecomment-1271930689_

celinval:
### Description of changes: 

Declare the function signatures instead of importing system definitions. This avoids conflict between our pre-build libraries and the C files that are built on the user machine.

Also remove the gen_c_lib.c file which seems to be dead code.

### Resolved issues:

Fixes #1774

### Call-outs:

1. Thanks for the suggestion @tedinski.
2. We should probably increase the amount of tests we run with our bundles. I think there is a gap here. We don't test standalone Kani, neither more complicated examples. I'll leave that for a follow up PR.

### Testing:

* How is this change tested? Existing tests + manually tested on NixOs

### Checklist


library/kani/kani_lib.c
```diff
@@ -1,12 +1,14 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-
-#include <assert.h>
 #include <stdbool.h>
```
tautschnig, line 3: While at it: do you really need/want to keep this one? I'd suggest you use `__CPROVER_bool` instead of `bool`.

tedinski (approved):

library/kani/kani_lib.c
celinval: yeah, I can do that.

celinval: Should I also use __CPROVER_size_t?

tautschnig: Good idea, though you’ll still need the header for uint8_t.

celinval: Yeah, I noticed that. I ended up just replacing the bool but I can replace size_t too.

danielsn (approved):
LGTM

celinval:
Declare the function signatures instead of importing system definitions. This avoids conflict between our pre-build libraries and the C files that are built on the user machine.

Also remove the gen_c_lib.c file which seems to be dead code.

Issue #1687, opened 2022-09-16
RFC: Resource limits for proof harnesses


tedinski:
This issue is a task to write/propose an RFC for resource limits.

Resource limits include:

1. Timeouts on proof harnesses. Including a default, and annotations to change them, or flags to manipulate them.
2. Possibly memory limits. Also including a default plus annotations/etc.

The need for these is two-fold:

1. Without resource limits, we get poor customer experiences. For instance, unwinding problems manifest as non-termination without a timeout. CI time can be unbounded when certain failures occur.
2. Resource limits are necessary to get a good experience from parallel harness runners. (If nothing else, we need to know which harnesses need a lot of memory, and limit concurrency so they have enough available. But we can also ensure best performance by starting with longest-timeout harnesses, to keep concurrency high.)

There are questions to resolve:

1. What is the default timeout? (I will initially propose an aggressive one. Maybe 15s)
2. Should memory limits be hard or soft or a combination? (e.g. use OS features to hard-limit to 2GB? Or just track usage and warn if it's over? Or both: hard limit X GB, warn when close to X?)
3. Should memory limits be numeric or "order of magnitude" (low, medium, high, extreme. Perhaps corresponding to 1 GB, 2 GB, 8GB, whole machine.) Or perhaps both.
4. How would memory limits interact with "library back-ends"? Perhaps we must always do solving in a separate process?
5. Should we have command line flags to manipulate them? Which flags? (e.g. we have strange flags around unwind bounds right now, because we start off thinking about only one proof harness at a time.)

and probably more.

Early comments about things I should think about welcome, but details can wait for the RFC.

tedinski:
* Related #885

tedinski:
We should ensure we're able to detect (at least) "timeout due to unwinding" versus "timeout due to problem size" and clarify then that an "adding `#[kani::unwind(x)]`" is the next step.

karkhaz:
One thing that occurs to me is that I wouldn't really want to specify particular memory sizes. What I'm really concerned about is that the usage doesn't exhaust the available RAM on my machine; I'd like to specify that kani terminate itself when available RAM is less than 5%, say. It feels cumbersome to me to have to specify that I want the mem limit is 24 GiB if I have to manually check that I actually have that much RAM available, it would be nice if the tool could check for me.

On the other hand, if I purposely want to constrain RAM to a very specific value for testing, it's probably good to be able to specify an absolute value too.

[...]

Issue #1808, opened 2022-10-27
Specify finer-grained unwinding values


zpzigi754:
Requested feature: a finer-grained unwinding values
Use case: a big harness containing multiple loops or iterators

Currently, the unwinding attribute(`#[kani::unwind(x)]`) could be applied to the granular of a harness. This scheme is good when the harness is small. However, if a target harness is big enough to contain multiple loops or iterators like integration tests, specifying one maximum unwinding value could affect the overall verification performance. For example, the existence of one loop or iterator requiring a big unwinding value would enforce using that value for the rest of loops or iterators which require a small unwinding value. 

The suggestion is the ability to assign different unwinding values to each loop or iterator: **statement-level unwinding value assignment**. Or different unwinding values could be assigned to each function or method which could be reachable by a harness: **function-level unwinding value assignment**. The second option could be achieved with the first option but might enhance the usability. In either cases, if an unwinding value has not been specified, the default unwinding value could be used. I hope that the suggestion could enhance the overall performance for big harnesses.

tedinski:
I think we're somewhat disinclined to add this specific feature. I'd like to know more about why you want it.

I understand that this approach is common for CBMC or other bounded model checkers, but for the way we see Kani getting used, it's likely to be a nightmare. It was never a truly good solution in the first place, and Rust code generally makes it a LOT harder to "address" specific loops in a comprehensible and reliable way. One of our goals is to keep proof harnesses independent from each other, so we don't want to put the annotation within the code being verified.

I think our tentative vision for resolving this general problem is to try to better infer loops bounds wherever we can. This is partially improving how CBMC detects them, but perhaps also finding ways for kani-compiler to pass "loop bound hints" from the higher level Rust code to the lower level IR. The hope would be that in many cases (certainly more than today) you wouldn't need an explicit global bound at all, and otherwise hopefully a single global bound would be "ok", as most loops still get tighter bounds inferred.

Did you run into a blocker in using Kani this would resolve? Maybe we could look at the specific case you have and see if there's something we can do...

PR #1815, opened 2022-10-28
RFC: Resource limits (timeouts) for proof harnesses

tedinski:
### Description of changes: 

[Rendered file can be read here](https://github.com/tedinski/rmc/blob/limits-rfc/rfc/src/rfcs/0003-resource-limits-timeouts.md)

Current status: Initial draft. I'm most interested in getting questions that the doc does not yet answer.

### Resolved issues:

Related to #1687 

### Checklist

### DIFF
--- a/rfc/src/rfcs/0003-resource-limits-timeouts.md
+++ b/rfc/src/rfcs/0003-resource-limits-timeouts.md
@@ -1,0 +1,224 @@
+- **Feature Name:** Resource limits (e.g. timeouts) for proof harnesses
+- **Feature Request Issue:** https://github.com/model-checking/kani/issues/1687
+- **RFC PR:** TBD
+- **Status:** Under Review
+- **Version:** 0
+- **Proof-of-concept:** N/A
+
+## Summary
+
+The RFC describes the introduction of resource limits for proof harnesses in Kani. This includes:
+
+1. Timeouts
+2. Memory limits
+
+## User Impact
+
+Timeouts are an often requested feature, with many reasons given for the request:
+
+1. Non-termination in any form is generally a bad user experience.
+2. Unwinding failures manifest as non-termination, and timeouts during the unwinding process would allow us to report the problem as an unwinding failure and give suggestions.
+3. Putting Kani in CI is often difficult if a failure may simply result in a CI process continuing out of control.
+4. Lack of timeouts encourages "abuse" of low unwinding values, to try to "force" termination early.
+5. (For our own purposes) We'd like to be able to create a "suite" of problems that are "too hard" for Kani to presently solve, and observe improvements over time.
+6. (Future) Parallel runners will be able to prioritize longer-running harnesses, so parallelism can be kept high throughout the verification process.
+
+Memory limits have not yet been requested, but are an essential feature to accomodate a parallel harness runner.
+Without memory limits, it would be easy for too many solvers in parallel to run a machine out of memory.
+Past tools for CBMC have required special annotations for "hard" harnesses to run serially.
+
+Further, to realize these benefits, we **must** implement resource limits by default.
+Consequently, this will likely be a "breaking change" for many customers.
+Either we should get feedback on whether this would be welcome, or we should consider a staged strategy for rolling out the feature (considered later in this doc).
+
+## User Experience
+
+The aim of this RFC is to introduce two new annotations to be placed along side proof harnesses.
+For example, from our unwinding tutorial:
+
+```rust
+#[cfg(kani)]
+#[kani::proof]
+#[kani::unwind(1)] // deliberately too low
+fn check_initialize_prefix()
+```
+
+This shows our abuse of `unwind` to force termination.
+We would prefer this to be unnecessary, and for the user to be introduced to unwinding via a default timeout.
+
+In the case where we do want to give higher limits, users should have to write something like:
+
+```rust
+#[cfg(kani)]
+#[kani::proof]
+#[kani::timeout(60)]
+#[kani::memory(10)]
+fn check_initialize_prefix()
+```
+
+> **Question 1: Should we consider alternative approach to specifying these limits?**
+>
+> I argue yes: We're now proliferating the number of attributes, so perhaps it's time to consider supporting:
+> `#[kani::proof(unwind = 5, timeout = 60, memory = 10)]`
+>
+> As we'll see later, the disadvantage of default limit choices is the friction of adding limit annotations to harnesses.
+> Making this as easy as possible helps mitigate those disadvantages.
+
+> **Question 2: Should we include units in the attributes?**
+>
+> I'm presently giving "seconds" and "gigabytes" and these seem totally reasonable units that we probably wouldn't ever deviate from.
+> However, it might be nice to either require units (to be explicit) or include units (if we think people might need them... perhaps just minutes).
+> I believe it would be reasonable to omit units however, and I'm not sure how difficult it would be to support `timeout = 2min` or `timeout = 4:00`.
+
+Beyond specifying limits on each proof harness, users would interact with these limits in two other ways:
+
+1. The output from running Kani.
+2. The command line arguments for manipulating limits.
+
+> **Question 3: How much flexibility should be provide on the command-line?**
+>
+> I believe we can keep this small.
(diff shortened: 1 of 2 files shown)


rfc/src/rfcs/0003-resource-limits-timeouts.md
```diff
@@ -0,0 +1,224 @@
...
+2. Keep this as the default for 2 kani releases (4 weeks).
+Perhaps longer if issues arise.
+3. Flip the default to enforcing.
+This will finally cause the breaking change, but will have had some warning, and we retain an option to revert to the old behavior.
+
+## Open questions
+
+TBD
+
+## Future possibilities
```
aaronbembenek-aws, line 222: A few ideas that came to mind as future possibilities:

- Could we add a way to set a machine-independent resource limit? [This thread](https://github.com/Z3Prover/z3/issues/56) has a pretty good discussion of why this can be useful for reproducible verification results. Machine-independent resource limits are part of the SMT-LIB standard and [have been proposed for CBMC](https://github.com/diffblue/cbmc/issues/4452). This would help ensure that harnesses don't time out in CI when they work fine locally.
- Could Kani provide some sort of functionality where it runs a harness (perhaps a couple of trials), and then adds resource bound annotations to the harness based off of the actual amount of resources it takes to complete? This could help catch regressions in prover performance. (Perhaps it'd be more useful for Kani internally than for clients.)

tedinski: GREAT ideas. This also brings to mind the mechanism I proposed to implement the limits currently is pretty primitive and will need adjusting if we start fully supporting external solvers (vs just minisat inside the cbmc process).

zhassan-aws:
Thanks for putting this together @tedinski.

I'm not in favor of having a default timeout; as a user, I would expect that if I don't specify any constraints, the tool should continue to run until completion. It also makes the usage a bit cumbersome. For CI, I would expect that users would have to pick an appropriate timeout anyway (even if we have a default timeout) since the necessary timeout could be quite different for different repos/tests.

I believe the unlimited unwind issue should be solved separately, e.g. through using a small unwind value by default (e.g. 1). We can make the current unwinding assertion message a bit more helpful by pointing to the documentation and/or suggesting adding a `#[kani::unwind]` annotation. Not specifying an unwind value rarely works, even for trivial program like this one:
```rust
#[kani::proof]
fn main() {
    let a = [1, 2, 3];
    let mut sum = 0;
    for i in a.iter() {
        sum += i;
    }
    assert_eq!(sum, 6);
}
```
The only cases it works AFAIK is when the loop is range-based (which I think is not very common in Rust since iterator-based loops are more common), or for loop-free programs (for which, a default unwind of 1 would still work).

On the other hand, having a default memory limit makes sense, because not having one can easily bring the machine down. A suggestion here is to make it a percentage of the machine's memory (e.g. 50%).

rfc/src/rfcs/0003-resource-limits-timeouts.md
```diff
@@ -0,0 +1,224 @@
...
+We might wish to write specific documentation to link to from (e.g.) the error message we emit when a limit was exceeded.
+
+## Detailed Design
+
+This will impact:
+
+1. Two new attributes in the `kani` library: `kani::timeout(X)` and `kani::memory(X)`.
+Also `kani::proof` if we decide to add convenient annotations there.
+2. `kani-compiler` and `kani_metadata` will be updated to record this metadata.
+3. `kani-driver` needs to updated to apply these limits.
```
celinval, line 118: How is `kani-driver` going to enforce those limits? How will this evolve if we start using CBMC as a library instead of a separate process?

rfc/src/rfcs/0003-resource-limits-timeouts.md
```diff
@@ -0,0 +1,224 @@
...
+
+Beyond specifying limits on each proof harness, users would interact with these limits in two other ways:
+
+1. The output from running Kani.
+2. The command line arguments for manipulating limits.
+
+> **Question 3: How much flexibility should be provide on the command-line?**
+>
+> I believe we can keep this small.
+> I propose we only add a few new options, and I'm not sure if we should even go that far:
```
celinval, line 81: Leave here the ones you think should be part of the proposal. The ones you are not so sure, you can add them to the open questions.

rfc/src/rfcs/0003-resource-limits-timeouts.md
```diff
@@ -0,0 +1,224 @@
...
+> 1. `--default-timeout`
+> 2. `--default-memory`
+> 3. `--timeout-multiplier=X`
+> 4. `--ignore-limits`
+>
+> The purpose of the first two is not really for the command line but instead `Cargo.toml` (which currently is just an alternative way to supply command line arguments).
```
celinval, line 87: I think default timeout makes a lot of sense for CIs.

rfc/src/rfcs/0003-resource-limits-timeouts.md
```diff
@@ -0,0 +1,224 @@
...
+>
+> The purpose of the multiplier is to allow a slow/inconsistent CI system to have an extra "grace period" for timeouts.
+> This way, CI can run `cargo kani --timeout-multiplier=1.3` to get 30% extra slack on timeouts.
```
celinval, line 90: This one I'm not so convinced. I can see where you are coming from (our github workers are rather slow), but I would also advocate for looser timeouts.

celinval (commented):
Thanks for putting this together, @tedinski. Are we planning to add any global limit on memory consumption?

As a user, I would prefer to set a maximum memory consumption for Kani, not per harness, so Kani doesn't kill my machine. I think setting a per harness limit might not be super intuitive for users unless if they have a good understanding on how Kani works in the background. What would be the guidelines for a good memory limit for a specific harness? It probably also depends on how many harnesses we are running in parallel.

On the document structure itself, I would suggest moving most of the questions into the open questions section. It's a bit hard to understand what is part of the current proposal, what are things that might still need some exploration and what's out of the scope.

zpzigi754:
This proposal seems to focus on less relying on the unwinding value. I am thinking of addressing the similar problem by allowing users to **assign finer-grained unwinding values** for each iterator or loop, briefly mentioned in (https://github.com/model-checking/kani/issues/1808). I think in this direction because some of resource consumption is caused by imposing the same big unwinding value to all iterators or loops even if it is needed for only a few iterators or loops in a harness.

tedinski:
(Note to self from another discussion:) We might want to add "filters" based on resource limits to our runner.

e.g. `cargo kani --filter="timeout<30"` or `cargo kani --filter="timout>=30"` or something, to support "run fast harnesses on pull request, run slow harnesses nightly" kind of workflows.

rfc/src/rfcs/0003-resource-limits-timeouts.md
```diff
@@ -0,0 +1,224 @@
...
+This shows our abuse of `unwind` to force termination.
+We would prefer this to be unnecessary, and for the user to be introduced to unwinding via a default timeout.
+
+In the case where we do want to give higher limits, users should have to write something like:
+
+```rust
+#[cfg(kani)]
+#[kani::proof]
+#[kani::timeout(60)]
+#[kani::memory(10)]
```
danielsn, line 55: Make the units explicit

tedinski:
Some prior art on timeouts from another tool: https://nexte.st/book/slow-tests.html

rfc/src/rfcs/0003-resource-limits-timeouts.md
```diff
@@ -0,0 +1,224 @@
...
+>
+> I argue yes: We're now proliferating the number of attributes, so perhaps it's time to consider supporting:
+> `#[kani::proof(unwind = 5, timeout = 60, memory = 10)]`
+>
+> As we'll see later, the disadvantage of default limit choices is the friction of adding limit annotations to harnesses.
+> Making this as easy as possible helps mitigate those disadvantages.
+
+> **Question 2: Should we include units in the attributes?**
+>
+> I'm presently giving "seconds" and "gigabytes" and these seem totally reasonable units that we probably wouldn't ever deviate from.
```
danielsn, line 69: Could we take a "Duration"?

Issue #1822, opened 2022-11-01
Compilation fails for s2n-quic after toolchain update to nightly-2022-10-25


zhassan-aws:
Steps to reproduce:

1. `git clone https://github.com/aws/s2n-quic`
2. `cd s2n-quic/quic/s2n-quic-core`
3. `cargo kani --tests --harness vectored_copy_fuzz_test`

```

error[E0080]: evaluation of constant value failed
  --> quic/s2n-quic-core/src/stream/testing.rs:20:13
   |
20 |             idx += 1;
   |             ^^^^^^^^ exceeded interpreter step limit (see `#[const_eval_limit]`)

error[E0080]: could not evaluate static initializer
  --> quic/s2n-quic-core/src/stream/testing.rs:25:25
   |
25 |     Bytes::from_static(&INNER)
   |                         ^^^^^ referenced constant has errors

For more information about this error, try `rustc --explain E0080`.
Error: "Failed to compile crate."
warning: `s2n-quic-core` (lib test) generated 2 warnings
error: could not compile `s2n-quic-core` due to 2 previous errors; 2 warnings emitted
Error: cargo exited with status exit status: 101
```

zhassan-aws:
Blocked by https://github.com/rust-lang/rust/issues/103814.

camshaft:
I was also able to fix this issue by simplifying the expression
https://github.com/aws/s2n-quic/pull/1559. You can revert your downgrade if this was the only blocker.

zhassan-aws:
Thanks @camshaft. Closing this issue.

PR #1830, opened 2022-11-01
Downgrade the toolchain to avoid blocking issue

celinval:
### Description of changes: 

Downgrade the rust toolchain for the nighly before the current version. This mitigates https://github.com/model-checking/kani/issues/1822 which is caused by https://github.com/rust-lang/rust/issues/103814.

Ps.: I also bumped cbmc min version since older versions don't work with Kani.

### Resolved issues:

Mitigates #1822 

### Related RFC:

N/A

### Testing:

* How is this change tested? New test + manually tested against s2n-quic.

### Checklist

### DIFF
--- a/scripts/kani-regression.sh
+++ b/scripts/kani-regression.sh
@@ -22,7 +22,7 @@ KANI_DIR=$SCRIPT_DIR/..
 export KANI_FAIL_ON_UNEXPECTED_DESCRIPTION="true"
 
 # Required dependencies
-check-cbmc-version.py --major 5 --minor 67
+check-cbmc-version.py --major 5 --minor 69
 check-cbmc-viewer-version.py --major 3 --minor 5
 
 # Formatting check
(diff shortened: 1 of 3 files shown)


tedinski (approved):

scripts/kani-regression.sh
```diff
@@ -22,7 +22,7 @@ KANI_DIR=$SCRIPT_DIR/..
 export KANI_FAIL_ON_UNEXPECTED_DESCRIPTION="true"
 
 # Required dependencies
-check-cbmc-version.py --major 5 --minor 67
+check-cbmc-version.py --major 5 --minor 69
```
tedinski, line 25: I was just running into this (without updating cbmc you get a weird `goto-cc` failure in the test suite) so glad you snuck this in.

zhassan-aws (approved):
Interesting to know that the offending commit occurred on 10/25 itself!

celinval:
> Interesting to know that the offending commit occurred on 10/25 itself!

Bad luck, I guess... :)

celinval:
Downgrade the rust toolchain for the nighly before the current version. This mitigates https://github.com/model-checking/kani/issues/1822 which is caused by https://github.com/rust-lang/rust/issues/103814.

Ps.: I also bumped cbmc min version since older versions don't work with Kani.

Issue #1531, opened 2022-08-16
Audit / refactor the cast code


celinval:
I don't think we have thoroughly tested the cast logic and we should probably revisit some assertions as well as the code structure in general. Issues I noticed while working on #1522:

- Incorrect code path. E.g.: casting `std::boxed::Box<dyn Byte + std::marker::Sync>` to `std::boxed::Box<dyn Byte>` was invoking `cast_sized_expr_to_unsized_expr`.
- Incorrect assumptions / misleading names: E.g.:
   - Functions like `cast_sized_pointer_to_fat_pointer` and  `cast_sized_expr_to_unsized_expr` operates under the assumption that we are casting sized to unsized, but at the same time they start checking if the type of source is the same type as destination.
   - Function `cast_sized_expr_to_unsized_expr` is not necessarily casting a sized expression to an unsized expression. A big part of it does thin pointer to fat pointer conversion. Both generate pointers which are sized expressions.
   - Trivially true assertions and (hopefully) unreachable code paths. E.g.:
     - https://github.com/celinval/kani-dev/blob/07bb6bcef2ed4cd46b3a04750bc60a75391c3b76/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs#L1088-L1105

Note: I renamed `cast_sized_expr_to_unsized_expr` to `cast_expr_to_unsized_expr` in #1522 to alleviate the first inconsistency, but it is still incorrect.

FYI: the issues we've seen so far trigger internal compiler errors (e.g.: #1528).

Issue #1528, opened 2022-08-16
Kani crashes when casting Rc<dyn Trait + Sync> to Rc<dyn Trait>.


celinval:
I tried this code (see [tests/kani/DynTrait/unsized_rc_cast_fixme.rs](https://github.com/model-checking/kani/pull/1522/files#diff-78b45c8407518ab37acaddeac260f599ebf6c4474bf01c43bce6f05f7587b46a)):

```rust
// Copyright Kani Contributors
// SPDX-License-Identifier: Apache-2.0 OR MIT
//! Check that we can cast between two unsized objects.
//! This fix-me is derived from unsized_rc_cast.rs and it should be merged with the original test.
use std::rc::Rc;

trait Byte {
    fn eq(&self, byte: u8) -> bool;
}

impl Byte for u8 {
    fn eq(&self, byte: u8) -> bool {
        *self == byte
    }
}

fn all_zero_rc(num: Rc<dyn Byte>) -> bool {
    num.eq(0x0)
}

#[kani::proof]
fn check_rc() {
    let num: u8 = kani::any();
    kani::assume(num != 0);
    let rc: Rc<dyn Byte + Sync> = Rc::new(num);
    assert!(!all_zero_rc(rc));
}
```

using the following command line invocation:

```
kani unsized_rc_cast_fixme
```

with Kani version:

I expected to see this happen: Verification succeed

<details>
<summary> Instead, this happened: Kani crashes during compilation. </summary>

```
thread 'rustc' panicked at 'Type std::rc::RcBox<dyn Byte> must be a trait type (a dynamic type)', kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs:594:9
stack backtrace:
   0:     0x7f85e38c1b70 - std::backtrace_rs::backtrace::libunwind::trace::h83eef7e1b99ceac2
                               at /rustc/fe3342816a282949f014caa05ea2e669ff9d3d3c/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   1:     0x7f85e38c1b70 - std::backtrace_rs::backtrace::trace_unsynchronized::h11443e110f08dce7
                               at /rustc/fe3342816a282949f014caa05ea2e669ff9d3d3c/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7f85e38c1b70 - std::sys_common::backtrace::_print_fmt::h98d30f5880b6fe77
[...]

Issue #566, opened 2021-10-19
DST example fails to codegen


adpaco-aws:
The following example from `The Rustonomicon/Data Layout/Exotically Sized Types/47.rs` shown below

```rust
// compile-flags: --edition 2018
#![allow(unused)]
struct MySuperSliceable<T: ?Sized> {
    info: u32,
    data: T,
}

pub fn main() {
    let sized: MySuperSliceable<[u8; 8]> = MySuperSliceable {
        info: 17,
        data: [0; 8],
    };

    let dynamic: &MySuperSliceable<[u8]> = &sized;

    // prints: "17 [0, 0, 0, 0, 0, 0, 0, 0]"
    println!("{} {:?}", dynamic.info, &dynamic.data);
}
```

fails to codegen with
```
thread 'rustc' panicked at 'Only an array can be cast to a slice.  Found types Adt(MySuperSliceable, [[u8; 8]]) and Adt(MySuperSliceable, [[u8]])', compiler/rustc_codegen_rmc/src/codegen/rvalue.rs:1076:37
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

zhassan-aws:
Still crashes with the same signature as of `af11d38d93a`.

PR #1837, opened 2022-11-02
Refactor coercion logic for future usage inside codegen

celinval:
### Description of changes: 

We currently have a few issues with how we are generating code for casting (#566, and #1528). The structure of the code is also hard to understand and maintain (see #1531 for more details).

This PR is the first part of the fix I developed. This change moves the coercion specific code to its own module and it introduces an iterator that traverses the coercion path.

### Resolved issues:

None

### Testing:

### Checklist


kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+///
+/// This is used to determine the vtable implementation that must be tracked by the fat pointer.
+///
+/// For example, if `&u8` is being converted to `&dyn Debug`, this method would return:
+/// `(u8, dyn Debug)`.
+///
+/// There are a few interesting cases (references / pointers are handled the same way):
+/// 1. Coercion between `&T` to `&U`.
+///    - This is the base case.
+///    - In this case, we extract the type that is pointed to.
```
danielsn, line 37: We extract a pair, not just one type

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+/// 2. Coercion between smart pointers like `Rc<T>` to `Rc<U>`.
+///    - Smart pointers implement the `CoerceUnsize` trait.
+///    - Use CustomCoerceUnsized information to traverse the smart pointer structure and find the
+///      underlying pointer.
+///    - Use base case to extract `T` and `U`.
+/// 3. Coercion between `&Wrapper<T>` to `&Wrapper<U>`.
```
danielsn, line 43: `Wrapper` here is a standin for some user defined type I assume? Maybe give an example of a user defined DST where this happens

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+///    - Apply base case to extract the pair `(Wrapper<T>, Wrapper<U>)`.
+///    - Extract the tail element of the struct which are of type `T` and `U`, respectively.
+/// 4. Coercion between smart pointers of wrapper structs.
+///    - Apply the logic from item 2 then item 3.
```
danielsn, line 47: can this nest arbitrarily deep? Smart pointer of wrapper of smart pointer?

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+pub fn extract_unsize_casting<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    src_ty: Ty<'tcx>,
+    dst_ty: Ty<'tcx>,
+) -> (Ty<'tcx>, Ty<'tcx>) {
```
danielsn, line 52: Unnamed pairs where both are of the same type are a 🦶🔫

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+    trace!(?dst_ty, ?src_ty, "find_trait_conversion");
+    let info = CoerceUnsizedIterator::new(tcx, src_ty, dst_ty).last().unwrap();
```
danielsn, line 54: `info` is not a very helpful name here

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+    let src_base_ty = extract_pointee(info.src_ty)
+        .expect(&format!("Expected source to be a pointer. Found {:?} instead", info.src_ty));
+    let dst_base_ty = extract_pointee(info.dst_ty)
+        .expect(&format!("Expected destination to be a pointer. Found {:?} instead", info.dst_ty));
+    tcx.struct_lockstep_tails_erasing_lifetimes(src_base_ty, dst_base_ty, ParamEnv::reveal_all())
```
danielsn, line 59: You described 4 cases, but all the magic happens here?

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+/// Iterates over the coercion path of a structure that implements `CoerceUnsized<T>` trait.
+/// The `CoerceUnsized<T>` trait indicates that this is a pointer or a wrapper for one, where
+/// unsizing can be performed on the pointee. More details:
+/// <https://doc.rust-lang.org/std/ops/trait.CoerceUnsized.html>
+///
+/// Given an unsized coercion between `impl CoerceUnsized<T>` to `impl CoerceUnsized<U>` where
+/// `T` is sized and `U` is unsized, this iterator will walk over the fields that lead to a
+/// pointer to `T`, which shall be converted from a thin pointer to a fat pointer.
+///
+/// This iterator will also include the name of the field that differs in the current iteration.
```
danielsn, line 71: where is this stored?

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+/// be coerced, as well as the current source and the destination.
+/// E.g.: The first iteration of casting `Rc<String>` -> `Rc<&dyn Debug>` will return
+/// ```rust,ignore
+/// CoerceUnsizedInfo { field: "ptr", src_ty /* Rc<String> */, dst_ty /* Rc<&dyn Debug> */}
+/// ```
+/// while the last iteration will return:
+/// ```rust,ignore
+/// CoerceUnsizedInfo {
+///   field: None,
+///   src_ty: Ty, /* *const RcBox<String> */,
```
danielsn, line 113: So it stops at the box, and doesn't go inside?  I guess that makes some sense cause the box is really a pointer?

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+                assert_eq!(src_def, dst_def);
+                let src_fields = &src_def.non_enum_variant().fields;
+                let dst_fields = &dst_def.non_enum_variant().fields;
+                assert_eq!(src_fields.len(), dst_fields.len());
+
+                let CustomCoerceUnsized::Struct(coerce_index) =
+                    custom_coerce_unsize_info(self.tcx, src_ty, dst_ty);
+                assert!(coerce_index < src_fields.len());
+
+                self.src_ty = Some(src_fields[coerce_index].ty(self.tcx, src_substs));
```
danielsn, line 143: assert that src_type is sized and des_type is unsized?

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+                None
+            }
+        };
+        Some(CoerceUnsizedInfo { field, src_ty, dst_ty })
+    }
+}
+
+/// Get information about an unsized coercion.
+/// This code was extracted from `rustc_monomorphize` crate.
+/// <https://github.com/rust-lang/rust/blob/4891d57f7aab37b5d6a84f2901c0bb8903111d53/compiler/rustc_monomorphize/src/lib.rs#L25-L46>
```
danielsn, line 162: Should we have a tracking issue to boost this into main rustc?

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,188 @@
...
+    target_ty: Ty<'tcx>,
+) -> CustomCoerceUnsized {
+    let def_id = tcx.require_lang_item(LangItem::CoerceUnsized, None);
+
+    let trait_ref = ty::Binder::dummy(TraitRef {
+        def_id,
+        substs: tcx.mk_substs_trait(source_ty, &[target_ty.into()]),
+    });
+
+    match tcx.codegen_select_candidate((ParamEnv::reveal_all(), trait_ref)) {
```
danielsn, line 175: This is magic.  I'm going to assume it works 🪄

kani-compiler/src/kani_middle/coercion.rs
celinval: No. The casting stop in the first layer.

kani-compiler/src/kani_middle/coercion.rs
celinval: Not really. Lines 55 / 57 would be enough for case 1. Line 54 is needed for case 2 and 4. This logic is needed for cases 3 and 4. I.e., this takes care of traversing the concrete type after we traversed the pointer type.

kani-compiler/src/kani_middle/coercion.rs
celinval: It's not stored. It's generated as part of each iteration.

kani-compiler/src/kani_middle/coercion.rs
celinval: This is not a `Box` though, it's an `RcBox` which is the internal structure of `Rc`.

kani-compiler/src/kani_middle/coercion.rs
celinval: `src_type` can be unsized though. This can happen when you are doing an unsized cast to remove an auto trait. E.g.: `Rc<dyn Trait + Send>` to `Rc<dyn Trait>`. See https://github.com/model-checking/kani/issues/1528 for an example where this is relevant.

kani-compiler/src/kani_middle/coercion.rs
celinval: We could, but it's fairly low in our priority list.

kani-compiler/src/kani_middle/coercion.rs
celinval: I.e.: You can't do:
```rust
    let concrete = Rc::new(Box::new(String::from("hi")));
    let dynamic : Rc<Box<dyn Display>> = concrete; 
```
You need to first do the coercion, then nest things:
```rust
    let boxed = Box::new(String::from("hi"));
    let dynamic : Box<dyn Display> = boxed; // Coercion happens here.
    let rc = Rc::new(dynamic); // The external Rc only contains a thin pointer.
```
See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0218b81a56f650edb521a3ec2ef72654

kani-compiler/src/kani_middle/coercion.rs
celinval: Ok ok... I'll add a new struct. =P

kani-compiler/src/kani_middle/coercion.rs
celinval: TBH, this is local to this function, so I don't think it's a big deal, but I'll rename it.

kani-compiler/src/kani_middle/coercion.rs
celinval: I added some comments so hopefully it helps understand the logic.

kani-compiler/src/kani_middle/coercion.rs
celinval: I changed to example to use `NonNull`, so hopefully it's a bit easier to understand.

kani-compiler/src/kani_middle/coercion.rs
celinval: I added the assertion in a different function.

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,232 @@
...
+//! There are many types of coercions in rust, they are described in the
+//! [RFC 401 Coercions](https://rust-lang.github.io/rfcs/0401-coercions.html).
+//!
+//! The more complicated coercions are DST Coercions (aka Unsized Coercions). These coercions
+//! allow rust to create references to dynamically sized types (such as Traits and Slices) by
+//! casting concrete sized types. Unsized coercions can also be used to cast unsized to unsized
+//! types. These casts work not only on the top of references, but it also handle
+//! references inside structures, allowing the unsized coercions of smart pointers. This last
+//! piece was defined by
+//! [RFC 982 DST Coercion](https://rust-lang.github.io/rfcs/0982-dst-coercion.html).
```
tedinski, line 14: Are all unsized coercions constructing fat pointers of some kinds?

Also nit: can you be explicit about "this last piece". It's a stupid writing thing that bugs me, but I like less indirection in technical writing even if it means slightly more repetition.

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,232 @@
...
+
+/// Extract the pair (`T`, `U`) for a unsized coercion where type `T` implements `Unsize<U>`.
+/// I.e., `U` is either a trait or a slice.
+/// For more details, please refer to:
+/// <https://doc.rust-lang.org/reference/type-coercions.html#unsized-coercions>
+///
+/// This is used to determine the vtable implementation that must be tracked by the fat pointer.
+///
+/// For example, if `&u8` is being converted to `&dyn Debug`, this method would return:
+/// `(u8, dyn Debug)`.
```
tedinski, line 32: I wonder if we could rephrase this opening to start with the concrete example instead of only coming to it later.

> Given an unsized coercion (e.g. from `&u8` to `&dyn Debug`), extract the pair of coresponding base types `T, U` (e.g. `u8`, `dyn Debug`), where the source base type `T` must implement `Unsize<U>` and `U` is either a trait or slice.

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,232 @@
...
+/// There are a few interesting cases (references / pointers are handled the same way):
+/// 1. Coercion between `&T` to `&U`.
+///    - This is the base case.
+///    - In this case, we extract the types that are pointed to.
+/// 2. Coercion between smart pointers like `NonNull<T>` to `NonNull<U>`.
+///    - Smart pointers implement the `CoerceUnsize` trait.
+///    - Use CustomCoerceUnsized information to traverse the smart pointer structure and find the
+///      underlying pointer.
+///    - Use base case to extract `T` and `U`.
+/// 3. Coercion between a pointer to a structure whose tail is being coerced.
```
tedinski, line 43: This explanation is the clearest I've seen yet! Nice!

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,232 @@
...
+    );
+    CoercionBase { src_ty: src_base_ty, dst_ty: dst_base_ty }
+}
+
+/// This structure represents the base of a coercion.
+///
+/// This base is used to determine the information that will be stored in the metadata.
+/// E.g.: In order to convert an `Rc<String>` into an `Rc<dyn Debug>`, we need to generate a
+/// vtable that represents the `impl Debug for String`. So this type will carry the `String` type
+/// as the `src_ty` and the `Debug` trait as `dst_ty`.
```
tedinski, line 94: Do we want to be consistent about saying `dyn Debug`?

tedinski (approved):
I know the answer is "it's too hard to write" but I wish we could write some unit tests for these functions. Oh well.

celinval:
> I know the answer is "it's too hard to write" but I wish we could write some unit tests for these functions. Oh well.

I believe the answer is "there is no way to write one". I don't think we can create a custom object of type `Ty`.

kani-compiler/src/kani_middle/coercion.rs
celinval: Yes. Done.

kani-compiler/src/kani_middle/coercion.rs
celinval: sure

kani-compiler/src/kani_middle/coercion.rs
```diff
@@ -0,0 +1,234 @@
...
+/// Iterates over the coercion path of a structure that implements `CoerceUnsized<T>` trait.
+/// The `CoerceUnsized<T>` trait indicates that this is a pointer or a wrapper for one, where
+/// unsizing can be performed on the pointee. More details:
+/// <https://doc.rust-lang.org/std/ops/trait.CoerceUnsized.html>
+///
+/// Given an unsized coercion between `impl CoerceUnsized<T>` to `impl CoerceUnsized<U>` where
+/// `T` is sized and `U` is unsized, this iterator will walk over the fields that lead to a
+/// pointer to `T`, which shall be converted from a thin pointer to a fat pointer.
+///
+/// This iterator will also include the name of the field that differs in the current iteration
```
danielsn, line 112: Isn't it the `Info` not the `iterator` that has the field name? So we should really say "the result of iterating with this iterator will ..."

danielsn (approved):
Few minor comments, but LGTM

kani-compiler/src/kani_middle/coercion.rs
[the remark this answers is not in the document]
celinval: `(u8, dyn Debug)`.

kani-compiler/src/kani_middle/coercion.rs
celinval: Indeed. Let me improve that.

Issue #1809, opened 2022-10-27
Tracking issue for function and method stubbing


aaronbembenek-aws:
# Summary

This is a tracking issue for the implementation of function and method stubbing, which will enable users to swap in alternate definitions of functions and methods during verification.

During development, this feature will be gated by the unstable option `--enable-stubbing`.

# Major Steps

Suggestion of major steps:

- [x] Get feedback on the `function_stubbing` RFC.
- [x] Implement `kani::stub` attribute.
- [x] Implement `--enable-stubbing` option. This will initially require `--harness` as well.
- [x] Implement basic stubbing mechanism, requiring a single harness, and doing no path resolution for function/method names.
- [X] Implement stub validation.
- [ ] Implement path resolution for function/method names.
- [ ] Implement mechanism removing single harness requirement.
- [ ] Stabilization PR.

PR #1838, opened 2022-11-02
Move attribute processing functions to kani_middle.

aaronbembenek-aws:
### Description of changes: 

This is a refactoring PR affecting `kani-compiler`. It moves functions for generically processing Rust attributes (like getting the argument of an attribute) from the `codegen_cprover_gotoc` module to the new `kani_middle` module. These functions are backend agnostic, and so should be in `kani_middle`. The stubbing code (which will also sit in `kani_middle`) will use these functions.

### Related issues:

Related to #1833 and #1809 

### Testing:

### Checklist


kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! This module contains code for processing Rust attributes (like `kani::proof`).
+
+use rustc_ast::{AttrKind, Attribute, LitKind};
+
+/// If the attribute is named `kanitool::name`, this extracts `name`
+fn kanitool_attr_name(attr: &Attribute) -> Option<String> {
```
celinval, line 8: nit: can you move the public methods to the top of the file? That's usually the entry point of the logic, so it's easier to read if they are on the top. :)

celinval (approved):
Sounds good to me

kani-compiler/src/kani_middle/attributes.rs
aaronbembenek-aws: Good idea! Done

PR #1840, opened 2022-11-03
Only copy to sysroot libraries from latest compilation

### LINKED ISSUE #1805: Fix multiple matching crates issue when rebuilding Kani
After upgrading the toolchain, I reran:

```
cargo build-dev
```

When I tried to run Kani, I get the following error:
```
error[E0464]: multiple matching crates for `kani`
  --> /kani/library/std/src/lib.rs:41:9
   |
41 |         kani::assert($cond, concat!("assertion failed: ", stringify!($cond)));
   |         ^^^^
   |
   = note: candidates:
           crate `kani`: /kani/target/kani/lib/rustlib/x86_64-unknown-linux-gnu/lib/libkani-709d4158b08b446e.rlib
           crate `kani`: /kani/target/kani/lib/rustlib/x86_64-unknown-linux-gnu/lib/libkani-f85c4bb36b420d60.rlib

```
I believe this is caused by the fact that cargo keep old rlib files behind when their hash changes AND we export `libkani` and all the other libraries without specifying their full path. So the compiler resolution can't pick which one to use.

I think the easiest solution is to only copy to target/kani/lib/ the files that are generated by the current build.

celinval:
### Description of changes: 

As part of `cargo build-dev`, we recreate the folder `target/kani` which will include libraries used by the compiler. We used to blindly copy all files inside the build folder, but this may contain older versions of certain libraries triggering issues while resolving dependencies.

Now, we only copy artifacts from the current build.

### Resolved issues:

Resolves #1805 

### Testing:

* How is this change tested? Current tests. Manually I created redundant files and made sure they don't get copied over.

### Checklist


tools/build-kani/src/sysroot.rs
```diff
@@ -111,96 +111,119 @@ pub fn build_lib() {
...
+    fs::create_dir_all(&std_path).expect(&format!("Failed to create {std_path:?}"));
 
-    //  Copy Kani libraries to inside sysroot folder.
-    let target_folder = Path::new(target_dir);
-    let macro_lib = format!("libkani_macros{}", lib_extension());
-    let kani_macros = path_buf!(target_folder, "debug", macro_lib);
-    cp(&kani_macros, &sysroot_lib).unwrap();
+    //  Copy Kani libraries to inside sysroot top folder.
+    copy_libs(&artifacts, &sysroot_lib, &is_kani_lib);
+    //  Copy Kani libraries to inside rustlib/<target>/lib/ folder.
```
celinval, line 126: I mean `std` libraries. I'll fix that in the next revision.

tools/build-kani/src/sysroot.rs
```diff
@@ -111,96 +111,119 @@ pub fn build_lib() {
...
+    is_rust_lib(artifact) && artifact.target.src_path.starts_with(env!("KANI_REPO_ROOT"))
+}
 
-    let dst_path = path_buf!(sysroot_lib, "rustlib", target, "lib");
-    fs::create_dir_all(&dst_path).unwrap();
-    cp_files(&src_path, &dst_path, &is_rlib).unwrap();
+/// Is this a std library or one of its dependencies.
+fn is_std_lib(artifact: &Artifact) -> bool {
+    is_rust_lib(artifact) && !is_kani_lib(artifact)
 }
```
tedinski, line 147: Non-blocking, but is there a more robust way to write these?

I also wonder if we could write unit tests purely as documentation (I'm less worried about actually testing behavior and more wondering "what files does this actually match?" It's very hard to understand just from code review.)

tedinski (approved):

PR #1844, opened 2022-11-03
Fix cbmc-viewer ci issues with Mac11

jaisnan:
### Description of changes: 

CI is broken because of an issue in actions - https://github.com/actions/runner-images/issues/6507. This is a patch to fix viewer installation in mac os till the issue is fixed.

### Call-outs:

We aren't specifying version number with Mac because that breaks the CI. Brew is unable to recognize specific version numbers of CBMC-Viewer and can only install the latest version.

### Testing:

* How is this change tested? Regression and CI in fork

### Checklist


scripts/setup/macos/install_viewer.sh
```diff
@@ -0,0 +1,20 @@
...
+
+if [ -z "${CBMC_VIEWER_VERSION:-}" ]; then
+  echo "$0: Error: CBMC_VIEWER_VERSION is not specified"
+  exit 1
+fi
+
+set -x
+
+# brew doesn't recognize specific versions of viewer
+brew install aws/tap/cbmc-viewer
```
danielsn, line 20: Is this always set at this point? Where is this tap added?

scripts/setup/macos/install_viewer.sh
```diff
@@ -0,0 +1,20 @@
...
+source kani-dependencies
+
+if [ -z "${CBMC_VIEWER_VERSION:-}" ]; then
+  echo "$0: Error: CBMC_VIEWER_VERSION is not specified"
+  exit 1
+fi
+
+set -x
+
+# brew doesn't recognize specific versions of viewer
```
danielsn, line 19: Are we likely to care? Should we ask the tap owners to keep pinned versions in the tap?

scripts/setup/macos/install_viewer.sh
jaisnan: `brew install repo` mitigates the need to add a tap but I could make that more explicit if it's safer.

scripts/setup/macos/install_viewer.sh
jaisnan: I can't imagine this becoming a problem since viewer isn't updated that often and also `brew install` installs the latest version of viewer anyway. This patch is also just temporary I imagine, till the issue in github-actions gets fixed or till we find a way to add a path to `pip install`. This was just the quickest solution, to get everyone's PR's unblocked by CI.

scripts/setup/macos/install_viewer.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+set -eu
+
+# Install cbmc-viewer
+
+# Source kani-dependencies to get CBMC_VIEWER_VERSION
+source kani-dependencies
```
zhassan-aws, line 10: If we're not installing a specific version, then we shouldn't need to source kani-dependencies.

scripts/setup/macos/install_viewer.sh
jaisnan: This is the issue that's causing ci to break, I imagine - https://github.com/actions/runner-images/issues/6507.

celinval: Can we push the fix to unblock us and we can figure out a better long term plan later? If https://github.com/actions/runner-images/issues/6507 gets fixed, we could probably just revert these changes.

celinval (approved):
We should figure out later what we want to do in the long term, but it looks like this will unblock us.

Issue #208, opened 2021-06-16
Standard library function `panicking::take_hook` does not codegen


danielsn:
```
            // [compiler/rustc_codegen_llvm/src/gotoc/rvalue.rs:885] self.current_fn().readable_name() = "panicking::take_hook"
            // error: internal compiler error: compiler/rustc_middle/src/ty/fold.rs:829:17: Trying to collect bound vars with a bound region: DebruijnIndex(0) BoundRegion { var: 0, kind: BrAnon(0) }

```

related #109

zhassan-aws:
This doesn't fail as of `4fe8370a2a8` on the following example:
```
use std::panic;

fn main() {
    panic::set_hook(Box::new(|_| {
        println!("Custom panic hook");
    }));

    let _ = panic::take_hook();

    panic!("Normal panic");
}
```

zhassan-aws:
This stopped working at some point (possibly due to the switch to using `assert-false-assume-false` for missing functions). It seems that the `set_hook/take_hook` functions need the standard library to be linked as their definition is missing. The example above now fails with:
```
Check 9: _ZN3std9panicking8set_hook17hd9797ff1f60891f8E.assertion.1
	 - Status: FAILURE
	 - Description: "assertion false"
	 - Location: Unknown File in function _ZN3std9panicking8set_hook17hd9797ff1f60891f8E
```
which suggests that the definition of `set_hook` is missing. If I remove `set_hook`:
```rust
fn main() {
    let _ = panic::take_hook();

    panic!("Normal panic");
}
```
I get a different failure due to a missing `take_hook` function:
```
Check 4: _ZN3std9panicking9take_hook17h69295202e6d47881E.assertion.1
	 - Status: FAILURE
	 - Description: "assertion false"
	 - Location: Unknown File in function _ZN3std9panicking9take_hook17h69295202e6d47881E
```

celinval:
There seems to still be an issue with this test even after we fixed the linking issue. The new failure has to do with `pthread` handling. With an unwind of `2`, this test now fails with the following error:

```
: function call: parameter "pthread_key_create::destructor" type mismatch:
got struct_tag
  * identifier: tag-_18057172096210256107
expected pointer
  * #source_location:
    * file: <builtin-library-pthread_key_create>
    * line: 11
    * function: pthread_key_create
    * working_directory: /tmp/StdLink
  * width: 64
  * #failed_symbol: pthread_key_create::destructor$object
  0: code
      * #source_location:
[...]

PR #1846, opened 2022-11-03
Remove fixme test because it may take a long time

celinval:
### Description of changes: 

This is an odd one, but I believe this test is causing some workflows to take hours. It doesn't always happen, and it depends on the machine, but I think for now we might as well remove this test.

E.g.: These runs took a long time, and the only test flagged as running above 60s is the one I'm removing now:
- https://github.com/model-checking/kani/actions/runs/3380137117/jobs/5612577106#step:5:238
- https://github.com/model-checking/kani/actions/runs/3348493009/jobs/5547606884#step:5:236

### Resolved issues:

N/A

### Related RFC:

N/A

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


celinval:
This fixme test was added to capture https://github.com/model-checking/kani/issues/208.

jaisnan (approved):
Thanks!

PR #1847, opened 2022-11-03
Use `<target>/kani` as the target directory and clean at every build

### LINKED ISSUE #1821: Target directories should be cleaned before codegen
I've run into trouble a couple of times:
 * When working on #1811 , `cargo-kani` regressions failed because conflicts were found in the GotoC programs I had generated. I think this is due to `cargo-kani` using the `--target-dir` option; not cleaning the directory first makes us reuse some code generated earlier.
 * When running the "unsupported features" experiment, it reported failures on half the crates because of some conflict. For this experiment, I need to run `--legacy-linker` (otherwise nothing is shown), and cleaning the experiments folder helped. I don't have a log file, but this one can be reproduced running the experiment first as is, and then with `--legacy-linker`.

But in general, whenever we are going to generate and link code, we should be cleaning the whole folder first. I believe some commands even use wildcards so it's no surprise this happens.

celinval:
### Description of changes: 

We can't really handle caching compilation artifacts today. So we are going to perform a `cargo clean` before building the goto program. I also changed our target to use the `kani` subfolder inside target.

We currently put cargo-kani compilation artifacts inside regular cargo target. This can conflict with users compilation artifacts and unnecessarily rebuild code). Now we force rebuild without modifying the target, that would always delete the entire user build folder, which seems like a terrible UX.

### Resolved issues:

Resolves #1821. 

### Testing:

* How is this change tested? Existing tests + manually tests.

### Checklist


kani-driver/src/call_cargo.rs
```diff
@@ -32,9 +32,14 @@ impl KaniSession {
...
             .as_ref()
             .unwrap_or(&metadata.target_directory.clone().into())
-            .clone();
+            .clone()
+            .join("kani");
         let outdir = target_dir.join(build_target).join("debug/deps");
 
+        // Clean directory before building since we are unable to handle cache today.
+        // TODO: https://github.com/model-checking/kani/issues/1736
+        self.cargo_clean(&target_dir)?;
```
zhassan-aws, line 41: Would deleting the `<target>/kani` directory instead work?

celinval: Sure. `cargo clean` seems like a good way to clean stuff generated by `cargo build`, but the effect is the same, so if you prefer, I can just remove the folder.

zhassan-aws: Thanks. I think the advantage of just removing the kani directory is to not wipe out the artifacts of a normal (non-kani) `cargo build`.

zhassan-aws: Which would hopefully make the UX a bit "less-terrible".

zhassan-aws (approved):
Thanks. Having another approval on this PR would be useful since this impacts the UX.

kani-driver/src/call_cargo.rs
celinval: BTW, we were providing the correct target directory to `cargo clean`, so it was only wiping out `target/kani` not the entire `target`. Anyway, not a biggie.

zhassan-aws: Sorry, I missed that.

PR #1848, opened 2022-11-04
Compiletest improvements

celinval:
### Description of changes: 

This adds `--timeout <val>` option to compiletest so we can add some hard timeout to some of our regression suites. My suggestion is to add timeout values for suites that are required to run in every PR. Suites like `slow_test` and `perf` may or may not have timeout.

I have also added a `--dry-run` that allow us to see the compiletest configuration and print the tests that were found.

For now, the only change I made to the regression is to print the compiletest configuration. This is to help us debug potential CI failures or timing issues.

### Resolved issues:

N/A

### Related RFC:

N/A

### Call-outs:

I think this is orthogonal to #1815 since this would be a per-harness timeout. Since one test could have multiple harnesses that together take longer than the suite timeout.

### Testing:

* How is this change tested? Manually ran compiletest in our suites with timeout of 60s. A bunch of tests failed. :(

### Checklist


tools/compiletest/src/common.rs
```diff
@@ -125,6 +126,12 @@ pub struct Config {
     /// Whether to rerun tests even if the inputs are unchanged.
     pub force_rerun: bool,
 
+    /// Timeout duration for each test.
+    pub timeout: Option<Duration>,
+
+    /// Flag that indicates whether we will run the tests or not.
```
adpaco-aws, line 132:
```suggestion
    /// Whether we will run the tests or not.
```

scripts/kani-regression.sh
```diff
@@ -64,6 +65,11 @@ else
   TESTS+=("no_unsound_experiments expected")
 fi
 
+# Build compiletest and print configuration. We pick suite / mode combo so there's no test.
+echo "--- Compiletest configuration"
+cargo run -p compiletest --quiet -- --suite kani --mode cargo-kani --dry-run --verbose
```
adpaco-aws, line 70: It's not clear to me why printing the configuration with an invalid suite/mode pair is useful here.

tools/compiletest/src/main.rs
```diff
@@ -205,6 +226,15 @@ pub fn run_tests(config: Config) {
         make_tests(c, &mut tests);
     }
 
+    if config.dry_run {
+        println!("Number of Tests: {}", tests.len());
+        for test in tests {
+            let ignore = if test.desc.ignore ^ config.run_ignored { "IGNORE" } else { "" };
+            println!(" - {} {}", test.desc.name.as_slice(), ignore);
```
adpaco-aws, line 233: Most of the output coming from compiletest doesn't print capital letters. Does this look similar to running without `--quiet`? Also, shouldn't `--quiet` or `--verbose` conflict with this flag?

adpaco-aws (commented):
Perhaps this discussion is more appropriate for the PR where we actually set a timeout, but... How do you plan on using this? I'd assume you want to detect slow regression tests before they're introduced, or get a heads-up (in the form of a CI failure) when we upgrade our dependencies.

celinval:
> Perhaps this discussion is more appropriate for the PR where we actually set a timeout, but... How do you plan on using this? I'd assume you want to detect slow regression tests before they're introduced, or get a heads-up (in the form of a CI failure) when we upgrade our dependencies.

That's a good question. I was hoping to change `kani-regression` and add timeout to each call to compiletest. Initially we would start with timeout that fits our current setup. I.e., the regression won't fail but if someone adds a test that takes even longer, the PR checks would fail.

scripts/kani-regression.sh
celinval: My main motivation to add this was to get more information out of the builders that are being used today. Specially, the number of threads used to run our tests. I thought it might be helpful to keep logging it to understand why some runs takes longer than others. So far, it looks like Mac uses 3 and Ubuntu workers run 2 tests in parallel. 

Not providing a `--suite` and a `--mode` is a hard failure today, and I wanted to limit the size of this change, since I'm really just trying to get some information to help us implement these time out. So I picked a combo that doesn't find any test but that doesn't fail the run.

I also found it very helpful to list the tests being executed, hence, the `--dry-run`. I'm happy to fine tune this or drop it. The timeout part is the most important anyway.

tools/compiletest/src/main.rs
celinval: I should probably just use `logv` here to ensure `verbose` and `quiet` are being respected. I'll change that.

I don't think there should be any conflict though.

adpaco-aws (approved):

Issue #1148, opened 2022-05-03
SIMD intrinsics: Restore and audit


adpaco-aws:
SIMD intrinsics were disabled in #1143 due to soundness concerns.

adpaco-aws:
Recent data (in #1400) has shown that SIMD intrinsics are among the top unsupported features. Some popular crates impacted by this unsupported feature are `rand` (random number generation) and `regex` (regular expressions). Therefore, we are planning to gradually add support for the following SIMD intrinsics:
 - [ ] `simd_add`
 - [ ] `simd_and`
 - [ ] `simd_div`
 - [ ] `simd_eq`
 - [ ] `simd_extract`
 - [ ] `simd_ge`
 - [ ] `simd_gt`
 - [ ] `simd_insert`
 - [ ] `simd_le`
 - [ ] `simd_lt`
 - [ ] `simd_mul`
 - [ ] `simd_ne`
 - [ ] `simd_or`
 - [ ] `simd_rem`
 - [ ] `simd_shl`
 - [ ] `simd_shr`
 - [ ] `simd_shuffle16`
 - [ ] `simd_shuffle2`
 - [ ] `simd_shuffle32`
 - [ ] `simd_shuffle4`
 - [ ] `simd_sub`
 - [ ] `simd_xor`

The tentative order is to first add support for comparison operators (e.g., `simd_eq`) which we never got to work. Then we'll continue with arithmetic/bit operators (e.g., `simd_add` and `simd_and`), for which we had support in the past but was never audited.

adpaco-aws:
Update:
 1. We added support in #1853 for SIMD vector comparison intrinsics: `simd_eq`, `simd_ne`, `simd_le`, `simd_ge`, `simd_lt` and `simd_gt`.
 2. We added support in #1893 for SIMD vector bitwise intrinsics: `simd_and`, `simd_or` and `simd_xor`.
 3. There is a PR up to add support for some SIMD vector arithmetic intrinsics: `simd_add`, `simd_sub` and `simd_mul`.

Bitshift intrinsics (`simd_lhs` and `simd_rhs`) are missing checks which apparently get automatically included for regular bitshift operations by the Rust compiler (see https://github.com/diffblue/cbmc/issues/7362). Same goes for the division and remainder operators (`simd_div` and `simd_rem`), for which we're missing overflow checks in the SIMD version of the operators (see https://github.com/diffblue/cbmc/issues/7367).

The remaining work is:
[...]

PR #1853, opened 2022-11-04
Add support for SIMD vector comparison intrinsics

adpaco-aws:
### Description of changes: 
 1. Declares `Vector*` expressions which generate `vector-*` ireps, as regular operators (e.g., `==` or `<=`) aren't picked up by CBMC. This also makes our own type-checking code cleaner.
 2. Re-enables SIMD intrinsics.
 3. Performs type-checking for cases where the Rust front-end doesn't check.
 4. Adds some utilities functions for vector types.
 6. Moves the `SIMD` tests folder to `Intrinsics`.
 7. Splits our test case into unsigned integers, signed integers, and floats. 

### Resolved issues:

Towards #1148 

### Testing:

* How is this change tested? Existing regression + more cases.

### Checklist


cprover_bindings/src/goto_program/typ.rs
```diff
@@ -71,7 +71,23 @@ pub enum Type {
...
+
+/// Contains information about SIMD vector types
+#[derive(Debug, Clone)]
+pub struct VectorData {
+    pub typ: Box<Type>,
+    pub size: u64,
+}
+
+impl PartialEq for VectorData {
+    /// Rust allows conversions between SIMD vector types with the same size. By
```
celinval, line 85: I'm a bit curious about this one. Do you have any reference? What kind of conversions are we talking about? I can see conversions between signed and unsigned, but does it allow conversions between float and integer types? Does it allow conversions between types with different sizes?

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -308,6 +323,14 @@ impl Type {
         }
     }
 
+    pub fn size(&self) -> Option<u64> {
```
celinval, line 326: Can we use `len()` instead? Can you also please add a comment to this method.

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -1041,6 +1064,19 @@ impl Type {
         Pointer { typ: Box::new(self) }
     }
 
+    /// Convert type to its signed counterpart if possible.
+    /// For types that are already signed, this will return self.
```
celinval, line 1068: I don't think this is returning `Self` for `ssize_t`

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1361,6 +1361,19 @@ impl<'tcx> GotocCtx<'tcx> {
         )
     }
 
+    fn codegen_intrinsic_simd_cmp<F: FnOnce(Expr, Expr, Type) -> Expr>(
```
celinval, line 1364: comment please. :)

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -611,21 +611,21 @@ impl<'tcx> GotocCtx<'tcx> {
...
-            "simd_eq" => unstable_codegen!(codegen_intrinsic_binop!(eq)),
+            "simd_eq" => self.codegen_intrinsic_simd_cmp(Expr::vector_eq, fargs, p, cbmc_ret_ty),
             "simd_extract" => {
                 let _vec = fargs.remove(0);
                 let _index = fargs.remove(0);
                 unstable_codegen!(self.codegen_expr_to_place(p, vec.index_array(index)))
             }
-            "simd_ge" => unstable_codegen!(codegen_intrinsic_binop!(ge)),
-            "simd_gt" => unstable_codegen!(codegen_intrinsic_binop!(gt)),
+            "simd_ge" => self.codegen_intrinsic_simd_cmp(Expr::vector_ge, fargs, p, cbmc_ret_ty),
```
celinval, line 620: I'm so happy you are not using a macro here! :)

tests/expected/intrinsics/simd-invalid-conversion/expected
```diff
@@ -0,0 +1,2 @@
+Kani unexpectedly panicked during compilation
```
celinval, line 1: Is this expected? Can we make this a user friendly error?

celinval (commented):
Can you please make sure the `fixme` tests have an explanation on why they are failing and an issue associated with it?

cprover_bindings/src/goto_program/typ.rs
adpaco-aws: Sorry about the confusion, I shouldn't have used "conversions" here nor in the tests. Conversions as such (i.e., with `as`) don't seem to be allowed for SIMD types.

Unfortunately, there isn't a lot of documentation on this topic. The most complete I've found, at least for this particular SIMD work (the first one, I think) is [this RFC](https://github.com/rust-lang/rfcs/blob/master/text/1199-simd-infrastructure.md), which I've been linking almost everywhere in this PR. But even things like type-checking are loosely defined there, and most of what I've learn has been through trial and error.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: Done.

cprover_bindings/src/goto_program/typ.rs
adpaco-aws: Done.

cprover_bindings/src/goto_program/typ.rs
adpaco-aws: Right. I've added a case to return `Self` for that case, also in the symmetric method.

tests/expected/intrinsics/simd-invalid-conversion/expected
adpaco-aws: The problem here is that we have to do the type-checking... Compiling this with `rustc` would cause an error, but doing so with Kani doesn't. I've added cases for this in `vector_cmp` and a couple of test cases. But in there I don't have anything to exit gracefully as we do in other cases (e.g., the session object). I agree with you this should just be a compiler error, but in general this is what we do when type-checking (assertions).

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -953,6 +955,9 @@ impl Expr {
                     || (lhs.typ.is_pointer() && rhs.typ.is_integer())
             }
             ROk => lhs.typ.is_pointer() && rhs.typ.is_c_size_t(),
+            VectorEqual | VectorNotequal | VectorGe | VectorLe | VectorGt | VectorLt => {
+                lhs.typ == rhs.typ && lhs.typ.is_vector()
```
celinval, line 959: I think we should be more explicit about the conditions we expect here. All the conditions bellow must hold:
- `lhs.len() == rhs.len()`
- `lhs.typ.is_vector()`
- `lhs.typ.base_type().is_integer()`

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1361,6 +1361,22 @@ impl<'tcx> GotocCtx<'tcx> {
         )
     }
 
+    /// Generates code for a SIMD vector comparison intrinsic. This is similar
+    /// to `binop`, but we pass along `ret_ty` because return type for vector
+    /// comparison depends on the place type.
+    fn codegen_intrinsic_simd_cmp<F: FnOnce(Expr, Expr, Type) -> Expr>(
```
celinval, line 1367: I think we should type check the expression and print user friendly errors, like the `llvm` backend does. 
- Check that return type length matches the rhs operands
- Check that the return type base type is an integer.

https://github.com/rust-lang/rust/blob/c1a859b25a95999ba276075bbd8e284ea675b41a/compiler/rustc_codegen_llvm/src/intrinsic.rs#L925-L953

celinval (commented):
Thank you for all the explanations. I added a few comments based on my understanding. Please let me know if that makes sense.

adpaco-aws:
Ready for review again!

Typecheck errors are now triggered in `codegen_intrinsic_simd_cmp` and they are very similar to the ones produced by `rustc`. I've also included the (redundant) typechecks in a new associated function `typecheck_vector_cmp_expr`.

I've also undone the `VectorData` wrapper change. It isn't needed anymore because we're passing down the place type as the return type, which we use to generate the vector comparison expression. Previously, I wasn't doing this so I need to bypass the typecheck for assign statements somehow.

I've also moved/edited/added some comments.

cprover_bindings/src/goto_program/expr.rs
adpaco-aws: Done, but using this stricter condition:
```
        lhs.typ.is_vector()
            && lhs.typ == rhs.typ
            && lhs.typ.len() == ret_typ.len()
            && ret_typ.base_type().unwrap().is_integer()
```

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: Done. The errors look like this now:
```
error: expected return type with length 2 (same as input type `Vector { typ: Unsignedbv { width: 64 }, size: 2 }`), found `Vector { typ: Unsignedbv { width: 32 }, size: 4 }` with length 4
  --> main.rs:37:35
   |
37 |         let invalid_simd: u32x4 = simd_eq(x, y);
   |                                   ^^^^^^^^^^^^^

```

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -1013,6 +1036,16 @@ impl Expr {
         expr!(BinOp { op, lhs: self, rhs }, Expr::binop_return_type(op, &self, &rhs))
     }
 
+    /// Like `binop`, but receives an additional parameter `ret_typ` with the expected
+    /// return type for the place, which is used as the return type.
+    pub fn vector_cmp(self, op: BinaryOperator, rhs: Expr, ret_typ: Type) -> Expr {
+        assert!(
+            Expr::typecheck_vector_cmp_expr(&self, &rhs, &ret_typ),
+            "vector comparison expression does not typecheck",
```
celinval, line 1044: nit: Can you print `rhs` and `ret_typ` here? It makes it a huge difference when debugging a failure.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
celinval: Nice! Any chance you can print the rust types instead of the `goto` ones? I.e.: print `u64x2` instead of `Vector { typ: Unsignedbv { width: 64 }, size: 2 }`

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1361,6 +1373,69 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        f: F,
+        mut fargs: Vec<Expr>,
+        p: &Place<'tcx>,
+        span: Option<Span>,
+        ret_typ: Type,
+    ) -> Stmt {
+        let arg1 = fargs.remove(0);
+        let arg2 = fargs.remove(0);
+
+        // The return type must be the same length as the input types. The
```
celinval, line 1392: nit: I personally prefer expanding the method documentation and keeping the inline documentation small. But it's up to you.

My reasoning is:
1. It's clear in the method signature what it checks.
2. It makes it easier to read the code.
3. Separation of concerns. Method signature documentation talks about the what and inline code comments talk about the how.

celinval (commented):
Thanks Adrian! I only have one ask, can you please make the errors relative to the Rust types instead of the `goto` ones?

Everything else is nit, and I'm OK with keeping as they are if you prefer.

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -999,8 +994,36 @@ impl Expr {
...
+    /// Comparison operators for SIMD vectors aren't typechecked as regular
+    /// comparison operators. First, the return type depends on the place's type
+    /// (i.e., the variable or expression type for the result).
+    ///
+    /// In addition, the return type must have:
+    ///  1. The same length (number of elements) as the operand types.
+    ///  2. An integer base type (or just "boolean"-y, as mentioned in
+    ///     <https://github.com/rust-lang/rfcs/blob/master/text/1199-simd-infrastructure.md#comparisons>).
+    ///     The signedness doesn't matter, as the result for each element is
+    ///     either "all ones" (false) or "all zeros" (true).
```
tedinski, line 1015: Is that backwards?

cprover_bindings/src/goto_program/expr.rs
adpaco-aws: Done. Prints `lhs`, `rhs` and `ret_typ` on failure.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: Done!

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: Thanks for the explanation, it made a lot of sense to me. Moved these comments to the method documentation!

adpaco-aws:
BTW this is ready for review again, all comments above have been addressed.

tedinski (approved):

PR #1862, opened 2022-11-07
Integrate latest s2n-quic proofs in Kani CI

### LINKED ISSUE #1773: Integrate the current s2n-quic proofs in our regressions
Need to integrate commit https://github.com/aws/s2n-quic/commit/ae29a32660741b4926d557920fa852a313aa4f86 or later.

zhassan-aws:
### Description of changes: 

Update s2n-quic submodule to integrate the current Kani proofs (11) and run them in our CI. This should bring down the `perf` workflow from 1 hour to 15-20 minutes.

### Resolved issues:

Resolves #1773 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

The s2n-quic Bolero/Kani harnesses are under test code, so I had to add a new `cargo-kani-test` mode to `compiletest` that runs `cargo kani --tests`.

### Testing:

* How is this change tested?

### Checklist


tools/compiletest/src/common.rs
```diff
@@ -124,11 +127,6 @@ pub struct Config {
 
     /// Whether to rerun tests even if the inputs are unchanged.
     pub force_rerun: bool,
-
-    /// Allow us to run the regression with the mir linker enabled by default. For that, set
-    /// `RUSTFLAGS=--cfg=mir_linker` while compiling `compiletest`.
-    /// Remove this as part of <https://github.com/model-checking/kani/issues/1677>
-    pub mir_linker: bool,
```
celinval, line 131: Thanks!

celinval (changes requested):
Thanks for doing this! Although I think it's very important for us to run these on CI, I don't think we should keep that granularity of checks though. I.e.: I think the `expected` checks are highly dependent on the s2n-quic implementation.

I would like to suggest that we keep one expected, and ensure that we successfully run all their harnesses in a timely manner. At most, we can list the harnesses being checked (but even that might require changes if they rename a harness), which is not ideal.

zhassan-aws:
> I would like to suggest that we keep one expected, and ensure that we successfully run all their harnesses in a timely manner. At most, we can list the harnesses being checked (but even that might require changes if they rename a harness), which is not ideal.

Makes sense.

tools/compiletest/src/common.rs
```diff
@@ -18,6 +18,7 @@ pub enum Mode {
     Kani,
     KaniFixme,
     CargoKani,
+    CargoKaniTest, // cargo kani --tests
```
celinval, line 21: Can you please add a note that this mode is temporary?

celinval (approved):
It looks good. I hope we can get rid of this new mode soon though.

PR #1849, opened 2022-11-04
PropProof: Choice strategy support.

YoshikiTakashima:
### Description of changes: 

This PR will add support for strategies that choose between multiple other strategies.

### Resolved issues:

Works towards; #1608 

### Call-outs:

- Placing the test case in `union.rs` triggered a bug where assertion macros were no longer properly replaced with Kani assertions. To avoid this, the test case was put into `runner.rs`. ~~TODO:~~ tracking issue: #1856.

### Testing:

* How is this change tested? `proptest/src/test_runner/runner.rs`

* Is this a refactor change? No

### Checklist
- [x] Each commit message has a non-empty body, explaining why the change was made
- [x] Methods or procedures are documented
- [x] Regression or unit tests are included, or existing tests cover the modified code
- [x] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #1873, opened 2022-11-08
PropProof: Vector support

YoshikiTakashima:
### Description of changes: 

Adds vector support to PropProof. Models are optimized to reduce the number of loops.

### Resolved issues:

Works towards #1608 

### Call-outs:

- ~~Only 07f5a73f37b06d086c4191d21c109ff13e57ef53 and after is new. Commits before it is from #1849 and will be removed by rebase.~~
- Redundant code warnings are suppressed to keep the diff clean. There will be one last commit that cleans these up.

### Testing:

* How is this change tested? Test case at the bottom of `proptest/src/collection.rs`

### Checklist

### DIFF
--- a/proptest/src/bits.rs
+++ b/proptest/src/bits.rs
@@ -451,6 +451,7 @@ pub(crate) mod varsize {
     pub(crate) struct VarBitSet(Inner);
 
     impl VarBitSet {
+        #[allow(dead_code)] // todo: cleanup redundant code.
         pub(crate) fn saturated(len: usize) -> Self {
             (0..len).collect::<VarBitSet>()
         }
--- a/proptest/src/bits.rs
+++ b/proptest/src/bits.rs
@@ -451,7 +451,7 @@ pub(crate) mod varsize {
     pub(crate) struct VarBitSet(Inner);
 
     impl VarBitSet {
-        #[allow(dead_code)] // todo: cleanup redundant code.
+        #[allow(dead_code)] // todo: cleanup redundant code. See issue #2004.
         pub(crate) fn saturated(len: usize) -> Self {
             (0..len).collect::<VarBitSet>()
         }
--- a/proptest/src/collection.rs
+++ b/proptest/src/collection.rs
@@ -534,7 +534,7 @@ where
     ))
 }
 
-#[allow(dead_code)] // todo: cleanup redundant code.
+#[allow(dead_code)] // todo: cleanup redundant code. See issue #2004.
 #[derive(Clone, Copy, Debug)]
 enum Shrink {
     DeleteElement(usize),
@@ -560,7 +560,7 @@ impl<T: Strategy> Strategy for VecStrategy<T> {
     type Value = Vec<T::Value>;
 
     /// Builds a new ValueTree<Vec<_>> by repeatedly generating values
-    /// from strategy stored in `self.element`. Kani Optimization: loop
+    /// from strategy stored in self.element. Kani Optimization: loop
     /// only once by storing the resulting vector in RefCell
     fn new_tree(&self, runner: &mut TestRunner) -> NewTree<Self> {
         let length: usize = kani::any();
@@ -599,14 +599,29 @@ impl<T: Strategy> Strategy for Vec<T> {
 impl<T: ValueTree> ValueTree for VecValueTree<T> {
     type Value = Vec<T::Value>;
 
+    /// The current value is extracted out of the RC-RefCell by
+    /// replacing it with an empty vector.
+    ///
+    /// Kani Optimization:This avoids having to copy
+    /// the vector and improves verification performance. Note that
+    /// this will clobber the current value, but that is fine because
+    /// Kani will run this function at most once.
     fn current(&self) -> Vec<T::Value> {
         self.current.as_ref().replace(vec![])
     }
 
+    /// Simplifies the current value.
+    ///
+    /// Kani Optimization: This is not necessary for Kani, and getting
+    /// rid of this loop improves verification performance.
     fn simplify(&mut self) -> bool {
         false
     }
 
+    /// Complicates the current value.
+    ///
+    /// Kani Optimization: This is not necessary for Kani, and getting
+    /// rid of this loop improves verification performance.
     fn complicate(&mut self) -> bool {
         false
     }
--- a/proptest/src/collection.rs
+++ b/proptest/src/collection.rs
@@ -20,7 +20,7 @@ use core::ops::{Add, Range, RangeInclusive, RangeTo, RangeToInclusive};
 use core::usize;
 
 use crate::std_facade::{
-    fmt, BTreeMap, BTreeSet, BinaryHeap, LinkedList, Vec, VecDeque,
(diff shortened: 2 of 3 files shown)


adpaco-aws:
@YoshikiTakashima is this ready for review?

YoshikiTakashima:
@adpaco-aws Yes. Please review.

YoshikiTakashima:
@danielsn @adpaco-aws Do you have an estimate on when you will have time to review?

YoshikiTakashima:
@adpaco-aws Could you get this?

proptest/src/bits.rs
```diff
@@ -451,6 +451,7 @@ pub(crate) mod varsize {
     pub(crate) struct VarBitSet(Inner);
 
     impl VarBitSet {
+        #[allow(dead_code)] // todo: cleanup redundant code.
```
adpaco-aws, line 454: Maybe add a tracking issue for clean-up?

proptest/src/collection.rs
```diff
@@ -534,40 +534,45 @@ where
     ))
 }
 
+#[allow(dead_code)] // todo: cleanup redundant code.
```
adpaco-aws, line 537: Same.

proptest/src/collection.rs
```diff
@@ -534,40 +534,45 @@ where
...
+        }
+    }
 }
 
 impl<T: Strategy> Strategy for VecStrategy<T> {
     type Tree = VecValueTree<T::Tree>;
     type Value = Vec<T::Value>;
 
+    /// Builds a new ValueTree<Vec<_>> by repeatedly generating values
+    /// from strategy stored in self.element. Kani Optimization: loop
```
adpaco-aws, line 563:
```suggestion
    /// from strategy stored in `self.element`. Kani Optimization: loop
```

proptest/src/collection.rs
```diff
@@ -597,90 +600,40 @@ impl<T: ValueTree> ValueTree for VecValueTree<T> {
     type Value = Vec<T::Value>;
 
     fn current(&self) -> Vec<T::Value> {
-        self.elements
-            .iter()
-            .enumerate()
-            .filter(|&(ix, _)| self.included_elements.test(ix))
-            .map(|(_, element)| element.current())
-            .collect()
+        self.current.as_ref().replace(vec![])
```
adpaco-aws, line 603: I don't understand what this is doing. Can you please add a comment?

proptest/src/collection.rs
```diff
@@ -597,90 +600,40 @@ impl<T: ValueTree> ValueTree for VecValueTree<T> {
...
-                // Move on to the next element
-                self.shrink = Shrink::ShrinkElement(ix + 1);
-            } else {
-                self.prev_shrink = Some(self.shrink);
-                return true;
-            }
-        }
-
-        panic!("Unexpected shrink state");
+        false
```
adpaco-aws, line 607: I understand why `simplify` (or `complicate`) isn't possible for Kani counter-examples, but can you add a short comment?

proptest/src/collection.rs
```diff
@@ -597,90 +600,40 @@ impl<T: ValueTree> ValueTree for VecValueTree<T> {
...
-#[cfg(all(test, not(kani)))]
+#[cfg(kani)]
+mod kani_tests {
+    use super::*;
+
+    crate::proptest! {
+    #[kani::unwind(5)]
+    #[kani::proof]
+    fn vector_even_sums(
+        vec_even in vec((0..10).prop_map(|x: i32| x << 1), 0..2),
```
adpaco-aws, line 627: `x * 2` would be clearer here (and below), but I'm not sure if you need `<<` for performance reasons.

proptest/src/collection.rs
```diff
@@ -597,90 +600,40 @@ impl<T: ValueTree> ValueTree for VecValueTree<T> {
...
+    ) {
+            let sum: i32 = vec_even.into_iter().sum();
+            assert!(sum  < 40, "each element is < 20, at most 2 elements");
```
adpaco-aws, line 630: Is this true? If you draw 10, you'll get 20 (which isn't `< 20`).

proptest/src/collection.rs
```diff
@@ -597,90 +600,40 @@ impl<T: ValueTree> ValueTree for VecValueTree<T> {
...
+            assert_eq!(sum % 2, 0, "Sum is even due to << 1.");
+    }
```
adpaco-aws, line 632: There are some formatting issues in this piece of code (e.g., `}` on the same line, space in `sum  < 40`). Can you fix those?

proptest/src/collection.rs
YoshikiTakashima: I don't see any difference. I will change to that.

proptest/src/collection.rs
YoshikiTakashima: Will do! TL;DR is that this allows us to output the the vector without copying it out. That's better for performance because original implementation and copying both need to iterate the vector twice.

proptest/src/collection.rs
YoshikiTakashima: I don't think it is possible to get 10. `0..10` is not inclusive of the upper bound 10, so the max will be 9. Multiply that by 2, and max is 18.

proptest/src/collection.rs
YoshikiTakashima: RC enables interior mut even though we only have `&self`, and RefCell has a replace API we can use to replace the contents and get back the current content. Combine both to get zero-copy values out of a static pointer.

proptest/src/bits.rs
YoshikiTakashima: #2004

proptest/src/collection.rs
YoshikiTakashima: #2004

proptest/src/bits.rs
YoshikiTakashima: 32ce7073475692638acfc594ebb94aa94d82c1d5

proptest/src/collection.rs
YoshikiTakashima: 32ce7073475692638acfc594ebb94aa94d82c1d5

proptest/src/collection.rs
YoshikiTakashima: 32ce7073475692638acfc594ebb94aa94d82c1d5

proptest/src/collection.rs
YoshikiTakashima: 32ce7073475692638acfc594ebb94aa94d82c1d5

proptest/src/collection.rs
YoshikiTakashima: c5b8571ba0e8c9f98106c1a281c8bd0c076f9975

proptest/src/collection.rs
YoshikiTakashima: c5b8571ba0e8c9f98106c1a281c8bd0c076f9975

YoshikiTakashima:
@adpaco-aws Fixed comments. Please check again and merge when you think it's ready.

adpaco-aws (approved):

Issue #1866, opened 2022-11-08
Resolve names during stubbing


aaronbembenek-aws:
This issue affects how we treat the function/method names supplied in `kani::stub(<original>, <replacement>)` attributes. We currently treat `<original>` and `<replacement>` verbatim as strings, so that they have to exactly match a name we find in the HIR. This means that we are unable to find many functions/methods in foreign crates, that we do not support relative paths, and that we are unable to take into account `use ... as ...` statements.

We should instead correctly resolve names during stubbing, as well as support path constructs like `self` and `super`.

celinval:
Would it be safe to assume that by correctly resolving names you would also be properly handling multiple versions of a crate? It might be nice to clarify that in the RFC.

aaronbembenek-aws:
Yes, we should be able to handle that.

celinval:
Just to double check, in the case of multiple versions, one stub would only be applied to one specific version right?

aaronbembenek-aws:
That would be the optimal behavior, but it's going to take some experimenting with our path resolution algorithm to make sure we get this right. Luckily, it looks like clippy is in the process of updating its path resolution algorithm to handle this correctly (https://github.com/rust-lang/rust-clippy/pull/9800), so we can probably draw from there.

aaronbembenek-aws:
It might actually be harder to handle multiple versions of the same crate than I thought, if they are included via the [renaming-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html?highlight=rename,depende#renaming-dependencies-in-cargotoml) feature of Cargo. As far as I can tell, the data we have available at the MIR level refers to the original name of the crate, and not what it has been renamed to...

celinval:
Closing this issue in favor of more specific ones.

Note that we have already implemented support for relative paths, including usage of `self` and `super`. I believe we are missing qualified paths needed for trait methods, and primitive type methods.

PR #1874, opened 2022-11-09
Implement basic function/method stubbing.

### LINKED ISSUE #1871: Implement basic function/method stubbing mechanism
Kani currently has support for the `kani::stub` attribute and the `--enable-stubbing` argument; however, it does not currently actually do any stubbing. This issue calls for implementing a basic stubbing mechanism for functions and methods (a step in #1809). It is out-of-scope to resolve function/method names using Rust-style path resolution, which deserves its own issue (see #1866).

aaronbembenek-aws:
### Description of changes: 

Kani currently has support for the `kani::stub` attribute and the `--enable-stubbing` argument; however, it does not actually do any stubbing. This PR implements a basic stubbing mechanism for functions and methods (a step in #1809). It is out-of-scope to resolve function/method names using Rust-style path resolution, which deserves its own PR (see #1866).

Stubbing happens as a MIR-to-MIR transformation in `kani-compiler`. We run rustc twice: first to figure out which stubs are requested for each harness, and a second time to actually do the stubbing and generate code. We might collapse these two steps into a single rustc call in the future, depending on Kani's approach to doing per-harness code generation.

### Resolved issues:

Resolves #1871 

### Testing:

### Checklist


kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -0,0 +1,59 @@
...
+        if let Some(captures) = RE.captures(arg) {
+            return Some(deserialize_mapping(captures.get(1).unwrap().as_str()));
+        }
+    }
+    None
+}
+
+/// Tries to find the `DefId` of a function/method that matches the path `path`.
+fn get_def_id(tcx: TyCtxt, path: &str) -> Option<DefId> {
+    tcx.iter_local_def_id().map(LocalDefId::to_def_id).find(|&id| tcx.def_path_str(id) == path)
```
celinval, line 58: Is there any other way to find this? This looks rather expensive.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -45,6 +45,32 @@ pub fn extract_integer_argument(attr: &Attribute) -> Option<u128> {
     }
 }
 
+/// Extracts a vector of path arguments from an attribute.
+/// Returns `None` if any argument is not syntactically a path.
+/// Paths are returned as strings.
+///
+/// For example, on `stub(foo::bar, baz)`, this returns `Some(vec!["foo::bar", "baz"])`.
+pub fn extract_path_arguments(attr: &Attribute) -> Option<Vec<String>> {
```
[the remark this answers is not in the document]
aaronbembenek-aws: I feel like that approach would lose information, as the caller would not know whether there are no arguments to the attribute, or whether there are arguments but one of them is not a path. Maybe it'd be cleaner to return a vector of `Option<String>`, where each entry in the vector corresponds to an argument of the attribute, and an entry is `Some` if it's a path.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -45,6 +45,32 @@ pub fn extract_integer_argument(attr: &Attribute) -> Option<u128> {
...
+            let path = meta_item
+                .path
+                .segments
+                .iter()
+                .map(|seg| seg.ident.as_str())
+                .collect::<Vec<&str>>()
+                .join("::");
+            paths.push(path);
+        } else {
+            return None;
```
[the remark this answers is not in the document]
aaronbembenek-aws: This happens if that argument to the attribute is not a path (for example, if it's an integer). Whether it is an error or not would probably depend on the attribute that is being processed. This would be cleaner with my suggestion above.

kani-compiler/src/kani_middle/stubbing/annotations.rs
```diff
@@ -0,0 +1,114 @@
...
+            Compilation::Stop
+        })
+    }
+}
+
+impl CollectorCallbacks {
+    /// Given a `kani::stub` attribute, tries to extract a pair of paths (the
+    /// original function/method, and its stub). Returns `None` and errors if
+    /// the attribute's arguments are not two paths.
+    fn extract_stubbing_pair(tcx: TyCtxt, attr: &Attribute) -> Option<(String, String)> {
```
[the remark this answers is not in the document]
aaronbembenek-aws: Fair enough - we could move it outside the struct.

kani-compiler/src/kani_middle/stubbing/annotations.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/// A rustc callback that is used to collect the stub mappings specified for
+/// each harness.
+#[derive(Default)]
+struct CollectorCallbacks {
+    stub_mapping: FxHashMap<String, FxHashMap<String, String>>,
+}
+
+impl Callbacks for CollectorCallbacks {
```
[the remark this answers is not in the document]
aaronbembenek-aws: Could you expand on this comment? The callback is used for the local call to rustc.

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -0,0 +1,59 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
```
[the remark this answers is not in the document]
aaronbembenek-aws: Will do!

kani-compiler/src/main.rs
```diff
@@ -104,10 +107,19 @@ fn main() -> Result<(), &'static str> {
...
 
     // Generate rustc args.
-    let rustc_args = generate_rustc_args(&matches);
+    let mut rustc_args = generate_rustc_args(&matches);
 
-    if matches.is_present(parser::ENABLE_STUBBING) {
-        eprintln!("warning: Kani currently does not perform any stubbing.");
+    // If appropriate, collect and set the stub mapping.
+    if matches.is_present(parser::ENABLE_STUBBING)
+        && queries.get_reachability_analysis() == ReachabilityType::Harnesses
```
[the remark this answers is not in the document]
aaronbembenek-aws: Sounds good!

kani-compiler/src/main.rs
```diff
@@ -104,10 +107,19 @@ fn main() -> Result<(), &'static str> {
...
+    {
+        queries.set_stubbing_enabled(true);
+        let all_stub_mappings =
+            stubbing::collect_stub_mappings(&rustc_args).or(Err("Failed to compile crate"))?;
+        let harness = matches.value_of(parser::HARNESS).unwrap();
+        let mapping = find_harness_stub_mapping(harness, all_stub_mappings)
```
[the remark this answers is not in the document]
aaronbembenek-aws: No, in that case the harness will have an empty stub mapping.

kani-compiler/src/kani_middle/stubbing/annotations.rs
```diff
@@ -0,0 +1,114 @@
...
+use rustc_driver::RunCompiler;
+use rustc_driver::{Callbacks, Compilation};
+use rustc_errors::ErrorGuaranteed;
+use rustc_interface::interface::Compiler;
+use rustc_interface::Queries;
+use rustc_middle::ty::TyCtxt;
+
+use crate::kani_middle::attributes::{extract_path_arguments, partition_kanitool_attributes};
+
+/// Collects the stubs from the harnesses in a crate, running rustc with the
```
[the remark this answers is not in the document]
aaronbembenek-aws: Will do!

kani-compiler/src/kani_middle/stubbing/transform.rs
aaronbembenek-aws: I'll look into this.

aaronbembenek-aws:
> 1. I have some mixed feelings about running the compiler from the stubbing module. I like that you stop after analysis though.

I'm open to moving it, or even combining it with the other call to rustc (so that we make only one rustc call - although in this case we'd have to think more about how the stub mappings get propagated from that pass to the actual transformation pass).

kani-compiler/src/kani_middle/attributes.rs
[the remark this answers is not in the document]
celinval: Yes, that would be much cleaner.

aaronbembenek-aws:
> 2\. Where do you validate that the stub types match? What happens if the user tries to stub something with a method that doesn't match its signature?

We don't do any validation currently; that case ends up being a type mismatch in GotoC. We could do more validation now; for example, to make sure that there are at least the same number of arguments. I think we'll need to think more about validation and address that carefully, as it'll be tricky to get right (unfortunately not as simple as seeing whether the two functions have the same types).

kani-compiler/src/main.rs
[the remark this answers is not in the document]
aaronbembenek-aws: Okay, sure. Kani will catch it later on if it can't find the harness, anyways.

celinval:
> > 2. Where do you validate that the stub types match? What happens if the user tries to stub something with a method that doesn't match its signature?
> 
> We don't do any validation currently; that case ends up being a type mismatch in GotoC. We could do more validation now; for example, to make sure that there are at least the same number of arguments. I think we'll need to think more about validation and address that carefully, as it'll be tricky to get right (unfortunately not as simple as seeing whether the two functions have the same types).

Sorry, why can't you compare the types of each argument and the return type? I really think we need to add some checks here, and not let that become a type mismatch in the GotoC.

aaronbembenek-aws:
> > > 2. Where do you validate that the stub types match? What happens if the user tries to stub something with a method that doesn't match its signature?
> > 
> > 
> > We don't do any validation currently; that case ends up being a type mismatch in GotoC. We could do more validation now; for example, to make sure that there are at least the same number of arguments. I think we'll need to think more about validation and address that carefully, as it'll be tricky to get right (unfortunately not as simple as seeing whether the two functions have the same types).
> 
> Sorry, why can't you compare the types of each argument and the return type? I really think we need to add some checks here, and not let that become a type mismatch in the GotoC.

It won't work if there are generic types involved, where (I think) we really want to see if the types match up after monomorphization. Another alternative would be to make sure that the type of the stub is a subtype of the function type, although this is more restrictive than looking at things post-monomorphization.

One option for now: we make sure that non-generic types match up, and if there are any generics, we issue a warning that a type mismatch might manifest in GotoC.

celinval:
> > > > 2. Where do you validate that the stub types match? What happens if the user tries to stub something with a method that doesn't match its signature?
> > > 
> > > 
> > > We don't do any validation currently; that case ends up being a type mismatch in GotoC. We could do more validation now; for example, to make sure that there are at least the same number of arguments. I think we'll need to think more about validation and address that carefully, as it'll be tricky to get right (unfortunately not as simple as seeing whether the two functions have the same types).
> > 
> > 
> > Sorry, why can't you compare the types of each argument and the return type? I really think we need to add some checks here, and not let that become a type mismatch in the GotoC.
> 
> It won't work if there are generic types involved, where (I think) we really want to see if the types match up after monomorphization. Another alternative would be to make sure that the type of the stub is a subtype of the function type, although this is more restrictive than looking at things post-monomorphization.
> 
> One option for now: we make sure that non-generic types match up, and if there are any generics, we issue a warning that a type mismatch might manifest in GotoC.

Are you thinking about allowing stubs of generic functions using a non-generic function?

aaronbembenek-aws:
> > > > > 2. Where do you validate that the stub types match? What happens if the user tries to stub something with a method that doesn't match its signature?
> > > > 
> > > > 
> > > > We don't do any validation currently; that case ends up being a type mismatch in GotoC. We could do more validation now; for example, to make sure that there are at least the same number of arguments. I think we'll need to think more about validation and address that carefully, as it'll be tricky to get right (unfortunately not as simple as seeing whether the two functions have the same types).
> > > 
> > > 
> > > Sorry, why can't you compare the types of each argument and the return type? I really think we need to add some checks here, and not let that become a type mismatch in the GotoC.
> > 
> > 
> > It won't work if there are generic types involved, where (I think) we really want to see if the types match up after monomorphization. Another alternative would be to make sure that the type of the stub is a subtype of the function type, although this is more restrictive than looking at things post-monomorphization.
> > One option for now: we make sure that non-generic types match up, and if there are any generics, we issue a warning that a type mismatch might manifest in GotoC.
> 
> Are you thinking about allowing stubs of generic functions using a non-generic function?

Yes, I was thinking that we should allow that. For example, a user might want to replace `rand::random` (which returns `T where Standard: Distribution<T>`) with a function that returns `u32`, if that's appropriate for the proof harness.

In the case of `rand::random`, it would be hard for a user to write a stub with the correct return type: there would be a type mismatch even if the function returned `kani::any()`.

kani-compiler/src/main.rs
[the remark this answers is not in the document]
aaronbembenek-aws: Actually, I'm not sure if we want to require `ReachabilityType::Harnesses` for `--enable-stubbing`. This will fail on a test I have involving `cargo kani` running on multiple crates (one of which is running with `ReachabilityType::None`).

kani-compiler/src/kani_middle/stubbing/transform.rs
aaronbembenek-aws: Added a TODO linked to #1894

kani-compiler/src/main.rs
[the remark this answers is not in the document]
celinval: Oh.. where is the code for calling this via cargo? Are you adding `--enable-stubbing` as part of `pkg_args`?

kani-compiler/src/kani_middle/stubbing/annotations.rs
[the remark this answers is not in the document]
celinval: I guess my concern is that this call is kind of hidden and it may catch developers by surprise that rustc is being invoked from this submodule.

kani-compiler/src/main.rs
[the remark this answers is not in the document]
aaronbembenek-aws: I don't think I've done anything to `pkg_args`. The test that invokes everything using cargo is `tests/cargo-kani/foreign-function-stubbing/`. The arguments (including `--enable-stubbing`) are set in the `Cargo.toml` there.

kani-compiler/src/kani_middle/stubbing/annotations.rs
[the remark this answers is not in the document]
aaronbembenek-aws: Fair. Some options:

1. Keep this preliminary call to rustc, but move the call itself into `main`, so that it is more visible.
2. Remove this call to rustc, and run the annotation collecting callback in the same rustc call that does code gen. I originally went with two calls to rustc because it'd allow us to figure out which harnesses use the same stub set, and then we'd able to code gen those ones together. However, if we are going to be doing per-harness code gen, figuring out the stub sets first doesn't really get us anything.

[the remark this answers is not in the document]
celinval: I think it's OK for this PR.

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Deserializes the stub mapping from the rustc argument value.
+fn deserialize_mapping(val: &str) -> FxHashMap<String, String> {
+    serde_json::from_str(val).unwrap()
+}
+
+/// Retrieves the stub mapping from the compiler configuration.
+fn get_stub_mapping(tcx: TyCtxt) -> Option<FxHashMap<String, String>> {
+    lazy_static! {
+        static ref RE: Regex = Regex::new(&format!("'{RUSTC_ARG_PREFIX}(.*)'")).unwrap();
+    }
```
adpaco-aws, line 52: Can't this regex be a local variable for some reason?

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Serializes the stub mapping into a rustc argument.
+pub fn mk_rustc_arg(stub_mapping: FxHashMap<String, String>) -> String {
+    // Store our serialized mapping as a fake LLVM argument (safe to do since
+    // LLVM will never see them).
+    format!("-Cllvm-args='{RUSTC_ARG_PREFIX}{}'", serde_json::to_string(&stub_mapping).unwrap())
+}
+
+/// Deserializes the stub mapping from the rustc argument value.
+fn deserialize_mapping(val: &str) -> FxHashMap<String, String> {
+    serde_json::from_str(val).unwrap()
```
adpaco-aws, line 45: Would it make sense to serialize/deserialize these arguments into structs?

kani-compiler/src/kani_middle/stubbing/annotations.rs
```diff
@@ -0,0 +1,105 @@
...
+    // <https://github.com/model-checking/kani/issues/1866>
+    Some((args[0].as_ref().unwrap().clone(), args[1].as_ref().unwrap().clone()))
+}
+
+/// Updates the running map `stub_pairs` that maps a function/method to its
+/// stub. Errors if a function/method is mapped more than once.
+fn update_stub_mapping(tcx: TyCtxt, attr: &Attribute, stub_pairs: &mut FxHashMap<String, String>) {
+    if let Some((original, replacement)) = extract_stubbing_pair(tcx, attr) {
+        let other = stub_pairs.insert(original.clone(), replacement.clone());
+        if let Some(other) = other {
```
adpaco-aws, line 95: Nit: I like to avoid `if let Some(x) = x` patterns (for clarity) by renaming the option type (e.g., `other_opt`)

kani-compiler/src/kani_middle/stubbing/annotations.rs
```diff
@@ -0,0 +1,105 @@
...
+                let mut stub_pairs = FxHashMap::default();
+                for (name, attr) in other {
+                    if name == "stub" {
+                        update_stub_mapping(tcx, attr, &mut stub_pairs);
+                    }
+                }
+                let harness_name = tcx.def_path_str(def_id);
+                self.stub_mapping.insert(harness_name, stub_pairs);
+            }
+            tcx.sess.abort_if_errors();
```
adpaco-aws, line 57: Was expecting this to be in the loop, I'm not sure you want to continue it if errors get spanned for any proof.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -45,6 +45,44 @@ pub fn extract_integer_argument(attr: &Attribute) -> Option<u128> {
     }
 }
 
+/// Extracts a vector with the path arguments of an attribute.
+/// The length of the returned vector is equal to the number of arguments in the
+/// attribute; an entry is `None` if the argument is not syntactically a path,
+/// and `Some(<path>)` otherwise. Paths are returned as strings.
+///
+/// For example, on `stub(foo::bar, 42, baz)`, this returns
+/// `vec![Some("foo::bar"), None, Some("baz")]`.
```
adpaco-aws, line 54: Nice comment! 💯

kani-compiler/src/main.rs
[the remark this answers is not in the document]
celinval: Sorry if I wasn't clear. The `--enable-stubbing` in that file will be passed to `kani-driver` only. I'm trying to understand how this is betting passed down to `kani-compiler`. There are basically two ways today to pass information to the compiler, one as part of `KANI_FLAGS` and a second one via `cargo rustc -- <extra_flags>`. I can provide more context later if you want, or you can find more details in the MIR Linker RFC.

That said, I think passing the `--enable-stubbing` as part of `pkg_args` will probably be more painful today, so let's just leave it as is for now.

kani-compiler/src/kani_middle/stubbing/annotations.rs
celinval: The nice thing about leaving this after the loop is that you print all errors at once. So users don't find a new error after they fix the first one.

kani-compiler/src/kani_middle/stubbing/annotations.rs
celinval: +1 on avoiding shadowing a variable.

kani-compiler/src/kani_middle/stubbing/transform.rs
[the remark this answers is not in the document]
celinval: Can you expand it a tiny bit on how it works?

tests/cargo-kani/foreign-function-stubbing/src/main.rs
```diff
@@ -0,0 +1,33 @@
...
+//
+//! This tests stubbing for foreign functions and methods.
+
+use other_crate;
+
+fn the_answer() -> u32 {
+    42
+}
+
+// TODO: Split up these assertions into separate harnesses, once stubbing is able to support that.
```
celinval, line 12: What about at least splitting this test into multiple ones? I think it will make it much more readable.

tests/expected/function-stubbing-error/expected
```diff
@@ -0,0 +1,4 @@
+Attribute `kani::stub` takes two path arguments; found 3
```
celinval, line 1: Can you please move this test to UI?

tests/kani/FunctionStubbing/main.rs
```diff
@@ -0,0 +1,75 @@
...
+    fn priv_fn(&self) -> u32 {
+        0
+    }
+
+    fn the_answer(&self) -> u32 {
+        42
+    }
+}
+
+// TODO: Split up these assertions into separate harnesses, once stubbing is able to support that.
```
celinval, line 58: Can you break this test into multiple ones? May local functions, private functions and methods?

celinval (approved):
I OK with merging this and work on further improvements in separate PRs. Thanks

kani-compiler/src/kani_middle/stubbing/transform.rs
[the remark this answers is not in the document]
aaronbembenek-aws: Done!

kani-compiler/src/main.rs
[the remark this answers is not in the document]
aaronbembenek-aws: Oh okay, sorry for the confusion: `--enable-stubbing` is passed through `KANI_FLAGS`.

kani-compiler/src/kani_middle/stubbing/transform.rs
aaronbembenek-aws: Definitely - this is just an optimization so that it the regex is compiled only once. I'll add a comment.

kani-compiler/src/kani_middle/stubbing/transform.rs
aaronbembenek-aws: Apparently the best way to pass information to a custom back end is through the rustc arguments themselves. We could try to deserialize the arguments just once (which would be nicer), but it's not clear where in the compiler we could actually store the result. In the past I've used globals for this, but this isn't best practice as rustc queries are supposed to be stateless (and this transformation becomes part of the `optimized_mir` query).

kani-compiler/src/kani_middle/stubbing/annotations.rs
aaronbembenek-aws: Fixed!

kani-compiler/src/kani_middle/stubbing/annotations.rs
aaronbembenek-aws: Yes, that's my justification for it. I think it would be annoying for the user to have to see errors one at a time.

adpaco-aws: That makes sense, thanks!

tests/kani/FunctionStubbing/main.rs
aaronbembenek-aws: Done!

tests/expected/function-stubbing-error/expected
aaronbembenek-aws: Done!

tests/cargo-kani/foreign-function-stubbing/src/main.rs
aaronbembenek-aws: Done - split into public functions, private functions, and methods.

adpaco-aws (approved):

PR #1886, opened 2022-11-10
Treat lack of proof harnesses as success

### LINKED ISSUE #1857: Allow kani to return EXIT_SUCCESS for empty proof harnesses
**Feature:** Treat empty harnesses as successes rather than failures, either by default or through a flag. This is helpful to make generic CI checkers that automatically pick up proofs when they exist for a project and ignore them when they don't. As an example, consider that `cargo test` does not fail on an empty harness.

**Current behaviour:** `kani` unconditionally errors out if there's no proof harness.

```
alexgv@mypc:~$ cargo new mwe
     Created binary (application) `mwe` package
alexgv@mypc:~$ cd mwe && cargo kani || echo NONZERO_EXIT
   Compiling mwe v0.1.0 (/home/alexgv/mwe)
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
Error: No proof harnesses (functions with #[kani::proof]) were found to verify.
NONZERO_EXIT
```

tedinski:
### Description of changes: 

Do not error out on lack of proof harnesses.

### Resolved issues:

* Resolves #1857

### Call-outs:

Our output is now consistent with `cargo test` on 0 harnesses: give the same summary information, but no tests/proofs were found. I think we could do better, but maybe now's not the time.

### Testing:

### Checklist


zhassan-aws (commented):
Thanks! Can we add a test?

celinval (commented):
Thank you for doing this. Can we still add a message though to say that no proof harness was found?

tedinski:
It seemed wasteful to write one test for this, much less the two tests needed to both check expected output and return value, but I guess I'm out voted

celinval:
Yeah, two tests do feel a bit overwhelming. Too bad this can't be done in one single test.

tedinski:
Compromise? I made this slightly more refactor-like by just moving the existing code to a new place as `println` instead of `bail`. And added the minimal test for exist status.

celinval (commented):
Can you please make sure this test actually runs? I remember having problems with `cargo kani` tests that don't have an `expected` file.

tedinski:
Oh yeah, definitely not working. Well this is disconcerting, there's actually NO way to write a test to ensure `cargo kani` succeeds.

zhassan-aws:
I think adding an `expected` file, even if empty, would guarantee that `cargo kani` runs.

tedinski:
It runs, but nothing pays attentions to the exit status

celinval (approved):

zhassan-aws (approved):

PR #1893, opened 2022-11-14
Restore support for bitwise SIMD intrinsics

adpaco-aws:
### Description of changes: 

Restores the support for bitwise SIMD intrinsics:
 * `simd_and`
 * `simd_or`
 * `simd_xor`

Also adds a new test for these specific intrinsics.

### Resolved issues:

Towards #1148 

### Testing:

* How is this change tested? Existing regression and 1 new test.

### Checklist

### DIFF
--- a/tests/kani/Intrinsics/SIMD/Operators/bitwise.rs
+++ b/tests/kani/Intrinsics/SIMD/Operators/bitwise.rs
@@ -13,43 +13,92 @@
 #[repr(simd)]
 #[allow(non_camel_case_types)]
 #[derive(Clone, Copy, PartialEq, Eq)]
+pub struct i8x2(i8, i8);
+
+#[repr(simd)]
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct i16x2(i16, i16);
+
+#[repr(simd)]
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct i32x2(i32, i32);
+
+#[repr(simd)]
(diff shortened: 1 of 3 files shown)


tests/kani/Intrinsics/SIMD/Operators/bitwise.rs
```diff
@@ -0,0 +1,55 @@
...
+//!  * `simd_xor`
+//! This is done by initializing vectors with the contents of 2-member tuples
+//! with symbolic values. The result of using each of the intrinsics is compared
+//! against the result of using the associated bitwise operator on the tuples.
+#![feature(repr_simd, platform_intrinsics)]
+
+#[repr(simd)]
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct i64x2(i64, i64);
```
danielsn, line 16: Should we check this for different bitwidths?

tedinski (approved):
I have the same question as Daniel, but I don't think merging should be blocked over it.

Were these delayed by waiting for changes on the cbmc side? Or were these the easy ones?

adpaco-aws:
Thanks all for you comments.

> Should we check this for different bitwidths?

The test now checks for this instead. My main concern was performance, but it seems to run pretty fast regardless.

> Were these delayed by waiting for changes on the cbmc side? Or were these the easy ones?

These are the easy ones: They don't need changes in CBMC nor Kani. We don't need to add checks for these operations, but that's not the case for arithmetic operators (which will be coming soon).

tedinski (approved):

danielsn (approved):

PR #1903, opened 2022-11-15
Encode string literals as a byte array

### LINKED ISSUE #1665: Kani encodes string slices as null-terminated C slices
In Kani's regression test suite, in [offset_u8_ok.rs](https://github.com/model-checking/kani/blob/main/tests/kani/Intrinsics/Offset/offset_u8_ok.rs#L21), an out-of-bound pointer is dereferenced, but not detected by Kani. Here's a reduced code
```rust
#![feature(core_intrinsics)]
use std::intrinsics::offset;

#[kani::proof]
fn test_offset() {
    let s: &str = "123";
    let ptr: *const u8 = s.as_ptr();
    unsafe {
        let x = *offset(ptr, 3);
    }
}
```

The pointer is 1-passed the bound of the object, so it is "weak_valid", in the sense defined by [CompCert](https://compcert.org/doc/html/compcert.common.Values.html#Val.COMPARISONS.weak_valid_ptr), meaning that it is either in-bound or one-passed-the bound of the object. According to C standard, that is not undefined behaviour, and shouldn't be detected as such. However, it seems that, in addition to check that the pointer can be created, this test also checks that it can be dereferenced which is always UB.

@danielsn mentioned it might be on purpose because the stdlib does that sometimes, but I haven't been able to find an example

zhassan-aws:
### Description of changes: 

Kani currently encodes string slices (`str`) as null-terminated C slices, which is unsound. This PR changes the encoding to a byte array to guarantee that Kani reports pointer accesses outside the slice as invalid dereferences.

### Resolved issues:

Resolves #1665 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? A new test was added, and existing tests were slightly modified

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
@@ -4,7 +4,7 @@ use crate::codegen_cprover_gotoc::utils::slice_fat_ptr;
 use crate::codegen_cprover_gotoc::GotocCtx;
 use crate::unwrap_or_return_codegen_unimplemented;
 use cbmc::btree_string_map;
-use cbmc::goto_program::{DatatypeComponent, Expr, Location, Stmt, Symbol, Type};
+use cbmc::goto_program::{DatatypeComponent, Expr, ExprValue, Location, Stmt, Symbol, Type};
 use rustc_ast::ast::Mutability;
 use rustc_middle::mir::interpret::{
     read_target_uint, AllocId, Allocation, ConstValue, GlobalAlloc, Scalar,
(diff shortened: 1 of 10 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -98,18 +98,45 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    assert_eq!(start, 0);
+                    // Create a static variable that holds its value
+                    let mem_var =
+                        self.codegen_allocation_auto_imm_name(data, |tcx| tcx.next_global_name());
+
+                    // Extract identifier for static variable.
+                    // codegen_allocation_auto_imm_name returns the *address* of
+                    // the variable, so need to pattern match to extract it.
+                    let ident = match mem_var.value() {
+                        ExprValue::AddressOf(address) => match address.value() {
```
celinval, line 119: nit: You could merge these matches into one.

kani-compiler/src/codegen_cprover_gotoc/utils/utils.rs
```diff
@@ -61,6 +42,30 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    /// codegen_slice_value), this will return the string constant. Otherwise,
+    /// return `None`.
+    pub fn extract_const_message(&self, arg: &Expr) -> Option<String> {
+        match arg.value() {
+            ExprValue::Struct { values } => match &values[0].value() {
+                ExprValue::Typecast(expr) => match expr.value() {
+                    ExprValue::AddressOf(address) => match address.value() {
+                        ExprValue::Symbol { identifier } => {
+                            // lookup the string literal in the goto context
+                            let name = self.str_literals.get(identifier).unwrap();
```
celinval, line 57: not for this PR, but I was wondering if we could map directly Expr -> String Literal.

celinval (approved):
Thanks for doing this

tests/kani/Intrinsics/ArithOffset/offset_u8_ok.rs
```diff
@@ -18,7 +18,7 @@ fn test_arith_offset() {
 
         // This is okay because it's one byte past the object,
```
celinval, line 19: We should probably update the comment in these tests and use a `_x` to show that this can't really be de-referenced.

kani-compiler/src/codegen_cprover_gotoc/utils/utils.rs
zhassan-aws: Yes, that would be better. I couldn't do that because `Expr` is not "hash"able. I'll add a comment though.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
celinval: I didn't realize there was a Box in there. It makes things a bit uglier, so what you have might just be cleaner. But here is how you could merge this:

```rust
let ident = match mem_var.value() {
                        ExprValue::AddressOf(Expr {
                            value: box ExprValue::Symbol { identifier },
                            ..
                        }) => identifier,
                        _ => unreachable!("Expecting a symbol for a string literal allocation"),
                    };
```

tests/kani/Intrinsics/ArithOffset/offset_u8_ok.rs
zhassan-aws: Done

Issue #1696, opened 2022-09-19
`ansi_term` crate is unmaintained


adpaco-aws:
We got a [dependabot alert](https://github.com/model-checking/kani/security/dependabot/2) because the `ansi_term` crate has been marked as "not maintained". This crate is not a direct dependency of our project, but is imported through the following crates which are actual dependencies:
 * `tracing-tree v0.2.1`
 * `tracing-subscriber v0.3.15`
 * `clap v2.34.0`

The security alert isn't critical, but we should remove this dependency as soon as possible to avoid issues in the future. To this end, we should upgrade the versions of the crates listed above to more recent version that don't use `ansi_term` or replace them with other crates that don't use it.

PR #1904, opened 2022-11-15
Update `kani-compiler` to use `clap` v4

### LINKED ISSUE #1902: Update `kani-compiler` to use `clap` v4
**Proposed change:**

Update `kani-compiler` to use `clap` v4.

**Motivation:**

Our current version of `clap` depends on the crate `ansi_term`, which is no longer maintained (see #1696).

aaronbembenek-aws:
### Description of changes: 

Update `kani-compiler` to use `clap` v4, which has a slightly different builder API than `clap` v2 (which is what it is currently using). I followed the [migration guide](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrating). This is part of removing our (transitive) dependency on the unmaintained `ansi_term` crate (#1696).

### Resolved issues:

Resolves #1902 

### Call-outs:

- The old code specified that `-?` should return the `kani-compiler` version (`-V` is used to return the rustc version). However, when I tried to actually use `-?` with the old parser, it does not work. The new parser does correctly handle this argument, although I'm not sure if it something we genuinely want (or if it was in the original code as some sort of hack).
- Do we want to allow non-valid UTF-8 in arguments? I tried to reproduce the behavior of the old parser, so that the new parser panics in the same places where the old parser would have. In particular, there are some cases where the panic would have occurred in `clap`, and other cases where the old parser extracts an `OsString` from `clap` and then panics when trying to convert it to a string. However, as far as I can tell, the old parser always panic one way or another on non-valid UTF-8; in this case, it seems cleaner just to instruct `clap` to disallow it entirely.

### Testing:

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,7 +95,7 @@ version = "0.15.0"
 dependencies = [
  "anyhow",
  "cargo_metadata",
- "clap 4.0.24",
+ "clap 4.0.25",
  "which",
 ]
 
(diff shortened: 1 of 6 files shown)


celinval (approved):
That's awesome! Thank you for doing this.

tedinski (approved):
We do not need to worry about non-UTF-8 for now, and the version change looks good.

Issue #1690, opened 2022-09-16
RFC: New default output for Kani


tedinski:
This issue is a task to write/propose an RFC for a new default output format for Kani.

There are a number of needs here:

1. #1194
2. #1653 (closed but has good discussion)
3. #598
4. #1675
5. #1460
6. We should give better output for unbounded unwinding that the raw CBMC output.
7. We need to think about the output for a parallel harness runner. (e.g. unwinding needs a solution because we currently rely on printing CBMC output as it comes, which only works while we run one harness at a time. But we should also think about what status information the runner should give. Perhaps a gradle-style "here's what running on each thread" in the console with output appearing above it, for instance.)

Being able to introduce a parallel runner is my focus, but the rest should be answered/solved through this RFC as well.

The goal should be (I think:)

1. Default output
2. Verbose output
3. Quiet output (probably nothing: just exit status)
4. Plus: possible differences for being interactive (terminal) versus batch (redirected to file / CI). (Colors + live status info as considerations)

I think I will designate as out-of-scope:

1. #942

There are likely a lot of decisions to be made here. I'm not even going to start the list of questions yet.

tedinski:
Some notes on inspiration for a new default output:

`cargo test` looks like this:

```
running 28 tests
test args::tests::check_abs_type ... ok
test args_toml::tests::check_merge_args_with_only_config_kani_args ... ok
[..snip..]
test call_cbmc::tests::check_resolve_unwind_value ... ok

test result: ok. 28 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

`cargo build` gives a status like this:

```
   Compiling indexmap v1.9.1
   Compiling rayon v1.5.3
   [..snip..]
   Compiling object v0.29.0
    Building [===================>     ] 132/165: clap, object, regex, num-bigint, syn, rayon
```

This nicely shows parallelism (`clap, object, regex, num-bigint, syn, rayon`) on just one line. Meanwhile, you can scroll the output above it.

I think a hybrid might be nice, augmented with more statistics

```
   Compiling crate v0.0.0
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
    Verified crate::harness ... ok, 0.95s, 355 props
    Verified crate::other_harness ... ok, 3.42s, 1660 props 
   Verifying [===================>     ] 2/4: crate::mod::harness, crate::mod::more_harness
```

tedinski:
[...]

PR #1906, opened 2022-11-15
RFC for cover ctatement

### LINKED ISSUE #1905: RFC: Cover statement
This is a tracking issue for the RFC: `cover_statement`.

zhassan-aws:
### Description of changes: 

An RFC for adding a new `cover` API to Kani.

A rendered version can be found here:

https://github.com/zhassan-aws/kani/blob/5f0fb98507d37bd13a900398d9a3f8a784f92e6c/rfc/src/rfcs/0003-cover-statement.md

### Resolved issues:

Resolves #1905 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,76 @@
...
+    ...
+    val_i => kani::cover(true), // verify that `x` can be `val_i`
+}
+```
+
+## User Experience
+
+The `cover` function instructs Kani to find _at least one_ possible execution that satisfies the specified condition at that line of code. If there is no such execution, verification *fails*.
+
+Each cover statements will be reported as a check whose description is "condition is satisfiable" and whose status is:
```
celinval, line 47: What about UNDETERMINED?

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,76 @@
...
+- `SUCCESS` (green): if Kani found an execution that would cover the condition
```
celinval, line 48: What about `COVERED` and `UNREACHABLE`?

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,76 @@
...
+
+- What is the impact of not doing this?
+The current workaround to accomplish the same effect of verifying that a condition can be covered is to use `assert!(!cond)`.
+However, if the condition can indeed be covered, verification would fail due to the failure of the assertion.
+
+## Open questions
+
+## Other Considerations
+
+We need to make sure the concrete playback feature can be used with `cover` statements that were found to be coverable.
```
celinval, line 68: +1

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Future possibilities
+
+Users may want to specify a custom message for the cover property, in which case, we may consider adding a separate function that takes a static `str`, e.g.
+```rust
+fn cover_msg(cond: bool, msg: &'static str)
```
celinval, line 74: Would it make sense to add a message now?

celinval (commented):
Awesome! I have 2 questions about UX:
1. What will be the `Description: ` of a cover statement?
2. What about the summary message?

rfc/src/rfcs/0003-cover-statement.md
celinval: I like the idea of adding the function and the macro. The function takes a `str` and the macro generates one using the user expression if no message was provided.

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,86 @@
...
+}
+```
+
+## User Experience
+
+The `cover` function instructs Kani to find _at least one_ possible execution that satisfies the specified condition at that line of code. If there is no such execution, verification *fails*.
+
+Each cover statements will be reported as a check whose description is "condition is satisfiable" and whose status is:
+- `SUCCESS` (green): if Kani found an execution that satisfies the condition
+- `FAILURE` (red): if Kani proved that the condition cannot be satisfied
```
tedinski, line 49: IMO we need better results here. There's three cases:

1. Success: reachable, satisfiable
2. Failure: reachable, not satisfiable
3. Failure: unreachable

I think it's very important for debugging to distinguish these.

rfc/src/rfcs/0003-cover-statement.md
celinval: I think SUCCESS and FAILURE are rather confusing today, so I was hoping to avoid that for covers. I'm OK with leaving as is for now though. @tedinski are you thinking about improving these statuses as part of #1690?

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,86 @@
...
+- `kani-compiler`: The `cover` function will be handled via a hook and codegen as an assertion (`cover(cond)` will be codegen as `assert(!cond)`)
+- `kani-driver`: The CBMC output parser will extract cover properties through their property class, and their result will be flipped.
+
+## Rationale and alternatives
+
+- What is the impact of not doing this?
+The current workaround to accomplish the same effect of verifying that a condition can be covered is to use `assert!(!cond)`.
+However, if the condition can indeed be covered, verification would fail due to the failure of the assertion.
+
+## Open questions
```
tedinski, line 74: Please comment on the ability to deprecate the current `expect_fail`. I think we could deprecate and replace this with cover, right?

rfc/src/rfcs/0003-cover-statement.md
zhassan-aws: That's a good point! This would require that we generate a "cover current location" check for every cover property the user writes. I'll add that.

rfc/src/rfcs/0003-cover-statement.md
zhassan-aws: Good point. I'll add that.

rfc/src/rfcs/0003-cover-statement.md
aaronbembenek-aws: Not for this RFC, but food for thought: How could we help users debug cases where a `cover` statement is not coverable? I suppose this would be similar to helping users understand why a program is in fact safe (i.e., why no bad state is reachable).

rfc/src/rfcs/0003-cover-statement.md
zhassan-aws: Done

rfc/src/rfcs/0003-cover-statement.md
zhassan-aws: Done

rfc/src/rfcs/0003-cover-statement.md
zhassan-aws: Done

camshaft:
How does this all work with with linker slicing and multiple harnesses? Say for example I have the following code

```rust
fn my_function() {
    kani::cover!(true):
}

#[kani::proof]
fn proof_a() {
    // Calls my function 
    my_function();
}

#[kani::proof]
fn proof_b() {
    // Doesn't call my function - the cover isn't reachable from this proof. 
    // Should it fail?
}
```

For `proof_b` I imagine it's possible to want two different answers to if it should fail or not, depending on what you're trying to accomplish. 

On the one hand, if I know `proof_b` will never call `my_function`, then I definitely don't want it to fail because they're unrelated. If it failed that would mean that I need to include every single `cover` statement in every single proof in the codebase.

On the other hand, what if I want to make sure that `my_function` is always reachable by `proof_b`? Let's say `my_function` is some critical code that must be called at some point in relation to another function. `proof_b` would pass and the only way I would know it's proving what I think it's proving is doing an audit of the properties output. And even if you see it at the time of writing the proof, there could be a change in the future where it's no longer reachable but the proof still passes.

zhassan-aws:
Good questions @camshaft!

I would expect cover properties to be primarily used within proof harnesses since `cover` is only interpreted by Kani. But in the example you gave, the cover property won't fail for `proof_b` because slicing (more specifically CBMC's `--drop-unused-functions`) is applied to each harness separately, so the cover property won't be checked for `proof_b`.

But if a function is not "obviously" unused, e.g.:
```rust
fn my_function() {
    kani::cover!(true):
}

fn foo(x: bool) {
    if x {
        my_function();
    }
}

#[kani::proof]
fn proof_a() {
    my_function(true);
}

#[kani::proof]
fn proof_b() {
    my_function(false);
}
```
The cover property will fail for `proof_b`.

aaronbembenek-aws:
> Good questions @camshaft!
> 
> I would expect cover properties to be primarily used within proof harnesses since `cover` is only interpreted by Kani. But in the example you gave, the cover property won't fail for `proof_b` because slicing (more specifically CBMC's `--drop-unused-functions`) is applied to each harness separately, so the cover property won't be checked for `proof_b`.
> 
> But if a function is not "obviously" unused, e.g.:
> 
> ```rust
> fn my_function() {
>     kani::cover!(true):
> }
> 
> fn foo(x: bool) {
>     if x {
>         my_function();
>     }
> }
> 
> #[kani::proof]
> fn proof_a() {
>     my_function(true);
> }
> 
> #[kani::proof]
> fn proof_b() {
>     my_function(false);
> }
> ```
> 
> The cover property will fail for `proof_b`.

Hm, this is sort of tricky. For example, what if CBMC changed the way it does program slicing? That could change whether a proof harness passes or fails, right? If so, that behavior strikes me as surprising. 

Maybe we could restrict `kani::cover` to be only in harnesses. Or alternatively we could add another argument to `kani::cover` identifying the harnesses it is applicable to, and make sure that CBMC considers the relevant code to be reachable when handling those harnesses.

camshaft:
Yes this is a bit tricky. These `cover!` statements should ideally be properties of a particular proof, not the general code base. And I think that's OK as long as we document this aspect.

As a side note, this got me thinking about how to reason about coverage of a program in relation to its inputs. I wrote up a quick sketch of what it could look like: https://gist.github.com/camshaft/9cc5e93c1bfdf96ec7dddfd1264a8b5d

celinval:
I don't know if we should limit cover statements to only be used inside a harness although I do agree that the current behavior would be rather cryptic. Ideally, we should report all properties that were dropped by CBMC as unreachable, and that should be the case for assertions and cover statements.

This discussion made me wonder if we should really fail the verification because a cover statement.

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Summary
+
+A new Kani API that allows users to check that a certain condition can occur at a specific location in the code.
+
+## User Impact
+
+Users typically want to gain confidence that a proof checks what it is supposed to check, i.e. that properties are not passing vacuously due to an over-constrained environment.
+
+A new Kani macro, `cover` will be created that can be used for checking that a certain condition _can_ occur at a specific location in the code.
```
adpaco-aws, line 16: Something that isn't clear to me after reading this section is how `cover` is different from `assert`. If my understanding is correct, `assert` checks a condition, while `cover` checks a condition AND ensures that the statement itself is reachable.

If we agree on that, then the "_can_" in this sentence should be "_will_" instead.

rfc/src/rfcs/0003-cover-statement.md
adpaco-aws: While I agree it's an important detail for debugging, I think this would complicate our verification output by adding another "status". Perhaps it would be simpler to just generate two checks for each cover statement, one for satisfiability and another one for reachability.

zhassan-aws:
Good discussion. How about we make an unsatisfiable cover property non-failing by default, but add an option, e.g. `--fail-uncoverable` that makes it a failure?

rfc/src/rfcs/0003-cover-statement.md
camshaft: I think conceptually they are similar. `assert` says this condition must hold for _all_ executions. `cover` says this condition must hold for _at least one_ execution.

zhassan-aws:
> Good discussion. How about we make an unsatisfiable cover property non-failing by default, but add an option, e.g. `--fail-uncoverable` that makes it a failure?

Updated RFC to reflect this discussion.

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,141 @@
...
+kani::assume(len < 5);
+for _i in 0..len {
+    v.push(kani::any());
+}
+// make sure we can get a vector of length 5
+kani::cover!(v.len() == 5);
+```
+This is typically used to ensure that verified checks are not passing _vacuously_, e.g. due to overconstrained pre-conditions.
+
+The special case of verifying that a certain line of code is reachable can be achieved using `kani::cover!(true)`, e.g.
```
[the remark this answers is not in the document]
zhassan-aws: Good idea! I'll update the RFC.

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,141 @@
...
+```
+Check 2: main.cover.2
+         - Status: SATISFIED
+         - Description: "cover condition "a == 0""
+         - Location: foo.rs:9:5 in function main
+```
+
+### Impact on Overall Verification Status
+
+By default, unsatisfiable and unreachable `cover` checks will not impact verification success or failure.
```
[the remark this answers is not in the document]
zhassan-aws: What I meant is that they won't cause an otherwise successful verification run to fail (i.e. if all normal checks are passing). Should I clarify the sentence a bit, or did you mean something else?

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,141 @@
...
+fn another_harness_that_doesnt_call_foo() {
+    // ...
+}
+
+fn foo() {
+    kani::cover!( /* some condition */);
+}
+```
+
+We can consider adding an option that would cause verification to fail if a cover property was unsatisfiable or unreachable, e.g. `--fail-uncoverable`.
```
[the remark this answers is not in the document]
zhassan-aws: I'll move it to the open questions.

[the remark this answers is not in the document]
zhassan-aws: Done.

rfc/src/rfcs/0003-cover-statement.md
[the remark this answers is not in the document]
celinval: I think it's OK

celinval (approved):
I'm OK with the initial proposal. I think we can tune it as we try it out. Please don't forget to get a second approval.

Thanks!

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+## User Experience
+
+The `cover` macro instructs Kani to find _at least one_ possible execution that satisfies the specified condition at that line of code.  If no such execution is possible, the check is reported as *unsatisfiable*.
+
+Each cover statements will be reported as a check whose description is `cover condition "cond"` and whose status is:
+- `SATISFIED` (green): if Kani found an execution that satisfies the condition
+- `UNSATISFIABLE` (yellow): if Kani proved that the condition cannot be satisfied
+- `UNREACHABLE` (yellow): if Kani proved that the cover statement itself cannot be reached
```
adpaco-aws, line 52: Just a few suggestions in this paragraph:
 1. "statements" (1st occurrence) -> "statement"
 2. Dots at the end of each sentence.
 3. Maybe use ``cover condition `<cond>` ``?

rfc/src/rfcs/0003-cover-statement.md
```diff
@@ -0,0 +1,141 @@
...
+For example, for the following `cover` statement:
+```rust
+kani::cover!(a == 0);
+```
+An example result is:
+```
+Check 2: main.cover.2
+         - Status: SATISFIED
+         - Description: "cover condition "a == 0""
+         - Location: foo.rs:9:5 in function main
```
adpaco-aws, line 63: Point 3 in my previous comment (use ``cover condition `<cond>` ``) also applies here, because otherwise you end up with `""` at the end.

adpaco-aws (approved):
My main concern with the current proposal is the fact that unsatisfiable/unreachable `cover` statements don't affect overall verification results, leading to surprises later. An alternative is to switch the proposal so the option becomes `--allow-uncoverable` instead (think of it as a sign-off from the user).

However, there's no way to know if that's what we want at the moment, let's get an initial version out and see where it goes from here.

rfc/src/rfcs/0003-cover-statement.md
zhassan-aws: Fixed the first two.

For the third, I changed it to:
```
cover condition: cond
```
to mimic the assert message:
```
assertion failed: 1 + 1 == 3
```
I hope this is OK.

PR #1918, opened 2022-11-18
Allow panic with args to be called in a const context

### LINKED ISSUE #1586: "cannot call non-const fn `kani::assert` in constant functions"

zhassan-aws:
### Description of changes: 

Calling `panic!` with args, e.g.
```rust
panic!("{}", msg)
```
is allowed by `rustc` in const contexts, but Kani compilation currently fails. This PR updates Kani's definition of the `panic` macro to allow this to go through.

### Resolved issues:

Resolves #1586 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

### Checklist


library/std/src/lib.rs
```diff
@@ -186,22 +186,12 @@ macro_rules! panic {
...
-    // ```
-    // panic!(concat!("Message {}", " split in two"), argument);
-    // ```
-    // The std implementation of `panic!()` macro is implemented in the compiler and it seems to
-    // be able to do things that we cannot do here.
-    // https://github.com/rust-lang/rust/blob/dc2d232c7485c60dd856f8b9aee83426492d4661/compiler/rustc_expand/src/base.rs#L1197
-    ($msg:expr, $($arg:tt)+) => {{
+    ($($arg:tt)+) => {{
         if false {
-            let _ = format_args!($msg, $($arg)+);
```
celinval, line 203: Any chance you can add a note on why this is here? I was also wondering if rustc will prune this. If not, can we use a noop assignment instead? Something like:

```
let _ = ($($arg)+)
```

tests/expected/panic/panic-2018/expected
```diff
@@ -1,10 +1,10 @@
...
 Failed Checks: msg
 Failed Checks: explicit panic
 Failed Checks: Panic message
 Failed Checks: "Panic message with arg {}", "str"
 Failed Checks: "{}", msg
 
-Failed Checks: concat!("ArrayVec::", "try_insert",\
-": index {} is out of bounds in vector of length {}"),\
-5, 3
+Failed Checks: concat!\
```
celinval, line 8: Is this new line on purpose?

celinval (commented):
Can you please also add a test that panic during const evaluation?

library/std/src/lib.rs
zhassan-aws: This line is supposed to make sure that any incorrect usage of `panic`'s formatting arguments is reported by Kani, e.g.
```rust
panic!("{}"); // positional argument but no argument given
```
However, I just found out that it's not doing that :( I'll have to figure out a fix.

tests/expected/panic/panic-2018/expected
zhassan-aws: TBH, I have no idea how this new line came about. It's not clear how where those new lines are coming from.

zhassan-aws:
> Can you please also add a test that panic during const evaluation?

Sure.

library/std/src/lib.rs
zhassan-aws: The error checking seems to be broken for the `assert` macro as well with https://github.com/model-checking/kani/pull/1858 :(

celinval (approved):
Awesome!! I was wondering if the compiler will also prune the assert before codegen. 🤞

zhassan-aws:
> Awesome!! I was wondering if the compiler will also prune the assert before codegen. 🤞

Is there a way to examine the MIR that the codegen module receives?

zhassan-aws:
Looks like it does get removed, e.g.
```
$ cat test.rs 
#[kani::proof]
fn main() {
    let x: i32 = kani::any();
    let y = x / 2;
    assert!(y <= x, "This is a message");
}
$ RUSTFLAGS="--emit mir" kani test.rs
```
This is what `main` looks like in the output:
```
fn main() -> () {
    let mut _0: ();                      // return place in scope 0 at test.rs:2:11: 2:11
    let _1: i32;                         // in scope 0 at test.rs:3:9: 3:10
    let mut _3: i32;                     // in scope 0 at test.rs:4:13: 4:14
    let mut _4: bool;                    // in scope 0 at test.rs:4:13: 4:18
    let mut _5: bool;                    // in scope 0 at test.rs:4:13: 4:18
    let mut _6: bool;                    // in scope 0 at test.rs:4:13: 4:18
    let _7: ();                          // in scope 0 at /home/ubuntu/git/kani/library/std/src/lib.rs:44:9: 44:59
    let mut _8: bool;                    // in scope 0 at test.rs:5:13: 5:19
    let mut _9: i32;                     // in scope 0 at test.rs:5:13: 5:14
    let mut _10: i32;                    // in scope 0 at test.rs:5:18: 5:19
    scope 1 {
        debug x => _1;                   // in scope 1 at test.rs:3:9: 3:10
        let _2: i32;                     // in scope 1 at test.rs:4:9: 4:10
        scope 2 {
            debug y => _2;               // in scope 2 at test.rs:4:9: 4:10
        }
    }

    bb0: {
        _1 = kani::any::<i32>() -> bb1;  // scope 0 at test.rs:3:18: 3:29
                                         // mir::Constant
                                         // + span: test.rs:3:18: 3:27
                                         // + literal: Const { ty: fn() -> i32 {kani::any::<i32>}, val: Value(<ZST>) }
    }

    bb1: {
        _3 = _1;                         // scope 1 at test.rs:4:13: 4:14
        _4 = const false;                // scope 1 at test.rs:4:13: 4:18
        _5 = Eq(_3, const i32::MIN);     // scope 1 at test.rs:4:13: 4:18
        _6 = BitAnd(move _4, move _5);   // scope 1 at test.rs:4:13: 4:18
        assert(!move _6, "attempt to compute `{} / {}`, which would overflow", _3, const 2_i32) -> bb2; // scope 1 at test.rs:4:13: 4:18
    }

    bb2: {
        _2 = Div(move _3, const 2_i32);  // scope 1 at test.rs:4:13: 4:18
        _9 = _2;                         // scope 2 at test.rs:5:13: 5:14
        _10 = _1;                        // scope 2 at test.rs:5:18: 5:19
        _8 = Le(move _9, move _10);      // scope 2 at test.rs:5:13: 5:19
        _7 = kani::assert(move _8, const "\"This is a message\"") -> bb3; // scope 2 at /home/ubuntu/git/kani/library/std/src/lib.rs:44:9: 44:59
                                         // mir::Constant
                                         // + span: /home/ubuntu/git/kani/library/std/src/lib.rs:44:9: 44:21
                                         // + literal: Const { ty: fn(bool, &'static str) {kani::assert}, val: Value(<ZST>) }
                                         // mir::Constant
                                         // + span: /home/ubuntu/git/kani/library/std/src/lib.rs:44:29: 44:58
                                         // + literal: Const { ty: &str, val: Value(Slice(..)) }
    }

    bb3: {
        return;                          // scope 0 at test.rs:6:2: 6:2
    }
}

library/std/src/lib.rs
zhassan-aws: Error checking was restored with the switch to using `core::assert`.

PR #1920, opened 2022-11-21
Validate function/method stubs

### LINKED ISSUE #1892: Validate function/method stubs
Right now, we are not doing any validation for function/method stubs. We should check and make sure that they are actually compatible with the functions/methods they are replacing. A relatively simple (but somewhat restrictive) check would make sure that the stub and the function/method it is replacing have the same types; this may involve some subtle reasoning for generic types.

aaronbembenek-aws:
### Description of changes: 

Currently we do not validate that a stub is actually compatible with the original function/method. When a user tries to use a non-compatible stub, this often shows up as a hard-to-debug type mismatch in the generated GotoC code.

This PR introduces a stub validation mechanism. The stub is required to have the _exact_ same types as the original function/method, including the naming of type variables. However, type variables are not required to implement the same traits, so long as the trait mismatch does not lead to a trait method being unresolvable during monomorphization.

Validation happens in two places. First, arities and types are checked during the MIR transformation when the original body is replaced with the stub's body. Second, the reachability analysis detects missing trait methods when it tries to resolve monomorphized calls.

I will separately update the RFC with a description of stub validation and examples.

### Resolved issues:

Resolves #1892

### Call-outs:

- I needed to customize the `collect_and_partition_mono_items` query. When the input program is treated as a library, this query is called to construct library metadata; this happens after the MIR transformation but before our GotoC code generation. With the default query, rustc code would panic on mismatched traits before we have a chance to gracefully handle that situation. The custom query runs the reachability analysis first (to catch mismatches) before running the default query.
- I renamed the module `kani_middle::mir_transform` to `kani_middle::provide`, since it now has code for hooking up rustc queries in general, and not just for hooking up new MIR transformations.

### Testing:

* How is this change tested? New regression tests

### Checklist


tests/kani/StubbingValidateMismatchedTraits/main.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright Kani Contributors
```
celinval, line 1: nit: Inside `kani/` suites we tend to use a subfolder for a feature. The name of the test reflects the scenario you are testing.

tests/ui/stubbing-type-validation/expected
```diff
@@ -0,0 +1,5 @@
+error: Arity mismatch: original function/method `f1` takes 1 argument(s), stub `f2` takes 0
+error: Type mismatch: stub `g2` has type `u32` where original function/method `g1` has type `i32`
```
celinval, line 2: I think this should be  `Type of parameter 0 differ`

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -13,24 +13,75 @@ use rustc_hir::def_id::{DefId, LocalDefId};
...
+    // indices [1, `arg_count`] refer to the parameters.
+    let mut matches = true;
+    for i in 0..=old_body.arg_count {
+        let old_arg = old_body.local_decls.get(i.into()).unwrap();
+        let new_arg = stub_body.local_decls.get(i.into()).unwrap();
+        if old_arg.ty != new_arg.ty {
+            tcx.sess.span_err(
+                new_arg.source_info.span,
+                format!(
+                    "Type mismatch: stub `{}` has type `{}` where original function/method `{}` has type `{}`",
```
celinval, line 74: Can you please specify which parameter / return type mismatch?

tests/ui/stubbing-type-validation/type_mismatch.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// kani-flags: --harness harness --enable-unstable --enable-stubbing
+//
+//! This tests that we catch type mismatches between the stub and the original
```
celinval, line 6: Can you please add coverage for the following:

- Return types differ.
- Wrong combination of generic parameters. Something like:
```rust
fn original<X, Y>(_x: &X, _y: &Y) -> X {}
fn stub<X, Y>(_x: &X, _y: &Y) -> Y {}
```
and
```rust
fn original<X, Y>(_x: &X, _y1: &Y, _y2: &Y) {}
fn stub<X, Y>(_x1: &X, _x2: &X, _y: &Y) {}
```

celinval (commented):
Thanks for doing this.

Not a problem for this PR, but we will have to think about the overall flow of this feature before stabilization.

kani-compiler/src/kani_middle/reachability.rs
```diff
@@ -400,11 +401,29 @@ impl<'a, 'tcx> MirVisitor<'tcx> for MonoItemsFnCollector<'a, 'tcx> {
...
                 let fn_ty = self.monomorphize(callee_ty);
                 if let TyKind::FnDef(def_id, substs) = *fn_ty.kind() {
-                    let instance =
+                    let instance_opt =
                         Instance::resolve(self.tcx, ParamEnv::reveal_all(), def_id, substs)
-                            .unwrap()
                             .unwrap();
-                    self.collect_instance(instance, true);
+                    match instance_opt {
+                        None => {
```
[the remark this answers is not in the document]
aaronbembenek-aws: I think we should be able to check whether the current function's body has been stubbed out, in which case it is likely to be a trait mismatch caused by stubbing. I believe that these types of trait mismatches can occur only in the bodies of stubbed out functions (although it's a little tricky to convince oneself of that).

Should I add a check and update the error message accordingly?

tests/kani/StubbingValidateMismatchedTraits/main.rs
aaronbembenek-aws: I can make that change. Should I do something similar for the other tests, or just the `kani/` ones?

tests/cargo-ui/function-stubbing-trait-mismatch/src/expected
```diff
@@ -0,0 +1,3 @@
+error: Unable to resolve call to `DoIt::do_it`; this is either an internal Kani bug, or there is a trait mismatch between a stub you are using and the original function/method.
```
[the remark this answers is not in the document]
aaronbembenek-aws: I had two tests because one treats the code as a library (the `ui` one) and one treats it as an executable (the `cargo-ui` one). These end up hitting the validation code along different code paths. Is there a better way to test this? Also, what sort of sanity check for `cargo-ui` did you have in mind?

kani-compiler/src/kani_middle/stubbing/transform.rs
aaronbembenek-aws: I've updated the messages to include this information, so that with the spans they now look like:
```
error: Type of parameter 0 differs: stub `h2` has type `T` where original function/method `h1` has type `S`
  --> type_mismatch.rs:29:10
   |
29 | fn h2<T>(_x: T) -> bool {
   |          ^^
```

and

```
error: Return type differs: stub `g2` has type `i32` where original function/method `g1` has type `bool`
  --> type_mismatch.rs:21:55
   |
21 | fn g2(_x: bool, _y: u32, _z: &mut bool, _zz: bool) -> i32 {
   |                                                       ^^^
```

tests/ui/stubbing-type-validation/type_mismatch.rs
aaronbembenek-aws: Added!

kani-compiler/src/kani_middle/reachability.rs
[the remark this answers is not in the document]
celinval: Yes. We need to have a clear message when the stub is incompatible.

kani-compiler/src/kani_middle/stubbing/transform.rs
celinval: That's awesome! Thanks

tests/kani/StubbingValidateMismatchedTraits/main.rs
celinval: Thanks! Unfortunately the expected tests and cargo-tests work a bit differently. You could create a folder `stubbing/` and move everything to subfolders there to keep things more organized, but we currently don't have a clear pattern. So I would say depends on how many tests you are adding to each suite. :)

kani-compiler/src/kani_middle/reachability.rs
```diff
@@ -400,11 +401,29 @@ impl<'a, 'tcx> MirVisitor<'tcx> for MonoItemsFnCollector<'a, 'tcx> {
...
+                            // force type variables in the stub's signature to
+                            // implement the same traits as those in the
+                            // original function/method. A trait mismatch shows
+                            // up here, when we try to resolve a trait method
+                            // that the type does not implement.
+                            tcx.sess.span_err(
+                                terminator.source_info.span,
+                                format!(
+                                    "Unable to resolve call to `{}`; \
+                        this is either an internal Kani bug, or there is a trait mismatch \
```
adpaco-aws, line 419: I'd avoid saying "this is either an internal Kani bug". Of course, it's possible we have a bug here (or anywhere in the tool, especially when it comes to unstable features), but let's first make clear what the error we found was, so the user can take action on it. They can always open an issue if they can't figure it out.

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -13,24 +13,75 @@ use rustc_hir::def_id::{DefId, LocalDefId};
...
+    old_body: &'a Body<'tcx>,
+    stub_def_id: DefId,
+    stub_body: &'a Body<'tcx>,
+) -> bool {
+    // Check whether the arities match.
+    if old_body.arg_count != stub_body.arg_count {
+        tcx.sess.span_err(
+            tcx.def_span(stub_def_id),
+            format!(
+                "Arity mismatch: original function/method `{}` takes {} argument(s), stub `{}` takes {}",
```
adpaco-aws, line 55: These error messages look pretty nice. Can we avoid capitalization (and punctuation, if any) so they follow [`rustc` guidelines](https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-structure)?

tests/cargo-kani/stubbing-validate-random/src/main.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright Kani Contributors
```
adpaco-aws, line 1: I don't this test is being run because it doesn't have an `main.expected` file.

tests/cargo-ui/function-stubbing-trait-mismatch/src/expected
adpaco-aws, line 1: Please ensure this test is being run as well.

adpaco-aws (commented):
Please make sure that these test are running. For example, in `cargo-kani` tests you are supposed to include a `<function>.expected` file for each function you're testing, but I don't think `compiletest` is complaining (but it should) if it doesn't find them.

kani-compiler/src/kani_middle/stubbing/transform.rs
aaronbembenek-aws: I've removed capitalization and any periods

kani-compiler/src/kani_middle/reachability.rs
aaronbembenek-aws: I've changed this, so if we're in the body of a function that has been stubbed it says something like:
```
error: unable to resolve call to `DoIt::do_it`; there is likely a trait mismatch between the signatures for the stub `bar` and the original function/method `foo`
```

tests/kani/StubbingValidateMismatchedTraits/main.rs
aaronbembenek-aws: Okay, I moved all the `kani/` stubbing tests into a subfolder

tests/cargo-kani/stubbing-validate-random/src/main.rs
celinval: BTW, you can just add a `expected` file where `Cargo.toml` lives and compiletest will run `cargo kani` without `--harness` and compare the output.

aaronbembenek-aws: Thanks! I took care of this

kani-compiler/src/kani_middle/reachability.rs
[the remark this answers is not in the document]
celinval: I posted a response to your original comment

kani-compiler/src/kani_middle/reachability.rs
aaronbembenek-aws: Okay, it now has this message:
```
error: `&&str` doesn't implement `DoIt`. The function `foo` cannot be stubbed by `bar` due to generic bounds not being met.
```
I could probably remove the `&`s if that's preferable.

celinval: It depends where the `&` comes from. The type you print should be the type you are trying to check the bounds against. In some cases it will have `&`, but others won't.

aaronbembenek-aws: Cool - I updated to that behavior, so that we now print this if `T == &str`:
```
error: `&str` doesn't implement `std::cmp::PartialEq`. The function `foo` cannot be stubbed by `bar` due to generic bounds not being met.
```
but this if `T == String`:
```
error: `std::string::String` doesn't implement `DoIt`. The function `foo` cannot be stubbed by `bar` due to generic bounds not being met.
```

celinval (approved):
Thanks!

adpaco-aws (approved):

Issue #1556, opened 2022-08-19
Spurious unused variable warnings for variables used in asserts


zhassan-aws:
I tried this code:

```rust
#[cfg_attr(kani, kani::proof)]
fn main() {
    let s = "foo";
    assert!(1 + 1 == 3, "arg is {}", s);
}
```

using the following command line invocation:

```
kani assert_arg.rs
```

with Kani version: 8b1d18c4363

I expected to see this happen: No warnings

Instead, this happened: Kani emits the following spurious warning:
```
warning: unused variable: `s`
 --> assert_arg.rs:3:9
  |
3 |     let s = "foo";
  |         ^ help: if this is intentional, prefix it with an underscore: `_s`
  |
  = note: `#[warn(unused_variables)]` on by default

warning: 1 warning emitted
```
This is due to Kani's overridden assert macros which stringify the arguments during macro expansion, but do not use the actual arguments in the expansion.

PR #1924, opened 2022-11-21
Ignore messages for assert, panic, and unreachable macros

### LINKED ISSUE #1586: "cannot call non-const fn `kani::assert` in constant functions"

zhassan-aws:
### Description of changes: 

We've had a myriad of issues related to the processing of messages for the `assert`, `panic`, and `unreachable` macros using the `format_args` macro. The purpose of using `format_args` in an unreachable (`if false`) block was to make sure the args are checked for errors/warnings. However, the usage of `format_args` caused Kani to fail to compile crates that use `assert` and `panic` in const contexts (e.g. https://github.com/time-rs/time/blob/4c13327d5fed8571316b181e828067fc54038e05/time/src/lib.rs#L362). Fixing this would require using a macro that has the following properties:
1. It processes the format args and checks for errors
2. It can be used in a const context
3. It is not overridden by Kani

We're unaware of a macro that satisfies those 3 properties. The `format_args` macro seems to be on its way to be usable in const contexts as the error message indicates:
```
error: `std::fmt::Arguments::<'a>::new_v1` is not yet stable as a const fn
   --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/time-0.3.17/src/lib.rs:356:5
    |
356 |     panic!("{}", message)
    |     ^^^^^^^^^^^^^^^^^^^^^
    |
    = help: add `#![feature(const_fmt_arguments_new)]` to the crate attributes to enable
    = note: this error originates in the macro `format_args` which comes from the expansion of the macro `panic` (in Nightly builds, run with -Z macro-backtrace for more info)
```
so in the meantime, we've decided to drop the processing of arguments. This may cause Kani to **not** produce compile-time errors in some cases.

### Resolved issues:

Resolves #1586 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

This PR will reintroduce two issues that were previously addressed by processing the messages using the `format_args` macro:
1. #803 
5. #1556 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


tests/expected/assert-arg-error/test.rs
```diff
@@ -1,10 +0,0 @@
-// Copyright Kani Contributors
```
celinval, line 1: Can you turn this into fixme test instead?

celinval (commented):
Didn't we talk about updating the documentation? We should also mention that we don't invoke the `fmt()` function of the arguments, since they could introduce some bad behavior or side effect.

zhassan-aws:
> Didn't we talk about updating the documentation?

Added.

> We should also mention that we don't invoke the fmt() function of the arguments, since they could introduce some bad behavior or side effect.

Can you clarify what you mean?

tests/expected/assert-arg-error/test.rs
zhassan-aws: We currently don't have a way to add fixme tests that are passing, but should actually fail (see discussion on https://github.com/model-checking/kani/pull/1666). That's why I deleted it. I'm not happy about that though, so I'm open to ideas on how we can keep the test in our regressions in a meaningful way.

docs/src/overrides.md
```diff
@@ -15,8 +15,11 @@ overrides and a description of their behavior compared to the `std` versions:
...
+`assert`, `assert_eq`, and `assert_ne` macros | Skips string formatting code[^skip-errors], generates a more informative message and performs some instrumentation.
 `debug_assert`, `debug_assert_eq`, and `debug_assert_ne` macros | Rewrites as equivalent `assert*` macro |
-`print`, `eprint`, `println`, and `eprintln` macros | Skips string formatting and I/O operations |
-`unreachable` macro | Skips string formatting and invokes `panic!()` |
+`panic` | Skips string formatting code[^skip-errors] |
+`print`, `eprint`, `println`, and `eprintln` macros | Skips I/O operations |
+`unreachable` macro | Skips string formatting[^skip-errors] and invokes `panic!()` |
 `std::process::{abort, exit}` functions | Invokes `panic!()` to abort the execution |
+
+[^skip-errors]: The effect of skipping string formatting code is that Kani fails to detect and report any compiler warnings/errors associated with them (see https://github.com/model-checking/kani/issues/803)
```
celinval, line 25: I think we should also call out that we may miss invalid behavior present in the formatting code and that the analysis result can be affected if the format code has any side effect.

tests/expected/assert-arg-error/test.rs
celinval: Honestly, I just rename the test, add a comment and change the expected file to the current output. It won't be part of a fixme suite, but I think it's better than deleting it.

zhassan-aws:
I found a way to make #1918 work fully, so might abandon this PR!

zhassan-aws:
Closing in favor of #1918.

PR #1930, opened 2022-11-22
Comment operand.rs

tedinski:
### Description of changes: 

This PR is 95% comments, but the diff might be hard to read there because some code really needed moving around. All function addition/removals are actually just moves. I left their bodies alone except once where I added a comment.

There were two meaningful changes:

1. A couple of function had `pub` removed because they didn't need it.
2. I changed the "unsupported feature" message for const slices: 

```diff
+                        let operation_name = format!("Constant slice for type {}", slice_ty);
                         return self.codegen_unimplemented_expr(
-                            "Constant slice value with 2+ bytes",
+                            &operation_name,
```

Everything else should largely be adding doc comments to functions, so kinda skipping through the diff to see what I'm adding there is the easiest way to proceed.

### Call-outs:

Some of this code looks like it could use revising...

### Testing:

* Is this a refactor change? yes, no behavior change

### Checklist


kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -23,6 +23,9 @@ enum AllocData<'a> {
 }
 
 impl<'tcx> GotocCtx<'tcx> {
+    /// A MIR operand is either a constant (literal or `const` declaration) or a place
```
celinval, line 26: Here is an example that I think you should first document what `codegen_operand` does. After that you can add more details on what the operand argument actually is.

celinval (commented):
Thank you!!

Can you just please make sure that you always start the documentation with a statement that says what the function is supposed to do? Here is a good ref: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
tedinski: I tend to leave that out when the name (and maybe type) is really good. What would you like to say about this function besides, uh, "codegens an operand" :)

celinval: Something like "generate a goto expression that represents a MIR operand."

celinval (approved):

PR #1932, opened 2022-11-23
Remove symbol table passes.

### LINKED ISSUE #439: Handle all reserved C names in --gen-c-runnable name transformer.
Currently, we only handle `case` and `default`, which was enough to cover our regression suite. Use a list of C reserved names to handle the rest.

### LINKED ISSUE #441: Ensure CBMC fix to --dump-c fixes --gen-c-runnable.
See https://github.com/diffblue/cbmc/issues/6268.

Test program:
```
fn help(x: u32) {
    let x: i32;
}
```

### LINKED ISSUE #444: Fix code generation for SIMD types
I'm not sure where in the translation process this should happen, but consider the following program:
```
#![feature(repr_simd)]

#[repr(simd)]
#[derive(PartialEq, Eq)]
pub struct A(i64, i64);

fn main() {
    let x = A(1, 2);
    let y = A(3, 4);
    assert!(x != y);
}
```

The generated C code from `--gen-c` does comparison like this:
```
_Bool _RNvXs_Csb7rQPrKk64Y_4mainNtB4_1ANtNtCs75dudyTCRYy_4core3cmp9PartialEq2neB4_(long int __attribute__((vector_size (2*sizeof(long int)))) *self, long int __attribute__((vector_size (2*sizeof(long int)))) *other)
{
  // ...
  long int *__self_1_0;
  long int *__self_1_1;
  long int *__self_0_0;
  long int *__self_0_1;
  // ...
  __self_1_0 = *other;
  __self_1_1 = &(*other)[1];
  __self_0_0 = *self;
  __self_0_1 = &(*self)[1];
  var_8 = *__self_0_0;
  var_9 = *__self_1_0;
  var_7 = var_8 != var_9;
  // ...
}
```

There are two main issues:
1) `__self_1_0 = *other;`: the value `*other` is a vector type, while `__self_1_0` is a `long int*`
2) `__self_1_1 = &(*other)[1];`: this leads to "address of vector element requested"; you cannot take reference to a vector element in C.

No longer relevant: <s>The `--gen-c-runnable` transformation handles this by casting the vector to a raw pointer and using array indexing, but long-term this should be fixed at the CBMC level.</s>

### LINKED ISSUE #445: Add testing script for --gen-c-runnable
Verify that generated C files are able to be compiled with gcc.

### LINKED ISSUE #1384: Clean up code related to broken --gen-c-runnable
I would like to propose that we clean the code related to the broken `--gen-c-runnable` flag. In fact, we have never enabled this flag in `kani-driver`. Please let me know if anyone has any objection.

celinval:
### Description of changes: 

This code hasn't been used since at least the creation of `kani-driver`, and I don't think we have any plans to use it in the near future. We might as well delete this since it's likely broken by now (I think it was already broken before `kani-driver`).

### Resolved issues:

Resolves #1384

This should also close the following opened issues related to `--gen-c-runnable`

Closes #445
Closes #441
Closes #439

### Call-outs:

Any objections to removing this code??? @danielsn 

### Testing:

* How is this change tested? None. There was no test coverage for this.

* Is this a refactor change?

### Checklist


adpaco-aws (commented):
I'm in favor of removing this code because it complicates development when doing changes to our CProver API, but it isn't used.

However, I don't see how this would resolve #444 . My understanding is that we're generating non-standard code for indexing SIMD vector components, as it's shown in `--gen-c` views, and the transformer just handles this case in a particular way.

celinval:
> I'm in favor of removing this code because it complicates development when doing changes to our CProver API and it isn't used.
> 
> However, I don't see how this would resolve #444 . My understanding is that we're generating non-standard code for indexing SIMD vector components, as it's shown in `--gen-c` views, and the transformer just handles this case in a particular way.

Sorry, it wasn't clear to me from the issue description or comments. I'll post your comment there too.

celinval:
Actually, I read the #444 thread again and it actually makes sense. Sorry about it.

adpaco-aws (approved):

adpaco-aws:
> Actually, I read the #444 thread again and it actually makes sense. Sorry about it.

No problem. Please wait for another review before merging.

tedinski (approved):
I think I agree with this as well. Maybe @danielsn ?

danielsn:
Agreed.  If we need this code, we can recreate it. Having unused code around adds toil without giving benifit.

PR #1934, opened 2022-11-23
Update function stubbing RFC

aaronbembenek-aws:
### Description of changes: 

This PR adds a section on stub compatibility and validation to our function/method stubbing RFC. It also addresses a few miscellaneous points that have come up in discussion.

A rendered version can be found [here](https://github.com/aaronbembenek-aws/kani/blob/update-function-stubbing-rfc/rfc/src/rfcs/0002-function-stubbing.md).

### Testing:

* How is this change tested? NA

* Is this a refactor change? NA

### Checklist


rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -156,7 +156,45 @@ Given a set of `original`-`replacement` pairs, Kani will exit with an error if
...
 3. the user specifies conflicting stubs for the same harness (e.g., if the same `original` function is mapped to multiple `replacement` functions); or
-4. the signature of the `replacement` stub is not compatible with the signature of the `original` function/method.
+4. the signature of the `replacement` stub is not compatible with the signature of the `original` function/method (see next section).
+
+### Stub compatibility and validation
+
+When considering whether a function/method can be replaced with some given stub, we want to allow some measure of flexibility, while also ensuring that we can provide the user with useful feedback if stubbing results in misformed code.
+We consider a stub to be compatible with some function/method if all the following conditions are met:
+
+- They have the same return type. The names of type variables need to match.
```
celinval, line 166: Can you please clarify: "The names of type variables need to match"? Also, I think this should be another item.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -156,7 +156,45 @@ Given a set of `original`-`replacement` pairs, Kani will exit with an error if
...
+- They have the same number of parameters.
+- Each parameter in the stub has the same type as the corresponding parameter in the original function/method.
+The names of type variables need to match.
+- During monomorphization, every statically dispatched call to a trait method can be resolved.
```
celinval, line 170: Can you please document this in terms of the language not the compiler phase? Something like: "The bounds for each type parameter don't need to match; however, all calls to the original function must also satisfy the bounds of the stub."

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -156,7 +156,45 @@ Given a set of `original`-`replacement` pairs, Kani will exit with an error if
...
+
+#[kani::proof]
+#[kani::stub(foo, bar)]
+fn harness() {
+    assert!(foo("hello"));
+}
+```
+The call to the trait method `DoIt::do_it` is unresolvable in the stub `bar` when the type variable `T` is instantiated with the type `&str`.
+On the other hand, our approach provides some flexibility, such as allowing our earlier example of mocking randomization: both `rand::random` and `my_random` have the type `() -> T`, but in the first case `T` is restricted such that the type `Standard` implements `Distribution<T>`, whereas in the latter case `T` has to implement `kani::Arbitrary`.
+This trait mismatch is harmless because it does not result in any unresolvable behavior when `T` is instantiated with `u32`.
```
celinval, line 197: Maybe remove this line. I think what you mean here is to say that this flexibility allow users to write expressive generic stubs with well defined behavior.

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: Good idea - updated to this language

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: I changed the last sentence to this:

> This trait mismatch is allowed because at this call site `T` is instantiated with `u32`, which implements `kani::Arbitrary`.

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -161,16 +161,17 @@ Given a set of `original`-`replacement` pairs, Kani will exit with an error if
...
-We consider a stub to be compatible with some function/method if all the following conditions are met:
+We consider a stub and a function/method to be compatible if all the following conditions are met:
 
-- They have the same return type. The names of type variables need to match.
 - They have the same number of parameters.
+- They have the same return type.
 - Each parameter in the stub has the same type as the corresponding parameter in the original function/method.
-The names of type variables need to match.
-- During monomorphization, every statically dispatched call to a trait method can be resolved.
+- We do **NOT** allow for the renaming of type parameters when checking whether types are equivalent; for example, the function `foo<S>(x: S) -> S` is considered to have a different type than the function `bar<T>(x: T) -> T`.
```
celinval, line 169: Oh, why is that?

celinval (commented):
It looks good to me. Just one small question

rfc/src/rfcs/0002-function-stubbing.md
aaronbembenek-aws: Not for a fundamental reason, and I've updated the RFC to be more flexible here: the stub and original function/method need to have the same number of generic parameters, but they can be named different things.

celinval (approved):
Please don't forget to get a second approval since this is a change to an RFC. Thanks!

rfc/src/rfcs/0002-function-stubbing.md
```diff
@@ -332,6 +370,11 @@ One possibility would be writing proofs about stubs (possibly relating their beh
 - It would increase the utility of stubbing if we supported stubs for types.
 The source code annotations could likely stay the same, although the underlying technical approach performing these substitutions might be significantly more complex.
 - It would probably make sense to provide a library of common stubs for users, since many applications might want to stub the same functions and mock the same behaviors (e.g., `rand::random` can be replaced with a function returning `kani::any`).
+- We could provide special classes of stubs that are likely to come up in practice:
+    - `unreachable`: assert the function is unreachable.
+    - `havoc_locals`: return non-det values and assign non-det to all mutable argument.
+    - `havoc`: similar to `havoc_locals` but also assign non-det to all mutable global.
```
adpaco-aws, line 376:
```suggestion
    - `havoc_locals`: return nondeterministic values and assign nondeterministic to all mutable arguments.
    - `havoc`: similar to `havoc_locals` but also assign nondeterministic to all mutable global variables.
```

adpaco-aws (approved):
LGTM - One suggestion.

PR #1935, opened 2022-11-23
Update `kani-driver` to use `clap` v4

### LINKED ISSUE #1696: `ansi_term` crate is unmaintained

### LINKED ISSUE #1901: Update `kani-driver` to use `clap` v4
**Proposed change:**

Update `kani-driver` to use `clap` v4.

**Motivation:**

Our current version of `clap` depends on the crate `ansi_term`, which is no longer maintained (see #1696).

aaronbembenek-aws:
### Description of changes: 

Update `kani-driver` to use `clap` v4, which has a slightly different derive API than `structopt` (which is what `kani-driver` is currently using). I followed the [migration guide](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrating). This is the final part of removing our (transitive) dependency on the unmaintained `ansi_term` crate (https://github.com/model-checking/kani/issues/1696).

### Resolved issues:

Resolves #1901 
Resolves #1696 

### Call-outs:

The only place where I am aware of a behavior mismatch between the original code and this PR is with the `--verbose` argument. In both cases, it is a boolean flag. However, in the original code, it also accepts an arbitrary number of arguments (e.g., `--verbose foo bar` is allowed). As far as I can tell, these arguments are never accessed. It is not clear how to reproduce this behavior with `clap` v4, so the new version does not allow `--verbose` to take any arguments.

### Testing:

* How is this change tested? Existing tests and some manual tests

* Is this a refactor change? Y

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,15 +33,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "anyhow"
 version = "1.0.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -95,7 +86,7 @@ version = "0.15.0"
 dependencies = [
  "anyhow",
  "cargo_metadata",
- "clap 4.0.26",
+ "clap",
  "which",
 ]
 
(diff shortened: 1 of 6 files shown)


celinval (approved):
Awesome!! Thanks

adpaco-aws (approved):
Thanks so much, @aaronbembenek-aws !

PR #1942, opened 2022-11-28
Restore `simd_extract` and `simd_insert` intrinsics

adpaco-aws:
### Description of changes: 

Restores the SIMD "construction" intrinsics `simd_extract` and `simd_insert`.

As in #1853, we must perform some type-checks that depend on the back-end. These type-checking errors are now caught and we include a couple of test cases to ensure that they remain that way.

### Resolved issues:

Towards #1148 

### Testing:

* How is this change tested? Existing regression. Adds two new tests, converts a "fixme".

* Is this a refactor change? Not completely, the main novelty here are the type-checks.

### Checklist


tests/kani/Intrinsics/SIMD/Construction/main.rs
```diff
@@ -34,8 +33,8 @@ fn main() {
     }
     {
         // Intrinsic updating
-        let m = unsafe { simd_insert(y, 0, 1) };
-        let n = unsafe { simd_insert(y, 1, 5) };
+        let m = unsafe { simd_insert(y, 0, 1_i64) };
+        let n = unsafe { simd_insert(y, 1, 5_i64) };
```
tedinski, line 37: Do we know if this limitation (needing that cast) will cause problems in how it gets used in real code?

tedinski (approved):
I think this is probably good to merge, my one concern might be a "followup PR" kind of thing if it's a problem at all...

tests/kani/Intrinsics/SIMD/Construction/main.rs
adpaco-aws: Hey @tedinski, sorry for the late reply.

Just wanted to make clear that this is just mimicking `rustc`'s current behavior. Some SIMD intrinsics require additional type-checking done by the backend, as in [here](https://github.com/rust-lang/rust/blob/c1a859b25a95999ba276075bbd8e284ea675b41a/compiler/rustc_codegen_llvm/src/intrinsic.rs#L1033-L1056). This would prevent us from accepting that, in principle, cannot be compiled with `rustc` (this was the case for this test).

PR #1943, opened 2022-11-28
Remove --dry-run from Kani.

### LINKED ISSUE #1936: Replace `--dry-run` by `--print-commands`
**Proposed change:** Replace the `--dry-run` option by something like `--print-commands`. Or even just `--verbose` (like `cargo`).

**Motivation:**

1.  First, this would be much simpler and we would be able to remove a bunch of hacks we add in `kani-driver` to support `--dry-run`. 
2. Usually when I run `--dry-run` is because I want to reproduce a similar sequence of commands with Kani, since `--dry-run` uses mock data, it forces me to edit the lines I get from Kani.
3. I also end up having to run Kani twice. `kani --dry-run` followed by a `kani --keep-temps`. This would make it easier to do one pass of commands only.

celinval:
### Description of changes: 

I also ensured that `--verbose` was always printing the commands `kani-driver` invokes and I added a `Running:` prefix to the command being printed so they are grepable.

### Resolved issues:

Resolves #1936

### Call-outs:

I kept the `--dry-run` and turned into an error for now as a reminder to people.

### Testing:

* Is this a refactor change?

### Checklist


kani-driver/src/args.rs
```diff
@@ -448,6 +446,13 @@ impl KaniArgs {
             ));
         }
 
+        if self.dry_run {
+            return Err(Error::raw(
+                ErrorKind::ValueValidation,
+                "We no longer support `--dry-run`. Use --verbose instead.",
```
zhassan-aws, line 452: Suggest saying:
```
The `--dry-run` option has been deprecated. Use `--verbose` instead.
```

zhassan-aws (approved):
Looks good. Thanks!

It might be useful to add a Kani-specific prefix to the "Running:" line, e.g.
```
[Kani] Running: ...
```
but the current form is fine too.

Also, might want to say "Executing command:" intead of "Running".

kani-driver/src/args.rs
celinval: Usually deprecated  means that it still works and it will be removed which isn't the case here.

celinval:
> Looks good. Thanks!
> 
> It might be useful to add a Kani-specific prefix to the "Running:" line, e.g.
> 
> ```
> [Kani] Running: ...
> ```
> 
> but the current form is fine too.
> 
> Also, might want to say "Executing command:" intead of "Running".

Sure, I can add a tag. I do prefer `Running` though since it is shorter and it matches cargo output. Is that OK?

kani-driver/src/args.rs
zhassan-aws: How about "obsolete"?

zhassan-aws:
> Sure, I can add a tag. I do prefer `Running` though since it is shorter and it matches cargo output. Is that OK?

Sure.

kani-driver/src/args.rs
celinval: Sure

PR #1947, opened 2022-11-28
Fix `kani-driver` error formatting

aaronbembenek-aws:
### Description of changes: 

With the update of `kani-driver` to use `clap` v4 (#1901), we have some regressions on formatting of error messages. Currently, the command `kani input.rs --enable-unstable --cbmc-args --function dummy` results in this message, with no newline, and no color highlighting (before it had the newline and `error` was colored red):
```
error: Invalid flag: --function should be provided to Kani directly, not via --cbmc-args.
```
With this PR, the command produces this output:
```
error: Invalid flag: --function should be provided to Kani directly, not via --cbmc-args.

Usage: kani [OPTIONS] <INPUT>

For more information try '--help'
```
There is a newline and `error` is once again colored red. It also gives the correct usage message if `cargo kani` has been used instead of `kani`.

### Testing:

* How is this change tested? Existing tests (although I'm not sure how much test coverage there is for error formatting)

* Is this a refactor change? Not exactly

### Checklist


kani-driver/src/main.rs
```diff
@@ -45,11 +46,12 @@ fn cargokani_main(input_args: Vec<OsString>) -> Result<()> {
     if matches!(args.command, Some(CargoKaniSubcommand::Assess)) || ctx.args.assess {
         // --assess requires --enable-unstable, but the subcommand needs manual checking
         if !ctx.args.enable_unstable {
-            clap::Error::raw(
-                clap::error::ErrorKind::MissingRequiredArgument,
-                "Assess is unstable and requires 'cargo kani --enable-unstable assess'".to_string(),
-            )
-            .exit()
+            args::CargoKaniArgs::command()
```
celinval, line 49: I'm assuming this won't print the command context, right?

celinval (approved):

kani-driver/src/main.rs
aaronbembenek-aws: This is what it prints:
```
error: Assess is unstable and requires 'cargo kani --enable-unstable assess'

Usage: cargo-kani [OPTIONS]

For more information try '--help'
```

Issue #661, opened 2021-11-22
RMC should use fully qualified name for harness selection and it should not rely on `no_mangle` attribute


celinval:
Users should be able to use RMC to target functions with the same name in different modules. For example, if user wants to verify the following code:

```rust
mod first {
    fn check() {
        // some check
    }
}

mod second {
    fn check() {
        // check for second module
    }
}

pub fn check() {
  // full check
}
```

They should be able to provide which function they want to check. I.e:
```
rmc multiple.rs --function [check | first::check | second::check]
```

Note that users cannot use `#[no_mangle]` for the check functions, since this would cause a name conflict. The compilation would fail with the following error:
```
error: symbol `check` is already defined
  --> multiple.rs:26:1
   |
26 | fn check() {
   | ^^^^^^^^^^^^^^

error: aborting due to previous error

```

tedinski:
1. Proof annotations should solve this, as we get a mapping in `rmc-metadata.json` from pretty names to mangled names. So when `cargo-rmc` exists and is looking at this mapping, we can easily find the function to use.
2. But right now proof annotations are using `no_mangle` as a hack in order to force codegen of the function. We need an alternative solution to this problem.

So far the only options I've come up with are:

1. Instead of putting `#[no_mangle]` in the case of `cfg(rmc)`, we could put `#[test]` and always have rmc build in "test mode" (i.e. `cargo test --no-run` not `cargo build`). One thing 'test' does is make the function get codegen'd, and that's all we need from it. This does have the advantage of being completely reliable.
2. We could see if `#[export_name = concat!(module_path!(), "::", #test_name))]` works and so generate fully qualified names ourselves. Might not be as reliable as Rust's native name mangling though.

tedinski:
Current status:

* We can run proof harnesses by short and full name.
* We **have not yet** removed `no_mangle` as we haven't picked a way to ensure Rust codegen's each function.

celinval:
This should work with the new MIR Linker. But we still need to remove the legacy one in order to remove it.

YoshikiTakashima:
@celinval, @tedinski this is blocking a 10/29 when evaluating proptest project. What is the remaining todos?

tedinski:
I'm not sure. I think it might work with `no_mangle` removed now, and the only reason we haven't is the lingering need to support the legacy linker...

YoshikiTakashima:
[...]

PR #1955, opened 2022-12-02
[DRAFT, Needs Discussion] Use fully qualified paths to avoid harness name clash.

YoshikiTakashima:
### Description of changes: 

Changes `kani_macros` to export harnesses with fully qualified paths, not name of function. Adjusts concrete-playback since it relies on this path.

### Resolved issues:

Partial resolution of #661 (full path, still has `#[no_mangle])

Optional #661 

### Call-outs:

- I don't think the Kani team agreed on which of the 2 options are preferable. That needs to happen before merge.

### Testing:

* How is this change tested? [tests/kani/FunctionCall_MultiModuleSameHarnessName/main.rs](https://github.com/model-checking/kani/compare/main...YoshikiTakashima:kani:661-different-harness-option2?expand=1#diff-8216cfac3e65171bbd6f681c590d1f61947aedf1c385c85d4a6f7e5ca43281ec)

### Checklist


tests/kani/FunctionCall_MultiModuleSameHarnessName/main.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
```
celinval, line 2: Can you please add a comment on what this is testing?

celinval (commented):
We should eventually get rid of the `[no_mangle]` stuff. I think only the legacy linker depends on it now. That said, I think this is a small enough change that is an improvement. We can remove the export name once we get rid of `[no_mangle]`.

My only ask is to add a similar test for concrete playback. Thanks

tests/kani/FunctionCall_MultiModuleSameHarnessName/main.rs
YoshikiTakashima: @celinval Resolved in https://github.com/model-checking/kani/pull/1955/commits/87d1214108935310626f5b6bea1e5cb169c32acf

tests/ui/concrete-playback/mult-harnesses/expected
```diff
@@ -17,11 +17,11 @@ VERIFICATION:- FAILED
...
 ```
 #[test]
-fn kani_concrete_playback_harness1
+fn kani_concrete_playback_first_harness
     let concrete_vals: Vec<Vec<u8>> = vec![
         // 0
         vec![0]
     ];
-    kani::concrete_playback_run(concrete_vals, harness1);
+    kani::concrete_playback_run(concrete_vals, first::harness);
```
celinval, line 25: Have you tried running this test? As far as I remember, the test is added to the same module has the harness, so I don't think the module should be added here.

YoshikiTakashima: Fixed with https://github.com/model-checking/kani/pull/1955/commits/bcbd79e137b0bdb8f72b3469e0f1a05560817153

Although the CI passed before this fix, and maybe we need to adjust the CI to not pass.

YoshikiTakashima:
@celinval Modulo any major bugs, I think this should fix the comments. 

Also, I can't press the merge button, so please merge when you think it's ready.

tests/ui/concrete-playback/mult-harnesses/expected
celinval: I totally agree. We need to add regression tests that actually runs the tests generated by  the concrete playback

celinval (approved):

Issue #1758, opened 2022-10-06
`kani-driver` needs its own notion of Kani Metadata


tedinski:
We handle metadata in three places:

1. In `kani-compiler`, we accumulate metadata in the `GotoCtx`. We then translate this into:
2. In `kani_metadata`, we record the "file format" for Kani metadata. We then read this into:
3. In `kani-driver`, we then apply a "merge" operation to produce one overall metadata structure.

This last step is an inappropriate hack. It "worked" when metadata consisted of nothing but a flat array of proof harnesses. But now we want richer structure allowing us to aggregate this information better.

Further, that was (maybe) never even the right behavior because it meant we accumulated together all proof harnesses from dependencies as well, when we really only care about the "main crate" we're building.

The solution is to define our own metadata data types in `kani-driver` and compute them from the flat serialization format from `kani_metadata`. (Though we might even want to put these types in `kani_metadata`, since they'd be generally usable functions for consuming this information...)

celinval:
I have two questions about the description:

1. Are you talking about the entire metadata or just the harness metadata?
2. What do you mean by "we record the file format"?

> Further, that was (maybe) never even the right behavior because it meant we accumulated together all proof harnesses from dependencies as well, when we really only care about the "main crate" we're building.

Yes, I don't think merging was the correct way here. I would strongly prefer if we don't have different notions of metadata format. This will likely cause some confusion.

celinval:
BTW, `#1` might not be clear, but I'm thinking about the vtable metadata. Note that with MIR Linker, there will be only one relevant metadata file.

tedinski:
Ooh, good point, mir linker will change how metadata is collected.

So originally I was just thinking this was an internal problem for `kani-driver`, but that might mean there are implications for `kani_metadata` too.

PR #1956, opened 2022-12-03
Introduce `project` and stop merging all files generated by the compiler

### LINKED ISSUE #1448: cargo kani --tests fails on crates with integration tests

### LINKED ISSUE #1910: Standalone `kani` on file without proof harnesses crashes
```
$ kani src/main.rs 
Error: No such file or directory (os error 2)
```

There are two fixes needed here:

1. That error message should be improved. We're not reporting what file was missing.
2. We should (in this situation) be reporting the same "No proof harnesses" result as `cargo kani`. I'm not sure why `kani` isn't producing the `kani-metadata.json` file that was expected (and that's what there's "no such file" for)

### LINKED ISSUE #1974: Kani gives wrong result if two harnesses have the same name
Here are the steps to reproduce them:

1. Create two crates, and add a harness to each of them with the same name, e.g.:

```
cargo init --lib lib1
cargo init --lib lib2
```

Where `lib1/src/lib.rs` include:

```rust
#[kani::proof]
fn check_lib() {
    assert!(1 == 1, "This should succeed");
}

```

And `lib2/src/lib.rs` include:

```rust
#[kani::proof]
fn check_lib() {
    assert_eq!(1 + 1, 1, "This should fail");
}
```

2. Create a workspace that includes both crates:

```toml
[workspace]
members = [
  "lib1",
  "lib2",
]
```

3. Run `cargo kani` in the workspace.

with Kani version: 0.16.0

I expected to see this happen: Kani should report that one harness failed and one succeeded.

Instead, this happened: Kani claims that both harnesses succeed.

```
$ cargo kani --output-format=terse
   Compiling lib1 v0.1.0 (/tmp/ws-name-conflict/lib1)
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
   Compiling lib2 v0.1.0 (/tmp/ws-name-conflict/lib2)
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
Checking harness check_lib...

VERIFICATION RESULT:
 ** 0 of 1 failed

VERIFICATION:- SUCCESSFUL
Verification Time: 0.01958095s

Checking harness check_lib...

VERIFICATION RESULT:
 ** 0 of 1 failed

VERIFICATION:- SUCCESSFUL
Verification Time: 0.019129628s

```

celinval:
### Description of changes: 

This PR is a pre-req for getting a per-harness reachability algorithm. First we need to stop treating all `goto` generated by the compiler as the same model which is what this PR does.

For that, I introduce the concept of a project, which is basically a container of artifacts produced by the build system. The goal is to abstracts the verification code from the build system we pick. It also provide methods to get artifacts relevant to a harness. Today this method will return the artifact related to the crate where it is contained, but in the future it can be easily modified to be the harness specific goto model.

As part of this refactoring, this fixes the issue with conflicting harnesses name. I also believe this fixes #1448 and it fixes #1910.

### Resolved issues:

Fixes #1974 
Fixes #1910

### Call-outs:

- I still use glob in two scenarios: `--legacy-linker` and `--function`. Fixing these two cases will take more effort, and I'm not sure it's worth it at this point. If people feel strong about them, I think they should be fixed in a separate PR since it will require more changes.
- I am still planning to add more comments and tests for the issues fixed, but I thought I could get some early feedback.
- I also believe that this change will make it very easy to enable `assess` for standalone Kani. But it is also out of scope right now.

### Testing:

* How is this change tested? New test + old test

* Is this a refactor change?

### Checklist

edit: Add #1974 as part of the fixes and remove #1448 since it was fixed by the MIR Linker.


kani-driver/src/project.rs
```diff
@@ -0,0 +1,282 @@
...
+
+impl Deref for Artifact {
+    type Target = Path;
+    fn deref(&self) -> &Self::Target {
+        &self.path
+    }
+}
+
+impl Artifact {
+    /// Create a new existing artifact.
```
zhassan-aws, line 104: "new" and "existing" sound a bit conflicting. Perhaps we can say "Create an artifact object for an artifact whose file exists on disk"?

tedinski (commented):
There is a lot going on here and I'm not sure how to review it. There seems to be lots of behavior changes, some potentially unintended? (e.g. --function?, are the files generated changed? I can't tell. Wait, they must be, that one test changed. Do restrictions still work? There's changes there too and I can't figure out if they're significant. Does this change the behavior of the legacy linker in subtle ways?)

This seems to try to factor out common code from the various entry points, which sure, but the result is just big and complicated and I can't figure it out. I can't tell if that's just a bad abstraction (there's a REASON I went with a small amount of copy/paste code between the entry points) or if that's part of the changes needed to make the multiple artifacts work, or both.

As for path forward, I guess I'd at least like to see:
1. Some thinking through of the behavior changes. This is I/O heavy code, so isn't (shouldn't won't) be something we can be confident in just by running the regression suite. I'm particularly curious about the deliberate deletion of code saying "hey we have to handle `--function` when it's not a harness" seemingly without concern?
2. I think at a minimum if `Project` is going to be a useful abstraction is needs to be cleaner than it is right now. I don't know how right now. This PR overflowed my ability to understand what was changing

kani-driver/src/project.rs
```diff
@@ -0,0 +1,282 @@
...
+    if path.exists() { Artifact::try_new(&path, typ).ok() } else { None }
+}
+
+fn dump_metadata(metadata: &KaniMetadata, path: &Path) {
+    let out_file = File::create(path).unwrap();
+    let writer = BufWriter::new(out_file);
+    serde_json::to_writer_pretty(writer, &metadata).unwrap();
+}
+
+pub fn cargo_project(session: &KaniSession) -> Result<Project> {
```
zhassan-aws, line 126: Please add documentation.

tests/cargo-kani/ws-name-conflict/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Test that Kani can handle harnesses in different crates with the same name
+
+[workspace]
```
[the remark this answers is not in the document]
celinval: Yes, this failed before in an unsound way. When we merge all the `goto` into one, Kani still know it has two harnesses, but their body become the same.

kani-driver/src/project.rs
```diff
@@ -0,0 +1,282 @@
...
+pub struct Project {
+    /// Each target crate metadata.
+    pub metadata: Vec<KaniMetadata>,
+    /// The directory where all outputs should be directed to. This path represents the canonical
+    /// version of outdir.
+    pub outdir: PathBuf,
+    /// The collection of artifacts kept as part of this project.
+    artifacts: Vec<Artifact>,
+    /// A flag that allow us to provide a consistent behavior for `--legacy-linker`.
+    /// If this flag is true, there should only be up to one artifact of a
```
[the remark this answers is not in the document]
celinval: sorry, let me finish that

kani-driver/src/project.rs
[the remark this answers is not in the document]
celinval: My main reason was simplicity and trying really hard to scope down this PR. I wanted to try to avoid changing `call_cargo.rs` because that code itself needs some refactoring (and this PR was already getting too big). So this code does a simple mapping between what `call_cargo.rs` generate.

FYI, the refactoring I'm talking about is to be able to get rid of cleaning the build artifacts.

tests/cargo-kani/ws-name-conflict/lib1/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Different harnesses with the same name. This one should pass. The simbling crate should fail.
```
zhassan-aws, line 4:
```suggestion
//! Different harnesses with the same name. This one should pass. The sibling crate should fail.
```

tests/cargo-kani/ws-name-conflict/lib2/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Different harnesses with the same name. This one should fail. The simbling crate should pass.
```
zhassan-aws, line 4:
```suggestion
//! Different harnesses with the same name. This one should fail. The sibling crate should pass.
```

kani-driver/src/metadata.rs
```diff
@@ -108,27 +108,17 @@ pub fn merge_kani_metadata(files: Vec<KaniMetadata>) -> KaniMetadata {
...
+        all_harnesses: &[&HarnessMetadata],
+    ) -> Result<Vec<HarnessMetadata>> {
+        if let Some(name) = self.args.harness.clone().or(self.args.function.clone()) {
             // Linear search, since this is only ever called once
-            let harness = find_proof_harness(name, &metadata.proof_harnesses)?;
+            let harness = find_proof_harness(&name, all_harnesses)?;
             return Ok(vec![harness.clone()]);
         }
-        Ok(metadata.proof_harnesses.clone())
+        Ok(all_harnesses.iter().cloned().cloned().collect())
```
zhassan-aws, line 121: Double `cloned`.

kani-driver/src/main.rs
```diff
@@ -25,109 +29,60 @@ mod cbmc_property_renderer;
...
-        linked_obj: &linked_obj,
-        report_base: &report_base,
-        symtabs: &outputs.symtabs,
-        retain_specialized_harnesses: true,
-    };
-
-    let results = runner.check_all_harnesses(&harnesses)?;
-
-    ctx.print_final_summary(&results)
+    let project = project::cargo_project(&session)?;
```
zhassan-aws, line 61: Very neat!

kani-driver/src/project.rs
zhassan-aws: Also, should we call this `build_project` or `compile_project`?

zhassan-aws (commented):
A few comments from a cursory look. Will take a more thorough look today.

kani-driver/src/metadata.rs
```diff
@@ -108,27 +108,17 @@ pub fn merge_kani_metadata(files: Vec<KaniMetadata>) -> KaniMetadata {
...
-        }
-        if let Some(name) = &self.args.harness {
+    pub fn determine_targets(
+        &self,
+        all_harnesses: &[&HarnessMetadata],
+    ) -> Result<Vec<HarnessMetadata>> {
+        if let Some(name) = self.args.harness.clone().or(self.args.function.clone()) {
             // Linear search, since this is only ever called once
-            let harness = find_proof_harness(name, &metadata.proof_harnesses)?;
+            let harness = find_proof_harness(&name, all_harnesses)?;
```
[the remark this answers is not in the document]
zhassan-aws: Do we still need `--function`?

kani-driver/src/project.rs
```diff
@@ -0,0 +1,282 @@
...
+            })
+            .collect()
+    }
+
+    /// Return the matching artifact for the given harness.
+    ///
+    /// If the harness has information about the goto_file we can use that to find the exact file.
+    /// For cases where there is no goto_file, we just assume that everything has been linked
+    /// together. I.e.: There should only be one artifact of the given type.
+    pub fn get_harness_artifact(
```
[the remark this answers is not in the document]
celinval: No, I don't think so. The HarnessMetadata should really just have properties of the harness. The fact that the harness metadata has information about the `goto_file` is also not great and it was a shortcut to get some of this work done.

I also don't think `kani-driver` should really be modifying that object. The metadata is a communication mechanism that is meant to go one way, the compiler writes them and the driver reads them. I believe #1758 kind of touches that point.

I really dislike that we create mock harnesses metadata, but this is a problem for later.

The one thing I don't like about this new code is that we are passing `HarnessMetadata` everywhere. I would like to eventually think about a better mechanism to index a harness or a better structure to pass around.

tests/cargo-kani/ws-name-conflict/Cargo.toml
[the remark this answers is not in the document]
celinval: https://github.com/model-checking/kani/issues/1974

kani-driver/src/project.rs
celinval: Unfortunately, we still have a branch at the beginning of main to see if something is `cargo kani` or `kani`, so I kept their names separate. I will try to add symmetry by adding a `standalone_project` that will replace the builder.

kani-driver/src/metadata.rs
celinval: oops

kani-driver/src/metadata.rs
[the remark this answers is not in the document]
celinval: We use `--function` for the bookrunner.

celinval:
> There is a lot going on here and I'm not sure how to review it. There seems to be lots of behavior changes, some potentially unintended? (e.g. --function?, are the files generated changed? I can't tell. Wait, they must be, that one test changed. Do restrictions still work? There's changes there too and I can't figure out if they're significant. Does this change the behavior of the legacy linker in subtle ways?)
> 
> This seems to try to factor out common code from the various entry points, which sure, but the result is just big and complicated and I can't figure it out. I can't tell if that's just a bad abstraction (there's a REASON I went with a small amount of copy/paste code between the entry points) or if that's part of the changes needed to make the multiple artifacts work, or both.
> 
> As for path forward, I guess I'd at least like to see:
> 
>     1. Some thinking through of the behavior changes. This is I/O heavy code, so isn't (shouldn't won't) be something we can be confident in just by running the regression suite. I'm particularly curious about the deliberate deletion of code saying "hey we have to handle `--function` when it's not a harness" seemingly without concern?
> 
>     2. I think at a minimum if `Project` is going to be a useful abstraction is needs to be cleaner than it is right now. I don't know how right now. This PR overflowed my ability to understand what was changing

I'm sorry this PR is bigger than I would like it to be too. I tried really hard to scope it down, but in some cases just didn't make sense to replicate a hack or purposely inject them in the new code to keep the behavior exactly as it was. It also didn't make sense to create an intermediate fix as a separate PR since the new code was going to fix them by design.

I believe the *only* behavioral change in this PR is that I disabled `--gen-c` for `--function` and `--legacy-linker`. This was to simplify the code since all these options are for internal usage or on track to be deleted. I could pull out the changes to the `args.rs` file if that helps to a separate PR.

All the other behavioral changes, they are actually fixes to inconsistent / incorrect behaviors. They are the following:

1. Kani is now sound when two or more targets contains harnesses or no mangled functions that have the same name (#1974). Note that this is not the case for `--legacy-linker` and `--function` options.
2. `kani` now respects the `--target-dir` argument provided by the customer.
3. `kani` no longer crash when the crate has no harness (#1910).

Regarding the deliberate deletion of "hey we have to handle `--function` when it's not a harness". It was replaced by, "hey, I don't need to care at this point whether the information is in the metadata files". One of the goals here is to have separation of concerns, where the project is built first, which includes linking. Then we verify the code. The code that determines which harness to verify, shouldn't be modifying the project metadata.

I'm happy to talk offline about a better abstraction for Project. I also agree that it could be better, but anything more sophisticated would require even more refactoring. I tried to keep it minimal, today it has 2 public members and 2 public functions.

kani-driver/src/args.rs
```diff
@@ -454,6 +478,14 @@ impl KaniArgs {
                 "The `--dry-run` option is obsolete. Use --verbose instead.",
             ));
         }
+        if let Some(out_dir) = &self.target_dir {
+            if out_dir.exists() && !out_dir.is_dir() {
+                return Err(Error::raw(
+                    ErrorKind::InvalidValue,
+                    "Invalid argument: `--target-dir` argument `{}` is not a directory",
```
zhassan-aws, line 485: The argument (`out_dir`) is missing from the error message.

celinval: And I'm missing the `format` call. That's why I need to add tests here.

kani-driver/src/args.rs
```diff
@@ -72,7 +72,7 @@ pub struct KaniArgs {
         value_enum
     )]
     pub concrete_playback: Option<ConcretePlaybackMode>,
-    /// Keep temporary files generated throughout Kani process
+    /// Keep temporary files generated throughout Kani process. This is a no-op for `cargo-kani`.
```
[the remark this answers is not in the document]
celinval: I updated the comment. Let me know if that makes sense.

kani-driver/src/project.rs
```diff
@@ -0,0 +1,282 @@
...
+                BUILD_ARTIFACTS.iter().filter_map(|typ| cargo_artifact(&canonical_metafile, *typ)),
+            );
+            debug!(?crate_name, ?crate_metadata, "cargo_project");
+            metadata.push(crate_metadata);
+        }
+        Ok(Project { outdir, artifacts, metadata, merged_artifacts: false })
+    }
+}
+
+pub struct StandaloneProjectBuilder<'a> {
```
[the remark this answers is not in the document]
celinval: I wrapped the builder into a `standalone_project()` function so the API is symmetric for `cargo kani` and `kani`.

kani-driver/src/args.rs
```diff
@@ -371,11 +372,34 @@ impl CheckArgs {
 impl StandaloneArgs {
     pub fn validate(&self) {
         self.common_opts.validate::<Self>();
+        self.valid_input().unwrap();
```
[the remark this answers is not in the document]
celinval: Yes, you were correct. This was crashing. I fixed it. unfortunately I cannot easily add a e2e test because compiletest already provides a valid input. I added a unit test for the `valid_input()` function.

kani-driver/src/args.rs
celinval: But I can't. :(

zhassan-aws: Are you still planning on adding a `format` call?

kani-driver/src/call_goto_instrument.rs
```diff
@@ -50,7 +57,8 @@ impl KaniSession {
             }
 
             let c_demangled = alter_extension(output, "demangled.c");
-            self.demangle_c(symtabs, &c_outfile, &c_demangled)?;
+            let symtabs = project.get_harness_artifact(&harness, ArtifactType::SymTab).unwrap();
```
zhassan-aws, line 60: Should this be a singular `symtab`?

kani-driver/src/project.rs
```diff
@@ -0,0 +1,315 @@
...
+use std::path::{Path, PathBuf};
+use tracing::{debug, trace};
+
+/// This structure represent the project information relevant for verification.
+/// A `Project` contains information about all crates under verification, as well as all
+/// artifacts relevant for verification.
+///
+/// For one specific harness, there should be up to one artifact of each type. I.e., artifacts of
+/// the same type are linked as part of creating the project. Except for `symtab` files used for
+/// demangling `C` when on legacy linker mode.
```
zhassan-aws, line 42: Does this comment still apply given that this PR makes `--legacy-linker` conflict with `--gen-c`?

zhassan-aws (approved):

kani-driver/src/args.rs
celinval: I can't believe I forgot the format. I must be going crazy. Sorry about that. Let me fix this

celinval: Ah, I did add format somewhere else. I guess I just missed this one. Thanks for catching this.

PR #1959, opened 2022-12-05
Add more info about debugging with rustc

celinval:
### Description of changes: 
Add information about emitting the MIR and printing the type layout.

### Resolved issues:

N/A

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/docs/src/rustc-hacks.md
+++ b/docs/src/rustc-hacks.md
@@ -201,9 +201,37 @@ rustup override set custom-toolchain
 cargo clean
 cargo build-dev
 ```
+# Rust compiler utilities to debug `kani-compiler`
 
-# Enable `rustc` logs
+## Enable `rustc` logs
 
 In order to enable logs, you can just define the `RUSTC_LOG` variable, as documented here: <https://rustc-dev-guide.rust-lang.org/tracing.html>.
 
-Note that depending on the level of logs you would like to enable, you'll need to build your own version of `rustc` as described above.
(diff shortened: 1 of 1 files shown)


docs/src/rustc-hacks.md
```diff
@@ -201,9 +201,37 @@ rustup override set custom-toolchain
...
+
+## Inspecting the MIR
+
+You can easily visualize the MIR that is used as an input to code generation by setting the rust flag `-emit mir`. I.e.:
+
+```
+RUSTFLAGS=--emit=mir kani test.rs
+```
+
+The compiler will generate a few files, but I recommend looking at the files that have the following suffix: `kani.mir`. Those files will include the entire MIR collected by our reachability analysis. It will include functions from all dependencies, including `std` library. One limitation is that we dump one copy of each specialization of the MIR function, even though the MIR body itself doesn't change.
```
adpaco-aws, line 237:
```suggestion
The compiler will generate a few files, but we recommend looking at the files that have the following suffix: `kani.mir`.
Those files will include the entire MIR collected by our reachability analysis.
It will include functions from all dependencies, including the `std` library.
One limitation is that we dump one copy of each specialization of the MIR function, even though the MIR body itself doesn't change.
```

adpaco-aws (approved):
Thanks, just leaving a few suggestions. Feel free to discard them if you don't agree.

Issue #1960, opened 2022-12-05
`simd_shuffle*` doesn't check that `indexes` contains constant values which are within expected bounds


adpaco-aws:
The LLVM backend [checks](https://github.com/rust-lang/rust/blob/c1a859b25a95999ba276075bbd8e284ea675b41a/compiler/rustc_codegen_llvm/src/intrinsic.rs#L1001-L1021) that all indexes supplied via the 3rd argument to `simd_shuffle` are:
 1. Constant values.
 2. Within bounds for the `simd_shuffle` operation (i.e., 2x the length of one of the first arguments).

At present, we do not perform this check because we don't have access to the `Builder` type they use for constant evaluation. But we can probably do it some other way, or just use `cprover` types for it.

PR #1961, opened 2022-12-05
Add support for `simd_shuffle*`

adpaco-aws:
### Description of changes: 

* Restores support for the `simd_shuffleN` intrinsic.
* Adds support for the variable-length `simd_shuffle` intrinsic.
* Performs additional type-checking for all `simd_shuffle*` operations (we only miss checks for `indexes` in #1960).
* Updates documentation.

### Resolved issues:

Towards #1148 
Opens #1960

### Testing:

* How is this change tested? Adds new tests, and restores one "fixme" test - in addition to existing regression.

### Checklist


docs/src/rust-feature-support/intrinsics.md
```diff
@@ -270,6 +270,6 @@ Name | Support | Notes |
 `simd_rem`  | No | |
 `simd_shl`  | No | |
 `simd_shr`  | No | |
-`simd_shuffle*`  | No | |
+`simd_shuffle*`  | Yes | Doesn't check indexes at compile time ([#1960](https://github.com/model-checking/kani/issues/1960))|
```
celinval, line 273: This gets checked during the verification, right?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -370,17 +370,44 @@ impl<'tcx> GotocCtx<'tcx> {
...
-            // `parse` fails on empty, so comment that bit of code out.
-            // To re-enable this we'll need to investigate how size is computed now.
-            // let n: u64 = stripped.parse().unwrap();
-            return unstable_codegen!(self.codegen_intrinsic_simd_shuffle(
-                fargs,
-                p,
-                cbmc_ret_ty,
-                n
-            ));
+        if let Some(stripped) = intrinsic.strip_prefix("simd_shuffle") {
```
celinval, line 373: nit: can you move this to its own function? Maybe call that from inside the codegen intrinsic function.

I think this function is already pretty big, so i would prefer not adding more logic in here.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -370,17 +370,44 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            let n: u64 = if stripped.is_empty() {
+                // Make sure that this is an array, since only the
+                // length-suffixed version of `simd_shuffle` (e.g.,
+                // `simd_shuffle4`) is type-checked
+                match farg_types[2].kind() {
```
celinval, line 378: please assert that fargs.len() == 3

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1501,21 +1528,37 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    ///  2. The return type's subtype is equal to the vector's subtype (i.e.,
+    ///     the 1st argument). Both input vectors are guaranteed to be of the
+    ///     same type when they get here due to the `simd_shuffle` definition.
+    ///
+    /// In the case of some SIMD intrinsics, the backend is responsible for
+    /// performing this and similar checks, and erroring out if it proceeds.
+    ///
+    /// TODO: Check that `indexes` contains constant values which are within the
+    /// expected bounds. See
+    /// <https://github.com/model-checking/kani/issues/1960> for more details.
```
celinval, line 1546: Do you have a test that covers this case?

celinval (commented):
Just a few comments.

docs/src/rust-feature-support/intrinsics.md
adpaco-aws: Yes, it shows as an "index out-of-bounds" failure.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: Agreed, moved into a method (since we need to span errors). Thanks!

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: The check appeared where we codegen, I've moved it so it's done before all of this.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
adpaco-aws: We didn't, added one!

docs/src/rust-feature-support/intrinsics.md
celinval: So I would propose that we don't bother adding this issue here.

celinval (approved):

docs/src/rust-feature-support/intrinsics.md
adpaco-aws: That makes sense, just updated #1960 to make this clear. Removed the issue while resolving merge conflicts.

Issue #1954, opened 2022-12-01
Windows support


Timmmm:
It would be nice if Kani worked on Windows. Also you should probably say somewhere near the start of the Readme or book that it doesn't.

tedinski:
Great point about the docs. I thought it was more prominent... but it's only in the installation instructions. I'll fix that.

Windows support isn't something the Kani team is currently prioritizing, but we'd definitely be happy to merge support if someone out there is motivated to add it.

We're quite interested in seeing Kani used more, of course, so if there are big application domains out there that would like to use Kani but consider lack of windows support a blocker, we'd like to hear about it.

rahulku:
@Timmmm, can you please provide some more details on what you mean by "Windows support"? Kani can be used on Windows using WSL. Is there something platform specific you are looking for?

Timmmm:
It's a reasonable workaround given that Kani doesn't support Windows but I meant native support.

codeOverFlow:
What about crates using `windows` dependency or function with `cfg(windows`, can it still be verified using WSL ?

tautschnig:
While you may be able to run Kani, I'd suspect you'd get limited verification value as we do not provide models of various Windows API functions.

PR #1969, opened 2022-12-06
Doc improvements: supported platforms and versions

tedinski:
### Description of changes: 

1. Very compactly mention which OSes (linux and mac, not windows) are supported in our readme.
2. Add a quick note about Rust 1.58+. I think we can remove this eventually from the README because it's pretty old, but temporarily we should note it. And it should stay noted in the longer-form doc.
3. The github action bit stated the wrong action name and was waaaaay to prominent in the read me. I bumped it down and fixed the name
4. minor tweaks

### Resolved issues:

This was motivated by an issue that pointed out we don't clearly say we don't support windows, so now at least the readme says "linux or mac"

* #1954 

### Call-outs:

* Re-reading the README carefully, I think we need to re-do our "elevator pitch" for Kani. I think the blurb over-emphasized unsafe code.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,9 @@ Kani verifies:
  * The absence of panics (e.g., `unwrap()` on `None` values)
  * The absence of some types of unexpected behavior (e.g., arithmetic overflows)
 
-## GitHub Action
-
-Use Kani in your CI with `model-checking/kani@VERSION`. See the
-[GitHub Action section in the Kani
-book](https://model-checking.github.io/kani/install-github-ci.html)
-for details.
-
 ## Installation
 
-To install the latest version of Kani, run:
+To install the latest version of Kani ([Rust 1.58+; Linux or Mac](https://model-checking.github.io/kani/install-guide.html)), run:
 
 ```bash
 cargo install --locked kani-verifier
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -9,16 +9,9 @@ Kani verifies:
...
-
-Use Kani in your CI with `model-checking/kani@VERSION`. See the
-[GitHub Action section in the Kani
-book](https://model-checking.github.io/kani/install-github-ci.html)
-for details.
-
 ## Installation
 
-To install the latest version of Kani, run:
+To install the latest version of Kani ([Rust 1.58+; Linux or Mac](https://model-checking.github.io/kani/install-guide.html)), run:
```
danielsn, line 14: Is it all linux? Or just tested on ubuntu?

tedinski: I don't think we've heard of any compatibility problems from others. We have tested on AL2, NixOS at least. I believe I also tested on CentOS 7.

But I'd definitely also state anything more specific on the install page, I think the README needs to stay really concise.

danielsn (approved):

PR #1730, opened 2022-09-29
Fix `cargo kani --debug` and remove debug logs from `--verbose`

celinval:
### Description of changes: 

- Fix #1348: Fix `cargo kani --debug` by redirecting kani-compiler logs to the STDERR so it doesn't conflict with cargo's output expectations.
- Fix #1631: Remove `kani-compiler` logs from the output of `--verbose`.

### Resolved issues:

Fix #1631
Fix #1348

### Related RFC:

N/A

### Call-outs:

These were tiny fixes so I combined them in one PR.

### Testing:

* How is this change tested? N/A

* Is this a refactor change?

### Checklist
- [ ] Each commit message has a non-empty body, explaining why the change was made
- [ ] Methods or procedures are documented
- [ ] Regression or unit tests are included, or existing tests cover the modified code
- [ ] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #1972, opened 2022-12-06
Add some performance instrumentation to our code

celinval:
### Description of changes: 

For runs using `--verbose`, Kani will now print the time spent in some parts of the code. This include calls to all CProver utilities and a few other places in the compiler.

### Resolved issues:

N/A

### Call-outs:

Note that the time measured in `run_cbmc` includes parsing the results. This is similar to the "Verification Time" we print, but it can be misleading when we are using `debug` mode, since it can more than double the actual time we report to be the verification time. We should probably improve that, but it's out of the scope of this PR. The overhead is also small when running a release build.

### Testing:

* How is this change tested? 

* Is this a refactor change?

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
@@ -83,7 +83,7 @@ impl CodegenBackend for GotocCodegenBackend {
         check_options(tcx.sess, need_metadata_module);
         check_crate_items(&gcx);
 
-        let items = with_timer(|| collect_codegen_items(&gcx), "reachability analysis");
+        let items = with_timer(|| collect_codegen_items(&gcx), "codegen reachability analysis");
         if items.is_empty() {
             // There's nothing to do.
             return codegen_results(tcx, rustc_metadata, gcx.symbol_table.machine_model());
@@ -504,6 +504,8 @@ where
     }
 }
 
+/// Execute the provided function and measure the clock time it took for its execution.
+/// Log the time with the given description.
 pub fn with_timer<T, F>(func: F, description: &str) -> T
 where
     F: FnOnce() -> T,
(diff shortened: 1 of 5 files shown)


kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -82,58 +83,66 @@ impl CodegenBackend for GotocCodegenBackend {
         check_options(tcx.sess, need_metadata_module);
         check_crate_items(&gcx);
 
-        let items = collect_codegen_items(&gcx);
+        let items = with_timer(|| collect_codegen_items(&gcx), "reachability analysis");
```
zhassan-aws, line 86: Can we be more specific, e.g. "codegen items reachability analysis"?

kani-driver/src/session.rs
```diff
@@ -191,6 +203,19 @@ impl KaniSession {
 
         Ok(Some(process))
     }
+
+    pub fn with_timer<T, F>(&self, func: F, description: &str) -> T
```
zhassan-aws, line 207: Might be worth creating a "common" or "utils" crate at some point that includes this function (and similar ones) to avoid duplication.

zhassan-aws (approved):
Thanks for doing this!

I only had a few minor comments.

kani-driver/src/session.rs
celinval: I totally agree. Not sure it's worth it in this PR though

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
celinval: I changed it to be "codegen reachability analysis". Let me know if that's OK.

zhassan-aws: Sounds good.

kani-driver/src/call_single_file.rs
```diff
@@ -117,6 +117,9 @@ impl KaniSession {
 
         if self.args.debug {
             flags.push("--log-level=debug".into());
+        } else if self.args.verbose {
+            // Print the symtab command being invoked.
+            flags.push("--log-level=info".into());
```
tedinski, line 122: This was removed in https://github.com/model-checking/kani/pull/1730 is the underlying issue fixed there?

PR #1991, opened 2022-12-13
Fix `--use-local-bundle` argument parsing

tedinski:
### Description of changes: 

There was a bug where it was using `arg[3]` but in one case should have used `arg[4]`. Fixed, factored out the parsing step to its own function, then wrote tests for it, since clearly we can get this wrong. :)

### Testing:

* How is this change tested? new unit tests! woo

* Is this a refactor change?

### Checklist


celinval (approved):
Thanks for the bug fix.

On the long run, do you think it is still worth to manually parse the args instead of using a more stable library?

zhassan-aws (approved):

tedinski:
> On the long run, do you think it is still worth to manually parse the args instead of using a more stable library?

I think the threshold is 'if we have to care about more than 1 argument'. I think we still benefit a lot from having very few deps in this proxy binary.

tedinski:
For reference, 

```
clap = { version = "4.0.29", default-features = false, features = ["std"] }
```

gives

```
$ cargo build
   Compiling os_str_bytes v6.4.1
   Compiling bitflags v1.3.2
   Compiling clap_lex v0.3.0
   Compiling clap v4.0.29
```

as the only deps and adds 3.5 seconds to build time on my laptop. So yeah, as soon as there's any more complexity to this, probably switch to (non-derive-based) clap is not super high-cost.

Issue #1998, opened 2022-12-13
Take into account `use` statements when resolving stub paths


aaronbembenek-aws:
Currently, we do not account for `use` statements when trying to resolve paths appearing in `kani::stub` attributes. For example, this harness fails to compile:

```rust
fn foo() -> bool {
    false
}

fn bar() -> bool {
    true
}

mod my_mod {
    use super::*;

    #[kani::proof]
    #[kani::stub(foo, bar)]
    fn harness() {
        assert!(foo());
    }
}
```

We are not able to resolve either `foo` or `bar`. We should fix this, both to make easier to specify stubs, and to ensure that stubs aren't resolving to the wrong functions/methods.

Issue #1997, opened 2022-12-13
Resolve trait methods during stubbing


aaronbembenek-aws:
Right now, we do not support stubbing methods declared in traits. Consider this example:

```rust
trait A {
    fn foo(&self) -> u32;

    fn bar(&self) -> u32;
}

trait B {
    fn bar(&self) -> u32;
}

struct X {}

impl X {
    fn new() -> Self {
        Self {}
    }
}

impl A for X {
    fn foo(&self) -> u32 {
        100
    }

    fn bar(&self) -> u32 {
        200
    }
}

impl B for X {
    fn bar(&self) -> u32 {
        300
    }
}

#[kani::proof]
fn harness() {
    let x = X::new();
    assert_eq!(x.foo(), 1);
    assert_eq!(A::bar(&x), 2);
    assert_eq!(<X as B>::bar(&x), 3);
}
```

It is currently not possible to stub `X`'s implementation of `A::foo`, `A::bar`, or `B::bar`. To do so, we'd need to do two things.

First, we'd need to come up with a way to refer to trait methods in our `kani::stub` attributes. In `rustc`, these are stringified as `<X as B>::bar`, but paths in attribute arguments are not allowed to have spaces or symbols like `<` and `>` (they are [simple paths](https://doc.rust-lang.org/reference/paths.html)). We could accept string arguments or come up with some other convention, e.g.:

```rust
#[kani::stub("<X as B>::bar", ...)]
// or
#[kani::stub(X__as__B::bar), ...)]
```

Second, we'd need to improve our path resolution algorithm to also search through trait implementations.

adpaco-aws:
Adding the `T-User` label as a few of users are hitting this.

celinval:
What if we create something like:
```rust
#[kani::stub(trait_impl(Trait::function for Type), stub_funtion)]
```

I.e.: The example would look like:
```rust
#[kani::stub(trait_impl(A::bar for X), empty_debug)]
```

JustusAdam:
I think it makes sense to use rust's syntax, e.g. `<Type<'lifetimes, Generics> as Trait>::method` for consistency, the tricky part is how do we deal with lifetimes and monomorphization which is almost independent from syntax. It may be necessary at least to distinguish `'a` and `'static`. On the flip side if you have an instance `impl<T> Trait for Type<T>`, how does this get stubbed and how does it get resolved? There are a few options here and it is not clear to me which is preferable. 

[...]

PR #1999, opened 2022-12-13
Resolve paths in `kani::stub` attributes (ignoring `use` statements)

### LINKED ISSUE #1894: Improve performance of stub lookup
**Proposed change:**

We should improve the way we get a handle on a stub during the MIR-to-MIR transformation pass; this is essentially the task of resolving the name of the stub (a string) to a `DefId`. Unfortunately [there is not an established rustc API for doing this](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/Resolve.20string.20to.20DefId), so this will take some work/creativity (we might also be able to draw on what other tools like clippy do).

**Motivation:**

Right now, we are iterating over all `LocalDefId`'s to see if we can find one that matches the name of the supplied stub. This approach is inefficient, as the linear lookup happens for every stub.

aaronbembenek-aws:
### Description of changes: 

Currently, when we resolve paths in `kani::stub` attributes we do not take into account the module of the harness and we do not support the path qualifiers `crate`, `self`, `super`, and initial `::`. This PR improves our path resolution algorithm to take into account the module of the harness (so we resolve relative paths correctly) and adds support for those qualifiers.

It does not handle `use` statements (as in `use XXX;` or `use XXX as YYY;`); they will be addressed in a separate PR (tracked in #1998).

### Resolved issues:

Related to #1866 
Resolves #1894 

### Call-outs:

Previously, we represented a stub mapping (mapping original functions/methods to their replacements) as a `FxHashMap<String, String>`, and then did a linear lookup in the HIR during the MIR transformation to try to find a function/method with the same string representation. Since we are actually resolving paths now when processing the attributes (and constructing the stub mappings), this PR uses a `FxHashMap<DefPathHash, DefPathHash>` instead, which is more robust and removes the need for that linear lookup. A [`DefPathHash`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_hir/definitions/struct.DefPathHash.html) is stable across crate and compilation boundaries.

### Testing:

* How is this change tested? Existing tests and new tests

### Checklist


kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -0,0 +1,236 @@
...
+    if segments.is_empty() {
+        return Some(Path::new(
+            Base::LocalModule { id: current_module, may_be_external_path: false },
+            segments,
+        ));
+    }
+
+    // Resolve qualifiers `crate`, initial `::`, and `self`. The qualifier
+    // `self` may be followed be `super` (handled below).
+    let mut may_be_external_path = true;
```
celinval, line 83: For readability, I would recommend inspecting the first element and assigning the proper value here. Something like:
```
let may_be_external_path = !matches!(first, CRATE, SELF, SUPER);
```

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -119,35 +113,47 @@ fn check_compatibility<'a, 'tcx>(
...
 }
 
 /// Deserializes the stub mapping from the rustc argument value.
-fn deserialize_mapping(val: &str) -> FxHashMap<String, String> {
-    serde_json::from_str(val).unwrap()
+fn deserialize_mapping(tcx: TyCtxt, val: &str) -> FxHashMap<DefId, DefId> {
+    type Item = (u64, u64);
+    let item_to_def_id = |item: Item| -> DefId {
+        let hash = DefPathHash(Fingerprint::new(item.0, item.1));
+        tcx.def_path_hash_to_def_id(hash, &mut || panic!())
```
celinval, line 135: Maybe a friendly error would be nice here. Something on the terms that this is likely due to a Kani issue, please file a bug report.

celinval (commented):
I still have to go through the tests in this PR, but overall I think this is a great improvement to the user experience.

The logic here is fairly complicated, so I would like to ask you to add some `debug!()` and `trace!()` statements to help future development. Especially when things go wrong. There are a lot of places inside `resolve.rs` that you return `None`, which basically means that we fail resolution. It would be great to know which path we took to get there.

Another thing to consider is to give a more informative error when resolution fails. E.g.: If you cannot find a crate with a given name, say that. Something like, we could find crate with name "Blah". If the issue was with a module / item, the error message could

kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -0,0 +1,236 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! This module contains code for resolving strings to `DefId`s.
```
celinval, line 3: Can you please expand this description? What kind of strings are we talking about? Does this module cover any string that represents a path? Do we support fully qualified paths? Maybe adding a link to the path syntax might help too: https://doc.rust-lang.org/reference/paths.html.

Which brings me a question, do we support qualified paths?

celinval: Looking through the open issues kind of answers my last question about qualified paths: https://github.com/model-checking/kani/issues/1997

But I think it would be great if we can explicitly state that we are only handling simple paths and add a link to its definition: https://doc.rust-lang.org/reference/paths.html#simple-paths

kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// `may_be_external_path` is false) and should be resolved relative to the
+    /// module identified by `id`.
+    LocalModule { id: LocalDefId, may_be_external_path: bool },
+}
+
+/// A path consisting of a starting point and a bunch of segments.
+#[derive(Debug)]
+struct Path {
+    base: Base,
+    segments: Segments,
```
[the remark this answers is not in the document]
aaronbembenek-aws: Done

kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -0,0 +1,236 @@
...
+            Base::LocalModule { id: current_module, may_be_external_path: false },
+            segments,
+        ));
+    }
+
+    // Resolve qualifiers `crate`, initial `::`, and `self`. The qualifier
+    // `self` may be followed be `super` (handled below).
+    let mut may_be_external_path = true;
+    let first = segments[0].as_str();
+    if first == ROOT {
```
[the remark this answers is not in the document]
aaronbembenek-aws: Done

kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -0,0 +1,236 @@
...
+    fn new(base: Base, segments: Segments) -> Self {
+        Path { base, segments }
+    }
+}
+
+/// Takes a string representation of a path and turns it into a `Path` data
+/// structure, resolving qualifiers (like `crate`, etc.) along the way.
+fn to_path(tcx: TyCtxt, current_module: LocalDefId, name: &str) -> Option<Path> {
+    const CRATE: &str = "crate";
+    const ROOT: &str = "{{root}}";
```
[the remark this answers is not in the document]
aaronbembenek-aws: Exactly. I added a comment

kani-compiler/src/kani_middle/resolve.rs
aaronbembenek-aws: This is cleaner! Done

kani-compiler/src/kani_middle/stubbing/transform.rs
aaronbembenek-aws: I checked, and the panic is caught by our panic handler, so the user does receive a message to that effect

kani-compiler/src/kani_middle/resolve.rs
aaronbembenek-aws: I've updated some comments to make this clearer (and included the link for simple paths)

kani-compiler/src/kani_middle/stubbing/transform.rs
celinval: Thanks!

celinval (approved):
Thank you! BTW, great tests!

celinval:
BTW, I create #2013 to track improvements to the stubbing error formats.

aaronbembenek-aws:
Thanks for taking a look! I've added traces so that we now get debug output like this:

```
┐kani_compiler::kani_middle::resolve::path_resolution
├───0ms DEBUG kani_compiler::kani_middle::resolve Normalizing path `local_mod::pub_fn`
├───0ms DEBUG kani_compiler::kani_middle::resolve Resolving `local_mod::pub_fn` in local crate root
├───0ms DEBUG kani_compiler::kani_middle::resolve Resolving `pub_fn` in local module `local_mod`
├───0ms DEBUG kani_compiler::kani_middle::resolve Resolved `pub_fn` as a function in local module `local_mod`
┘
```

> BTW, I create #2013 to track improvements to the stubbing error formats.

Thanks! I think we should be able to support better errors by (as you suggest) changing the `Option`s to `Result`s where the error case records where resolution gets stuck.

PR #2011, opened 2022-12-16
Introduce a kani macro for checking coverage

### LINKED ISSUE #696: Create a `cover` statement to RMC

zhassan-aws:
### Description of changes: 

Introduce a `kani::cover` macro that can be used for checking whether certain conditions can occur without causing verification to fail.

### Resolved issues:

Resolves #696 

### Related RFC:

Optional #1905 

### Call-outs:

This PR does not yet introduce reachability checks for cover properties. Thus, both unreachable and reachable but unsatisfiable properties will be reported as "UNSATISFIABLE". I'm planning to introduce reachability checks in a follow-up PR because I didn't want to complicate this one too much.

### Testing:

* How is this change tested? Added a few tests.

### Checklist


tedinski:
Hmm, was this always going to be a macro, not just a function, and I missed that when I glanced at the RFC before?

Why a macro over just a function? Is it just the messages part? Is that essential?

I want to make sure the disadvantages of a macro (complexity, compile time, IDE autocomplete problems) are actually needed for this...

zhassan-aws:
> Hmm, was this always going to be a macro, not just a function, and I missed that when I glanced at the RFC before?

No, it was originally a function, but the discussion on the RFC led to introducing a macro. See https://github.com/model-checking/kani/pull/1906#discussion_r1023433867.

The function still exists though, and is public, so can be used directly.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -240,6 +240,12 @@ pub fn format_result(properties: &Vec<Property>, show_checks: bool) -> String {
     let mut number_tests_undetermined = 0;
```
celinval, line 240: I wonder if we should rename this to `number_checks_*` to contrast with `number_covers_*`.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -240,6 +240,12 @@ pub fn format_result(properties: &Vec<Property>, show_checks: bool) -> String {
...
     let mut failed_tests: Vec<&Property> = vec![];
 
+    // cover checks
```
celinval, line 243: Can you please add to the function documentation that we separate these two in the report?

library/kani/src/lib.rs
```diff
@@ -64,6 +64,17 @@ pub const fn assert(_cond: bool, _msg: &'static str) {
     }
 }
 
+/// Creates a cover property with the specified condition and message.
```
celinval, line 67: Can you please expand a bit on what a cover property is?

tests/expected/cover/cover-pass/expected
```diff
@@ -0,0 +1,11 @@
+Status: SATISFIED\
+Description: "cover condition: sorted == arr"\
+main.rs:12:5 in function cover_sorted
+
+Status: SATISFIED\
+Description: "cover condition: sorted != arr"\
+main.rs:13:5 in function cover_sorted
+
+ ** 0 of 2 cover properties unsatisfiable
```
celinval, line 9: Can we invert the report? I think we should report the number of covers that were satisfied.

celinval (commented):
Thanks for doing this!

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -46,6 +46,39 @@ fn matches_function(tcx: TyCtxt, instance: Instance, attr_name: &str) -> bool {
     false
 }
 
+struct Cover;
+impl<'tcx> GotocHook<'tcx> for Cover {
```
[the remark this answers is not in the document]
zhassan-aws: Added.

kani-driver/src/cbmc_property_renderer.rs
zhassan-aws: Done.

tests/expected/cover/cover-pass/expected
zhassan-aws: I agree an inverted report is better (I made the change). The reason I didn't do it initially, is that an inverted report is a bit inconsistent with the report for normal checks where we reported the number of things that went wrong (failures) as opposed to those that were successful (similarly, the unsatisfiable covers are the ones that are unexpected).

I'm pro inverting the summary for normal checks as well (perhaps in a separate PR).

library/kani/src/lib.rs
zhassan-aws: Done

kani-driver/src/cbmc_property_renderer.rs
zhassan-aws: Done.

celinval (approved):
Thanks!

Issue #1922, opened 2022-11-21
Tracking issue: Replacing supported features scripts with assess


tedinski:
There are a few tasks that need doing to replace the scripts ([located here](https://github.com/model-checking/kani/tree/main/scripts/exps)) for measuring unsupported features with Assess.

* [ ] Develop a JSON output format for Assess.
* [ ] #1819 is required to reproduce the same metrics we currently produce. (This is a consequence of the switch to MIR linker.)
* [ ] We should have a new mode for measuring unsupported features only. (As opposed to the current goal for assess, of running tests under Kani. To reproduce the current behavior we only need build information, not test running results. We'll want those later.)
* [ ] The reachability mode needs changing as well. We should treat public functions as roots and NOT build in test mode for this.
* [ ] We should produce a real aggregator that runs assess on all targets, reads the json outputs, and outputs aggregate results.
* [ ] Finally, we should run this nightly.

Over time I suspect we'll want to experiment with the mode changing behaviors, to move towards more useful metrics than the current ones we're tracking. But to start we need to reproduce our current metrics, then starting changing things. So I suspect we'll want all of (codegen only, tests mode, pubfns/test fns) to be flags we can turn on/off.

PR #2014, opened 2022-12-16
Refactor Assess table building, introduce TableBuilder.

tedinski:
### Description of changes: 

This has no behavior changes, just code refactoring.

1. Move assess code into a submodule, so we can start organizing it together with several files.
2. Introduce `TableBuilder` (`TableRow`, `RenderableTableRow`) to generalize over how we build tables. The primary purpose of this change is to refactor how we currently build tables so that we can produce programmatic output in the future. (The `build` vs `render` method of `TableBuilder` - not currently used, but identifies the capability.)
3. Break all 3 tables out into their own files and use the common table building approach.

### Resolved issues:

Part of #1922

### Call-outs:

Rust's traits are pretty nice, I like how this worked out, mostly.

### Testing:

* How is this change tested? existing regression

### Checklist


kani-driver/src/assess/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::metadata::merge_kani_metadata;
+use crate::project;
+use crate::session::KaniSession;
+
+mod table_builder;
+mod table_failure_reasons;
+mod table_promising_tests;
+mod table_unsupported_features;
+
+pub(crate) fn cargokani_assess_main(mut session: KaniSession) -> Result<()> {
```
tedinski, line 15: Diff can be hard to read with code moving, so N.B. this function isn't changed other than calling the build functions from the table modules instead of locally.

kani-driver/src/assess/table_builder.rs
```diff
@@ -0,0 +1,168 @@
...
+    Text,
+    /// Numbers should be right-justified.
+    Number,
+}
+
+/// Describes how a table should be printed visually.
+///
+/// This is an extension to [`TableRow`] (which is solely concerned with computing a table's contents.)
+/// In this trait, we are solely concerned with how the table should look when printed.
+pub trait RenderableTableRow {
```
zhassan-aws, line 60: I would prefer if we return a single vector of tuples or something similar. Having different methods return a vector that should be of the same size for all of them makes it a bit error prone.

kani-driver/src/assess/table_promising_tests.rs
```diff
@@ -0,0 +1,84 @@
...
+/// =============================================================================
+/// ```
+pub(crate) fn build(results: &[HarnessResult]) -> Table {
+    let mut builder = TableBuilder::new();
+
+    for r in results {
+        // For now we're just reporting "successful" harnesses as candidates.
+        // In the future this heuristic should be expanded. More data is required to do this, however.
+        if r.result.failed_properties().is_empty() {
+            // The functions we extract are actually the closures inside the test harness macro expansion
```
zhassan-aws, line 36: Can you add an example of the name of a closure, or the template of the closure name?

kani-driver/src/assess/table_failure_reasons.rs
```diff
@@ -0,0 +1,87 @@
...
+
+impl TableRow for FailureReasonsTableRow {
+    type Key = String;
+
+    fn key(&self) -> Self::Key {
+        self.reason.clone()
+    }
+
+    fn merge(&mut self, new: Self) {
+        assert!(new.count == 1);
```
zhassan-aws, line 66: Not sure I understand why this assertion holds. Can you add a comment to clarify?

kani-driver/src/assess/table_failure_reasons.rs
```diff
@@ -0,0 +1,87 @@
...
+/// and the heuristics for "promising tests" are improved.
+///
+/// Example:
+///
+/// ```text
+/// ================================================
+///  Reason for failure           | Number of tests
+/// ------------------------------+-----------------
+///  unwind                       |              61
+///  success                      |               6
```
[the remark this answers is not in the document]
tedinski: This is how it was before, but I'm happy to bikeshed now. I changed it to `none (success)`. If you have a better idea, I'm open.

I do think it's important to represent in this table directly the relative proportion that succeeded/failed.

kani-driver/src/assess/table_builder.rs
tedinski: Unfortunately it complicates code a little to aggregate them, just due to how the table rendering library works.

I did add test cases checking length, and as a bonus I realized with just `#[derive(Default)]` I could also check the length of the `row` result, too.

(I also thought about some more advanced type magic to return fixed-size arrays instead of vectors, but I felt that would be complicating this code too much.)

kani-driver/src/assess/table_failure_reasons.rs
tedinski: It's just due to how it was constructed above. I removed the assertion and generalized the function (because I ended up needing that in the next PR, hah. Easy enough to move that forward into this one.)

zhassan-aws (approved):

PR #2016, opened 2022-12-16
Implement derive macro for Arbitrary

### LINKED ISSUE #726: Create custom derive macro for Arbitrary type.

celinval:
### Description of changes: 

Introduce a `derive` macro that allows users to annotate their structs and enums with `#[derive(kani::Arbitray)]`. This will allow users to easily generate implementation of the `Arbitrary` types for their custom structs and enums when their arbitrary value is a combination of its fields arbitrary values.

### Resolved issues:

Resolves #726

### Call-outs:

- Since we don't publish our crates yet, users will have to use `cfg_attr` to actually use the derive in production code. I.e.:
```rust
#[cfg_attr(kani, derive(kani::Arbitrary))]
```
- I will add documentation as a follow up PR.
- I made a tiny change to our build script to print user friendly errors instead of json ones.

### Testing:

### Checklist


tedinski:
I want to make it clear I don't object, but do we want to duplicate stuff like this ourselves versus directing people towards Bolero and `bolero-generator-derive`?

celinval:
I see your point, but I don't think we should have a hard dependency on `bolero`. We could think about merging the implementations into a separate crate though.

library/kani_macros/src/derive.rs
```diff
@@ -0,0 +1,144 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! This module provides Kani's derive macro for Arbitrary.
```
zhassan-aws, line 3:
```suggestion
//! This module provides Kani's `derive` macro for `Arbitrary`.
```

library/kani_macros/src/derive.rs
```diff
@@ -0,0 +1,144 @@
...
+/// }
+/// ```
+fn fn_any_body(ident: &Ident, data: &Data) -> TokenStream {
+    match data {
+        Data::Struct(struct_data) => init_item(ident, &struct_data.fields),
+        Data::Enum(enum_data) => fn_any_enum(ident, enum_data),
+        Data::Union(_) => {
+            abort!(Span::call_site(), "Cannot derive `Arbitrary` for `{}` union", ident;
+                note = ident.span() =>
+                "The derive `Arbitrary` cannot be used for unions such as `{}`", ident
```
zhassan-aws, line 72:
```suggestion
                "`#[derive(Arbitrary)]` cannot be used for unions such as `{}`", ident
```

library/kani_macros/src/derive.rs
```diff
@@ -0,0 +1,144 @@
...
+            )
+        }
+    }
+}
+
+/// Generate an item initialization where an item can be a struct or a variant.
+/// For named fields, this will generate: `Item { field1: kani::any(), field2: kani::any(), .. }`
+/// For unnamed fields, this will generate: `Item (kani::any(), kani::any(), ..)`
+/// For unit field, generate an empty initialization.
+fn init_item(ident: &Ident, fields: &Fields) -> TokenStream {
```
zhassan-aws, line 82: Rename to `construct_symbolic_item` or `construct_nondet_item`? The `init` gives an impression that it's getting initialized to some (concrete) value.

library/kani_macros/src/derive.rs
```diff
@@ -0,0 +1,144 @@
...
+        }
+        Fields::Unit => {
+            quote! {
+                #ident
+            }
+        }
+    }
+}
+
+/// Generate the body of the function `any()` for structures.
```
zhassan-aws, line 118:
```suggestion
/// Generate the body of the function `any()` for enums.
```

library/kani_macros/src/derive.rs
```diff
@@ -0,0 +1,144 @@
...
+fn fn_any_enum(ident: &Ident, data: &DataEnum) -> TokenStream {
+    if data.variants.is_empty() {
+        abort!(Span::call_site(), "Cannot derive `Arbitrary` for `{}`", ident;
```
zhassan-aws, line 121: Should we make this case a no-op? It seems the compiler doesn't complain about zero-variant enums, e.g.:
```rust
#[derive(Debug)]
enum Foo {}

fn main() {}
```
```
$ rustc enum.rs && ./enum
$ 
```

library/kani_macros/src/derive.rs
```diff
@@ -0,0 +1,144 @@
...
+            note = ident.span() => "`{}` has zero variants and cannot be instantiated", ident
+        );
+    }
+    let arms = data.variants.iter().enumerate().map(|(idx, variant)| {
```
zhassan-aws, line 125: I think I see what this is doing, but can you add to the function documentation an example of the output?

library/kani_macros/src/derive.rs
```diff
@@ -0,0 +1,144 @@
...
+            }
+        } else {
+            quote! {
+                _ => #ident::#init,
+            }
+        }
+    });
+
+    quote! {
+        match kani::any() {
```
zhassan-aws, line 140: Is this guaranteed to use a type that is wide enough for the number of variants?

tests/ui/derive-arbitrary/empty_enum/empty_enum.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! Check that Kani can automatically derive an empty Arbitrary enum.
```
zhassan-aws, line 3: "cannot" instead of "can"? But see my comment on making those a no-op.

tests/ui/derive-arbitrary/enum/expected
```diff
@@ -0,0 +1,24 @@
...
+SUCCESS\
+"assertion failed: disc == -1"
+
+SUCCESS\
+"assertion failed: disc == 0"
+
+
+Checking harness check_enum_wrapper...
+
+SUCCESS\
```
zhassan-aws, line 17: Are you relying on this being reported as "UNREACHABLE" if the variant is not possible? If so, you might want to use the brand new `cover` macro instead :)

tests/ui/derive-arbitrary/generic_struct/generic_struct.rs
```diff
@@ -0,0 +1,23 @@
...
+struct Point<X, Y> {
+    x: X,
+    y: Y,
+}
+
+#[kani::proof]
+fn check_arbitrary_point() {
+    let point: Point<i32, i8> = kani::any();
+    if kani::any() {
+        assert!(point.x >= 0);
```
zhassan-aws, line 15: Might want to use `cover` instead.

zhassan-aws (commented):
Very nice!

library/kani_macros/src/derive.rs
celinval: But invoking `kany::any::<Foo>()` would an error with a not so friendly message.

library/kani_macros/src/derive.rs
celinval: Yes. Because the type will be decided based on the number of match conditions

tests/ui/derive-arbitrary/enum/expected
celinval: Yes. I was thinking about that while writing all these tests. =)

tests/ui/derive-arbitrary/generic_struct/generic_struct.rs
celinval: Yes! I will rewrite them.

library/kani_macros/src/derive.rs
zhassan-aws: I see. I guess it doesn't matter much.

library/kani_macros/src/derive.rs
celinval: Is `init_symbolic_item` OK?

zhassan-aws: Sure.

library/kani_macros/src/derive.rs
celinval: Done

tests/ui/derive-arbitrary/empty_enum/empty_enum.rs
celinval: I guess it can now.

zhassan-aws (approved):
Thanks!

PR #2019, opened 2022-12-18
Implement arbitrary for up to size-12 tuples.

### LINKED ISSUE #1965: Provide `Arbitrary` implementation for tuples
Requested feature: It would be nice if Kani implements `Arbitrary` for tuples. 
Use case: To allow assigning symbolic constants to tuples, e.g.:
```rust
#[kani::proof]
fn main() {
    let x: (u8, i32) = kani::any();
}
```
instead of having to do:
```rust
    let x: (u8, i32) = (kani::any(), kani::any());
```
It perhaps makes sense to provide the implementation up to a certain tuple size, for example 12 similar to how `std` provides an implementation for `Debug` for tuples up to twelve items long:  https://doc.rust-lang.org/std/primitive.tuple.html#impl-Debug-for-(T%2C)

YoshikiTakashima:
### Description of changes: 

Implements `kani::Arbitrary`  for tuples

### Resolved issues:

Resolves #1965 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? tests/kani/Intrinsics/SIMD/Operators/bitwise.rs

### Checklist


library/kani/src/tuple.rs
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// Modifications Copyright Kani Contributors
```
zhassan-aws, line 3: Is there a reason this is using the "modifed-file" license header as opposed to the standard one?

library/kani/src/tuple.rs
```diff
@@ -0,0 +1,38 @@
...
+// See GitHub history for details.
+
+//! Support for arbitrary tuples where each element implements
+//! `kani::Arbitrary`. Tuples of size up to 11 are supported in this
```
zhassan-aws, line 7: Isn't this supporting up to 12 (A - L)?

library/kani/src/tuple.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use crate::Arbitrary;
+
+/// This macro implements `kani::Arbitrary` on a tuple whose elements
+/// already implement `kani::Arbitrary` by running `kani::any()` on
+/// each index of the tuple.
+macro_rules! tuple {
+    ($($typ:ident),*) => {
+	impl<$($typ : Arbitrary),*>  Arbitrary for ($($typ,)*) {
+            #[inline(always)]
```
zhassan-aws, line 18: Is the indentation off on this line?

library/kani/src/tuple.rs
```diff
@@ -0,0 +1,38 @@
...
+	    fn any() -> Self {
+		($(crate::any::<$typ>(),)*)
```
zhassan-aws, line 20: The indentation seems off on this line as well. Perhaps because of tabs?

zhassan-aws (commented):
Thanks @YoshikiTakashima! I have a few comments.

library/kani/src/tuple.rs
YoshikiTakashima: No. I just copied the license header of whatever I had open next to it, which was likely the modified proptest. 

I'll change it back.

YoshikiTakashima:
@zhassan-aws Thanks for the review. I'm a bit busy with other parts of kani, so I would get back to this PR around Jan. 2023. Have a nice holidays.

library/kani/src/tuple.rs
YoshikiTakashima: 72a48f243631eef863686f1726db395a867df4a0

library/kani/src/tuple.rs
YoshikiTakashima: 72a48f243631eef863686f1726db395a867df4a0

library/kani/src/tuple.rs
YoshikiTakashima: 72a48f243631eef863686f1726db395a867df4a0, also fixed PR title so the squash commit says 12.

library/kani/src/tuple.rs
YoshikiTakashima: 72a48f243631eef863686f1726db395a867df4a0

zhassan-aws:
Thanks @YoshikiTakashima! Can we add a couple of tests for the minimum and maximum tuple sizes (i.e. 1 and 12) to make sure these are working correctly, and do not break in the future?

zhassan-aws (approved):
Thanks!

PR #2027, opened 2022-12-21
Delete kani::expect_fail and associated hook

### LINKED ISSUE #1197: Rename `expect_fail` to `should_fail` or other phrases.
### Use Case 
`expect_fail` can be slightly confusing as an expression in some cases. For example, in the case ->

``` rust
#[kani::proof]
fn main() {
    let x: i32 = kani::any();
    kani::expect_fail(x == 5, "x is not 5!");
}
```

It is unclear whether `expect_fail(x == 5) ;- SUCCESS. Failed as Expected` here implies that x is equal to 5 or if x is not equal to 5 when it fails. While the developer can leave a description that can help themselves understand the output better, the phrase itself should be enough 

### Proposed Solution 
If we were to rename it to `should_fail`, it becomes easier to interpret this code as -
"the code should fail as x != 5 . SUCCESS means we can ensure, FAILURE means that we can't ensure."

Alternative titles - 
1. `should_fail`
2. `check_fail`
3. `ensure_fail`
4. `verify_fail`

Note : This could be my personal biases and this might not be an issue for others.

### LINKED ISSUE #1538: Handling of `expect_fail` in the regression is broken
I tried adding the following code to our test suite:

```rust
#[kani::proof]
fn check_obvious() {
    kani::expect_fail(true, "this won't fail");
}
```

```
./scripts/kani-regression.sh
```

with Kani version:

I expected to see this happen: Regression should fail since it has a `kani::expect_fail()` statement that does not fail.

Instead, this happened: Regression passed.

My suggestion is to actually deprecate `kani::expect_fail()` and have a proper `kani::cover()` implementation. In this case, `kani` itself should fail if a `cover` statement is unreachable or unsatisfied.

Note that a bunch of out tests depend on this mechanism to work properly and it is possible that we have regressions not being detected today. One example of this happening was #1533.

zhassan-aws:
### Description of changes: 

Get rid of `kani::expect_fail`, and replace its usage in the regressions by `kani::cover`.

### Resolved issues:

Resolves #1197 
Resolves #1538 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

There were many `_fail` tests that used `kani::expect_fail` that had a corresponding non-failing test that used `assert`. For example, a test name `foo.rs` would have:
```rust
let x = ...
assert!(x == 2);
```
and there would be a corresponding test named `foo_fail.rs` that has:
```rust
let x = ...
kani::expect_fail(x != 2);
```
These tests were essentially useless (since the main test verifies that the assertion passes), so I deleted all of them.

I only kept tests that were meaningful, and replaced `kani::expect_fail` with `kani::cover`.

### Testing:

* How is this change tested? Updated regressions

### Checklist


tests/kani/DynTrait/boxed_closure_fail.rs
```diff
@@ -1,15 +0,0 @@
-// Copyright Kani Contributors
-// SPDX-License-Identifier: Apache-2.0 OR MIT
-// kani-verify-fail
-
-// Check that we can codegen a boxed dyn closure and fail an inner assertion
-#[kani::proof]
-fn main() {
-    // Create a boxed once-callable closure
-    let f: Box<dyn FnOnce(i32)> = Box::new(|x| {
-        kani::expect_fail(x == 2, "Wrong int");
```
celinval, line 10: What happened to this test? Did you forget to do a git add on the new ones??

tests/expected/arbitrary/integers/expected
```diff
@@ -0,0 +1,222 @@
...
+Description: "This may also be true"\
+main.rs:21:1 in function check_i16
+
+Status: SATISFIED\
+Description: "v1 may be MAX"\
+main.rs:21:1 in function check_i16
+
+Status: SATISFIED\
+Description: "v1 may be MIN"\
+main.rs:21:1 in function check_i16
```
celinval, line 179: Please don't include line and column in `expected` files, unless they are extremely important. It can be annoying to update them. E.g.: Adding a comment in this file would require changing all the lines in the `expected` file.

celinval (changes requested):
That's awesome! Thanks for doing this. Question, did you mean to delete or replace the `DynTrait` tests?

tests/kani/DynTrait/boxed_closure_fail.rs
zhassan-aws: I deliberately deleted most of the `*_fail` tests because they are useless. For example, there is a corresponding test called `boxed_closure.rs` which is almost identical to this one except that it uses asserts instead of `expect_fail`:

https://github.com/model-checking/kani/blob/main/tests/kani/DynTrait/boxed_closure.rs

This test is passes, and thus subsumes the `boxed_closure_fail.rs` test. This is the same case for the tests I deleted (I checked them one by one).

See the call-outs for more details.

tests/expected/arbitrary/integers/expected
zhassan-aws: Sure, I'll remove them. What is important to include is the `in function <function-name>` on the next line to make sure this check passes in all harnesses. For example, if the `expected` file only had:
```
Status: SATISFIED\
Description: "v1 may be MIN"
```
then the test will pass if this check was satisfied in _any_ harness. We need to make sure it passes in _all_ harnesses.

zhassan-aws: Done.

tests/kani/DynTrait/boxed_closure_fail.rs
celinval: Ok. Sorry about that

celinval (approved):

PR #2033, opened 2022-12-22
Eliminate a source of code bloat in the kani library

### LINKED ISSUE #2010: Warning about unsupported `caller_location` intrinsic

zhassan-aws:
### Description of changes: 

Same trick applied in #2032, but this time to the `kani::panic` function. In addition to eliminating the `caller_location` warning on a small program like this one:
```rust
#[kani::proof]
fn main() {
    panic!("Some message");
}
```
it also results in a speedup and a significant reduction in the MIR size.

Before:
```bash
$ /usr/bin/time -p kani simp_panic.rs
...
real 3.92
user 3.54
sys 0.37

$ RUSTFLAGS="--emit mir" kani simp_panic.rs
$ wc simp_panic.kani.mir 
  33248  237545 4173632 simp_panic.kani.mir
```
After:
```bash
$ /usr/bin/time -p kani simp_panic.rs
...
real 0.15
user 0.11
sys 0.04

$ RUSTFLAGS="--emit mir" kani simp_panic.rs
$ wc simp_panic.kani.mir 
  86  530 5399 simp_panic.kani.mir
```

### Resolved issues:

Resolves #2010 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? Couldn't find a way to test this.

### Checklist

### DIFF
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -136,13 +136,7 @@ pub(crate) unsafe fn any_raw_internal<T, const SIZE_T: usize>() -> T {
 #[inline(never)]
 #[allow(dead_code)]
 fn any_raw_inner<T>() -> T {
-    // while we could use `unreachable!()` or `panic!()` as the body of this
-    // function, both cause Kani to produce a warning on any program that uses
-    // kani::any() (see https://github.com/model-checking/kani/issues/2010).
-    // This function is handled via a hook anyway, so we just need to put a body
-    // that rustc does not complain about. An infinite loop works out nicely.
-    #[allow(clippy::empty_loop)]
-    loop {}
+    __kani_dummy_never()
 }
 
 /// Function used to generate panic with a static message as this is the only one currently
@@ -154,8 +148,22 @@ fn any_raw_inner<T>() -> T {
 #[inline(never)]
 #[rustc_diagnostic_item = "KaniPanic"]
 #[doc(hidden)]
-pub const fn panic(message: &'static str) -> ! {
-    panic!("{}", message)
+pub const fn panic(_message: &'static str) -> ! {
+    __kani_dummy_never()
+}
+
+/// This is a dummy function that is intended to be called from functions with
+/// `rustc_diagnostic_item`, i.e. that have kani hooks, to make sure they
+/// compile.
+/// While we could use `unreachable!()` or `panic!()` as the body of this
+/// function, both cause Kani to produce a warning on missing `caller_location`
+/// intrinsic (see https://github.com/model-checking/kani/issues/2010) and
+/// produce unnecessarily large MIR.  This function is handled via a hook
+/// anyway, so we just need to put a body that rustc does not complain about. An
+/// infinite loop works out nicely.
+const fn __kani_dummy_never() -> ! {
+    #[allow(clippy::empty_loop)]
+    loop {}
 }
 
 /// A macro to check if a condition is satisfiable at a specific location in the


library/kani/src/lib.rs
```diff
@@ -154,8 +148,22 @@ fn any_raw_inner<T>() -> T {
...
+/// This is a dummy function that is intended to be called from functions with
+/// `rustc_diagnostic_item`, i.e. that have kani hooks, to make sure they
+/// compile.
+/// While we could use `unreachable!()` or `panic!()` as the body of this
+/// function, both cause Kani to produce a warning on missing `caller_location`
+/// intrinsic (see https://github.com/model-checking/kani/issues/2010) and
+/// produce unnecessarily large MIR.  This function is handled via a hook
+/// anyway, so we just need to put a body that rustc does not complain about. An
+/// infinite loop works out nicely.
+const fn __kani_dummy_never() -> ! {
```
tedinski, line 164: `#[doc(hidden)]` for things we don't want customer-facing.

tedinski (approved):
I'm slightly concerned about playing whack-a-mole with this.

With assess, I think we also see a lot of code bloat and caller_location. I haven't looked into it yet, but it'd be nice if we could integrate a hook or something in Kani that fixed all this everywhere, and then we could revert these back to being basic `unreachable!()` or whatever.

library/kani/src/lib.rs
zhassan-aws: Is this needed for a private function?

tedinski: Oh, good point. Nevermind then :)

celinval:
I don't think hooks will help here, since it would run after the reachability analysis. We could potentially plug hooks into the reachability analysis or use stubs here.

For the caller location warning, we should probably just fix it.

celinval:
Also, I mentioned this to @zhassan-aws. Fixing our library is not really going to fix the problem in real user code, since any call to the std library that may eventually panic will bring in all that code. E.g.: `unwrap()`.

zhassan-aws:
> Fixing our library is not really going to fix the problem in real user code, since any call to the std library that may eventually panic will bring in all that code. E.g.: unwrap().

True. One interesting thing though is that `panic` in particular seems to be excessively causing a MIR code bloat:
```
$ cat simp_panic.rs 
#[kani::proof]
fn main() {
    panic!("Some message");
}
$ RUSTFLAGS="--emit mir" kani simp_panic.rs
$ wc simp_panic.kani.mir 
  33248  237545 4173632 simp_panic.kani.mir
```

```
$ cat simp_unwrap.rs
#[kani::proof]
fn main() {
    let x: Option<i32> = kani::any();
    x.unwrap();
}
$ RUSTFLAGS="--emit mir" kani simp_unwrap.rs
$ wc simp_unwrap.kani.mir 
  1083   7418 124029 simp_unwrap.kani.mir
```
I'm not sure what else `panic` brings other than formatting code (which `unwrap` also bring as @celinval pointed out) that is causing the number of lines of MIR to be 33X that of the unwrap program.

These numbers are using c135fa9017f.

zhassan-aws:
I evaluated the impact of this change on a few customer tests and found the reduction in the MIR size to be minimal. Abandoning this PR.

PR #2034, opened 2022-12-22
Report bookrunner failures for each stage

adpaco-aws:
### Description of changes: 

Adds a Python script that can be used to collect failures for each stage based on the `bookrunner` HTML report. The script is used to print this data in the bookrunner workflow.

### Testing:

* How is this change tested? Testing through CI.

### Checklist

### DIFF
--- a/.github/workflows/kani.yml
+++ b/.github/workflows/kani.yml
@@ -96,7 +96,10 @@ jobs:
       - name: Print book runner text results
         run: cat build/output/latest/html/bookrunner.txt
 
-      - name: Detect unpexpected book runner failures
+      - name: Print book runner failures grouped by stage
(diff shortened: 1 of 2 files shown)


celinval (approved):
It looks good to me. How hard would it be to actually print the failure? E.g.: Timeout x Crash.

adpaco-aws:
> It looks good to me. How hard would it be to actually print the failure? E.g.: Timeout x Crash.

This works on the HTML output, which means it would be much harder.

I submitted this as a "temporary" solution (that I was already using) so all of us can collect bookrunner metrics in just a few clicks. But actually, the best way to do this would be working on the `run.json` file that Litani produces. It's much easier to get the exact failures (or any other data, really) in there. However, I think `run.json` was being removed before uploading the results because of space constraints on GH pages. I'll look into that.

PR #2039, opened 2022-12-23
Add support to cdylib, dylib and staticlib

### LINKED ISSUE #1915: Add support to libraries with `cdylib` type
Requested feature: Users may want to verify crates that also support the generation of a `cdylib`. One good example can be found at #1896. 
Use case: https://github.com/nyx-space/hifitime 

Test case:

```bash
git clone https://github.com/nyx-space/hifitime.git
cd hifitime
cargo kani
```

celinval:
### Description of changes: 

Refactor how we were creating the `rlib` files in the link stage. We cannot invoke rustc's linker directly for the native types since they will call the native linker that will fail. Instead, we just manually create an `rlib` file that includes only a `metadata` file.

I cleaned up our archive builder since we no longer need to conform to rustc's interface.

### Resolved issues:

Resolves #1915

### Testing:

### Checklist


kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -188,38 +188,54 @@ impl CodegenBackend for GotocCodegenBackend {
...
         codegen_results: CodegenResults,
         outputs: &OutputFilenames,
     ) -> Result<(), ErrorGuaranteed> {
-        // In `link`, we need to do two things:
-        // 1. We need to emit `rlib` files normally. Cargo expects these in some circumstances and sends
-        //    them to subsequent builds with `-L`.
-        // 2. We MUST NOT try to invoke the native linker, because that will fail. We don't have real objects.
-        // This is normally not a problem: usually we only get one requested `crate-type`.
-        // But let's be careful and fail loudly if we get conflicting requests:
+        use crate::codegen_cprover_gotoc::archive::ArchiveBuilder;
```
celinval, line 207: Shoot, I forgot to move these to the top of the file. I'll do it in the next iteration.

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -188,38 +194,46 @@ impl CodegenBackend for GotocCodegenBackend {
...
-        //    them to subsequent builds with `-L`.
-        // 2. We MUST NOT try to invoke the native linker, because that will fail. We don't have real objects.
-        // This is normally not a problem: usually we only get one requested `crate-type`.
-        // But let's be careful and fail loudly if we get conflicting requests:
         let requested_crate_types = sess.crate_types();
-        // Quit successfully if we don't need an `rlib`:
-        if !requested_crate_types.contains(&rustc_session::config::CrateType::Rlib) {
+        if !requested_crate_types.contains(&CrateType::Rlib) {
+            // Quit successfully if we don't need an `rlib`:
+            println!("No need for rlib");
```
zhassan-aws, line 216: Is this print statement intentional, or a left-over from debugging?

kani-driver/src/call_cargo.rs
```diff
@@ -211,55 +211,49 @@ impl VerificationTarget {
...
-        .collect();
+        }
+        match (supported_lib, unsupported_lib) {
+            (true, true) => println!(
+                "warning: Skipped verification of `{}` due to unsupported crate-type: \
+                        `proc-macro`.",
+                target.name,
+            ),
+            (true, false) => verification_targets.push(VerificationTarget::Lib),
+            (_, _) => {}
```
zhassan-aws, line 254: What happens in the case of `(false, true)`?

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
tedinski: Agree, I don't think we want a print here.

celinval: Leftovers... :)

kani-compiler/src/codegen_cprover_gotoc/archive.rs
```diff
@@ -2,114 +2,42 @@
...
-use object::ReadCache;
-
-#[derive(Debug)]
-enum ArchiveEntry {
-    FromArchive { archive_index: usize, file_range: (u64, u64) },
-    File(PathBuf),
-}
-
-pub(crate) struct ArArchiveBuilder<'a> {
+pub(crate) struct ArchiveBuilder<'a> {
```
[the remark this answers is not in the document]
celinval: We don't have access to this yet, since we are still using an older version of `rustc`. But once we do update to a newer version, I agree that we should be able to completely remove this file and just use `ArArchiveBuilder` directly.

kani-driver/src/call_cargo.rs
celinval: For `(false, true)`, if `verbose`, we will print that the current target was skipped due to unsupported type. Otherwise, we will silently ignore the target.

celinval: Note that the behavior hasn't changed here and it is still covered by this test: https://github.com/model-checking/kani/tree/main/tests/cargo-ui/unsupported-lib-types/proc-macro

zhassan-aws (approved):

Issue #1342, opened 2022-07-06
Sinf/Cosf not being correctly evaluated.


ssoudan:
Might have missed something but so far I think the following is happening:
Rust `intrinsics::sinf32()` is substituted with C `sinf()` which itself seems to be substituted inside CBMC.

This code does some __CPROVER_assume() but no actual calculation of the sinus (in https://github.com/diffblue/cbmc/blob/16240afb9a855951533a57a0343b1737dd3f0168/src/ansi-c/library/math.c#L508):
```c
/* FUNCTION: sinf */

float __VERIFIER_nondet_float();

float sinf(float x)
{
  // gross over-approximation
  float ret=__VERIFIER_nondet_float();

  if(__CPROVER_isinff(x) || __CPROVER_isnanf(x))
    __CPROVER_assume(__CPROVER_isnanf(ret));
  else
  {
    __CPROVER_assume(ret<=1);
    __CPROVER_assume(ret>=-1);
    __CPROVER_assume(x!=0 || ret==0);
  }

  return ret;
}
```

The result from kani perspective is that the following code fails the verification:
```rust
use std::f32::consts::FRAC_PI_2;

#[cfg_attr(kani, kani::proof)]
fn main() {
    let a = FRAC_PI_2;
    println!("a={}", a);

    let sin_a = a.sin();
    println!("sin_a={:.}", sin_a);
    assert_eq!(sin_a, 1.);
}
```

using the following command line invocation:
```
kani src/sinf.rs    
Checking harness main...
CBMC 5.60.0 (cbmc-5.60.0)
CBMC version 5.60.0 (cbmc-5.60.0) 64-bit arm64 macos
Reading GOTO program from file src/sinf.out.for-main
Generating GOTO Program
Adding CPROVER library (arm64)
Removal of function pointers and virtual functions
Generic Property Instrumentation
Running with 16 object bits, 48 offset bits (user-specified)
Starting Bounded Model Checking
Runtime Symex: 0.00116479s
size of program expression: 76 steps
slicing removed 39 assignments
Generated 2 VCC(s), 2 remaining after simplification
Runtime Postprocess Equation: 3.25e-05s
Passing problem to propositional reduction
converting SSA
Runtime Convert SSA: 0.00020675s
Running propositional reduction
Post-processing
Runtime Post-process: 1.9166e-05s
Solving with MiniSAT 2.2.1 with simplifier
97 variables, 455 clauses
SAT checker: instance is SATISFIABLE
[...]

PR #2050, opened 2022-12-29
Add 'float' note to limitations doc

tedinski:
### Description of changes: 

Add an additional limitation about some floating point operations.

### Resolved issues:

Related to #1342

### Testing:

* How is this change tested? n/a

* Is this a refactor change? n/a

### Checklist


docs/src/rust-feature-support.md
```diff
@@ -200,3 +200,12 @@ related to [advanced features](#advanced-features).
 
 Please refer to [Intrinsics](rust-feature-support/intrinsics.md) for information
 on the current support in Kani for Rust compiler intrinsics.
+
+### Floating point operations
+
+Kani supports floating point numbers, but some kinds of operations on floats are "over-abstracted."
```
zhassan-aws, line 206: I think we need to specify the exact set of operations. Otherwise, it may sound that none of the floating-point operations can be trusted, which is not the case. From looking at https://github.com/diffblue/cbmc/blob/16240afb9a855951533a57a0343b1737dd3f0168/src/ansi-c/library/math.c#L508, it seems only trigonometric functions and sqrt use abstractions (and consequently, any functions that call them).

docs/src/rust-feature-support.md
```diff
@@ -200,3 +200,12 @@ related to [advanced features](#advanced-features).
...
+This means the verifier can raise errors that cannot actually happen when the code is run normally.
+For instance, ([#1342](https://github.com/model-checking/kani/issues/1342)) the sin/cos functions basically return a nondeterministic value between -1 and 1.
```
zhassan-aws, line 208:
```suggestion
For instance, ([#1342](https://github.com/model-checking/kani/issues/1342)) the `sin`/`cos` functions basically return a nondeterministic value between -1 and 1.
```

docs/src/rust-feature-support.md
```diff
@@ -200,3 +200,12 @@ related to [advanced features](#advanced-features).
...
+In other words, they largely ignore their input and give very conservative answers.
```
zhassan-aws, line 209: I would add to the end "that guarantee verification soundness".

docs/src/rust-feature-support.md
```diff
@@ -200,3 +200,12 @@ related to [advanced features](#advanced-features).
...
+This range certain includes the "real" value, but it means Kani could raise spurious errors that cannot actually happen.
+This makes Kani unsuitable for verifying some kinds of properties (e.g. precision) about numerical algorithms.
```
zhassan-aws, line 211: Should we include a suggestion of using stubs to replace the usage of those functions with versions that suit the verification needs?

tedinski: I initially decided no, but I thought again and went for it. :)

docs/src/rust-feature-support.md
tedinski: Hmm, I added a sentence saying this, but I tried to leave a tiny bit of wiggle room for there to be more operations. Without having deeply looked into it myself, I don't really have fully confidence we definitely know the exact set, and... well, it might change over time.

docs/src/rust-feature-support.md
```diff
@@ -200,3 +200,14 @@ related to [advanced features](#advanced-features).
...
+### Floating point operations
+
+Kani supports floating point numbers, but some supported operations on floats are "over-approximated."
+These are the trigonometric functions like `sin` and `cos` and the `sqrt` function as well.
+This means the verifier can raise errors that cannot actually happen when the code is run normally.
+For instance, ([#1342](https://github.com/model-checking/kani/issues/1342)) the `sin`/`cos` functions basically return a nondeterministic value between -1 and 1.
+In other words, they largely ignore their input and give very conservative answers.
+This range certainly includes the "real" value, so proof soundness is still preserved, but it means Kani could raise spurious errors that cannot actually happen.
+This makes Kani unsuitable for verifying some kinds of properties (e.g. precision) about numerical algorithms.
+Proofs that fail because of this problem can sometimes be repaired by introducing "stubs" for these functions that return a more acceptable approximation.
```
tautschnig, line 213: It may be worth adding a note to the effect of what @martin-cs said in https://github.com/diffblue/cbmc/issues/6999#issuecomment-1220773186: actual implementations may vary from platform to platform, and more precise models (stubs) may well introduce unsoundness.

martin-cs: Is there even a standard for numerical accuracy for Rust or is it implementation defined?  This:

https://stackoverflow.com/questions/21177436/is-there-a-published-language-format-standard-for-rust-yet

makes me a little concerned.  I note that:

https://doc.rust-lang.org/stable/reference/types/numeric.html

links to IEEE-754 2008 and that should be correctly supported.  The documentation for the numerical functions is ... a bit thin:

https://doc.rust-lang.org/std/primitive.f64.html#method.sin

zhassan-aws (approved):

PR #2057, opened 2023-01-03
Add documentation for cover properties

### LINKED ISSUE #2056: Add documentation for `cover `macro

zhassan-aws:
### Description of changes: 

Add a description of `kani::cover` and how its results are reported to the documentation.

### Resolved issues:

Resolves #2056 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? 

* Is this a refactor change?

### Checklist


docs/src/verification-results.md
```diff
@@ -94,3 +94,66 @@ Check 2: undetermined_example.assertion.2
...
+```
+which is not possible as such string will contain at least two characters.
+```
+Check 46: cover_unsatisfiable_example.cover.1
+         - Status: UNSATISFIABLE
+         - Description: "cover condition: s.chars().count() <= 1"
+         - Location: src/main.rs:75:9 in function cover_unsatisfiable_example
+```
+
+3. `UNREACHABLE`: This indicates that the `cover` property itself is unreachable (i.e. it is _vacuously_ unsatisfiable).
```
tedinski, line 132: I would add something here like:

> In contrast to an `UNREACHABLE` result for assertions, an unreachable cover check is considered a proof failure.

tedinski (approved):
1 comment, otherwise looks good.

docs/src/verification-results.md
zhassan-aws: Thanks. I weakened the statement a bit, since we don't fail verification if a cover property is unreachable or unsatisfiable:

> In contrast to an `UNREACHABLE` result for assertions, an unreachable (or an unsatisfiable) cover property may indicate an incomplete proof.

Good?

tedinski: Oh, I thought that was largely the point of cover, but if that's not the behavior then that looks good for the doc.

zhassan-aws: Yeah, that was originally the plan, but the discussion [here ](https://github.com/model-checking/kani/pull/1906#issuecomment-1318036336)led to changing it.

Issue #1217, opened 2022-05-24
Combine proof annotation with attributes


zhassan-aws:
Requested feature: Currently, harnesses need to be annotated with `#[kani::proof]`, and with `#[kani::unwind(n)]` if the harness requires an unwinding. Would be nice if those two can be combined, e.g.:
```
#[kani::proof(unwind = 8)]
```

tedinski:
iirc, we explicitly decided against this, I don't remember why though

camshaft:
combining the attributes would make it a bit cleaner when using `cfg_attr`:
```rust
#[cfg_attr(kani, kani::proof(unwind = 8))]
```

vs

```rust
#[cfg_attr(kani, kani::proof)]
#[cfg_attr(kani, kani::unwind(8))]
```

camshaft:
oh nice. i didn't know you could do that!

PR #2088, opened 2023-01-06
Allow user to specify CBMC's solver

### LINKED ISSUE #1656: Feature request: make it easier to use other SAT solvers (e.g. kissat)
In some cases, [kissat](https://github.com/arminbiere/kissat) is an order of magnitude faster than minisat, which is the default used by CBMC. In one example, I've seen a reduction in running time by a factor of 15.

To use kissat currently, you need to build it yourself and run

```
kani --enable-unstable --cbmc-args --external-sat-solver path/to/kissat
```

It would be nice to be able to run `kani --solver kissat` instead.

Even better, we could even look into shipping Kani with kissat, so the user does not have to build it himself. The license of kissat (MIT) is compatible with Kani's.

As a further improvement, we could consider running multiple solvers in parallel and cancel the runs as soon as one of them finishes.

zhassan-aws:
### Description of changes: 

Adds two methods that allow the user to select the solver to use with CBMC:
1. A new command-line `--solver `option
2. A `kani::solver` attribute

The `kani::solver` attribute allows selecting a different solver for each harness, e.g.
```rust
#[kani::proof]
#[kani::solver(kissat)]
fn check() { ... }
```
If the command-line `--solver` option is specified, it overrides the solver specified in the attribute (if any).

Valid values are `kissat` and `minisat`. Passing a custom SAT solver binary can be done with:
```
#[kani::solver(custom = "my_sat_solver")]
```
This case gets passed to CBMC's `--external-sat-solver` option directly.

### Resolved issues:

Resolves #1656 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? Added several tests

### Checklist


celinval:
I would prefer if the attribute is based on an ID, not a path. I don't think it's a good practice to hardcode paths in the code, since it is hard to ensure that the path works for every developer and the CI.

To make this extensible for users that want to try something new that doesn't ship with Kani, I would prefer adding installation instructions, command line or a utility that configures a new engine via its path into a format / path that Kani expects.

zhassan-aws:
> I would prefer if the attribute is based on an ID, not a path. I don't think it's a good practice to hardcode paths in the code, since it is hard to ensure that the path works for every developer and the CI.

Point taken. A `kani::solver` attribute should be ID-based to support future plans of integrating multiple backends.

For the purposes of allowing the user to control CBMC's SAT solver specifically (which this PR is targeting), I can think of two options:
1. Introduce a CBMC-specific annotation, e.g. `kani::cbmc_solver` or even `kani::cbmc_sat_solver` where the argument would be `kissat`, `cadical`, etc.
2. Use `kani::solver`, and prefix each SAT solver option with `cbmc` (as we discussed offline), e.g. `cbmc_kissat`, `cbmc_cadical`, etc.

The advantage of option 1 is that it would keep the more valuable/more generic `kani::solver` attribute free, allowing us some time to design it based on future plans for integrating more backends.

The advantage of option 2 is that the user only needs to learn about one attribute for specifying the solver (i.e. fewer control knobs) which makes the usage simpler.

Thoughts?

rahulku:
Option 2 sounds better.

celinval:
Just a nit... maybe use "::" as a convention, it is also less likely that "::" will be in the same of a binary.

I was thinking, since we only have cbmc today, we could consider that things that an id that doesn't have "::" uses cbmc. I.e.: `kani::solver("minisat")` is the same as `kani::solver("cbmc::minisat")`.

Another option would be to start with what you have, and once we add more options, we use key / value for the attribute. And we support the previous mode as legacy purpose. I.e.:
`kani::solver("minisat")` == `kani::solver(engine="cbmc", sat_solver="minisat")`

adpaco-aws:
Thanks for pushing on this, @zhassan-aws !

Why are we adding a new attribute? We could just make `engine` and `solver` optional arguments in `kani::proof`, your example would look like this:
```
#[kani::proof(solver = kissat)]
fn check() { ... }
```

This idea comes from @fzaiser in #1656, and in my opinion is much cleaner than the other options considered here. It'd work in a similar way to @celinval 's last suggestion, assuming `engine = cbmc` and `solver = minisat` at the moment.

I don't think engines and solvers should be coupled in any of these options neither (i.e., things like `cbmc_kissat` or `cbmc::kissat`). Is there a technical reason for us to keep them that way? Even if newer backends only work with certain solvers, we can always limit these choices from Kani, but in principle we should aim for maximum flexibility.

zhassan-aws:
> Why are we adding a new attribute? We could just make `engine` and `solver` optional arguments in `kani::proof`:

I fully agree with making attributes arguments to `kani::proof`, and have proposed it in #1217.

zhassan-aws:
> Are we going in the direction of adding a new attribute vs attribute arguments?

We agreed that we will support both approaches.

celinval:
I have a few comments regarding the UX from the PR description:
1. We should also accept `minisat` as a value.
2. We should create a `--list-sovers` that prints all the solvers known to Kani.
3. I was wondering if we should consider adding a `--default-solver` similar to the `--default-unwind` option.

zhassan-aws:
> * We should also accept `minisat` as a value.
> 
Make sense.

> * We should create a `--list-sovers` that prints all the solvers known to Kani.
> 
Good idea.

> * I was wondering if we should consider adding a `--default-solver` similar to the `--default-unwind` option.

TBH, I don't like `--default-unwind`, and I haven't seen it being used much. The behavior that makes sense to me is that command-line options should _override_ attributes.

celinval:
> > * We should also accept `minisat` as a value.
> 
> Make sense.
> 
> > * We should create a `--list-sovers` that prints all the solvers known to Kani.
> 
> Good idea.
> 
> > * I was wondering if we should consider adding a `--default-solver` similar to the `--default-unwind` option.
> 
> TBH, I don't like `--default-unwind`, and I haven't seen it being used much. The behavior that makes sense to me is that command-line options should _override_ attributes.

Sounds good to me.

zhassan-aws:
Added minisat in e426f7905ff7a967f1321fe736eacc234baba191.

For `--list-solvers`, I just recalled that the help already displays the options:
```
$ kani --help
...
      --solver <SOLVER>
          Specify the CBMC solver to use. Overrides the harness `solver` attribute
          [possible values: "SAT solver binary", kissat, minisat]
...
```
Is this sufficient?

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -4,6 +4,18 @@
 
 use rustc_ast::{AttrKind, Attribute, LitKind, MetaItem};
 
+/// An enum for possible errors with attribute extraction
+pub enum AttrError {
```
adpaco-aws, line 8: Should this be `KaniToolAttrError` or `KaniAttrError`? I thought these came from `rustc_ast` but then found this enum.

kani-driver/src/args.rs
```diff
@@ -150,6 +152,9 @@ pub struct KaniArgs {
     /// Specify the value used for loop unwinding for the specified harness in CBMC
     #[arg(long, requires("harness"))]
     pub unwind: Option<u32>,
+    /// Specify the CBMC solver to use. Overrides the harness `solver` attribute.
+    #[arg(long, value_parser = clap::builder::PossibleValuesParser::new(CbmcSolver::VARIANTS))]
+    pub solver: Option<String>,
```
adpaco-aws, line 157: I was expecting this to be `Option<Solver>`, and the annotation to allow only a set of literals (associated to solvers) and not strings (e.g., `#[kani::solver(kissat)]`)

adpaco-aws (commented):
I don't think the arguments to the annotation should be strings. If we're being forced to use strings because we need to support custom solvers, then I think we should consider other solutions for that scenario (e.g., have a `custom` literal and a way to pass the path).

In any case, I don't think we need to support that use-case in this PR, unless there is no way to use a different solver after this is merged (but one can still pass `--external-sat-solver` through `--cbmc-args`, right?).

adpaco-aws:
> I think we should consider other solutions for that scenario (e.g., have a `custom` literal and a way to pass the path).

One option that I really like is passing the path to the solver with an environment variable `KANI_CUSTOM_SOLVER`, similar to how we allow an alternative setup path with `KANI_HOME`.

tests/ui/solver-attribute/not-found/expected
```diff
@@ -0,0 +1 @@
+Error: Specified solver "non_existing_solver" not found in path
```
celinval, line 1: I was wondering if there is a better message here. It would be odd if the user just misspelled "kissat" and they get this error.

celinval (commented):
I do wonder if we should make a distinction between a builtin solver and a custom one. So:
```
#[kani::solver("minisat")]
```
and 
```
#[kani::solver(custom="my_custom_sat")]
```
In that case, we could have the following argument structure:
```bash
# Run with kissat
$ kani --solver kissat
# Run with custom solver
$ kani --solver custom=my_sat
```

Also, instead of `custom`, I'm fine with `bin` or some other keyword.

zhassan-aws:
> I don't think the arguments to the annotation should be strings.

I updated the PR to not use strings, e.g. `#[kani::solver(kissat)]`. Custom solver binaries can be specified using `#[kani::solver(custom = "my_solver_binary")]`.

kani-compiler/src/kani_middle/attributes.rs
zhassan-aws: I removed this attribute.

kani-driver/src/args.rs
zhassan-aws: Changed to `Option<CbmcSolver>`.

zhassan-aws:
> I do wonder if we should make a distinction between a builtin solver and a custom one. So:

Done. This **only** works for the attribute though, but not the command-line option :( It seems that to support passing `--solver custom="solver_binary"`, we would need to parse the argument ourselves. In this case, we would lose the automatic listing of variants in the help message that `strum` provides.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -444,4 +449,41 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        if harness.solver.is_some() {
+            self.tcx.sess.span_err(
+                attr.span,
+                format!("Only one '{ATTRIBUTE}' attribute is allowed per harness"),
+            );
+            return;
+        }
+        match extract_string_arguments(attr) {
+            Ok(arg) => {
+                if arg.len() == 1 {
```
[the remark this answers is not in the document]
zhassan-aws: This code no longer exists.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -444,4 +449,41 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        attr.span,
+                        format!("The `{ATTRIBUTE}` attribute expects a single string argument. Got {} arguments.", arg.len()),
+                    );
+                }
+            }
+            Err(err) => {
+                let msg = match err {
+                    AttrError::Empty => format!("No arguments were specified to `{ATTRIBUTE}`."),
+                    AttrError::NonLiteral(_) => {
+                        format!("A non-literal argument was provided to `{ATTRIBUTE}`.")
```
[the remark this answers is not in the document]
zhassan-aws: This code no longer exists.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -45,6 +57,27 @@ pub fn extract_integer_argument(attr: &Attribute) -> Option<u128> {
     }
 }
 
+/// Extracts the string arguments from the attribute provided
+/// For example, `solver("foo")` return `Some("foo")`
+pub fn extract_string_arguments(attr: &Attribute) -> Result<Vec<String>, AttrError> {
```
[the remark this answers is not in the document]
zhassan-aws: Per our offline discussion, I filed https://github.com/model-checking/kani/issues/2192 for tracking the moving of argument validation to the macro.

kani-driver/src/call_cbmc.rs
```diff
@@ -186,6 +208,45 @@ impl KaniSession {
...
+                args.push("kissat".into());
+            }
+            CbmcSolver::Minisat => {
+                // Minisat is currently CBMC's default solver, so no need to
+                // pass any arguments
+            }
+            CbmcSolver::Custom(custom_solver) => {
+                // Check if the specified binary exists in path
+                if !Path::new(&custom_solver).exists() {
+                    eprintln!("Error: Specified solver \"{custom_solver}\" not found in path");
```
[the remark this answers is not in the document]
zhassan-aws: Can you clarify what you mean?

tests/ui/solver-attribute/multiple-args/expected
```diff
@@ -0,0 +1,2 @@
+error: The `#[kani::solver]` attribute expects a single string argument. Got 2 arguments.
```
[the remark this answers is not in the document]
zhassan-aws: Yes. I added those to the `expected` files.

tests/ui/solver-attribute/non-string/expected
```diff
@@ -0,0 +1,2 @@
+error: Invalid argument type. The `#[kani::solver]` attribute expects a string literal argument. Got Int(123, Unsuffixed).
```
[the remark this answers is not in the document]
zhassan-aws: This case was removed.

tests/ui/solver-attribute/not-found/expected
zhassan-aws: I couldn't think of a better one. Any suggestions?

kani_metadata/src/harness.rs
```diff
@@ -19,6 +19,8 @@ pub struct HarnessMetadata {
     pub original_start_line: usize,
     /// The line in that file where the proof harness ends.
     pub original_end_line: usize,
+    /// Optional data to store solver.
+    pub solver: Option<String>,
```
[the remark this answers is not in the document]
zhassan-aws: Done.

adpaco-aws:
> I updated the PR to not use strings, e.g. `#[kani::solver(kissat)]`. Custom solver binaries can be specified using `#[kani::solver(custom = "my_solver_binary")]`.

Thanks, @zhassan-aws !

One question: How would the user specify the solver in `#[kani::proof(solver = ...)]` if he wants to use a custom solver? I don't know if there's a syntax that'd support this case, since `custom` receives a string now, right?

zhassan-aws:
> How would the user specify the solver in `#[kani::proof(solver = ...)]` if he wants to use a custom solver?

I'm not sure :) We could perhaps do `kani::proof(solver = custom("solver-binary"))`.

tests/ui/solver-attribute/no-arg/expected
```diff
@@ -0,0 +1,6 @@
+error: The `#[kani::solver]` attribute expects a single argument. Got 0 arguments.
+test.rs:5:1
+  |
+5 | #[kani::solver]
```
celinval, line 4: nit: remove the line number from the expected files and add "\" to the end of each line.

celinval (commented):
It looks good to me. Just some minor tweaks of the error messages.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -445,4 +449,69 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    }
+
+    fn extract_solver_argument(&mut self, attr: &Attribute) -> Option<CbmcSolver> {
+        // TODO: Argument validation should be done as part of the kani_macros crate
+        // https://github.com/model-checking/kani/issues/2192
+        const ATTRIBUTE: &str = "#[kani::solver]";
+        let invalid_arg_err = |attr: &Attribute| {
+            self.tcx.sess.span_err(
+                attr.span,
+                format!("Invalid argument to `{ATTRIBUTE}` attribute. Expecting one of the supported solvers, e.g. `kissat` or a SAT solver binary, e.g. `custom=\"sat-solver-binary\"`.")
```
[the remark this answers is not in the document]
zhassan-aws: Done.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -445,4 +449,69 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        None
+                    }
+                }
+            }
+            MetaItemKind::NameValue(lit) if ident_str == "custom" && lit.kind.is_str() => {
+                //Some(CbmcSolver::Custom(meta_item.name_value_literal().unwrap().token_lit.symbol.to_string()))
+                Some(CbmcSolver::Custom(lit.token_lit.symbol.to_string()))
+            }
+            _ => {
+                invalid_arg_err(attr);
```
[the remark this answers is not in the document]
zhassan-aws: This message is not specific to `custom`. It is emitted when the argument received is neither a `MetaItemKind::Word` or `MetaItemKind::NameValue`.

kani-driver/src/call_cbmc.rs
```diff
@@ -186,6 +188,45 @@ impl KaniSession {
...
+                args.push("kissat".into());
+            }
+            CbmcSolver::Minisat => {
+                // Minisat is currently CBMC's default solver, so no need to
+                // pass any arguments
+            }
+            CbmcSolver::Custom(custom_solver) => {
+                // Check if the specified binary exists in path
+                if which::which(custom_solver).is_err() {
+                    eprintln!(
```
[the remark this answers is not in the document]
zhassan-aws: Removed the `eprintln` and moved its message to `bail`.

tests/ui/solver-attribute/no-arg/expected
zhassan-aws: Done.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -445,4 +449,68 @@ impl<'tcx> GotocCtx<'tcx> {
...
         }
     }
+
+    /// Set the solver for this proof harness
+    fn handle_kanitool_solver(&mut self, attr: &Attribute, harness: &mut HarnessMetadata) {
+        // Make sure the solver is not already set
+        if harness.solver.is_some() {
+            self.tcx
+                .sess
+                .span_err(attr.span, "Only one '#[kani::solver]' attribute is allowed per harness");
```
adpaco-aws, line 459: Can you change the error messages to start with a lowercase letter? I've been doing this following the [diagnostic output style guide](https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-output-style-guide) from `rustc`, and I'd like us to start applying a similar style (will post a PR after this review).

kani_metadata/src/cbmc_solver.rs
```diff
@@ -0,0 +1,33 @@
...
+    AsRefStr,
+    EnumString,
+    EnumVariantNames,
+    PartialEq,
+    Eq,
+    Serialize,
+    Deserialize
+)]
+#[strum(serialize_all = "snake_case")]
+pub enum CbmcSolver {
```
adpaco-aws, line 22: What's the reason for this to be `CbmcSolver` and not just `Solver`?

adpaco-aws (commented):
Just a few comments about error messages.

I'd like us to iterate on the best option to pass the path for `custom` solvers. In my opinion, there shouldn't be any differences between the CLI and the annotation, which AFAIK is not true at the moment. But I don't want this to block the contribution since it's a special, non-required case for me.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
zhassan-aws: Done

kani_metadata/src/cbmc_solver.rs
zhassan-aws: Since the solver is currently specific for CBMC, and we're not yet clear on future plans, and whether this solver will apply to other engines as well, I added CBMC to its name.

zhassan-aws:
> In my opinion, there shouldn't be any differences between the CLI and the annotation, which AFAIK is not true at the moment. 

I implemented a clap parser for the `--solver` option, so that we allow `--solver kissat`, `--solver minisat`, or `--solver custom=<any-binary>`. I added a couple of tests for it.

@celinval this is something that you have requested as well.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -445,4 +449,68 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                .sess
+                .span_err(attr.span, "only one '#[kani::solver]' attribute is allowed per harness");
+            return;
+        }
+        harness.solver = self.extract_solver_argument(attr);
+    }
+
+    fn extract_solver_argument(&mut self, attr: &Attribute) -> Option<CbmcSolver> {
+        // TODO: Argument validation should be done as part of the kani_macros crate
+        // https://github.com/model-checking/kani/issues/2192
```
adpaco-aws, line 467:
```suggestion
        // TODO: Argument validation should be done as part of the `kani_macros` crate
        // <https://github.com/model-checking/kani/issues/2192>
```

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -445,4 +449,68 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        const ATTRIBUTE: &str = "#[kani::solver]";
+        let invalid_arg_err = |attr: &Attribute| {
+            self.tcx.sess.span_err(
+                attr.span,
+                format!("invalid argument to `{ATTRIBUTE}` attribute. Expecting one of the supported solvers, e.g. `kissat` or a SAT solver binary, e.g. `custom=\"<SAT_SOLVER_BINARY>\"`.")
```
adpaco-aws, line 472:
```suggestion
                format!("invalid argument for `{ATTRIBUTE}` attribute, expected one of the supported solvers (e.g. `kissat`) or a SAT solver binary (e.g. `custom=\"<SAT_SOLVER_BINARY>\"`)")
```

kani-driver/src/call_cbmc.rs
```diff
@@ -186,6 +188,42 @@ impl KaniSession {
...
         args
     }
+
+    fn handle_solver_args(
+        &self,
+        harness_solver: &Option<CbmcSolver>,
+        args: &mut Vec<OsString>,
+    ) -> Result<()> {
+        let solver = if let Some(solver) = &self.args.solver {
+            // --solver option takes precedence over attributes
```
adpaco-aws, line 198:
```suggestion
            // `--solver` option takes precedence over attributes
```

kani_metadata/src/cbmc_solver.rs
```diff
@@ -0,0 +1,33 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use serde::{Deserialize, Serialize};
+use strum_macros::{AsRefStr, EnumString, EnumVariantNames};
+
+/// An enum for CBMC solver options. All variants are handled by Kani, except for
+/// the "Custom" one, which it passes as is to CBMC's `--external-sat-solver`
```
adpaco-aws, line 8:
```suggestion
/// the `Custom` one, which it passes as is to CBMC's `--external-sat-solver`
```

library/kani_macros/src/lib.rs
```diff
@@ -148,6 +148,29 @@ pub fn stub(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+
+/// Select the SAT solver to use with CBMC for this harness
+/// The attribute `#[kani::solver(arg)]` can only be used alongside `#[kani::proof]``
+///
+/// arg - name of solver, e.g. kissat
+#[cfg(kani)]
+#[proc_macro_attribute]
+pub fn solver(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut result = TokenStream::new();
+    // Translate #[kani::solver(arg)] to #[kanitool::stub(arg)]
```
adpaco-aws, line 166:
```suggestion
    // Translate `#[kani::solver(arg)]` to `#[kanitool::solver(arg)]`
```

tests/perf/btreeset/insert_same/src/main.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This test checks the performance of pushing the same element onto a BTreeSet
+//! The test is from https://github.com/model-checking/kani/issues/2022.
```
adpaco-aws, line 5:
```suggestion
//! This test checks the performance of pushing the same element onto a `BTreeSet`
//! The test is from <https://github.com/model-checking/kani/issues/2022>
```

adpaco-aws (approved):
Thanks, @zhassan-aws ! Looks great overall, just a few suggestions 😄

kani_metadata/src/cbmc_solver.rs
celinval: I tend to agree with @adpaco-aws here, but this is an implementation detail that is internal to our code, so we can easily rename this later.

tests/ui/solver-attribute/invalid/expected
```diff
@@ -0,0 +1,6 @@
+error: invalid argument for `#[kani::solver]` attribute, expected one of the supported solvers (e.g. `kissat`) or a SAT solver binary (e.g. `custom="<SAT_SOLVER_BINARY>"`)\
+test.rs:\
+  |\
+5 | #[kani::solver(123)]\
+  | ^^^^^^^^^^^^^^^^^^^^
```
celinval, line 5: Removing the line number:

```suggestion
|\
| #[kani::solver(123)]\
| ^^^^^^^^^^^^^^^^^^^^
```

tests/ui/solver-attribute/multiple-args/expected
```diff
@@ -0,0 +1,6 @@
+error: the `#[kani::solver]` attribute expects a single argument. Got 2 arguments.\
+test.rs:\
+  |\
+5 | #[kani::solver(kissat, minisat)]\
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
celinval, line 5:
```suggestion
|\
| #[kani::solver(kissat, minisat)]\
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

tests/ui/solver-attribute/multiple-attrs/expected
```diff
@@ -0,0 +1,6 @@
+error: only one '#[kani::solver]' attribute is allowed per harness\
+test.rs:\
+  |\
+5 | #[kani::solver(kissat)]\
+  | ^^^^^^^^^^^^^^^^^^^^^^^
```
celinval, line 5:
```suggestion
|\
| #[kani::solver(kissat)]\
| ^^^^^^^^^^^^^^^^^^^^^^^
```

tests/ui/solver-attribute/no-arg/expected
```diff
@@ -0,0 +1,6 @@
+error: the `#[kani::solver]` attribute expects a single argument. Got 0 arguments.\
+test.rs:\
+  |\
+5 | #[kani::solver]\
+  | ^^^^^^^^^^^^^^^
```
celinval, line 5:
```suggestion
|\
| #[kani::solver]\
| ^^^^^^^^^^^^^^^
```

tests/ui/solver-attribute/unknown/expected
```diff
@@ -0,0 +1,6 @@
+error: unknown solver `foo`\
+test.rs:\
+  |\
+5 | #[kani::solver(foo)]\
+  | ^^^^^^^^^^^^^^^^^^^^
```
celinval, line 5:
```suggestion
|\
| #[kani::solver(foo)]\
| ^^^^^^^^^^^^^^^^^^^^
```

celinval (commented):
It looks good. Just please remove the remaining line numbers from the expected tests.

PR #2089, opened 2023-01-06
Allow custom path for installation/setup

### LINKED ISSUE #1735: Kani doesn't provide an optional flag for alternative install path
Requested feature: cargo-kani setup to install to `$KANI_HOME` if set, and do what it does now (install to `$HOME/.kani`) otherwise.

Use case: I'm installing Kani in a docker container for use in a containerized development environment (Coder). The environment quashes the $HOME directory upon import of the container, so a pre-installed Kani at the default location for Kani (`$HOME/.kani`) doesn't exist after the container is launched.

Thank you!

adpaco-aws:
### Description of changes: 

Adds the option to install Kani using an alternative path by using the `KANI_HOME` environment variable.

The downside is that the variable needs to be set in subsequent Kani invocations, otherwise the proxy binaries will fail to detect a previous installation.

### Resolved issues:

Resolves #1735 

### Testing:

* How is this change tested? Tested manually and in CI (using a new docker)

### Checklist

### DIFF
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -19,22 +19,35 @@ const VERSION: &str = env!("CARGO_PKG_VERSION");
 const TARGET: &str = env!("TARGET");
 
 /// The directory where Kani is installed, either:
-///  * `${KANI_HOME}/kani-<VERSION>` if the environment variable
+///  * (custom) `${KANI_HOME}/kani-<VERSION>` if the environment variable
 ///    `KANI_HOME` is set.
-///  * `${HOME}/.kani/kani-<VERSION>` where `HOME` is the canonical definition
-///    of home directory used by Cargo and rustup.
+///  * (default) `${HOME}/.kani/kani-<VERSION>` where `HOME` is the canonical
+///    definition of home directory used by Cargo and rustup.
 pub fn kani_dir() -> Result<PathBuf> {
-    let (home_dir, origin) = match env::var("KANI_HOME") {
-        Ok(val) => (PathBuf::from(val), "KANI_HOME"),
-        Err(_) => (home::home_dir().expect("couldn't find home directory").join(".kani"), "HOME"),
+    let kani_dir = match env::var("KANI_HOME") {
+        Ok(val) => custom_kani_dir(val)?,
+        Err(_) => default_kani_dir()?,
     };
-    if !home_dir.is_dir() {
-        bail!(
-            "got home directory `{}` (from `{origin}`) which isn't a directory",
-            home_dir.display()
-        );
+    let kani_dir = kani_dir.join(format!("kani-{VERSION}"));
+    Ok(kani_dir)
+}
+
+/// Returns the custom Kani home directory: `${KANI_HOME}`
+fn custom_kani_dir(path: String) -> Result<PathBuf> {
+    let kani_dir = PathBuf::from(path);
+    if !kani_dir.is_dir() {
+        bail!("got custom Kani directory `{}` which isn't a directory", kani_dir.display());
     }
-    let kani_dir = home_dir.join(format!("kani-{VERSION}"));
+    Ok(kani_dir)
+}
+
+/// Returns the default Kani home directory: `${HOME}/.kani`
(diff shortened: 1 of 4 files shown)


src/setup.rs
```diff
@@ -18,19 +18,31 @@ const VERSION: &str = env!("CARGO_PKG_VERSION");
 /// Set by our `build.rs`, reflects the Rust target triple we're building for
 const TARGET: &str = env!("TARGET");
 
-/// Where Kani has been installed. Typically `~/.kani/kani-1.x/`
-pub fn kani_dir() -> PathBuf {
-    home::home_dir().expect("Couldn't find home dir?").join(".kani").join(format!("kani-{VERSION}"))
+/// The directory where Kani is installed, either:
+///  * `${KANI_HOME}/.kani/kani-<VERSION>` if the environment variable
```
celinval, line 22: As a user, I would expect `${KANI_HOME}/kani-<version>`. I.e., I wouldn't expect `.kani` to be inside the set `$KANI_HOME`. That is how `CARGO_HOME`, `RUSTUP_HOME`, and other tools usually work.

src/setup.rs
```diff
@@ -18,19 +18,31 @@ const VERSION: &str = env!("CARGO_PKG_VERSION");
...
+///    `KANI_HOME` is set.
+///  * `${HOME}/.kani/kani-<VERSION>` where `HOME` is the canonical definition
+///    of home directory used by Cargo and rustup.
+pub fn kani_dir() -> Result<PathBuf> {
+    let home_dir = match env::var("KANI_HOME") {
+        Ok(val) => PathBuf::from(val),
+        Err(_) => home::home_dir().expect("couldn't find home directory"),
+    };
+    if !home_dir.is_dir() {
+        bail!("got home directory which isn't a directory");
```
celinval, line 32: Can you please improve the error message to include the value of `home_dir`? You can also include whether you got it from KANI_HOME.

celinval (commented):
Just a few comments. Thanks

src/setup.rs
```diff
@@ -18,19 +18,34 @@ const VERSION: &str = env!("CARGO_PKG_VERSION");
...
+///  * `${KANI_HOME}/kani-<VERSION>` if the environment variable
+///    `KANI_HOME` is set.
+///  * `${HOME}/.kani/kani-<VERSION>` where `HOME` is the canonical definition
+///    of home directory used by Cargo and rustup.
+pub fn kani_dir() -> Result<PathBuf> {
+    let (home_dir, origin) = match env::var("KANI_HOME") {
+        Ok(val) => (PathBuf::from(val), "KANI_HOME"),
+        Err(_) => (home::home_dir().expect("couldn't find home directory").join(".kani"), "HOME"),
+    };
+    if !home_dir.is_dir() {
```
celinval, line 31: Can you please test if this works even when the user is first installing Kani. I.e., delete `~/.kani` and try to run setup.

I'm asking because on line `54` we are explicitly creating `kani_dir` if it doesn't exist.

src/setup.rs
adpaco-aws: Makes a lot of sense, thanks! Changed to add "/.kani" on the `HOME` arm only.

celinval (commented):
I just noticed one thing. Can you check please if this patch won't break setup if the `${HOME}/.kani` doesn't exist yet?

src/setup.rs
celinval: Actually, I think the regression is failing because of that: https://github.com/model-checking/kani/actions/runs/3858944820/jobs/6578010318

adpaco-aws: There's an easy way to fix this (pushing ".kani" after the check), but I think the code & error message are too confusing then. Let me try something else.

adpaco-aws: Added two separate functions to get the custom/default Kani home directory since the two cases are treated differently in terms of error handling. This is hopefully clearer and works.

celinval (approved):
I would prefer if we don't fail if `$KANI_HOME` doesn't exist, since we try to create the folder later on. But not a blocker. Thanks!

PR #2092, opened 2023-01-09
Add a workflow that runs regressions nightly with CBMC's latest commit

### LINKED ISSUE #2090: Mini RFC: Run the Kani regressions nightly with CBMC's tip of tree
**Motivation:** We've now had multiple occasions where we failed to integrate a new CBMC release because it causes our regressions to fail, due to a CBMC bug, or due to a breaking change that requires some changes to the Kani code. In the former case, we often have to wait two weeks for the next CBMC release to come out with the fix. If those issues are caught earlier, it would give us more time to investigate/fix the issues rather than learning about the failures after the release comes out.

**Proposed change:** Create a new nightly workflow that checks out CBMC, builds it, and runs the Kani regressions with the built version. The result can be displayed on the GitHub page using a status badge and/or an e-mail can be sent to the team if there's a failure.

The following command sequence can be used:
```
git clone --depth=1 https://github.com/diffblue/cbmc
cd cbmc/
git submodule update --init
cmake -S . -Bbuild -DWITH_JBMC=OFF
cmake --build build -- -j 4
# Run Kani regressions (at least the normal and perf regressions)
```
These need to be done both on Ubuntu and macos.

zhassan-aws:
### Description of changes: 

Add a nightly workflow that checks out the CBMC repo, builds it, and runs the Kani regressions with it. This is to detect any issues that the upcoming CBMC release may introduce and address them in a timely manner.

The result of the run is displayed in a status badge on the main page.

See how the badge is displayed [here](https://github.com/zhassan-aws/kani#readme), and the results of running the action [here](https://github.com/zhassan-aws/kani/actions/runs/3875659522).

### Resolved issues:

Resolves #2090 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? Tested in my fork. 

### Checklist


.github/workflows/cbmc-latest.yml
```diff
@@ -0,0 +1,63 @@
...
+        uses: ./.github/actions/setup
+        with:
+          os: ubuntu-20.04
+
+      - name: Build Kani using release mode
+        run: cargo build-dev -- --release
+
+      - name: Checkout and build CBMC
+        run: |
+          git clone --depth=1 https://github.com/diffblue/cbmc
```
celinval, line 56: FYI, you could use the checkout action to replace all the git calls.

celinval (approved):

.github/workflows/cbmc-latest.yml
zhassan-aws: Done.

.github/workflows/cbmc-latest.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Checkout CBMC under "cbmc"
+        uses: actions/checkout@v3
+        with:
+          repository: diffblue/cbmc
+          path: cbmc
+
+      - name: Build CBMC
+        run: |
+          cd cbmc
+          cmake -S . -Bbuild -DWITH_JBMC=OFF
```
zhassan-aws, line 38: @feliperodri @tautschnig is this a good recipe for building CBMC for Kani?

PR #2102, opened 2023-01-11
Add `exec` mode for `compiletest`

### LINKED ISSUE #2059: Compiletest mode for executing arbitrary commands
We have several ad-hoc bash scripts for testing in Kani, it would be nice if we had a compiletest mode that:

1. Could execute an arbitrary command, and
2. Performed `expect` test validation on the output.

Given:

* #1895 

This solution should probably check exit status (for a particular value, just default 0) as well.

adpaco-aws:
### Description of changes: 

This PR adds a new `exec` mode for `compiletest`. This mode is useful to execute script-based tests, and allows checking expected output and exit codes at the same time.

Unlike `cargo-kani` tests, this mode expects a folder with a `config.yml` file within (not adding one will cause an error). The `config.yml` file should contain:
 * `script`: The path to the script to be executed.
 * `expected` (optional): The path to the `.expected` file to use for output comparison.
 * `exit_code` (optional): The exit code to be returned by executing the script (a zero exit code is expected if not specified).

A new test `error-code` has been added with the purpose of testing the `exec` mode itself. This PR also migrates the `check-output` test that used to run at the beginning of our regression. Other scripts can be migrated in subsequent PRs, but I'm not doing it here since it might require some work.

### Resolved issues:

Resolves #2059 

### Call-outs:

 * `compiletest` actually runs `time script`, since running `script` alone didn't seem to be successful. However, many of our existing script-based tests run with `time`, so I figured that's the best to do there (previously was just running `sh script`).

### Testing:

* How is this change tested? CI + new tests.

### Checklist


tests/exec-init/check-output/.gitignore
```diff
@@ -1,4 +1,4 @@
 target/
 Cargo.lock
 *.c
-build/
+build/
```
zhassan-aws, line 4: Is the removal of the new line intentional?

tools/compiletest/src/main.rs
```diff
@@ -385,6 +386,46 @@ fn collect_expected_tests_from_dir(
     Ok(())
 }
 
+/// Collect `exec` tests from a directory.
+///
+/// Note that this method isn't recursive like
+/// `collect_expected_tests_from_dir`, as it allows us to ensure that newly
```
zhassan-aws, line 392: It is nice to be able to check that newly tests are running, but the inability to add tests under sub-directories could be a bit restrictive.

tools/compiletest/src/main.rs
```diff
@@ -385,6 +386,46 @@ fn collect_expected_tests_from_dir(
...
+    // Ensure the mode we're running is `Mode::Exec`
+    assert!(config.mode == Mode::Exec);
+
+    // Each test is expected to be in its own folder (i.e., we don't do recursion)
+    for file in fs::read_dir(dir)? {
+        // Look for `config.yml` file in folder
+        let file = file?;
+        let file_path = file.path();
+        let has_config_yml = file_path.join("config.yml").exists();
+        assert!(has_config_yml, "couldn't find `config.yml` file for `exec` test");
```
zhassan-aws, line 419: Can we error out instead of panic?

tools/compiletest/src/runtest.rs
```diff
@@ -310,6 +324,57 @@ impl<'test> TestCx<'test> {
...
 
+    /// Runs an executable file and:
+    ///  * Checks the expected output if an expected file is specified
+    ///  * Checks the exit code (assumed to be 0 by default)
+    fn run_exec_test(&self) {
+        // Open the `config.yml` file and extract its values
+        let path_yml = self.testpaths.file.join("config.yml");
+        let config_file = std::fs::File::open(path_yml).expect("couldn't open `config.yml`");
+        let exec_config: ExecConfig =
+            serde_yaml::from_reader(config_file).expect("couldn't read `config.yml` values");
```
zhassan-aws, line 335: Panicking would make it hard to debug the issues with the yaml file. Perhaps we should emit the error returned by `serde_yaml::from_reader`?

tools/compiletest/src/runtest.rs
```diff
@@ -310,6 +324,57 @@ impl<'test> TestCx<'test> {
...
+
+        // Compare with expected output if it was provided
+        if let Some(output) = expected_output {
+            self.verify_output(&proc_res, &output);
+        }
+
+        // Compare with exit code (0 if it wasn't provided)
+        let expected_code = exec_config.code.or(Some(0));
+        if proc_res.status.code() != expected_code {
+            self.fatal_proc_rec("test failed: expected different code", &proc_res);
```
zhassan-aws, line 374: Can we add the expected code and the actual code to the message?

zhassan-aws (commented):
Thanks! I have a few comments. Can we also add this to the documentation?

tools/compiletest/src/main.rs
adpaco-aws: I tried using the `error!` macro, but this prints an error message (see 2nd line below) and continues, only to panic when executing the individual tests.

```
Check compiletest suite=exec-init mode=exec
2023-01-11T18:58:20.278416Z ERROR compiletest: couldn't find `config.yml` file for `exec` test

running 2 tests
F.
failures:

---- [exec] exec-init/error-code stdout ----
thread '[exec] exec-init/error-code' panicked at 'couldn't open `config.yml`: Os { code: 2, kind: NotFound, message: "No such file or directory" }', tools/compiletest/src/runtest.rs:333:57
```

In my opinion, it's better to have an early failure than waiting for the test to be run. Do you have a specific suggestion to replace the `assert!`?

tests/exec-init/check-output/.gitignore
adpaco-aws: No (restored).

tools/compiletest/src/runtest.rs
adpaco-aws: Done.

tools/compiletest/src/runtest.rs
adpaco-aws: Done.

tools/compiletest/src/main.rs
adpaco-aws: Yes, but I wouldn't like to allow tests like these ones (which are quite complex) to be missed. Of course, this can be changed if it becomes too restrictive.

adpaco-aws:
Thanks @zhassan-aws for the review! I'm still working on documentation for this, but feel free to provide more comments in the meantime.

tools/compiletest/src/main.rs
zhassan-aws: Yes, it's better to fail early. Perhaps use something similar to `fatal_proc_rec` in `runtest.rs`?

adpaco-aws:
Added some documentation in the "regression testing" section and rebased the PR against `main`.

docs/src/regression-testing.md
```diff
@@ -56,7 +56,16 @@ In particular, the Kani testing suites are composed of:
...
                 for each harness in the package.
-* `cargo-ui`: Similar to `cargo-kani`, but focuses on the user interface like the `ui` test suite.
+ * `cargo-ui`: Similar to `cargo-kani`, but focuses on the user interface like the `ui` test suite.
+ * `exec-init`: This suite is useful to execute script-based tests, and also
+                allows checking expected output and exit codes after running them. The suite
+                expects a folder with a `config.yml` file within, which should contain:
+                 - `script`: The path to the script to be executed.
+                 - `expected` (optional): The path to the `.expected` file to
+                 use for output comparison.
+                 - `code` (optional): The exit code to be returned by executing
```
zhassan-aws, line 66: Can we rename this to `exit_code` to make it clearer?

docs/src/regression-testing.md
```diff
@@ -56,7 +56,16 @@ In particular, the Kani testing suites are composed of:
                 Arguments can be specified in the `Cargo.toml` configuration file.
                 Similar to the `expected` suite, we look for `*.expected` files
                 for each harness in the package.
-* `cargo-ui`: Similar to `cargo-kani`, but focuses on the user interface like the `ui` test suite.
+ * `cargo-ui`: Similar to `cargo-kani`, but focuses on the user interface like the `ui` test suite.
+ * `exec-init`: This suite is useful to execute script-based tests, and also
+                allows checking expected output and exit codes after running them. The suite
+                expects a folder with a `config.yml` file within, which should contain:
```
zhassan-aws, line 62: Can we add an example `config.yml` below?

zhassan-aws, line 62: We should clarify the restriction on the directory structure, e.g. "Tests are expected to be placed directly under `exec-init` in a directory with a `config.yml`.".

scripts/kani-regression.sh
```diff
@@ -41,13 +41,9 @@ cargo test -p kani-compiler
...
 cargo test -p kani_metadata
 
-# Check output files (--gen-c option)
-echo "Check GotoC output file generation"
-time "$KANI_DIR"/tests/output-files/check-output.sh
-echo ""
-
 # Declare testing suite information (suite and mode)
 TESTS=(
+    "exec-init exec"
```
zhassan-aws, line 46: Are we planning to add the other script-based scripts to the same suite? If so, maybe rename `exec-init` to `script-based` or something more generic?

tools/compiletest/src/runtest.rs
```diff
@@ -310,6 +324,66 @@ impl<'test> TestCx<'test> {
...
+    ///  * Checks the expected output if an expected file is specified
+    ///  * Checks the exit code (assumed to be 0 by default)
+    fn run_exec_test(&self) {
+        // Open the `config.yml` file and extract its values
+        let path_yml = self.testpaths.file.join("config.yml");
+        let config_file = std::fs::File::open(path_yml).expect("couldn't open `config.yml`");
+        let exec_config_res = serde_yaml::from_reader(config_file);
+        if let Err(error) = &exec_config_res {
+            let err_msg = format!("couldn't parse `config.yml` file: {error}");
+            self.error(&err_msg);
```
zhassan-aws, line 337: Will this continue execution after the error? If so, then it would be better to use `fatal_proc_rec` so that this doesn't go unnoticed.

tools/compiletest/src/runtest.rs
```diff
@@ -310,6 +324,66 @@ impl<'test> TestCx<'test> {
...
+        }
+        let exec_config: ExecConfig = exec_config_res.unwrap();
+
+        // Check if the `script` file exists
+        let script_rel_path = PathBuf::from(exec_config.script);
+        let script_path = self.testpaths.file.join(script_rel_path);
+        if !script_path.exists() {
+            let err_msg = format!("test failed: couldn't find script in {}", script_path.display());
+            self.error(&err_msg);
```
zhassan-aws, line 346: Same here.

docs/src/regression-testing.md
adpaco-aws: Renamed.

docs/src/regression-testing.md
adpaco-aws: I felt like this bullet point was getting too complex, so I made a subsection in `Script-based tests` where I provide more details on how it works, including an example `config.yml` file.

adpaco-aws: Done (see new subsection mentioned in `Script-based tests`).

scripts/kani-regression.sh
adpaco-aws: Renamed it to `script-based-init` - the `init` is supposed to indicate that these are the initial tests, because I intend to migrate the script-based tests that occur in the end (e.g., `std` test) to another suite `script-based-end`. How does that sound?

tools/compiletest/src/runtest.rs
adpaco-aws: You're right, it continues execution. But we couldn't use `fatal_proc_rec` because one of the arguments is `proc_res` which is the process we spawn to run the test. So I used the `fatal_error` function added in the previous revision.

With this change, the output (if I remove `config.yml` for one test) shows like this:
```
running 2 tests
F.
failures:

---- [exec] script-based-init/error-code stdout ----
error: test failed: couldn't find script in /home/ubuntu/kani/tests/script-based-init/error-code/exit-one.sh
```

tools/compiletest/src/runtest.rs
adpaco-aws: Done (same as above).

scripts/kani-regression.sh
zhassan-aws: Nit: In this case "script-based-pre" might be a bit clearer.

tools/compiletest/src/runtest.rs
```diff
@@ -376,7 +376,7 @@ impl<'test> TestCx<'test> {
         let expected_code = exec_config.exit_code.or(Some(0));
         if proc_res.status.code() != expected_code {
             let err_msg = format!(
-                "test failed: expected code {}, got code {}",
+                "test failed: exp   ected code {}, got code {}",
```
zhassan-aws, line 379: Typo

docs/src/regression-testing.md
```diff
@@ -166,3 +171,50 @@ In fact, most of them are equivalent to running `cargo kani` and performing
 checks on the output. The downside to scripting is that these tests will always
 be run, even if there have not been any changes since the last time the
 regression was run.
+
+> **NOTE**: With the addition of the `exec` mode for `compiletest` (described
+> below), we'll be migrating these script-based tests to other suites using the
+> `exec` mode. The `exec` mode allows us to take advantage of `compiletest`
+> features while executing script-based tests (e.g., parallel execution).
+
+### The `exec` mode
```
zhassan-aws, line 180: Much clearer it in a separate section! Thanks!

zhassan-aws (approved):

scripts/kani-regression.sh
adpaco-aws: Sure, done.

tools/compiletest/src/runtest.rs
adpaco-aws: Oops. Thanks!

PR #2103, opened 2023-01-11
Add a kani::any_value_where function to enable easy generation of constrained symbolic values

danielsn:
### Description of changes: 

Currently, users must first create a value using `kani::any`, then constain it using `kani::assume`.  This is unergonomic.  
### Resolved issues:

Resolves #ISSUE-NUMBER

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

Feel free to bikeshed the name

### Testing:

* How is this change tested? New unit test

### Checklist

### DIFF
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -109,6 +109,31 @@ pub fn any<T: Arbitrary>() -> T {
     T::any()
 }
 
+/// This creates an symbolic *valid* value of type `T`.
+/// The value is constrained to be a value accepted by the predicate passed to the filter.
+/// You can assign the return value of this function to a variable that you want to make symbolic.
+/// The explanation field gives a mechanism to explain why the assumption is required for the proof.
+///
+/// # Example:
+///
+/// In the snippet below, we are verifying the behavior of the function `fn_under_verification`
+/// under all possible `NonZeroU8` input values between 0 and 12.
+///
+/// ```rust
+/// let inputA = kani::filter_any::<std::num::NonZeroU8>(|x| x < 12, "explanation");
+/// fn_under_verification(inputA);
+/// ```
+///
+/// Note: This is a safe construct and can only be used with types that implement the `Arbitrary`
+/// trait. The Arbitrary trait is used to build a symbolic value that represents all possible
+/// valid values for type `T`.
+#[inline(always)]
+pub fn filter_any<T: Arbitrary, F: FnOnce(&T) -> bool>(f: F, _msg: &'static str) -> T {
+    let result = T::any();
+    assume(f(&result));
+    result
+}
+
 /// This function creates a symbolic value of type `T`. This may result in an invalid value.
 ///
 /// # Safety
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -120,7 +120,7 @@ pub fn any<T: Arbitrary>() -> T {
 /// under all possible `NonZeroU8` input values between 0 and 12.
 ///
 /// ```rust
-/// let inputA = kani::any_with::<std::num::NonZeroU8>(|x| *x < 12, "explanation");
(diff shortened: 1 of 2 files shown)


zhassan-aws:
Thanks! Are the updates to the `s2n-quic` submodule intentional?

danielsn:
> Thanks! Are the updates to the `s2n-quic` submodule intentional?

No, removed them

library/kani/src/lib.rs
```diff
@@ -109,6 +109,31 @@ pub fn any<T: Arbitrary>() -> T {
...
+/// You can assign the return value of this function to a variable that you want to make symbolic.
+/// The explanation field gives a mechanism to explain why the assumption is required for the proof.
+///
+/// # Example:
+///
+/// In the snippet below, we are verifying the behavior of the function `fn_under_verification`
+/// under all possible `NonZeroU8` input values between 0 and 12.
+///
+/// ```rust
+/// let inputA = kani::filter_any::<std::num::NonZeroU8>(|x| x < 12, "explanation");
```
zhassan-aws, line 123: I think the predicate needs to dereference x: `*x < 12`.

library/kani/src/lib.rs
```diff
@@ -109,6 +109,31 @@ pub fn any<T: Arbitrary>() -> T {
...
+/// fn_under_verification(inputA);
+/// ```
+///
+/// Note: This is a safe construct and can only be used with types that implement the `Arbitrary`
+/// trait. The Arbitrary trait is used to build a symbolic value that represents all possible
+/// valid values for type `T`.
+#[inline(always)]
+pub fn filter_any<T: Arbitrary, F: FnOnce(&T) -> bool>(f: F, _msg: &'static str) -> T {
```
zhassan-aws, line 131: I'm not sure if the message adds any value since it's not used anywhere (vs. writing a comment).

zhassan-aws (commented):
I have a few comments.

Here's my bikeshedding 😄:  `kani::any_with` or `kani::any_such_that`.

adpaco-aws:
> Here's my bikeshedding 😄: `kani::any_with` or `kani::any_such_that`.

THIS. I'd prefer `kani::any_with` over `kani::any_such_that` because it's shorter and avoids mathematical jargon.

tedinski:
Thirding the bikeshedding over the name `filter_any` doesn't seem right to me. I like `any_with`

proptest has an `any_with`, but it works differently: it allows supplying arguments to the generator/strategy being used. In practice, it seems like these are often constraints: e.g. size for a vector. So I think it fits in the spirit here: we don't need/have a generator, and so we can just supply arbitrary constraints directly. So I count this as precedent for this choice of name.

zhassan-aws:
I like `any_where` as well, but agree it's a bit funny.

`crux-mir` uses `symbolic_where`: https://github.com/GaloisInc/crucible/blob/79428012dae4e3c922f75c6f41f418c8aacca665/crux-mir/lib/crucible/symbolic.rs#L8

tests/kani/Assume/main.rs
```diff
@@ -7,3 +7,9 @@ fn main() {
     kani::assume(i < 10);
     assert!(i < 20);
 }
+
+#[kani::proof]
+fn verify_filter_assume() {
+    let i: i32 = kani::filter_any(|x| *x < 10, "Only single digit values are legal");
```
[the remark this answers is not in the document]
celinval: rust doesn't support optional parameters. We would have to introduce a `any!()` macro for that.

zhassan-aws:
Yet another option: `any_value_where`.

rahulku:
We need to pick one and move on. 

@danielsn please advise.

danielsn:
I like @zhassan-aws's suggestion of `any_value_where`

rahulku (approved):

PR #2110, opened 2023-01-12
Fix deserialization of property IDs for special case

### LINKED ISSUE #2106: Kani regression test expected/dynamic-error-trait/main.rs fails with latest CBMC
**Proposed change:**
Using diffblue/cbmc@6a7b97cc89 (which is current develop, and between the already released 5.74.0 and the upcoming 5.74.0) results in a regression test failure that I believe is due to a limitation in Kani. Running the regression tests I get:
```
[...]

stderr:
warning: unused imports: `Read`, `Write`
 --> /home/ub-backup/tautschn/kani.git/tests/expected/dynamic-error-trait/main.rs:6:21
  |
6 | use std::io::{self, Read, Write};
  |                     ^^^^  ^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: field `addr` is never read
  --> /home/ub-backup/tautschn/kani.git/tests/expected/dynamic-error-trait/main.rs:11:5
   |
10 | pub struct MemoryMapping {
   |            ------------- field in this struct
11 |     addr: *mut u8,
   |     ^^^^
   |

             - caller_location (1)

warning: 3 warnings emitted

thread '<unnamed>' panicked at 'Found property counter that doesn't match number format', kani-driver/src/cbmc_output_parser.rs:172:9

failures:
    [expected] expected/dynamic-error-trait/main.rs

test result: FAILED. 159 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 15.16s

Some tests failed in compiletest suite=expected mode=expected host=(none) target=(none)
```
Looking at the CBMC output, and comparing it to an earlier version, the only difference is:
* New: `[_ZN4core3ptr42drop_in_place$LT$std..io..error..Error$GT$17hd75fb0a3517a80c2E.recursion] recursion unwinding assertion: SUCCESS`
* Before: `[.recursion] recursion unwinding assertion: SUCCESS`

The latter seems bogus, and the new output produced by CBMC strictly seems to be an improvement. Either the output parser needs fixing, or we need to check why this mangled function name is seen in the output in the first place, which then perhaps returns the output to something that _is_ compatible with Kani's expected property format.

**Motivation:**
diffblue/cbmc@5c4503f2ef2fd altered the way we compute some property names. CBMC reserves the discretion to change property names, as long as they remain unique. On this occasion, however, it seems the previous property name was bogus, and Kani just seemed to apply a workaround.

adpaco-aws:
### Description of changes: 

Extends the deserialization of property IDs to handle the "recursion" special case so it now accepts both
 * `<function>.recursion` (new in CBMC 5.74.0)
 * `.recursion` (in earlier CBMC versions)
as valid property IDs and returns the expected values.

One unit test with this special case is included.

### Resolved issues:

Resolves #2106 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? Existing tests + new one.

### Checklist

### DIFF
--- a/kani-driver/src/cbmc_output_parser.rs
+++ b/kani-driver/src/cbmc_output_parser.rs
@@ -133,7 +133,7 @@ impl<'de> serde::Deserialize<'de> for PropertyId {
         //
         // As of CBMC 5.74.0, the property ID is `<function>.recursion`.
         // In earlier versions, it would just be `.recursion`.
-        if id_str.ends_with(".recursion")  {
+        if id_str.ends_with(".recursion") {
             let attributes: Vec<&str> = id_str.splitn(2, '.').collect();
             let fn_name = if attributes[0] == "" {
                 None
@@ -639,6 +639,38 @@ mod tests {
             trace: None,
         };
         assert_eq!(dummy_prop.property_name(), "recursion.1");
+    }
+
+    #[test]
+    fn check_property_id_deserialization_special_name() {
+        let prop_id_string = "\"alloc::raw_vec::RawVec::<u8>::allocate_in.recursion\"";
+        let prop_id_result: Result<PropertyId, serde_json::Error> =
+            serde_json::from_str(prop_id_string);
+        let prop_id = prop_id_result.unwrap();
+        assert_eq!(
+            prop_id.fn_name,
+            Some(String::from("alloc::raw_vec::RawVec::<u8>::allocate_in"))
+        );
+        assert_eq!(prop_id.class, String::from("recursion"));
+        assert_eq!(prop_id.id, 1);
+
+        let dummy_prop = Property {
+            description: "".to_string(),
+            property_id: prop_id,
+            source_location: SourceLocation {
+                function: None,
+                file: None,
+                column: None,
+                line: None,
+            },
+            status: CheckStatus::Success,
+            reach: None,
+            trace: None,
+        };
+        assert_eq!(
+            dummy_prop.property_name(),
+            "alloc::raw_vec::RawVec::<u8>::allocate_in.recursion.1"
+        );
     }
 
     #[test]
--- a/kani-driver/src/cbmc_output_parser.rs
+++ b/kani-driver/src/cbmc_output_parser.rs
@@ -127,15 +127,15 @@ impl<'de> serde::Deserialize<'de> for PropertyId {
     {
         let id_str = String::deserialize(d)?;
 
-        // Handle two special cases that doen't respect the format, and appears
-        // at least in the test `tests/expected/dynamic-error-trait/main.rs`
-        // with the description "recursion unwinding assertion".
+        // Handle a special case that doesn't respect the format, and appears at
+        // least in the test `tests/expected/dynamic-error-trait/main.rs` with
+        // the description "recursion unwinding assertion".
         //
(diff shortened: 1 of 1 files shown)


celinval (approved):
Interesting. Question unrelated to the PR, do we handle this the same way as we handle unwind assertion failures?

adpaco-aws:
> Do we handle this the same way as we handle unwind assertion failures?

Just checked this. They aren't considered when we determine if there were failed unwinding assertions during post-processing.

zhassan-aws:
> Do we handle this the same way as we handle unwind assertion failures?

I could swear they were in the past, but I just checked, and indeed, they're not handled the same!
```rust
fn factorial(x: u32) -> u32 {
    if x == 1 {
        x
    } else {
        x * factorial(x - 1)
    }
}


#[kani::proof]
#[kani::unwind(3)]
fn main() {
    let x = 5;
    assert_eq!(x, 5);
    let f = factorial(x);
    assert_eq!(f, 120);
}
```

```bash
$ kani recursion.rs 
Checking harness main...
CBMC 5.72.0 (cbmc-5.72.0)
CBMC version 5.72.0 (cbmc-5.72.0) 64-bit x86_64 linux
Reading GOTO program from file /home/ubuntu/examples/recursion.for-main.out
Generating GOTO Program
Unwinding recursion factorial iteration 1
Unwinding recursion factorial iteration 2
Unwinding recursion factorial iteration 3
Not unwinding recursion factorial iteration 4
Runtime Symex: 0.00701862s
size of program expression: 407 steps
slicing removed 258 assignments
Generated 11 VCC(s), 6 remaining after simplification
Runtime Postprocess Equation: 6.0421e-05s
converting SSA
Runtime Convert SSA: 0.000235395s
Post-processing
Runtime Post-process: 7.561e-06s
0 variables, 0 clauses
Runtime Solver: 4.6641e-05s
Runtime decision procedure: 0.000357648s

RESULTS:
Check 1: main.assertion.1
         - Status: SUCCESS
         - Description: "assertion failed: x == 5"
         - Location: recursion.rs:14:5 in function main

Check 2: main.assertion.2
         - Status: UNREACHABLE
         - Description: "assertion failed: f == 120"
         - Location: recursion.rs:16:5 in function main

Check 3: factorial.assertion.1
         - Status: SUCCESS
         - Description: "attempt to subtract with overflow"
         - Location: recursion.rs:5:23 in function factorial

Check 4: factorial.assertion.2
         - Status: UNREACHABLE
         - Description: "attempt to multiply with overflow"
         - Location: recursion.rs:5:9 in function factorial

Check 5: recursion.1
         - Status: FAILURE
         - Description: "recursion unwinding assertion"


SUMMARY:
 ** 1 of 5 failed (2 unreachable)
Failed Checks: recursion unwinding assertion

VERIFICATION:- FAILED
Verification Time: 0.027254s
```
Notice how some checks were still reported as successful, and the missing message to increase the unwind value.

celinval:
We should create an issue to fix that. @adpaco-aws can you please do that?

adpaco-aws:
> We should create an issue to fix that. @adpaco-aws can you please do that?

Almost got the PR to fix. Thanks for the test, @zhassan-aws !

PR #2117, opened 2023-01-12
Allow `&bool` in asserts

### LINKED ISSUE #2108: Kani assert override is more restrictive than standard assert
This code works as a test but fails with Kani:

```rust
#[cfg_attr(test, test)]
fn check() {
    let v: bool = true;
    assert!(&v);
}
```

With test: builds successfully, no warnings/errors, test passes.

With kani:

```text
  --> test_bool_ref.rs:7:13
   |
7  |     assert!(&v);
   |     --------^^-
   |     |       |
   |     |       expected `bool`, found `&bool`
   |     arguments to this function are incorrect
   |
note: function defined here
  --> /home/ubuntu/rmc/library/kani/src/lib.rs:62:14
   |
62 | pub const fn assert(_cond: bool, _msg: &'static str) {
   |              ^^^^^^
```

zhassan-aws:
### Description of changes: 

From investigating #2108, it turns out that due to how the standard library expands `assert!(x)` as
```rust
    if !x {
        ::core::panicking::panic("assertion failed: x")
    };
```
the following is valid Rust code:
```rust
fn main() {
    let b: bool = true;
    assert!(&b);
}
```
because it gets expanded to:
```rust
fn main() {
    let b: bool = true;
    if !&b {
        ::core::panicking::panic("assertion failed: &b")
    }
}
```
which, however, is rejected by Kani because `&b` is not a boolean (only `!&b` is!).

This PR adds a hacky fix, which is to inject `!!` (i.e. double negation) before the condition.

### Resolved issues:

Resolves #2108 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

TBH, I'm not sure about whether this should be merged as it is quite hacky. We should look into where this occurs, and why it is valid.

### Testing:

### Checklist

### DIFF
--- a/library/std/src/lib.rs
+++ b/library/std/src/lib.rs
@@ -44,10 +44,11 @@ pub mod process;
 #[macro_export]
 macro_rules! assert {
     ($cond:expr $(,)?) => {
-        kani::assert($cond, concat!("assertion failed: ", stringify!($cond)));
+        // The double negation is to resolve https://github.com/model-checking/kani/issues/2108
+        kani::assert(!!$cond, concat!("assertion failed: ", stringify!($cond)));
     };
     ($cond:expr, $($arg:tt)+) => {{
-        kani::assert($cond, concat!(stringify!($($arg)+)));
+        kani::assert(!!$cond, concat!(stringify!($($arg)+)));
         // Process the arguments of the assert inside an unreachable block. This
         // is to make sure errors in the arguments (e.g. an unknown variable or
         // an argument that does not implement the Display or Debug traits) are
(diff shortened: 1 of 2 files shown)


celinval (commented):
Did you try defining `kani::assert()` using `AsRef<bool>` instead?

celinval:
> Did you try defining `kani::assert()` using `AsRef<bool>` instead?

Actually, I don't think that will work. :(

celinval (approved):

tedinski:
Seems fine, can we change the PR name to something slightly more customer-facing in advance of merging? "Accept `&bool` in asserts, like Rust's assert macro does" perhaps

zhassan-aws:
> can we change the PR name to something slightly more customer-facing in advance of merging?

Done.

PR #2127, opened 2023-01-17
Remove `--legacy-linker` support

### LINKED ISSUE #2126: Remove `--legacy-linker`
**Proposed change:** Remove support to `--legacy-linker`. As far as I know, no user should be using `--legacy-linker` anymore. Internally, we still use it for testing codegen of crates that don't have harnesses. We should start using `assess` for that or other existing facilities.

**Motivation:** The legacy linker requires special handling inside `kani-driver` which is no longer needed. This will simplify some of the code.

celinval:
### Description of changes: 

I updated the firecracker codegen test to use `cargo kani assess` instead and I removed the old scripts to collect info on the top X crates since we no longer use them.

I kept the compiler option `--reachability=legacy` since we still use it to test the std library code generation. I also kept the `--mir-linker` option as a no-op for now.

### Resolved issues:

Fixes #2126 

### Related RFC:

N/A

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


zhassan-aws (approved):
Happy to see this go. Thanks.

kani-driver/src/args.rs
```diff
@@ -135,19 +135,9 @@ pub struct KaniArgs {
     #[arg(long, hide_short_help = true)]
     pub only_codegen: bool,
 
-    /// Disable the new MIR Linker. Using this option may result in missing symbols from the
-    /// `std` library. See <https://github.com/model-checking/kani/issues/1213> for more details.
+    /// Enable the new MIR Linker. This is a no-op since we no longer support any other linker.
     #[arg(long, hide = true)]
-    pub legacy_linker: bool,
```
tedinski, line 141: I'm fine with removing it outright, but as a pattern for future removals, maybe we should leave the flag but issue warnings/errors (i.e. a good error message is better than a bad one. "This flag no longer does anything, remove it" is better than "unknown flag")

But I suspect no one is using this one but us. Still, maybe we should add the warning (or error, probably) for `--mir-linker` now?

kani-driver/src/project.rs
```diff
@@ -152,8 +149,8 @@ pub fn cargo_project(session: &KaniSession) -> Result<Project> {
     let outputs = session.cargo_build()?;
     let mut artifacts = vec![];
     let outdir = outputs.outdir.canonicalize()?;
-    if session.args.legacy_linker || session.args.function.is_some() {
-        // For the legacy linker or `--function` support, we still use a glob to link everything.
+    if session.args.function.is_some() {
+        // For the `--function` support, we still use a glob to link everything.
```
tedinski, line 153: Hmm, do we we need to? We can choose semantics for this, since it's not customer-facing. The only problem is trying to find the right target that has that symbol, right? Maybe we can just look it up! (Doesn't need to be this Pr, but open issue?)

scripts/codegen-firecracker.sh
```diff
@@ -23,12 +23,11 @@ echo
...
 # At the moment, we only test codegen for the virtio module
 cd $KANI_DIR/firecracker/src/devices/src/virtio/
-# Disable warnings until https://github.com/model-checking/kani/issues/573 is fixed
+export KANI_LOG=error
 export RUSTC_LOG=error
 export RUST_BACKTRACE=1
-# Use the legacy linker for now since we want to maximize the code that we are compiling from firecracker.
-# The MIR Linker will by default only collect code relevant to proof harnesses, however, firecracker has none.
-cargo kani --only-codegen --legacy-linker
+# Use cargo assess since this is now our default way of assessing Kani suitability to verify a crate.
```
tedinski, line 29: Not for this PR, but think we should do this for `std` too? (Or did we already and I missed it...)

tedinski (approved):

kani-driver/src/args.rs
celinval: Sure. I can do the same for the legacy linker too. I debated whether to do that or not, but it's probably a good habit anyways.

kani-driver/src/project.rs
celinval: Totally. I think it's a separate issue, but I was thinking that we should just record the public functions as part of the metadata, like we do for tests. We could even restrict them to only include functions that don't take any parameter for now to avoid UB while using --function.

celinval: https://github.com/model-checking/kani/issues/2129

scripts/codegen-firecracker.sh
celinval: We haven't. I added a note to the description of the PR.

I don't know how easy it would be to compile the tests of the `std` library without the bootstrap script. We currently rely on the special cargo support to build the standard library. So even just building the standard library as part of `cargo kani` flow would require a way to extend our cargo call to include the `-Z build-std`.

kani-driver/src/args.rs
```diff
@@ -535,6 +539,14 @@ impl KaniArgs {
 
         Ok(())
     }
+
+    fn print_deprecated(&self, option: &str) {
+        if !self.quiet {
+            warning(&format!(
+                "The `{option}` flag is deprecated and it will be removed in future versions"
```
tedinski, line 546: I would try to state outright that it does nothing anymore and should be removed.

tedinski (approved):

PR #2128, opened 2023-01-17
Fix nightly CBMC workflow

zhassan-aws:
### Description of changes: 

The CBMC repo recently introduced some `Cargo.toml` files (for the Rust API), and if CBMC is built after checking out Kani, the build fails because it thinks CBMC should be added as a workspace member to Kani:

https://github.com/model-checking/kani/actions/runs/3937611951/jobs/6735339605

This PR updates the nightly CBMC workflow to install CBMC _before_ checking out Kani to avoid this issue.

It also introduces an input parameter to the setup action to control whether to install CBMC as part of the setup. This is to avoid installing two CBMC versions (nightly and integrated version).

### Resolved issues:

Resolves #ISSUE-NUMBER

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/.github/workflows/cbmc-latest.yml
+++ b/.github/workflows/cbmc-latest.yml
@@ -21,6 +21,22 @@ jobs:
       matrix:
         os: [macos-11, ubuntu-18.04, ubuntu-20.04, ubuntu-22.04]
     steps:
+      - name: Checkout CBMC under "cbmc"
+        uses: actions/checkout@v3
+        with:
+          repository: diffblue/cbmc
+          path: cbmc
+
+      - name: Build CBMC
+        run: |
+          cd cbmc
+          cmake -S . -Bbuild -DWITH_JBMC=OFF
+          cmake --build build -- -j 4
+          sudo cmake --build build --target install
+          # Cleanup cbmc directory
+          cd ..
+          rm -rf cbmc
+
       - name: Checkout Kani
         uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


.github/workflows/cbmc-latest.yml
```diff
@@ -44,34 +67,21 @@ jobs:
...
-    runs-on: ubuntu-20.04
-    steps:
       - name: Checkout Kani
         uses: actions/checkout@v3
 
       - name: Setup Kani Dependencies
         uses: ./.github/actions/setup
         with:
           os: ubuntu-20.04
+          install_cbmc: 'false'
```
karkhaz, line 81: Are you sure this works? I've seen cases where inputs and outputs are not converted to booleans correctly when used in expressions like `if: ${{ inputs.install_cbmc }}`. Since `'false'` is a non-empty string, it evaluates to `true`.

The [ubuntu-20.04 run](https://github.com/model-checking/kani/actions/runs/3942005640/jobs/6745140138#step:3:5) for this PR does indeed seem to set `install_cbmc` to true... 

Suggestion [here](https://github.com/actions/runner/issues/1483#issuecomment-996183227) is to do:
```
if: ${{format('{0}', inputs.install_cbmc) != 'false'}}
```

karkhaz (changes requested):

.github/workflows/cbmc-latest.yml
zhassan-aws: Indeed, it failed! Thanks for pointing out the fix. I'll apply your suggestion.

karkhaz (approved):

.github/workflows/cbmc-latest.yml
zhassan-aws: I ran the workflow in my fork and confirmed that the CBMC installation was skipped:

https://github.com/zhassan-aws/kani/actions/runs/3942199886/jobs/6745566157

PR #2132, opened 2023-01-17
Reject proof harnesses with arguments

### LINKED ISSUE #661: RMC should use fully qualified name for harness selection and it should not rely on `no_mangle` attribute

### LINKED ISSUE #689: Introduce rmctool attribute validation

### LINKED ISSUE #1919: Inconsistent behavior when functions with parameters are used as harnesses
Consider the following program:
```rust
#[kani::proof]
fn foo(c: char) {
    assert!(char::from_u32(c as u32).is_some());
}

#[kani::proof]
fn bar() {
    let c = kani::any();
    foo(c);
}
```
When run with:
```
kani entry.rs --harness bar
```
verification succeeds as expected:
```
SUMMARY:
 ** 0 of 67 failed

VERIFICATION:- SUCCESSFUL
```
However, when run with
```
kani entry.rs --harness foo
```
verification fails:
```
SUMMARY:
 ** 1 of 11 failed
Failed Checks: assertion failed: char::from_u32(c as u32).is_some()
 File: "/home/ubuntu/examples/entry.rs", line 3, in foo

VERIFICATION:- FAILED
```
Kani version: 4db1b0964a3

This is caused by not applying the validity constraints on `char` in the `foo` version.

celinval:
### Description of changes: 

- Change `#[kani::proof]` expansion so it doesn't include `#[no_mangle]` but includes `[allow(dead_code)]`. (#661 and #689).

- Add a check for harnesses with arguments and merge the checks into one function (#1919). 

### Resolved issues:

Fixes #661
Fixes #689
Fixes #1919

### Testing:

* Is this a refactor change?

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
@@ -238,7 +238,7 @@ impl<'tcx> GotocCtx<'tcx> {
         if !proof_attributes.is_empty() {
             let span = proof_attributes.first().unwrap().span;
             if proof_attributes.len() > 1 {
-                self.tcx.sess.span_warn(proof_attributes[0].span, "Duplicated attribute");
+                self.tcx.sess.span_warn(proof_attributes[0].span, "Duplicate attribute");
             }
 
             if self.tcx.def_kind(def_id) != DefKind::Fn {
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
@@ -13,6 +13,7 @@ use rustc_ast::Attribute;
 use rustc_hir::def::DefKind;
 use rustc_hir::def_id::DefId;
 use rustc_middle::mir::{HasLocalDecls, Local};
+use rustc_middle::ty::layout::FnAbiOf;
 use rustc_middle::ty::{self, Instance};
 use std::collections::BTreeMap;
 use std::convert::TryInto;
@@ -232,11 +233,14 @@ impl<'tcx> GotocCtx<'tcx> {
         self.reset_current_fn();
     }
 
-    pub fn is_proof_harness(&self, def_id: DefId) -> bool {
-        let all_attributes = self.tcx.get_attrs_unchecked(def_id);
-        let (proof_attributes, _) = partition_kanitool_attributes(all_attributes);
+    /// Check that if an item is tagged with a proof_attribute, it is a valid harness.
+    fn check_proof_attribute(&self, def_id: DefId, proof_attributes: Vec<&Attribute>) {
         if !proof_attributes.is_empty() {
             let span = proof_attributes.first().unwrap().span;
+            if proof_attributes.len() > 1 {
+                self.tcx.sess.span_warn(proof_attributes[0].span, "Duplicated attribute");
+            }
+
             if self.tcx.def_kind(def_id) != DefKind::Fn {
                 self.tcx
                     .sess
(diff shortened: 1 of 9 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -245,33 +249,37 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                if !self.fn_abi_of_instance(instance, ty::List::empty()).args.is_empty() {
+                    self.tcx
+                        .sess
+                        .span_err(span, "Functions used as harnesses can not have any arguments.");
+                }
             }
+        }
+    }
+
+    pub fn is_proof_harness(&self, def_id: DefId) -> bool {
```
tedinski, line 263: Since this function is called from e.g. reachability, it might not be appropriate to the put the check here. It may get called multiple times?

tedinski (approved):

tedinski:
One other thing: Can we change the PR title? Want to ensure that whoever write the release notes will notice that this is a possible "breaking" change.

Maybe:

> Properly reject proof harnesses with arguments, and don't change harness symbol names

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -232,11 +233,14 @@ impl<'tcx> GotocCtx<'tcx> {
         self.reset_current_fn();
     }
 
-    pub fn is_proof_harness(&self, def_id: DefId) -> bool {
-        let all_attributes = self.tcx.get_attrs_unchecked(def_id);
-        let (proof_attributes, _) = partition_kanitool_attributes(all_attributes);
+    /// Check that if an item is tagged with a proof_attribute, it is a valid harness.
+    fn check_proof_attribute(&self, def_id: DefId, proof_attributes: Vec<&Attribute>) {
         if !proof_attributes.is_empty() {
```
zhassan-aws, line 238: This condition is also checked by the callers. Should we remove it and turn it into an assert?

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -232,11 +233,14 @@ impl<'tcx> GotocCtx<'tcx> {
...
             let span = proof_attributes.first().unwrap().span;
+            if proof_attributes.len() > 1 {
+                self.tcx.sess.span_warn(proof_attributes[0].span, "Duplicated attribute");
```
zhassan-aws, line 241:
```suggestion
                self.tcx.sess.span_warn(proof_attributes[0].span, "Duplicate attribute");
```

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
celinval: Sure

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
celinval: It already was being checked twice. But I'm ok removing from here.

zhassan-aws (approved):

PR #2134, opened 2023-01-18
Enable concrete playback for satisfiable cover properties

### LINKED ISSUE #2133: Support concrete playback for `kani::cover`
Requested feature: Allow `--concrete-playback` to generate harnesses for satisfied cover properties specified using `kani::cover`.
Use case: The `kani::cover` macro can be used to check if certain scenario are possible. When they are, it would be useful to determine the input that can lead to such scenarios.
Link to relevant documentation (Rust reference, Nomicon, RFC): https://model-checking.github.io/kani/crates/doc/kani/macro.cover.html

Test case:

```rust
#[kani::proof]
fn main() {
    let x: i32 = kani::any();
    // What is a possible value of `x` that satisfies the following condition?
    kani::cover!(x != 0 && x / 2 == 0);
}
```

zhassan-aws:
### Description of changes: 

Include satisfiable cover properties as part of what concrete playback can generate tests for.

### Resolved issues:

Resolves #2133 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

### Checklist

### DIFF
--- a/kani-driver/src/harness_runner.rs
+++ b/kani-driver/src/harness_runner.rs
@@ -116,7 +116,10 @@ impl KaniSession {
         let failing = failures.len();
         let total = succeeding + failing;
 
-        if self.args.concrete_playback.is_some() && !self.args.quiet && failures.is_empty() {
+        if self.args.concrete_playback.is_some()
+            && !self.args.quiet
+            && results.iter().all(|r| !r.result.generated_concrete_test)
+        {
             println!(
                 "INFO: The concrete playback feature never generated unit tests because there were no failing harnesses."
             )
--- a/tests/ui/concrete-playback/cover/main.rs
+++ b/tests/ui/concrete-playback/cover/main.rs
@@ -5,6 +5,6 @@
 
 #[kani::proof]
 pub fn harness() {
-    let x: i32 = kani::any();
+    let x: u32 = kani::any();
     kani::cover!(x != 0 && x / 2 == 0);
 }
--- a/tests/ui/concrete-playback/cover/main.rs
+++ b/tests/ui/concrete-playback/cover/main.rs
@@ -1,0 +1,10 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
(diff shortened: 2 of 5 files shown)


kani-driver/src/harness_runner.rs
```diff
@@ -116,7 +116,10 @@ impl KaniSession {
         let failing = failures.len();
         let total = succeeding + failing;
 
-        if self.args.concrete_playback.is_some() && !self.args.quiet && failures.is_empty() {
+        if self.args.concrete_playback.is_some()
+            && !self.args.quiet
+            && results.iter().all(|r| !r.result.generated_concrete_test)
+        {
             println!(
                 "INFO: The concrete playback feature never generated unit tests because there were no failing harnesses."
```
celinval, line 124: Do you think we even need this? I was also wondering if we print that when a unit tests wasn't generated because it already existed.

tests/ui/concrete-playback/cover/main.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// kani-flags: --enable-unstable --concrete-playback=print
+
+#[kani::proof]
+pub fn harness() {
+    let x: i32 = kani::any();
+    kani::cover!(x != 0 && x / 2 == 0);
```
celinval, line 9: Can you make this check to have only one satisfying value, so we avoid having to change the test whenever CBMC decides to produce a different value.

zhassan-aws: Good point. Done.

kani-driver/src/harness_runner.rs
zhassan-aws: > I was also wondering if we print that when a unit tests wasn't generated because it already existed.

We do that [here](https://github.com/model-checking/kani/blob/f635f7ba833d532ca31d4eec95ff2339e2289869/kani-driver/src/concrete_playback.rs#L98).

celinval (approved):

PR #2141, opened 2023-01-23
Add documentation for performance comparison tool

karkhaz:
(not to be merged).

The only really interesting file to comment on is 0100-content.md, which has bee hidden from the diff because it's too long.

### DIFF
--- a/tools/performance-benchmarking/src/md/0010-content.md
+++ b/tools/performance-benchmarking/src/md/0010-content.md
@@ -1,0 +1,916 @@
+`benchcomp` allows you to:
+
+* Run two or more sets of benchmark suites under different configurations;
+* declaratively define visualizations to be generated from the resulting data;
+* run the tool with sensible defaults, or hook into every stage of the pipeline when you need more flexibility.
+
+This documentation contains three sections.
+The [user walkthrough](#user-walkthrough) takes you through the process of setting up an entirely new benchmark run and dashboard.
+The [developer reference](#developer-reference) describes `benchcomp`'s architecture and the different data formats that it uses, enabling you to author a benchmark run and dashboard of your own.
+The [visualization viewer guide](#benchcomp-visualization-viewer-guide) describes how to use dashboards that `benchcomp` generates.
+Although dashboards are highly customizable, the guide describes the common elements.
+
+
+# User Walkthrough
+
+This section presents a single user story, showing how the user can quickly get started with a simple benchmark comparison run, through to a more complex setup with extensive customization.
+This section explains the syntax in prose without attempting to exhaustively document it; the [Development Reference](#developer-reference) below is a full reference for the syntax.
+
+
+## Comparing two variants
+
+Alice wants to compare the performance of two versions of CBMC: the latest release, with and without a new optimization that she's implemented.
+Initially, Alice tries to do this entirely using `benchcomp`'s built-in parsers and filters.
+She performs the following steps:
+
+* Create a directory containing all AWS CBMC proofs, with a top-level script that runs all of them called `run-all-cbmc-proofs.py`
+* Create a configuration file (shown below)
+* Run `benchcomp -o result.json`
+
+The initial configuration file is:
+
+<div class="twocolumn">
+<div class="col col-66">
+
+```yaml
+# benchcomp.yaml
+
+run:
+  suites:
+    all_cbmc_proofs:
+      parser:
+        type: built_in
+        name: litani_to_benchcomp
+        directory: all_cbmc_proofs
+      variants:
+        optimized:
+          provenance: inline
+          config:
+            command_line: ./run-cbmc-proofs.py
+            env:
+              CBMC: ~/src/cbmc/build/bin/cbmc
+        release:
+          provenance: inline
+          config:
+            command_line: ./run-cbmc-proofs.py
+            env:
+              CBMC: /usr/local/bin/cbmc
+```
+
+</div>
+<div class="col col-33">
+
+This says: run two *variants* (`optimized` and `release`) of a single benchmark *suite* (`all_cbmc_proofs`).
+The variants are distinguished in this case by the `CBMC` environment variable, which the AWS [proof build system](https://github.com/model-checking/cbmc-starter-kit) uses when invoking CBMC.
+
+</div>
+</div>
+
+`benchcomp` copies the `all_cbmc_proofs` directory to two temporary directories, one for each variant, and runs the command line. It uses the built-in `litani_to_benchcomp` parser to assemble the results. `benchcomp` then writes this data to the output file in JSON format (here in YAML for readability):
+
+<div class="twocolumn">
+<div class="col col-75">
+
+```yaml
+# result.json
+
+metrics:
+  runtime:
+    lower_is_better: true
+    unit: s
+  passed: {}
+  coverage:
+    unit: %
+benchmarks:
+  s2n_init:
+    tags: [s2n, crypto, cbmc]
+    variants:
+        optimized:
+          metrics:
+            runtime: 3
+            passed: true
+            coverage: 100
+          path: ./tmp/benchcomp/runs/abcd1234/optimized/all_cbmc_proofs/s2n/tests/cbmc/proofs/s2n_init
+        release:
+          metrics:
+            runtime: 4
+            passed: true
+            coverage: 100
+          path: ./tmp/benchcomp/runs/abcd1234/release/all_cbmc_proofs/s2n/tests/cbmc/proofs/s2n_init
+  freertos_list_replace:
+    tags: [cbmc, freertos, freertos_kernel]
+    variants:
+      optimized:
+        metrics:
+          # ...
+```
+</div>
+<div class="col col-25">
+
+The output file first declares the various *metrics* that the union of all the parsers emitted.
+In later sections, we will see how Alice writes her own parser or extends an existing one with new metrics.
+The file then maps benchmark names to values for each of the metrics for each of the variants that the benchmark ran under.
+
+</div>
+</div>
+
+## Visualizing the results
+
+Alice now wants to visualize the difference in runtime between the two versions of CBMC, ignoring other metrics for now.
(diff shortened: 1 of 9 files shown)


celinval (commented):
Can you please move everything to be inside either `tools/` or `scripts/`?

celinval:
It might be helpful to add a link to the rendered file: https://github.com/karkhaz/kani/blob/kk-performance-benchmarking/performance-benchmarking/src/md/0010-content.md

karkhaz:
Thanks @celinval, I've moved the files to `tools` so the rendered page is now [here](https://github.com/karkhaz/kani/blob/kk-performance-benchmarking/tools/performance-benchmarking/src/md/0010-content.md).

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
...
+</div>
+</div>
+</div>
+<div class="central-container">
+<div class="central-view">
+
+
+~include tmp/box_whisker/cbmc.html
+
+</div>
```
tautschnig, line 186: Is this just not rendered properly on GitHub, but otherwise works fine (including the `~include` line)?

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
...
+
+```yaml
+filter:
+  - type: executable
+    path: ./scripts/drop_fast_benchmarks.py
+```
+</div>
+</div>
+
+With either of these examples, `benchcomp` will automatically invoke the filter whenever Alice runs `benchcomp` or `benchcomp` visualize.
```
tautschnig, line 249: Should it be "`benchcomp visualize`"?

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
...
+This development reference describes `benchcomp`'s architecture and gives complete schemas for each of its data formats.
+
+
+## Architecture and overview
+
+This section gives a high-level sequence diagram of the entire tool pipeline; describes the components of a single run; and then describes how benchmarks are run in more detail.
+
+At the highest level, users invoke `benchcomp`, a unified front-end that runs several other sub-tools in the background.
+`benchcomp` first executes `benchcomp run`, which runs one or more _benchmark suites_ several times (each time using a different _variant_).
+`bc run` eventually returns a JSON document in the [`result.json`](#result.json-schema) format, containing the union of results from all benchmark runs under every variant.
```
tautschnig, line 517: How do `bc run` and `benchcomp run` relate to each other?

tautschnig (approved):
This is a beautiful piece of work! Just one high-level question that I couldn't immediately spot the answer to: how is concurrency/isolation handled here? Does `benchcomp` just not introduce any concurrency itself and it's all left to the underlying tool (which might be `litani`)?

karkhaz:
> This is a beautiful piece of work!

Thank you @tautschnig ! :)

> Just one high-level question that I couldn't immediately spot the answer to: how is concurrency/isolation handled here? Does `benchcomp` just not introduce any concurrency itself and it's all left to the underlying tool (which might be `litani`)?

My intention was that benchcomp would run each benchmark serially by default, but users could choose to run them in parallel with a `-j` flag.

Serially by default seems sensible because I assume that a "suite" contains many benchmarks, usually more than the number of cores, and so each benchmark suite would contain its own mechanism for parallelizing its own benchmarks. This is true for Litani and also for `cargo kani`, and most other testing frameworks and build systems I guess.

I can't see how benchcomp would be able to run multiple suites in parallel without risking either over- or under-utilizing CPU cores, because each suite has its own parallelism scheduler with no central control. I haven't thought about this deeply though, it would be good if this were possible. Nevertheless, I think giving users a `-j` switch if they want to use it is probably okay.

(Roughly speaking, I intend that benchcomp will emit a Ninja file that will take care of: running each suite; parsing the result; combining the results; filtering the combined result; and generating visualizations. All in dependency order of course. Now, the benchmark suite runs don't depend on each other so could run in parallel. I was intending to put the ninja jobs for the suite runs in a pool called `suite_runs`, which would have a depth of 1 by default.)

zhassan-aws (commented):
Very good overall! I added a few comments. I also suggest adding a section that lists the inputs the user needs to provide to `benchcomp`, e.g. the yaml configuration file, a script to run a benchmark suite, a script to parse output results into json, etc.

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
...
+This section explains the syntax in prose without attempting to exhaustively document it; the [Development Reference](#developer-reference) below is a full reference for the syntax.
+
+
+## Comparing two variants
+
+Alice wants to compare the performance of two versions of CBMC: the latest release, with and without a new optimization that she's implemented.
+Initially, Alice tries to do this entirely using `benchcomp`'s built-in parsers and filters.
+She performs the following steps:
+
+* Create a directory containing all AWS CBMC proofs, with a top-level script that runs all of them called `run-all-cbmc-proofs.py`
```
celinval, line 26: Should we expand a bit on what this script is? Doesn't benchcomp rely on the output of this script?

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
+`benchcomp` allows you to:
+
+* Run two or more sets of benchmark suites under different configurations;
+* declaratively define visualizations to be generated from the resulting data;
+* run the tool with sensible defaults, or hook into every stage of the pipeline when you need more flexibility.
+
+This documentation contains three sections.
+The [user walkthrough](#user-walkthrough) takes you through the process of setting up an entirely new benchmark run and dashboard.
+The [developer reference](#developer-reference) describes `benchcomp`'s architecture and the different data formats that it uses, enabling you to author a benchmark run and dashboard of your own.
```
zhassan-aws, line 9: We've primarily been using the "dashboard" term to mean a page that tracks performance over time. It's used here to refer to the results of a single benchcomp run, which confused me a bit. Perhaps replace it with "Visual results page" or something similar?

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
...
+</div>
+<div class="col col-33">
+
+This says: run two *variants* (`optimized` and `release`) of a single benchmark *suite* (`all_cbmc_proofs`).
+The variants are distinguished in this case by the `CBMC` environment variable, which the AWS [proof build system](https://github.com/model-checking/cbmc-starter-kit) uses when invoking CBMC.
+
+</div>
+</div>
+
+`benchcomp` copies the `all_cbmc_proofs` directory to two temporary directories, one for each variant, and runs the command line. It uses the built-in `litani_to_benchcomp` parser to assemble the results. `benchcomp` then writes this data to the output file in JSON format (here in YAML for readability):
```
celinval, line 69: Also, isn't the `JSON` an intermediate step for benchcomp at this point? Should we even talk about it in this use case?

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
...
+        metric: vccs
+        pairs:
+          - [release, optimized]
+```
+
+</div>
+</div>
+
+
+## Adding another benchmark suite
```
celinval, line 321: Shouldn't you talk about adding a benchmark suite first?

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
...
+* run the tool with sensible defaults, or hook into every stage of the pipeline when you need more flexibility.
+
+This documentation contains three sections.
+The [user walkthrough](#user-walkthrough) takes you through the process of setting up an entirely new benchmark run and dashboard.
+The [developer reference](#developer-reference) describes `benchcomp`'s architecture and the different data formats that it uses, enabling you to author a benchmark run and dashboard of your own.
+The [visualization viewer guide](#benchcomp-visualization-viewer-guide) describes how to use dashboards that `benchcomp` generates.
+Although dashboards are highly customizable, the guide describes the common elements.
+
+
+# User Walkthrough
```
feliperodri, line 14: It'd be nice to have in the User Walk-though a small example that I could follow and try the tool on my on.

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
...
+
+<div class="subpage">
+<div class="sidebar">
+<div class="side-header">
+
+Run ID: `abc123`
+
+`2023-01-01T18:42:54`
+
+[JSON version](/) of this dashboard
```
celinval, line 160: ?

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
...
+This says to generate a single graph in a single HTML dashboard visualization.
+The graph compares the runtimes of all benchmarks run under the `release` variant pairwise with the same benchmarks run under the `optimized` variant.
+
+</div>
+</div>
+
+Alice runs the entire suite, together with generating and writing out the visualization, by running `benchcomp` again.
+Alternatively, she can run `benchcomp visualize < result.json`, which loads the result of the run she did in the previous section.
+
+The resulting dashboard looks like this:
```
celinval, line 150: I have a few questions about this graph, not sure it's out of the scope of this PR.

tools/performance-benchmarking/src/md/0010-content.md
```diff
@@ -0,0 +1,916 @@
...
+<div class="central-view">
+
+
+~include tmp/box_whisker/cbmc.html
+
+</div>
+</div>
+</div>
+
+Benchmarks whose runtime did not change between the two versions have a 'runtime increase' of 1.0.
```
feliperodri, line 190: Does that mean that if I see s2n(1.0) I get no change in runtime? How precise is that?

celinval (commented):
I haven't finished yet. So far this looks promising. Thanks!

PR #2045, opened 2022-12-28
Update rust toolchain to 2022-12-11

zhassan-aws:
- Update Rust toolchain to version `2022-12-04`
- Conform with changes to `GenericArg`
- Conform with changes to `NestedMetaItem`
- Make new `ConstKind::Expr` return on reach visitor
- Conform with changes to `Layout`
- Update rust toolchain version to 2022-12-11
- Clippy fixes
- More clippy/format fixes

### Description of changes: 

This PR builds on @adpaco-aws's commits in #1983, but bumps the version further to 2022-12-11 (instead of 2022-12-04), to include some of the later commits related to generators, which are necessary to fix how we codegen them. 

The relevant PR is https://github.com/rust-lang/rust/pull/105082 which was merged on 2022-12-10.

Another PR that required changes is https://github.com/rust-lang/rust/pull/105234.

### Resolved issues:

Towards #2044 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

I suggest reviewing by commit because the clippy fix commits touch many files.

### Testing:

* How is this change tested? Current regressions

* Is this a refactor change? No

### Checklist
- [X] Each commit message has a non-empty body, explaining why the change was made
- [X] Methods or procedures are documented
- [X] Regression or unit tests are included, or existing tests cover the modified code
- [X] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #2145, opened 2023-01-23
Add `--no-fail-fast` to compiletest

### LINKED ISSUE #2144: Add a flag to `compiletest` to control whether it should fail fast or not
**Proposed change:**  Add a `--no-fail-fast` to compiletest which will execute the entire suite regardless of failure.

**Motivation:** This used to be the default behavior but it has changed to always failing fast after #2045. I think both modes are useful, but for CI I would prefer printing all failures in the suite before aborting. The more information the better.

celinval:
### Description of changes: 

Compiletest behavior has changed to always failing fast after #2045. This PR introduces a `--no-fail-fast` flag to compiletest which will execute the entire suite regardless of failure.

The regression script uses `--no-fail-fast` so all failures in a suite are printed as part of the CI.

### Resolved issues:

Resolves #2144 

### Testing:

* How is this change tested? Manually tested the default behavior. The regression uses the new flag.

### Checklist

### DIFF
--- a/scripts/kani-regression.sh
+++ b/scripts/kani-regression.sh
@@ -74,7 +74,8 @@ for testp in "${TESTS[@]}"; do
   suite=${testl[0]}
   mode=${testl[1]}
   echo "Check compiletest suite=$suite mode=$mode"
-  cargo run -p compiletest --quiet -- --suite $suite --mode $mode --quiet
+  cargo run -p compiletest --quiet -- --suite $suite --mode $mode \
+      --quiet --no-fail-fast
 done
 
 # Check codegen for the standard library
--- a/tools/compiletest/src/common.rs
+++ b/tools/compiletest/src/common.rs
@@ -134,6 +134,10 @@ pub struct Config {
     /// Timeout duration for each test.
     pub timeout: Option<Duration>,
 
+    /// Whether we will abort execution when a failure occurs.
+    /// When set to false, this will execute the entire test suite regardless of any failure.
+    pub fail_fast: bool,
+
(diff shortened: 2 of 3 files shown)


scripts/kani-regression.sh
```diff
@@ -74,7 +74,8 @@ for testp in "${TESTS[@]}"; do
   suite=${testl[0]}
   mode=${testl[1]}
   echo "Check compiletest suite=$suite mode=$mode"
-  cargo run -p compiletest --quiet -- --suite $suite --mode $mode --quiet
+  cargo run -p compiletest --quiet -- --suite $suite --mode $mode \
+      --quiet --no-fail-fast
```
adpaco-aws, line 78: Sorry if I didn't make myself clear earlier, but some people are actually annoyed if they have to wait for all tests in a suite to complete before they're able to see what went wrong.

adpaco-aws: I guess there are two scenarios here:
 * Developers running the regression: They want it to fail fast so they don't have to wait for the suite to complete before seeing the errors they got.
 * CI running the regression: Developers want it to NOT fail fast so they get more comprehensive results.

adpaco-aws (approved):

scripts/kani-regression.sh
celinval: Discussed offline. We'll keep the script simple for now and keep the default as it is meant to be run in CI (with `--no-fail-fast`). We can customize the script later or developers can individually run `compiletest` separately.

Issue #1365, opened 2022-07-12
Audit function signature codegen


fzaiser:
The code of `closure_sig` is copy-pasted from `fn_sig_for_fn_abi` in `compiler/rustc_middle/src/ty/layout.rs`. We may have to do something similar for generators (#416), so it would be good if we could use the rustc function `fn_abi_of_instance` directly (as the comment on `fn_sig_for_fn_abi` suggests. This would prevent problems with keeping in sync with rustc.

fzaiser:
The relevant function `fn_sig_for_fn_abi` is [here](https://github.com/rust-lang/rust/blob/848090dcd18553b790461132ca9d2a020aeea9a2/compiler/rustc_middle/src/ty/layout.rs#L2754). It has a comment saying:
```rust
    // NOTE(eddyb) this is private to avoid using it from outside of
    // `fn_abi_of_instance` - any other uses are either too high-level
    // for `Instance` (e.g. typeck would use `Ty::fn_sig` instead),
    // or should go through `FnAbi` instead, to avoid losing any
    // adjustments `fn_abi_of_instance` might be performing.
```

So the question is whether we should use `Ty::fn_sig` or `fn_abi_of_instance`. To find out, I looked at the cranelift backend. They use [this function](https://github.com/rust-lang/rust/blob/b629c85bd74dfb730a3e9308312b007c0bf027cb/compiler/rustc_codegen_cranelift/src/abi/mod.rs#L52) to obtain the signature of a function/closure/generator. I presume we should do something similar.

Note that the `fn_abi_of_instance` function makes some adjustments to the signature returned by the private function `fn_sig_for_fn_abi`: they happen in the call to `fn_abi_new_uncached` [here](https://github.com/rust-lang/rust/blob/848090dcd18553b790461132ca9d2a020aeea9a2/compiler/rustc_middle/src/ty/layout.rs#L3163), in that function they call `fn_abi_adjust_for_abi` [here](https://github.com/rust-lang/rust/blob/848090dcd18553b790461132ca9d2a020aeea9a2/compiler/rustc_middle/src/ty/layout.rs#L3364), which is defined just below.

fzaiser:
@celinval The adjustments to the signature could be related to #1350. These adjustments don't currently happen in the Kani code, but they probably should.

celinval:
Totally! We should investigate if using `fn_abi_of_instance` would simplify how we handle functions today. I believe this is also related to the issue we were having with closure to function pointer casting: #274

[...]

PR #2149, opened 2023-01-24
Upgrade toolchain to `nightly-2023-01-23`

### LINKED ISSUE #2113: Update rust toolchain version for Kani 0.20
Biweekly update for the rust toolchain version. For Kani 0.20, we will target the new year version: nightly-2023-01-16.

celinval:
### Description of changes: 
Upgrade our toolchain to `nightly-2023-01-23`. The changes here are related to the following changes:
- https://github.com/rust-lang/rust/pull/104986
- https://github.com/rust-lang/rust/pull/105657
- https://github.com/rust-lang/rust/pull/105603
- https://github.com/rust-lang/rust/pull/105613
- https://github.com/rust-lang/rust/pull/105977.

### Resolved issues:

Resolves #2113 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

I initially upgraded the toolchain to `nightly-2023-01-16` as planned in #2113. However, the AsyncAwait tests were failing due to wrong monomorphization. To fix that, I decided to upgrade to `nightly-2023-01-23` instead to include the changes related to https://github.com/rust-lang/rust/pull/105977. That fixed the regression.

Note: We should consider prioritizing #1365 to stop having to amend the function abi code.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/kani-compiler/kani_queries/src/lib.rs
+++ b/kani-compiler/kani_queries/src/lib.rs
@@ -15,7 +15,7 @@ use {
     std::sync::{Arc, Mutex},
 };
 
-#[derive(Debug, Clone, Copy, AsRefStr, EnumString, EnumVariantNames, PartialEq, Eq)]
+#[derive(Debug, Default, Clone, Copy, AsRefStr, EnumString, EnumVariantNames, PartialEq, Eq)]
 #[strum(serialize_all = "snake_case")]
 pub enum ReachabilityType {
     /// Start the cross-crate reachability analysis from all harnesses in the local crate.
@@ -23,17 +23,12 @@ pub enum ReachabilityType {
     /// Use standard rustc monomorphizer algorithm.
     Legacy,
     /// Don't perform any reachability analysis. This will skip codegen for this crate.
+    #[default]
     None,
     /// Start the cross-crate reachability analysis from all public functions in the local crate.
     PubFns,
     /// Start the cross-crate reachability analysis from all *test* (i.e. `#[test]`) harnesses in the local crate.
     Tests,
-}
-
-impl Default for ReachabilityType {
-    fn default() -> Self {
-        ReachabilityType::None
-    }
 }
 
 pub trait UserInput {
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
@@ -358,20 +358,41 @@ impl<'tcx> GotocCtx<'tcx> {
         // `Generator::resume(...) -> GeneratorState` function in case we
         // have an ordinary generator, or the `Future::poll(...) -> Poll`
         // function in case this is a special generator backing an async construct.
-        let ret_ty = if self.tcx.generator_is_async(*did) {
-            let state_did = self.tcx.require_lang_item(LangItem::Poll, None);
-            let state_adt_ref = self.tcx.adt_def(state_did);
-            let state_substs = self.tcx.intern_substs(&[sig.return_ty.into()]);
-            self.tcx.mk_adt(state_adt_ref, state_substs)
+        let tcx = self.tcx;
+        let (resume_ty, ret_ty) = if tcx.generator_is_async(*did) {
+            // The signature should be `Future::poll(_, &mut Context<'_>) -> Poll<Output>`
+            let poll_did = tcx.require_lang_item(LangItem::Poll, None);
+            let poll_adt_ref = tcx.adt_def(poll_did);
+            let poll_substs = tcx.intern_substs(&[sig.return_ty.into()]);
+            let ret_ty = tcx.mk_adt(poll_adt_ref, poll_substs);
+
+            // We have to replace the `ResumeTy` that is used for type and borrow checking
+            // with `&mut Context<'_>` which is used in codegen.
+            #[cfg(debug_assertions)]
+            {
+                if let ty::Adt(resume_ty_adt, _) = sig.resume_ty.kind() {
+                    let expected_adt = tcx.adt_def(tcx.require_lang_item(LangItem::ResumeTy, None));
+                    assert_eq!(*resume_ty_adt, expected_adt);
+                } else {
+                    panic!("expected `ResumeTy`, found `{:?}`", sig.resume_ty);
+                };
+            }
+            let context_mut_ref = tcx.mk_task_context();
+
+            (context_mut_ref, ret_ty)
         } else {
-            let state_did = self.tcx.require_lang_item(LangItem::GeneratorState, None);
-            let state_adt_ref = self.tcx.adt_def(state_did);
-            let state_substs = self.tcx.intern_substs(&[sig.yield_ty.into(), sig.return_ty.into()]);
-            self.tcx.mk_adt(state_adt_ref, state_substs)
+            // The signature should be `Generator::resume(_, Resume) -> GeneratorState<Yield, Return>`
+            let state_did = tcx.require_lang_item(LangItem::GeneratorState, None);
+            let state_adt_ref = tcx.adt_def(state_did);
+            let state_substs = tcx.intern_substs(&[sig.yield_ty.into(), sig.return_ty.into()]);
+            let ret_ty = tcx.mk_adt(state_adt_ref, state_substs);
+
+            (sig.resume_ty, ret_ty)
         };
+
         ty::Binder::bind_with_vars(
-            self.tcx.mk_fn_sig(
-                [env_ty, sig.resume_ty].iter(),
+            tcx.mk_fn_sig(
+                [env_ty, resume_ty].iter(),
                 &ret_ty,
                 false,
                 Unsafety::Normal,
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
@@ -813,7 +813,7 @@ impl<'tcx> GotocCtx<'tcx> {
                     )
                 }
             }
-            ty::Projection(_) | ty::Opaque(_, _) => {
+            ty::Alias(..) => {
                 unreachable!("Type should've been normalized already")
             }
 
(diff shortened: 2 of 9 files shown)


kani-compiler/kani_queries/src/lib.rs
```diff
@@ -15,27 +15,22 @@ use {
...
-#[derive(Debug, Clone, Copy, AsRefStr, EnumString, EnumVariantNames, PartialEq, Eq)]
+#[derive(Debug, Default, Clone, Copy, AsRefStr, EnumString, EnumVariantNames, PartialEq, Eq)]
 #[strum(serialize_all = "snake_case")]
 pub enum ReachabilityType {
     /// Start the cross-crate reachability analysis from all harnesses in the local crate.
     Harnesses,
     /// Use standard rustc monomorphizer algorithm.
     Legacy,
     /// Don't perform any reachability analysis. This will skip codegen for this crate.
+    #[default]
```
zhassan-aws, line 26: Nice!

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -358,20 +358,41 @@ impl<'tcx> GotocCtx<'tcx> {
         // `Generator::resume(...) -> GeneratorState` function in case we
         // have an ordinary generator, or the `Future::poll(...) -> Poll`
         // function in case this is a special generator backing an async construct.
-        let ret_ty = if self.tcx.generator_is_async(*did) {
-            let state_did = self.tcx.require_lang_item(LangItem::Poll, None);
-            let state_adt_ref = self.tcx.adt_def(state_did);
-            let state_substs = self.tcx.intern_substs(&[sig.return_ty.into()]);
-            self.tcx.mk_adt(state_adt_ref, state_substs)
+        let tcx = self.tcx;
```
zhassan-aws, line 361: Oh no, did the upstream code change again?

zhassan-aws (approved):

zhassan-aws:
Weird: one s2n-quic perf harness is failing.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
celinval: Yep. We really need to fix that.

celinval:
> Weird: one s2n-quic perf harness is failing.

I'm trying to find out what's going on here. CBMC detects a non-existing loop inside `Option<>::ok_or()`. This is an example of test that is failing: 

```
#[kani::proof]
#[kani::unwind(2)]
fn check_ok_or() {
    let opt : Option<bool> = kani::any();
    let val_or = opt.ok_or(());
    assert!(val_or.is_ok());
}
```

Without the changes from this PR, CBMC still detects a loop, but it only fails for `--unwind 1`.

zhassan-aws:
> This is an example of test that is failing:

Does it pass with a larger unwind value?

celinval:
> Does it pass with a larger unwind value?

Yes. It does. It also passes if I don't specify a bound. But why there is a loop here in the first place is a mystery. Let me see if I can reproduce this in C.

celinval:
I did some digging and created a tiny C example that triggers the same issue: https://github.com/diffblue/cbmc/issues/7506.

tedinski:
Is the decision to block this, waiting on a CBMC fix and release? From that CBMC issue's discussion it wasn't clear this was going to be fixed soon...

celinval:
> Is the decision to block this, waiting on a CBMC fix and release? From that CBMC issue's discussion it wasn't clear this was going to be fixed soon...

That's a great question. I don't think we have decided it yet. I don't think this change is urgent, so I think we can wait a few more days to see if CBMC can scope out the fix. Once we have an idea of when this can be resolved, we can decide if a work around is granted. I rather delay this update for a week or two than push a work around to our users code.

tautschnig:
> > Is the decision to block this, waiting on a CBMC fix and release? From that CBMC issue's discussion it wasn't clear this was going to be fixed soon...
> 
> That's a great question. I don't think we have decided it yet. I don't think this change is urgent, so I think we can wait a few more days to see if CBMC can scope out the fix. Once we have an idea of when this can be resolved, we can decide if a work around is granted. I rather delay this update for a week or two than push a work around to our users code.
> 
> Let me know if that makes sense or if you feel like we should just move ahead.

Given the discussion we had in a meeting yesterday, a fix may be several weeks out. @martin-cs has a much more complete view of the concept of "loops" and their challenges, and tells me there are a lot of possible pitfalls. So I don't question that we need to find a way to address this for our customers, but it may well be worth upgrading with a workaround right now and then keeping open an issue to track reverting this workaround once a fix has been implemented.

martin-cs:
On a previous project I learnt, to my cost, that there are a lot of weird loop configurations that turn up if you analyse a large enough code base.  Either A. you play whack-a-mole and continuously tweak your loop handling or B. you need a dedicated loop normalisation pass.  An A solution to this would be pretty quick but it will not be the last, a B solution to this will take longer but will benefit many other users ( @remi-delmas-3000 for one ).

At the moment I don't have the time allocation to solve the whole problem myself but am happy to meet with / share experience and benchmarks with anyone who does.

celinval:
> On a previous project I learnt, to my cost, that there are a lot of weird loop configurations that turn up if you analyse a large enough code base. Either A. you play whack-a-mole and continuously tweak your loop handling or B. you need a dedicated loop normalisation pass. An A solution to this would be pretty quick but it will not be the last, a B solution to this will take longer but will benefit many other users ( @remi-delmas-3000 for one ).
> 
> At the moment I don't have the time allocation to solve the whole problem myself but am happy to meet with / share experience and benchmarks with anyone who does.

I would love to hear more about it @martin-cs  but in this case there is no loop. A simple DFS in the CFG should be enough to at least rule out false positives in the loop detection.

martin-cs:
@celinval I completely agree with your point that it will solve this specific problem.  My experience with this and other loop-related weirdness is that "We just need a small tweak to deal with this one" is a very long road of spiralling complexity and increasing fragility of the algorithms.  Although it might be a little more work initially, I would strongly recommend a robust loop normalisation pass and robust loop detection algorithms because the list of "just one more tweak" seems to never end.

celinval:
> @celinval I completely agree with your point that it will solve this specific problem. My experience with this and other loop-related weirdness is that "We just need a small tweak to deal with this one" is a very long road of spiralling complexity and increasing fragility of the algorithms. Although it might be a little more work initially, I would strongly recommend a robust loop normalisation pass and robust loop detection algorithms because the list of "just one more tweak" seems to never end.

I get your point. We might try doing a topological sort on our side to see if we overcome the current issues we are facing.

Rust doesn't support `goto` statements, so I am hoping that the weird cases you are talking about won't be a problem here.

martin-cs:
> Rust doesn't support goto statements, so I am hoping that the weird
> cases you are talking about won't be a problem here.

Unfortunately ... you can still cause chaos without `goto`.  Once you
have inlining then returns start to behave a lot like localised gotos.
Also sensible normalisations like "if you have a GOTO instruction who's
target is an unconditional GOTO then you can jump directly to the
second target" can create some surprisingly anomalous situations like
loops with multiple back edges.

celinval:
> Rust doesn't support goto statements, so I am hoping that the weird cases you are talking about won't be a problem here.
> Unfortunately ... you can still cause chaos without `goto`. Once you have inlining then returns start to behave a lot like localised gotos. Also sensible normalisations like "if you have a GOTO instruction who's target is an unconditional GOTO then you can jump directly to the second target" can create some surprisingly anomalous situations like loops with multiple back edges.

Interesting. Do you have any examples for these cases? Shouldn't the return statements behave like break statements at that point?

celinval:
Running the failing test in my local machine with `cargo kani` I got the following:

| Configuration  | Time | Memory |
|--------------------|---------|------------|
| This PR           | 200s | 20GB |
| 55fe61c9fb9*  | 164s | 13GB |
| main               | 198s | 16GB |

\* This is the commit that preceded any of the changes in this PR.

It's not clear to me why it is only failing now since GitHub workers are only supposed to have 7GB of RAM. Let me keep digging.

celinval:
Running the failing test in my local machine with `cargo kani` and CBMC 5.75, I got the following:

| Configuration  | Time | Memory |
|--------------------|---------|------------|
| main  (CBMC 5.75)    | 5.74s | 165MB |
| this PR (CBMC 5.75) |  206s  | 16GB    |
| main  (CBMC 5.70)    | 200.s | 16GB |

Note that the time in this PR is very similar to the one I obtained when running `main` with an older CBMC version (before https://github.com/diffblue/cbmc/pull/7230).

I wonder if whatever has changed neutralized the optimization that @tautschnig introduced. 😕

celinval:
I create https://github.com/model-checking/kani/issues/2191 to track the performance regression. This PR is blocked till we have a better picture of what's going on.

celinval:
Since the fix has already been merged to CBMC and we should update the CBMC version before the release, we decided to mark the performance regression tests as ignored. In the mean time, the CBMC latest workflow will run the ignored tests.

I manually triggered the job in my fork to see if it works: https://github.com/celinval/kani-dev/actions/runs/4368456571

.github/workflows/cbmc-latest.yml
```diff
@@ -93,4 +93,4 @@ jobs:
       - name: Execute Kani performance ignored tests
         working-directory: ./kani
         continue-on-error: true
-        run: cargo run -p compiletest -- --suite perf --mode cargo-kani-test ignore --ignored
+        run: cargo run -p compiletest -- --suite perf --mode cargo-kani-test ignore --ignored --no-fail-fast
```
zhassan-aws, line 96: Do we need both the `ignore` and `--ignored` arguments?

zhassan-aws (approved):

.github/workflows/cbmc-latest.yml
celinval: Yes, we do. The "ignore" filters all tests that don't have ignore in their name. The `--ignored` tells rust to execute tests even if they are marked as an ignore test.

I.e., without the `ignore` argument, all tests will run. Without the `--ignored`, the ignore tests will be ignored. 🙂

PR #2142, opened 2023-01-23
Minor refactoring for CI scripts

zhassan-aws:
### Description of changes: 

This is a follow-up to @celinval's [comment](https://github.com/model-checking/kani/pull/2087#discussion_r1083025802) from #2087. Instead of having to call multiple installation scripts (CBMC, Viewer, and Kissat), `install_deps.sh` now installs them, so one only needs to invoke one script to install all the dependencies.

### Testing:

* How is this change tested? CI

* Is this a refactor change? Yes

### Checklist
- [X] Each commit message has a non-empty body, explaining why the change was made
- [X] Methods or procedures are documented
- [X] Regression or unit tests are included, or existing tests cover the modified code
- [X] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #2158, opened 2023-01-25
Fix nightly CBMC regressions

zhassan-aws:
### Description of changes: 

#2142 broke the nightly CBMC regressions, as the refactoring caused the CBMC release to be installed, which shouldn't be installed in the nightly CBMC regressions since we build it from source.

This PR fixes this issue by updating the CBMC nightly workflow to call the individual setup scripts (instead of relying on the setup action) so that it can skip installing CBMC.

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* Is this a refactor change? Partly.

### Checklist

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -5,10 +5,6 @@ inputs:
   os:
     description: In which Operating System is this running
     required: true
-  install_cbmc:
-    description: Whether to install CBMC
-    required: false
-    default: 'true'
 runs:
   using: composite
   steps:
--- a/.github/workflows/cbmc-latest.yml
+++ b/.github/workflows/cbmc-latest.yml
@@ -41,10 +41,16 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Setup Kani Dependencies
-        uses: ./.github/actions/setup
-        with:
-          os: ${{ matrix.os }}
-          install_cbmc: 'false'
+        run: |
+          ./scripts/setup/${{ matrix.os }}/install_packages.sh
+          ./scripts/setup/${{ matrix.os }}/install_viewer.sh
+          ./scripts/setup/install_kissat.sh
+          ./scripts/setup/install_rustup.sh
+
+      - name: Update submodules
+        run: |
+          git submodule update --init --depth 1
+        shell: bash
 
       - name: Build Kani
         run: cargo build-dev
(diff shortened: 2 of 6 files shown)


tautschnig:
Wouldn’t it make for a much smaller change if you just uninstalled CBMC after running the scripts?

zhassan-aws:
> Wouldn’t it make for a much smaller change if you just uninstalled CBMC after running the scripts?

Thanks for the suggestion @tautschnig. The current failure occurs while installing CBMC via the scripts, because it's already installed (by building from source). We could perhaps fix that by reversing the order so that the CBMC installation through the scripts is done first, and then uninstalled before installing from source. But we would also need uninstallation scripts that are platform dependent (e.g. `brew uninstall`/`dpkg -r`).

tautschnig:
> > Wouldn’t it make for a much smaller change if you just uninstalled CBMC after running the scripts?
> 
> Thanks for the suggestion @tautschnig. The current failure occurs while installing CBMC via the scripts, because it's already installed (by building from source). We could perhaps fix that by reversing the order so that the CBMC installation through the scripts is done first, and then uninstalled before installing from source. But we would also need uninstallation scripts that are platform dependent (e.g. `brew uninstall`/`dpkg -r`).

But isn't it really just a (literal) one-liner for each platform? So I'd support the idea of reversing the order in the interest of keeping things as simple as possible.

celinval:
I have a question.... if we do revert the order, shouldn't the installation from source override the system one?

zhassan-aws:
I just recalled that the current order (installing CBMC from source first) was to avoid an issue with building CBMC from source when it is checked out after Kani is. See #2128 for details.

tautschnig:
> I just recalled that the current order (installing CBMC from source first) was to avoid an issue with building CBMC from source when it is checked out after Kani is. See #2128 for details.

What I don't understand is why these relationships/conflicts need to exist. Can't we just fix this by checking out and building CBMC in a different directory, somewhere next to the `kani` folder rather than as a subdirectory? Or work around the problem by deleting the `Cargo.toml` file? Maybe I'm too eager in applying it, but I do prefer to abide by [KISS](https://en.wikipedia.org/wiki/KISS_principle).

zhassan-aws:
> if we do revert the order, shouldn't the installation from source override the system one?

It should, but I wasn't 100% sure whether the installation from source, which is done using `cmake --target install` places the binaries in the same path that `brew install` and `dpkg` do so that they get overridden, and if they don't, that the built-from-source binaries are the ones that get picked up. That's why I thought it's safer/cleaner to not perform the CBMC system installation in the first place.

karkhaz:
Again advocating for the KISS approach: to avoid even the need to re-order, would it be easier to build CBMC from source... then don't even install it, just set `PATH=$(pwd)/cbmc-src/build/bin:${PATH}` in subsequent steps. Kani isn't linking against CBMC, right, it's just calling binaries?

zhassan-aws:
> So I'd support the idea of reversing the order in the interest of keeping things as simple as possible.

I opened an alternative PR (#2161) which does that. I'm OK with either approach.

zhassan-aws:
> then don't even install it, just set PATH=$(pwd)/cbmc-src/build/bin:${PATH} in subsequent steps. 

That's a good idea. I'll update #2161 to do that.

celinval:
I think #2161 is much cleaner.

zhassan-aws:
Closing in favor of #2161.

Issue #2165, opened 2023-01-27
Assess scan should include failures to build a package


celinval:
Requested feature: When running `cargo kani assess scan` users should have visibility to the packages that Kani failed to assess. 
Use case: We should provide more visibility to the user why `assess` failed to analyze some crates.
Link to relevant documentation (Rust reference, Nomicon, RFC):

Test case: Run the assess script to analyze the top 100 failures. The results will have the following info:

```rust
Assessed 82 successfully, with 18 failures. 
```

However, there is no visibility to why those failures occurred. Manually inspecting some of the logs, it looks like there are a diverse set of failures that include some `kani-compiler` crashes, some ambiguous package selection issues while invoking Cargo...

celinval:
This is related to #2058 but it is the work related to the `scan` part of it.

tedinski:
Also related to #2053: scan has an easy path to doing this, but regular assess did not. Switching behaviors so that regular assess is always single-package only is probably the best path forward.

Issue #2058, opened 2023-01-03
Assess should classify failures to build/analyze packages


tedinski:
When building a lot of packages, assess can fail for a variety of reasons. It'd be great to summarize these and identify the causes.

These include:

1. Kani-compiler panics, which we can aggregate based on the `file:line` of the panic
2. Build failures, where we can both have a "generic" build failure, and ad-hoc recognizers for specific known kinds of failure. (e.g. recognizing "`assert` is ambiguous" from `futures-0.1` as a common specific reason. Building these ad-hoc recognizers might be "unprincipled" in that we just build them for everything we see in practice, but I think that's a good way to proceed: make sure we understand everything because we'll have a metric for "unknown build failure" that should trend towards 0.)
3. Common types of project misconfigurations. E.g. sometimes a native library development package needs to be installed on the host. Or projects can't build with their default features.
4. CBMC / goto-cc / validate-goto-model failures.
5. Other unknown failures.

celinval:
Can we also store more information about the problem on the metadata? For example, the stack trace of a crash.

PR #2166, opened 2023-01-28
Store failure reason into assess metadata

celinval:
### Description of changes: 

We now store Kani's version and the failure reason (if there's one) into the assess metadata. For now, we still need to manually parse this information, but this moves us closer to #2058 and #2165.

In order to collect more details about the compilation errors, we now inspect the cargo diagnostic messages.

### Resolved issues:

N/A

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


celinval:
I still need to implement when and how to disable the text decoration while invoking cargo. Redirecting the output causes a bunch of escape characters to leak.

tests/cargo-kani/asm/global_error/doesnt_call_crate_with_global_asm.expected
```diff
@@ -1,3 +1 @@
 error: Crate crate_with_global_asm contains global ASM, which is not supported by Kani. Rerun with `--enable-unstable --ignore-global-asm` to suppress this error (**Verification results may be impacted**).
-error: could not compile `crate_with_global_asm` due to previous error
```
zhassan-aws, line 2: This is a useful line that points out the name of the crate that had an error. Is the crate name available elsewhere?

kani-driver/src/assess/metadata.rs
```diff
@@ -27,6 +27,12 @@ use super::AssessArgs;
 /// This is not a stable interface.
 #[derive(Serialize, Deserialize)]
 pub struct AssessMetadata {
+    // TODO: The assess cache should check this field to know whether the version should change.
```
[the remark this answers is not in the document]
celinval: Assess scan has an `--existing-only` option that won't rerun assess but only parse existing metadata files. What I'm trying to state here, is that this option shouldn't reuse metadata files that were generated by older versions of `Kani`, since the results might no longer hold.

I can rewrite this comment to make it clearer.

tests/cargo-kani/asm/global_error/doesnt_call_crate_with_global_asm.expected
celinval: There seems to be a bug on Cargo that it counts an extra error if you use `--message-format=json`. Basically, the compiler prints an error message saying how many errors were found. In regular mode, cargo excludes this message, but when emitting `json`, cargo skips that, and counts it as an extra error.

I.e., for this test to pass, I have three options:

1. Remove the check from here.
2. Update the error count:
```
error: could not compile `crate_with_global_asm` due to 2 previous errors
```
3. Update the expected message to not include the number of errors:
```
error: could not compile `crate_with_global_asm`
```

I picked the first since it was the simplest, but I can change. Any preference?

zhassan-aws: I see. The option you went with is fine. I just wanted to make sure the message still gets printed.

celinval: FYI, I updated this one expected test to have the message that includes the name of the crate and the counter as a sanity check, but I didn't bother adding to all the other tests, because one would hope that Cargo will update that counter eventually. In that case, we would only need to update one test.

zhassan-aws (approved):

PR #2167, opened 2023-01-28
Support ICE reporting using json diagnostics

celinval:
### Description of changes: 

This will allow us to retrieve crash information from Kani driver once we merge #2166.

### Resolved issues:

Part of #2166 

### Call-outs:

For now, I'm only added the line where the crashed occurred. I think it's probably our best mechanism today to categorize crashes. Does anyone have another suggestion?

Unfortunately this doesn't cover issues that happen before callback configuration, which is mostly argument parsing.

### Testing:

* How is this change tested? I manually tested this one by injecting a failure in the compiler.

### Checklist

### DIFF
--- a/kani-compiler/src/main.rs
+++ b/kani-compiler/src/main.rs
@@ -44,6 +44,9 @@ use kani_queries::{QueryDb, ReachabilityType, UserInput};
 use rustc_data_structures::fx::FxHashMap;
 use rustc_driver::{Callbacks, RunCompiler};
 use rustc_hir::definitions::DefPathHash;
+use rustc_interface::Config;
+use rustc_session::config::ErrorOutputType;
+use session::json_panic_hook;
 use std::ffi::OsStr;
 use std::path::PathBuf;
 use std::rc::Rc;
@@ -141,7 +144,13 @@ fn main() -> Result<(), &'static str> {
 struct KaniCallbacks {}
 
 /// Use default function implementations.
-impl Callbacks for KaniCallbacks {}
+impl Callbacks for KaniCallbacks {
+    fn config(&mut self, config: &mut Config) {
+        if matches!(config.opts.error_format, ErrorOutputType::Json { .. }) {
+            json_panic_hook();
+        }
+    }
+}
 
 /// The Kani root folder has all binaries inside bin/ and libraries inside lib/.
 /// This folder can also be used as a rustc sysroot.
--- a/kani-compiler/src/session.rs
+++ b/kani-compiler/src/session.rs
@@ -5,6 +5,10 @@
 
 use crate::parser;
 use clap::ArgMatches;
+use rustc_errors::{
+    emitter::Emitter, emitter::HumanReadableErrorType, fallback_fluent_bundle, json::JsonEmitter,
+    ColorConfig, Diagnostic,
+};
 use std::panic;
 use std::str::FromStr;
 use std::sync::LazyLock;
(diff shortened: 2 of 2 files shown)


kani-compiler/src/main.rs
```diff
@@ -141,7 +144,13 @@ fn main() -> Result<(), &'static str> {
 struct KaniCallbacks {}
 
 /// Use default function implementations.
-impl Callbacks for KaniCallbacks {}
+impl Callbacks for KaniCallbacks {
+    fn config(&mut self, config: &mut Config) {
+        if matches!(config.opts.error_format, ErrorOutputType::Json { .. }) {
```
tedinski, line 149: Is `error_format` the right place to hook? I thought that was for compiler errors. Does rustc format panic like this?

celinval: Yes, to the best of my knowledge, that is the best place that works today. That is the option that `rustc` exposes to generate messages in `json` format: https://doc.rust-lang.org/rustc/command-line-arguments.html#--error-format-control-how-errors-are-produced. Which is used by `cargo` when you select `json` message format: https://doc.rust-lang.org/cargo/commands/cargo-build.html#display-options.

I initially tried to invoke `report_ice` API from `rustc`, but that actually did not produce a `json` message while reporting the ICE. I actually think that is a `rustc` issue. When `rustc` crashes, it is not emitting the panic message in `json` format. This is probably because the panic hook is initialized before parsing the arguments.

kani-compiler/src/session.rs
```diff
@@ -30,9 +34,36 @@ static PANIC_HOOK: LazyLock<Box<dyn Fn(&panic::PanicInfo<'_>) + Sync + Send + 's
...
+// Custom panic hook when executing under json error format `--error-format=json`.
+#[allow(clippy::type_complexity)]
+static JSON_PANIC_HOOK: LazyLock<Box<dyn Fn(&panic::PanicInfo<'_>) + Sync + Send + 'static>> =
+    LazyLock::new(|| {
+        let hook = panic::take_hook();
+        panic::set_hook(Box::new(|info| {
+            // Print stack trace.
+            let msg = format!(
+                "Kani unexpectedly panicked at {}",
+                info.location().map_or(String::from("Unknown Location"), |loc| loc.to_string())
```
tedinski, line 51: What does this format as? What a panic typically prints? e.g. `{path}:{line}:{column}` (I think?)

tedinski (approved):

kani-compiler/src/session.rs
celinval: This is an example:
```
{"message":"Kani unexpectedly panicked at compiler/rustc_borrowck/src/region_infer/mod.rs:2196:9","code":null,"level":"error: internal compiler error","spans":[],"children":[],"rendered":"error: internal compiler error: Kani unexpectedly panicked at compiler/rustc_borrowck/src/region_infer/mod.rs:2196:9\n\n"}
```

kani-compiler/src/main.rs
celinval: BTW, I created https://github.com/rust-lang/rust/issues/107484 to see what they way about the panic message format.

kani-compiler/src/session.rs
celinval: btw, I'm changing this message to print the panic info instead, which will print the message and the location.

celinval: So now it will be:
```
{"message":"Kani unexpectedly panicked at panicked at 'no entry found for key', compiler/rustc_borrowck/src/region_infer/mod.rs:2196:9.","code":null,"level":"error: internal compiler error","spans":[],"children":[],"rendered":"error: internal compiler error: Kani unexpectedly panicked at panicked at 'no entry found for key', compiler/rustc_borrowck/src/region_infer/mod.rs:2196:9.\n\n"}
```

PR #2168, opened 2023-01-28
Gracefully exit from kani-driver process

### LINKED ISSUE #2064: Compilation errors produce stacktraces
When `RUST_BACKTRACE=1` is set, Kani produces a stacktrace when there is a compile error. This is annoying for packages that have `RUST_BACKTRACE=1` turned on by default.

Steps to reproduce:

1. `cargo new foo`
2. Put the following in `src/main.rs`:
```rust
fn f(x: i32) {
    assert!(x >= 0);
}

fn main() {
    f('c');
}
```
3. Set the `RUST_BACKTRACE` environment variable to 1 and run `cargo kani`, e.g. `RUST_BACKTRACE=1 cargo kani`. This produces:
```bash
$ RUST_BACKTRACE=1 cargo kani
   Compiling foo v0.1.0 (/home/ubuntu/examples/foo)
 --> src/main.rs:7:7
  |
7 |     f('c');
  |     - ^^^ expected `i32`, found `char`
  |     |
  |     arguments to this function are incorrect
  |
note: function defined here
 --> src/main.rs:1:4
  |
1 | fn f(x: i32) {
  |    ^ ------
help: you can cast a `char` to an `i32`, since a `char` always occupies 4 bytes
  |
7 |     f('c' as i32);
  |           ++++++

For more information about this error, try `rustc --explain E0308`.
error: could not compile `foo` due to previous error

Stack backtrace:
             at /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/anyhow-1.0.68/src/error.rs:83:36
             at /home/ubuntu/git/kani/kani-driver/src/session.rs:141:13
             at /home/ubuntu/git/kani/kani-driver/src/call_cargo.rs:124:17
             at /home/ubuntu/git/kani/kani-driver/src/project.rs:152:19
   4: kani_driver::cargokani_main
             at /home/ubuntu/git/kani/kani-driver/src/main.rs:61:19
   5: kani_driver::main
             at /home/ubuntu/git/kani/kani-driver/src/main.rs:44:44
   6: core::ops::function::FnOnce::call_once
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/core/src/ops/function.rs:507:5
   7: std::sys_common::backtrace::__rust_begin_short_backtrace
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/sys_common/backtrace.rs:121:18
   8: std::rt::lang_start::{{closure}}
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/rt.rs:166:18
   9: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/core/src/ops/function.rs:606:13
  10: std::panicking::try::do_call
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/panicking.rs:483:40
  11: std::panicking::try
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/panicking.rs:447:19
  12: std::panic::catch_unwind
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/panic.rs:137:14
  13: std::rt::lang_start_internal::{{closure}}
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/rt.rs:148:48
  14: std::panicking::try::do_call
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/panicking.rs:483:40
  15: std::panicking::try
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/panicking.rs:447:19
  16: std::panic::catch_unwind
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/panic.rs:137:14
  17: std::rt::lang_start_internal
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/rt.rs:148:20
  18: std::rt::lang_start
             at /rustc/c6fcdb690609769a240fc8ab0de0ce68d5ea7dba/library/std/src/rt.rs:165:17
  19: main
  20: __libc_start_main
  21: _start
$
```

This does not happen with `cargo build` or `cargo run`:
```bash
$ RUST_BACKTRACE=1 cargo run
   Compiling foo v0.1.0 (/home/ubuntu/examples/foo)
 --> src/main.rs:7:7
  |
7 |     f('c');
  |     - ^^^ expected `i32`, found `char`
  |     |
  |     arguments to this function are incorrect
  |
note: function defined here
 --> src/main.rs:1:4
  |
1 | fn f(x: i32) {
  |    ^ ------
help: you can cast a `char` to an `i32`, since a `char` always occupies 4 bytes
  |
7 |     f('c' as i32);
  |           ++++++

For more information about this error, try `rustc --explain E0308`.
error: could not compile `foo` due to previous error
$
```

with Kani version: 48569a261b8

celinval:
### Description of changes: 

When an error that was properly handled ocurrs in `kani-driver` we now gracefully exit and return ExitStatus::FAILURE.

We still print all error context to keep a similar experiece to what it was before. However, this will no longer include the backtrace if users set RUST_BACKTRACE=1.

The backtrace will still be included if you enable the debug logs and also enable the backtrace. I.e.:

```bash
export RUST_BACKTRACE=1
export KANI_LOG=kani_driver=debug
cargo kani
```

### Resolved issues:

Resolves #2064 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/kani-driver/src/main.rs
+++ b/kani-driver/src/main.rs
@@ -48,7 +48,7 @@ fn main() -> ExitCode {
 
     if let Err(error) = result {
         // We are using the debug format for now to print the all the context.
-        // We should consider create a standard for error reporting.
+        // We should consider creating a standard for error reporting.
         debug!(?error, "main_failure");
         util::error(&format!("{error:#}"));
         ExitCode::FAILURE
--- a/kani-driver/src/main.rs
+++ b/kani-driver/src/main.rs
@@ -4,6 +4,7 @@
 #![feature(array_methods)]
 
 use std::ffi::OsString;
+use std::process::ExitCode;
 
 use anyhow::Result;
 
@@ -39,10 +40,20 @@ mod unsound_experiments;
 /// The main function for the `kani-driver`.
 /// The driver can be invoked via `cargo kani` and `kani` commands, which determines what kind of
 /// project should be verified.
-fn main() -> Result<()> {
-    match determine_invocation_type(Vec::from_iter(std::env::args_os())) {
+fn main() -> ExitCode {
+    let result = match determine_invocation_type(Vec::from_iter(std::env::args_os())) {
         InvocationType::CargoKani(args) => cargokani_main(args),
         InvocationType::Standalone => standalone_main(),
+    };
+
+    if let Err(error) = result {
+        // We are using the debug format for now to print the all the context.
+        // We should consider create a standard for error reporting.
+        debug!(?error, "main_failure");
+        util::error(&format!("{error:#}"));
+        ExitCode::FAILURE
+    } else {
+        ExitCode::SUCCESS
     }
 }
 
--- a/kani-driver/src/util.rs
+++ b/kani-driver/src/util.rs
@@ -95,6 +95,13 @@ pub fn warning(msg: &str) {
     println!("{warning} {msg_fmt}")
 }
 
+/// Print a warning message. This will add a "warning:" tag before the message and style accordinly.
+pub fn error(msg: &str) {
+    let error = console::style("error:").bold().red();
+    let msg_fmt = console::style(msg).bold();
+    println!("{error} {msg_fmt}")
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/kani-driver/src/util.rs
+++ b/kani-driver/src/util.rs
@@ -88,14 +88,14 @@ pub fn specialized_harness_name(linked_obj: &Path, harness_filename: &str) -> Pa
     alter_extension(linked_obj, &format!("for-{harness_filename}.out"))
 }
 
-/// Print a warning message. This will add a "warning:" tag before the message and style accordinly.
+/// Print a warning message. This will add a "warning:" tag before the message and style accordingly.
 pub fn warning(msg: &str) {
     let warning = console::style("warning:").bold().yellow();
     let msg_fmt = console::style(msg).bold();
     println!("{warning} {msg_fmt}")
 }
 
-/// Print a warning message. This will add a "warning:" tag before the message and style accordinly.
+/// Print an error message. This will add an "error:" tag before the message and style accordingly.
 pub fn error(msg: &str) {
     let error = console::style("error:").bold().red();
     let msg_fmt = console::style(msg).bold();
--- a/tests/cargo-kani/asm/global_error/doesnt_call_crate_with_global_asm.expected
+++ b/tests/cargo-kani/asm/global_error/doesnt_call_crate_with_global_asm.expected
@@ -1,3 +1,3 @@
 error: Crate crate_with_global_asm contains global ASM, which is not supported by Kani. Rerun with `--enable-unstable --ignore-global-asm` to suppress this error (**Verification results may be impacted**).
 error: could not compile `crate_with_global_asm` due to previous error
-Error: cargo exited with status exit status: 101
+error: cargo exited with status exit status: 101
--- a/tests/cargo-kani/simple-proof-annotation/main.expected
+++ b/tests/cargo-kani/simple-proof-annotation/main.expected
@@ -1,1 +1,1 @@
-Error: A proof harness named main was not found
+error: A proof harness named main was not found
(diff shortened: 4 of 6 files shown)


celinval:
It makes sense. What you are suggesting is already the current behavior though. The standard library checks those variables before generating the backtrace.

That said, it doesn't fix the problem at hand. So I think the question here is whether #2064 should be fixed in the first place. @zhassan-aws ?

tedinski:
I think we should fix it. The behavior I was suggesting is this:

> If you want only panics to have backtraces, set `RUST_BACKTRACE=1` and `RUST_LIB_BACKTRACE=0`.

So to fix this in `kani-driver`, if the former is set, and the latter is not, then we explicitly set the latter to 0.

celinval:
Ah, I see what you are saying, you think we should change the behavior only when `RUST_BACKTRACE=1` and `RUST_LIB_BACKTRACE` isn't set, right? Why is that? Do you think people just aren't used to the second variable?

This PR mimics `cargo` and `rustc` behavior, which I think makes more sense. No matter the value of RUST_BACKTRACE and RUST_LIB_BACKTRACE, these two tools won't print the stack trace of an error that is due to a user error (like compilation error). Note that today passing the wrong argument will not print the error stack, no matter the values of these two variables.

No matter which approach we take, we should consider improving the driver error handling to make a distinction between internal tool error and user error, so this can be handled in a better way.

zhassan-aws:
> So I think the question here is whether #2064 should be fixed in the first place. @zhassan-aws ?

I think it should be fixed, because a backtrace gives an impression that something is broken in the tool.

kani-driver/src/main.rs
```diff
@@ -39,10 +40,20 @@ mod unsound_experiments;
...
-fn main() -> Result<()> {
-    match determine_invocation_type(Vec::from_iter(std::env::args_os())) {
+fn main() -> ExitCode {
+    let result = match determine_invocation_type(Vec::from_iter(std::env::args_os())) {
         InvocationType::CargoKani(args) => cargokani_main(args),
         InvocationType::Standalone => standalone_main(),
+    };
+
+    if let Err(error) = result {
+        // We are using the debug format for now to print the all the context.
```
zhassan-aws, line 50:
```suggestion
        // We are using the debug format for now to print all the context.
```

kani-driver/src/main.rs
```diff
@@ -39,10 +40,20 @@ mod unsound_experiments;
...
+        // We should consider create a standard for error reporting.
```
zhassan-aws, line 51: "consider create" -> "create" or "consider creating"

kani-driver/src/util.rs
```diff
@@ -95,6 +95,13 @@ pub fn warning(msg: &str) {
     println!("{warning} {msg_fmt}")
 }
 
+/// Print a warning message. This will add a "warning:" tag before the message and style accordinly.
```
zhassan-aws, line 98:
```suggestion
/// Print an error message. This will add an "error:" tag before the message and style accordingly.
```

zhassan-aws (approved):
A few minor comments. Otherwise, looks good (if @tedinski is OK with the changes).

tedinski:
> Do you think people just aren't used to the second variable?

Pretty much.

Also, just want to make it clear: I don't object to this change.

celinval:
> > Do you think people just aren't used to the second variable?
> 

Yeah, tbh, I am not a big fan of the design they came up with these two variables. It's not very intuitive.


> Pretty much.
> 
> Also, just want to make it clear: I don't object to this change.

Sounds good. Thanks!

PR #2171, opened 2023-01-31
Fix typo in Kani result

qinheping:
### Description of changes: 

Fix a typo in user-visible results.


zhassan-aws (approved):
Good catch!

PR #2172, opened 2023-01-31
Increase width's `width` for trace values

### LINKED ISSUE #2169: Unhandled variant in ParserItem JSON
I'm working on a [lock-free, async spsc channel in s2n-quic](https://github.com/aws/s2n-quic/pull/1614) and wanted to write a kani proof for it. I have a [harness written that simply allocates the channel](https://github.com/aws/s2n-quic/blob/8c79376f82e6281b7dc808e08c5cdf6a417a951d/quic/s2n-quic-core/src/sync/spsc/tests.rs#L270-L296). After running Kani, I get a panic inside the kani-driver crate:

```
...snip...
Runtime Symex: 1.36531s
size of program expression: 83108 steps
slicing removed 47540 assignments
Generated 4442 VCC(s), 2676 remaining after simplification
Runtime Postprocess Equation: 0.0723243s
converting SSA
Runtime Convert SSA: 0.339988s
Post-processing
Runtime Post-process: 1.14201s
672691 variables, 3828307 clauses
Runtime Solver: 70.2977s
Runtime decision procedure: 70.6452s
672691 variables, 1113105 clauses
Runtime Solver: 0.0784595s
Runtime decision procedure: 0.0795185s
672691 variables, 1111673 clauses
SAT checker: instance is UNSATISFIABLE
Runtime Solver: 5.03282s
Runtime decision procedure: 5.04014s
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Error("data did not match any variant of untagged enum ParserItem", line: 0, column: 0)', kani-driver/src/cbmc_output_parser.rs:425:21
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: <kani_driver::cbmc_output_parser::Parser as core::iter::traits::iterator::Iterator>::next
   4: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   5: kani_driver::cbmc_output_parser::process_cbmc_output
   6: kani_driver::harness_runner::<impl kani_driver::session::KaniSession>::check_harness
   7: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &mut F>::call_once
   8: <alloc::vec::Vec<T,A> as alloc::vec::spec_extend::SpecExtend<T,I>>::spec_extend
   9: rayon::iter::plumbing::bridge_producer_consumer::helper
  10: <rayon::iter::plumbing::bridge::Callback<C> as rayon::iter::plumbing::ProducerCallback<I>>::callback
  11: rayon::iter::from_par_iter::collect_extended
  12: rayon::iter::ParallelIterator::collect
  13: std::panic::catch_unwind
  14: <rayon_core::job::StackJob<L,F,R> as rayon_core::job::Job>::execute
  15: rayon_core::registry::WorkerThread::wait_until_cold
  16: rayon_core::registry::ThreadBuilder::run
```

To reproduce, you can run the following command:

```sh
$ git clone --branch camshaft/spsc https://github.com/aws/s2n-quic.git
$ cd s2n-quic/quic/s2n-quic-core
$ RUST_BACKTRACE=1 RUSTFLAGS="--cfg kani_slow" cargo kani --harness alloc_test --tests
```

with Kani version: 0.20

adpaco-aws:
### Description of changes: 

 * Changes `TraceValue.width` to be `Option<u32>` instead of `Option<u8>` (this fixes #2169).
 * Adds a case in `parse_item` to provide better error messages for `ParserItem::Result`.

### Resolved issues:

Resolves #2169 

### Testing:

* How is this change tested? CI + manually tested on example from #2169 (the example takes a few minutes to run, which is why I'm not including it in the regression).

### Checklist


kani-driver/src/cbmc_output_parser.rs
```diff
@@ -35,6 +35,8 @@ use std::os::unix::process::ExitStatusExt;
 use std::path::PathBuf;
 use std::process::{Child, ChildStdout};
 
+const RESULT_ITEM_PREFIX: &str = "  {\n    \"result\":";
```
celinval, line 38: Should we use regex instead or will this run all the time? I'm worried this is very fragile.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -420,6 +433,24 @@ impl<'a, 'b> Parser<'a, 'b> {
...
+        // Error("data did not match any variant of untagged enum ParserItem", line: 0, column: 0)'
+        // ```
+        // This error message doesn't provide information about what went wrong
+        // while parsing due to `ParserItem` being an untagged enum. A more
+        // informative error message will be produced if we attempt to
+        // deserialize it into a struct. The attempt will still fail, but it
+        // shouldn't be hard to debug with that information. The same strategy
+        // can be used for other `ParserItem` variants, but they're normally
+        // easier to debug.
+        if string_without_delimiter.starts_with(RESULT_ITEM_PREFIX) {
```
celinval, line 450: Have you considered changing the signature of this function to Result<ParserItem> and in the case of failure you can add the context that we failed on (i.e., what was the item that we were trying to parse).

celinval (commented):
A few notes.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -285,7 +298,7 @@ pub struct TraceValue {
     pub name: String,
     pub binary: Option<String>,
     pub data: Option<TraceData>,
-    pub width: Option<u8>,
+    pub width: Option<u32>,
```
tedinski, line 301: Could we stick a small unit test in this file with a reduce version of the string we saw? It should be possible to just deserialize a string to a `TraceValue` to unit test just this code. It might help when we make changes to this in the future to accumulate a list of corner cases we ran into.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: We don't know which item we're trying to parse, that's why we use the `ParserItem` enum. I can change the signature to use `Result<ParserItem>` but I don't know what to with the result other than `unwrap()`.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: Yes, let me do that.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: I made the change but then dropped it because the new case would return a `ResultStruct`. I'm open to suggestions here.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: There were a few things I considered here, for example trimming the string without delimiters (so we can check against `"{\"result\":"`). The problem is that trimming that string could impact performance: the one that caused the crash in #2169 is 921K lines (33MB). So I went for the most performant check (this one), which should run all the time.

kani-driver/src/cbmc_output_parser.rs
celinval: Could you use `expect()` and add to the error message the string that we failed to parse.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -681,4 +713,17 @@ mod tests {
             serde_json::from_str(prop_id_string);
         let _prop_id = prop_id_result.unwrap();
     }
+
+    #[test]
+    fn check_trace_value_deserialization_works() {
+        let data = r#"{
+            "binary": "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
```
celinval, line 720: nit: can you generate this string using padding format? Something like:

```rust
   let data = format!(r#"{{
            "binary": "{:0>1000}",
            "data": "0",
            "name": "integer",
            "type": "unsigned __CPROVER_bitvector[960]",
            "width": 960
        }}"#, 0);
```

Just replace the 1000 by whatever number you want.

kani-driver/src/cbmc_output_parser.rs
celinval: Thanks for explaining the issue offline. I'll resolve this comment.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: Done.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: I cannot get rid of `#[allow(unused)]` even if use `ResultStruct.result` in the test.

kani-driver/src/cbmc_output_parser.rs
adpaco-aws: Done.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -681,4 +713,45 @@ mod tests {
...
+            0
+        );
+        let trace_value: Result<TraceValue, _> = serde_json::from_str(&data);
+        assert!(trace_value.is_ok());
+    }
+
+    /// Checks that a valid CBMC "result" item can be deserialized into a
+    /// `ParserItem` or `ResultStruct`.
+    #[test]
+    fn check_result_deserialization_works() {
```
celinval, line 736: Thanks!

celinval (approved):
Thanks

Issue #2137, opened 2023-01-19
Improve how we pass Kani codegen specific options to `kani-compiler`


celinval:
**Proposed change:** Instead of using an environment variable or relying on the arguments order to pass information to `kani-compiler`, we should either use `-C llvm-args=` or (ideally) help create `-C backed-args=` and use them with `kani` backend (as [discussed here](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/Tools.20.2F.20custom.20backend.20arguments)). 

**Motivation:** This is a blocker for reusing compilation artifacts in Kani (#2036). The current mechanism doesn't work well with Cargo's caching. Cargo uses the provided `rustc` flags to compute its artifacts cache, and to compute when it needs to recompile an artifact. This is one of the reasons we explicitly do a `cargo clean` every time users invoke `cargo kani`.

This is also hacky and rather fragile.

celinval:
Note that for us to use `-C llvm-args` today, we need to make sure we only set them for the target platform, not the host. Even though today they are the same, this will avoid us passing Kani specific flags to LLVM backend when we compile build scripts and proc-macros.

Issue #2036, opened 2022-12-23
Kani needlessly recompiles everything every time


Kixunil:
When trying to get kani to work one often has to do small changes to a file. Kani recompiles all crates all over again even when only one file changed. Including proc macros, so if there's some technical reason why the code can't be reused for conversion to CBMC at least proc macros should be reusable.

zhassan-aws:
Hi @Kixunil. This is indeed very annoying. We've had to disable incremental compilation because of a few bugs that we're working on fixing. This is tracked in #1736.

PR #2174, opened 2023-02-02
Move all rustc arguments to kani-driver

celinval:
### Description of changes: 

Move all rustc arguments to kani-driver

1. It's good to have one source of truth.
2. This is needed for us to properly support caching compiler artifacts

### Resolved issues:

None. It's part of #2137 and needed for #2036.

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

### Checklist

### DIFF
--- a/kani-driver/src/call_cargo.rs
+++ b/kani-driver/src/call_cargo.rs
@@ -114,6 +114,8 @@ impl KaniSession {
                     .args(&pkg_args)
                     .env("RUSTC", &self.kani_compiler)
                     .env("KANIFLAGS", kani_args.join(" "))
+                    // Use CARGO_ENCODED_RUSTFLAGS instead of RUSTFLAGS is preferred. See
+                    // https://doc.rust-lang.org/cargo/reference/environment-variables.html
                     .env("CARGO_ENCODED_RUSTFLAGS", rustc_args.join(OsStr::new("\x1f")))
                     .env("CARGO_TERM_PROGRESS_WHEN", "never");
 
--- a/kani-driver/src/call_cargo.rs
+++ b/kani-driver/src/call_cargo.rs
@@ -2,11 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use crate::args::KaniArgs;
-use crate::session::{KaniSession, ReachabilityMode};
+use crate::session::KaniSession;
 use anyhow::{bail, Context, Result};
 use cargo_metadata::diagnostic::{Diagnostic, DiagnosticLevel};
 use cargo_metadata::{Message, Metadata, MetadataCommand, Package};
-use std::ffi::OsString;
+use std::ffi::{OsStr, OsString};
 use std::fs;
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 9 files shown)


kani-driver/src/call_cargo.rs
```diff
@@ -125,8 +113,8 @@ impl KaniSession {
                     .args(&target.to_args())
                     .args(&pkg_args)
                     .env("RUSTC", &self.kani_compiler)
-                    .env("RUSTFLAGS", "--kani-flags")
-                    .env("KANIFLAGS", &crate::util::join_osstring(&kani_args, " "))
+                    .env("KANIFLAGS", kani_args.join(" "))
+                    .env("CARGO_ENCODED_RUSTFLAGS", rustc_args.join(OsStr::new("\x1f")))
```
zhassan-aws, line 117: Can you clarify what this does?

kani-driver/src/call_single_file.rs
```diff
@@ -93,19 +85,61 @@ impl KaniSession {
...
         flags
     }
 
-    /// These arguments are arguments passed to kani-compiler that are `rustc` specific.
-    /// These are also used by call_cargo to pass as the env var KANIFLAGS.
+    /// This function generates all rustc configurations required by our goto-c codegen.
+    ///
+    /// We override the `std` library using the mechanism described here:
+    /// <https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/.E2.9C.94.20Globally.20override.20an.20std.20macro/near/268873354>
+    /// for more details.
```
zhassan-aws, line 105: Remove this line.

scripts/std-lib-regression.sh
```diff
@@ -50,8 +50,7 @@ cd std_lib_test
 
 # Add some content to the rust file including an std function that is non-generic.
 echo '
-#[kani::proof]
```
zhassan-aws, line 53: Why was this removed?

scripts/std-lib-regression.sh
```diff
@@ -68,7 +67,7 @@ echo "Starting cargo build with Kani"
 export RUST_BACKTRACE=1
 export RUSTC_LOG=error
 export KANIFLAGS="--goto-c --ignore-global-asm --reachability=legacy"
-export RUSTFLAGS="--kani-flags"
+export RUSTFLAGS="--kani-flags -C panic=abort"
```
zhassan-aws, line 70: Why this change?

kani-driver/src/call_cargo.rs
celinval: It's the same as RUSTFLAGS. Per [their documentation](https://doc.rust-lang.org/cargo/reference/environment-variables.html), it is the recommended way. I don't know if it matters much here, but I figured why not:

> This string is split by whitespace; for a more robust encoding of multiple arguments, see CARGO_ENCODED_RUSTFLAGS.

scripts/std-lib-regression.sh
celinval: So we don't have to manually inject Kani dependency here. Since our goal for this script is to make sure we can generate code for the standard library, this seems much easier to maintain than duplicating a bunch of the compiler flags here.

scripts/std-lib-regression.sh
celinval: I had to change this because in this script we are invoking `kani-compiler` via `cargo build` not via `cargo kani`. This is the only relevant flag for the codegen to go through.

kani-driver/src/call_cargo.rs
zhassan-aws: Please add a link to the documentation

zhassan-aws (approved):

kani-driver/src/call_single_file.rs
zhassan-aws: Sorry, I just meant the last line of the comment: 
```
/// for more details.
```
to fix the sentence and not the entire comment. The rest of the comment is still relevant.

Issue #1194, opened 2022-05-12
Cleanup Kani's output


celinval:
**Requested feature:** Provide a cleaner user interface by default and allow users to control its verbosity.
**Use case:** Kani output can be very verbose which makes it hard to read.

**Test case:** Kani prints 2000+ lines (400+ properties) in this testcase taken from tests/kani/Vectors/vector_extend.rs:

```rust
#[kani::proof]
fn main() {
    let mut v: Vec<u32> = Vec::new();
    v.extend(42..=42);
    assert!(v[0] == 42);
}
```
**Proposed solution:**

Provide the following verbosity options to `kani` and `cargo-kani`:

- **Default:** Similar to today's output but without CBMC stats and only include user assertions for each harness.
- **--quiet**: Similar interface to `cargo test --quiet`. I.e.:
```
running 5 harnesses
.....
verification result: ok. 5 passed; 0 failed; finished in 0.00s
```
  with failure Kani displays the *terse* output for the failed tests:
```
running 5 harnesses
..F..
verification result: FAILED. 4 passed; 1 failed; finished in 0.00s

---- harness: <name> ----
VERIFICATION RESULT: 
 ** 1 of 100 failed (10 unreacheable)
Failed Checks: assertion failed: 1 + 1 == 3
 File: "/home/ANT.AMAZON.COM/celinval/workspace/kani-project/kani/tests/ui/terse-output-format-fail/fail.rs", line 7, in main

VERIFICATION:- FAILED

failures:
    <harness_name>
```
- **--verbose**: Today's default interface including all checks.
- **--debug**: Verbose output that also include debug logs.
- **--raw**: Hidden option to print CBMC's output as is. Implies debug.

For simplicity, I would prefer if we retire today's `--output-format`.

Alternative solution:

We could use `--output-format [quiet | regular | verbose | debug | raw]`.

jaisnan:
What about warnings? Do we still display them by default?

celinval:
That's a good point. For question 2, I would say we print warnings for everything except the tests that pass when using the `--quiet` option. 

For question 3, you got it right. I was thinking that it should only be used for debug purpose; hence, we assume debug verbosity.

zhassan-aws:
I agree with the proposal.

> For simplicity, I would prefer if we retire today's --output-format.

+1

> --raw: Hidden option to print CBMC's output as is. Implies debug.

Can we reduce it to imply `--verbose` only? Debug logs can contain too much information.

[...]

PR #2177, opened 2023-02-02
Update assertion description with a less confusing alternative

### LINKED ISSUE #1240: Proper message for successful assertion
Requested feature:
Replace `assertion failed` with another clearer message like `assertion holds` when an assertion is indeed provable.

Test case:
This example in the doc:

https://github.com/model-checking/kani/blob/6213b0b2200f9e468f2494c738bdc19cb31ef222/docs/src/verification-results.md?plain=1#L37

```rust
fn success_example() {
    let mut sum = 0;
    for i in 1..4 {
        sum += i;
    }
    assert_eq!(sum, 6);
}
```

outputs:

```bash
Check 4: success_example.assertion.4
         - Status: SUCCESS
         - Description: "assertion failed: sum == 6"
```

But the assertion did not fail, it actually holds.

Note: This may be linked to [https://github.com/model-checking/kani/issues/189](https://github.com/model-checking/kani/issues/189), but the issue was closed.

adpaco-aws:
### Description of changes: 

Replaces `"assertion failed"` with `"assertion holds"` in the description of checks from the `assertion` class.

The new description is less confusing for users.

### Resolved issues:

Resolves #1240 

### Call-outs:
 * [This comment](https://github.com/model-checking/kani/issues/1240#issuecomment-1144180339) mentions that there are other cases where we don't have control over the message, but I don't if this was considering doing it during post-processing.
 * We can discuss other options to update the message, it's not a problem. For example, I initially though about replacing `"assertion failed"` with just `"assertion"`, but the message is short and doesn't really describe anything. So I'm using the proposed text instead, but I'm open to suggestions.

### Testing:

* How is this change tested? Existing tests.

### Checklist

### DIFF
--- a/docs/src/getting-started/verification-results/failure_example.expected
+++ b/docs/src/getting-started/verification-results/failure_example.expected
@@ -1,2 +1,2 @@
 FAILURE\
-Description: "assertion failed: arr.len() != 3"
+Description: "assertion holds: arr.len() != 3"
--- a/docs/src/getting-started/verification-results/success_example.expected
+++ b/docs/src/getting-started/verification-results/success_example.expected
@@ -1,2 +1,2 @@
 SUCCESS\
-Description: "assertion failed: sum == 6"
+Description: "assertion holds: sum == 6"
--- a/docs/src/getting-started/verification-results/undetermined_example.expected
+++ b/docs/src/getting-started/verification-results/undetermined_example.expected
@@ -1,2 +1,2 @@
 UNDETERMINED\
-Description: "assertion failed: x == 0"
+Description: "assertion holds: x == 0"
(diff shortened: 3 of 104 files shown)


celinval (changes requested):
Nope. Sorry, this is not how we fix this problem. Panic messages are meant to describe the error that was found, not a property. Just renaming the prefix won't change anything. Also, after your change, the summary will make zero sense. We will print something like:
```
Failed Checks: assertion holds: x < 8
```

adpaco-aws:
After a discussion with @celinval , I'm convinced that this is only a short-term patch and doesn't solve the actual problem, which is our overloaded and sometimes confusing output.

There are clear differences between properties (checks) and panic messages, and this is point to be considered in a future proposal to re-design our output. Of course, we must not overwhelm users with too much output, so it's necessary that we collapse the output as much as possible. Regardless, I'm closing this PR in favor of #1194 / #1690.

PR #2181, opened 2023-02-02
Work around loop identification issues with CBMC

### LINKED ISSUE #2164: Kani fails to verify harness with `for` loop with bounded iterations
I tried this code after looking into how Kani is handling `Option::ok_or()`:

```rust
/// Dummy function with a for loop that only runs 2 iterations.
fn bounded_loop<T: Default>(b: bool, other: T) -> T {
    let mut ret = other;
    for i in 0..2 {
        ret = match b {
            true => T::default(),
            false => ret,
        };
    }
    return ret;
}

/// Harness that should succeed. We add a conservative loop bound.
#[kani::proof]
#[kani::unwind(10)]
fn harness() {
    let _ = bounded_loop(kani::any(), ());
}

```

```
kani harness.rs
```

with Kani version: 0.20

I expected to see this happen: The verification should succeed.
Instead, this happened: Verification fails due to unwinding assertion failure.

<details><summary>Open here for a few notes:</summary>

### Possibly related issue

I bumped into this issue while debugging a failure in #2149 where Kani incorrectly identifies a loop in `Option::ok_or()`. Verification may fail if user sets a low loop bound. This is an example of the original issue:

```rust
fn loop_free<T: Default>(b: bool, other: T) -> T {
    match b {
        true => T::default(),
        false => other,
    }
}

#[kani::proof]
#[kani::unwind(2)]
fn check_our_ok_or() {
    let b : bool = kani::any();
    let _ = loop_free(b, 5));
}
```
See https://github.com/diffblue/cbmc/issues/7506 for the related CBMC issue.

With that in mind, I decided to see what would happen if the problematic code was inside a loop, and it looks like things just don't work.

### Drop support

I believe both issues are related to the CFG generated when the code includes drop statements. Since the function `bounded_loop` is generic, its MIR will include drop statements for cases where the type `T` must be dropped. The drop statement is added to a basic block at the end (in order of declaration) of the function.

If we add `Copy` to the function signature, the issue goes away and the CFG is much simpler.

```rust
fn bounded_loop<T: Default + Copy>(b: bool, other: T) -> T;
```
</details>

celinval:
### Description of changes: 

CBMC has a heuristic to identify loops that identifies any jump to a previously declared basic block as the back edge of a loop.

For rust, this happens often due to drop elaboration. See https://rustc-dev-guide.rust-lang.org/mir/drop-elaboration.html#drop-elaboration-1 for more details.

To avoid that issue, we now codegen basic blocks in topological order.

### Resolved issues:

Resolves #2164 

### Testing:

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
@@ -12,6 +12,7 @@ use kani_queries::UserInput;
 use rustc_ast::Attribute;
 use rustc_hir::def::DefKind;
 use rustc_hir::def_id::DefId;
+use rustc_middle::mir::traversal::reverse_postorder;
 use rustc_middle::mir::{Body, HasLocalDecls, Local};
 use rustc_middle::ty::layout::FnAbiOf;
 use rustc_middle::ty::{self, Instance};
@@ -88,7 +89,7 @@ impl<'tcx> GotocCtx<'tcx> {
             self.codegen_function_prelude();
             self.codegen_declare_variables();
 
-            mir.basic_blocks.iter_enumerated().for_each(|(bb, bbd)| self.codegen_block(bb, bbd));
+            reverse_postorder(mir).for_each(|(bb, bbd)| self.codegen_block(bb, bbd));
 
             let loc = self.codegen_span(&mir.span);
             let stmts = self.current_fn_mut().extract_block();
--- a/tests/kani/Loops/loop_with_drop.rs
+++ b/tests/kani/Loops/loop_with_drop.rs
@@ -1,0 +1,23 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+//! Ensure that Kani correctly unwind the loop with drop instructions which may .
+//! This was related to https://github.com/model-checking/kani/issues/2164
+/// Dummy function with a for loop that only runs 2 iterations.
+fn bounded_loop<T: Default>(b: bool, other: T) -> T {
+    let mut ret = other;
+    for i in 0..2 {
+        ret = match b {
+            true => T::default(),
+            false => ret,
+        };
+    }
+    return ret;
+}
+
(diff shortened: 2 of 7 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -88,7 +89,7 @@ impl<'tcx> GotocCtx<'tcx> {
             self.codegen_function_prelude();
             self.codegen_declare_variables();
 
-            mir.basic_blocks.iter_enumerated().for_each(|(bb, bbd)| self.codegen_block(bb, bbd));
+            reverse_postorder(mir).for_each(|(bb, bbd)| self.codegen_block(bb, bbd));
```
tedinski, line 92: Sweet.

tedinski (approved):
Want to add a comment about removing this, with an issue related to the more fundamental CBMC fix? Or should we consider just keeping this change permanently?

zhassan-aws (approved):
Ha! A lot simpler than we thought!

celinval:
> Want to add a comment about removing this, with an issue related to the more fundamental CBMC fix? Or should we consider just keeping this change permanently?

I don't see a reason to remove this. I don't think there is any cost here, since rustc caches this information. Even if we have to rerun this, it will be a simple DFS on the basic blocks.

On the plus side, besides working around the CBMC issue, it tends to simplify the CFG for the code we generate which makes it easier to debug later.

tests/kani/Loops/loop_with_drop.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+//! Ensure that Kani correctly unwind the loop with drop instructions which may .
```
adpaco-aws, line 4: "unwind" -> "unwinds"

Also, the sentence isn't finished?

celinval: oops.

PR #2185, opened 2023-02-03
Change kani-compiler to use `-C llvm-args` for Kani specific flags

### LINKED ISSUE #2137: Improve how we pass Kani codegen specific options to `kani-compiler`

celinval:
### Description of changes: 

Changed how `kani-driver` passess arguments to `kani-compiler`. We now have one option, `--kani-compiler`, that changes how we configure rustc, and all the other Kani specific arguments are added to `-C llvm-args`. We no longer need to keep the arguments in a certain order or need to use KANI_FLAGS environment variable.

In the `kani-compiler` we no longer parse the arguments before invoking rustc, we initially detect whether users want to use the Kani flavour of the compiler by looking for a `--kani-compiler` flag.

If `--kani-compiler` flag is found, we configure rustc's driver to use our custom backend and our custom callback. We then invoke the driver with the command line arguments excluding the `--kani-compiler` flag. During the callback configuration we initialize the Kani specific flags that are passed via `-C llvm-args`.

If there isn't any `--kani-compiler` flag, we just invoke rustc with all the given arguments.

### Resolved issues:

Resolves #2137 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

Since we initialize the QueryDb after sharing it, I put the entire structure behind a mutex, and converted its members to simple non-synchronized values.

### Testing:

### Checklist


kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -45,19 +45,19 @@ use std::io::Write as IoWrite;
...
-use std::rc::Rc;
+use std::sync::{Arc, Mutex};
 use std::time::Instant;
 use tempfile::Builder as TempFileBuilder;
 use tracing::{debug, error, info};
 
 #[derive(Clone)]
 pub struct GotocCodegenBackend {
-    queries: Rc<QueryDb>,
+    queries: Arc<Mutex<QueryDb>>,
```
zhassan-aws, line 55: Can you add a comment that explains why this needs Arc/Mutex?

kani-compiler/src/main.rs
```diff
@@ -31,152 +31,27 @@ extern crate tempfile;
...
-        assert_eq!(os_str.to_str(), None);
-
-        let matches = parser::parser().get_matches_from(vec![
-            OsString::from("kani-compiler").as_os_str(),
-            OsString::from("--sysroot").as_os_str(),
-            OsString::from("any").as_os_str(),
-            os_str,
-        ]);
-        generate_rustc_args(&matches);
+    if kani_compiler {
```
zhassan-aws, line 50: Should this be an assert instead? Are there cases where we invoke the driver without `--kani-compiler`?

kani-compiler/src/kani_compiler.rs
```diff
@@ -0,0 +1,182 @@
...
+    Box::new(GotocCodegenBackend::new(queries))
+}
+
+/// Fallback backend. It will trigger an error if no backend has been enabled.
+#[cfg(not(feature = "cprover"))]
+fn backend(queries: Arc<Mutex<QueryDb>>) -> Box<CodegenBackend> {
+    compile_error!("No backend is available. Only supported value today is `cprover`");
+}
+
+/// Empty struct since we don't support any callbacks yet.
```
zhassan-aws, line 65: The comment is misplaced/outdated.

kani-driver/src/call_single_file.rs
```diff
@@ -160,6 +166,17 @@ impl KaniSession {
...
+        // rustc driver is invoked.
+        flags.push("--kani-compiler".into());
+
         flags
     }
 }
+
+/// This function can be used to convert a Kani compiler specific argument into a rustc one.
+/// We currently pass Kani specific arguments using the `--llvm-args` structure which is the
+/// hacky mechanism used by other rustc backend to receive arguments unknown to rustc.
```
zhassan-aws, line 179: Can you elaborate on how llvm-args affects the cache?

zhassan-aws (commented):
LGTM. A few minor comments.

kani-compiler/src/main.rs
celinval: Yes. To build proc_macro dependencies and build scripts. Cargo also invokes without ``--kani-compiler` to retrieve information about the compiler such as version.

kani-driver/src/call_single_file.rs
celinval: Sure

celinval:
I updated the comments. Let me know if that makes sense.

kani-driver/src/call_single_file.rs
```diff
@@ -174,9 +174,16 @@ impl KaniSession {
     }
 }
 
-/// This function can be used to convert a Kani compiler specific argument into a rustc one.
+/// This function can be used to convert Kani compiler specific arguments into a rustc one.
 /// We currently pass Kani specific arguments using the `--llvm-args` structure which is the
 /// hacky mechanism used by other rustc backend to receive arguments unknown to rustc.
+///
+/// Note that Cargo caching mechanism takes the building context into consideration, which
```
zhassan-aws, line 181: Thanks!

zhassan-aws (approved):

PR #1625, opened 2022-09-01
refactor(kani-driver, concrete_playback): process unit tests line by line, modularize functions, and add unit tests

sanjit-bhat:
### Description of changes: 

This PR does the following:
1. It refactors the kani-driver concrete playback code to generate/process unit tests line by line. As discussed with @celinval, this is a more intuitive/clean way to do things.
2. It moves the following functionality out of the `KaniSession` functions: 1) generating rustfmt args, 2) writing the new src code to a tmp file, 3) extracting the parent directory and source file from a file path.
3. It adds unit tests for kani-driver concrete playback functions that live outside `KaniSession`.

Note: I'm about to start PhD research next week, so trying to finish up some Kani PRs before then:)

### Call-outs

1. This should probably be reviewed by at least @celinval.
2. @adpaco-aws, you might want to look at the unit test for the concrete vals extractor. Maybe we can generalize that by making mock initializations for some of the CBMC output parser structs?

### Resolved issues:

Resolves #1502.

### Testing:

* How is this change tested?
Added kani-driver unit tests for 1) the formatting functions that generate the concrete playback unit test, 2) a file path helper function, 3) the concrete vals extractor.

* Is this a refactor change?
Yes

### Checklist
- [x] Each commit message has a non-empty body, explaining why the change was made
- [x] Methods or procedures are documented
- [x] Regression or unit tests are included, or existing tests cover the modified code
- [x] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #2188, opened 2023-02-06
Add unit tests for concrete playback, refactor util functions for easier testing

### LINKED ISSUE #1502: Add unit tests for `kani-driver` concrete playback components
See title. This needs to:
1. Separate `modify_src_code` into smaller functions.
2. Add unit tests for `modify_src_code` and `format_unit_test`.

jaisnan:
### Description of changes: 

This PR adds
1. Unit tests related to concrete playback outside of the existing regression tests.
2. Removes randomized layout and metadata as a parameter to generate the unit tests
3. Changes hash to use the `pretty_name` instead of the `mangled_name`.

### Resolved issues:

Partly resolves #1502 

### Call-outs:

1. Thanks to @sanjit-bhat for the work in PR #1625 
2. It changes the string that is to be hashed and added as a hash to the function. The naming schema hasn't yet changed to the user by this PR but will be in the coming few PR's.
4. The refactoring (removing dependency on metadata and randomized_layout) allows the functions to be tested and in the future, will allow the unit tests to be generated line by line.

### Testing:

* How is this change tested? Unit and regression tests

### Checklist


adpaco-aws (commented):
I've read the code but I'm still confused about the change in the function names of the tests. Are the generated unit tests runnable using these names without any modifications?

jaisnan:
> I've read the code but I'm still confused about the change in the function names of the tests. Are the generated unit tests runnable using these names without any modifications?

No, those generated unit tests are not runnable. I can make another revision fixing the names as well (will just need one more layer of formatting the names of the functions).

celinval:
> > I've read the code but I'm still confused about the change in the function names of the tests. Are the generated unit tests runnable using these names without any modifications?
> 
> No, those generated unit tests are not runnable. I can make another revision fixing the names as well (will just need one more layer of formatting the names of the functions).

I love the idea of adding the name of the original harness. However, I think this new name schema will break the `--inplace` UX though, wouldn't it?

jaisnan:
> > > I've read the code but I'm still confused about the change in the function names of the tests. Are the generated unit tests runnable using these names without any modifications?
> > 
> > 
> > No, those generated unit tests are not runnable. I can make another revision fixing the names as well (will just need one more layer of formatting the names of the functions).
> 
> I love the idea of adding the name of the original harness. However, I think this new name schema will break the `--inplace` UX though, wouldn't it?

Yes, I am trying to find a way to make the UX work, but if not, I'll just go back to the old naming scheme.

celinval:
Can you remove all the modules and only keep the function name?

celinval (approved):
That's awesome. Thanks @jaisnan

Once you merge this, can you please update our documentation as well to reflect the new convention please?

kani-driver/src/concrete_playback.rs
```diff
@@ -383,3 +322,198 @@ mod concrete_vals_extractor {
         None
     }
 }
+
+const SPACES_4: &str = "    ";
+const SPACES_8: &str = "        ";
+
+/// Format a unit test for a number of concrete values.
+fn format_unit_test(harness_name: &str, concrete_vals: &[ConcreteVal]) -> UnitTest {
```
zhassan-aws, line 330: Is there a reason for moving this function down here from its original position in the file? Changing its position is making it hard to see what changed in it?

zhassan-aws (commented):
Since the PR changes the name of concrete playback tests, the PR title should mention that.

kani-driver/src/concrete_playback.rs
jaisnan: No real reason, must have been shuffled around while editing. Should I make a revision placing it back into the original place?

zhassan-aws: Yes, I would prefer doing that to be able to easily spot what changed.

celinval (commented):
BTW, can you please either extend an existing test or add a new test for the new name schema? Thanks

kani-driver/src/concrete_playback.rs
```diff
@@ -383,3 +371,146 @@ mod concrete_vals_extractor {
...
+    struct SplitUnitTestName {
+        before_hash: String,
+        hash: String,
+    }
+
+    /// Unit test names are formatted as "kani_concrete_playback_{harness_name}_{hash}".
+    /// This function splits the name into "kani_concrete_playback_{harness_name}" and "{hash}".
+    fn split_unit_test_name(unit_test_name: &str) -> SplitUnitTestName {
+        let underscore_locs: Vec<_> = unit_test_name.match_indices('_').collect();
+        let last_underscore_idx = underscore_locs[underscore_locs.len() - 1].0;
```
zhassan-aws, line 416:
```suggestion
        let last_underscore_idx = underscore_locs.last().unwrap().0;
```

kani-driver/src/concrete_playback.rs
```diff
@@ -383,3 +371,146 @@ mod concrete_vals_extractor {
...
+        let hash_diff_harness_name = extract_hash_from_unit_test(harness_name_2, &concrete_vals_1);
+        let hash_diff_concrete_byte = extract_hash_from_unit_test(harness_name_1, &concrete_vals_2);
+        let hash_diff_interp_val = extract_hash_from_unit_test(harness_name_1, &concrete_vals_3);
+
+        assert_ne!(hash_base, hash_diff_harness_name);
+        assert_ne!(hash_base, hash_diff_concrete_byte);
+        assert_ne!(hash_base, hash_diff_interp_val);
+    }
+
+    #[test]
```
zhassan-aws, line 474: Can you add a comment that explains what this test is checking?

tests/ui/concrete-playback/use-function-name/main.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// kani-flags: --enable-unstable --concrete-playback=print
+
+// This test case checks that the new naming schema (i.e using the pretty name instead of mangled name)
```
zhassan-aws, line 6: As far as I've seen, this PR doesn't actually change to using the pretty name instead of the mangled name. This was previously the case. What changed though is that now the pretty name is hashed instead of the mangled name. Can you explain the reason for this change?

zhassan-aws (commented):
I'm a bit confused about the description in the PR:
> Functions weren't given their scope prefixes before which lead to multiple generated unit tests in the same file being called "kani_concrete_playback_harness", which is confusing to the user. This change renames them to use their scope as prefixes if available (using the "pretty name"). For example - "kani_concrete_playback_second::harness".

Can you clarify with an example?

tests/ui/concrete-playback/use-function-name/main.rs
jaisnan: It's a refactoring change. Passing the entire metadata to generate a function name seemed inefficient. This change will come in use for the upcoming PR which allows us to generate the unit test line by line as well.

celinval: Please update this comment to something like:
```rust
//! This test checks that the generated test includes the harness name.
```

I would avoid talking about "new" schema, because the comment may make sense as part of this PR, but it won't make sense in the future. If you want to mention the new schema, I would say something like:
```
This test checks the change to the name schema introduced in PR #2188. The concrete playback test case now includes the name of the harness.
```
I.e., explicitly mention the change this is related to.

zhassan-aws (approved):

tests/ui/concrete-playback/use-function-name/main.rs
jaisnan: Wanted to clarify that the final result of this PR does not bring any visible change to the name of the generated unit test (it did earlier, but i have removed that change since the generated unit tests weren't runnable), so i have removed the tests from this PR as well. It does change the way the hash suffix is generated, but not the overall naming scheme. I will submit the same tests at a later time, when I do add the new naming scheme. Hope that's a bit more clear.

PR #2194, opened 2023-02-07
Fix std overrides in crates where std is used as a feature

### LINKED ISSUE #2187: Issue finding `debug_assert` in `no_std` crates when std enabled
I tried this code:

```rust
#![no_std]

#[cfg(feature = "std")]
extern crate std;

fn foo() {
    std::debug_assert!(true, "text");
}
```

```
cargo kani --features std
```

with Kani version: 0.20.0

I expected to see this happen: verification should succeed

Instead, this happened: 

I get an error:
```
error: cannot find macro `__kani__workaround_core_assert` in this scope
 --> src/lib.rs:7:5
  |
7 |     std::debug_assert!(true, "text");
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `$crate::assert` which comes from the expansion of the macro `std::debug_assert` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `playground` due to previous error

Stack backtrace:
   4: kani_driver::main
   5: std::sys_common::backtrace::__rust_begin_short_backtrace
   6: std::rt::lang_start::{{closure}}
   7: std::rt::lang_start_internal
   8: main
   9: __libc_start_call_main
  10: __libc_start_main_impl
  11: _start
```

I initially ran into this while trying to compile a transitive dependency: `unicode-bidi`, which has a similar snippet (it's found at: https://github.com/servo/unicode-bidi/blob/master/src/implicit.rs#L494).

All of the following make the bug disappear:
 - removing `#![no_std]` 
 - removing `extern crate std` 
 - removing the `std::` from `std::debug_assert!(...)` 
 - removing the `"text"` from `std::debug_assert!(true, "text")`

All of the following have no effect:
 - adding `#[cfg(feature = "std")]` to the debug assert
 - adding `#[cfg(kani)]` to `fn foo()`
 - adding `#[kani::proof]` to `fn foo()`
 - changing the assert to `std::debug_assert!(false, "text");`

I've tried to reduce it further, but it's pretty bare at this point.

The key ingredients seem to be:
 - `extern crate std;` gated by a feature flag
 - `std::debug_assert!`
 - some actual text in the error message

I can't tell if this is caused by user error, or if there's a workaround I can use to verify my program even in the presence of this. I'm quite new to using kani, and am evaluating it for use at work, so apologies if there's some docs that explains this :sweat_smile: 

Thanks in advance :grin:

zhassan-aws:
### Description of changes: 

Kani's std overrides relied on creating an alias for `core::assert` up front to use for error checking in various macros (`assert`, `panic`, etc.). Creating the alias upfront doesn't work properly in crates with `no_std` where `std` is used as a feature (see #2187). This PR updates the std overrides to use `core::assert` directly when an `std` feature is enabled.

### Resolved issues:

Resolves #2187 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

To avoid code duplication, I tried creating a macro that calls `__kani__workaround_core_assert` under `not(std)` and `core::assert` otherwise. However, I ran into the same error reported in #2187.

### Testing:

### Checklist


tests/cargo-kani/no_std/src/main.rs
```diff
@@ -0,0 +1,19 @@
...
+#![no_std]
+
+#[cfg(feature = "std")]
+extern crate std;
+
+#[kani::proof]
+fn foo() {
+    let x: i32 = kani::any();
+    let y = 0;
+    std::debug_assert!(x + y == x, "Message");
```
danielsn, line 16: Should we also have a test where the assertion fails?

celinval (approved):
I'm worried that this is getting out of control. We should investigate a more robust solution.

Thanks for the quick fix though.

library/std/src/lib.rs
```diff
@@ -61,7 +61,10 @@ macro_rules! assert {
         // strategy, which is tracked in
         // https://github.com/model-checking/kani/issues/692
         if false {
+            #[cfg(not(feature = "std"))]
```
danielsn, line 64: Comment to the issue this is fixing, for future maintainers to know why we did this

tests/cargo-kani/no_std/src/main.rs
zhassan-aws: I don't think this is necessary, since we're mainly checking that compilation goes through. There are many other tests that check for failing assertions.

library/std/src/lib.rs
zhassan-aws: Done.

Issue #1562, opened 2022-08-20
Doc request: "How Kani works"


tedinski:
Customer feedback in response to a blog post:

https://old.reddit.com/r/rust/comments/wrd5bg/using_the_kani_rust_verifier_on_tokio_bytes/

> So it seems to use [cmbc](https://github.com/diffblue/cbmc) and a bunch of other tools from cprover under the hood (bundled in the github release and setup on first run...). I would really like to have this "how" more visible in the documentation, it's essential to hint at the limitations of such an automated prover, even if the underlying system is rather powerful.

We talk about this a lot in our talks about Kani, but not presently in our documentation. We should fix that. Part of this might also involve documenting a bit about how CBMC works as well.

* For example: https://www.youtube.com/watch?v=vE5Jwf_Lx94

If anyone has good ideas for "questions customers want answered about how Kani works" post em here. :)

PR #2195, opened 2023-02-07
Clarify the analysis technique used by Kani

tautschnig:
We use model checking - make clear what its limitations are when talking about what we may be able to prove.

### Description of changes: 

Clarification of what Kani is in our documentation.

### Resolved issues:

n/a

### Related RFC:

n/a

### Call-outs:

n/a

### Testing:

n/a

### Checklist
- n/a Regression or unit tests are included, or existing tests cover the modified code

### DIFF
--- a/docs/src/getting-started.md
+++ b/docs/src/getting-started.md
@@ -1,9 +1,11 @@
 # Getting started
 
-Kani is an open-source verification tool that uses automated reasoning to analyze Rust programs.
+Kani is an open-source verification tool that uses [model checking](./tool-comparison.md) to analyze Rust programs.
 Kani is particularly useful for verifying unsafe code in Rust, where many of the Rust’s usual guarantees are no longer checked by the compiler.
 Some example properties you can prove with Kani include memory safety properties (e.g., null pointer dereferences, use-after-free, etc.), the absence of certain runtime errors (i.e., index out of bounds, panics), and the absence of some types of unexpected behavior (e.g., arithmetic overflows).
 Kani can also prove custom properties provided in the form of user-specified assertions.
+As Kani uses model checking, Kani will either prove the property, disprove the
+property (with a counterexample), or may run out of resources.
 
 Kani uses proof harnesses to analyze programs.
 Proof harnesses are similar to test harnesses, especially property-based test harnesses.
@@ -15,7 +17,7 @@ Releases are published [here](https://github.com/model-checking/kani/releases).
 Major changes to Kani are documented in the [RFC Book](https://model-checking.github.io/kani/rfc).
 
 There is support for a fair amount of Rust language features, but not all (e.g., concurrency).
-Please see [Limitations - Rust feature support](./rust-feature-support.md) for a detailed list of supported features.
+Please see [Limitations](./limitations.md) for a detailed list of supported features.
 
 Kani releases every two weeks.
 As part of every release, Kani will synchronize with a recent nightly release of Rust, and so is generally up-to-date with the latest Rust language features.
--- a/docs/src/limitations.md
+++ b/docs/src/limitations.md
@@ -1,7 +1,8 @@
 # Limitations
 
 Like other tools, Kani comes with some limitations. In some cases, these
-limitations are inherent because of the techniques it's based on. In other
+limitations are inherent because of the techniques it's based on, or the
+undecidability of the properties that Kani seeks to prove. In other
 cases, it's just a matter of time and effort to remove these limitations (e.g.,
 specific unsupported Rust language features).
 


docs/src/getting-started.md
```diff
@@ -1,9 +1,12 @@
 # Getting started
 
-Kani is an open-source verification tool that uses automated reasoning to analyze Rust programs.
+Kani is an open-source verification tool that uses model checking to analyze Rust programs.
```
feliperodri, line 3: Suggestion: I think it would be nice to add a link (reference) here for model checking. Imagine a developer that has never really try or heard of it before. I don't have any great suggestions on what we could point here, but perhaps something even more general like this [article](https://www.amazon.science/publications/model-checking-as-a-human-endeavor).

feliperodri (approved):

docs/src/getting-started.md
zhassan-aws: +1

docs/src/getting-started.md
```diff
@@ -1,9 +1,12 @@
...
 Kani is particularly useful for verifying unsafe code in Rust, where many of the Rust’s usual guarantees are no longer checked by the compiler.
 Some example properties you can prove with Kani include memory safety properties (e.g., null pointer dereferences, use-after-free, etc.), the absence of certain runtime errors (i.e., index out of bounds, panics), and the absence of some types of unexpected behavior (e.g., arithmetic overflows).
 Kani can also prove custom properties provided in the form of user-specified assertions.
+Many of these properties are undecidable in the general case.
```
zhassan-aws, line 7: I feel this is a bit too heavy for the getting started section. I suggest we move it to the [the comparison with other tools section](https://model-checking.github.io/kani/tool-comparison.html) or paraphrase it a bit, e.g.:
```
Kani may not be able to solve all properties, in which case it may not terminate or may run out of memory.
Otherwise, Kani will either prove the property, or find a trace that leads to the violation of the property (e.g. the failure of an assertion).
```

celinval: +1 to moving this to a different section. I was also wondering if we should call out here that the proof is under some constraints and add a link to the limitations section.

docs/src/getting-started.md
```diff
@@ -1,9 +1,12 @@
...
+As Kani uses model checking, Kani will either prove the property, disprove the
+property (with a counterexample), or may run out of resources.
```
karkhaz, line 9:
```suggestion
property (providing you with a trace showing what input led to the property violation), or may run out of resources.
```
I wonder if "trace" is more familiar to the reader than "counterexample"?

docs/src/getting-started.md
celinval: I like the article idea, and that is a great article to mention here, but I would also suggest something more basic that explains what model checking is.

docs/src/getting-started.md
celinval: I don't mind counterexample, since we are talking about what not how. Trace is one way of representing a counterexample. We also offer the concrete playback in which case the counterexample will be represented as a unit test.

docs/src/getting-started.md
adpaco-aws: I couldn't find a good resource that is basic enough. I recall we've discussed having a section that explains the basics of model checking and other automated reasoning technologies in the past, but I couldn't find an issue related to it. Maybe https://github.com/model-checking/kani/issues/1562?

zhassan-aws: Why not the [wikipedia page](https://en.wikipedia.org/wiki/Model_checking)?

docs/src/getting-started.md
adpaco-aws: +1 to moving it.

This paragraph should focus on what Kani *can* do, as opposed to what it *cannot* do. All tools have their limitations and it's OK with me to explain it in other sections, but we should keep it brief in order to keep readers engaged.

docs/src/getting-started.md
jaisnan: We do have the [gentle introduction to automated reasoning](https://www.amazon.science/blog/a-gentle-introduction-to-automated-reasoning), which to me as a beginner, is easier to read and understand than the wikipedia page.

docs/src/getting-started.md
jaisnan: I think I like counterexample but not as a technical word, just as a phrase i.e `disprove the property (with the example that fails a proof)` or may run out of resources. Too many technical words in a blog causes reading fatigue imo.

docs/src/getting-started.md
jaisnan: I agree with @adpaco-aws too that we should add a page about the basics of model checking in the future.

tautschnig: I instead opted to link to our own `tool-comparison.md`.

docs/src/getting-started.md
tautschnig: I have removed the mention of "undecidable," and generalised the link to the Limitations section so as not to link into just a subsection.

zhassan-aws (approved):
A small typo. Otherwise looks good.

PR #2199, opened 2023-02-07
Create `Arbitrary::any_array()`

feliperodri:
We can take advantage of `any_raw_internal` to set all values in an array to nondet.
Kani users may now use `any_array()` with a loop-free operation for trivial types,
which improves performance.

Signed-off-by: Felipe R. Monteiro <felisous@amazon.com>

### Checklist

### DIFF
--- a/library/kani/src/arbitrary.rs
+++ b/library/kani/src/arbitrary.rs
@@ -3,6 +3,7 @@
 
 //! This module introduces the Arbitrary trait as well as implementation for primitive types and
 //! other std containers.
+
 use std::{
     marker::{PhantomData, PhantomPinned},
     num::*,
@@ -10,8 +11,19 @@ use std::{
 
 /// This trait should be used to generate symbolic variables that represent any valid value of
 /// its type.
-pub trait Arbitrary {
+pub trait Arbitrary
+where
+    Self: Sized,
+{
     fn any() -> Self;
+    fn any_array<const MAX_ARRAY_LENGTH: usize>() -> [Self; MAX_ARRAY_LENGTH]
+    // the requirement defined in the where clause must appear on the `impl`'s method `any_array`
+    // but also on the corresponding trait's method
+    where
+        [(); std::mem::size_of::<[Self; MAX_ARRAY_LENGTH]>()]:,
+    {
+        [(); MAX_ARRAY_LENGTH].map(|_| Self::any())
+    }
 }
 
 /// The given type can be represented by an unconstrained symbolic value of size_of::<T>.
@@ -20,8 +32,21 @@ macro_rules! trivial_arbitrary {
         impl Arbitrary for $type {
             #[inline(always)]
             fn any() -> Self {
-                // This size_of call does not use generic_const_exprs feature. It's inside a macro, and $type isn't generic.
-                unsafe { crate::any_raw_internal::<$type, { std::mem::size_of::<$type>() }>() }
+                // This size_of call does not use generic_const_exprs feature. It's inside a macro, and Self isn't generic.
+                unsafe { crate::any_raw_internal::<Self, { std::mem::size_of::<Self>() }>() }
+            }
+            fn any_array<const MAX_ARRAY_LENGTH: usize>() -> [Self; MAX_ARRAY_LENGTH]
+            where
+                // `generic_const_exprs` requires all potential errors to be reflected in the signature/header.
+                // We must repeat the expression in the header, to make sure that if the body can fail the header will also fail.
+                [(); { std::mem::size_of::<[$type; MAX_ARRAY_LENGTH]>() }]:,
+            {
+                unsafe {
+                    crate::any_raw_internal::<
+                        [Self; MAX_ARRAY_LENGTH],
+                        { std::mem::size_of::<[Self; MAX_ARRAY_LENGTH]>() },
+                    >()
+                }
             }
         }
     };
@@ -99,9 +124,10 @@ nonzero_arbitrary!(NonZeroIsize, isize);
 impl<T, const N: usize> Arbitrary for [T; N]
 where
     T: Arbitrary,
+    [(); std::mem::size_of::<[T; N]>()]:,
 {
     fn any() -> Self {
-        [(); N].map(|_| T::any())
+        T::any_array()
     }
 }
 
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -3,6 +3,9 @@
 
 // Used for rustc_diagnostic_item.
 #![feature(rustc_attrs)]
+// This is required for the optimized version of `any_array()`
+#![feature(generic_const_exprs)]
+#![allow(incomplete_features)]
 
 pub mod arbitrary;
 #[cfg(feature = "concrete_playback")]
--- a/tests/expected/object-bits/insufficient/main.rs
+++ b/tests/expected/object-bits/insufficient/main.rs
@@ -7,6 +7,9 @@
 
 #[kani::proof]
 fn main() {
-    let arr: [i32; 100] = kani::any();
-    assert_eq!(arr[0], arr[99]);
+    let mut arr: [i32; 100] = kani::Arbitrary::any_array();
+    for i in 0..30 {
+        arr[i] = kani::any();
+    }
+    assert!(arr[0] > arr[0] - arr[99]);
 }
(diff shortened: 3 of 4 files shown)


celinval:
I'm curious, how does that work?

library/kani/src/slice.rs
```diff
@@ -60,20 +60,9 @@ impl<T, const MAX_SLICE_LENGTH: usize> AnySlice<T, MAX_SLICE_LENGTH> {
...
-            //     for i in 0..slice_len {
-            //         *(ptr as *mut T).add(i) = any();
-            //     }
-            while i < any_slice.slice_len && i < MAX_SLICE_LENGTH {
-                *any_slice.ptr.add(i) = any();
-                i += 1;
-            }
+            let i = any();
+            assume(i < any_slice.slice_len);
+            *any_slice.ptr.add(i) = any();
```
zhassan-aws, line 65: Wouldn't that leave the entries uninitialized (i.e. poison) as opposed to initialized with a non-deterministic value?

feliperodri: Yes, but do we care about the entire content of the slice at his point or we want to make sure that if we use this content in an operation later, we will consider the case where `any()` part of the slice could have a non-deterministic value?

Instead of using a loop, we can create a non-deterministic index, make sure (with an assumption) that the index is within bounds, and then non-deterministically select a section of slice and assign it to `any()`. During symbolic execution, whenever we access this slice, it will consider the path where the section is assign to `any()`. Thus, we are basically taking advantage of the symbolism in symbolic execution instead of unrolling the loop during GOTO transformation.

For instance,
```rust
//consider p = poison value;
slice := [p, p, …, p, …, p]

i = any()
assume( i < slice.len())

slice[i] = any()
slice:= [p,p, …, p, any(), p, …, p, p] // any() could be placed in any() location up to slice.len()

// now let's assert something about slice
j = any();
assert(slice[j] == p); // any assertion on the content of slice will fail, since any location could be equal to any()
```

feliperodri: @celinval does that clarify your question too? ☝️

zhassan-aws: @feliperodri Reading uninitialized memory in Rust is undefined behavior: https://doc.rust-lang.org/reference/behavior-considered-undefined.html

The compiler may do anything with those values unless we initialize them.

danielsn: I suspect the real solution is to move `any_slice` from a rust implementation to a kani intrinsic, like we do for `any`.

feliperodri: Oh! So we should not implement this here in Rust, maybe at the GOTO level only. Is there a way where I only include the body of this function later during codegen?

zhassan-aws: We use hooks for implementing functions in codegen. For example, `kani::any_raw_inner` is handled here: https://github.com/model-checking/kani/blob/51b758b246198b47dcb3ed3c485aa47ec9398ba1/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs#L168

Out of curiosity, have you run into a case where `any_slice` in its current form is causing a performance issue?

zhassan-aws:
Can you clarify what the new version of the PR is doing? Is it providing a new way to create non-det arrays:
```rust
let arr: [u8; 10] = kani::Arbitrary::any_array();
```
? If so, how does it compare to the existing one:
```rust
let arr: [u8; 10] = kani::any();
```
and are there any advantages to the new approach?

tests/expected/object-bits/insufficient/main.rs
```diff
@@ -7,6 +7,9 @@
 
 #[kani::proof]
 fn main() {
-    let arr: [i32; 100] = kani::any();
```
celinval, line 10: Why did you have to change this test?

celinval (commented):
Weren't you planning to also make changes to any_slice?

celinval:
> Can you clarify what the new version of the PR is doing? Is it providing a new way to create non-det arrays:
> 
> ```rust
> let arr: [u8; 10] = kani::Arbitrary::any_array();
> ```
> 
> ? If so, how does it compare to the existing one:
> 
> ```rust
> let arr: [u8; 10] = kani::any();
> ```
> 
> and are there any advantages to the new approach?

With the new approach, we can directly assign a nondet array without initializing each element individually for types where this operation is safe.

zhassan-aws:
> With the new approach, we can directly assign a nondet array without initializing each element individually for types where this operation is safe.

Can you clarify how this is achieved? The `any_array` implementation:
```rust
[(); MAX_ARRAY_LENGTH].map(|_| Self::any())
```
seems to be doing the same thing that the `Arbitrary` implementation for arrays was doing:
```rust
[(); N].map(|_| T::any())
```

feliperodri:
> Weren't you planning to also make changes to any_slice?

@celinval no changes to `any_slice()` on this PR. I want to focus this one only on implementing a version of `any_array()`. We can address any changes to `any_slice()` on a separate PR.

library/kani/src/arbitrary.rs
```diff
@@ -3,15 +3,25 @@
...
 /// its type.
-pub trait Arbitrary {
+pub trait Arbitrary
+where
+    Self: Sized,
+{
     fn any() -> Self;
+    fn any_array<const MAX_ARRAY_LENGTH: usize>() -> [Self; MAX_ARRAY_LENGTH]
+    where
+        [(); std::mem::size_of::<[Self; MAX_ARRAY_LENGTH]>()]:,
```
zhassan-aws, line 21: Can you add a comment on what this `where` constraint is doing?

celinval:
> > With the new approach, we can directly assign a nondet array without initializing each element individually for types where this operation is safe.
> 
> Can you clarify how this is achieved? The `any_array` implementation:
> 
> ```rust
> [(); MAX_ARRAY_LENGTH].map(|_| Self::any())
> ```
> 
> seems to be doing the same thing that the `Arbitrary` implementation for arrays was doing:
> 
> ```rust
> [(); N].map(|_| T::any())
> ```

The difference is in the implementation of `arbitrary_trivial!()` macro, which is how we define the implementation for the trivial types which have no contraint in their values:

```rust
            fn any_array<const MAX_ARRAY_LENGTH: usize>() -> [Self; MAX_ARRAY_LENGTH]
            where
                [(); { std::mem::size_of::<[$type; MAX_ARRAY_LENGTH]>() }]:,
            {
                unsafe {
                    crate::any_raw_internal::<
                        [Self; MAX_ARRAY_LENGTH],
                        { std::mem::size_of::<[Self; MAX_ARRAY_LENGTH]>() },
                    >()
                }
```
Note that we directly call `any_raw_internal` with the array type.

zhassan-aws (commented):
I see. Can you point out how `any_slice` will use it?

celinval (approved):
It looks good to me. Please address @zhassan-aws comments before merging.

library/kani/src/lib.rs
```diff
@@ -3,6 +3,9 @@
 
 // Used for rustc_diagnostic_item.
 #![feature(rustc_attrs)]
+// This will enable us to implement an optmized version of anyArray()
```
zhassan-aws, line 6:
```suggestion
// This is required for the optimized version of `any_array()`
```

zhassan-aws (approved):
Thanks. Please update the PR description as well.

library/kani/src/arbitrary.rs
feliperodri: I added the following comment
```rust
    // the requirement defined in the where clause must appear on the `impl`'s method `any_array`
    // but also on the corresponding trait's method
```

I also added a comment at the `impl`.
```rust
    // generic_const_exprs requires all potential errors to be reflected in the signature/header.
    // We must repeat the expression in the header, to make sure that if the body can fail the header will also fail.
```

library/kani/src/lib.rs
feliperodri: Implemented as suggested.

zhassan-aws (approved):
Looks good. Thanks @feliperodri!

zhassan-aws:
It would be useful to add a test (perhaps in a follow-up PR) that makes sure that a program involving an array of a primitive type does not involve loops. This can be verified by adding a `#[unwind(0)]` annotation so that the test would fail if a loop is generated.

PR #2202, opened 2023-02-09
Allow users to select a subset of harness to run

### LINKED ISSUE #1778: Kani should support multiple harnesses as arguments
Requested feature: Users should be able to select a subset of harnesses to run either via wildcard or multiple usage of `--harness`.
Use case: A user may want to run all the harnesses from one module or that targets a specific function instead of running all harnesses in the crate.

My suggestion is that:
1. We add a new argument: `--filter [FILTER_STR]` that allows users to use wildcard syntax. 
2. Multiple usages of `--filter` should implement an AND logic, i.e., a harness has to match all filters to be included in the run.
3. Using both `--filter` and `--harness` should not be allowed.
4. Users should be able to use `--harness` multiple times to select more than one specific harness to run.
5. The change 4 should also be applied to `--function`.

celinval:
### Description of changes: 

Users can now select a subset of harnesses to run using `--harness` multiple times. Previously users could either run a single harness or they had to run them all.

For the example provided in #1778:
```rust
// example.rs
#[kani::proof]
fn a() {}
#[kani::proof] 
fn b() {}
#[kani::proof] 
fn c() {std::unimplemented!();}
```

Users can select harnesses `a` and `b` by running:
```bash
$ kani example.rs --harness a --harness b
```

In case of multiple matches, Kani will run all harnesses that matches at least one of the `--harness` argument.

### Resolved issues:

Resolves #1778 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

### Checklist


jaisnan:
Awesome change! this will come in really handy for the kani extension. Could you add an example invocation of the change to the description as well?

jaisnan:
I have a couple of questions - 

1. Does this change anything about `cargo kani`? It doesn't seem that way. Afaik `cargo kani ` always had the option to run multiple harnesses using `cargo kani --harness ['a', 'b']` or is that not the case?
2. This doesn't add the --filter option yet, is that still coming in a future PR? That is something that could be very useful for the kani extension as well.

celinval:
> I have a couple of non-blocking questions -
> 
>     1. Does this change anything about `cargo kani`? It doesn't seem that way. Afaik `cargo kani` has the option to run multiple harnesses using `cargo kani --harness ['a', 'b']` or is that not the case? (and if not, Why add this change to single file `kani` as opposed to `cargo kani`?)
> 
>     2. This doesn't add the `--filter [FILTER_STR]` option yet, is that still coming in a future PR? That is something that could be very useful for the kani extension as well.

Great questions! I need to check how this argument configuration will work for configuration via `Cargo.toml`.

1. All the changes here should apply to `cargo kani` as well. I don't believe `cargo kani` allows you to specify multiple harnesses today.
2. Since we changed to match against substring, I don't think we need the `--filter` option. What do you think?

jaisnan:
> > I have a couple of non-blocking questions -
> > ```
> > 1. Does this change anything about `cargo kani`? It doesn't seem that way. Afaik `cargo kani` has the option to run multiple harnesses using `cargo kani --harness ['a', 'b']` or is that not the case? (and if not, Why add this change to single file `kani` as opposed to `cargo kani`?)
> > 
> > 2. This doesn't add the `--filter [FILTER_STR]` option yet, is that still coming in a future PR? That is something that could be very useful for the kani extension as well.
> > ```
> 
> Great questions! I need to check how this argument configuration will work for configuration via `Cargo.toml`.
> 
>     1. All the changes here should apply to `cargo kani` as well. I don't believe `cargo kani` allows you to specify multiple harnesses today.
> 
>     2. Since we changed to match against substring, I don't think we need the `--filter` option. What do you think?

Substrings work great as well. Would it be possible to change the invokation to use the form `--harness ['a' 'b']` format instead of `--harness 'a' --harness 'b'`? Seems more ergonomic but just something to maybe add in the future, not too important for now.

adpaco-aws:
> Would it be possible to change the invokation to use the form `--harness ['a' 'b']` format instead of `--harness 'a' --harness 'b'`? Seems more ergonomic but just something to maybe add in the future, not too important for now.

I'd recommend against this.

Allowing multiple `--harness` arguments gives more flexibility to the user. The other problem is that characters `'` and `[` often need to be escaped depending on the context (the current proposal doesn't use `'` as far as I know).

celinval:
@jaisnan I totally understand your point, and I totally agree that it is painful to pass multiple `--harness` arguments, but I think this is more robust, as @adpaco-aws mentioned. It is also not the recommended to accept multiple values if you have positional arguments (see [this doc on why](https://docs.rs/structopt/0.2.14/structopt/clap/struct.Arg.html#method.multiple)). To go around this, we could consider doing one of the following:

- Create a short version of `--harness`, I don't know which letter to use though, since `-h` is help. 🙂
- We could consider using `,` delimiter to allow users to provide multiple arguments to `--harness`. So users could use `--harness first,second`.

jaisnan:
I see, makes sense to me. Yeah, the second idea would be ergonomic as well, but honestly what we have works too.

kani-driver/src/harness_runner.rs
```diff
@@ -139,11 +139,31 @@ impl KaniSession {
...
+                    }
+                    ([harness], None) => {
+                        println!("No harnesses matched the harness filter: `{harness}`",)
+                    }
+                    (harnesses, None) => {
+                        println!(
+                            "No harnesses matched the harness filters: `{}`",
+                            harnesses.join("`, `")
+                        )
+                    }
```
adpaco-aws, line 159: I'm concerned about this scenario:
 * User has Kani proofs running in CI
 * User renames proofs or embed them in a non-visible module for Kani.
 * Kani running in CI doesn't find proofs, or the filters cause it to not run them.
In this scenario, Kani would silently continue to run no proofs until someone takes a look at the CI logs.

Of course, the same scenario can happen while still running some proofs. In that case, there isn't much we can do about. But we can prevent a scenario like the one I described earlier by returning an error here, and an exit code different to 0 in the caller of this function (not sure how this structured, just giving ideas).

I'll admit that this is a very conservative approach, but I think users would appreciate the effort to prevent this problem.

kani-driver/src/metadata.rs
```diff
@@ -158,40 +165,22 @@ pub fn mock_proof_harness(
...
-            } else {
-                result = Some(h);
-            }
+fn find_proof_harnesses<'a>(
+    targets: BTreeSet<&String>,
+    all_harnesses: &[&'a HarnessMetadata],
+) -> Vec<&'a HarnessMetadata> {
+    debug!(?targets, "find_proof_harness");
+    let mut result = vec![];
+    for md in all_harnesses.iter() {
```
adpaco-aws, line 174: What is `md`?

kani-driver/src/metadata.rs
```diff
@@ -206,13 +195,24 @@ mod tests {
             mock_proof_harness("module::not_check_three", None, None),
         ];
         let ref_harnesses = harnesses.iter().collect::<Vec<_>>();
-        assert!(find_proof_harness("check_three", &ref_harnesses).is_err());
+        assert_eq!(
+            find_proof_harnesses(BTreeSet::from([&"check_three".to_string()]), &ref_harnesses)
```
adpaco-aws, line 199: Curious: Are the harness names normalized before we try to find them?

kani-driver/src/args_toml.rs
```diff
@@ -106,7 +106,8 @@ fn toml_to_args(tomldata: &str) -> Result<(Vec<OsString>, Vec<OsString>)> {
     for (flag, value) in map {
         if flag == "cbmc-args" {
             // --cbmc-args has to come last because it eats all remaining arguments
-            insert_arg_from_toml(&flag, &value, &mut cbmc_args)?;
+            cbmc_args.push("--cbmc-args".into());
+            cbmc_args.append(&mut cbmc_arg_from_toml(&value)?);
         } else {
             insert_arg_from_toml(&flag, &value, &mut args)?;
```
adpaco-aws, line 112: Why is the new function needed? Why aren't you using it in the `else` branch?

adpaco-aws (commented):
How does this impact the features that are supposed to only work for a single harness (e.g., stubbing)? I couldn't find the code that deals with that case, apologies if I missed it.

kani-driver/src/harness_runner.rs
adpaco-aws: BTW I know that we had https://github.com/model-checking/kani/issues/1857 in the past, but I'm just wondering if we should reconsider it.

kani-driver/src/metadata.rs
celinval: metadata... I can expand. The 'h' was annoying me, but I guess `md` is not much better. :)

kani-driver/src/metadata.rs
celinval: The harnesses `pretty_name` are fully qualified. Does that answer you question?

kani-driver/src/args_toml.rs
celinval: That's a great question. The problem here is that `--cbmc-args` will be the only case where we won't replicate the flag name when flattening an array. Here is an example:

```toml
[package.metadata.kani.flags]
package = ["p1", "p2"]
harness = ["foo", "bar"]
cbmc_args = ["--outfile", "filename"]
```

We will convert this into these args:
```rust
["--package", "p1", "--package", "p2", "--harness", "foo", "--harness", "bar", "--cbmc-args", "--outfile", "filename"]
```

Does that make sense?

kani-driver/src/harness_runner.rs
celinval: I totally understand the concern. There is a question here about what the level of flexibility should be.

I think your concern is legit, and it is something that I noticed happens sometimes when we add tests in our regression and we forget to add the `expected` files. We think something is being tested, but it's not.

I'm not sure that throwing an error when the filter doesn't apply to any harness really solves the root problem though. Let's say the user filter use to match one harness and it doesn't anymore because that harness got renamed. If that filter also matches another harness, Kani will run without throwing an error. So in this case, we are now missing a harness that we didn't before.

celinval:
> How does this impact the features that are supposed to only work for a single harness (e.g., stubbing)? I couldn't find the code that deals with that case, apologies if I missed it.

It's there, but it's not pretty. I'm currently working on removing that restriction, so I didn't want to spend too much time with this.

Let me improve this to at least throw an error if more than one filter is given and if more than one harness matches the filter. Is that OK?

Also, is there any other feature besides stubbing that relies on a single harness? I remember that we used to have that restriction for concrete playback, but that's no longer the case.

kani-driver/src/harness_runner.rs
celinval: Maybe we can report the harnesses that we skipped. On verbose mode, we could even print them. What do you think?

celinval: I noticed that `cargo tests` include the number of tests filtered out:
```
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 38 filtered out; finished in 0.00s
```

zhassan-aws: Including the number of filtered-out harnesses sounds like a reasonable solution.

kani-driver/src/args_toml.rs
adpaco-aws: Yes, thank you!

kani-driver/src/harness_runner.rs
adpaco-aws: Yes, as said earlier: there isn't a lot we can do when other harnesses match the filter. But the case where the user runs `--harness` and filters out all proof harnesses should be signaled somehow.

kani-driver/src/metadata.rs
adpaco-aws: Yes, thanks!

adpaco-aws:
> It's there, but it's not pretty. I'm currently working on removing that restriction, so I didn't want to spend too much time with this.
> 
> Let me improve this to at least throw an error if more than one filter is given and if more than one harness matches the filter. Is that OK?

Sounds good. I don't think you need to check both conditions though, just the latter.

> Also, is there any other feature besides stubbing that relies on a single harness? I remember that we used to have that restriction for concrete playback, but that's no longer the case.

I wasn't sure about concrete playback, but that's the only two that I had mind. Sorry about that.

kani-driver/src/harness_runner.rs
celinval: Fair... do you want it to be an error or a warning?

adpaco-aws: An error.

kani-compiler/src/kani_compiler.rs
```diff
@@ -101,9 +102,18 @@ impl KaniCompiler {
...
             FxHashMap::default()
-        } else if let Some(harness) = self.args.as_ref().unwrap().get_one::<String>(parser::HARNESS)
+        } else if let Some(harnesses) =
+            self.args.as_ref().unwrap().get_many::<String>(parser::HARNESS)
         {
-            find_harness_stub_mapping(harness, all_stubs).unwrap_or_default()
+            let mappings = filter_stub_mapping(harnesses.collect(), all_stubs);
+            if mappings.len() > 1 {
+                tcx.sess.err(format!(
+                    "Failed to apply stubs. Harnesses with stubs must be verified separately. Found: `{}`",
```
zhassan-aws, line 111:
```suggestion
                    "Failed to apply stubs. Harnesses with stubs must be verified separately. The `--harness` option matched multiple harnesses: `{}`",
```
or anything shorter you can think of.

kani-driver/src/args.rs
```diff
@@ -642,6 +648,25 @@ mod tests {
         .unwrap();
         // no assertion: the above might fail if it fails to allow 0 args to cbmc-args
     }
+
+    #[test]
+    fn check_multiple_harnesses() {
+        // accepts repeated:
```
adpaco-aws, line 654: Nit: Add a comment for the function instead.

adpaco-aws (approved):

kani-driver/src/main.rs
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 #![feature(let_chains)]
 #![feature(array_methods)]
-
+#![feature(map_try_insert)]
```
zhassan-aws, line 5: Is this needed? Didn't see `try_insert` used in the PR.

zhassan-aws (approved):
Thanks for implementing this @celinval! A couple of minor comments.

kani-compiler/src/kani_compiler.rs
celinval: Yeah, I was trying to keep it short. Also, I am planning to get rid of this restriction sooner rather than later. Would you mind if I just leave it as is for this PR?

kani-driver/src/main.rs
celinval: oh... yeah, that was from a previous iteration.

kani-compiler/src/kani_compiler.rs
zhassan-aws: Sure.

PR #2204, opened 2023-02-09
Integrate goto-synthesizer into Kani

### LINKED ISSUE #2156: Intergrate ```goto-synthesizer``` into Kani
The support for Kani-compiled GOTOs is implemented in https://github.com/diffblue/cbmc/pull/7487 and https://github.com/diffblue/cbmc/pull/7488. Once both of them have been merged. We will integrate ```goto-synthesizer``` into Kani so that users can choose to synthesize loop invariants for loop with a new flag.

qinheping:
### Description of changes: 

This PR integrate ```goto-synthesizer``` into Kani, and add an option to synthesize loop contracts for all loops.

### Resolved issues:

Resolves #2156 

### Call-outs:

The purpose of this PR is to allow us to experiment the loop-contracts synthesizer on Kani benchmarks. However, there are still a few issues we need to address with later PRs:

- The synthesizer may not terminate. When there is no loop invariants in the candidate space, the enumerator in the synthesizer will run forever trying to find a solution. To address this problem, we might want to allow users to limit the resource used by the synthesizer. On the ```goto-synthesizer``` side, we will add an option to limit the size of candidates we enumerate, so that it will not keep enumerating when there is no solution in the candidate space. On the Kani side, we can also add an option to limit the running time of the synthesizer.
- The synthesizer is not compatible with unwinding numbers. Now the synthesizer ignores unwinding numbers, and will synthesize loop-contracts for all loops no mater if there are loop numbers specified for them. However, we might sometimes want to unwind some loops while synthesize loop contracts for other loops. We will allow users to specify unwinding numbers in ```goto-synthesizer```. Then users can choose to synthesize loop contracts only for those loops without unwinding number specified.

### Testing:

* How is this change tested?

This PR include a regression test.

* Is this a refactor change?

No.

### Checklist


kani-driver/src/harness_runner.rs
```diff
@@ -64,6 +64,12 @@ impl<'sess> HarnessRunner<'sess> {
                         &harness,
                     )?;
 
+                    if self.sess.args.synthesize_loop_contracts {
+                        println!("Start to synthesize loop contracts for all loops.");
+                        println!("WARNING: this process may not terminate.");
```
celinval, line 69: How come?

kani-driver/src/harness_runner.rs
```diff
@@ -64,6 +64,12 @@ impl<'sess> HarnessRunner<'sess> {
                         &harness,
                     )?;
 
+                    if self.sess.args.synthesize_loop_contracts {
+                        println!("Start to synthesize loop contracts for all loops.");
```
celinval, line 68: Please check if we are on quiet mode before printing this message.

celinval (commented):
That's awesome! Thanks for adding this. I have a few questions:
1. How does the loop synthesizer interacts with loop unwind argument?
2. Is there anyway we can avoid the non-termination? Even if that means failing the synthesis?
3. We need to add some documentation around this. I would suggest writing an RFC explaining the user experience, the value it adds and and its limitations.

qinheping:
> 

Yes, in the Call-outs part of the PR description I introduced my plan to handle non-termination and how to deal with unwinding numbers. More comments about them are appropriated. 

And yes, I will create a RFC about it.

celinval:
I think an RFC will help. From what you described, there is a conflict between unwind and loop synthesis, at least for now. In this case, we should figure out what the message should be and how we handle a harness that is annotated with `unwind`. There's also a question whether this should be applied to the entire system or to specific harnesses.

zhassan-aws:
Is it possible to add tests?

qinheping:
> Is it possible to add tests?

I added two simple tests to check the synthesizer is correctly called and enumerates required loop invariants. More experiments are needed to see how the synthesizer works on more complicated tests.

kani-driver/src/harness_runner.rs
qinheping: Add

kani-driver/src/call_goto_synthesizer.rs
```diff
@@ -0,0 +1,31 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
```
celinval, line 2: please add a little doc to this file

kani-driver/src/call_goto_synthesizer.rs
```diff
@@ -0,0 +1,31 @@
...
+            output.to_owned().into_os_string(), // input
+            output.to_owned().into_os_string(), // output
+        ];
+
+        self.call_goto_synthesizer(args)?;
+
+        Ok(())
+    }
+
+    /// Non-public helper function to actually do the run of goto-synthesizer
```
celinval, line 24: Is there a reason why you have this?

kani-driver/src/call_goto_synthesizer.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use anyhow::Result;
+use std::ffi::OsString;
+use std::path::Path;
+use std::process::Command;
+
+use crate::session::KaniSession;
+
+impl KaniSession {
+    pub fn synthesize_loop_contracts(&self, output: &Path) -> Result<()> {
```
celinval, line 12: doc please

kani-driver/src/harness_runner.rs
```diff
@@ -64,6 +64,14 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
                         &harness,
                     )?;
 
+                    if self.sess.args.synthesize_loop_contracts {
+                        if !self.sess.args.quiet {
```
celinval, line 68: please move this code to be inside the synthesize_loop_contracts function.

kani-driver/src/harness_runner.rs
```diff
@@ -64,6 +64,14 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
...
+                            println!("Start to synthesize loop contracts for all loops.");
+                            println!("WARNING: this process may not terminate.");
```
celinval, line 70: please use `util::warn()`

tests/kani/LoopContractsSynthesizer/main_signed.rs
```diff
@@ -0,0 +1,17 @@
...
+
+// Check if goto-synthesizer is correctly called, and synthesizes the required
+// loop invariants.
+
+#[kani::proof]
+fn main() {
+    let mut y: i32 = kani::any_where(|i| *i > 0, "y is positive");
+
+    while y > 0 {
+        y = y - 1;
```
celinval, line 14: can you please move these tests to the `ui` suite and ensure that the last assert is reachable.

kani-driver/src/args.rs
```diff
@@ -224,6 +224,10 @@ pub struct KaniArgs {
     #[arg(long, hide_short_help = true, requires("enable_unstable"))]
     pub no_slice_formula: bool,
 
+    /// Synthesize loop contracts for all loops.
+    #[arg(long, hide_short_help = true, requires("enable_unstable"))]
```
celinval, line 228: you should add that this conflicts with `--unwind`

qinheping:
> 

Thank you Celina! Yes, I plan to add more tests. I am doing the experiment of running the synthesizer on all current Kani benchmarks, which will provide us better insight about what we can and cannot synthesize. I will create an issue and add tests when the experiment is done.

tests/kani/LoopContractsSynthesizer/main_signed.rs
qinheping: I moved them to the ```ui``` suite. I tried to use ```kani::cover``` instead of ```assert!```, but the synthesizer fail to solve the problem. I will create an issue about it and fix this issue.

kani-driver/src/call_goto_synthesizer.rs
```diff
@@ -0,0 +1,36 @@
...
+use std::path::Path;
+use std::process::Command;
+
+use crate::session::KaniSession;
+
+impl KaniSession {
+    /// Synthesize loop contracts for a goto binary `input` and produce a new goto binary `output`
+    /// The synthesizer we use is `goto-synthesizer` built in CBMC codebase, which is an enumerative
+    /// loop-contracts synthesizer. `goto-synthesizer` enumerates and checks if a candidate can be
+    /// used to prove some assertions, and return found invariants when all checks pass.
```
zhassan-aws, line 16:
```suggestion
    /// used to prove some assertions, and applies found invariants when all checks pass.
```
?

kani-driver/src/call_goto_synthesizer.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub fn synthesize_loop_contracts(&self, input: &Path, output: &Path) -> Result<()> {
+        if !self.args.quiet {
+            warning("Start to synthesize loop contracts for all loops.");
```
zhassan-aws, line 19: I would say "Running loop contract synthesizer".

zhassan-aws, line 19: Does this need to be a warning?

kani-driver/src/call_goto_synthesizer.rs
```diff
@@ -0,0 +1,36 @@
...
+        let args: Vec<OsString> = vec![
+            "--loop-contracts-no-unwind".into(),
+            input.to_owned().into_os_string(),  // input
+            output.to_owned().into_os_string(), // output
+        ];
+
+        let mut cmd = Command::new("goto-synthesizer");
+        cmd.args(args);
+
+        self.run_suppress(cmd)?;
```
zhassan-aws, line 32: Does the `goto-synthesizer` produce any output while it's running? If so, `run_terminal` might be more suitable.

tests/ui/LoopContractsSynthesizer/main_signed/expected
```diff
@@ -0,0 +1 @@
+VERIFICATION:- SUCCESSFUL
```
zhassan-aws, line 1: The test file itself is missing from the PR.

tests/ui/LoopContractsSynthesizer/main_unsigned/expected
```diff
@@ -0,0 +1 @@
+VERIFICATION:- SUCCESSFUL
```
zhassan-aws, line 1: Same for this one.

zhassan-aws (commented):
Thanks @qinheping!

kani-driver/src/call_goto_synthesizer.rs
qinheping: Not actually, I will use ```println!``` instead.

kani-driver/src/call_goto_synthesizer.rs
qinheping: It only output debug messages when a high ```verbosity``` is specified.

zhassan-aws: I see. In this case, I suggest doing:
```rust
       if self.args.quiet {
            self.run_suppress(cmd)?;
        } else {
            self.run_terminal(cmd)?;
        }
```
so that we can see its output if needed.

qinheping: Sounds great!

zhassan-aws (approved):
Thanks @qinheping! Please address any remaining comments from @celinval and wait for her approval before merging.

kani-driver/src/call_goto_synthesizer.rs
```diff
@@ -0,0 +1,40 @@
...
+        ];
+
+        let mut cmd = Command::new("goto-synthesizer");
+        cmd.args(args);
+
+        if self.args.quiet {
+            self.run_suppress(cmd)?;
+        } else {
+            self.run_terminal(cmd)?;
+        }
```
celinval, line 36: You don't need this. In fact, `run_suppress` invokes `run_terminal` if `quiet` is true.

kani-driver/src/call_goto_synthesizer.rs
```diff
@@ -0,0 +1,40 @@
...
+
+impl KaniSession {
+    /// Synthesize loop contracts for a goto binary `input` and produce a new goto binary `output`
+    /// The synthesizer we use is `goto-synthesizer` built in CBMC codebase, which is an enumerative
+    /// loop-contracts synthesizer. `goto-synthesizer` enumerates and checks if a candidate can be
+    /// used to prove some assertions, and applies found invariants when all checks pass.
+    pub fn synthesize_loop_contracts(&self, input: &Path, output: &Path) -> Result<()> {
+        if !self.args.quiet {
+            println!("Running loop contract synthesizer.");
+            warning("WARNING: this process may not terminate.");
```
celinval, line 20: the `warning` function already adds the `warning` prefix.

kani-driver/src/args.rs
```diff
@@ -565,6 +569,12 @@ impl KaniArgs {
             }
         }
 
+        if (natives_unwind || extra_unwind) && self.synthesize_loop_contracts && !self.quiet {
```
celinval, line 572: I'm OK with the warning if the user has sets `default-unwind` but everything else I would highly recommend an error. I prefer if we don't have an implicit precedence for our arguments. If they are incompatible, just throw an error.

kani-driver/src/call_goto_synthesizer.rs
zhassan-aws: My bad (this was my suggestion) :(

zhassan-aws: We should clean up this instance then: https://github.com/model-checking/kani/blob/035225ca362b0a80c807af0f567c3513237cb0ee/kani-driver/src/concrete_playback.rs#L191

PR #2087, opened 2023-01-06
Include Kissat in the Kani bundle

zhassan-aws:
### Description of changes: 

Install Kissat as part of the Kani setup, and include it in the Kani release bundle.

This is required by a PR that adds a `kani::solver` attribute (whic will follow shortly).

**Keeping this as a draft until we integrate a CBMC version that fixes #1962.**

### Resolved issues:

Towards #1656 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? A test is included in a PR that will follow shortly.

* Is this a refactor change? No

### Checklist
- [X] Each commit message has a non-empty body, explaining why the change was made
- [X] Methods or procedures are documented
- [X] Regression or unit tests are included, or existing tests cover the modified code
- [X] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #2207, opened 2023-02-10
fix: add kissat to the nixos patchelf list

camshaft:
### Description of changes: 

In #2087, kissat was added to the kani-verifier distribution bundle. However, the binary wasn't included in the NixOS list of binaries to patch with `patchelf` and results in it not being executable on that platform. This change simply adds it to the list.

### Call-outs:

I wonder if we need a better mechanism of keeping the list of binaries we ship in sync?

### Testing:

I tested it locally by running the `patchelf` command manually. After running it I was able to use the solver as expected.

### Checklist


zhassan-aws (approved):
Thanks!

adpaco-aws (approved):

PR #2211, opened 2023-02-15
Move attribute handling to `kani_middle` and polish warnings

celinval:
### Description of changes: 

As part of removing one harness restriction from stubbing, we will need to run the rust compiler driver multiple times. However, we want to only produce one metadata file with information about all the compiled artifacts.

Thus, I'm planning to move the harness metadata generation to the KaniCompiler as opposed to the GotoCCtxt. As a first step, I'm moving the attributes handling out of the GotoCCtxt.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

The only behavior change in this PR is related to some of the warnings we print. I tried to standardize it a bit (following our new guidelines) and I moved the stub warning to be part of the driver.

### Testing:

### Checklist


kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -345,64 +254,27 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                _ => {
-                    self.tcx.sess.span_err(
-                        attr.1.span,
-                        format!("Unsupported Annotation -> {}", attr.0.as_str()).as_str(),
-                    );
-                }
-            }
+        let attributes = extract_harness_attributes(self.tcx, def_id);
+        if let Some(attrs) = attributes {
+            let mut harness = self.default_kanitool_proof();
```
zhassan-aws, line 270: Nit: It would be slightly cleaner to add another method that takes attributes as a parameter and use it to set the attributes field.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -2,33 +2,278 @@
...
 
-    for attr in all_attributes {
+use rustc_middle::ty::layout::FnAbiOf;
+use tracing::{debug, trace};
+
+use crate::kani_middle::CompilerHelpers;
+
+#[derive(Debug, Clone, Copy, AsRefStr, EnumString, PartialEq, Eq, PartialOrd, Ord)]
+#[strum(serialize_all = "snake_case")]
+enum KaniAttributeKind {
```
zhassan-aws, line 22: Nice!

zhassan-aws (approved):
Thanks @celinval. Only one nit. Otherwise looks good.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -2,33 +2,278 @@
...
+
+    if tcx.def_kind(def_id) != DefKind::Fn {
+        tcx.sess.span_err(span, "the `proof` attribute can only be applied to functions");
+    } else if tcx.generics_of(def_id).requires_monomorphization(tcx) {
+        tcx.sess.span_err(span, "the `proof` attribute cannot be applied to generic functions");
+    } else {
+        let instance = Instance::mono(tcx, def_id);
+        let helper = CompilerHelpers { tcx };
+        if !helper.fn_abi_of_instance(instance, ty::List::empty()).args.is_empty() {
+            tcx.sess.span_err(span, "functions used as harnesses cannot have any arguments");
```
jaisnan, line 148: Should this be an error? Is there a way to ignore the arguments and throw a warning? Not sure if that's feasible or not but it doesn't seem like something we should throw an error for.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -2,33 +2,278 @@
...
-                other_attributes.push((attribute_string.to_string(), attr));
+                tcx.sess.span_err(attr.span, "value above maximum permitted value - u32::MAX");
+                None
             }
         }
     }
+}
 
-    (proof_attributes, other_attributes)
+fn parse_stubs(tcx: TyCtxt, attributes: Vec<&Attribute>) -> Vec<Stub> {
```
jaisnan, line 192: I'm assuming this function ensures the right format for the `kani::Stub` attribute? Could you add a comment above the function?

celinval: This is converting the attributes into a list of stubs to be applied. Yes, I can totally add a comment!

kani-compiler/src/kani_middle/attributes.rs
celinval: This was already an error. Unfortunately, we can't run harnesses with arguments in a sound way today.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
celinval: Totally! What if I replace `default_kanitool_proof(&self)` by something like `generate_metadata(&self, attributes: Option<HarnessAttributes>` and set the attributes as I'm creating the metadata?

zhassan-aws: Sounds good.

PR #2213, opened 2023-02-16
Generate unit test line by line (for --inplace)

jaisnan:
### Description of changes: 

This PR adds -

1. Generating the unit tests (concrete playback tests) line by line using a sync buffered reader/writer.
2. Util functions and related unit tests.
3. Minor refactoring

The reason to generate the unit tests line by line is -
1. to avoid storing the entire unit test, no matter how large in a string variable before dumping it.
2. To make sure users still have access to the generated unit tests in case generation fails midway.

### Resolved issues:

Closes the changes proposed by #1625 

### Call-outs:

1. Uses the tempfile crate to take care of the file handling. The documentation says that the tempfile is deleted automatically whenever it goes out of scope.
2. Changes the definition of the UnitTest struct.
3. There are no regression tests for the changes made.

### Testing:

* How is this change tested? unit tests but not 

### Checklist


kani-driver/src/concrete_playback.rs
```diff
@@ -146,11 +149,13 @@ impl KaniSession {
...
 
         // Renames are usually automic, so we won't corrupt the user's source file during a crash.
-        fs::rename(temp_file.path(), source_path).with_context(|| {
+        fs::rename(&temp_path, source_path).with_context(|| {
             format!("Couldn't rename tmp source file to actual src file `{source_path}`.")
         })?;
 
-        // temp file gets deleted automatically by the NamedTempFile handler
+        drop(temp_file);
+        remove_file(temp_path)?;
```
celinval, line 159: This will only remove the file if the function succeeds though.

kani-driver/src/concrete_playback.rs
```diff
@@ -79,88 +81,86 @@ impl KaniSession {
...
+        // Use a buffered reader/writer to generate the unit test line by line
+        for line in source_reader.lines().flatten() {
+            if line.contains(&unit_test.name) {
+                if !self.args.quiet {
+                    println!(
+                        "Concrete playback unit test `{}/{}` already found in source code, so skipping modification.",
+                        source_path, unit_test.name,
+                    );
+                }
+                // temp file gets deleted automatically when function goes out of scope
```
celinval, line 134: Is that still the case.

celinval (commented):
Thanks

kani-driver/src/concrete_playback.rs
jaisnan: no, this is a redundant comment :/ .

kani-driver/src/concrete_playback.rs
jaisnan: Yes, thanks for catching that. Is the use of drop here valid? I don't know if explicitly calling drop was good practice or not but I thought it's better to be explicit than assume that it would be dropped when going out of scope.

celinval: Usually you should only invoke drop if you need it to happen at a certain time or order. I guess you could also manually invoke drop if something is taking a lot of memory. Otherwise, just leave it to the compiler. I think in this case you do want to drop `temp_file` before removing the file. You could wrap these two things into one object and drop everything in the correct order as part of its drop logic.

kani-driver/src/concrete_playback.rs
```diff
@@ -11,13 +11,44 @@ use anyhow::{Context, Result};
...
-use std::fs::{self, File};
+use std::fs::{self, remove_file, File};
 use std::hash::{Hash, Hasher};
-use std::io::{Read, Write};
-use std::path::Path;
+use std::io::{BufRead, BufReader, BufWriter, Error, Write};
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
+struct TempFile {
```
celinval, line 22: doc

kani-driver/src/concrete_playback.rs
```diff
@@ -11,13 +11,44 @@ use anyhow::{Context, Result};
...
+    file: File,
+    temp_path: PathBuf,
+}
+
+impl TempFile {
```
celinval, line 27: nit: can you either move this to `util` for now or just create a module for it.

kani-driver/src/concrete_playback.rs
```diff
@@ -11,13 +11,44 @@ use anyhow::{Context, Result};
...
+    fn new() -> Result<Self, Error> {
+        // Create temp file
```
celinval, line 29: move this comment to the function doc

kani-driver/src/concrete_playback.rs
```diff
@@ -11,13 +11,44 @@ use anyhow::{Context, Result};
...
+            Err(e) => return Err(e),
+        };
+        Ok(Self { file: temp_file, temp_path })
+    }
+}
+
+impl Drop for TempFile {
+    fn drop(&mut self) {
+        if let Err(e) = self.file.sync_all() {
+            eprintln!("Error syncing tempfile: {e}");
```
celinval, line 43: please use `util::error()`

kani-driver/src/concrete_playback.rs
```diff
@@ -79,88 +111,84 @@ impl KaniSession {
...
+    ) -> Result<bool> {
+        // Read from source
+        let source_file = File::open(source_path).unwrap();
+        let source_reader = BufReader::new(source_file);
+
+        // Create temp file
+        let mut temp_file = match TempFile::new() {
+            Ok(f) => f,
+            Err(_) => return Ok(false),
+        };
```
celinval, line 155: nit:
```rust
let mut temp_file = TempFile::try_new().map_err(|_| false)?;
```

kani-driver/src/concrete_playback.rs
```diff
@@ -11,13 +11,44 @@ use anyhow::{Context, Result};
...
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
+struct TempFile {
+    file: File,
+    temp_path: PathBuf,
+}
+
+impl TempFile {
+    fn new() -> Result<Self, Error> {
```
celinval, line 28: nit: rename this to `try_new`. Usually `try_new()` means that the creation can fail.

kani-driver/src/concrete_playback.rs
```diff
@@ -79,88 +111,84 @@ impl KaniSession {
...
+            writeln!(temp_writer, "{line}")?;
+            if curr_line_num == proof_harness_end_line {
+                for unit_test_line in unit_test.code.iter() {
+                    curr_line_num += 1;
+                    writeln!(temp_writer, "{unit_test_line}")?;
+                }
+            }
+        }
+
+        // Have to drop the bufreader to be able to reuse and rename the moved temp file
```
celinval, line 182: bufwriter

kani-driver/src/concrete_playback.rs
```diff
@@ -79,88 +114,84 @@ impl KaniSession {
...
+        drop(temp_writer);
+
+        // Renames are usually automic, so we won't corrupt the user's source file during a crash.
+        if fs::rename(&temp_file.temp_path, source_path).is_err() {
```
celinval, line 186: note: if we rename the file, we shouldn't be trying to remove it.

kani-driver/src/concrete_playback.rs
```diff
@@ -79,88 +114,84 @@ impl KaniSession {
...
+            eprintln!("Couldn't rename tmp source file to actual src file `{source_path}`.");
```
celinval, line 187: replace this by `?` operator

kani-driver/src/concrete_playback.rs
```diff
@@ -11,13 +11,44 @@ use anyhow::{Context, Result};
...
+    }
+}
+
+impl Drop for TempFile {
+    fn drop(&mut self) {
+        if let Err(e) = self.file.sync_all() {
+            eprintln!("Error syncing tempfile: {e}");
+        }
+
+        if let Err(e) = remove_file(&self.temp_path) {
```
celinval, line 46: I believe this will fail if we move the file (which happens in the success case).

celinval (commented):
It's looking good. I was wondering if we should modify the `TempFile` slightly. So, we don't want to delete the file if it doesn't exist anymore. We should also consider bringing the `BufWriter` to be inside this structure since we want to explicitly call `flush()` before dropping it and before removing the file.

kani-driver/src/concrete_playback.rs
```diff
@@ -12,12 +12,84 @@ use concrete_vals_extractor::{extract_harness_values, ConcreteVal};
...
 use std::hash::{Hash, Hasher};
-use std::io::{Read, Write};
+use std::io::{BufRead, BufReader, Write};
 use std::path::Path;
 use std::process::Command;
 
+use self::tempfile::TempFile;
+
+/// Handle a tempfile functionality
+/// for example - used during concrete playback's inplace run to write to the source file
```
celinval, line 24: please document what it does... something like
```rust
/// Handle a writable temporary file which will be deleted when the object is dropped.
/// To save the contents of the file, users can invoke `rename` which will move the file to
/// its new location and no further deletion will be performed.
```
also, please move this documentation to document the struct itself not the module.

kani-driver/src/concrete_playback.rs
```diff
@@ -12,12 +12,84 @@ use concrete_vals_extractor::{extract_harness_values, ConcreteVal};
...
+        pub temp_path: PathBuf,
+        pub writer: Option<BufWriter<File>>,
+        renamed: bool,
+    }
+
+    impl TempFile {
+        /// Create a temp file
+        pub fn try_new() -> Result<Self, Error> {
+            let mut temp_path = env::temp_dir();
+            temp_path.push("concrete_overwrite.tmp");
```
celinval, line 46: please take the suffix name as an input to the function

kani-driver/src/concrete_playback.rs
```diff
@@ -12,12 +12,84 @@ use concrete_vals_extractor::{extract_harness_values, ConcreteVal};
...
+        pub file: File,
+        pub temp_path: PathBuf,
+        pub writer: Option<BufWriter<File>>,
+        renamed: bool,
+    }
+
+    impl TempFile {
+        /// Create a temp file
+        pub fn try_new() -> Result<Self, Error> {
+            let mut temp_path = env::temp_dir();
```
celinval, line 45: You need to generate a somewhat random name to avoid collision. See the [`temp_dir` documentation](https://doc.rust-lang.org/std/env/fn.temp_dir.html):

> thus, the creation of any files or directories in the temporary directory must use a secure method to create a uniquely named file.

kani-driver/src/concrete_playback.rs
```diff
@@ -12,12 +12,84 @@ use concrete_vals_extractor::{extract_harness_values, ConcreteVal};
...
+            temp_path.push("concrete_overwrite.tmp");
+            let temp_file = File::create(&temp_path)?;
+            let writer = BufWriter::new(temp_file.try_clone()?);
+
+            Ok(Self { file: temp_file, temp_path, writer: Some(writer), renamed: false })
+        }
+
+        /// Replace a source file with the current tempfile
```
celinval, line 53:
```rust
Rename the temporary file to the new path, replacing the original file if the path points to a file that already exists.
```

kani-driver/src/concrete_playback.rs
```diff
@@ -12,12 +12,84 @@ use concrete_vals_extractor::{extract_harness_values, ConcreteVal};
...
+        pub fn rename(&mut self, source_path: &str) -> Result<(), Error> {
+            // flush here
+            self.writer.as_mut().unwrap().flush()?;
+            self.writer = None;
+            // Renames are usually automic, so we won't corrupt the user's source file during a crash.
+            if let Err(_e) = rename(&self.temp_path, source_path) {
+                util::error(&format!("Error renaming file {}", self.temp_path.to_string_lossy()));
```
celinval, line 60: in this case, you should just return an error with that message.
```rust
rename(&self.temp_file,source_path).with_context(|| format!("Error renaming file {}", self.temp_path.to_string_lossy())?;
```

kani-driver/src/concrete_playback.rs
```diff
@@ -12,12 +12,84 @@ use concrete_vals_extractor::{extract_harness_values, ConcreteVal};
...
+    /// Ensure that the bufwriter is flushed and temp variables are dropped
+    /// everytime the tempfile is out of scope
+    /// note: the fields for the struct are dropped automatically by destructor
+    impl Drop for TempFile {
+        fn drop(&mut self) {
+            // if writer is not flushed, flush it
+            if self.writer.as_ref().is_some() {
+                // couldn't use ? as drop does not handle returns
+                if let Err(_e) = self.writer.as_mut().unwrap().flush() {
+                    util::error("Couldn't flush inside drop");
```
celinval, line 76: Please log the original error
```rust
format!("failed to flush {}: {e}", self.temp_path.display())
```

kani-driver/src/concrete_playback.rs
```diff
@@ -12,12 +12,84 @@ use concrete_vals_extractor::{extract_harness_values, ConcreteVal};
...
+            let mut temp_path = env::temp_dir();
+            temp_path.push("concrete_overwrite.tmp");
+            let temp_file = File::create(&temp_path)?;
+            let writer = BufWriter::new(temp_file.try_clone()?);
+
+            Ok(Self { file: temp_file, temp_path, writer: Some(writer), renamed: false })
+        }
+
+        /// Replace a source file with the current tempfile
+        pub fn rename(&mut self, source_path: &str) -> Result<(), Error> {
```
celinval, line 54: Can you change this to take `mut self` instead of `&mut self`? Otherwise, this function will panic if called twice in a roll.

kani-driver/src/concrete_playback.rs
```diff
@@ -12,12 +12,84 @@ use concrete_vals_extractor::{extract_harness_values, ConcreteVal};
...
+                    util::error("Couldn't flush inside drop");
+                }
+                self.writer = None;
+            }
+
+            if !self.renamed {
+                if let Err(_e) = fs::remove_file(&self.temp_path.clone()) {
+                    util::error(&format!(
+                        "Error removing file {}",
+                        self.temp_path.to_string_lossy()
```
celinval, line 85: self.temp_path.display()

kani-driver/src/util.rs
```diff
@@ -14,6 +14,84 @@ use std::ffi::OsString;
...
+
+            Ok(Self { file: temp_file, temp_path, writer: Some(writer), renamed: false })
+        }
+
+        /// Rename the temporary file to the new path, replacing the original file if the path points to a file that already exists.
+        pub fn rename(mut self, source_path: &str) -> Result<(), String> {
+            // flush here
+            self.writer.as_mut().unwrap().flush().unwrap();
+            self.writer = None;
+            // Renames are usually automic, so we won't corrupt the user's source file during a crash.
```
celinval, line 61:
```suggestion
            // Renames are usually atomic, so we won't corrupt the user's source file during a crash.
```

kani-driver/src/util.rs
```diff
@@ -14,6 +14,84 @@ use std::ffi::OsString;
...
+        pub writer: Option<BufWriter<File>>,
+        renamed: bool,
+    }
+
+    impl TempFile {
+        /// Create a temp file
+        pub fn try_new(suffix_name: &str) -> Result<Self, Error> {
+            let mut temp_path = env::temp_dir();
+
+            // Generate a unique name for the temporary directory
```
celinval, line 45:
```suggestion
            // Generate a unique name for the temporary file
```

kani-driver/src/util.rs
```diff
@@ -14,6 +14,84 @@ use std::ffi::OsString;
...
+        }
+
+        /// Rename the temporary file to the new path, replacing the original file if the path points to a file that already exists.
+        pub fn rename(mut self, source_path: &str) -> Result<(), String> {
+            // flush here
+            self.writer.as_mut().unwrap().flush().unwrap();
+            self.writer = None;
+            // Renames are usually automic, so we won't corrupt the user's source file during a crash.
+            rename(&self.temp_path, source_path)
+                .with_context(|| format!("Error renaming file {}", self.temp_path.display()))
```
celinval, line 63: Can you also please add the same of the source_path?

celinval (approved):
Thanks @jaisnan. Just tiny comments

PR #2215, opened 2023-02-17
RFC: Loop-contract synthesis

### LINKED ISSUE #2214: RFC: Loop-contract synthesis
## Summary

This is a tracking issue for the RFC loop-contract synthesizer which will enable users to use ```goto-synthesizer``` to synthesize loop contracts and verify programs without unwinding loops.

## Major Steps

Suggestion of major steps:

- [ ] Get feedback on the RFC PR.
- [x] Implement the integration of `goto-synthesizer` into Kani (#2204).
- [ ] Add options to limit the resource used by the synthesizer to guarantee termination.
- [ ] Add compatibility of the synthesizer and the unwinding numbers specified by users.
- [ ] Implement the mechanism in the synthesizer to allow it synthesize loop-contracts even there are some invalid assertions.
- [ ] (Optional) Implement the quick-filter to improve the performance of the synthesizer.
- [ ] Stabilization PR.

qinheping:
### Description of changes: 

Create an RFC for the loop-contract synthesis. The synthesizer will allow users to synthesize loop contracts with which Kani can verify programs without loop unwinding bounds.

The first step of the integration of ```goto-synthesizer``` is implemented in #2204. It allows us to call the synthesizer from Kani. There are more options to be implemented in future PRs including implementations of the options to limit resource usage by the synthesizer, the compatibility of the synthesizer and unwinding numbers, and the mechanism in the synthesizer to prove some assertions to be invalid.

You can also find a rendered version of this document [here](https://github.com/qinheping/kani/blob/rfc-goto-synthesizer/rfc/src/rfcs/0004-loop-contract-synthesis.md).

### Resolved issues:

Resolves #2214
### Related RFC:

Optional #2214.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+- **Version:** 0
+- **Proof-of-concept:** <https://github.com/qinheping/kani/tree/kani-synthesizer>
+
+## Summary
+
+An new option that allows users to verify programs without unwinding loops by synthesizing loop contracts for those loops.
+
+
+## User Impact
+Currently Kani does not support proving correctness (i.e. assertions never fail) on programs with unbounded control flow (e.g. loops with dynamic bounds). Kani unrolls all unbounded loops until a few iterations and then verifies this unrolled program — it thus provides a much weaker guarantee on correctness.
```
feliperodri, line 14: I would re-phrase "it thus provides a much weaker guarantee on correctness." to see bounds as part of the assumptions.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+
+The loop-contract synthesizer will synthesize loop contracts for loops without unwinding numbers specified. With the synthesized loop contracts, CBMC abstracts those loops as non-loop structures that can cover arbitrary iterations of the loops. As a result, with the synthesizer proposed in this RFC, Kani can verify programs without unwind loops, and provide a stronger guarantee on correctness. For example, the number of executed iterations of the loop in the following harness is dynamically bounded by an unbounded variable `y`; hence no unwinding number can guarantee the correctness of the loop.
```
celinval, line 16: In the text you mention "correctness of the loop" but in the examples, there is nothing inside the loop that gets proven. What happens if there is an assertion inside the loop or an array access? Do we still prove their correctness? If so, can you please add such an example?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+    while y > 0 {
+        y = y - 1;
+    }
+    assert!(y == 0);
+}
+```
+With the loop-contract synthesizer, Kani can synthesize the loop invariant `y >= 0`, with which it can prove the post-condition `y == 0` without unwinding the loop.
+
+
+Also, these contracts, when provided, would also significantly improve Kani’s verification time since all loops would be unrolled only to a single iteration, as opposed to a number of iterations which is Kani’s current behavior. For example, although the following loop is bounded and we can easily find out the unwinding number `50000` for it, unwinding the loop 50000 times will lead to a large query, and hence slow down Kani. With loop contract, we only verify the single abstract iteration of the loop, which leads to a smaller query.
```
feliperodri, line 31: > "would also significantly improve Kani’s verification time"

What about havoc statements and their impact in performance?

feliperodri: I think it would be important to state upfront which kind of loops you want to target with this approach first and add a list of limitations. For example, could this approach still handle all memory safety checks? What if I have `break` or `continue`? What if I have dynamic allocation in the loop? What about function pointers?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+        y = y - 1;
+    }
+    assert!(y == 0);
+}
+```
+
+
+## User Experience
+
+Once this RFC has been stabilized, users will be able to synthesize loop contracts for loops and verify programs with the synthesized loop contracts with a new command line flag `--synthesize-loop-contracts`.
```
feliperodri, line 48: That's why it's important to specify for which loops will Kani support in this RFC.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+        x = 2;
+    }
+    assert!(y == 0 || y == -1);
+}
+
+#[kani::unwind(inner:2)]
+fn bar() {
+    let mut y: i32 = 50000;
+    let mut x: i32 = 2;
+    `outer`: while y > 0 {
```
celinval, line 73: that's a neat idea... can you please expand how this works when the user also provides the command line `--unwind` option? Also please make it clear what happens if the user uses a loop specific bound and a generic one for the harness.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+        while x > 0 {
+            y = y - 1;
+            x = x - 1;
+        }
+        x = 2;
+    }
+    assert!(y == 0 || y == -1);
+}
+
+#[kani::unwind(inner:2)]
```
celinval, line 69: Are you planning to make this work for loops outside of the harness?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+
+
+## User Experience
+
+Once this RFC has been stabilized, users will be able to synthesize loop contracts for loops and verify programs with the synthesized loop contracts with a new command line flag `--synthesize-loop-contracts`.
+
+### How the Synthesizer Works with Unwinding Numbers
+For each loop, users will be able to choose to unwind it for a certain number of times by specifying the unwinding numbers explicitly, or to synthesize loop contracts for it when no unwinding number specified. 
+
+We introduce a new way for users to specify unwinding numbers for a single loops rather than just for the whole functions: specifying loop numbers using labels. E.g., in the following program, both loops in the function `foo` will be unwound `5` times, while only the loop `inner` in the function `bar` will be unwound `2` times.
```
feliperodri, line 53:
```suggestion
We introduce a new way for users to specify unwinding numbers for a single loop rather than just for the whole function: specifying loop numbers using labels. In the following program, both loops in the function `foo` will be unwound `5` times, while only the loop `inner` in the function `bar` will be unwound `2` times.
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
celinval: Also, you might want to use "=" instead of ":"

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+        `inner`: while x > 0 {
+            y = y - 1;
+            x = x - 1;
+        }
+        x = 2;
+    }
+    assert!(y == 0 || y == -1);
+}
+```
+Then, we the flag `--synthesize-loop-contracts` is present, the synthesizer will synthesize loop contracts for only the loop `outer` in the function `bar` and unwind the other three loops.
```
celinval, line 83: nit: s/we/when

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+
+## User Experience
+
+Once this RFC has been stabilized, users will be able to synthesize loop contracts for loops and verify programs with the synthesized loop contracts with a new command line flag `--synthesize-loop-contracts`.
+
+### How the Synthesizer Works with Unwinding Numbers
+For each loop, users will be able to choose to unwind it for a certain number of times by specifying the unwinding numbers explicitly, or to synthesize loop contracts for it when no unwinding number specified. 
+
+We introduce a new way for users to specify unwinding numbers for a single loops rather than just for the whole functions: specifying loop numbers using labels. E.g., in the following program, both loops in the function `foo` will be unwound `5` times, while only the loop `inner` in the function `bar` will be unwound `2` times.
+```rust
```
celinval, line 54: where is the proof in this example?

rfc/src/rfcs/0004-loop-contract-synthesis.md
feliperodri, line 83:
```suggestion
Then, when using the `--synthesize-loop-contracts` flag, Kani will synthesize loop contracts for only the loop `outer` in the function `bar` and unwind the other three loops.
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+    }
+    assert!(y == 0 || y == -1);
+}
+```
+Then, we the flag `--synthesize-loop-contracts` is present, the synthesizer will synthesize loop contracts for only the loop `outer` in the function `bar` and unwind the other three loops.
+
+### Limit Resource Used by Synthesizer for Termination
+Without resource limit, an enumerative synthesizer may run forever to exhaust search space consist of infinite number of candidates, especially when there is no solution in the search space. So, for the guarantee of termination, we provide users options: <code>--limit-synthesis-time <ins>T</ins></code> to limit the running time of the synthesizer, and <code>--limit-synthesis-size <ins>N</ins></code> to limit the sizes (number of nodes in the AST of the candidate predicate) of loop-invariant candidates we search for during the enumeration.
+
+We encourage users to specify such limit while we hardcode some default limit at the same time.
```
celinval, line 88: what will be the hardcoded limit?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+        }
+        x = 2;
+    }
+    assert!(y == 0 || y == -1);
+}
+```
+Then, we the flag `--synthesize-loop-contracts` is present, the synthesizer will synthesize loop contracts for only the loop `outer` in the function `bar` and unwind the other three loops.
+
+### Limit Resource Used by Synthesizer for Termination
+Without resource limit, an enumerative synthesizer may run forever to exhaust search space consist of infinite number of candidates, especially when there is no solution in the search space. So, for the guarantee of termination, we provide users options: <code>--limit-synthesis-time <ins>T</ins></code> to limit the running time of the synthesizer, and <code>--limit-synthesis-size <ins>N</ins></code> to limit the sizes (number of nodes in the AST of the candidate predicate) of loop-invariant candidates we search for during the enumeration.
```
feliperodri, line 86: A link for the theoretical approach here would be nice. Wider audience might not know what is a "candidate" in this context.

rfc/src/rfcs/0004-loop-contract-synthesis.md
feliperodri, line 88: What are the default limits?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Detailed Design
+The synthesizer ```goto-synthesizer``` is implemented in the repository of `CBMC`, takes as input a goto binary, and outputs a new goto binary with the synthesized loop contracts applied. Currently, Kani invoke `goto-instrument` to instrument the goto binary `main.goto` into a new goto binary `main_instrumented.goto`, and then invoke ```cbmc``` on `main_instrumented.goto` to get the verification result. The synthesis will happen between calling `goto-instrument` and calling ```cbmc```. That is, we invoke ```goto-synthesizer``` on ```main_instrumented.goto``` to produce a new goto binary ```main_synthesized.goto```, and then call ```cbmc``` on `main_synthesized.goto` instead. 
+
+When invoking ```goto-synthesizer```, we pass the following parameters to it with the flags built in `goto-synthesizer`:
+- the resource limit of the synthesis;
+- the unwinding numbers of loops.
+Note that the way we identify loops in goto and the way we identify loop in rust are different. So we also need to parse the label-based loop id to the index-based loop id used by goto.
```
celinval, line 103: Can you also expand how you are planning to do this?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+            
+NT_int  -> NT_int + NT_int | terminals_int | LOOP_ENTRY(terminals_int)
+            | POINTER_OFFSET(terminals_ptr) | OBJECT_SIZE(terminals_ptr)
+            | POINTER_OFFSET(LOOP_ENTRY(terminals_ptr)) | 1
+```
+where `terminals_ptr` are all pointer variables in the scope, and `terminal_int` are all integers variables in the scope.
+
+## Rationale and alternatives
+
+- Different way to identify loops. Instead of using labels for loops, we can use loop number as what CBMC does. Loop numbers are indexes of loops according to the syntactic order. Also, we can alternatively use inlined pragmas to specify unwinding numbers for loops . However, inlined pragmas will hurt the user experience of no annotations in users code while using Kani.
```
feliperodri, line 120: Using labels for loops might be a major concern. Can a Rust developer add labels to all loops in their code? What if the loop comes from a dependency?

celinval (commented):
That's pretty neat idea. Can you please provide a link or a bit more context on what loop synthesis is? It would be great if you could also link to the CBMC documentation for this feature.

rfc/src/rfcs/0004-loop-contract-synthesis.md
zhassan-aws, line 53: How would loops due to iterators be handled?

rfc/src/rfcs/0004-loop-contract-synthesis.md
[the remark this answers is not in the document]
remi-delmas-3000: all of this to say that we might need to locate loop occurrences in the call graph of the program like `harness.foo_call.bar_call1.BAR_LOOP`

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+
+We encourage users to specify such limit while we hardcode some default limit at the same time. 
+
+### Output of Kani when Synthesizer Enable
+When the flag `--synthesize-loop-contracts` is provided, Kani will report different result for different cases
+1. When there exists some loop invariant in the candidate space with which all assertions can be proved, Kani will synthesize the loop contracts, verify the program with the synthesized loop contracts, and report verification SUCCESS;
+2. When there exists no loop invariants in the candidate space that can be used to prove all assertions within the resource limit, Kani will report the verification result with the best-effort-synthesized loop contracts.
```
remi-delmas-3000, line 93: When no working candidate has been found in the search space within the specified limits

remi-delmas-3000 (changes requested):

celinval:
Another few questions:

1. What happens in the error scenarios? What happens if the synthesizer fails? What kind of outcomes do we have?
2. Does this work for recursion?
3. List all the current limitations.
4. When applying the synthesis, can we ensure that havoc the variables take into consideration type constraints?
5. Maybe leave the loop selection as an open question for now and focus on polishing the UX for cases where we either apply or not apply the loop synthesis.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
+- **Feature Name:** Loop-contract synthesis
+- **Feature Request Issue:** <https://github.com/model-checking/kani/issues/2214>
+- **RFC PR:** <https://github.com/model-checking/kani/pull/2215>
+- **Status:** Under Review
+- **Version:** 0
+- **Proof-of-concept:** <https://github.com/qinheping/kani/tree/kani-synthesizer>
+
+## Summary
+
+An new option that allows users to verify programs without unwinding loops by synthesizing loop contracts for those loops.
```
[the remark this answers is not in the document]
qinheping: No, we will apply synthesized loop contracts on all loops instead of unwinding them.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+```rust
+#[kani::proof]
+#[kani::unwind(50000)]
+fn main() {
+    let mut y: i32 = 50000;
+
+    while y > 0 {
+        y = y - 1;
+    }
+    assert!(y == 0);
```
karkhaz, line 41: Does the contract assume that the loop will terminate after an arbitrary but finite number of iterations?

I wonder about the case where users write an infinite loop. Kani would catch this by default because of the unwinding assertion, but if this feature abstracts away that loop and replaces it with one that will terminate, it could hide bugs.

In this code snippet, if you change the loop body to `y = 1;` so that the loop doesn't terminate, would kani correctly identify that the assertion cannot be reached?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,145 @@
...
+The synthesizer ```goto-synthesizer``` is implemented in the repository of `CBMC`, takes as input a goto binary, and outputs a new goto binary with the synthesized loop contracts applied. Currently, Kani invoke `goto-instrument` to instrument the goto binary `main.goto` into a new goto binary `main_instrumented.goto`, and then invoke ```cbmc``` on `main_instrumented.goto` to get the verification result. The synthesis will happen between calling `goto-instrument` and calling ```cbmc```. That is, we invoke ```goto-synthesizer``` on ```main_instrumented.goto``` to produce a new goto binary ```main_synthesized.goto```, and then call ```cbmc``` on `main_synthesized.goto` instead. 
+
+When invoking ```goto-synthesizer```, we pass the following parameters to it with the flags built in `goto-synthesizer`:
+- the resource limit of the synthesis;
+- the unwinding numbers of loops.
+Note that the way we identify loops in goto and the way we identify loop in rust are different. So we also need to parse the label-based loop id to the index-based loop id used by goto.
+
+
+The enumerator used in the synthesizer enumerate candidates from the language of the following grammar template
```
karkhaz, line 106: I don't really understand what this means, I agree with @feliperodri comment about explaining about candidates.

I wonder, would it be practical to give an example here? I.e. for a particular snippet of code, show what the resulting candidate would be? It's difficult for me to relate these constructs to what I normally think of as an assertion

rfc/src/rfcs/0004-loop-contract-synthesis.md
qinheping: It should be "Only the unwinding number `i32::MAX` can guarantee to cover all iteration of the loop, and hence satisfies the unwinding assertions.".

rfc/src/rfcs/0004-loop-contract-synthesis.md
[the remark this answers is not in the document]
qinheping: Even thought there two occurrences of `bar`, we only synthesize one loop invariant for the loop in `bar` to cover both occurrences.

rfc/src/rfcs/0004-loop-contract-synthesis.md
qinheping: There are decreases clauses in loop contracts with which we can prove the termination of loops. The idea of decreases clauses is to prove that some quantity monotonically decrease along the execution of loops.
However, synthesizing decreases clauses is not yet supported in `goto-synthesizer`.

qinheping:
Thanks for the comments!

> Another few questions:
> 
>     1. What happens in the error scenarios? What happens if the synthesizer fails? What kind of outcomes do we have?
The synthesizer will still try to synthesize loop contracts in error scenarios. But it probably runs forever if no resource limit specified.
If the synthesizer fails within the given resource limits, it will still apply what loop contracts it found in the goto model. Kani will still run CBMC on the goto model and report the verification result. 
With the flag `--dump-loop-contracts`, it will also dump synthesized loop contracts in a json file. However, without a good way to identify loops, the dump loop contracts will use loop numbers, which could be a bit confusing for Rust programs.
>     2. Does this work for recursion?
No. We need function contracts for recursion,
>     3. List all the current limitations.
I listed limitations in the User Impact section.
>     4. When applying the synthesis, can we ensure that havoc the variables take into consideration type constraints?
The type constraints of havocing variables are the type of them declared in the goto model.
>     5. Maybe leave the loop selection as an open question for now and focus on polishing the UX for cases where we either apply or not apply the loop synthesis.
Yes.

zhassan-aws (commented):
What was the conclusion on allowing unwinding of individual loops?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Limit Resource Used by Synthesizer for Termination
+Without resource limit, an enumerative synthesizer may run forever to exhaust search space consist of infinite number of candidates, especially when there is no solution in the search space.
+So, for the guarantee of termination, we provide users options: `--limit-synthesis-time T` to limit the running time of the synthesizer.
+
+
+#### Output of Kani when Synthesizer Enable
+When the flag `--synthesize-loop-contracts` is provided, Kani will report different result for different cases
+1. When there exists some loop invariant in the candidate space with which all assertions can be proved, Kani will synthesize the loop contracts, verify the program with the synthesized loop contracts, and report verification SUCCESS;
+2. When no working candidate has been found in the search space within the specified limits, Kani will report the verification result with the best-effort-synthesized loop contracts.
```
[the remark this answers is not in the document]
qinheping: Yes, I will clarify in the text.

rfc/src/rfcs/0004-loop-contract-synthesis.md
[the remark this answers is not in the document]
qinheping: >    Does the synthesizer need to be applied to a whole program?

Yes, it will synthesize loop invariants for all loops at the same time.

>    Can we get something out of the synthesizer before it gets killed?

Yes, we plan to implement best-effort-synthesis.
 
>   Why not use memory instead of AST to refer to size?

We will use running time only.

rfc/src/rfcs/0004-loop-contract-synthesis.md
qinheping: It is an open problem for `goto-synthesizer`. Once `goto-synthesizer` supports loops over iterators, loop-contract synthesis in Kani will also support loops over iterators.

qinheping:
@zhassan-aws 
> What was the conclusion on allowing unwinding of individual loops?

We don't allow unwinding of individual loops when the synthesizer is enable now. We will discuss more about how to identify loops when we introduce loop contracts to Kani.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,188 @@
...
+y == LOOP_ENTRY(y)
+y == 1
+...
+1 <= y + 1
+...
+```
+The synthesizer then verifies with CBMC if the candidate is an inductive invariant that can be used to prove the post-condition `y == 0`.
+For example, the candidate `y == y` is verified to be an inductive invariant, but cannot be used to prove the post-condition `y == 0`.
+The candidate `y == 1` is not inductive.
+The synthesizer rejects all candidates until it finds the candidate `1 <= y + 1`, which can be simplified to `y >= 0`.
```
zhassan-aws, line 137: Does the synthesizer call `cbmc` every time it finds an inductive invariant?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,188 @@
...
+The candidate grammar introduced above now only contains a restricted set of operators, which works well for array-manipulating programs with only pointer-checks instrumented by `goto-instrument`, but probably not enough for other user-written checks.
+We may want to include array-indexing, pointer-dereference, or other arithmetic operators in the candidate grammar for synthesizing a larger set of loop invariants.
+However, there is a trade-off between the size of candidate we enumerate and the running time of the enumeration.
+We will collect more data to decide what operators we should include in the candidate grammar.
+
+## Open questions
+
+**How does the synthesizer work with unwinding numbers?**
+There may exist some loops for which the synthesizer cannot find loop contracts, but some small unwinding numbers are enough to cover all executions of the loops.
+In this case, we may want to unwind some loops in the program while synthesizing loop contracts for other loops.
```
zhassan-aws, line 154: It would be nice if we can introduce a macro to enable/disable unwinding, e.g.
```rust
#[kani::proof]
#[kani::unwind(10)]
fn check() {
    // unwinding starts as enabled, so all loops in this code block will be unwound to 10
    #[kani::disable_unwinding]
    // unwinding is disabled for all loops in this block of code
    #[kani::enable_unwinding]
    // it is enabled in this block of code until the end of the program
}
```
but not sure if this is feasible since it may require incorporating the synthesizer within the symbolic execution engine.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,188 @@
...
+It enumerates candidate invariants from a pre-designed search space described by a given regular tree grammar and verifies if the candidate is an inductive invariant.
+Therefore it has the following limitations:
+1. the search space is not complete, so it may fail to find a working candidate. The current search space consists of only conjunction of linear inequalities built from the variables in the loop, which is not expressive enough to capture all loop invariants.
+For example, the loop invariant `a[i] == 0` contains array access and cannot be captured by the current search space.
+However, we can easily extend the search space to include more complex expressions with the cost of an exponential increase of the running time of the synthesizer.
+2. the synthesizer suffers from the same limitation as the loop contract verification in CBMC. For example, it dose not support unbounded quantifiers, or dynamic allocations in the loop body. 
+
+## User Experience
+
+Once this RFC has been stabilized, users will be able to synthesize loop contracts for all loops and verify programs with the synthesized loop contracts with a new command line flag `--synthesize-loop-contracts`.
```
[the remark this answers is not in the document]
qinheping: Good point!

rfc/src/rfcs/0004-loop-contract-synthesis.md
qinheping: Yes, the synthesizer run `CBMC` every time it needs to check if a candidate is inductive.

zhassan-aws: And if CBMC says it is inductive, will it call CBMC again to check the program?

rfc/src/rfcs/0004-loop-contract-synthesis.md
qinheping: Good point, in C we similarly use pragmas to enable/disable certain checks for a block of code.
What I am not sure about this is how common it is where we want to unwind some loops while synthesize loop contracts for others.
I will add it as future possibilities.

rfc/src/rfcs/0004-loop-contract-synthesis.md
qinheping: They are checked in the same call to CBMC.
For a candidate invariant, we apply it to the program and run CBMC to verify the programs including the inductiveness checks and all other checks.
If all checks pass, the candidate invariant is good to return.
If inductive checks pass, the candidate invariant is inductive, we can keep it as the strongest inductive invariant we have seen.
If inductive checks fail, we discard the candidate.

zhassan-aws: I see, thanks for explaining. Can you clarify this a bit in the RFC as it wasn't clear to me?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,188 @@
...
+The synthesis will happen between calling `goto-instrument` and calling ```cbmc```.
+That is, we invoke ```goto-synthesizer``` on ```main_instrumented.goto``` to produce a new goto binary ```main_synthesized.goto```, and then call ```cbmc``` on `main_synthesized.goto` instead. 
+
+When invoking ```goto-synthesizer```, we pass the following parameters to it with the flags built in `goto-synthesizer`:
+- the resource limit of the synthesis;
+- the solver options to specify what SAT solver we use to verify invariant candidates.
+
+The enumerator used in the synthesizer enumerates candidates from the language of the following grammar template
+```
+NT_Bool -> NT_Bool && NT_Bool | NT_int == NT_int
```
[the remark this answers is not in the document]
qinheping: For better performance, we want to use a candidate grammar that is as smaller as possible. This grammar performs well on the C benchmarks. 
Maybe we should provide options to users and let they choose what kind of grammar they want to use.

rfc/src/rfcs/0004-loop-contract-synthesis.md
qinheping: Sure.

rfc/src/rfcs/0004-loop-contract-synthesis.md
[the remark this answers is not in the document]
qinheping: I will add more discussion about it in **Different candidate space** under **Rationale and alternatives**.

zhassan-aws (approved):
Thanks @qinheping! Please wait for another approval because RFCs require 2 approvals.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+- **Version:** 0
+- **Proof-of-concept:** <https://github.com/qinheping/kani/tree/kani-synthesizer>
+
+## Summary
+
+A new option that allows users to verify programs without unwinding loops by synthesizing loop contracts for those loops.
+
+
+## User Impact
+Currently Kani does not support proving correctness (i.e. assertions never fail) on programs with unbounded control flow (e.g. loops with dynamic bounds). Kani unrolls all unbounded loops until a global threshold (unwinding number) specified by the user and then verifies this unrolled program, which limits the set of programs it can verify.
```
adpaco-aws, line 14:
```suggestion
Currently Kani does not support proving correctness (i.e. assertions never fail) on programs with unbounded control flow (e.g. loops with dynamic bounds).
Kani unrolls all unbounded loops up to a global threshold (unwinding number) specified by the user and then verifies this unrolled program, which limits the set of programs it can verify.
```

adpaco-aws: Also, why did you chose "proving correctness" here? If I'm not mistaken, the same is true for any verification you're trying to do with Kani.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+
+A new Kani flag `--synthesize-loop-contracts` will be created that can be used to enable the goto-level loop-contract synthesizer `goto-synthesizer`.
+The idea of [loop contracts](https://arxiv.org/pdf/2010.05812.pdf) is, instead of unwinding loops, we abstracts those loops as non-loop structures that can cover arbitrary iterations of the loops.
```
adpaco-aws, line 17:
```suggestion
The idea of [loop contracts](https://arxiv.org/pdf/2010.05812.pdf) is, instead of unwinding loops, we abstract those loops as non-loop structures that can cover arbitrary iterations of the loops.
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+The loop contract synthesizer, when enabled, will attempt to synthesize loop contracts for all loops.
+CBMC can then apply the synthesized loop contracts and verify the program without unwinding any loop.
+So, the synthesizer will help to verify the programs that require Kani to unwind loops for a very large number of times to cover all iterations.
+
+
+For example, the number of executed iterations of the loop in the following harness is dynamically bounded by an unbounded variable `y`.
```
adpaco-aws, line 23: Actually, the variable `y` is bounded on the lower end.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+Only an unwinding value of `i32::MAX` can guarantee to cover all iteration of the loop, and hence satisfies the unwinding assertions.
```
adpaco-aws, line 24:
```suggestion
Only an unwinding value of `i32::MAX` can guarantee to cover all iterations of the loop, and hence satisfy the unwinding assertions.
```
? Not sure about "satisfies -> satisfy", but I think you're talking about the unwinding value.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+Unwinding the loop `i32::MAX` number of time will result in a too large goto program to be verified by CBMC.
```
adpaco-aws, line 25:
```suggestion
Unwinding the loop an `i32::MAX` number of times will result in a too large goto program to be verified by CBMC.  
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+        y = y - 1;
+    }
+    assert!(y == 0);
+}
+```
+
+The `goto-synthesizer` is an [enumeration-based synthesizer](https://www.cis.upenn.edu/~alur/SyGuS13.pdf).
+It enumerates candidate invariants from a pre-designed search space described by a given regular tree grammar and verifies if the candidate is an inductive invariant.
+Therefore it has the following limitations:
+1. the search space is not complete, so it may fail to find a working candidate. The current search space consists of only conjunction of linear inequalities built from the variables in the loop, which is not expressive enough to capture all loop invariants.
```
adpaco-aws, line 61:
```suggestion
1. the search space is not complete, so it may fail to find a working candidate. The current search space consists of only conjunctions of linear inequalities built from the variables in the loop, which is not expressive enough to capture all loop invariants.
```
I'm guessing you build all possible conjunctions and not just one? This seems like a typo but may lead to confusion.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+For example, the loop invariant `a[i] == 0` contains array access and cannot be captured by the current search space.
```
adpaco-aws, line 62:
```suggestion
For example, the loop invariant `a[i] == 0` contains an array access and cannot be captured by the current search space.
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+However, we can easily extend the search space to include more complex expressions with the cost of an exponential increase of the running time of the synthesizer.
+2. the synthesizer suffers from the same limitation as the loop contract verification in CBMC. For example, it does not support unbounded quantifiers, or dynamic allocations in the loop body. 
+
+## User Experience
+
+Once this RFC has been stabilized, users will be able to use the new command-line flag `--synthesize-loop-contracts` to run the synthesizer, which will attempt to synthesize loop contracts, and verify programs with the synthesized loop contracts.
```
adpaco-aws, line 68: Why are we discussing stabilization now? In my opinion, this proposal should cover the unstable feature (enabled through `--Zsynthesize-loop-contracts` or similar?), and once it's ready to stabilize we can discuss further.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+
+
+#### Limit Resource Used by Synthesizer for Termination
+Without a resource limit, an enumerative synthesizer may run forever to exhaust a search space consisting of an infinite number of candidates, especially when there is no solution in the search space.
+So, for the guarantee of termination, we provide users options: `--limit-synthesis-time T` to limit the running time of the synthesizer.
```
adpaco-aws, line 73: Please mention what `T` represents: seconds?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+
+
+#### Output of Kani when the Synthesizer is Enabled
+When the flag `--synthesize-loop-contracts` is provided, Kani will report different result for different cases
+1. When there exists some loop invariant in the candidate space with which all assertions can be proved, Kani will synthesize the loop contracts, verify the program with the synthesized loop contracts, and report verification SUCCESS;
```
adpaco-aws, line 78: Question: Are programs with failures (e.g., assertions) not good candidates for loop-contract synthesis? In other words, is it possible to deduce loop contracts for programs which would result in `FAILURE`?

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Detailed Design
+The synthesizer ```goto-synthesizer``` is implemented in the repository of `CBMC`, takes as input a goto binary, and outputs a new goto binary with the synthesized loop contracts applied.
+Currently, Kani invokes `goto-instrument` to instrument the goto binary `main.goto` into a new goto binary `main_instrumented.goto`, and then invokes ```cbmc``` on `main_instrumented.goto` to get the verification result.
+The synthesis will happen between calling `goto-instrument` and calling ```cbmc```.
+That is, we invoke ```goto-synthesizer``` on ```main_instrumented.goto``` to produce a new goto binary ```main_synthesized.goto```, and then call ```cbmc``` on `main_synthesized.goto` instead. 
+
+When invoking ```goto-synthesizer```, we pass the following parameters to it with the flags built in `goto-synthesizer`:
+- the resource limit of the synthesis;
+- the solver options to specify what SAT solver we use to verify invariant candidates.
```
adpaco-aws, line 93: Is this still true? I recall you changed it so the SAT solver is just the one used for verification.

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+
+The enumerator used in the synthesizer enumerates candidates from the language of the following grammar template
```
adpaco-aws, line 95:
```suggestion
The enumerator used in the synthesizer enumerates candidates from the language of the following grammar template.
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+```
+NT_Bool -> NT_Bool && NT_Bool | NT_int == NT_int 
+            | NT_int <= NT_int | NT_int < NT_int 
+            | SAME_OBJECT(terminals_ptr, terminals_ptr)
+            
+NT_int  -> NT_int + NT_int | terminals_int | LOOP_ENTRY(terminals_int)
+            | POINTER_OFFSET(terminals_ptr) | OBJECT_SIZE(terminals_ptr)
+            | POINTER_OFFSET(LOOP_ENTRY(terminals_ptr)) | 1
+```
+where `terminals_ptr` are all pointer variables in the scope, and `terminal_int` are all integers variables in the scope.
```
adpaco-aws, line 105:
```suggestion
where `terminals_ptr` are all pointer variables in the scope, and `terminal_int` are all integer variables in the scope.
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+```
+The synthesizer will enumerate candidates derived from `NT_Bool` in the following order.
+```
+y == y
+y == LOOP_ENTRY(y)
+y == 1
+...
+1 <= y + 1
+...
+```
```
adpaco-aws, line 139: Nice step-by-step example, thanks!

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+The synthesizer then verifies with CBMC if the candidate is an inductive invariant that can be used to prove the post-condition `y == 0`.
+For example, the candidate `y == y` is verified to be an inductive invariant, but cannot be used to prove the post-condition `y == 0`.
+The candidate `y == 1` is not inductive.
+The synthesizer rejects all candidates until it finds the candidate `1 <= y + 1`, which can be simplified to `y >= 0`.
+`y >= 0` is an inductive invariant that can be used to prove the post-condition.
+So the synthesizer will return `y >= 0` and apply it in the goto model to get `main_synthesized.goto`.
```
adpaco-aws, line 145:
```suggestion
So the synthesizer will return `y >= 0` and apply it to the goto model to get `main_synthesized.goto`.
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+Once we decide more kinds of candidate grammars, we will provide users options to choose which candidate grammar they want to use.
+
+## Open questions
+
+**How does the synthesizer work with unwinding numbers?**
+There may exist some loops for which the synthesizer cannot find loop contracts, but some small unwinding numbers are enough to cover all executions of the loops.
+In this case, we may want to unwind some loops in the program while synthesizing loop contracts for other loops.
+It requires us to have a way to identify and specify which loops we want to unwind. 
+
+In C programs, we identify loops by **loop id**, which is a pair (function name, loop number).
```
adpaco-aws, line 164:
```suggestion
In C programs, we identify loops by the **loop ID**, which is a pair `(function name, loop number)`.
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+However, it Rust programs, loops are usually in library functions such as `Iterator::for_each`.
```
adpaco-aws, line 165:
```suggestion
However, in Rust programs, loops are usually in library functions such as `Iterator::for_each`.
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+And a library function may be called from different places in the program.
+We may want to unwind the loop in some calls but not in other calls.
+
+**How do we output the synthesized loop contracts?**
+To better earn users' trust, we want to be able to report what loop contracts we synthesized and used to verify the given programs.
+Now `goto-synthesizer` can dump the synthesized loop contracts into a json file.
```
adpaco-aws, line 171:
```suggestion
Now `goto-synthesizer` can dump the synthesized loop contracts into a JSON file.
```

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+Here is an example of the dumped loop contracts.
+It contains the location of source files of the loops, the synthesized loop invariants and loop assigns for loops identified by loop numbers.
```
adpaco-aws, line 173: "loops" appears 5 times in this sentences, but "loop assigns" doesn't mean anything. Can you rewrite this sentence? For example, I believe "loop assigns" can be replaced with "assign clauses".

rfc/src/rfcs/0004-loop-contract-synthesis.md
```diff
@@ -0,0 +1,209 @@
...
+      }
+    ],
+    "output": "stdout"
+}
+```
+
+There are two challenges here if we want to also dump synthesized loop contracts in Kani.
+1. We need to have a consistent way to identify loops.
+2. We need to dump loop invariants in `rust` instead of `c`.
+3. There are many auxiliary variables we added in Kani-compiled GOTO, such as `var_9`, `var_10`, `var_11`, and `var_12` in the above json file.
```
adpaco-aws, line 190:
```suggestion
3. There are many auxiliary variables we added in Kani-compiled GOTO, such as `var_9`, `var_10`, `var_11`, and `var_12` in the above JSON file.
```

adpaco-aws (changes requested):

rfc/src/rfcs/0004-loop-contract-synthesis.md
qinheping: I added a footnote to clarify it.

rfc/src/rfcs/0004-loop-contract-synthesis.md
qinheping: Loop contracts over-approximate the states that we can reach at any iteration of the loops. 
We say a candidate is valid if it is inductive and it satisfy the base case.
But if there are still violated property, we can't distinguish whether the candidate is not strong enough to prove the property, or the property is always violated with no matter what loop contracts.
So we say a candidate good only when there is no violation.

Synthesis with limited resource will return the strongest valid candidate found when the limit reached.

rfc/src/rfcs/0004-loop-contract-synthesis.md
qinheping: Yes, we propagate the solver options from Kani to goto-synthesizer.

adpaco-aws (approved):
Thanks @qinheping for your work on this! The RFC looks good to me, but we'll need to iterate on the UX.

In particular, for future versions, I'd like to see a discussion on invariant caching (i.e., attempt to reuse previously working invariants) and user-provided loop-contract annotations. These features are difficult to attempt at the moment given the limitations on loop identification, but let's not forget about them. Please mention these features on the **Future possibilities** section.

Other than that, I'm excited to see what loop-contract synthesis can enable!

rfc/src/rfcs/0004-loop-contract-synthesis.md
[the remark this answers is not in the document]
qinheping: Add these to the future possibilities once when have a good answer to how we identify loops in Rust programs.

rfc/src/rfcs/0004-loop-contract-synthesis.md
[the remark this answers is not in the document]
qinheping: Removed the AST nodes related text.

remi-delmas-3000 (approved):

PR #2221, opened 2023-02-20
Don't rely on JSON symtabs for --gen-c, --function

remi-delmas-3000:
### Description of changes: 

When `--write-goto-binary` is used, JSON symtab files are not produced, so we should not rely on them.
For `--gen-c` we now store the `name` to `prettyName` mapping in its own JSON `ArtifactType::PrettyNameMap`.
For `--function`, we modify `CargoOutputs` so that it tracks GOTO symtabs instead of JSON symtabs.

### Testing:

* How is this change tested?

CI regression run.

* Is this a refactor change?

No.

### Checklist


kani-driver/src/call_goto_instrument.rs
```diff
@@ -57,7 +57,8 @@ impl KaniSession {
             }
 
             let c_demangled = alter_extension(output, "demangled.c");
-            let symtab = project.get_harness_artifact(&harness, ArtifactType::SymTab).unwrap();
+            let symtab =
```
zhassan-aws, line 60: Can you update the variable name (`symtab`)?

zhassan-aws (approved):
Thanks! One minor comment. Otherwise looks good.

kani-driver/src/call_goto_instrument.rs
remi-delmas-3000: fixed

celinval (approved):

PR #2228, opened 2023-02-22
Add `--fix-expected` to help batch fix to expected tests (Internal tool only)

celinval:
### Description of changes: 

This option will basically override the expected file with Kani's output for every test that fails the expectation. Developers should still prune the expected file to ensure only the information that is needed gets pushed.

My motivation is that I always find myself updating expected tests manually by rerunning Kani for each failure, piping the output to the expected file and edit the file after. With this hack, I only need to do the last step manually.

Since we still fail the test, the files that were updated are listed, which makes it very easy to track down the changes.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

Note that with great power comes great responsibility. Developers still need to manually edit the expected files before publishing a PR.

### Testing:

* How is this change tested? Manually tested in the last PR I created.

### Checklist

### DIFF
--- a/tools/compiletest/src/common.rs
+++ b/tools/compiletest/src/common.rs
@@ -140,6 +140,11 @@ pub struct Config {
 
     /// Whether we will run the tests or not.
     pub dry_run: bool,
+
+    /// Whether we should update expected tests when there is a mismatch. This is helpful for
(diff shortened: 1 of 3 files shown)


zhassan-aws (approved):
Thanks.

Issue #2231, opened 2023-02-23
Add a test that the build cache is sensitive to different Kani versions


celinval:
**Proposed change:** We need a way to test that change to Kani versions invalidate the build cache.

**Motivation:** Failure to do so may cause inconsistent behavior.

Issue #2230, opened 2023-02-23
Split "codegen" flags to reduce recompilation


celinval:
Requested feature: Only recompile the artifacts that are affected by changes to Kani options.
Use case: After enabling the build cache, any change to the Kani arguments that influence how we invoke `kani-compiler` will trigger a full recompilation. This is unnecessary in some cases (e.g.: reachability checks), which are only applied to the target crate being compiled.

Issue #2229, opened 2023-02-23
Build cache should consider changes to the compiler


celinval:
**Proposed change:** For internal development, we should include information about the current compiler version (e.g.: the commit hash or a hash of the `kani-compiler` binary. We should also include the toolchain version that's being used.

**Motivation:** This will alleviate the development pain of remembering to do a `cargo clean` whenever making changes to the compiler. Not doing so can manifest in weird / inconsistent results that end up taking time to debug.

celinval:
Some context: https://github.com/rust-lang/cargo/pull/8073

celinval:
For Kani developers, if you make any change to `kani-compiler` or the rust toolchain, you might need to clean the build cache. For the regression, remove the `build/` folder in the repo root, for one off test, delete the `target` directly.

PR #2232, opened 2023-02-23
Enable build cache in `cargo kani`

### LINKED ISSUE #1736: Implement proper caching mechanism for files during compilation
Currently, we "do it the way its always been done".  https://github.com/model-checking/kani/pull/1686#discussion_r984694402

We should have a principled choice for what to keep when.

### LINKED ISSUE #2036: Kani needlessly recompiles everything every time

### LINKED ISSUE #2140: Kani compiler should include its backend version
**Proposed change:** The `kani-compiler` should include its version when execution includes `--version --verbose`.

**Motivation:** Cargo uses the compiler version to create a hash code for its compiler. This currently does not include the backend version, but it sounds like they are opened to change it. So this will probably be the best mechanism to implement proper caching invalidation when our users update Kani in case the new version still relies on the same version rustc version.

https://github.com/rust-lang/cargo/blob/ee755e77dae6cd419babe6f778dfff9337a4c7cf/src/cargo/core/compiler/context/compilation_files.rs#L662-L664

celinval:
### Description of changes: 

This change re-enable the build cache in Kani. For that, we also added a new dummy option that includes Kani's version, which will ensure that the cache gets refreshed when the user upgrades Kani.

### Resolved issues:

Resolves #1736 
Resolves #2140
Resolves #2036

### Call-outs:

1. I added an option to force rebuild as suggested by @zhassan-aws to overcome any instability that we may find with the cache. I also used this new option to overcome a somewhat buggy test. In the test `cargo-ui/supported-lib-types/lib-rlib`, we declare redundant types which generate a conflict name warning:
  ```
warning: output filename collision.                                                                                                        
The lib target `lib` in package `supported-lib v0.1.0 (lib-rlib)` has the same output filename as the lib target `lib` in package `supported-lib v0.1.0 (lib-rlib)`.                                                                   
Colliding filename is: lib-rlib/target/debug/d
eps/liblib-56b09ad964e3cddd.rlib                                                                                                           
The targets should have unique names.                                                                                                      
Consider changing their names to be unique or compiling them separately.                                                                   
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>. 
  ```
  while subsequent runs print the same warning and crash:
  ```
thread 'main' panicked at 'assertion failed: mtimes.insert(output.clone(), mtime).is_none()', src/tools/cargo/src/cargo/core/compiler/fingerprint.rs:1013:13
  ```
The same behavior can be reproduced using `cargo build`.

2. I created a few follow up issues to improve the caching mechanism: #2229, #2230, and #2231

### Testing:

### Checklist

edit: No longer delete the problematic test.


tests/script-based-pre/build-cache-bin/cache_works.expected
```diff
@@ -0,0 +1,25 @@
+Initial compilation
+target/initial.log:1
```
adpaco-aws, line 2: No idea where this number is printed from 😄

tests/script-based-pre/build-cache-bin/cache_works.sh
```diff
@@ -0,0 +1,67 @@
...
+# Move the original source to the output folder since it will be modified
+cp -r ${ORIG} ${OUT_DIR}
+
+echo "Initial compilation"
+check_kani --only-codegen initial.log
+
+echo "Re-execute the same command"
+check_kani --only-codegen same.log
+
+echo "Run with new arg that affects kani-driver workflow only"
```
adpaco-aws, line 55: You aren't checking that these cases do not trigger re-compilation, right? Or did I miss something?

tests/script-based-pre/build-cache-dirty/target_lib/Cargo.lock
```diff
@@ -0,0 +1,46 @@
+# This file is automatically @generated by Cargo.
```
adpaco-aws, line 1: Do we need to commit this file?

tests/script-based-pre/build-cache-dirty/target_lib/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! We don't use any of our dependency to keep the test fast
```
adpaco-aws, line 3: "dependencies"?

tests/script-based-pre/build-cache-bin/cache_works.sh
adpaco-aws: Oh wait, you actually do... You use "All fresh" for that case.

adpaco-aws (commented):
Thanks, @celinval !

The tests in this PR are looking great but I don't know where the first line comes from in, for example, this test:
```
target/same.log:0
target/same.log: All fresh
target/same.log: No harness
target/same.log: ok
```

Other than that, only minor comments.

tests/cargo-ui/supported-lib-types/lib-rlib/Cargo.toml
```diff
@@ -10,3 +10,10 @@ description = "Test that Kani correctly handle supported crate types"
 name = "lib"
 crate-type = ["lib", "rlib"]
 path = "../src/lib.rs"
+
+[package.metadata.kani.flags]
+# This test doesn't work with the cache due to naming conflict generated by
+# declaring ["lib", "rlib"] which is in fact redundant.
+# However, this still works for a fresh build and it only prints a warning.
```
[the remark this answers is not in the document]
celinval: No... this warning actually explicitly states that Cargo might turn the warning into error in the near future. So I highly doubt they will invest time to fix this issue.

tests/script-based-pre/build-cache-bin/cache_works.sh
```diff
@@ -0,0 +1,67 @@
...
+    if [ -z "${args}" ]
+    then
+        cargo kani --manifest-path "${MANIFEST}" --target-dir "${OUT_DIR}" \
+            2>&1 | tee "${log_file}"
+    else
+        cargo kani --manifest-path "${MANIFEST}" --target-dir "${OUT_DIR}" \
+            "${args}" 2>&1 | tee "${log_file}"
+    fi
+
+    # Check for occurrances of "Compiling" messages in the log files
```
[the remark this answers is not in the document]
celinval: oops

tests/script-based-pre/build-cache-bin/cache_works.expected
celinval: hahahaha.... it is printed from the "grep -c Compiling". I will improve the message, but it is printing the number of crates that were compiled. :)

tests/script-based-pre/build-cache-bin/cache_works.sh
celinval: And the number of compiled packages. Let me improve the message here.

tests/script-based-pre/build-cache-dirty/target_lib/Cargo.lock
celinval: I guess not

tests/script-based-pre/build-cache-dirty/rebuild.expected
```diff
@@ -0,0 +1,22 @@
+Initial compilation
+target/initial.log:5
+target/initial.log:Checking harness check_u8_i16...
+target/initial.log:Checking harness check_u8_u32...
+target/initial.log:Complete - 2 successfully verified harnesses, 0 failures, 2 total.
+Run with a new argument that affects compilation
+target/enable_checks.log:4
```
zhassan-aws, line 7: Why did the number go down by 1 from `initial.log`?

tests/script-based-pre/build-cache-fresh/cache_works.sh
```diff
@@ -0,0 +1,53 @@
...
+# The cases we cover here are:
+# - Exact same input being invoked 2x.
+# - Different options that do not influence the compilation only the Kani flow.
+# - Different options that do not influence the compilation only the CBMC execution.
+
+MANIFEST=lib/Cargo.toml
+OUT_DIR=target
+
+# Expects two arguments: "kani arguments" "output_file"
+function check_kani {
```
zhassan-aws, line 15: It might be better to put this function in a different file and source it in each of the scripts to avoid repetition.

zhassan-aws (commented):
Very cleverly crafted tests! A couple of comments.

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -81,6 +81,10 @@ impl CodegenBackend for GotocCodegenBackend {
         provide::provide_extern(providers);
     }
 
+    fn print_version(&self) {
```
zhassan-aws, line 84: This doesn't seem to be used anywhere.

kani-compiler/src/parser.rs
```diff
@@ -137,6 +131,12 @@ pub fn parser() -> Command {
                 .help("Instruct the compiler to perform stubbing.")
                 .requires(HARNESS)
                 .action(ArgAction::SetTrue),
+        )
+        .arg(
+            Arg::new("check-version")
```
zhassan-aws, line 136: Nit: should this be called `kani-version`?

kani-driver/src/call_single_file.rs
```diff
@@ -187,3 +187,11 @@ impl KaniSession {
 pub fn to_rustc_arg(kani_args: Vec<String>) -> String {
     format!(r#"-Cllvm-args={}"#, kani_args.join(" "))
 }
+
+/// Function that returns a `--check-version` argument to the compiler flags.
```
zhassan-aws, line 191:
```suggestion
/// Function that returns a `--check-version` argument to be added to the compiler flags.
```

zhassan-aws (commented):
A few more comments

tests/script-based-pre/build-cache-dirty/rebuild.expected
celinval: Sorry, I updated the test to use a crate with less dependencies. But that was a great question, so I reverted the test locally to double check. The reason is that one of the crates (`autocfg`) is a build-dependency. Here is the result of `cargo tree` for the original test:
```target_lib v0.1.0 (/kani/tests/script-based-pre/build-cache-dirty/target_lib)
└── num-bigint v0.4.3
    ├── num-integer v0.1.45
    │   └── num-traits v0.2.15
    │       [build-dependencies]
    │       └── autocfg v1.1.0
    │   [build-dependencies]
    │   └── autocfg v1.1.0
    └── num-traits v0.2.15 (*)
    [build-dependencies]
    └── autocfg v1.1.0
```

The cache for build dependencies doesn't get affected by the argument change, since the build dependency configuration uses host configuration.

celinval: And here is the full compilation log:
```
Run with a new argument that affects compilation
   Compiling num-traits v0.2.15
   Compiling num-integer v0.1.45
   Compiling num-bigint v0.4.3
   Compiling target_lib v0.1.0 (/kani/tests/script-based-pre/build-cache-dirty/target/target_lib)
```

tests/script-based-pre/build-cache-fresh/cache_works.sh
celinval: I totally agree, but this compiletest mode fails if it finds any subdirectory that is not a test. So to reuse this file, I would have to define the function inside one test folder and make all the other tests point to that folder.

So I decided to keep it simple and just use redundancy for now.

tests/script-based-pre/build-cache-dirty/rebuild.expected
zhassan-aws: Got it, thanks for clarifying.

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
celinval: In theory this is a function that the compiler uses to print the codegen version. Note that it is part of the implementation of the `CodegenBackend` trait.

There is a limitation in the compiler today where this doesn't get invoked for codegen backend that was configured via the driver, but I figured we might as well implement it and once this issue gets fixed, we get our version printed correctly.

kani-compiler/src/parser.rs
celinval: This is just a dummy argument, so I don't think it matters. I figured that we could eventually ensure that the kani-compiler matches the driver version, but I didn't bother here.

tests/script-based-pre/build-cache-fresh/cache_works.sh
zhassan-aws: OK

zhassan-aws (approved):
Thanks for all the hard work and numerous PRs to allow re-enabling the cache @celinval!

zhassan-aws:
Oh, one more thing: please make sure to add this to the `CHANGELOG` under "Major Changes" (perhaps in a follow-up PR) 🙂

PR #2246, opened 2023-02-28
Remove glob to collect all artifacts introduced by cargo kani

### LINKED ISSUE #1478: Extra Symtabs leads to conflicting proof harnesses
I tried this code: cargo-kani/vecdeque-cve/

./scripts/kani-regressions.sh

with Kani version: `git clone --branch 1478-bug-replication https://github.com/YoshikiTakashima/kani.git`

I expected to see this happen: This test case to pass.

Instead, this happened: The test case fails b/c spurious warnings
differ from expected output.

It seems he cause appears to be that passing extra symtabs in
`kani-driver/src/call_cargo.rs`. If that is removed, then the warnings go away.

### LINKED ISSUE #2234: Improve build cache handling that does not rely on Cargo build-plan feature
**Proposed change:** I don't have a great solution for this; thus, I decided to use the Cargo's [`--build-plan` feature](https://doc.rust-lang.org/cargo/reference/unstable.html#build-plan) for now. However, we need to figure out a more robust solution for the long term.

Ideally, we should use some API provided by Cargo. 

**Motivation:** Build plan seems to be deprecated (https://github.com/rust-lang/cargo/issues/7614), so we need a better solution ASAP. I posted a comment asking for some guidance: https://github.com/rust-lang/cargo/issues/7614#issuecomment-1444692932 

As I mentioned in the comment, I was hoping to use compiler artifact messages for that, but unfortunately the information provided is incomplete. I tried to extract the artifact name from the [artifact messages](https://doc.rust-lang.org/cargo/reference/external-tools.html?highlight=artifact#artifact-messages), but for things like binaries, static / shared libraries, the list of filenames do not include the original files that were generated by the compiler.

E.g.: For a static library, this is what I get:
```json
"filenames":["/tmp/test/static_lib/target/debug/libstatic_lib.a"]
```
while for a rust library, I get:
```json
"filenames":["/tmp/test/libA/target/debug/liblibA.rlib","/tmp/test/libA/target/debug/deps/liblibA-728fd0386e8e49d7.rmeta"]
```

On the other hand, `cargo build --build-plan` includes the name of the file:
```json
"outputs": ["/tmp/test/static_lib/target/debug/deps/libstatic_lib-edb6526891c85cb2.a" ],
```

celinval:
### Description of changes: 

This change allows us to pin-point exactly which build artifacts are related to a cargo build run. This is required to re-enable the build cache (#2036) where multiple artifacts can co-exist in the same folder.

The solution implemented here is rather hacky, but it's more reliable than other alternatives I've tried (see Call-outs). Now, `kani-compiler` will generate stubs in the place of binaries, shared, and static libraries when those types are requested. Those stubs will contain a JSON representation of the new type `CompilerArtifactStub`, which basically contain the path for the `metadata.json` file where the compiler stores the metadata related to a given crate.

`kani-driver` will parse `CompilerArtifact` messages to figure out which artifacts were built for the given build. For libraries, it will derive the name of the metadata file from the `rmeta` filepath. For other types of artifacts, it will parse the output file, convert it to `CompilerArtifactStub`, and extract the path for the `rmeta` filepath.

### Resolved issues:

Resolves #1478
Resolves #2234 

### Related RFC:

N/A

### Call-outs:

I'm not proud of this solution, and I am opened to suggestions. This is very hacky, but it is very similar to how MIRI solves this problem. MIRI also writes a JSON in the place of the binary that is meant to be generated by the compiler. This solution also works when all artifacts are fresh, since Cargo uses hardlink / copy to generate the final artifact, so the content of the file should point to the correct metadata file.

I investigated a few other options, but here is why they were not suitable:

1. Parse `CompilerArtifact` and derive the name of the metadata file from it. Unfortunately, for things like static libraries, the only artifact reported is different than the one generated by the compiler which does not include the file hash.
2. Emit a compiler message that `kani-driver` could parse and retrieve the name of the metadata. For this to work, we wouldn't be able to cache the final compilation artifacts, since we would need the compiler to be invoked again.
3. Use the deprecated `cargo build --build-plan`. Besides being deprecated, this option doesn't quite work for `cargo rustc`, and the results actually differed when we build with tests enabled. 
4. Implement our own cache by using different target directory. This would be overly complicated, it would require fine tuning and it would be more error prone.
5. Use a similar mechanism to the implemented one, and override the definition of `--print=file-names` to include `.json` to the artifact names. This would fix the main annoyance I have with the current solution which is to write plain text to file paths that are meant to be ELF. However, there is no easy to extend the compiler to do that, and this would require us to intercept the cargo call to the compiler for the call that gathers multiple information like file name, system configuration and others. IMO, not worth it.
6. Create dummy ELF files with some information about the build. We could either extract information from the ELF or even just compare the inode / file hash of the artifact reported and all the artifacts available in the build directory. This solution would likely be as reliable as the one we implemented, but it is also way more complicated.

For more context, see this Zulip thread I created: https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Files.20generated.20by.20the.20compiler

### Testing:

* How is this change tested? Existing tests + manually tested together with #2232

### Checklist


kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -215,34 +216,48 @@ impl CodegenBackend for GotocCodegenBackend {
     /// Types such as `bin`, `cdylib`, `dylib` will trigger the native linker.
     ///
     /// Thus, we manually build the rlib file including only the `rmeta` file.
+    ///
+    /// For cases where no metadata file was requested, we stub the file requested by writing the
+    /// path of the `metadata.json` file so `kani-driver` can safely find the latest metadata.
```
zhassan-aws, line 221: Too many metadatas :) So just clarify this a bit:
```suggestion
    /// path of the `kani-metadata.json` file so `kani-driver` can safely find the latest metadata.
```

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -215,34 +216,48 @@ impl CodegenBackend for GotocCodegenBackend {
...
+                let (metadata, _metadata_position) = create_wrapper_file(
+                    sess,
+                    b".rmeta".to_vec(),
+                    codegen_results.metadata.raw_data(),
+                );
+                let metadata = emit_wrapper_file(sess, &metadata, &path, METADATA_FILENAME);
+                builder.add_file(&metadata);
+                builder.build(&out_path);
+            } else {
+                // Write the location of the metadata file to the requested type.
```
zhassan-aws, line 252:
```suggestion
                // Write the location of the kani metadata file in a file of the requested type.
```

kani-driver/src/call_cargo.rs
```diff
@@ -196,9 +190,10 @@ impl KaniSession {
                             }
                         }
                     },
-                    Message::CompilerArtifact(_)
-                    | Message::BuildScriptExecuted(_)
-                    | Message::BuildFinished(_) => {
+                    Message::CompilerArtifact(rustc_artifact) => {
+                        artifact = Some(rustc_artifact);
```
zhassan-aws, line 194: Is it guaranteed that there's a single artifact per cargo invocation? If so, I suggest adding an assertion that checks that, e.g:
```rust
assert_eq!(artifact, None);
```

kani-driver/src/call_cargo.rs
```diff
@@ -236,15 +231,6 @@ fn print_msg(diagnostic: &Diagnostic, use_rendered: bool) -> Result<()> {
     Ok(())
 }
 
-/// Given a `path` with glob characters in it (e.g. `*.json`), return a vector of matching files
-fn glob(path: &Path) -> Result<Vec<PathBuf>> {
```
zhassan-aws, line 240: Woohoo!

zhassan-aws (commented):
Thanks for figuring out a solution @celinval! Just have a few comments.

kani-driver/src/call_cargo.rs
celinval: Not really. Every package will generate at least one artifact. However, we only care about the last one, which is the one that we pass the codegen flag to. Thus, I am only running the map in the last artifact that was collected.

zhassan-aws: I see. Perhaps add a comment to mention this?

celinval: Done

zhassan-aws (approved):

celinval:
I was finally able to debug this flaky issue!! It seems that we are getting the artifacts out of order. In [this log](https://pipelines.actions.githubusercontent.com/serviceHosts/14dcd0bf-cf10-416a-af82-d1e50ea29e0a/_apis/pipelines/1/runs/1631/signedlogcontent/4?urlExpires=2023-03-03T01%3A58%3A30.1121353Z&urlSigningMethod=HMACV1&urlSignature=tyo%2Bjqq39undGhuzDIAEhqnL3H8l6hsLXSOVDaneTrM%3D), it seems that we are using the `librand` artifact:
```
2023-03-03T01:45:58.1861520Z TRACE kani_driver::call_cargo map_kani_artifact, rmeta="/Users/runner/work/kani-dev/kani-dev/build/tests/cargo-kani/dependencies/check_dummy/target/kani/x86_64-apple-darwin/debug/deps/librand-51ad98e9ae258aeb.rmeta", kani_meta="/Users/runner/work/kani-dev/kani-dev/build/tests/cargo-kani/dependencies/check_dummy/target/kani/x86_64-apple-darwin/debug/deps/rand-51ad98e9ae258aeb.kani-metadata.json
```
instead of the `foo`
```
2023-03-03T01:45:58.1854750Z DEBUG kani_compiler::codegen_cprover_gotoc::compiler_interface write_json, filename="/Users/runner/work/kani-dev/kani-dev/build/tests/cargo-kani/dependencies/check_dummy/target/kani/x86_64-apple-darwin/debug/deps/foo-c9378ac126d6db94.kani-metadata.json"
```

zhassan-aws (approved):
Thanks.

PR #2257, opened 2023-03-02
Remove unused msg parameter from any_where

zhassan-aws:
### Description of changes: 

The `_msg` parameter of `kani::any_where` has no effect and does not appear in the Kani output. Removing it to make the usage of `any_where` more ergonomic.

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

This is a *breaking* change, and needs to be called out in the release notes.

### Testing:

### Checklist

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,21 @@
+# Changelog
+
+This file contains notable changes (e.g. breaking changes, major changes, etc.) in Kani releases.
+
+This file was introduced starting Kani 0.23.0, so it only contains changes from version 0.23.0 onwards.
+
+## [0.23.0]
+
+### **Breaking Changes**
+
+- The second parameter in the `kani::any_where` function (`_msg: &'static str`) was removed to make the function more ergonomic to use.
+We suggest moving the explanation for why the assumption is introduced into a comment.
+For example, the following code:
+```rust
+    let len: usize = kani::any_where(|x| *x < 5, "Restrict the length to a value less than 5");
+```
+should be replaced by:
+```rust
+    // Restrict the length to a value less than 5
+    let len: usize = kani::any_where(|x| *x < 5);
+```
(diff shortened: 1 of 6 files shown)


library/kani/src/lib.rs
```diff
@@ -120,15 +120,15 @@ pub fn any<T: Arbitrary>() -> T {
 /// under all possible `NonZeroU8` input values between 0 and 12.
 ///
 /// ```rust
-/// let inputA = kani::any_where::<std::num::NonZeroU8>(|x| *x < 12, "explanation");
+/// let inputA = kani::any_where::<std::num::NonZeroU8>(|x| *x < 12);
```
adpaco-aws, line 123: Can you please remove the line `/// The explanation field gives a mechanism to explain why the assumption is required for the proof.` that appears above?

adpaco-aws (approved):
Thanks, @zhassan-aws !

library/kani/src/lib.rs
zhassan-aws: Good catch @adpaco-aws! Removed.

karkhaz (approved):

CHANGELOG
```diff
@@ -0,0 +1,21 @@
+# Changelog
```
celinval, line 1: Nice!

celinval (approved):
Quick question... do we want to add this file to our installation package?

CHANGELOG
zhassan-aws: Good point! I'll add it.

zhassan-aws: Done.

celinval:
I would add it here instead:

https://github.com/model-checking/kani/blob/e53ed6389223c693837e41fdc305ecd1937b6e57/Cargo.toml#L17

zhassan-aws:
> I would add it here instead:

That makes more sense. Done.

celinval (approved):
Thanks @zhassan-aws

PR #2272, opened 2023-03-06
RFC: The `kani::should_panic` attribute

adpaco-aws:
### Description of changes: 

Users may want to express that a verification harness should panic.
This RFC proposes a new harness attribute `#[kani::should_panic]` that informs Kani about this expectation.

Rendered preview [here](https://github.com/adpaco-aws/rmc/blob/rfc-should-fail/rfc/src/rfcs/0005-should-fail-attr.md).

### Resolved issues:

Related #600 

### Testing:

### Checklist


rfc/src/rfcs/0005-should-fail-attr.md
```diff
@@ -0,0 +1,197 @@
...
+Complete - 0 successfully verified harnesses, 3 failures, 3 total.
+```
+
+Harnesses marked with `#[kani::should_fail]` won't show unless the expected result was different from the actual result.
+The summary will consider harnesses that match their expectation as "successfully verified harnesses".
+
+Therefore, if we added `#[kani::should_fail]` to all harnesses in the previous example, we'd see this output:
+
+```
+Complete - 3 successfully verified harnesses, 0 failures, 3 total.
```
celinval, line 104: Doesn't this conflict with the detailed representation?

rfc/src/rfcs/0005-should-fail-attr.md
```diff
@@ -0,0 +1,197 @@
...
+ 1. **(Expectation)** Kani was expecting the harness to fail.
+ 2. **(Outcome)**: The actual verification result that Kani produced after the analysis.
+This will avoid a potential scenario where the user doesn't know for sure if the attribute has had an effect when running the verification harness.
+
+As mentioned, we've considered two ways to represent this result.
+
+#### Representation #1 (Recommended): No changes to overall result
+
+```rust
+VERIFICATION:- FAILED (expected FAILED)
```
celinval, line 76: I'm so glad you are tackling this here, but I wonder if we can still improve this. Double negation is hard to parse.

rfc/src/rfcs/0005-should-fail-attr.md
```diff
@@ -0,0 +1,197 @@
...
+Therefore, if we added `#[kani::should_fail]` to all harnesses in the previous example, we'd see this output:
+
+```
+Complete - 3 successfully verified harnesses, 0 failures, 3 total.
+```
+
+### Availability
+
+This feature will only be available as an attribute.
+That means this feature won't be available as a CLI option (i.e., `--should-fail`).
```
celinval, line 110: +1.

rfc/src/rfcs/0005-should-fail-attr.md
```diff
@@ -0,0 +1,197 @@
...
+To be honest, we avoid `expect` altogether for two reasons:
+ - We may consider a more granular approach to indicate expectations regarding individual checks and cover statements in the future. One possible name for the attribute is `#[kani::expect]`.
+ - We heavily use this word for testing in Kani: there is an `expected` mode, which works with `*.expected` files. Other modes also use such files.
+
+### Alternative #2: Granularity
+
+As mentioned earlier, users may want to express more granular expectations for their harnesses.
+For example, a user may want to specify that a given check results in a failure.
+
+Again, there is a relation between this and the [the `#[should_panic]` attribute](https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#testing-panics) for Rust tests.
```
celinval, line 159: Did you consider adding `#[should_panic]` instead of `#[should_fail]` for now. I am trying to see what is the motivation for users to knowingly leave code that triggers UB.

rfc/src/rfcs/0005-should-fail-attr.md
```diff
@@ -0,0 +1,197 @@
...
+    fn test_specific_panic() {
+        divide_non_zero_result(1, 10);
+    }
+```
+
+In principle, it wouldn't be a problem to extend this proposal to include the `expected` argument.
+The implementation could compare the `expected` string against property descriptions.
+
+However, there could be some problems later with this approach:
+ * What if users want to specify multiple `expected` strings?
```
celinval, line 174: could users have multiple annotations?

rfc/src/rfcs/0005-should-fail-attr.md
```diff
@@ -0,0 +1,197 @@
...
+
+We don't know if these will be requirements in the future.
+If they were, then we'd need to come up with language to state expectations about properties (checks or cover statements).
+A good option to consider here is adding the single-string `expected` argument (as in `#[should_panic]`) and see how it's used.
+It could be the cause of a breaking change in the future, but it would provide us with useful data in the meatime.
+
+## Open questions
+
+The main question I'd like to get resolved are:
+ * Do we want to extend `#[kani::should_fail]` with an `expected` field?
```
celinval, line 186: I'm worried that `should_fail` is too open since it includes any UB.

rfc/src/rfcs/0005-should-fail-attr.md
adpaco-aws: Another good point, I'll add that to this list.

rfc/src/rfcs/0005-should-fail-attr.md
adpaco-aws: This was answered in my long comment.

rfc/src/rfcs/0005-should-fail-attr.md
adpaco-aws: No, this shows up for both the default and terse outputs.

rfc/src/rfcs/0005-should-fail-attr.md
adpaco-aws: True, it's also not clear to me how this result will look in #1690. I need to spend more time on this one.

adpaco-aws: Couldn't find a simpler representation. It's now a question in the "Open questions", and I'll explicitly look for feedback on this.

rfc/src/rfcs/0005-should-fail-attr.md
adpaco-aws: The RFC now promotes a discussion around multiple panics, and suggests the panic expectation be limited to one. It isn't clear to me that we want to allow multiple panics with this attribute, but there's an open question related to this aspect.

rfc/src/rfcs/0005-should-fail-attr.md
```diff
@@ -0,0 +1,250 @@
...
+ - It'd only be useful when combined with `--harness` to filter negative harnesses.
+ - We could have trouble extending its functionality (see [Future possibilities](#future-possibilities) for more details).
+
+### Pedagogy
+
+The `#[kani::should_panic]` attribute will become one of the most basic attributes in Kani.
+As such, it'll be mentioned in the tutorial and added to the dedicated section planned in [#2208](https://github.com/model-checking/kani/issues/2208).
+
+In general, **we'll also advise against negative verification** when a harness can be written both as a regular (positive) harness and a negative one.
+The feature, as it's presented in this proposal, won't allow checking that the panic failure is due to the panic we expected.
```
karkhaz, line 146: This seems like a big deal TBH. From basically the first example, I was wondering "what happens if a new unintentional `panic` later gets inserted earlier in the code?"

Is there a way to 'fix' this? Instead of an attribute, could we have a magic comment (or some Rusty equivalent) directly above the line that is expected to panic, instead of the whole harness?

rfc/src/rfcs/0005-should-fail-attr.md
```diff
@@ -0,0 +1,250 @@
...
+}
+```
+
+This is what a negative harness may look like.
+The user wants to verify that calling `device.init()` more than once should result in a panic.
+
+> **NOTE**: We could convert this into a Rust unit test and add the `#[should_panic]` attribute to it.
+> However, there are two good reasons to have a verification-specific attribute that does the same:
+>  1. To ensure that other unexpected behaviors don't occur (e.g., overflows).
+>  2. Because `#[should_panic]` cannot be used if the test harness contains calls to Kani's API.
```
robdockins, line 72: A third potential case is that the user has used `cfg` attributes to stub out part of the code, and wants to be sure the expected panic occurs both in the code as exposed to Kani and in the "normal" code.

In the case I'm concerned with, I used stubs to implement the low-level concurrency primitives that are the basis of the library, so I was wanting to be sure the code with stubs also behaved as expected.

adpaco-aws: Thanks @robdockins , added the scenario you describe as a third reason.

rfc/src/rfcs/0005-should-fail-attr.md
adpaco-aws: The main problem with the approach you mention (and similar ones like the "function that takes a closure", which is now mentioned as an alternative) is that you can't use them on code that's external to your packages.

Moreover, it seems to me that users just want a coarse mechanism to check that a panic is reachable. It's possible that users want a more fine-grained mechanism, as noted in the RFC, but we simply can't get much data on that unless we have the coarse one first.

adpaco-aws:
Addressed comments on the PR and from our discussion yesterday:
 * Moved discussion about multiple panics to UX section, and extended it significantly.
 * Provide answers for all questions that were open, and move them to a subsection "resolved questions".
 * Add new alternative to mention the Kani-API-that-takes-closure approach.
 * Modified representations, and selected the second one as the recommended.
 * Added third motivating reason for the attribute.

rfc/src/rfcs/0005-should-fail-attr.md
```diff
@@ -88,24 +88,24 @@ Since we added `#[kani::should_panic]`, running this example would produce a suc
...
 ```
 
 We recommend this representation so the user receives clear information about both the outcome and the expectation.[^footnote-representation]
 
-#### Representation #2: Changes to overall result
+#### Representation #2 (Recommended): Changes to overall result
 
 ```rust
-VERIFICATION:- SUCCESSFUL (expected FAILED and it FAILED)
+VERIFICATION:- SUCCESSFUL (expected and verified one panic at least)
```
zhassan-aws, line 111: I would avoid associating "verified" with "panic". Maybe say `encountered one or more panics as expected`?

zhassan-aws (approved):
Looks good. Just one comment.

rfc/src/rfcs/0005-should-fail-attr.md
adpaco-aws: Thanks! I've changed it and extended this section with data from the actual implementation (see #2315)

karkhaz (approved):

Issue #2279, opened 2023-03-07
Create mechanism to enable / disable APIs


celinval:
**Proposed change:** Add a mechanism for us to mark an API as unstable, and to ensure that users explicit enable them.

**Motivation:** We want to enable experiments and fast development for our libraries the same way we do for our software without affecting users trust. An experimental API may bring instability and unexpected results, so users should be aware of that when they choose to enable them.

The initial use case would be async APIs introduced in https://github.com/model-checking/kani/pull/1659.

PR #2281, opened 2023-03-08
RFC: Unstable APIs

celinval:
### Description of changes: 

Create a RFC for feature gate for APIs included in the Kani library.

### Related RFC:

Optional #2279 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


adpaco-aws (commented):
Thanks for writing this, @celinval !

Overall, I agree with the proposal. I've left some comments regarding motivation and other procedures (e.g., a warning) that I'd like you to consider.

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,98 @@
+- **Feature Name:** Unstable APIs
```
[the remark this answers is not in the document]
celinval: indeed... thanks

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,98 @@
...
+## User Impact
+
+The impact should be minimal, and it is very similar to the current experience when they are trying an experimental feature in Kani.
+ The unstable API will be given a unique identifier, which the user will enable either via command line or via command argument.
+
+## User Experience
+
+Users will have to invoke Kani with:
+```
+--enable-unstable --unstable-feature <feature_identifier>
```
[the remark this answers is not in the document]
celinval: Yeah, I agree that this is rather verbose. The problem with `--enable-unstable` today is that it doesn't specify which unstable feature it enables. We could add arguments to it, but it would be a breaking change. We could use `-Z` like the compiler and cargo to specify the actual feature.

Yes, we do need one for each feature identifier. Handling multiple identifiers is not recommended since we accept positional arguments.

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,98 @@
...
+In order to mark an API as unstable, we will add the following attribute to the APIs marked as unstable:
+
+```rust
+#[kani::unstable(feature="<IDENTIFIER>", issue="<TRACKING_ISSUE_NUMBER>", reason="<OPTIONAL_DESCRIPTION>")]
+pub fn unstable_api() {}
+```
+
+This is similar to the interface used by [the standard library](https://rustc-dev-guide.rust-lang.org/stability.html#unstable).
+
+If the user tries to run Kani without enabling an unstable feature that is reachable,
```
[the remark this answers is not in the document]
celinval: Yes, I am planning to print a warning like the one we have for unsupported features. I'll make it clear here. It will be a new warning. I prefer not to mix the warnings though.

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,98 @@
...
+  bar()
+}
+```
+
+The compiler will replace the body of the API with a reachability check if the feature `foo_bar` was
+not enabled. I.e., this will generate the function body similar to running:
+
+```rust
+pub fn unstable_foo<T>() -> T {
+  kani::unstable("unstable_foo", reason, issue")
```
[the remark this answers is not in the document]
celinval: oops... it should, since it is a syntax error. 🙂

rfc/src/rfcs/0006-unstable-api.md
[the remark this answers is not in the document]
zhassan-aws: I would prefer if we make this an error if it's possible (e.g. if it's reachable as determined by the MIR linker).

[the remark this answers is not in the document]
celinval: Yeah, we can totally do that... @adpaco-aws, what do you think?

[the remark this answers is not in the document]
adpaco-aws: Makes sense to me.

One case where we don't error at the moment (but for some reason, my intuition says we should) is when we run a harness annotated with `#[kani::stub(...,...)]` but we don't pass `--enable-stubbing`. But I don't have a good sense about how many "spurious errors" will be caused by emitting an error here. I think it'd be OK to start with errors and be prepared to downgrade to warnings if errors become too annoying.

rfc/src/rfcs/0006-unstable-api.md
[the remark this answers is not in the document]
adpaco-aws: I like using `-Z`, and it wouldn't be a bad time to remove `--enable-unstable` if we decided to use `-Z<unstable_feature_ID>` for all unstable features.

rfc/src/rfcs/0006-unstable-api.md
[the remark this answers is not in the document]
celinval: Sounds good... let me change the proposal and add the downgrade as future work. This actually makes my problem easier to solve. =)

rfc/src/rfcs/0006-unstable-api.md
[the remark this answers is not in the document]
adpaco-aws: Yes, that'd be great. Another benefit is that we could follow the [same rules they use](https://doc.rust-lang.org/cargo/reference/unstable.html) to require `-Z` for unstable options without an interface.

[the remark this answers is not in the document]
celinval: Yes, I was thinking about adding a `[unstable]` table under `kani`.

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,98 @@
...
+```rust
+#[kani::unstable(feature="<IDENTIFIER>", issue="<TRACKING_ISSUE_NUMBER>", reason="<OPTIONAL_DESCRIPTION>")]
+pub fn unstable_api() {}
+```
+
+This is similar to the interface used by [the standard library](https://rustc-dev-guide.rust-lang.org/stability.html#unstable).
+
+If the user tries to run Kani without enabling an unstable feature that is reachable,
+the verification may fail if the unstable feature is reachable.
+I.e., the behavior will be similar to unsupported features.
```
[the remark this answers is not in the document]
celinval: I removed this since we decided to treat this situation as a compilation error.

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
+- **Feature Name:** Unstable APIs (unstable-api)
```
adpaco-aws, line 1:
```suggestion
- **Feature Name:** Unstable APIs (`unstable-api`)
```

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
...
+
+```rust
+#[kani::unstable(feature="<IDENTIFIER>", issue="<TRACKING_ISSUE_NUMBER>", reason="<OPTIONAL_DESCRIPTION>")]
+pub fn unstable_api() {}
+```
+
+This is similar to the interface used by [the standard library](https://rustc-dev-guide.rust-lang.org/stability.html#unstable).
+
+If the user tries to use an unstable feature in Kani without explicitly enabling it,
+Kani will trigger an error. For unstable APIs, the error will be trigger during the crate
```
adpaco-aws, line 65: "will be triggered"

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
...
+Add an opt-in model for users to try experimental APIs.
+The goal is to enable users to try features that aren't stable yet,
+which allow us to get valuable feedback during the development of new features and APIs.
+
+The opt-in model empower the users to control when some instability is acceptable,
+which makes Kani UX more consistent and safe.
+
+In order to reduce friction, we will also standardize how users opt-in to any Kani unstable feature.
+We will use similar syntax to the one used by the Rust compiler and Cargo.
+As part of this work, we will also deprecate and remove `--enable-unstable` option.
```
adpaco-aws, line 25: The section doesn't explain the current approach to unstable options. I think it's worth explaining here for commenters who aren't familiar with it. That'll also make clearer what the benefits of this proposal are (UX standardization, an unstable feature catalog, etc.).

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
...
+## Detailed Design
+
+We will add the `-Z` option to both `kani-driver` and `kani-compiler`.
+Kani driver will pass the information to the compiler.
+
+For unstable APIs, the compiler will check if any reachable function uses a unstable feature that was not enabled.
+If that is the case, the compiler will trigger a compilation error.
+
+We will also change the compiler to only generate code for harnesses that match the harness filter.
+The filter is already passed to the compiler, but it is currently only used for stubbing.
```
adpaco-aws, line 77: It's not clear to me why this change is needed.

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
...
+whether it is stabilized or removed.
+In those cases, users would have to edit each harness that enables the affected feature.
+
+## Open questions
+
+
+## Future possibilities
+
+- Delay the error due to the usage of a unstable API, and only fail at runtime if the API is reachable.
+- Allow users to enable unstable features on a per-harness base.
```
adpaco-aws, line 110: "base" -> "basis"?

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
...
+
+### API Removal
+
+If we decide to remove an API that is marked as unstable, we should follow a regular deprecation
+path (using `#[deprecated]` attribute), and keep the `unstable` flag + attributes, until we are
+ready to remove the feature completely.
+
+## Rational and Alternatives
+
+For this RFC, the suggestion is to only enable experimental features globally for simplicity of use and implementation.
```
adpaco-aws, line 93: +1

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
...
+
+In order to reduce friction, we will also standardize how users opt-in to any Kani unstable feature.
+We will use similar syntax to the one used by the Rust compiler and Cargo.
+As part of this work, we will also deprecate and remove `--enable-unstable` option.
+
+Note that although Kani is still on v0, which means that everything is somewhat unstable,
+this allow us to set different bars when it comes to what kind of changes is expected,
+as well as what kind of support we will provide for a feature.
+
+## User Experience
```
adpaco-aws, line 31: Really liked how this section is written!

adpaco-aws (commented):
It's looking great!

I'd only extend the "User Impact" section with an explanation on the current approach to unstable options, briefly mentioning its limitations.

rfc/src/rfcs/0006-unstable-api.md
celinval: Let's say you have a harness "A" that uses a unstable feature and a harness "B" that doesn't. Should the following command fail?
```
kani --harness B input.rs
```

zhassan-aws: I think this will be useful for other things as well, e.g. reducing warnings due to unsupported constructs that are not reachable from the selected harnesses.

zhassan-aws (approved):
Looks good. Thanks @celinval!

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
...
+Provide a standard option for users to enable experimental APIs and features in Kani,
+and ensure that those APIs are off by default.
+
+## User Impact
+
+Add an opt-in model for users to try experimental APIs.
+The goal is to enable users to try features that aren't stable yet,
+which allow us to get valuable feedback during the development of new features and APIs.
+
+The opt-in model empower the users to control when some instability is acceptable,
```
fzaiser, line 20: empowers

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
...
+If the user tries to use an unstable feature in Kani without explicitly enabling it,
+Kani will trigger an error. For unstable APIs, the error will be trigger during the crate
+compilation.
+
+## Detailed Design
+
+We will add the `-Z` option to both `kani-driver` and `kani-compiler`.
+Kani driver will pass the information to the compiler.
+
+For unstable APIs, the compiler will check if any reachable function uses a unstable feature that was not enabled.
```
fzaiser, line 73: an unstable

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
...
+If that is the case, the compiler will trigger a compilation error.
+
+We will also change the compiler to only generate code for harnesses that match the harness filter.
+The filter is already passed to the compiler, but it is currently only used for stubbing.
+
+### API Stabilization
+
+Once an API has been stabilized, we will remove the `unstable` attributes from the given API.
```
fzaiser, line 81: I think the Rust library adds an attribute to stabilized functions with the version number where it was stabilized. Would this make sense for Kani too?

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,110 @@
...
+If we decide to remove an API that is marked as unstable, we should follow a regular deprecation
+path (using `#[deprecated]` attribute), and keep the `unstable` flag + attributes, until we are
+ready to remove the feature completely.
+
+## Rational and Alternatives
+
+For this RFC, the suggestion is to only enable experimental features globally for simplicity of use and implementation.
+
+For now, we will trigger a compilation error if an unstable API is reachable from a user crate
+unless if the user opt in for the unstable feature.
```
fzaiser, line 96: unless the user opts in to

fzaiser:
Thanks @celinval! I took a look as well as my PR #1659 will be the first API to be affected by this. It sounds good to me, I only added some minor comments (mostly typos).

fzaiser:
Could I ask for an update given that #1659 is blocked on this?

rfc/src/rfcs/0006-unstable-api.md
celinval: That's a great suggestion... we could do something similar, but we cannot use the exact same attribute as rust. We will likely need to consider something similar when / if we start shipping Kani library separate from our binaries.

celinval:
> Thanks @celinval! I took a look as well as my PR #1659 will be the first API to be affected by this. It sounds good to me, I only added some minor comments (mostly typos).

That's awesome! Thanks for the feedback. Sorry for the late reply, I was out for a couple of weeks.

rfc/src/rfcs/0006-unstable-api.md
```diff
@@ -0,0 +1,119 @@
...
+The opt-in model empowers the users to control when some instability is acceptable,
+which makes Kani UX more consistent and safe.
+
+Currently, each new unstable feature will introduce a new switch, some of them will look like `--enable-<feature>`,
+while others will be a plain switch which allows further feature configuration `--<feature-config>=[value]`.
+For example, we today have the following unstable switches `--enable-stubbing`, `--concrete-playback`, `--gen-c`.
+In all cases, users are still required to provide the additional `--enable-unstable` option.
+Some unstable features are included in the `--help` section, and only a few mention the requirement
+to include `--enable-unstable`. There is no way to list all unstable features.
+The the transition to stable switches is also ad-hoc.
```
adpaco-aws, line 29:
```suggestion
The transition to stable switches is also ad-hoc.
```

adpaco-aws (approved):
Thank you, @celinval !

rfc/src/rfcs/0006-unstable-api.md
celinval: BTW, I added it to the open questions so we can explore the idea as part of the implementation.

PR #2297, opened 2023-03-10
Add "kani" configuration key to enable conditional compilation in build scripts

celinval:
### Description of changes: 

We now pass `--cfg=kani` to build scripts, which allow users to use constructs such as `if cfg!(kani)` to conditionally compile their build scripts.

The build script may have logic that is not redundant to Kani, or even unsupported. Users can now change how their build works based on conditional compilation.

### Resolved issues:

N/A

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

### Checklist


celinval:
There is one issue with this change is that we require nightly cargo since this is a unstable feature. However, we invoke whatever cargo is in the system. I.e.: I'll need to change our code to invoke the cargo we install.

karkhaz (approved):

karkhaz (approved):

PR #2313, opened 2023-03-23
Add scripts to automate dependencies

jaisnan:
### Description of changes: 

These scripts help automate part of the dependency scripts. While initially tested for manual use, it can be developed and used as part of the CD.

### Call-outs:

These scripts only automate the part of the deployment that deal with updating dependencies and updating version numbers. If approved, the rest of the deployment can be added easily (the PR for that is ready to be added).

### Testing:

* How is this change tested? Manually

### Checklist

### DIFF
--- a/scripts/ci/preflight-steps/check-outdated.py
+++ b/scripts/ci/preflight-steps/check-outdated.py
@@ -6,7 +6,13 @@ import subprocess
 def run_cargo_command(command, working_directory):
 
     try:
-        process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True, text=True, cwd=working_directory)
+        process = subprocess.Popen(
+            command,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            shell=True,
+            text=True,
+            cwd=working_directory)
         stdout, stderr = process.communicate()
 
         if process.returncode != 0:
@@ -24,6 +30,7 @@ def main():
     target_directory = "kani"
     run_cargo_command("cargo install cargo-outdated", working_directory)
     run_cargo_command("cargo outdated --workspace", working_directory)
+
 
 if __name__ == "__main__":
     main()
--- a/scripts/ci/preflight-steps/check-outdated.py
+++ b/scripts/ci/preflight-steps/check-outdated.py
@@ -1,0 +1,29 @@
+#!/usr/bin/python3
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+import subprocess
+
+def run_cargo_command(command, working_directory):
+
+    try:
+        process = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True, text=True, cwd=working_directory)
+        stdout, stderr = process.communicate()
+
+        if process.returncode != 0:
+            print(f"Command '{command}' executed unsuccesfully, in '{working_directory}'.")
+            print(stderr)
+        else:
+            print(f"Command '{command}' executed succesfully, in '{working_directory}'.")
+            print(stdout)
+    except Exception as e:
+        print(f"Error: {str(e)}")
+
+
+def main():
+
+    target_directory = "kani"
+    run_cargo_command("cargo install cargo-outdated", working_directory)
+    run_cargo_command("cargo outdated --workspace", working_directory)
+
+if __name__ == "__main__":
+    main()
--- a/scripts/ci/preflight-steps/dependency_links.py
+++ b/scripts/ci/preflight-steps/dependency_links.py
@@ -1,0 +1,23 @@
+#!/usr/bin/python3
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+dependencies_links = {
+    'cbmc': {
+        'dependency_name': 'cbmc',
+        'dependency_string': 'CBMC_VERSION',
+        'org_name': 'diffblue',
+        'link': 'https://github.com/diffblue/cbmc'
+    },
+    'cbmc_viewer': {
+        'dependency_name': 'cbmc-viewer',
+        'dependency_string': 'CBMC_VIEWER_VERSION',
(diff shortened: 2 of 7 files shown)


scripts/ci/preflight-steps/dependency_updater.py
```diff
@@ -0,0 +1,58 @@
...
+        with open(self.file, 'w') as f:
+            for line in lines:
+                match = re.search(r'(\w+)="\d+(\.\d+)+"', line.strip())
+                if match and match.group(1).strip() == self.dependency["dependency_string"]:
+                    f.write(f'{self.dependency["dependency_string"]}="{self.dependencies[self.dependency["dependency_string"]]}"\n')
+                else:
+                    f.write(line)
+
+
+    def run_process(self):
```
karkhaz, line 55: It's not clear to me that you need to walk through the file, updating the existing values. It seems much quicker to overwrite the entire file.

If you have a dict like this:
```pr
new_versions = {
  "CBMC_VIEWER": 3.9,
  "CBMC": 5.99.0,
  ...
}
```
then I think it would be much clearer to do this:
```
with open("kani-dependencies", "w") as handle:
    for tool, version in new_versions.items():
        print(f"{tool}={version}", file=handle)
```
then you don't need the if-statement on line 27, which hard-codes the tool names and will thus need to be updated whenever we add a new dependency.

so my suggestion is:
1. Get the list of existing tools by opening `kani-dependencies`, read each line, split on `=`, return the first part
2. Turn that list of tools into the `new versions` dict, using a variation of your `get_latest_version` method
3. Write the dict out using the 3 lines of code above, no need to update the existing file, just overwrite it.

I think this would make the code a lot shorter and clearer, and avoid hardcoding the tool names here. `kani-dependencies` is the only place those names should be written.

scripts/ci/preflight-steps/dependency_links.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/python3
```
karkhaz, line 1: I'm going to propose that we don't duplicate this information, see [this](https://github.com/model-checking/kani/pull/2314/files#r1146533803)

scripts/ci/preflight-steps/check-outdated.py
```diff
@@ -0,0 +1,36 @@
...
+            cwd=working_directory)
+        stdout, stderr = process.communicate()
+
+        if process.returncode != 0:
+            print(f"Command '{command}' executed unsuccesfully, in '{working_directory}'.")
+            print(stderr)
+        else:
+            print(f"Command '{command}' executed succesfully, in '{working_directory}'.")
+            print(stdout)
+    except Exception as e:
```
karkhaz, line 24: Not sure that there's an need for try-except here. If `Popen` raises, I think the program should crash out early.

There's also no need to use `shell=True` here, please pass the command as a list.

More generally I'm not sure that there's any need to buffer the output and then immediately print it. I think this code could be a lot shorter:

```python
cmds = [
  ["cargo", "install", "cargo-outdated"],
  ["cargo", "outdated", "--workspace"],
]
exit_code = 0
for cmd in cmds:
    proc = subprocess.run(cmd, cwd=working_directory)
    exit_code += proc.returncode
sys.exit(exit_code)
```

scripts/ci/preflight-steps/update-cbmc.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/python3
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+from job_runner import dependencies_links
+from dependency_updater import VersionUpdater
+
+# Set the name of the dependency you want to update
+kani_dependencies_path = "/Users/jaisnan/kani/kani-dependencies"
```
karkhaz, line 8: This shouldn't be using the full path, just `kani-dependencies` should be fine

karkhaz (changes requested):
Apart from all the specific comments:

1. Please run these files through a linter like pylint or pep-8
2. Would you be able to write a top-level script that runs all of these? It's not clear to me how they should be run, or on what order. Some of my comments suggest that the small `update-$TOOL.py` duplicate a lot of code, I honestly think this whole PR could fit in a single file
3. I think we should avoid duplicating data as much as possible, in particular the names of dependencies that are written elsewhere in the source tree. Ideally this code should be iterating over a JSON file or other file that contains the list of dependencies. The tool names should not be hardcoded anywhere in this PR. I'm working on making that possible [here](https://github.com/model-checking/kani/pull/2314/files#r1146533803)

zhassan-aws:
Another option to consider for updating Rust dependencies is [dependabot](https://github.com/dependabot). It automatically opens a PR when a dependency has a new version. For example, s2n-quic uses it, and yesterday it [opened a PR ](https://github.com/aws/s2n-quic/pull/1674) to upgrade Kani to 0.24.

This is an example of how it can be integrated: https://github.com/model-checking/kani/pull/1909

The annoying thing at the moment is that it would open one PR for every version update as opposed to batching all of them together. There's an [open feature request ](https://github.com/dependabot/dependabot-core/issues/1190) to add support for grouped updates.

scripts/ci/preflight-steps/check-outdated.py
```diff
@@ -0,0 +1,36 @@
...
+        print(f"Error: {str(e)}")
+
+
+def main():
+
+    target_directory = "kani"
```
zhassan-aws, line 30: Should this be `working_directory`?

jaisnan:
> Another option to consider for updating Rust dependencies is [dependabot](https://github.com/dependabot). It automatically opens a PR when a dependency has a new version. For example, s2n-quic uses it, and yesterday it [opened a PR ](https://github.com/aws/s2n-quic/pull/1674) to upgrade Kani to 0.24.
> 
> This is an example of how it can be integrated: #1909
> 
> The annoying thing at the moment is that it would open one PR for every version update as opposed to batching all of them together. There's an [open feature request ](https://github.com/dependabot/dependabot-core/issues/1190) to add support for grouped updates.

I personally think dependabot is a good candidate for this job, it seems like good practice. I could look into how to use it for Kani.

PR #2314, opened 2023-03-23
Remove duplicated CBMC Viewer version string

karkhaz:
This ensures that the version string does not need to be updated in two places, either manually or via a release script.

### Testing:

```sh
$ source kani-dependencies
$ echo $CBMC_VIEWER_VERSION                                                           
3.8
```

### Checklist

### DIFF
--- a/kani-dependencies
+++ b/kani-dependencies
@@ -1,4 +1,7 @@
 CBMC_VERSION="5.79.0"
-# If you update this version number, remember to bump it in `src/setup.rs` too
-CBMC_VIEWER_VERSION="3.8"
+CBMC_VIEWER_VERSION=$(
+  grep "cbmc-viewer==" < src/setup.rs |
+  awk -F== '{print $2}' |
+  awk -F '"' '{print $1}'
+)
 KISSAT_VERSION="3.0.0"
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -137,7 +137,6 @@ fn setup_python_deps(kani_dir: &Path, os: &os_info::Info) -> Result<()> {
     println!("[4/5] Installing Kani python dependencies...");
     let pyroot = kani_dir.join("pyroot");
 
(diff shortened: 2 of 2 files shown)


feliperodri (approved):

jaisnan (approved):
Thanks!

kani-dependencies
```diff
@@ -1,4 +1,7 @@
 CBMC_VERSION="5.79.0"
-# If you update this version number, remember to bump it in `src/setup.rs` too
-CBMC_VIEWER_VERSION="3.8"
+CBMC_VIEWER_VERSION=$(
```
zhassan-aws, line 2: We wanted to use `kani-dependencies` as the single source of truth for versions of dependencies, so I would rather keep the version explicit here and instead update `src/setup.rs` to read the version from this file.

jaisnan: Thanks for catching that.

karkhaz: @zhassan-aws as part of @jaisnan's release script work, we will also need to store the GitHub repositories of each of these tools somewhere. It would be nice not to have to store them in two different places.

Can I suggest that we convert this file into JSON? The file is only ever used from the various `install_*.sh` scripts that run in CI. Thus, we could continue to read the versions if the CI has `jq` available. My proposal is to write the file like this:

```json
{
  "CBMC_VIEWER": {
    "version": "3.8",
    "url": "https://github.com/model-checking/cbmc-viewer",
  }
}
```

Then `install_cbmc.sh` would do
```sh
CBMC_VERSION=$(jq -e CBMC.version < kani-dependencies.json)
```
and `setup.rs` would also parse the file with a JSON parser to find the CBMC Viewer version.

What do you think?

zhassan-aws: Converting it to a JSON file is definitely a good idea.

celinval: Actually, I don't understand why we want JSON here.

PR #2322, opened 2023-03-28
Run "slow" tests as part of pull request checks

tautschnig:
### Description of changes: 

The execution of this check takes less than 15 minutes, which is well below our overall CI time for pull requests (around 45 minutes). Making these tests part of pull request checks will ensure we catch problems early.

### Resolved issues:

n/a

### Related RFC:

n/a

### Call-outs:

n/a

### Testing:

* How is this change tested? Part of this pull request.

* Is this a refactor change? Sort of - refactoring of CI.

### Checklist

### DIFF
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -1,14 +1,10 @@
 # Copyright Kani Contributors
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
-# We use block scalar notation to allow us to add ":" to the workflow name.
-name: >-
-  Nightly: Slow tests
-
+name: Slow tests
 on:
-  schedule:
-    - cron: "30 5 * * *" # Run this every day at 05:30 UTC
-  workflow_dispatch:     # Allow manual dispatching for a custom branch / tag.
+  push:
+    branches: [ main ]
 
 env:
   RUST_BACKTRACE: 1


zhassan-aws:
Also worth mentioning that there are many harnesses in the `tokio-proofs` regression that are currently disabled because they ran for more than 2 hours. I just tried [one of them](https://github.com/model-checking/kani/blob/a7580847a3525f07330dae60510fc2fc78354af3/tests/slow/tokio-proofs/src/tokio/io_chain.rs#L15) and it completed in 3 seconds (after updating the unwind to 12). So we should revisit them and check which ones we can enable.

tautschnig:
It seems the tests themselves (tokio-proofs) take between 2 to 3 minutes, but then I don't know what those tests are specifically testing - I could just move them to a new folder under `tests/` (`tests/tokio-proofs`), but I don't know whether that's justified. Also, in doing so `tests/slow` would be left empty, so I'd have to add a dummy file to keep the folder in git.

For the moment I took a different approach and moved the runs from "Nightly" to run-on-push-to-main: there is nothing in these tests where a different day would make a difference, it's all about the code we have in the repository. So we should detect problems as early as possible, and if we think that (eventually) those tests will take too long for pull-request CI then let's run them immediately after the merge so that we know as soon as possible when we messed up.

karkhaz (approved):

celinval (changes requested):
I agree with @zhassan-aws. We should move those harnesses out of the slow folder. If that leaves us with no slow harness, then we should change this job to not fail if there is no test.

tautschnig:
Closing as I will instead move no-longer-slow tests into the regular test suite as suggested.

PR #2334, opened 2023-04-03
Delete Generator/discriminant test

karkhaz:
This test is flaky in CI; removing until we can diagnose and fix the failure.

### Description of changes: 

Describe Kani's current behavior and how your code changes that behavior. If there are no issues this PR is resolving, explain why this change is necessary.

### Checklist


qinheping (approved):

qinheping:
Can we instead make it a fixme test and create an issue to remind us to fix it later? https://model-checking.github.io/kani/regression-testing.html#fixme-tests

karkhaz:
I'm not sure if fixme would work here because the test is flaky. It sometimes appears to memout in CI, but otherwise works. So we can't add an expected result I think...

zhassan-aws:
I agree adding `fixme` won't work because the test sometimes passes, so the `kani-fixme` suite may fail.

But, as @qinheping pointed out, can we at least file an issue to track re-enabling this test?

karkhaz:
Yes, opened https://github.com/model-checking/kani/issues/2337

zhassan-aws (approved):

PR #2347, opened 2023-04-04
Upgrade toolchain to 2023-02-05

tautschnig:
### Description of changes: 

Implement AggregateKind::{Adt,Closure,Generator} which are being used when moving toolchains from nightly 2023-02-04 to 2023-02-05 (or any later date).

### Resolved issues:

Attempting to upgrade the toolchain to 2023-02-05.

### Related RFC:

n/a

### Call-outs:

n/a

### Testing:

* How is this change tested? All CI checks pass, which includes two additional tests.

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use super::typ::pointee_type;
+use crate::codegen_cprover_gotoc::codegen::place::{ProjectedPlace, TypeOrVariant};
 use crate::codegen_cprover_gotoc::codegen::PropertyClass;
 use crate::codegen_cprover_gotoc::utils::{dynamic_fat_ptr, slice_fat_ptr};
 use crate::codegen_cprover_gotoc::{GotocCtx, VtableCtx};
@@ -13,14 +14,13 @@ use cbmc::goto_program::{Expr, Location, Stmt, Symbol, Type};
 use cbmc::MachineModel;
 use cbmc::{btree_string_map, InternString, InternedString};
 use num::bigint::BigInt;
-use rustc_index::vec::IndexVec;
 use rustc_middle::mir::{AggregateKind, BinOp, CastKind, NullOp, Operand, Place, Rvalue, UnOp};
 use rustc_middle::ty::adjustment::PointerCast;
 use rustc_middle::ty::layout::LayoutOf;
-use rustc_middle::ty::{self, AdtDef, Instance, IntTy, Ty, TyCtxt, UintTy, VtblEntry};
-use rustc_target::abi::{FieldsShape, LayoutS, Size, TagEncoding, VariantIdx, Variants};
+use rustc_middle::ty::{self, Instance, IntTy, Ty, TyCtxt, UintTy, VtblEntry};
+use rustc_target::abi::{FieldsShape, Size, TagEncoding, VariantIdx, Variants};
 use std::collections::BTreeMap;
-use tracing::{debug, warn};
+use tracing::{debug, trace, warn};
 
 impl<'tcx> GotocCtx<'tcx> {
     fn codegen_comparison(&mut self, op: &BinOp, e1: &Operand<'tcx>, e2: &Operand<'tcx>) -> Expr {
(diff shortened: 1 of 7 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -315,8 +500,62 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                } else {
+                    // enum
+                    debug!("IS ENUM");
+                    debug!(?variant_index, "VARIANT INDEX");
+                    let layout = self.layout_of(res_ty);
+                    debug!(?layout, "LAYOUT");
+                    if operands.is_empty() {
+                        self.codegen_rvalue_enum(&variant_index, None, res_ty)
+                    } else {
+                        assert_eq!(operands.len(), 1);
```
celinval, line 537: Why do you have this assertion? I believe the number of operands here will match the number of fields of that variant. For example:
```
enum Enum {
    One(bool),
    Two(bool, bool),
}

fn check() {
    // This will have one operand.
    let _var = Enum::One(false);
    // This will have two operands -- true and false
    let _var = Enum::Two(true, false);
}
```

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -279,6 +279,191 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                if let ty::Adt(adt, subst) = &ty.kind() {
+                    debug!(?subst);
+                    // here we must have one variant
+                    let variant = &adt.variants()[*index];
+
+                    match variant.fields.len() {
+                        0 => {
+                            // would neither use the variant index nor the operand (if any),
+                            // no clear what this expression should mean
+                            unimplemented!()
```
celinval, line 317: I believe the compiler optimizes this since it is a zero size type.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -279,6 +279,191 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                            // let fty = variant.fields[0].ty(self.tcx, subst);
+                            // TODO: this is questionable - no cast, unconditional use of the
+                            // operand
+                            Expr::struct_expr_from_values(
+                                overall_t,
+                                vec![self.codegen_operand(operand_opt.unwrap())],
+                                &self.symbol_table,
+                            )
+                        }
+                        _ => unreachable!(),
```
celinval, line 330: The variant itself can have many fields. Here you want to handle the assignment the same way you handled struct

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Thank you for this example! I was naively assuming that the `Two` case would result in a single, struct-typed operand.

tautschnig: I am now including this as a test case.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: I have completely revised this piece of code.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Done.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -279,6 +282,385 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
+    fn codegen_rvalue_enum_single(
```
celinval, line 285: Can you please add documentation to your functions? It really helps to review the logic.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -279,6 +282,385 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            );
+            Expr::struct_expr_from_values(
+                overall_t,
+                vec![case_value, cases_value],
+                &self.symbol_table,
+            )
+        }
+    }
+
+    fn codegen_niche_struct_from_operands_rec(
```
celinval, line 376: I don't think we should duplicate the niche calculation. I was wondering if we could leverage the `SetDiscriminant` logic instead. If you move that logic to its own function, you could reuse it here.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -279,6 +282,385 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                self.codegen_rvalue_enum_direct(variant_index, operands, overall_t, adt, variants)
+            }
+            Variants::Multiple {
+                tag_encoding:
+                    TagEncoding::Niche { untagged_variant, ref niche_variants, niche_start },
+                variants,
+                tag_field,
+                ..
+            } => {
+                // 1. Find the type of the member corresponding to the variant index.
```
celinval, line 527: What if instead we do the following:
1. Create a temporary variable of the expected type.
2. Initialize the member of the corresponding variant.
3. Invoke the SetDiscriminant logic.
4. Return the temporary variable. 

Another possibility is to refactor the `codgen_rvalue` so it has access to the lvalue. So we wouldn't need the temporary variable. I believe that's closer to the rust backend logic.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -279,6 +282,385 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        let direct_fields = overall_t.lookup_field("direct_fields", &self.symbol_table).unwrap();
+        let direct_fields_components =
+            direct_fields.typ().lookup_components(&self.symbol_table).unwrap().clone();
+        let mut operands_iter = operands.iter();
+        let direct_fields_expr = Expr::struct_expr_with_nondet_fields(
+            direct_fields.typ(),
+            direct_fields_components
+                .iter()
+                .filter_map(|c| {
+                    if c.is_padding() {
```
celinval, line 644: you might be able to simplify this logic if you iterate over the object layout instead.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -304,20 +686,72 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        .fields
+                        .index_by_increasing_offset()
+                        .map(|idx| self.codegen_operand(&operands[idx]))
+                        .collect(),
+                )
+            }
+            AggregateKind::Adt(_, variant_index, _, _, _) if res_ty.is_enum() => {
+                self.codegen_rvalue_enum(&variant_index, &operands, res_ty)
+            }
+            AggregateKind::Adt(_, _, _, _, _) => {
```
celinval, line 728: btw, you can replace this by `AggregateKind::Adt(..)`

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -304,20 +686,72 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    layout
+                        .fields
+                        .index_by_increasing_offset()
+                        .map(|idx| self.codegen_operand(&operands[idx]))
+                        .collect(),
+                    &self.symbol_table,
+                )
+            }
+            AggregateKind::Closure(_, _) => {
+                let typ = self.codegen_ty(res_ty);
```
celinval, line 742: This logic is replicated in at least three different places. You could just join the match statements, like:
```rust
AggregateKind::Tuple | AggregateKind::Closure(..) | AggregateKind::Adt(..) => {
   // code
}
```

celinval (commented):
Thanks for doing this @tautschnig!

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Thanks, done.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Implemented as suggested -- it's closer to MIR now rather than trusting information provided by the type codegen layer. I don't know whether this is good or bad.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Done, will produce a separate PR for this change.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: I now have an implementation of this proposal, and it indeed makes the implementation much easier (but might increase the burden on CBMC, which has to disentangle all the field updates).

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Done.

celinval (commented):
I think we can still simplify the enum case. Here is an example using a temporary variable:

https://github.com/celinval/kani-dev/blob/843f17fcd60b0f360cf035966aa47efbd5d6fa16/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs#L283-L351

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -48,9 +48,32 @@ impl<'tcx> GotocCtx<'tcx> {
                         .goto_expr
                         .assign(self.codegen_rvalue(r, location).cast_to(Type::c_bool()), location)
                 } else {
-                    unwrap_or_return_codegen_unimplemented_stmt!(self, self.codegen_place(l))
+                    match r {
```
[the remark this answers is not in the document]
tautschnig: But this would no longer solve the problem: I need to modify the left-hand side expression (well, I should rather say: I'd prefer to modify the left-hand side expression, else I need to express the right-hand side as a byte-update expression, which so far hasn't been introduced in Kani).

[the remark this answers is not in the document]
tautschnig: ... but then your proposed implementation would do the trick and we could get rid of this hack.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -279,13 +280,120 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 
+    /// Create an initializer for a generator struct.
+    fn codegen_rvalue_generator(&mut self, operands: &[Operand<'tcx>], ty: Ty<'tcx>) -> Expr {
+        let layout = self.layout_of(ty);
+        let discriminant_field = match &layout.variants {
+            Variants::Multiple { tag_encoding: TagEncoding::Direct, tag_field, .. } => tag_field,
+            _ => unreachable!("Generators have more than one variant and use direct encoding"),
```
celinval, line 288: Can you please add the layout to this message? Something like:
```rust
unreachable!("Expected generators to have multiple variants and direct encoding, but found: {layout:?}")
```

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -279,13 +280,120 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        };
+        let overall_t = self.codegen_ty(ty);
+        let direct_fields = overall_t.lookup_field("direct_fields", &self.symbol_table).unwrap();
+        let mut operands_iter = operands.iter();
+        let direct_fields_expr = Expr::struct_expr_with_nondet_fields(
```
celinval, line 293: can you use `struct_expr_from_values` instead?

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -279,13 +280,120 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        // to initialize the non-existing invariant.
+                        trace!(?res_ty, ?variant_index, "Unreachable invariant");
+                        return Expr::nondet(typ);
+                    }
+                    &layout.fields
+                }
+                Variants::Multiple { variants, .. } => &variants[variant_index].fields,
+            };
+
+            debug!(?variant_expr, ?fields, ?operands, "codegen_aggregate enum");
```
celinval, line 364:
```suggestion
            trace!(?variant_expr, ?fields, ?operands, "codegen_aggregate enum");
```

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -279,13 +280,120 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            let init_struct = Expr::struct_expr_from_values(
+                variant_expr.typ().clone(),
+                fields
+                    .index_by_increasing_offset()
+                    .map(|idx| {
+                        let op = self.codegen_operand(&operands[idx]);
+                        debug!(?op, ?idx, "codegen_aggregate enum op");
+                        op
+                    })
```
celinval, line 373: Not sure the formatting is right...

```suggestion
                    .map(|idx| self.codegen_operand(&operands[idx]))
```

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -304,7 +412,42 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    typ,
+                    layout
+                        .fields
+                        .index_by_increasing_offset()
+                        .map(|idx| {
+                            let cgo = self.codegen_operand(&operands[idx]);
+                            if *cgo.typ() == vector_element_type {
+                                cgo
+                            } else {
+                                cgo.transmute_to(vector_element_type.clone(), &self.symbol_table)
```
celinval, line 441: Can you please add a comment about why we need this?

celinval (commented):
It looks good! Thanks @tautschnig.  I just have a few minor comments.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Comment added.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Done.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Formatting was taken care of by the auto-formatter, but I think you're now suggesting to just remove the debug output? Could you please clarify?

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Good call, this is a left-over from the time where I was still trying to figure out what the operands were about. Fixed.

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Done, thank you for the suggestion!

PR #2348, opened 2023-04-04
Add 'benchcomp visualize' & error on regression

### LINKED ISSUE #2338: Implement error return on regression action
Implement a benchcomp 'action' (previously called 'visualization') that terminates benchcomp with a return code of 1 if the result contains a regression.
This is required to be able to terminate CI runs when a regression is detected. It is also useful when running locally, for example when benchcomp is used as part of a makefile or script that checks whether a regression happened.
Users should be able to specify the action like this in their benchcomp.config file:
```yaml actions:
  - type: error_return_on_regression
    checks:
      - metric: passed
        check: boolean_not
      - metric: runtime
        check: ratio
        max: 1.1
```
This says: perform two actions on the results of running the two variants. Check if the 'passed' metric has flipped from true to false or vice versa; and check that the ratio of new to old runtime does not exceed 1.1.
If any pair of results fail either of these checks, the action sets the returncode of benchcomp to 1.

karkhaz:
This commit adds an implementation for the `benchcomp visualize`
command. Currently, there is one visualization, "error_on_regression",
which causes `benchcomp` or `benchcomp visualize` to terminate with a
return code of 1 if there was a regression in any of the metrics.

Users can specify the following in their config file:

    visualize:
    - type: error_on_regression
      variant_pairs:
      - [variant_1, variant_2]
      - [variant_1, variant_3]
      checks:
      - metric: runtime
        test: "lambda old, new: new / old > 1.1"
      - metric: passed
        test: "lambda old, new: False if not old else not new"

This says to check whether any benchmark regressed when run under
variant_2 compared to variant_1. A benchmark is considered to have
regressed if the value of the 'runtime' metric under variant_2 is 10%
higher than the value under variant_1. Furthermore, the benchmark is
also considered to have regressed if it was previously passing, but is
now failing. These same checks are performed on all benchmarks run under
variant_3 compared to variant_1. If any of those lambda functions
returns True, then benchcomp will terminate with a return code of 1.

This commit fixes #2338.
### Testing:

* How is this change tested? Two new regression tests

### Checklist

### DIFF
--- a/tools/benchcomp/benchcomp/entry/benchcomp.py
+++ b/tools/benchcomp/benchcomp/entry/benchcomp.py
@@ -15,3 +15,5 @@ def main(args):
 
     args.suites_dir = run_result.out_prefix / run_result.out_symlink
     results = benchcomp.entry.collate.main(args)
+
+    benchcomp.entry.visualize.main(args)
--- a/tools/benchcomp/benchcomp/entry/visualize.py
+++ b/tools/benchcomp/benchcomp/entry/visualize.py
@@ -4,5 +4,16 @@
 # Entrypoint for `benchcomp visualize`
 
 
-def main(_):
-    pass
+import sys
+
+import yaml
+
+import benchcomp.visualizers.utils
+
+def main(args):
+    with open(args.result_file, encoding="utf-8") as handle:
+        results = yaml.safe_load(handle)
+
+    generate_visualizations = benchcomp.visualizers.utils.Generator(args.config)
+    generate_visualizations(results)
+    sys.exit(benchcomp.visualizers.utils.EXIT_CODE)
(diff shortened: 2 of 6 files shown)


qinheping (commented):
Is it possible to report the name of the regressed benchmark in the error message?
Also, it would be great to have a test that contains metric of more than one benchmark.

karkhaz:
Done both, thank you! I'm printing it out as a warning to stderr.

Note that this visualization is supposed to be fairly minimal, it's not supposed to be doing fancy output... the idea is that we'd implement more visualizations for Markdown output (for GitHub Actions), HTML reports, etc when we need them. But I do think a warning message to terminal is a good idea here

qinheping (approved):
Thank you @karkhaz!

Issue #1609, opened 2022-09-01
Concrete playback requires nightly rustc


celinval:
I tried the following:

```
$ cargo init dummy
$ cd dummy
$ cargo --version
cargo 1.63.0 (fd9c4297c 2022-07-01)
```

I added the following line to `Cargo.toml`
```toml
[dev-dependencies]
kani = { git = "https://github.com/model-checking/kani", features = ["concrete_playback"] }
```

I opened src/main.rs

```rust
#[kani::proof]
fn force_failure() {
  assert!(kani::any());
}
```

using the following command line invocation:

```
cargo kani --enable-unstable --concrete-playback
cargo test
```

with Kani version: 0.9.0

I expected to see this happen: Cargo runs the new test and fail.

Instead, this happened: Cargo failed to build `kani` library with the following error:
```
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> ~/.cargo/git/checkouts/kani-0ce0dacf5e98886d/281d0bb/library/kani/src/lib.rs:5:1
  |
5 | #![feature(rustc_attrs)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^
```

celinval:
We could guard the usage of `rustc_attrs` and `rustc_diagnostic_item` with `cfg_attr`, but I was wondering if there's a cleaner solution.

celinval:
This is very similar to https://github.com/model-checking/kani/issues/1581

adpaco-aws:
> I was wondering if there's a cleaner solution.

I haven't been able to find one, but IMHO the `cfg_attr` solution is quite clean already.

celinval:
> > I was wondering if there's a cleaner solution.
> 
> I haven't been able to find one, but IMHO the `cfg_attr` solution is quite clean already.

We could probably use a `macro` or a `proc_macro` to do that. Unfortunately, https://github.com/model-checking/kani/pull/2199 has added another unstable feature.

jaisnan:
Would removing the transmute remove the dependency on nightly features entirely?

celinval:
The way we are doing transmute is the reason why we need the `incomplete_features` for arbitrary.

celinval:
The transmute could be replaced by storing `Any` as described here: https://github.com/model-checking/kani/issues/1527

PR #2353, opened 2023-04-06
Enable Concrete Playback Unit tests to run with  `--cfg kani`

### LINKED ISSUE #1610: Concrete playback instructions don't work if user sets `--cfg=kani`
I tried the following:

```
$ cargo init dummy
$ cd dummy
```

I added the following line to `Cargo.toml`
```toml
[dev-dependencies]
kani = { git = "https://github.com/model-checking/kani", features = ["concrete_playback"] }
```

I opened src/main.rs

```rust
#[cfg(kani)]
mod harnesses {
  #[kani::proof]
  fn force_failure() {
    assert!(kani::any());
  }
}
```

```
cargo kani --enable-unstable --concrete-playback
RUSTFLAGS="--cfg=kani" cargo +nightly test
```

with Kani version: 0.9.0

I expected to see this happen: Cargo runs the new test and fail.

Instead, this happened: Cargo failed to build the harness with the following error:
```
error[E0433]: failed to resolve: use of undeclared crate or module `kanitool`
  --> src/main.rs:36:5
   |
36 |     #[kani::proof]
   |     ^^^^^^^^^^^^^^ use of undeclared crate or module `kanitool`
   |
   = note: this error originates in the attribute macro `kani::proof` (in Nightly builds, run with -Z macro-backtrace for more info)

```

jaisnan:
### Description of changes: 

Fixes #1610 by creating a `concrete_playback` feature in `kani_macros`. Users can now call their unit tests with the command - `RUSTFLAGS="--cfg=kani" cargo +nightly test`.

### Resolved issues:

Resolves #1610 

### Call-outs:

Still need to pass `+nightly ` with the tests. Already recorded in #1609.

### Testing:

* How is this change tested? Regression tests
 
### Checklist


tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
```diff
@@ -0,0 +1,44 @@
...
+echo "Running cargo test on the unit test ..."
+echo
+
+cd bin/
+
+output=$(grep 'channel = ' ../../../../rust-toolchain.toml | cut -d '"' -f 2)
+echo "$output"
+
+# Run cargo test on the unit test
+cargo_output=$(RUSTFLAGS="--cfg=kani" cargo +${output} test 2>/dev/null)
```
karkhaz, line 36: This appears to be saving to a variable whose value is then printed out, but not otherwise used. Suggest that you just run the command without saving it to a variable.

Also, is there a need to sink stderr? If there are error messages, it would be good to see them in CI.

If there's a need to parse stdout and stderr separately, I would suggest to redirect them to two different files, and then cat both files. Otherwise, I suggest just run the command and don't sink stderr and don't save stdout to a variable.

tests/script-based-pre/concrete_playback_test/bin/src/main.rs
```diff
@@ -0,0 +1,29 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
adpaco-aws, line 3: The test needs a description. I'm OK with having it here or in the script.

tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
```diff
@@ -0,0 +1,44 @@
...
+# Check if cargo is installed
+if ! command -v cargo &> /dev/null; then
+    echo "cargo command not found. Please install Rust and Cargo."
+    exit 1
+fi
+
+# Check if the directory containing cargo is in the PATH environment variable
+if ! echo "$PATH" | grep -q "$(dirname "$(command -v cargo)")"; then
+    echo "Directory containing cargo is not in PATH. Adding directory to PATH..."
+    echo "export PATH=$PATH:$(dirname "$(command -v cargo)")" >> ~/.bashrc
```
karkhaz, line 23: Is this intended to be run by local users? I don't think it's a good idea to contaminate users' bashrc.

If you want to add to the path, you can do that in the script without changing .bashrc, by directly calling `export PATH...`.

tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
```diff
@@ -0,0 +1,44 @@
...
+    source ~/.bashrc
+fi
+
+echo "Running cargo test on the unit test ..."
```
karkhaz, line 27: TBH I think most of these echo statements should go to stderr:

```sh
>&2 echo "Running cargo test on the unit test ..."
```

tests/script-based-pre/concrete_playback_test/bin/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+#[cfg(kani)]
+mod harnesses {
+    #[kani::proof]
+    fn harness() {
+        let result_1: Result<u8, u8> = kani::any();
+        let result_2: Result<u8, u8> = kani::any();
+        assert!(!(result_1 == Ok(101) && result_2 == Err(102)));
+    }
+
```
adpaco-aws, line 14: Can we add a reference to the `ui` sibling of this test?

tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
```diff
@@ -0,0 +1,44 @@
...
+# Run cargo test on the unit test
+cargo_output=$(RUSTFLAGS="--cfg=kani" cargo +${output} test 2>/dev/null)
+echo "$cargo_output"
+
+cd ..
+
+# Try to leave a clean output folder at the end
+rm -rf ${OUT_DIR}
+
+set -e
```
karkhaz, line 44: Not sure that this is necessary?

karkhaz (changes requested):

tests/script-based-pre/concrete_playback_test/bin/Cargo.toml
adpaco-aws, line 1: Why is this in a `bin` folder?

adpaco-aws, line 1: Can we rename the test to `playback_with_cfg_kani` or something along those lines?

tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
```diff
@@ -0,0 +1,44 @@
...
+echo "$output"
+
+# Run cargo test on the unit test
+cargo_output=$(RUSTFLAGS="--cfg=kani" cargo +${output} test 2>/dev/null)
+echo "$cargo_output"
+
+cd ..
+
+# Try to leave a clean output folder at the end
+rm -rf ${OUT_DIR}
```
karkhaz, line 42: Is `bin/target` used by any other test or build artifact? I'm not sure it's a good idea to delete it

tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
jaisnan: Those error messages from `stderr` are warning messages about unused imports and would just add noise. I think it's better to sink them here. But I can directly print out stdout instead of saving to a variable.

tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
jaisnan: We set +e in the script to prevent the script from getting blocked so I thought it'd be best to undo it at the end of the script.

tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
jaisnan: I can remove this section entirely, it was a check that necessary for my local environment but I don't think it's relevant to CI.

tests/script-based-pre/concrete_playback_test/bin/src/main.rs
jaisnan: That's a good idea. I'll add a comment referencing the original UI test.

tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
karkhaz: > Those error messages from stderr are warning messages about unused imports

that’s in the case where the test works. What if there’s an error, wouldn’t stderr be useful there? But sure, if you think there’s no useful information there, I’m happy for you to drop stderr

tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
karkhaz: I think removing it is fine. But it’s fine to keep it too, it’s a reasonable check to have. But definitely don’t append to bashrc. If you just do `export PATH=$PATH:$(dirname "$(command -v cargo)")` inside that block, instead of appending it to bashrc, then it will take effect until the end of the script

tests/script-based-pre/concrete_playback_test/concrete_playback_test.sh
karkhaz: `+e`/`-e` only takes effect for the duration of the script, it won't affect any other scripts. So it's redundant to put this statement on the last line of the script

jaisnan: Oh cool. I shall remove it then! Thanks for pointing it out.

tests/script-based-pre/playback_with_cfg_kani/sample_crate/src/main.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// This module contains a harness and it's associated unit test that is generated
+// by running concrete-playback on it. There is an existing UI test to generate the unit test
+// itself (in kani/tests/ui/concrete-playback/result). The current module runs `cargo test` on the unit test
+// on the unit test and checks the output to test the rest of the concrete-playback flow.
```
adpaco-aws, line 7:
```suggestion
// This test checks that running the unit test generated using the concrete playback feature
// with `RUSTFLAGS="--cfg=kani" cargo +nightly test doesn't cause a compilation error.
// There is an existing UI test to generate the unit test itself (in kani/tests/ui/concrete-playback/result).
```
Much simpler. The `--cfg=kani` detail is important here.

adpaco-aws (approved):

celinval:
Thanks Jai for fixing this. I was also thinking that we could improve this logic by making the default the behaviour that is today activated by concrete playback.

Instead of using `#[cfg(kani)]` in our libraries, we could use something like `#[cfg(kani_symex)]` which is set by our compiler together with the `kani` cfg.

I didn't want to suggest this before so this fix wouldn't get too big. But I think it would address @karkhaz important concern of cfg explosion.

jaisnan:
> Thanks Jai for fixing this. I was also thinking that we could improve this logic by making the default the behaviour that is today activated by concrete playback.
> 
> Instead of using `#[cfg(kani)]` in our libraries, we could use something like `#[cfg(kani_symex)]` which is set by our compiler together with the `kani` cfg.
> 
> I didn't want to suggest this before so this fix wouldn't get too big. But I think it would address @karkhaz important concern of cfg explosion.

Should I address this in this PR or can we raise an issue and address it later?

PR #2355, opened 2023-04-07
Add --no-copy flag to benchcomp run

karkhaz:
This flag makes benchcomp not create a fresh copy of each benchmark directory before running each variant. The default behaviour is to make a new copy each time, to avoid built artefacts from silently being skipped, and to ensure that patches apply cleanly. Users can use this flag when benchmark runs are guaranteed to be idempotent, to avoid waiting for benchmark suites to be copied to a fresh directory.

### Call-outs:

Should --no-copy actually be the default behaviour?

### Testing:

* How is this change tested? I ran it

* Is this a refactor change?

### Checklist


qinheping (approved):
I think it is OK to keep copy as default behavior as it is safer.

PR #2370, opened 2023-04-13
Add performance regression comparison in CI

### LINKED ISSUE #2277: Add a CI pass for pull requests that detects perfomance regressions compared to HEAD

karkhaz:
This commit adds a CI job that runs the `benchcomp` tool on the
performance regression suite, comparing the HEAD of the pull request to
the branch that the PR targets.

The CI job fails if any performance benchmark regresses when run using
the HEAD version of Kani with respect to the 'base' branch. Regression
is defined as a regression in symex or solver time of 10% for any
benchmark for which this value is >2s, or if any performance benchmark
fails with the HEAD version while passing with the base.

This fixes #2277.

### Call-outs:

This is currently running 'old' kani on the old perf suite, and 'new' kani on the new suite.
There seems to be a problem with finding the stdlib otherwise. This means that benchmarks that get renamed will not be checked in the PR that renames them (though they will then start to be checked again when the PR is merged).

### Testing:

* How is this change tested?

[Here](https://github.com/karkhaz/kani/actions/runs/4684181964/jobs/8300057143) is a run that failed because one of the benchmarks regressed. Compared with that run, I've changed this PR so that numerical metrics that are less than 10 are not compared between the old and new checkouts.

### Checklist

### DIFF
--- a/.github/workflows/kani.yml
+++ b/.github/workflows/kani.yml
@@ -226,3 +226,61 @@ jobs:
           if-no-files-found: error
           # Aggressively short retention: we don't really need these
           retention-days: 3
+
+  perf-benchcomp:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Save push event HEAD and HEAD~ to environment variables
+        if: ${{ github.event_name == 'push' }}
+        run: |
+          echo "NEW_REF=${{ github.event.after}}" | tee -a "$GITHUB_ENV"
+          # We want to compare with $NEW_REF~. But we can't know what
+          # that ref actually is until we clone the repository, so for
+          # now make it equal to $NEW_REF
+          echo "OLD_REF=${{ github.event.after }}" | tee -a "$GITHUB_ENV"
+
+      - name: Save pull request HEAD and target to environment variables
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          echo "OLD_REF=${{ github.event.pull_request.base.sha }}" | tee -a "$GITHUB_ENV"
+          echo "NEW_REF=${{ github.event.pull_request.head.sha }}" | tee -a "$GITHUB_ENV"
+
+      - name: Check out Kani (old variant)
+        uses: actions/checkout@v3
+        with:
+          path: ./old
+          ref: ${{ env.OLD_REF }}
+          fetch-depth: 2
+
+      - name: Check out HEAD~ of push event as 'old' variant
+        if: ${{ github.event_name == 'push' }}
+        run: pushd old && git checkout "${NEW_REF}^"
+
+      - name: Check out Kani (new variant)
+        uses: actions/checkout@v3
+        with:
+          path: ./new
+          ref: ${{ env.NEW_REF }}
+
+      - name: Set up Kani Dependencies (old variant)
+        uses: ./old/.github/actions/setup
+        with:
+          os: ubuntu-20.04
+          kani_dir: old
+
+      - name: Set up Kani Dependencies (new variant)
+        uses: ./new/.github/actions/setup
+        with:
+          os: ubuntu-20.04
+          kani_dir: new
+
+      - name: Build Kani (new variant)
+        run: pushd new && cargo build-dev
+
+      - name: Build Kani (old variant)
+        run: pushd old && cargo build-dev
+
+      - name: Run benchcomp
+        run: |
+          new/tools/benchcomp/bin/benchcomp \
+            --config new/tools/benchcomp/configs/perf-regression.yaml
--- a/tools/benchcomp/configs/perf-regression.yaml
+++ b/tools/benchcomp/configs/perf-regression.yaml
@@ -1,0 +1,41 @@
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+# Run the Kani perf suite twice, erroring out on regression. This config
+# file is primarily intended to be used in CI, because it assumes that
+# there are two Kani checkouts in the 'old' and 'new' directories;
+# benchcomp compares the performance of these two checkouts.
+
+variants:
+  kani_new:
+    config:
+      directory: new
+      command_line: PATH=$(realpath new/scripts):$PATH scripts/kani-perf.sh
+      env:
+        RUST_TEST_THREADS: "1"
+  kani_old:
+    config:
+      directory: old
+      command_line: PATH=$(realpath old/scripts):$PATH scripts/kani-perf.sh
+      env:
+        RUST_TEST_THREADS: "1"
+
+run:
+  suites:
+    kani_perf:
+      parser:
+        module: kani_perf
+      variants: [kani_old, kani_new]
+
+visualize:
+  - type: error_on_regression
+    variant_pairs: [[kani_old, kani_new]]
+    checks:
+      - metric: success
+        # Compare the old and new variants of each benchmark. The
+        # benchmark has regressed if the lambda returns true.
+        test: "lambda old, new: False if not old else not new"
+      - metric: solver_runtime
+        test: "lambda old, new: False if new < 2 else new/old > 1.1"
+      - metric: symex_runtime
+        test: "lambda old, new: False if new < 2 else new/old > 1.1"
--- a/tools/benchcomp/benchcomp/entry/run.py
+++ b/tools/benchcomp/benchcomp/entry/run.py
@@ -57,7 +57,8 @@ class _SingleInvocation:
 
         if self.copy_benchmarks_dir:
             shutil.copytree(
-                self.directory, self.working_copy, ignore_dangling_symlinks=True)
+                self.directory, self.working_copy,
+                ignore_dangling_symlinks=True, symlinks=True)
 
         try:
             subprocess.run(
--- a/tools/benchcomp/benchcomp/visualizers/utils.py
+++ b/tools/benchcomp/benchcomp/visualizers/utils.py
@@ -3,6 +3,7 @@
 
 
 import dataclasses
+import logging
 import typing
 
 import benchcomp.visualizers
@@ -82,7 +83,7 @@ class AnyBenchmarkRegressedChecker:
                 new = bench["variants"][new_variant]["metrics"][self.metric]
 
                 if has_regressed(old, new):
-                    logging.warining(
+                    logging.warning(
                         "Benchmark '%s' regressed on metric '%s' (%s -> %s)",
                         bench_name, self.metric, old, new)
                     ret = True
--- a/tools/benchcomp/configs/README.md
+++ b/tools/benchcomp/configs/README.md
@@ -1,0 +1,4 @@
+Example Benchcomp Configurations
+================================
+
+The files in this directory can be passed to benchcomp's -c/--config flag.


.github/workflows/kani.yml
```diff
@@ -245,3 +245,75 @@ jobs:
...
+              variant_pairs: [[kani_base, kani_pr]]
+              checks:
+                - metric: success
+                  test: "lambda old, new: False if not old else not new"
+                - metric: solver_runtime
+                  test: "lambda old, new: False if new < 1 else new / old > 1.1"
+                - metric: symex_runtime
+                  test: "lambda old, new: False if new < 1 else new / old > 1.1"
+                - metric: verification_time
+                  test: "lambda old, new: False if new < 1 else new / old > 1.1"
```
adpaco-aws, line 315: This may frequently block on benchmarks which take a small amount of time (e.g., less than 5s). Maybe we should define different scales, for example:
 * Less than 1s: Block on >3s
 * Less than 30s: Block on >1min
 * Anything else: Block on >1.1x
Of course, I haven't experimented on these particular values, so more scales or tweaks might be needed.

qinheping:
> This may frequently block on benchmarks which take a small amount of time (e.g., less than 5s). Maybe we should define different scales, for example:
> 
>     * Less than 1s: Block on >3s
> 
>     * Less than 30s: Block on >1min
> 
>     * Anything else: Block on >1.1x
> 
> 
> Of course, I haven't experimented on these particular values, so more scales or tweaks might be needed.

I agree with the idea that we have different scales for different benchmarks. But I am not sure about the numbers. How about running this check in CI a couple times to collect the data of running, time and then use the data to decide the scale numbers.

.github/workflows/kani.yml
```diff
@@ -245,3 +245,75 @@ jobs:
...
+          os: ubuntu-20.04
+          kani_dir: pr
+
+      - name: Build Kani (pull request HEAD)
+        run: pushd pr && cargo build-dev
+
+      - name: Build Kani (pull request base)
+        run: pushd base && cargo build-dev
+
+      - name: Write benchcomp config file
```
celinval, line 282: Is there a reason why the config file is hard coded here instead of adding it to the repository?

.github/workflows/kani.yml
```diff
@@ -245,3 +245,75 @@ jobs:
...
           retention-days: 3
+
+  perf-benchcomp:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout Kani (pull request HEAD)
+        uses: actions/checkout@v3
+        with:
+          path: ./pr
+          ref: ${{ github.event.pull_request.head.sha }}
```
celinval, line 256: What will happen when this job is triggered due to a push?

celinval (commented):
Just a nit: This PR is adding a new workflow for PRs but it isn't blocking yet. AFAIK, this is a github configuration which I think we should activate only after we adjust the scale that ensures this job won't give too many false positives.

.github/workflows/kani.yml
karkhaz: Thanks, I've added a condition so that this only runs on `pull_request` events.

I experimented with also making it run on pushes. The trouble is that [push events](https://docs.github.com/webhooks-and-events/webhooks/webhook-events-and-payloads#push) don't have the 'before' commit that we need (i.e. the HEAD of the branch that we're trying to merge into). There is a 'before' field in that documentation, but it refers to something else---it's the commit that was previously at the tip of the pull request branch.

That is, if you post a PR to merge commit `abc123` into `main`, and then later force-push so that the PR is now trying to merge `def456` into `main`, then the push event will have `github.event.push.after == def456` and `github.event.push.before == abc123`. There is no `main` in the event, so we are unable to check it out to compare it against `def456`.

karkhaz:
> Just a nit: This PR is adding a new workflow for PRs but it isn't blocking yet. AFAIK, this is a github configuration which I think we should activate only after we adjust the scale that ensures this job won't give too many false positives.

Thanks, I've changed the commit message. Agreed about leaving this to bake for a while before we make it blocking

karkhaz:
@qinheping here are my thoughts on which numbers to use:

I've tried a few experiments with CI and locally. I get a lot less variance locally, often all benchmarks (or all but one) have a variance <10%. This is when comparing identical Kanis.

In CI, the numbers that @adpaco-aws seem to work well for now. I do think it would be a good idea to optimize these numbers, but I would prefer to see what the range is on realistic PRs---that is, PRs that actually change the implementation of Kani and so cause and actual performance difference.

What I propose:
* In a future PR, I'll save the results.yaml file as an artifact. This will allow me to download the results of several runs over the last 30 days.
* `benchcomp visualize` can be used locally on a results.yaml file, without re-running the benchmarks. So if we download a lot of results.yaml files, we can try lots of different numbers in the config file and see which ones are best suited for checking that wide range of result files.

I suggest that we merge this PR first, though. I would be inclined to merge quickly and start getting data as quickly as we can, rather than trying to get it right from now, especially since we don't have data for realistic runs.

tools/benchcomp/configs/perf-regression.yaml
```diff
@@ -0,0 +1,41 @@
...
+    variant_pairs: [[kani_base, kani_pr]]
+    checks:
+      - metric: success
+        test: "lambda old, new: False if not old else not new"
+      - metric: solver_runtime
+        test: "lambda old, new: new/old>3 if new < 1 else new/old>2 if new<30 else new/old>1.1"
+      - metric: symex_runtime
+        test: "lambda old, new: new/old>3 if new < 1 else new/old>2 if new<30 else new/old>1.1"
+      - metric: verification_time
+        test: "lambda old, new: new/old>3 if new < 1 else new/old>2 if new<30 else new/old>1.1"
```
tautschnig, line 41: It seems that symex and solver time produce more stable numbers. So can we please not go ahead with this without first making sure that we are using a metric exhibiting very low noise?

tautschnig (changes requested):
I think we need to make sure we have access to metrics that are very low noise.

qinheping:
> @qinheping here are my thoughts on which numbers to use:
> 
> I've tried a few experiments with CI and locally. I get a lot less variance locally, often all benchmarks (or all but one) have a variance <10%. This is when comparing identical Kanis.
> 
> In CI, the numbers that @adpaco-aws seem to work well for now. I do think it would be a good idea to optimize these numbers, but I would prefer to see what the range is on realistic PRs---that is, PRs that actually change the implementation of Kani and so cause and actual performance difference.
> 
> What I propose:
> 
>     * In a future PR, I'll save the results.yaml file as an artifact. This will allow me to download the results of several runs over the last 30 days.
> 
>     * `benchcomp visualize` can be used locally on a results.yaml file, without re-running the benchmarks. So if we download a lot of results.yaml files, we can try lots of different numbers in the config file and see which ones are best suited for checking that wide range of result files.
> 
> 
> I suggest that we merge this PR first, though. I would be inclined to merge quickly and start getting data as quickly as we can, rather than trying to get it right from now, especially since we don't have data for realistic runs.

Sounds great! Thank you @karkhaz.
Yes, we also want to save the results from realistic runs and review them periodically to detect accumulating performance regression.

qinheping (approved):

.github/workflows/kani.yml
celinval: Can you checkout `HEAD~`?

I would actually really like to run this when a release tag is created and compare with the previous release.

tools/benchcomp/configs/perf-regression.yaml
karkhaz: I just tried removing Verification Time, and instead checked for solver and symex time >5%. [But one benchmark failed](https://github.com/model-checking/kani/actions/runs/4701956238/jobs/8338630914?pr=2370) ('vec/string/main' regressed on metric 'symex_runtime' (1.5912 -> 1.69417))... so I'm really not convinced that the tool itself is the problem here, the noise seems somewhat intrinsic to the CI machine.

I would like to push back a bit by noting what this PR is for: we're trying to catch changes to Kani (or upgrades to CBMC or the solver) that cause typical benchmarks to regress. I'm not sure that the noise itself is a problem if we can ignore it. If runs of identical Kani are persistently varying by this much, then I think either setting the threshold higher or ignoring fast benchmarks would still allow us to catch real regressions. (I prefer the second approach)

.github/workflows/kani.yml
karkhaz: By HEAD I guess you mean the tip of the pull request? I tried this, but this only does the right thing if there is only 1 commit in the pull request.

Of course, we squash when merging, so the push event for the merge commit would do the right thing. However, while the PR is open, checking out HEAD~ would not be checking against the correct base.

> I would actually really like to run this when a release tag is created and compare with the previous release.

I suggest that we have a separate workflow for this that triggers on pushing a particular tag. We would then checkout the tag, and the previous tag. For what you're asking for, I don't think HEAD~ is correct anyway, HEAD~ of a tag is just the commit before the tag in Kani because we're using a linear history.

I think having multiple different workflows that each run benchcomp is a good idea, because they would run in parallel but on different machines. If we do this, I would refactor the first steps of the job into a separate workflow---i.e. checkout two different refs in two directories---and we'd use that workflow from several different benchcomp jobs.

karkhaz: If we had a separate 'release' branch where every commit was a tagged release, then `HEAD~` would have the semantics you suggest, but with a linear history I'm not aware of a symbolic way to jump from one tag to the previous (or next) one, you have to parse the log.

celinval: By `HEAD~`, I'm talking about running this on `push` events not PRs.

For the tag workflow, we would need to find the previous / stable release.

karkhaz: Screen Shot 2023-04-14 at 18 51 57
```
# This only works because releases are on their own branch so can be traversed with '~'

$ git show -s --pretty='%(describe:)' '1.27.0~'
1.26.0

$ git show -s --pretty='%(describe:)' '1.27.0~~'
1.25.0

$ git show -s --pretty='%(describe:)' '1.27.0^2'
34c30a9

$ git show -s --pretty='%(describe:)' '1.27.0^2~'
9bbb729
```

celinval: BTW, you could create an action that takes two references that will be compared. Then each workflow will just have a logic to figure out the references and call the action.

tautschnig (approved):
Got convinced that this is a start, but I do maintain that we need to dig deeper to figure out what the sources of performance variability are.

celinval:
> Got convinced that this is a start, but I do maintain that we need to dig deeper to figure out what the sources of performance variability are.

I agree. I think we should merge the PR but only make it required once we figure that out.

.github/workflows/kani.yml
karkhaz: Thanks a lot for these comments @celinval.

I changed the job so that it does the right thing for both PRs and pushes, assuming that push events contain a single commit.
* For pull requests, benchcomp compares the HEAD of the PR with the HEAD of the branch that the PR is to be merged into. The workflow [sets `OLD_REF` and `NEW_REF` environment variables](https://github.com/model-checking/kani/actions/runs/4705590189/jobs/8346243267?pr=2370#step:3:8) with logic dependent on whether the event was a push or pull request, and uses these variables later for checking out.
* For pushes (here is an [example run](https://github.com/karkhaz/kani/actions/runs/4705582390/jobs/8346229668) in my personal fork), initially the 'old' and 'new' refs are the same, because we don't know what the 'old' ref is from the push event. Once the action clones both repositories, it checks out `HEAD~` in the 'old' one.

.github/workflows/kani.yml
```diff
@@ -226,3 +226,61 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Save push event HEAD and HEAD~ to environment variables
+        if: ${{ github.event_name == 'push' }}
+        run: |
+          echo "NEW_REF=${{ github.event.after}}" | tee -a "$GITHUB_ENV"
+          # We want to compare with $NEW_REF~. But we can't know what
+          # that ref actually is until we clone the repository, so for
+          # now make it equal to $NEW_REF
+          echo "OLD_REF=${{ github.event.after }}" | tee -a "$GITHUB_ENV"
```
celinval, line 240: what about making this something like:
```suggestion
          echo "OLD_REF=${{ github.event.after }}~" | tee -a "$GITHUB_ENV"
```
and removing the extra checkout step?

.github/workflows/kani.yml
```diff
@@ -226,3 +226,61 @@ jobs:
...
+      - name: Check out Kani (old variant)
+        uses: actions/checkout@v3
+        with:
+          path: ./old
+          ref: ${{ env.OLD_REF }}
+          fetch-depth: 2
+
+      - name: Check out HEAD~ of push event as 'old' variant
+        if: ${{ github.event_name == 'push' }}
+        run: pushd old && git checkout "${NEW_REF}^"
```
celinval, line 257:
```suggestion
      - name: Check out Kani (old variant)
        uses: actions/checkout@v3
        with:
          path: ./old
          ref: ${{ env.OLD_REF }}
```

.github/workflows/kani.yml
celinval: If this doesn't work, another possibility would be:
```suggestion
          echo "OLD_REF=HEAD~" | tee -a "$GITHUB_ENV"
```

tools/benchcomp/configs/perf-regression.yaml
```diff
@@ -0,0 +1,39 @@
...
+        RUST_TEST_THREADS: "1"
+
+run:
+  suites:
+    kani_perf:
+      parser:
+        module: kani_perf
+      variants: [kani_base, kani_pr]
+
+visualize:
```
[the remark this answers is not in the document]
karkhaz: The `error_on_regression` visualization does print out any benchmark that regressed as a warning to stderr.

It doesn't attempt to print anything out in a pretty/parseable way. The idea would be that we'd write a different visualization to do that. This is because you might want it to print something else to stdout, e.g. the `dump_yaml` visualization or a different kind of visualization.

.github/workflows/kani.yml
karkhaz: These don't work because the 'checkout' action needs to send the actual ref to the server in order to do a shallow clone. We can't specify things like `HEAD~` or `abc123~` as an argument to the checkout@v3 action. So I'm instead specifying HEAD, and cloning to a depth of 2 rather than 1 so that I can subsequently checkout HEAD~.

tools/benchcomp/configs/perf-regression.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - type: error_on_regression
+    variant_pairs: [[kani_base, kani_pr]]
+    checks:
+      - metric: success
+        test: "lambda old, new: False if not old else not new"
```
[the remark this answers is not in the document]
karkhaz: Done, thanks

.github/workflows/kani.yml
celinval: sorry, I missed the `fetch-depth: 2`. You would still need that

PR #2386, opened 2023-04-18
Enforce unstable APIs can only be used if the related feature is enabled

### LINKED ISSUE #2357: Add unstable attribute to tag experimental APIs
**Proposed change:** Add the attribute `#[kani:: unstable]` as part of #2279. This will introduce an opt-in model for users to be able to use unstable APIs.

**Motivation:** Allow users to try new APIs, but only enable them when explicitly requested.

celinval:
### Description of changes: 

This PR adds the new opt-in option for users to enable unstable features (via `-Z` option or `unstable` table in Cargo.toml). This PR also adds logic to parse and verify APIs tagged with `unstable`.

### Resolved issues:

Fixes #2357 

### Related RFC:

#2279 

### Call-outs:

1. We still need to finish migrating all unstable features to the new model and deprecate the old ones.
2. This PR turned out a big big. I can break it down if needed.

### Testing:

### Checklist


kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -128,6 +147,55 @@ pub fn extract_harness_attributes(tcx: TyCtxt, def_id: DefId) -> Option<HarnessA
     }
 }
 
+/// Check that any unstable API has been enabled. Otherwise, emit an error.
```
adpaco-aws, line 150: This function is doing many things: It's checking that all reachable unstable features have been enabled, emits an error if that's not the case, and emits an error also if it wasn't able to parse `#[kani::unstable]` annotations. Please ensure that the description includes this. It can also be more readable if we refactor error handling for the `Ok(unstable_attr)` case.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -128,6 +147,55 @@ pub fn extract_harness_attributes(tcx: TyCtxt, def_id: DefId) -> Option<HarnessA
...
+                        let mut err = tcx.sess.struct_err(format!(
+                            "Use of unstable feature `{}`{reason}",
+                            unstable_attr.feature
+                        ));
+                        err.span_note(
+                            tcx.def_span(def_id),
+                            format!("the function `{fn_name}` is unstable:"),
+                        );
+                        if let Some(link) = unstable_attr.issue {
+                            err.note(format!("see issue {link} for more information"));
```
adpaco-aws, line 171: Do we always expect links to issues?

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -128,6 +147,55 @@ pub fn extract_harness_attributes(tcx: TyCtxt, def_id: DefId) -> Option<HarnessA
...
+                        }
+                        err.help(format!(
+                            "use `-Z {}` to enable use this API.",
```
adpaco-aws, line 174: I don't like using the term "API", maybe we can just use `{fn_name}`?

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -128,6 +147,55 @@ pub fn extract_harness_attributes(tcx: TyCtxt, def_id: DefId) -> Option<HarnessA
...
+                        let fn_name = tcx.def_path_str(def_id);
+                        let reason =
+                            unstable_attr.reason.map_or(String::default(), |r| format!(": {r}"));
+                        let mut err = tcx.sess.struct_err(format!(
+                            "Use of unstable feature `{}`{reason}",
+                            unstable_attr.feature
+                        ));
+                        err.span_note(
+                            tcx.def_span(def_id),
+                            format!("the function `{fn_name}` is unstable:"),
```
adpaco-aws, line 168: Why the colon at the end?

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -347,6 +454,23 @@ fn parse_path(meta_item: &MetaItem) -> Option<String> {
...
+fn parse_key_values(attr: &Attribute) -> Result<BTreeMap<String, String>, String> {
+    trace!(list=?attr.meta_item_list(), ?attr, "parse_key_values");
+    let args = attr.meta_item_list().ok_or("malformed attribute input")?;
+    args.iter()
+        .map(|arg| match arg.meta_item() {
+            Some(MetaItem { path: key, kind: MetaItemKind::NameValue(val), .. }) => {
+                Ok((key.segments.first().unwrap().ident.to_string(), val.symbol.to_string()))
+            }
+            _ => Err(format!(
+                r#"expected "key = value" pair, but found `{}`"#,
```
adpaco-aws, line 467: Does Rust reports these errors the same way?

kani-compiler/src/kani_middle/mod.rs
```diff
@@ -50,6 +54,23 @@ pub fn check_crate_items(tcx: TyCtxt, ignore_asm: bool) {
...
+/// This method will exhaustively print any error / warning and it will abort at the end if any
+/// error was found.
+pub fn check_reachable_items(tcx: TyCtxt, queries: &QueryDb, items: &[MonoItem]) {
+    // Avoid printing the same error multiple times for different instantiations of the same item.
+    let mut def_ids = HashSet::new();
+    for item in items {
+        let def_id = item.def_id();
+        if !def_ids.contains(&def_id) {
+            // Check if any unstable attribute was reached.
+            check_unstable_features(tcx, queries.get_unstable_features(), def_id);
```
adpaco-aws, line 67: It's assumed here that two or more instantiations of the same item don't need to be checked. Why are they given then? How can they differ?

kani-driver/src/args.rs
```diff
@@ -343,6 +347,16 @@ pub enum ConcretePlaybackMode {
     InPlace,
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, ValueEnum, strum_macros::Display)]
+#[strum(serialize_all = "kebab-case")]
+pub enum UnstableFeatures {
+    /// Allow replace certain items with stubs (mocks).
+    /// See (RFC-0002)[https://model-checking.github.io/kani/rfc/rfcs/0002-function-stubbing.html]
```
adpaco-aws, line 354:
```suggestion
    /// See [RFC-0002](https://model-checking.github.io/kani/rfc/rfcs/0002-function-stubbing.html)
```

kani-driver/src/args.rs
```diff
@@ -343,6 +347,16 @@ pub enum ConcretePlaybackMode {
     InPlace,
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, ValueEnum, strum_macros::Display)]
+#[strum(serialize_all = "kebab-case")]
+pub enum UnstableFeatures {
+    /// Allow replace certain items with stubs (mocks).
```
adpaco-aws, line 353:
```suggestion
    /// Allows to replace certain items with stubs (mocks).
```
Not sure if it's grammatically correct though.

kani-driver/src/args_toml.rs
```diff
@@ -116,6 +136,14 @@ fn toml_to_args(tomldata: &str) -> Result<(Vec<OsString>, Vec<OsString>)> {
     Ok((args, cbmc_args))
 }
 
+fn unstable_entry(name: &String, value: &Value) -> Result<Option<OsString>> {
```
adpaco-aws, line 139: I don't know what this function is doing.

kani-compiler/src/kani_middle/mod.rs
```diff
@@ -50,6 +54,23 @@ pub fn check_crate_items(tcx: TyCtxt, ignore_asm: bool) {
     tcx.sess.abort_if_errors();
 }
 
+/// Check that all given items are supported and there's no misconfiguration.
+/// This method will exhaustively print any error / warning and it will abort at the end if any
+/// error was found.
+pub fn check_reachable_items(tcx: TyCtxt, queries: &QueryDb, items: &[MonoItem]) {
```
zhassan-aws, line 60: The function name and description are a bit too general. Do you foresee expanding the checks that this function does in the future? If not, I would suggest renaming it to something more specific, e.g. `check_reachable_unstable_items`.

kani-compiler/src/kani_middle/attributes.rs
zhassan-aws, line 174:
```suggestion
                            "use `-Z {}` to enable using this API.",
```

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -128,6 +147,55 @@ pub fn extract_harness_attributes(tcx: TyCtxt, def_id: DefId) -> Option<HarnessA
...
+                            err.note(format!("see issue {link} for more information"));
+                        }
+                        err.help(format!(
+                            "use `-Z {}` to enable use this API.",
+                            unstable_attr.feature
+                        ));
+                        err.emit();
+                    }
+                }
+                Err(msg) => {
```
zhassan-aws, line 180: Since this case is meant for developers and not users, you might want to change it to a panic.

kani-driver/src/args.rs
zhassan-aws: Perhaps
```suggestion
    /// Allow replacing certain items with stubs (mocks).
```

kani-driver/src/args.rs
```diff
@@ -343,6 +347,16 @@ pub enum ConcretePlaybackMode {
...
+    /// See (RFC-0002)[https://model-checking.github.io/kani/rfc/rfcs/0002-function-stubbing.html]
+    Stubbing,
+    /// Generate a C-like file equivalent to inputted program used for debugging purpose.
```
zhassan-aws, line 356:
```suggestion
    /// Generate a C-like file equivalent to input program used for debugging purpose.
```

kani-compiler/src/kani_middle/attributes.rs
celinval: Not according to the RFC, but we can change that.

kani-compiler/src/kani_middle/attributes.rs
celinval: Ok

kani-driver/src/args_toml.rs
celinval: Let me add a comment. But this is parsing the new table `#[package.metadata.kani.unstable]` and converting it into `-Z <unstable_feature>` argument.

kani-compiler/src/kani_middle/mod.rs
celinval: Yes, I was thinking that this will likely be extended in the future. But I'm OK renaming it for now if you prefer.

kani-compiler/src/kani_middle/attributes.rs
celinval: I thought about it, but my worry is that we don't have anyway of actually restricting the usage of `kani::unstable` in the user code. But I can try to come up with something.

kani-compiler/src/kani_middle/mod.rs
celinval: Because generic items have the same `def_id` and they will have the same attributes attached to them. The input to this function is working on the top of monomorphized items though, so we really only need to check the attributes once.

kani-compiler/src/kani_middle/attributes.rs
celinval: Great question... probably a typo or I might've changed the message and forgot to update the it. 🤷‍♀️

kani-compiler/src/kani_middle/attributes.rs
zhassan-aws: I think it's very unlikely for users to use it.

kani-compiler/src/kani_middle/mod.rs
zhassan-aws: Either way

kani-compiler/src/kani_middle/attributes.rs
celinval: For sure. It would be a mistake, but since it is public, I rather treat it as so. What if I add a debug assert?

zhassan-aws: Sure

kani-compiler/src/kani_middle/attributes.rs
celinval: That's fair... let me cleanup this a bit.

celinval: TBH, the main problem with this code is that we are not really validating the format of the attributes until the codegen stage. In this case, we will detect issues that may be from a dependency. I think the long term solution is to move the attribute handling to happen before codegen.

There are a few issues with that though. The main one for this case is that we don't even use `kani-compiler` today to build the sysroot. So doing this the "proper" way will actually take some time.

kani-compiler/src/kani_middle/attributes.rs
celinval: No, but I don't think it's worth to walk the AST just for this. I don't think the error that rustc emits is that much more friendly. Here are a few examples:
```
error: expected unsuffixed literal or identifier, found `<eof>`
 --> src/lib.rs:5:12
  |
5 | #[cfg(test =)]
  |            ^
```
or
```
error: expected one of `(`, `,`, `::`, or `=`, found `?`
 --> src/lib.rs:5:12
  |
5 | #[cfg(test ? aff)]
  |            ^ expected one of `(`, `,`, `::`, or `=`

```

kani-driver/src/args.rs
```diff
@@ -343,6 +347,16 @@ pub enum ConcretePlaybackMode {
...
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, ValueEnum, strum_macros::Display)]
+#[strum(serialize_all = "kebab-case")]
+pub enum UnstableFeatures {
+    /// Allow replacing certain items with stubs (mocks).
+    /// See [RFC-0002](https://model-checking.github.io/kani/rfc/rfcs/0002-function-stubbing.html)
+    Stubbing,
+    /// Generate a C-like file equivalent to input program used for debugging purpose.
+    GenC,
```
zhassan-aws, line 357: Are we planning to hook these up in a follow-up PR?

It might also be worth adding a variant for concrete playback.

zhassan-aws (approved):
LGTM. Adding a debug assert when the unstable annotation is not parseable would be nice, but not necessary in this PR.

kani-driver/src/args.rs
celinval: Yes, we need to move all the unstable features to this new model and deprecate the old `--enable-unstable`.

celinval:
> LGTM. Adding a debug assert when the unstable annotation is not parseable would be nice, but not necessary in this PR.

I just added in the last revision. :)

kani-compiler/src/kani_middle/attributes.rs
adpaco-aws: Agree, thanks for checking!

adpaco-aws (approved):
Thanks, @celinval !

PR #2389, opened 2023-04-18
Fix benchcomp path to Kani binary

karkhaz:
This commit fixes an error in setting the $PATH variable in the benchcomp perf regression config file. The script was working by accident because kani-perf.sh uses its own logic to find the Kani directory.

### Testing:

* How is this change tested? I tried running it with an intentionally-missing kani in one of the variants to ensure that it worked. (Previously the correct Kani was being used anyway, but this test ensures that the new behavior is correct)

* Is this a refactor change?

### Checklist
- [ x My PR is restricted to a single feature or bugfix

### DIFF
--- a/tools/benchcomp/configs/perf-regression.yaml
+++ b/tools/benchcomp/configs/perf-regression.yaml
@@ -10,13 +10,13 @@ variants:
   kani_new:
     config:
       directory: new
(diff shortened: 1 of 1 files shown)


celinval (approved):
TBH, I would just remove the code that sets the PATH since it has no effect on the result.

karkhaz:
done, thanks.

PR #2390, opened 2023-04-18
Fix error_on_regression only using last check

karkhaz:
This commit ensures that error_on_regression checker checks every one of the checks specified in the config file. Prior to this commit it would only check the last one.

The commit also increases the threshold that benchcomp uses to report a regression. Benchmarks that take less than 10 seconds (previously 2 seconds) are now ignored, and a 20% slowdown (previously 10%) is considered a regression.

### Checklist

### DIFF
--- a/tools/benchcomp/configs/perf-regression.yaml
+++ b/tools/benchcomp/configs/perf-regression.yaml
@@ -36,6 +36,6 @@ visualize:
         # benchmark has regressed if the lambda returns true.
         test: "lambda old, new: False if not old else not new"
       - metric: solver_runtime
-        test: "lambda old, new: False if new < 2 else new/old > 1.1"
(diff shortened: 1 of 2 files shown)


tools/benchcomp/configs/perf-regression.yaml
```diff
@@ -36,6 +36,6 @@ visualize:
         # benchmark has regressed if the lambda returns true.
         test: "lambda old, new: False if not old else not new"
       - metric: solver_runtime
-        test: "lambda old, new: False if new < 2 else new/old > 1.1"
+        test: "lambda old, new: False if new < 2 else new/old > 1.5"
```
qinheping, line 39: The new threshold 1.5 is not consistent as the description (20%). Which one is the right one?

qinheping (approved):

Issue #1286, opened 2022-06-16
MPSC Channel test case times out.


YoshikiTakashima:
Tracking issue

_Originally posted by @danielsn in https://github.com/model-checking/kani/pull/1285#discussion_r899169729_

Running 
```bash'
./scripts/kani mpsc.rs
```
on the following code times out as of Kani 0.4.0 on M1 Mac, cargo 1.63.0-nightly (85e457e15 2022-06-07). Note that `kani::unwind(1)` fails to verify, and `kani::unwind(2)` or not putting an unwind limit leads to timeout.
```rust
// mpsc.rs
use std::sync::mpsc::*;

static mut CELL: i32 = 0;

struct DropSetCELLToOne {}

impl Drop for DropSetCELLToOne {
    fn drop(&mut self) {
        unsafe {
            CELL = 1;
        }
    }
}

#[kani::unwind(2)]
#[kani::proof]
fn main() {
    {
        let (send, recv) = channel::<DropSetCELLToOne>();
        send.send(DropSetCELLToOne {}).unwrap();
        let _to_drop: DropSetCELLToOne = recv.recv().unwrap();
    }
    assert_eq!(unsafe { CELL }, 1, "Drop should be called");
}
```

tedinski:
Just to clarify: It times out _with_ the `unwind(2)`? Or it times out without it and seems weird to need it?

celinval:
Interesting, with the fix for #2164, this test passes on Mac but fails on Linux due to #1781. :(

Issue #1781, opened 2022-10-13
function call: parameter "pthread_key_create::destructor" type mismatch when hitting `thread::current()`


roypat:
When a kani is run with `--mir-linker --enable-unstable` and encounters `thread::current()` (or rather, the destructor of a thread object?) it yields the error message from the title. 

I tried this code:
```rust
fn main() {}

#[kani::proof]
fn crash() {
    std::thread::current();
}
```

using the following command line invocation:

```
cargo clean; cargo kani --mir-linker --enable-unstable
```

with Kani version: cargo-kani 0.12.0

I expected to see this happen: As `std::thread` interfaces with the OS API (pthread on linux) I'd have expected kani to complain about it reaching some function for which it does not have access to the source code. If I run kani without `--mir-linker`, this is indeed what happens. 

Instead, this happened: 

<details>

```bash
Checking harness crash...
CBMC 5.67.0 (cbmc-5.67.0)
CBMC version 5.67.0 (cbmc-5.67.0) 64-bit x86_64 linux
Reading GOTO program from file /home/ec2-user/kani-reproducer/target/x86_64-unknown-linux-gnu/debug/deps/cbmc-linked.for-crash.out
Generating GOTO Program
Adding CPROVER library (x86_64)
Removal of function pointers and virtual functions
Generic Property Instrumentation
Running with 16 object bits, 48 offset bits (user-specified)
Starting Bounded Model Checking
: function call: parameter "pthread_key_create::destructor" type mismatch:
got struct_tag
  * identifier: tag-_3583338355805481440
expected pointer
  * #source_location: 
    * file: <builtin-library-pthread_key_create>
    * line: 11
    * function: pthread_key_create
    * working_directory: /home/ec2-user/kani-reproducer/src
  * width: 64
  * #failed_symbol: pthread_key_create::destructor$object
  0: code
      * #source_location: 
        * file: <builtin-library-pthread_key_create>
        * line: 11
        * function: pthread_key_create
        * working_directory: /home/ec2-user/kani-reproducer/src
      * return_type: empty
          * #source_location: 
            * file: <builtin-library-pthread_key_create>
            * line: 11
            * function: pthread_key_create
            * working_directory: /home/ec2-user/kani-reproducer/src
      * parameters: 
        0: parameter
            * type: pointer
                * #source_location: 
                  * file: <builtin-library-pthread_key_create>
                  * line: 11
                  * function: pthread_key_create
[...]

PR #2395, opened 2023-04-20
Add size_of annotation to help CBMC's allocator

tautschnig:
### Description of changes: 

CBMC's heap allocator will produce objects that aren't just a byte array when a prior call to `size_of` was involved in computing the number of bytes to be allocated via malloc/calloc/realloc. This, in turn, permits type-safe member accesses to heap-allocated objects. This should speed up copying as well as all other operations on the resulting object (which will then not have to go through byte extract/byte update operations).

It also seems that there no longer is a "size_of" intrinsic.

### Resolved issues:

Fixes: #1286

### Related RFC:

n/a

### Call-outs:

n/a

### Testing:

* How is this change tested? Locally reviewed CBMC's `--program-only` output on some Vector examples.

### Checklist


cprover_bindings/src/goto_program/expr.rs
```diff
@@ -293,6 +294,10 @@ impl Expr {
         &self.value
     }
 
+    pub fn size_of_annotation(&self) -> &Option<Type> {
```
zhassan-aws, line 297: It's more idiomatic to return `Option<&Type>` as opposed to `&Option<Type>` (see https://users.rust-lang.org/t/what-is-the-preferred-way-to-return-option-fields/25100/5). You can use `as_ref` to do the conversion.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -611,7 +611,6 @@ impl<'tcx> GotocCtx<'tcx> {
                 loc,
             ),
             "simd_xor" => codegen_intrinsic_binop!(bitxor),
-            "size_of" => codegen_intrinsic_const!(),
```
zhassan-aws, line 614: Are we sure this is no longer possible (it still exists in `core`:  https://doc.rust-lang.org/src/core/intrinsics.rs.html#887). If so, I think it's better if we replace this with `unreachable!` similar to what we do for `copy_nonoverlapping`: https://github.com/model-checking/kani/blob/d9cf03a2ac1bf80f339ebec2edf465a43108f735/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs#L469 so that we get a panic at compile time as opposed to verification failure.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1267,11 +1266,12 @@ impl<'tcx> GotocCtx<'tcx> {
     /// This function computes the size and alignment of a dynamically-sized type.
     /// The implementations follows closely the SSA implementation found in
     /// `rustc_codegen_ssa::glue::size_and_align_of_dst`.
-    fn size_and_align_of_dst(&self, t: Ty<'tcx>, arg: Expr) -> SizeAlign {
+    fn size_and_align_of_dst(&mut self, t: Ty<'tcx>, arg: Expr) -> SizeAlign {
         let layout = self.layout_of(t);
         let usizet = Type::size_t();
         if !layout.is_unsized() {
-            let size = Expr::int_constant(layout.size.bytes_usize(), Type::size_t());
+            let size = Expr::int_constant(layout.size.bytes_usize(), Type::size_t())
```
zhassan-aws, line 1273: Can we add a comment to explain what this is doing and how the annotation helps CBMC?

cprover_bindings/src/goto_program/expr.rs
tautschnig: Thank you!

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
tautschnig: Good call, done!

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
tautschnig: I have added a comment in `expr.rs`.

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -41,6 +50,7 @@ pub struct Expr {
     value: Box<ExprValue>,
     typ: Type,
     location: Location,
+    size_of_annotation: Option<Type>,
```
adpaco-aws, line 53: Why do we need an additional `Option<Type>` here? We already have the type in `typ`, and if I'm not mistaken this hint is something you always want to have for aggregate types.

cprover_bindings/src/irep/to_irep.rs
```diff
@@ -162,6 +162,10 @@ impl ToIrep for Expr {
         } else {
             self.value().to_irep(mm).with_location(self.location(), mm).with_type(self.typ(), mm)
         }
+        .with_named_sub_option(
+            IrepId::CCSizeofType,
+            self.size_of_annotation().map(|ty| ty.to_irep(mm)),
+        )
```
adpaco-aws, line 168: My internal type-checker is complaining: I'd expect a `size_of_annotation` to return an unsigned number (i.e., a `usize`), but it returns a type instead. The irep ID is `CCSizeofType`, but you're encoding an `Option<&Type>`?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1267,11 +1267,12 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                .with_size_of_annotation(self.codegen_ty(t));
```
adpaco-aws, line 1275: Are there any other places where we need to add annotations?

adpaco-aws (commented):
Thanks for this detailed explanation, @tautschnig ! A few questions.

cprover_bindings/src/goto_program/expr.rs
tautschnig: No, the annotation really says (and I have now expanded the comment to say this): "... the expression is the result of computing `size_of(type)`."

cprover_bindings/src/irep/to_irep.rs
tautschnig: The actual size is the value of the expression, the annotation just tells you the size of which type was computed to arrive at that value.

tautschnig:
With thanks to @karkhaz for making this happen, and with the hope that this isn't just about some GitHub runner anomaly (but at least the numbers stack up with my local experiment):
```
  s2n-quic/quic/s2n-quic-core/slice::tests::vectored_copy_fuzz_test:
    variants:
      kani_new:
        metrics:
          solver_runtime: 74.86000000000001
          success: true
          symex_runtime: 7.85538
          verification_time: 93.35263
      kani_old:
        metrics:
          solver_runtime: 178.3941
          success: true
          symex_runtime: 7.67614
          verification_time: 196.657
```

adpaco-aws:
It's looking great, @tautschnig ! Just wondering if you think we should add any more tests to the performance regression or not. Is the `vectored_copy_fuzz_test` already in there?

zhassan-aws (approved):
Thanks @tautschnig!

One small point. The OP says:
> Fixes: https://github.com/model-checking/kani/issues/1781

But #1781 seems unrelated?

tautschnig:
> Thanks @tautschnig!
> 
> One small point. The OP says:
> 
> > Fixes: #1781
> 
> But #1781 seems unrelated?

My apologies, looks like I wrongly inferred this from https://github.com/model-checking/kani/issues/1286#issuecomment-1414676278. I just confirmed that (as expected) #1781 is _not_ fixed.

tautschnig:
> It's looking great, @tautschnig ! Just wondering if you think we should add any more tests to the performance regression or not. Is `vectored_copy_fuzz_test` already in there?

That test is already there.

PR #2396, opened 2023-04-20
Override std::ptr::align_offset

### LINKED ISSUE #2363: Improve symbolic execution performance when running against concrete values
**Requested feature:** When running Kani assess, many unit tests take a long time to run. Since these tests run with concrete inputs, I would expect the performance to be somewhat similar to MIRI.
**Use case:** Running tests to find undefined behavior + potential improvement to harnesses.
**Link to relevant documentation (Rust reference, Nomicon, RFC):**

**Test case:**

```rust
// TODO
```

tautschnig:
This hook intercepts calls to `std::ptr::align_offset<T>` as CBMC's memory model has no concept of alignment of allocations, so we would have to non-deterministically choose an alignment of the base pointer, add the pointer's offset to it, and then do the math that is done in `library/core/src/ptr/mod.rs`. Instead, we choose to always return either 0 when the pointer points to the beginning of an object (and, therefore, is necessarily aligned), or `usize::MAX`, per `align_offset`'s documentation, which states: "It is permissible for the implementation to always return usize::MAX. Only your algorithm’s performance can depend on getting a usable offset here, not its correctness."

Towards: #2363


tautschnig:
Perhaps @karkhaz:
1. The alleged performance regression (https://github.com/model-checking/kani/actions/runs/4753944159/jobs/8446153448?pr=2396) is the same as previously reported upon a merge into main: https://github.com/model-checking/kani/actions/runs/4747428362/jobs/8432323295. 
2. Would benchcomp also tell us about any performance improvements?

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -358,6 +358,41 @@ impl<'tcx> GotocHook<'tcx> for MemCmp {
...
+/// `library/core/src/ptr/mod.rs`. Instead, we choose to always return `usize::MAX`, per
+/// `align_offset`'s documentation, which states: "It is permissible for the implementation to
+/// always return usize::MAX. Only your algorithm’s performance can depend on getting a usable
+/// offset here, not its correctness."
+pub struct AlignOffset;
+
+impl<'tcx> GotocHook<'tcx> for AlignOffset {
+    fn hook_applies(&self, tcx: TyCtxt<'tcx>, instance: Instance<'tcx>) -> bool {
+        let name = with_no_trimmed_paths!(tcx.def_path_str(instance.def_id()));
+        name == "std::ptr::align_offset"
```
zhassan-aws, line 373: I think we should include `core::ptr::align_offset` as well.

zhassan-aws (commented):
Is there any potential unsoundness with this change?

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
celinval: Actually, they are two different methods. the `std` one is safe, while the `core` one is unsafe. Instead of using the name here, I would recommend using the `align_offset` lang_item instead, since it is way more reliable.

In this case, you would be overriding the `core` version, which is unsafe and it is UB to execute it with an input that is not power of 2. So please add a check for that.

celinval (dismissed):
I don't think this implementation is sound. The code should panic if the input is not a power of two.
For example, the following code [panic in Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=18a4945b51194b8f66d15a9d0f63caa4):

```rust
use std::mem::align_of;

fn main() {
  let x = 10;
  let ptr = &x as *const i32;
  let _ = ptr.align_offset(5);
}
```

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
tautschnig: The `lang_item` appears to no longer exist today and Miri has also changed their approach. Now using `core::ptr::align_offset` as all other functions eventually call into this one.

feliperodri:
I've pushed two commits here to get this unstuck (2daf1272, 8fa15a36). @tautschnig, the second one changes the semantics of your hook, the reasoning is below and in the commit message.

### The mechanical failures (2daf1272)

Three of the four red jobs had nothing to do with the feature:

- **`hooks.rs` had lost its trailing newline.** That fails the format check, which is the *first* step of `kani-regression.sh` — so all five `regression` jobs and `llbc-regression` died before running a single test. One missing byte was taking out six jobs.
- **`ptr` became unused** when the hook stopped inspecting the pointer, failing `clippy-check` under `-D warnings`.
- **The test contradicted the implementation.** `tests/kani/StdOverrides/align_offset.rs` still asserted `0` for a pointer at an object's base while the hook had been changed to return `usize::MAX` unconditionally.

I also merged current `main` (the branch was `behind`), and made `hook_applies` match `core::ptr::align_offset::<` as well as `std::ptr::align_offset::<`. Kani renders the path as `std::ptr` for an ordinary crate, but a `no_std` crate sees `core::ptr` and the hook would silently not apply, that's the durable half of @zhassan-aws's original request.

### The real blocker (8fa15a36)

The last commit before mine switched the hook to return `usize::MAX` unconditionally. That's sound, but it makes Kani report failures in correct code, and it's what broke `perf`:

```
Failed Checks: "contents must be aligned to cmsghdr"
 File: "quic/s2n-quic-platform/src/message/cmsg/decode.rs", line 136
```

s2n-quic does this:

```rust
debug_assert_eq!(
    cursor.align_offset(align_of::<cmsghdr>()),
    0,
    "contents must be aligned to cmsghdr"
);
```

The documentation does say callers must not depend on a usable offset *for correctness*, but real crates depend on it in debug assertions, and Kani honours those. Reduced to a standalone case, a buffer that is provably aligned:

```rust
#[repr(align(8))]
struct Aligned([u8; 32]);

#[kani::proof]
fn asserts_alignment_of_an_aligned_buffer() {
    let buf = Aligned([0u8; 32]);
    let cursor = buf.0.as_ptr();
    assert_eq!(cursor.align_offset(align_of::<u64>()), 0);
}
```

This fails with the hook and passes without it. I confirmed causation by toggling only the hook registration on one build. CI's own history agrees: `perf` **passed** on 32e07315 (the merge commit, with the original offset-based behaviour) and **failed** on 50f4a37a (the switch to unconditional `usize::MAX`).

So the hook now reports aligned when `__CPROVER_pointer_offset(ptr)` is a multiple of `align`, and `usize::MAX` otherwise, which `align_offset` explicitly permits. The offset is the only part of an address CBMC's memory model represents, so this is what the real implementation concludes under that model. That makes it a shortcut rather than a change in behaviour, which is the point of the hook.

The caveat is documented on the hook rather than papered over: it reports "aligned" for a pointer whose object base is less aligned than `align`. CBMC cannot represent that either way, so the real implementation does not discover it under this model either. @tautschnig, you know that model well, so if you think that's the wrong trade, the alternatives are to model alignment properly with a non-deterministic base (which defeats the performance purpose, and would make any
alignment assertion fail anyway) or to accept the false positives and adjust the s2n-quic expectations.

### Test

The previous test could not actually detect the hook: with alignment 1 every pointer is aligned, so the hook and the real implementation agree. It now checks offset 0 against alignment 4 (aligned, so `0`) and offset 1 against alignment 4, where the hook answers `usize::MAX` and the real implementation would answer `3`.

tests/kani/StdOverrides/align_offset.rs
```diff
@@ -0,0 +1,15 @@
...
+
+#[kani::proof]
+fn align_offset() {
+    let x = [10, 42];
+    let base_ptr = &x[0] as *const i32;
+    let base_alignment = base_ptr.align_offset(1);
+    assert_eq!(base_alignment, 0);
+    let offset_ptr = &x[1] as *const i32;
+    let offset_alignment = offset_ptr.align_offset(1);
+    assert_eq!(offset_alignment, usize::MAX);
```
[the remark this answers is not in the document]
feliperodri: This was applied in 50f4a37a, and it turned out to be the thing blocking the PR, so recording why it's been changed in 8fa15a36 rather than silently reverted.

The reasoning is right that returning `0` because `__CPROVER_pointer_offset(ptr) == 0` does not *prove* alignment to the requested `align`. But unconditional `usize::MAX` trades a theoretical unsoundness for a concrete false positive: `perf` began failing on aws/s2n-quic, which does

```rust
debug_assert_eq!(cursor.align_offset(align_of::<cmsghdr>()), 0, "contents must be aligned to cmsghdr");
```

Kani honours debug assertions, so this reports a failure in code that is provably fine. A `#[repr(align(8))]` buffer asserting `align_offset(align_of::<u64>()) == 0` fails with the hook and passes without it. CI confirms the attribution: `perf` passed on the commit before this suggestion was applied and failed on it.

The hook now answers from the pointer's offset — aligned when the offset is a multiple of `align`, `usize::MAX` otherwise. The key point the original comment misses is that the alternative isn't "conservative vs. optimistic": CBMC's memory model does not represent object base alignment at all, so the real implementation reaches the same conclusion under that model. Answering from the offset keeps the hook a shortcut for what CBMC would conclude anyway, instead of contradicting it.

The caveat is now written on the hook itself, since it's a real limitation and worth stating rather than assuming a future reader will rederive it.

PR #2398, opened 2023-04-20
Make cargo build-dev produce a proper exit code

tautschnig:
### Description of changes: 

Build failures of Kani-related components did not result in a non-zero exit code.

### Resolved issues:

n/a

### Related RFC:

n/a

### Call-outs:

n/a

### Testing:

* How is this change tested? Tested locally, see https://github.com/model-checking/kani/actions/runs/4753845746/jobs/8445926203 for an example that I had not previously noticed in my local builds.

### Checklist

### DIFF
--- a/tools/build-kani/src/main.rs
+++ b/tools/build-kani/src/main.rs
@@ -19,10 +19,7 @@ fn main() -> Result<()> {
     let args = parser::ArgParser::parse();
 
     match args.subcommand {
-        parser::Commands::BuildDev(build_parser) => {
-            build_lib();
-            build_bin(&build_parser.args);
-        }
+        parser::Commands::BuildDev(build_parser) => build_lib().and(build_bin(&build_parser.args)),
         parser::Commands::Bundle(bundle_parser) => {
             let version_string = bundle_parser.version;
             let kani_string = format!("kani-{version_string}");
@@ -45,10 +42,10 @@ fn main() -> Result<()> {
             std::fs::remove_dir_all(dir)?;
 
             println!("\nSuccessfully built release bundle: {bundle_name}");
+
+            Ok(())
         }
     }
-
-    Ok(())
 }
 
 /// Ensures everything is good to go before we begin to build the release bundle.
@@ -70,10 +67,9 @@ fn prebundle(dir: &Path) -> Result<()> {
     }
 
     // Before we begin, ensure Kani is built successfully in release mode.
-    build_bin(&["--release"]);
-    // And that libraries have been built too.
-    build_lib();
-    Ok(())
+    build_bin(&["--release"])
+        // And that libraries have been built too.
+        .and(build_lib())
 }
 
 /// Copy Kani files into `dir`
--- a/tools/build-kani/src/sysroot.rs
+++ b/tools/build-kani/src/sysroot.rs
@@ -12,6 +12,7 @@
 //! Note: We don't cross-compile. Target is the same as the host.
 
 use crate::{cp, AutoRun};
+use anyhow::{bail, format_err, Result};
 use cargo_metadata::{Artifact, Message};
 use std::ffi::OsStr;
 use std::fs;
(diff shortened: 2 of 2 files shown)


tools/build-kani/src/sysroot.rs
```diff
@@ -106,7 +106,12 @@ pub fn build_lib() {
 
     // Collect the build artifacts.
     let artifacts = build_artifacts(&mut cmd);
-    let _ = cmd.wait().expect("Couldn't get cargo's exit status");
+    let exit_status = cmd.wait().expect("Couldn't get cargo's exit status");
+    // exit_ok is an experimental API where we could do .exit_ok().expect("Build failed")
+    // Until it is stable:
+    if !exit_status.success() {
+        panic!("Build failed");
```
adpaco-aws, line 113: Can you please extend this message as:
```suggestion
        panic!("Build failed because `cargo build` didn't return a successful exit code");
```
or something similar. This information may help in debugging build issues.

tools/build-kani/src/sysroot.rs
```diff
@@ -106,7 +106,12 @@ pub fn build_lib() {
 
     // Collect the build artifacts.
     let artifacts = build_artifacts(&mut cmd);
-    let _ = cmd.wait().expect("Couldn't get cargo's exit status");
+    let exit_status = cmd.wait().expect("Couldn't get cargo's exit status");
+    // exit_ok is an experimental API where we could do .exit_ok().expect("Build failed")
+    // Until it is stable:
```
adpaco-aws, line 111: Please backticks when writing code or function names in comments, e.g.:
```suggestion
    // `exit_ok` is an experimental API where we could do `.exit_ok().expect("Build failed")`
    // Until it is stable:
```
Regardless, you could just say we could do the same with `exit_ok` (i.e., without the example) but we avoid it since it's experimental.

adpaco-aws (approved):
Thanks, @tautschnig ! Just a couple of minor comments.

If I'm not mistaken, this will prevent this error that @karkhaz had a while ago:
```
error: extern location for std does not exist: /home/runner/work/kani/kani/target/kani/lib/libstd.rlib
```

tools/build-kani/src/sysroot.rs
tautschnig: Thanks, done.

tools/build-kani/src/sysroot.rs
tautschnig: Done.

tools/build-kani/src/sysroot.rs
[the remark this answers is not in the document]
zhassan-aws: IMHO, the current usage of `expect` is for very unusual situations (I/O related), e.g. failing to spawn a process, failing to wait for a child process, failing to create a directory, etc. The case of a non-zero exit status for `cargo build` is not as unusual as it could happen due to an error in the Kani library during development. Thus returning an error here seems more appropriate.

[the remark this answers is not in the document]
tautschnig: Fair point, this now actually makes for a much more reasonable behaviour (such as no longer suggesting the use of RUST_BACKTRACE for debugging...). Fixed, but I'd appreciate you taking another look whether this is the most appropriate way to do it.

tools/build-kani/src/main.rs
```diff
@@ -19,10 +19,7 @@ fn main() -> Result<()> {
     let args = parser::ArgParser::parse();
 
     match args.subcommand {
-        parser::Commands::BuildDev(build_parser) => {
-            build_lib();
-            build_bin(&build_parser.args);
-        }
+        parser::Commands::BuildDev(build_parser) => build_lib().and(build_bin(&build_parser.args)),
```
zhassan-aws, line 22: I would've used the [question mark operator](https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html):
```rust
build_lib()?;
build_bin(&build_parser.args);
```
but your version looks cleaner.

zhassan-aws (approved):
Looks good.

tools/build-kani/src/main.rs
zhassan-aws: If you don't want to modify `build_bin`, you can use `and_then` instead of `and`:
```rust
build_lib().and_then(|_| build_bin(&build_parser.args))
```

tautschnig: I actually tried the above at first, but this wouldn't compile.

PR #2399, opened 2023-04-20
Allow excluding packages from verification with `--exclude`

### LINKED ISSUE #2375: Feature request: exclude crates from workspace at build time
## Requested feature

When building a workspace under kani, add the option to exclude certain crates from the build. Right now this functionality can be achieved by providing a list of the crates that should be built with kani and excluding the target crates from that list.
```
cargo kani -p crate1 crate2 crate3
```

This is not feasible to use when you have a workspace with lots of crates, out of which you want to exclude a comparatively small number. Instead, it would be nice to be able to specify a list of crates to exclude.

```
cargo kani --exclude test-crate
```

## Use case

We are using kani in our project where we have 17 crates in one workspace. Out of these crates 2 binary crates are only relevant for Rust integration tests and we don't want to build them under kani. It is not feasible for us to pass 15 crates to the `--package` option, so what we do instead is that we use `cfg(kani)` to define a dummy main for the test related crates. This is making the code harder to read that it would be necessary, and it would be nice to exclude the test crates directly at build time instead.

## Link to relevant documentation (Rust reference, Nomicon, RFC)

Cargo also defines a `--exclude` flag: https://doc.rust-lang.org/cargo/commands/cargo-build.html#package-selection

adpaco-aws:
### Description of changes: 

This PR adds a `--exclude` option to Kani and allows excluding packages from verification (requested in #2375). It also adds validation for the packages passed to `--package` and `--exclude` so we print a user-friendly error with all unknown packages, instead of using `expect` as we were doing for arguments to `--package`.

### Resolved issues:

Resolves #2375 

### Testing:

* How is this change tested? Existing tests + 4 new ones.

### Checklist


kani-driver/src/args.rs
```diff
@@ -314,9 +315,14 @@ pub struct CargoArgs {
     /// Build all packages in the workspace
     #[arg(long)]
     pub workspace: bool,
+
     /// Run Kani on the specified packages.
     #[arg(long, short, conflicts_with("workspace"), num_args(1..))]
     pub package: Vec<String>,
+
+    /// Run Kani on the specified packages.
```
zhassan-aws, line 323: Message should be "Exclude the specified packages".

kani-driver/src/call_cargo.rs
```diff
@@ -103,8 +103,13 @@ impl KaniSession {
         let mut pkg_args: Vec<String> = vec![];
         pkg_args.extend(["--".to_string(), self.reachability_arg()]);
 
+        let packages_to_verify = packages_to_verify(&self.args, &metadata);
+        if let Err(package_error) = packages_to_verify {
```
zhassan-aws, line 107: Why not use question mark operator?
```rust
        let packages_to_verify = packages_to_verify(&self.args, &metadata)?;
```

tests/cargo-ui/ws-package-exclude-unknown/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+# Copyright Kani Contributors
```
zhassan-aws, line 1: The tests seem to be missing the expected files.

zhassan-aws (commented):
Thanks @adpaco-aws!

> because it doesn't require --workspace (but it's implied anyway)

I'd prefer if we make `--workspace` required to keep the same interface as `cargo build/test`.

kani-driver/src/call_cargo.rs
adpaco-aws: Not sure what happened here, sorry about that.

kani-driver/src/args.rs
adpaco-aws: Done.

tests/cargo-ui/ws-package-exclude-unknown/Cargo.toml
adpaco-aws: Good catch! Added them.

adpaco-aws:
> I'd prefer if we make `--workspace` required to keep the same interface as `cargo build/test`.

Thanks, it's required now.

kani-driver/src/call_cargo.rs
```diff
@@ -261,35 +261,57 @@ fn print_msg(diagnostic: &Diagnostic, use_rendered: bool) -> Result<()> {
...
 ///     See <https://github.com/rust-lang/cargo/issues/8033>.
-fn packages_to_verify<'b>(args: &KaniArgs, metadata: &'b Metadata) -> Vec<&'b Package> {
-    debug!(package_selection=?args.cargo.package, workspace=args.cargo.workspace, "packages_to_verify args");
+/// In addition, if either `--package <pkg>` or `--exclude <pkg>` is given,
+/// validate that `<pkg>` is a package name in the workspace, or return an error
+/// otherwise.
+fn packages_to_verify<'b>(args: &KaniArgs, metadata: &'b Metadata) -> Result<Vec<&'b Package>> {
+    debug!(package_selection=?args.cargo.package, package_exclusion=?args.cargo.exclude, workspace=args.cargo.workspace, "packages_to_verify args");
     let packages = if !args.cargo.package.is_empty() {
+        validate_package_names(args.cargo.package.clone(), metadata.packages.clone())?;
```
zhassan-aws, line 298: Nit: the two arguments can be passed by reference (borrow) to avoid cloning them.

tests/cargo-ui/ws-package-exclude-unknown/src/main.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+fn main() {}
+
+#[kani::proof]
+fn harness_in_ws_package() {
+    assert!(1 + 1 == 2);
```
zhassan-aws, line 8: Perhaps additionally make the assertion incorrect (e.g. `1 + 1 == 3`) so that verification fails if the package is incorrectly excluded.

tests/cargo-ui/ws-package-select/src/main.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+fn main() {}
+
+#[kani::proof]
+fn harness_in_ws_package() {
+    assert!(1 + 1 == 2);
```
zhassan-aws, line 8: Same here.

zhassan-aws (approved):

kani-driver/src/call_cargo.rs
adpaco-aws: Thanks!

tests/cargo-ui/ws-package-exclude-unknown/src/main.rs
adpaco-aws: That's a great idea, thanks!

PR #2404, opened 2023-04-24
Implement `kani::Arbitrary` for `Box<T>`

### LINKED ISSUE #2403: Feature request: Implement `kani::Arbitrary` for `Box<T>`
Requested feature: Implement `kani::Arbitrary` for `Box<T>`
Use case: I'm experimenting with user code that defines a struct similar to:
```rust
struct Foo<T> {
    boxed: Box<T>,
}
```
If I attempt to `#[derive(kani::Arbitrary)]` on that struct, I get this error:

```
error[E0277]: the trait bound `std::boxed::Box<T>: kani::Arbitrary` is not satisfied
   --> test.rs:9:5
    |
9   |     boxed: Box<T>,
    |     ^^^^^^^^^^^^^ the trait `kani::Arbitrary` is not implemented for `std::boxed::Box<T>`
    |
    = help: the following other types implement trait `kani::Arbitrary`:
              ()
              (A, B)
              (A, B, C)
              (A, B, C, D)
              (A, B, C, D, E)
              (A, B, C, D, E, F)
              (A, B, C, D, E, F, G)
              (A, B, C, D, E, F, G, H)
            and 39 others
note: required by a bound in `kani::any`
   --> /home/ubuntu/kani/library/kani/src/lib.rs:114:15
    |
114 | pub fn any<T: Arbitrary>() -> T {
    |               ^^^^^^^^^ required by this bound in `any`

```

Since `Box<T>` comes from Rust, the Kani library should implement `Arbitrary` for it (similar to #2224 / #2225).

adpaco-aws:
### Description of changes: 

This PR implements `Arbitrary` for `Box<T>`. This is required for structs that `#[derive(kani::Arbitrary)]` and have members with `Box<T>` types.

### Resolved issues:

Resolves #2403 

### Call-outs:

 - I haven't thought too much about the best way to test these changes, so I'm open to suggestions.
 - The same-purpose tests added in #2225 were in the `ui` folder. I moved them to `expected` because that seems to be a better place (no UI testing is done in those tests).

### Testing:

* How is this change tested? Existing tests + new one.

### Checklist


library/kani/src/arbitrary.rs
```diff
@@ -161,3 +161,12 @@ impl Arbitrary for std::marker::PhantomPinned {
         PhantomPinned
     }
 }
+
+impl<T> Arbitrary for std::boxed::Box<T>
+where
+    T: Arbitrary,
+{
+    fn any() -> Self {
+        Box::new(T::any())
```
zhassan-aws, line 170: Does this cover all possible states a `Box` can have? What about a box created using `Box::new_uninit`?

tests/expected/derive-arbitrary/box/test.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[derive(kani::Arbitrary)]
+struct Foo<T> {
+    boxed: Box<T>,
+}
+
+#[kani::proof]
+fn main() {
+    let foo: Foo<i32> = kani::any();
+    kani::cover!(*foo.boxed >= i32::MIN && *foo.boxed <= i32::MAX);
```
zhassan-aws, line 15: It might be better to cover specific corner values, e.g. `kani::cover!(*foo.boxed == i32::MIN)`. Same for 0 and `i32::MAX`.

zhassan-aws (commented):
I'm concerned about the soundness of the `Arbitrary` implementation. We need to make sure the implementation covers all possible states a box can be in. I'm a bit unclear on whether this is the case with the current implementation.

library/kani/src/arbitrary.rs
celinval: `Box::new_uninit` wouldn't fit this signature though, since T would have to be `MaybeUninit<U>`

celinval (approved):

tests/expected/derive-arbitrary/box/test.rs
adpaco-aws: Done: split this cover statement into the three cases you mentioned.

Issue #1483, opened 2022-08-09
Replace remaining usage of old arithmetic overflow operations


zhassan-aws:
In https://github.com/model-checking/kani/pull/1426, new methods for arithmetic overflow operations were added that rely on CBMC's new, more efficient operations: `add_overflow_result`, `sub_overflow_result`, and `mul_overflow_result`. There are still a few usages of the old methods, e.g. `add_overflow` in `kani_compiler::codegen_cprover_gotoc::codegen::intrinsic::GotocCtx::codegen_offset`. We should switch all the usages to the new methods, and delete the old methods.

PR #2405, opened 2023-04-25
Use optimized overflow operation everywhere

celinval:
### Description of changes: 

This is related to the performance degradation we are seeing in the toolchain upgrade: https://github.com/model-checking/kani/pull/2293

### Resolved issues:

Towards https://github.com/model-checking/kani/issues/1483. There are a few remaining instances of add_overflow and mul_overflow in intrinsics.rs.

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

This PR has a side effect of removing redundant checks inserted by CBMC for checked arithmetic operations.

### Testing:

* How is this change tested? Existing tests and manual tests with my other branch where I'm doing the upgrade

### Checklist


celinval:
It looks like the performance is not always better... :(

celinval:
I do wonder if this proof is also somewhat unstable. I noticed that in my other PR (#2407) that had minimum changes to the model, there was also a performance degradation of 47% for the same test.

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -1448,6 +1456,7 @@ impl Expr {
 
     /// `ArithmeticOverflowResult r; >>>r.overflowed = builtin_sub_overflow(self, e, &r.result)<<<`
     pub fn mul_overflow(self, e: Expr) -> ArithmeticOverflowResult {
+        // TODO: Should we always use the one below?
         let result = self.clone().mul(e.clone());
         let overflowed = self.mul_overflow_p(e);
         ArithmeticOverflowResult { result, overflowed }
```
tautschnig, line 1462: I'd answer that "TODO" with a "yes, unless there really is a need to only compute the Boolean would-this-operation-overflow result." That's not the case here, which does the multiplication, so that's likely just a left-over from the time where we didn't yet have the OverflowResult* expressions. We should get rid of this.

tests/ui/cbmc_checks/signed-overflow/expected
```diff
@@ -7,12 +7,5 @@ Failed Checks: attempt to calculate the remainder with a divisor of zero
...
 Failed Checks: attempt to shift right with overflow
-Failed Checks: arithmetic overflow on signed addition
-Failed Checks: arithmetic overflow on signed subtraction
-Failed Checks: arithmetic overflow on signed multiplication
-Failed Checks: shift distance is negative
-Failed Checks: shift distance too large
-Failed Checks: shift operand is negative
 Failed Checks: arithmetic overflow on signed shl
-Failed Checks: shift distance is negative
-Failed Checks: shift distance too large
```
tautschnig, line 18: Isn't that absence of failing checks concerning?

tautschnig (changes requested):
I'd like to ask for an explanation as to why the no-longer-failing checks are ok.

cprover_bindings/src/goto_program/expr.rs
```diff
@@ -1403,9 +1403,9 @@ impl Expr {
...
     }
 
-    /// Uses CBMC's add-with-overflow operation that performs a single addition
+    /// Uses CBMC's [binop]-with-overflow operation that performs a single addition
```
zhassan-aws, line 1406:
```suggestion
    /// Uses CBMC's [binop]-with-overflow operation that performs a single arithmetic
```

cprover_bindings/src/goto_program/expr.rs
zhassan-aws: Yes, the intention was to replace all the usage of `mul_overflow` with `mul_overflow_result` and get rid of `mul_overflow`.

tests/ui/cbmc_checks/signed-overflow/expected
zhassan-aws: I think those checks were redundant, i.e. each failure condition was being checked twice.

zhassan-aws (approved):
This has been on my to-do list for too long. Thanks for taking care of it @celinval!

You should add in the PR description "Towards https://github.com/model-checking/kani/issues/1483". There are a few remaining instances of `add_overflow` and `mul_overflow` in `intrinsics.rs` (and possibly elsewhere).

tests/ui/cbmc_checks/signed-overflow/expected
celinval: Sorry, I should've added that to the call-out. Yes, @zhassan-aws is correct that these checks were redundant.

These operations that we changed are checked operations, which means that the codegen adds an overflow check already. Since the implementation does not invoke the arithmetic operation directly, CBMC doesn't add its own overflow check anymore.

That said, I think I deleted some of the shift errors by accident. They shouldn't be affected.

celinval:
> I'd like to ask for an explanation as to why the no-longer-failing checks are ok.

That's fair! My bad, I should've added an explanation before. I added it as part of an answer to one of your question and addressed the issue in the call-out section. In a nutshell, the arithmetic check that was being added by CBMC was redundant.

tests/ui/cbmc_checks/signed-overflow/expected
tautschnig: Ok, thanks, I should have looked at more context in that file (which has all those codegen-added overflow checks also as "Failed Checks".

Issue #1855, opened 2022-11-04
Per-harness code generation


celinval:
Requested feature: Generate code on a per-harness basis instead of aggregating the reachable code for all harnesses.
Use case: Improve the verification performance since `goto-instrument` may not be able to slice some stuff that is actually unreachable. See #1659 for an example on how adding a second harness can interfere with the time of the first.
Link to relevant documentation (Rust reference, Nomicon, RFC):

PR #2406, opened 2023-04-26
Upgrade the toolchain to nightly-2023-04-16 

### LINKED ISSUE #560: Structure mismatch in "Generics - Phantom type parameters" example

### LINKED ISSUE #2383: Update rust toolchain version to nightly-2023-04-16
We are running a bit late on our toolchain version, so we have some catching up to do.

celinval:
### Description of changes: 
Update the toolchain to use nightly-2023-04-16. Changes were related to the following changes to the toolchain:

- https://github.com/rust-lang/rust/pull/108944
- https://github.com/rust-lang/rust/pull/108148
- https://github.com/rust-lang/rust/pull/108471
- https://github.com/rust-lang/rust/pull/109358
- https://github.com/rust-lang/rust/pull/109849
- https://github.com/rust-lang/rust/pull/109819
- https://github.com/rust-lang/rust/pull/109718
- https://github.com/rust-lang/rust/pull/109092
- https://github.com/rust-lang/rust/pull/108856
- https://github.com/rust-lang/rust/pull/105613
- https://github.com/rust-lang/rust/pull/103042
- https://github.com/rust-lang/rust/pull/109716
- https://github.com/rust-lang/rust/pull/108340
- https://github.com/rust-lang/rust/pull/102906
- https://github.com/rust-lang/rust/pull/98112
- https://github.com/rust-lang/rust/pull/108080

### Resolved issues:

Resolves #2383 
Resolves #560

### Call-outs:

The checked shift right / left operations were optimized in this commit: https://github.com/rust-lang/rust/pull/108282. The Rust compiler now adds an overflow check before actually performing the shift. With that, some of the CBMC shift overflow checks (which were already redundant) no longer fail. I updated the tests to either remove the expectation or to use the `unchecked_[shl|shr]`. However, these checks take `u32` for the shift distance, hence the check for negative distance is probably useless now.

### Testing:

* How is this change tested? Existing tests + manually executed `--visualize` to check the trace

* Is this a refactor change?

### Checklist


cprover_bindings/src/goto_program/expr.rs
```diff
@@ -131,7 +131,7 @@ pub enum ExprValue {
     StringConstant {
         s: InternedString,
     },
-    /// Struct initializer  
+    /// Struct initializer
```
tautschnig, line 134: I'm all for whitespace cleanup, but can this perhaps go in a PR of its own?

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -154,11 +154,11 @@ impl<'tcx> GotocCtx<'tcx> {
         if self.queries.get_check_assertion_reachability() {
             // Generate a unique ID for the assert
             let assert_id = self.next_check_id();
-            // Generate a message for the reachability check that includes the unique ID
-            let reach_msg = assert_id.clone();
             // Also add the unique ID as a prefix to the assert message so that it can be
             // easily paired with the reachability check
             let msg = GotocCtx::add_prefix_to_msg(&msg, &assert_id);
+            // Generate a message for the reachability check that includes the unique ID
+            let reach_msg = assert_id;
```
tautschnig, line 161: Hmm, yes, getting rid of a `.clone()` is nice, but seems unrelated to toolchain updates?

cprover_bindings/src/goto_program/expr.rs
celinval: Sure... my IDE just does that automatically

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
celinval: I agree that it's not ideal, but also not worth the review cycle to remove these trivial changes, don't you think? Anyway, I'll revert these for now.

cprover_bindings/src/goto_program/expr.rs
celinval: btw, this change will go away once I merge #2405

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
tautschnig: I do not at all intend to block this PR over this, it's just (like the whitespace change) something that tripped me up while reading the changes. Might be a matter of personal preference that I'd rather do 5 PRs that each do one thing.

tests/expected/array/main.rs
```diff
@@ -6,11 +6,17 @@ fn foo(x: [i32; 5]) -> [i32; 2] {
     [x[0], x[1]]
 }
 
+/// Generate an out-of-bound index with the given length.
+/// We use a function so the constant propagation
```
zhassan-aws, line 10: Incomplete comment

tests/ui/should-panic-attribute/unexpected-failures/test.rs
```diff
@@ -3,10 +3,11 @@
 
 //! Checks that verfication fails when `#[kani::should_panic]` is used but not
 //! all failures encountered are panics.
+#![feature(unchecked_math)]
 
 fn trigger_overflow() {
     let x: u32 = kani::any();
-    let _ = 42 << x;
+    let _ = unsafe { 42u32.unchecked_shl(x) };
```
zhassan-aws, line 10: Why this change?

zhassan-aws (approved):
LGTM

tests/ui/should-panic-attribute/unexpected-failures/test.rs
celinval: See the call-out for details. The rust compiler changed how shift is encoded.

celinval:
Great... 😕

It looks like the rust compiler API to retrieve attributes no longer works for non-local items. @oli-obk, do you know if this was by design? I'm happy to try to submit a fix if needed (I should also try the latest rustc to see if it hasn't been fixed yet).

celinval:
> Great... confused
> 
> It looks like the rust compiler API to retrieve attributes no longer works for non-local items. @oli-obk, do you know if this was by design? I'm happy to try to submit a fix if needed (I should also try the latest rustc to see if it hasn't been fixed yet).

By the way, the error we are getting looks like:
```
thread 'rustc' panicked at 'assertion failed: `(left == right)`
  left: `ClosureExpr`,
 right: `Ctor`', compiler/rustc_metadata/src/rmeta/decoder.rs:1127:17
stack backtrace:
   0: rust_begin_unwind
             at /rustc/5cdb7886a5ece816864fab177f0c266ad4dd5358/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/5cdb7886a5ece816864fab177f0c266ad4dd5358/library/core/src/panicking.rs:67:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed::<rustc_hir::definitions::DefPathData, rustc_hir::definitions::DefPathData>
   4: <rustc_metadata::creader::CrateMetadataRef>::get_item_attrs
   5: rustc_metadata::rmeta::decoder::cstore_impl::provide_extern::item_attrs
   6: rustc_query_system::query::plumbing::try_execute_query::<rustc_query_impl::queries::item_attrs, rustc_query_impl::plumbing::QueryCtxt>
   7: <rustc_query_impl::Queries as rustc_middle::ty::query::QueryEngine>::item_attrs
   8: <rustc_middle::ty::context::TyCtxt>::get_attrs_unchecked
```

oli-obk:
> It looks like the rust compiler API to retrieve attributes no longer works for non-local items

It looks to me like this is specifically about closures, as we can definitely access attributes from non local items in general.

I don't know how that happened, none of the changes look like they could be at fault. https://github.com/rust-lang/rust/pull/108944 is the most suspicious one, but it makes no sense for it to affect attributes.

I'll need to investigate with a computer, as it's too hard to do on mobile (I'm on sick leave). It may be until next week until I can get to it.

celinval:
Just to give an update, the issue seems related to the unwinding of a recursive drop logic. I am trying to figure out the culprit. I am investigating if the issue could be related to an update to the standard library

oli-obk:
I only see the perf CI failing. How do I reproduce this failure you're seeing?

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -53,7 +53,7 @@ impl<'tcx> GotocCtx<'tcx> {
                 var_type,
                 self.codegen_span(&ldata.source_info.span),
             )
-            .with_is_hidden(!ldata.is_user_variable())
+            .with_is_hidden(!self.is_user_variable(&lc))
```
oli-obk, line 56: Could this change be causing your issues somehow? Since you're getting this information from debug information now, it may be different from what you had before, so some logic continues to try to extract attributes from closures when it wasn't before? Or did it actually work before to extract attributes from closures?

celinval: I don't think this is related, since this code runs after the code that check for attributes.



> Or did it actually work before to extract attributes from closures?

Yes, this code actually used to run for closures before without crashing. I just double checked our main branch. That said, it always returns an empty list, so I don't think we are missing much. It was just cleaner before.

celinval:
I've done some analysis with the patch from #2433, and I believe this issue might be related to how we encode `drop_in_place`, and how CBMC identifies candidate function pointers. This is very similar to the issue described in #1855. 

To test this hypothesis, I commented out all the other harnesses from the s2n_quic code and ran:
```
cargo kani --tests --harness sync::spsc::tests::alloc_test
```
on my local machine, the verification takes about 42s.

oli-obk:
I found the issue. https://github.com/rust-lang/rust/pull/109765 added `should_encode_attrs` and put `Closure` into the `false` arm.

While this makes sense for rustc, it's obviously an issue for analysis tools like kani. Unfortunately it's also performance relevant. I'll open a PR to see if there's a perf impact, if there is, we need to figure out how to let kani decide to encode that (or store these attributes in some other way).

oli-obk:
The latest nightly should include https://github.com/rust-lang/rust/pull/111381 now, if there are similar failures again in the future, you can use that PR to look up where to edit the match to keep more attributes in metadata

celinval:
New update: My initial hypothesis was actually wrong. Sorry! The reachability per harness made a huge difference in my preliminary analysis because I had rewritten the `alloc_test` harness to invoke Kani's functions directly instead of via Bolero. In that case, the Bolero logic from other harnesses were affecting the analysis.

However, when rolling back my changes, the `alloc_test` usage of Bolero was enough to bring all the `drop_in_places` that are interfering with CBMC's analysis.

That said, @zhassan-aws helped me debug this issue further and understand why Bolero was having such a big impact in our analysis with this upgrade. We narrowed down to an issue with how `if cfg!()` statements are handled by this new compiler version.

We first noticed this issue during this upgrade and even created https://github.com/model-checking/kani/pull/2407 to fix it for our local crates.

I have now created https://github.com/camshaft/bolero/pull/148 to update Bolero, which will hopefully unblock this PR.

PR #2408, opened 2023-04-26
Explicitly pass out_file to benchcomp visualizers

karkhaz:
This commit adds an `out_file` to the benchcomp visualizes that emit output (currently just one visualization, dump_yaml). This is to allow different visualizations to be written to different output files, including stdout.

The intention (for a future commit) is that CI will save certain visualizations as artifacts, which users can then download.

### Testing:

* How is this change tested? Ran it locally

* Is this a refactor change?

### Checklist

### DIFF
--- a/tools/benchcomp/benchcomp/visualizers/__init__.py
+++ b/tools/benchcomp/benchcomp/visualizers/__init__.py
@@ -6,6 +6,7 @@ import dataclasses
 
 import yaml
 
+import benchcomp
 import benchcomp.visualizers.utils as viz_utils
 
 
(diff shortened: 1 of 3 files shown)


tautschnig:
> This is not currently possible because the Action will fail if there were any regressions, meaning that the artifact would not be subsequently uploaded. This will be addressed in a future commit that allows particular visualizations to run.

I don't think this statement is true: https://github.com/actions/upload-artifact will happily upload artifacts even if an earlier step failed (there even is provision to _only_ do so when an earlier step failed, see their docs). We make good use of this in https://github.com/diffblue/cbmc/blob/develop/.github/workflows/build-and-test-Linux.yaml, which will upload files to debug build failures.

tautschnig (approved):

karkhaz:
You're right, thank you. I've amended the commit message.

PR #2412, opened 2023-04-27
Refactor slow benchcomp regression tests

karkhaz:
This commit moves the regression tests that run Kani into a separate suite that runs after the other regression tests, so that users get quick feedback about any of the fast tests that fail.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


celinval (commented):
Question, where is the order of tests run being defined? Is that why you added a "Z" to the name of the second suite?

PR #2425, opened 2023-05-03
The unstable `--c-lib` option now requires `-Z c-ffi` to enable C-FFI support

celinval:
### Description of changes: 

This will also give visibility to the compiler to whether this feature is enabled or not which is needed to avoid type mismatch issue (https://github.com/model-checking/kani/issues/1781).

### Resolved issues:

Related to #1781

### Related RFC:

#2281 

### Call-outs:

1. I am not sure I should use `-Z unstable-options` here. This seems like over-killing having to use `-Z c-ffi` and `-Z unstable-options`.
2. While updating the tests, I notice that the test for #3 wasn't really testing anything. So I removed it and replaced it with a minimum test for the `Option<&T>` -> `*T` decay.

### Testing:

### Checklist


kani-driver/src/args.rs
```diff
@@ -361,6 +361,8 @@ pub enum UnstableFeatures {
     Stubbing,
     /// Generate a C-like file equivalent to input program used for debugging purpose.
     GenC,
+    /// Allow Kani to link against C code.
+    CFfi,
```
adpaco-aws, line 365: I'm wondering if we should just have a generic FFI option (i.e., not restricted to C) as the gate for all FFI-related options. What do you think?

celinval: Yeah, but does Kani work at all with any other type of FFI beyond C?

tests/kani/ForeignItems/fixme_option_ref.rs
```diff
@@ -0,0 +1,26 @@
...
+//! Check how Kani handles declaration of Nullable reference (Option<&T>).
+//! See <https://github.com/model-checking/kani/issues/2152> for more details.
+
+extern "C" {
+    // In rust, you say nullable pointer by using option of reference.
+    // Rust guarantees that this has the bitwise represntation
+    // Some(&x) => &x;
+    // None => NULL;
+    // FIXME: we need to notice when this happens and do a bitcast, or C is unhappy
+    // https://github.com/model-checking/kani/issues/3
```
adpaco-aws, line 14:
```suggestion
    // In Rust, you say nullable pointer by using option of reference.
    // Rust guarantees that this has the bitwise representation:
    //  - Some(&x) => &x;
    //  - None => NULL;
    // FIXME: we need to notice when this happens and do a bitcast, or C is unhappy
    // <https://github.com/model-checking/kani/issues/3>
```

adpaco-aws (approved):

kani-driver/src/args.rs
celinval: Talked offline...

PR #2427, opened 2023-05-03
Get rid of the legacy mode

celinval:
### Description of changes: 

Get rid of the Legacy mode and modify the std regression. Since we produce goto now directly, we can afford using `pub_fns`. There's just a small performance penalty.

### Resolved issues:
N/A

### Related RFC:
N/A

### Testing:

### Checklist


scripts/std-lib-regression.sh
```diff
@@ -70,9 +70,10 @@ export RUSTC_LOG=error
 RUST_FLAGS=(
     "--kani-compiler"
     "-Cpanic=abort"
+    "-Zalways-encode-mir"
```
zhassan-aws, line 73: What does this do?

celinval: This stores the MIR body of every function as part of the `.rlib` file. We use this option to build the sysroot as described [here](https://model-checking.github.io/kani/rfc/rfcs/0001-mir-linker.html#kanis-sysroot).

zhassan-aws (approved):

PR #2428, opened 2023-05-03
Limit FFI calls by default to explicitly supported ones

### LINKED ISSUE #1781: function call: parameter "pthread_key_create::destructor" type mismatch when hitting `thread::current()`

celinval:
### Description of changes: 

C-FFI is fairly unstable right now and the lack of error handling causes a poor experience when things go wrong. 

In this PR, we limit its usage to only support CBMC built-in functions that are explicitly included in Kani and alloc functions from `kani_lib.c`. User's should still be enable to go back to previous behavior by enabling unstable C-FFI support

### Resolved issues:

Resolves #1781 

### Call-outs:

This PR is currently blocked by https://github.com/model-checking/kani/pull/2427. Once that gets merged, the regression should pass.

### Testing:

### Checklist


zhassan-aws (commented):
Looks good. Mostly minor comments.

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
```diff
@@ -0,0 +1,155 @@
...
+use cbmc::goto_program::{Expr, Location, Stmt, Symbol, Type};
+use cbmc::{InternString, InternedString};
+use kani_queries::UserInput;
+use lazy_static::lazy_static;
+use rustc_middle::ty::Instance;
+use rustc_target::abi::call::Conv;
+use tracing::{debug, trace};
+
+lazy_static! {
+    static ref RUST_ALLOC_FNS: HashSet<InternedString> = {
```
zhassan-aws, line 24: Might want to add a comment that this should be kept in sync with `kani_lib.c`. And perhaps add a comment in `kani_lib.c` itself.

tests/kani/ForeignItems/extern_fn_ptr.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Z c-ffi --c-lib tests/kani/ForeignItems/lib.c
+
+//! Check that Kani correctly handle function pointers to C functions.
```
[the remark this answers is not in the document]
celinval: Not sure I understand your question, can you please clarify?

tests/ui/missing-function/extern_c/extern_c.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-// kani-flags: --harness harness
+// kani-flags: --harness harness -Z c-ffi
 
 // This test is to check Kani's error handling for missing functions.
 // TODO: Verify that this prints a compiler warning:
```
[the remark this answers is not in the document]
celinval: Let me check

tests/cargo-kani/libc/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+# Copyright Kani Contributors
```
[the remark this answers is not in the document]
celinval: Ah, good catch! Let me do a `git add`. :)

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
celinval: Actually, the `kani_lib.c` already has proper comments in place.

cprover_bindings/src/goto_program/builtin.rs
```diff
@@ -17,6 +17,8 @@ pub enum BuiltinFn {
     Copysignf,
     Cos,
     Cosf,
+    Error,
+    ErrorNoLocation,
```
feliperodri, line 21: There is no comment at the top of this enum documenting it. Is this the list of standard function supported by CPROVER? Can I add any builtin function from GCC that CBMC supports as well? Like anything from [here](https://github.com/diffblue/cbmc/tree/71d07e0d507b31220379f54eefb1bf3bb0f7a184/src/ansi-c/compiler_headers).

cprover_bindings/src/goto_program/builtin.rs
```diff
@@ -195,6 +204,8 @@ impl BuiltinFn {
             Copysignf => Type::float(),
             Cos => Type::double(),
             Cosf => Type::float(),
+            Error => Type::c_int().to_pointer(),
+            ErrorNoLocation => Type::c_int().to_pointer(),
```
feliperodri, line 208: Are these the return type?

tests/cargo-kani/libc/Cargo.toml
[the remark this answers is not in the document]
feliperodri: I'm a bit confuse if this [rule](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) applies in this case.

tests/ui/missing-function/extern_c/extern_c.rs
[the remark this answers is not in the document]
feliperodri: why `-Z`?

tests/expected/foreign-function/ffi_ptr.rs
```diff
@@ -0,0 +1,28 @@
...
+}
+
+fn may_not_call(call: bool, input: u32, func: unsafe extern "C" fn(u32) -> u32) -> Option<u32> {
+    call.then(|| unsafe { func(input) })
+}
+
+#[kani::proof]
+fn check_fn_ptr_called() {
+    let input: u32 = kani::any();
+    assert_eq!(call_on(input, foreign), input);
```
feliperodri, line 21: how can we always guarantee that the return o `call_on` is equal to `input`?

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -743,19 +743,15 @@ impl<'tcx> GotocCtx<'tcx> {
...
-            )
-            .with_is_extern(true)
-        });
+        let sym = if self.tcx.is_foreign_item(instance.def_id()) {
+            // Get the symbol that represents a foreign instance.
+            self.codegen_foreign_fn(instance)
+        } else {
+            // All non-foreign functions should've been declared beforehand.
+            let func = self.symbol_name(instance);
+            self.symbol_table.lookup(func).unwrap()
```
feliperodri, line 753: Shouldn't we add a debug statement here?

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
```diff
@@ -0,0 +1,155 @@
...
+                        .with_is_parameter(true);
+                    self.symbol_table.insert(sym);
+                    arg_type.as_parameter(Some(arg_name.into()), Some(base_name.into()))
+                })
+            })
+            .collect();
+        let ret_type = self.codegen_ty(fn_abi.ret.layout.ty);
+
+        if fn_abi.c_variadic {
+            Type::variadic_code(params, ret_type)
```
feliperodri, line 123: Do we have a test case for variadic code?

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
```diff
@@ -0,0 +1,155 @@
...
+            "https://github.com/model-checking/kani/issues/2423"
+        } else {
+            "https://github.com/model-checking/kani/issues/new/choose"
+        };
+        self.codegen_assert_assume(
+            Expr::bool_false(),
+            PropertyClass::UnsupportedConstruct,
+            &GotocCtx::unsupported_msg(&msg, Some(url)),
+            loc,
+        )
```
feliperodri, line 153: I assume this will be reported as an assertion failure with the same source location of the unsupported function, correct?

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    /// Generate code for a foreign function shim.
+    fn codegen_ffi_shim(&mut self, shim_name: InternedString, instance: Instance<'tcx>) -> Stmt {
+        debug!(?shim_name, ?instance, sym=?self.symbol_table.lookup(shim_name), "generate_foreign_shim");
+
+        let loc = self.codegen_span(&self.tcx.def_span(instance.def_id()));
+        let unsupported_check = self.codegen_ffi_unsupported(instance, loc);
+        Stmt::block(vec![unsupported_check], loc)
+    }
```
feliperodri, line 97: Do we have a test case for this?

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
```diff
@@ -0,0 +1,155 @@
...
+use rustc_target::abi::call::Conv;
+use tracing::{debug, trace};
+
+lazy_static! {
+    static ref RUST_ALLOC_FNS: HashSet<InternedString> = {
+        HashSet::from([
+            "__rust_alloc".into(),
+            "__rust_alloc_zeroed".into(),
+            "__rust_dealloc".into(),
+            "__rust_realloc".into(),
```
feliperodri, line 29: Where does this list come from?

tests/ui/missing-function/extern_c/extern_c.rs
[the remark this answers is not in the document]
celinval: Updated the comment.

cprover_bindings/src/goto_program/builtin.rs
celinval: Yes. You can also see that this model doesn't scale really well. The other problem is that the Rust `libc` signature doesn't always matches CBMC's (although they have the same ABI).

cprover_bindings/src/goto_program/builtin.rs
celinval: yep

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
celinval: I am adding a comment to describe this in the next version. Basically these are declared in the [Rust alloc module](https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/alloc/alloc/index.html), but they are defined today in [kani_lib.c](https://github.com/model-checking/kani/blob/main/library/kani/kani_lib.c).

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
celinval: yes. The new tests `expected/foreign-function/ffi_ptr.rs` and `cargo-kani/libc/` both exercise this code.

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
celinval: Yes, we have a fix-me test today. However, it wasn't really checking what it was meant to. I'll create a separate PR that fixes that test since this PR is already big enough.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
celinval: Actually, let me add an expected message so if it ever crashes we have more information.

tests/expected/foreign-function/ffi_ptr.rs
celinval: This dummy proof is incomplete since we don't have the body of `foreign`. This would only work if foreign is an identity function.

celinval: btw, I fixed the test here, but what I think we should do later is turn this test into an `expected` test and make sure we are emitting the right error.

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
celinval: btw, I fixed the test here, but what I think we should do later is turn this test into an expected test and make sure we are emitting the right error.

kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
```diff
@@ -0,0 +1,161 @@
...
+use kani_queries::UserInput;
+use lazy_static::lazy_static;
+use rustc_middle::ty::Instance;
+use rustc_target::abi::call::Conv;
+use tracing::{debug, trace};
+
+lazy_static! {
+    /// The list of Rust allocation functions that are declared in the `core::alloc` module
+    /// but defined by each backend.
+    /// For out `goto-program` backend, these functions are defined inside `kani_lib.c`.
```
zhassan-aws, line 26:
```suggestion
    /// For our `goto-program` backend, these functions are defined inside `kani_lib.c`.
```

zhassan-aws (approved):

Issue #2440, opened 2023-05-12
Create a command to play a counter example


celinval:
Requested feature: User should be able to run a unit test generated by Kani with one command which would incorporate stubs / kani library and other steps that today don't work or require extra configuration.
Use case: Running concrete playback without this extra setup step: https://model-checking.github.io/kani/debugging-verification-failures.html#setup
Link to relevant documentation (Rust reference, Nomicon, RFC):

Notes: We should accept a similar arguments as `cargo test` and accept a runner.

Issue #1831, opened 2022-11-01
`kani-driver` argument parsing needs restructuring


tedinski:
Assess further compounded an already existing issue that we'll need to address before it gets too much worse.

Here are a few known issues, in no particular order:

* [ ] Several kinds of options should be grouped together better: for example, `--workspace` `--package` `--target-dir` `--all-features`
* [ ] Several flags only apply to `cargo-kani` or `kani` but appear in both: for example, the above do nothing for `kani`
* [ ] `assess` has a hack to add both a subcommand and `--assess`; this is solely to make it run in the test suite and the flag should be removed.
* [ ] We should put together a specific set of supported options in `Cargo.toml` instead of trying to treat them as flags
* [ ] The arguments/configuration should be more decoupled from `KaniSession` in `kani-driver`. (I'm not sure this means complete separation: we don't want pointlessly duplicate boilerplate, but we might try to break the configuration down into multiple `structs` instead of having just `args`.)
* [ ] We badly need a "bless"-like UI test suite, to test all the help options to `kani` and `cargo-kani`.
* [ ] Assess currently takes "prepended flags" (like `cargo kani --flag assess`) and should not actually permit this

celinval:
Regarding the last topic:

>  [ ] Assess currently takes "prepended flags" (like cargo kani --flag assess) and should not actually permit this

we could use https://docs.rs/clap/latest/clap/struct.Command.html#method.args_conflicts_with_subcommands.

PR #2449, opened 2023-05-17
Move common arguments to their own structure

celinval:
### Description of changes: 

This is a pre-work needed for implementing #2440 as a subcommand. Today, the only other subcommand we have is `assess` and the way it takes arguments is rather broken as explained in #1831.

For the playback sub-command, I expect that we will need at least CargoArgs and CommonArgs to control the build process and to control the process verbosity.

We would probably benefit from properly isolating the build arguments, and not allowing arguments when subcommands are provided, but that will be a problem for another day since it will require changes to assess logic and Kani sesssion. I am hoping this will be enough for a first working version of the playback subcommand (famous last words... 🙂).

### Resolved issues:

Related to #2440 

### Call-outs:

I created an `args/` folder for all argument related logic because I believe it will be easier to maintain and it creates a better separation of responsibility and cleaner dependency tree (i.e.: args module no longer depends on assess module).

Update: I'll probably need to create one more PR that enables [args_conflicts_with_subcommands](https://docs.rs/clap/latest/clap/struct.Command.html#method.args_conflicts_with_subcommands) to fix the issue with the argument position.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


jaisnan:
Question: If I understand this right, we plan to have two subcommands (assess and concrete-playback) and one default one for verify after this PR, is that correct? That makes sense. Does this also mean that running concrete-playback, will block the verification flow and just print the unit test?

jaisnan (approved):
Looks like a re organization, so I'm approving with a few questions. Thanks!

kani-driver/src/args/mod.rs
```diff
@@ -44,20 +73,20 @@ pub struct CargoKaniArgs {
...
+    Assess(Box<crate::assess::AssessArgs>),
 }
 
-// Common arguments for invoking Kani. This gets put into KaniContext, whereas
-// anything above is "local" to "main"'s control flow.
+// Common arguments for invoking Kani for verification purpose. This gets put into KaniContext,
+// whereas anything above is "local" to "main"'s control flow.
 #[derive(Debug, clap::Args)]
-pub struct KaniArgs {
+pub struct VerifyArgs {
```
qinheping, line 89: We have `ValidateArgs` the trait for validate arguments. Here `VerifyArgs` sounds like we are verifying the arguments. How about Rename it to `VerificationArgs`

qinheping (approved):

celinval:
> I have just one comment, which is about the regression test that was changed. I didn't understand the motivation behind removing `--dry-run`. Is it because you moved it into the `ValidateArgs` and so `dry-run` doesn't work now as expected, but it will with the upcoming PR?

Great question! The test at hand had two errors:
1. Conflicting unwinding arguments (which is what we want to test).
2. Use of `--dry-run` which has been removed already.

Before this change, we checked # 1 first, then we checked # 2. With this change, the order has changed, and the test was failing now because of # 2. So I removed the `--dry-run` argument so it still tests # 1.

kani-driver/src/args/mod.rs
celinval: I really like that suggestion. Thanks!

celinval:
> Question 1: This change is a simple refactor to put all the args into two structures, one for common ones (Which will be used by the subcommands) and one structure for the verification related arguments, am i understanding that correct?

Yes! I was thinking that we could even add a subcommand `verify`, which is the default one when no subcommand is given.

> Question 2: The plan to have two subcommands (assess and concrete-playback) and one default one for verify after this PR, is that correct? That makes sense. Does this also mean that running concrete-playback, will block the verification flow and just print the unit test after the coming PR?

Oh, I guess you already answered the first question here. 😉 So the testcase generation won't change, and it will still be part of the "verify" subcommand. The new `playback` subcommand will compile and run this testcase.

> Question 3: I didn't understand the motivation behind removing --dry-run in the test case. Is it because you moved it into the ValidateArgs and so dry-run doesn't work now as expected, but it will with the upcoming PR?

Per my previous answer, `--dry-run` doesn't work today. We should really just remove it completely, but I didn't want to make that change as part of this PR.

tests/expected/dry-run-flag-unwind-conflict/main.rs
```diff
@@ -1,10 +1,7 @@
 // Copyright Kani Contributors
```
zhassan-aws, line 1: We should probably renamed the test since it has `dry-run` in its name.

zhassan-aws (approved):

kani-driver/src/args/mod.rs
```diff
@@ -595,9 +576,11 @@ impl KaniArgs {
...
         }
 
-        if !self.c_lib.is_empty() && !self.unstable_features.contains(&UnstableFeatures::CFfi) {
-            if self.enable_unstable {
-                self.print_deprecated("`--enable-unstable`");
+        if !self.c_lib.is_empty()
+            && !self.common_args.unstable_features.contains(&UnstableFeatures::CFfi)
+        {
+            if self.common_args.enable_unstable {
+                print_deprecated(&self.common_args, "`--enable-unstable`");
```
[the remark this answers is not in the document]
celinval: Fair... let me update this. I had actually crated a version where I distinguish, "retired" and "deprecated". Usually deprecated means that something will stop working, not that it has already stopped working.

I just wasn't sure about the retired term. Any thoughts?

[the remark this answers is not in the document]
zhassan-aws: Instead of "retired", we can use "obsolete". Deprecated may still have an effect, but will be removed in the future. Obsolete has no effect.

PR #2470, opened 2023-05-24
Slightly change playback arguments

celinval:
### Description of changes: 

Remove `--test` and add trailing test arguments to allow us to pass extra arguments to the test binary.

### Resolved issues:

N/A

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:
As I was working in the integration with the VSCode extension, I noticed that it passes arguments to the test binary. Instead of manually adding support to each one of them, just handle them as `cargo test` does and pass them directly.

### Testing:

### Checklist


jaisnan (approved):
Thanks!

PR #2474, opened 2023-05-25
Do not download x86 mdbook on aarch64

karkhaz:
Users who attempt to build the documentation on Apple ARM will be prompted to install mdbook using cargo if it is not already installed, as pre-built binaries for this platform are not available.

### Testing:

* How is this change tested? Ran it a few times on an M1 laptop, inverting the conditions

* Is this a refactor change?

### Checklist

### DIFF
--- a/scripts/build-docs.sh
+++ b/scripts/build-docs.sh
@@ -11,15 +11,25 @@ set -eu
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 cd $SCRIPT_DIR
 
-# Download mdbook release (vs spending time building it via cargo install)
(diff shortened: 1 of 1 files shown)


zhassan-aws:
The latest mdbook releases ship with aarch64 binaries, e.g. https://github.com/rust-lang/mdBook/releases/tag/v0.4.29

Should we instead upgrade to a more recent version?

karkhaz:
If you're referring to mdbook-v0.4.29-aarch64-unknown-linux-musl.tar.gz, this doesn't seem to work on my M1
```
kani/scripts/mdbook: cannot execute binary file
```
and the binary is an ELF file.

zhassan-aws:
I see

PR #2476, opened 2023-05-25
NOP concrete playback to type check during verification mode.

### LINKED ISSUE #2475: Kani cannot verify bolero harness after concrete test has been added
I tried this code:

```rust

#[cfg(test)]
mod bolero_harness {
    #[test]
    fn dummy_harness() {
        bolero::check!()
            .with_type()
            .cloned()
            .for_each(|(a, b): (u8, u8)| assert!(a < b || a > b));
    }
}
```

using the following command line sequence:

```
cargo kani --tests -Z concrete-playback --concrete-playback=inplace
cargo kani --tests
```

with Kani version:

I expected to see this happen: Both commands should succeed

Instead, this happened: The second command failed due to a compilation error
```
error[E0425]: cannot find function `concrete_playback_run` in crate `kani`
  --> src/lib.rs:
   |
   |         kani::concrete_playback_run(concrete_vals, fuzz_add);
   |               ^^^^^^^^^^^^^^^^^^^^^ not found in `kani`

```

The problem is that the harness uses a function that is only included when we execute concrete playback. We should still implement that function in verification mode.

YoshikiTakashima:
### Description of changes: 

This PR adds a NOP concrete playback function so that code inserted by the concrete playback feature that uses `concrete_playback` function still type checks during verification. Previously, this function was not available during verification, causing type fails during verification after playback code was injected.

### Resolved issues:

Resolves #2475 

### Call-outs:

@YoshikiTakashima is not sure where the tests should go given that we want to test behavior when the kani is not in playback mode.

### Testing:

* How is this change tested? `tests/cargo-kani/simple-lib`

### Checklist

### DIFF
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -21,6 +21,10 @@ pub mod vec;
 pub use arbitrary::Arbitrary;
 #[cfg(feature = "concrete_playback")]
 pub use concrete_playback::concrete_playback_run;
+#[cfg(not(feature = "concrete_playback"))]
+/// NOP `concrete_playback` for type checking during verification mode.
+pub fn concrete_playback_run<F: Fn()>(_: Vec<Vec<u8>>, _: F) {}
+
 pub use futures::block_on;
 
 /// Creates an assumption that will be valid after this statement run. Note that the assumption
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -23,7 +23,9 @@ pub use arbitrary::Arbitrary;
 pub use concrete_playback::concrete_playback_run;
 #[cfg(not(feature = "concrete_playback"))]
 /// NOP `concrete_playback` for type checking during verification mode.
-pub fn concrete_playback_run<F: Fn()>(_: Vec<Vec<u8>>, _: F) {}
+pub fn concrete_playback_run<F: Fn()>(_: Vec<Vec<u8>>, _: F) {
+    unreachable!("Concrete playback does not work during verification")
+}
 
 pub use futures::block_on;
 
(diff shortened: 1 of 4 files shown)


library/kani/src/lib.rs
```diff
@@ -21,6 +21,10 @@ pub mod vec;
 pub use arbitrary::Arbitrary;
 #[cfg(feature = "concrete_playback")]
 pub use concrete_playback::concrete_playback_run;
+#[cfg(not(feature = "concrete_playback"))]
+/// NOP `concrete_playback` for type checking during verification mode.
+pub fn concrete_playback_run<F: Fn()>(_: Vec<Vec<u8>>, _: F) {}
```
celinval, line 26: Can you please add an `unreachable!("Concrete playback does not work during verification")` or something like that, just in case users ever reach this code.

celinval (approved):
Thanks @YoshikiTakashima. Do you mind adding a separate test instead?

library/kani/src/lib.rs
YoshikiTakashima: Done [c057854](https://github.com/model-checking/kani/pull/2476/commits/c057854d26cdf961d3ed778cdf9e1582e0318dd8)

YoshikiTakashima:
new tests in 4575586d40bbb7b718ac41e5cd5e55744b247cf4.

YoshikiTakashima:
Something about this PR seems to be degrading performance.

celinval:
> Something about this PR seems to be degrading performance. Might need a deep dive into benchcomp.

No worries, the test that failed is very flaky. You should be safe.

YoshikiTakashima:
Rerunning didn't seem to fix it. The performance degradation might be something a bit more permanent...

PR #2477, opened 2023-05-25
Fix playback with build scripts

celinval:
### Description of changes: 

I'm not 100% sure on why we need to set target, but without it, cargo is not properly running the build scripts as part of `playback` subcommand. I added that + made sure we propagate RUSTFLAGS to playback as well.

### Resolved issues:

N/A

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

Adding `--target` also caused compilation for our tests to generate an extra artifact. Thus, I had to change an existing test.

### Testing:

* Is this a refactor change?

### Checklist

### DIFF
--- a/kani-driver/src/call_single_file.rs
+++ b/kani-driver/src/call_single_file.rs
@@ -136,12 +136,6 @@ impl KaniSession {
             }
         }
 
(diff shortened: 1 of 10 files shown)


jaisnan (approved):
Lgtm!

PR #2478, opened 2023-05-25
Improve assess and regression time

celinval:
### Description of changes: 

The per-reachability analysis can cause a big penalty when assessing a project that has lots of tests.

Change that for now.

Despite that improvement, assessing the firecracker still takes a long time, even with --only-codegen, due to the number of calls done to goto-cc. Thus, don't use assess here.

### Resolved issues:

N/A

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

Due to https://github.com/model-checking/kani/pull/2472, we need a specialized name for each test harness. For now, we just replicate the model for each harness.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist

### DIFF
--- a/scripts/codegen-firecracker.sh
+++ b/scripts/codegen-firecracker.sh
@@ -6,7 +6,13 @@ set -eu
 
 # Test for platform
 PLATFORM=$(uname -sp)
-if [[ $PLATFORM != "Linux x86_64" ]]; then
+
+if [[ $PLATFORM == "Linux x86_64" ]]
+then
+  TARGET="x86_64-unknown-linux-gnu"
+  # 'env' necessary to avoid bash built-in 'time'
+  WRAPPER="env time -v"
+else
   echo
   echo "Firecracker codegen regression only works on Linux x86 platform, skipping..."
   echo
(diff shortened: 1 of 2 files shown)


scripts/codegen-firecracker.sh
```diff
@@ -23,11 +29,27 @@ echo
...
+    "-Cpanic=abort"
+    "-Zalways-encode-mir"
+    "-Cllvm-args=--goto-c"
+    "-Cllvm-args=--ignore-global-asm"
+    "-Cllvm-args=--reachability=pub_fns"
+    "--sysroot=${KANI_DIR}/target/kani"
+)
+export RUSTFLAGS="${RUST_FLAGS[@]}"
+export RUSTC="$KANI_DIR/target/kani/bin/kani-compiler"
+$WRAPPER cargo build --verbose --lib --target $TARGET
```
zhassan-aws, line 52: Do we want to keep `--verbose` in CI?

zhassan-aws (approved):

scripts/codegen-firecracker.sh
celinval: I think so. It's not that verbose and it helps debugging issues.

PR #2481, opened 2023-05-26
Change how we translate Array to GOTO

### LINKED ISSUE #2479: Kani compiles arrays into a struct wrapping arrays in GOTO
**Proposed change:** We should evaluate if that is needed. We have the following comment in our code:
https://github.com/model-checking/kani/blob/c6e91691d5ad2dd4bfb364ace2b6df92a2ba88c7/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs#L783-L787

which points to the reason why it was done it in the first place, but it's not clear if we measured the performance implications.

**Motivation:** CBMC doesn’t enable array theory for arrays wrapped in struct.

celinval:
### Description of changes: 

Directly codegen a rust Array into a GOTO array. Before we used to wrap it in a struct.

According to the code documentation, we used to wrap it so we could assign an array to another using the struct copy; however, I didn't see any place where this is needed.

The cprover bindings crate incorrectly returned that an array was not an lvalue type, which maybe is the reason why we thought that was needed before. However, array is an lvalue, just not a modifiable lvalue, according to the C standard.
This simplifies the code and hopefully will help with loop systhesis.

### Resolved issues:

Resolves #2479 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

We might want to add more tests before merging this one. Keeping it as draft for now.

### Testing:

### Checklist


qinheping:
I experimented this PR with the loop contracts for the hidden loop in `memcmp` (implemented in https://github.com/qinheping/cbmc/tree/invariant-for-memcmp). 

The task is to prove the following program that compares two nondeterministic arrays with loop contracts.
```rust
#[kani::proof]
fn main() {
    let a: [i8; 4096] = kani::any();
    let b: [i8; 4096] = kani::any();
    // assert!(!unsafe { std::intrinsics::raw_eq(&a, &b) });
    assert_eq!(a, b);
}
```
The command I used (with the synthesizer built from https://github.com/qinheping/cbmc/tree/invariant-for-memcmp):
```
> kani --enable-unstable --synthesize-loop-contracts main.rs --keep-temps --cbmc-args --arrays-uf-always
```
Kani return in 2 seconds
```
SUMMARY:
 ** 1 of 38 failed
Failed Checks: assertion failed: a == b
 File: "/Users/qinhh/Repos/kani/qinheping/tests/ui/loop-contracts-synthesis/array_eq/main.rs", line 68, in main

VERIFICATION:- FAILED
Verification Time: 1.8414005s
```

If we increase the size of arrays from `4096` to `65536`, the running time goes up to 14 seconds.

So I think it is good to merge this PR if it doesn't slow down any other benchmarks.

There are still issues to use loop contracts for array-manipulating programs even with this PR:
1. Array theory won't be enable if the type of array is not `i8` or `u8`,
2. Although array theory allow us to prove programs with large-sized (~2^16) arrays, the running time still grows as the growth of the size of the arrays.

qinheping (approved):

Issue #2488, opened 2023-05-30
Concrete Playback command generates incorrect output 


jaisnan:
Running concrete playback using `cargo kani playback --package test-concrete -Z concrete-playback` on a unit test produces the following output - 

```
failures:

---- kani_concrete_playback_harness_i64_13477596495325479973 stdout ----
---- kani_concrete_playback_harness_i64_13477596495325479973 stderr ----
thread 'main' panicked at 'assertion failed: !(i64_1 == i64::MIN && i64_2 == -101 && i64_3 == 0 && i64_4 == 101 &&
            i64_5 == i64::MAX)', library/kani/src/lib.rs:90:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- other::harnesses::kani_concrete_playback_my_harness_5205172929174653629 stdout ----
---- other::harnesses::kani_concrete_playback_my_harness_5205172929174653629 stderr ----
thread 'main' panicked at 'assertion failed: !(result_1 == Ok(101) && result_2 == Err(102))', library/kani/src/lib.rs:90:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- kani_concrete_playback_harness_i8_10415494690275622521 stdout ----
---- kani_concrete_playback_harness_i8_10415494690275622521 stderr ----
thread 'main' panicked at 'assertion failed: !(i8_1 == i8::MIN && i8_2 == -101 && i8_3 == 0 && i8_4 == 101 &&
            i8_5 == i8::MAX)', library/kani/src/lib.rs:90:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    kani_concrete_playback_harness_i64_13477596495325479973
    kani_concrete_playback_harness_i8_10415494690275622521
    other::harnesses::kani_concrete_playback_my_harness_5205172929174653629
```

There are two issues with this output - 

1. All the traces contain stdout and stderr in the output, even though only the stdout usually is displayed to the user. 
2. All the trace information refers to kani's source code as the origin of panic, not the unit test itself, which is erronous.

I tried this code:

```rust
<code>
```

using the following command line invocation:

```
cargo kani playback --package test-concrete -Z concrete-playback
```

with Kani version: 0.28

I expected to see this happen: 

1. The output should contain trace information about the package, not kani. 
2. Only stdout should be displayed to the user.

This is what the output of cargo test usually looks like - 

```
running 3 tests
test test_example_1 ... ok
[...]

PR #2490, opened 2023-05-30
Do not use panic abort strategy for playback

celinval:
### Description of changes: 

The VSCode debugger isn't working when we build the std library without unwind support. Build with unwind instead.

### Resolved issues:

#2488 

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


jaisnan (approved):
Tested the debugger with these changes and it works locally for the extension as well.

Issue #2494, opened 2023-05-30
Reduce the compilation overhead for harnesses with the same set of stubs.


celinval:
Requested feature: With the fix to #1841 the compiler will be able to handle multiple harnesses with stubs, but it will rerun the Rust compiler multiple times independent on the set of stubs used on each harness.
Use case: Users may use the same stubs for a set of harnesses, and they can all be processed in the same compiler session. This will reduce the duplicated warnings emitted by the compiler, as well as the time compiling a crate with multiple harnesses that use the same stubs.
Link to relevant documentation (Rust reference, Nomicon, RFC):

Issue #2493, opened 2023-05-30
Finish refactoring on how we handle the metadata file for Tests and PubFns


celinval:
**Proposed change:** To solve #1841 we moved the metadata generation to the `kani_compiler` for the Harness reachability mode. We should also change how we handle Tests and PubFns.

**Motivation:** The code has two very distinct logic to handle metadata, which is more fragile and harder to maintain. We only partitioned this work into separate issues to reduce the size of the PR.

Note: We also need to ensure we store unsupported features to the metadata files.

celinval:
This is no longer needed since #3245.

carolynzech:
Closing since this is no longer needed.

PR #2486, opened 2023-05-29
Simplify compiler queries structure (internal)

celinval:
### Description of changes: 

1. Move the query database out of its own package. 
2. Remove UserInput trait for now.
3. Remove unnecessary functions.

### Resolved issues:

N/A

### Call-outs:

The query database was initially introduced to eventually help us to communicate between all Kani components, but it became just a compiler internal communication mechanism. Thus, move it to be inside compiler.

### Testing:

* How is this change tested? No user behavior changes is expected.

* Is this a refactor change? Yes

### Checklist
- [ ] Each commit message has a non-empty body, explaining why the change was made
- [ ] Methods or procedures are documented
- [ ] Regression or unit tests are included, or existing tests cover the modified code
- [ ] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #2495, opened 2023-05-30
Remove --harness requirement from stubbing

### LINKED ISSUE #1841: Remove `--harness` argument from `kani-compiler`
The `--harness` argument has been temporarily added to `kani-compiler` to ease initial support for stubbing. We should remove it once the stubbing mechanism can handle multiple harnesses at once.

celinval:
### Description of changes: 

The compiler can now generate code for different combinations of harnesses with and without stubs. For that, we implemented a little state machine that may execute the underlying rust compiler multiple times. The first run is always for harnesses that don't have stub, while the subsequent runs are for harnesses with stubs.

### Resolved issues:

Resolves #1841

### Related RFC:

#1809 

### Call-outs:

1. I have not changed how the compiler handles Tests and PublicFns reachability yet to minimize the changes in this PR. I created https://github.com/model-checking/kani/issues/2493 to capture that work.
2. We should also optimize the number of compiler sessions we execute when multiple harnesses use the same set of stubs. For simplicity, we only handle one harness at a time in this PR: https://github.com/model-checking/kani/issues/2494
~~3. Keeping this as a draft since we are blocked waiting for #2486. Once this change goes through, I'll drop the first commit.~~

### Testing:

### Checklist


tests/cargo-kani/stubbing-ws-packages/top/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+[package]
+name = "top"
```
zhassan-aws, line 5: The name is very hardware-ish :)

kani-compiler/src/kani_compiler.rs
```diff
@@ -64,68 +63,262 @@ fn backend(queries: Arc<Mutex<QueryDb>>) -> Box<CodegenBackend> {
...
+}
+
+/// Represents the current compilation stage.
+///
+/// The Kani compiler may run the Rust compiler multiple times since stubbing has to be applied
+/// to the entire Rust compiler session.
+///
+/// - We always start in the [CompilationStage::Init].
+/// - After [CompilationStage::Init] we transition to either
+///   - [CompilationStage::CodegenNoStubs] on a regular crate compilatin, this will follow Init.
```
zhassan-aws, line 85:
```suggestion
///   - [CompilationStage::CodegenNoStubs] on a regular crate compilation, this will follow Init.
```

kani-compiler/src/kani_compiler.rs
```diff
@@ -64,68 +63,262 @@ fn backend(queries: Arc<Mutex<QueryDb>>) -> Box<CodegenBackend> {
...
+    fn next_stage(&mut self) {
+        self.stage = match &mut self.stage {
+            CompilationStage::Init => {
+                // This may occur when user passes arguments like --version or --help.
+                CompilationStage::Done
+            }
+            CompilationStage::CodegenNoStubs { next_harnesses, all_harnesses, .. }
+            | CompilationStage::CodegenWithStubs { next_harnesses, all_harnesses, .. } => {
+                if let Some(target_harness) = next_harnesses.pop() {
+                    CompilationStage::CodegenWithStubs {
```
zhassan-aws, line 196: Does it move to the `CodegenWithStubs` stage even if there are no harnesses with stubs?

zhassan-aws (commented):
Neat! Thanks @celinval! I have a few comments.

tests/cargo-kani/stubbing-ws-packages/expected
```diff
@@ -0,0 +1,25 @@
+Checking harness verify::check_no_stub...
+Status: SUCCESS\
+Description: "assertion failed: config == StubConfig::NoStub"
+VERIFICATION:- SUCCESSFUL
+
+Checking harness verify::check_stub_1...
+Status: SUCCESS\
+Description: "assertion failed: config == StubConfig::Stub1"
+VERIFICATION:- SUCCESSFUL
```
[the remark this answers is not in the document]
celinval: Yeah I know... this is more for human consumption. At least it helps me when I'm debugging a test to know what the expected result is. What do you think?

tests/cargo-kani/stubbing-ws-packages/top/Cargo.toml
celinval: LoL... I did not even think about it. Any other suggestion?

kani-compiler/src/kani_compiler.rs
celinval: If there is no harness with stubs, the `next_harnesses` vector will be empty, so we will move to Done in the `else` block.

All harnesses without stubs are processed together in one single codegen pass as part of the `CodegenNoStubs` stage.

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -234,13 +235,17 @@ impl CodegenBackend for GotocCodegenBackend {
         match reachability {
             ReachabilityType::Harnesses => {
                 // Cross-crate collecting of all items that are reachable from the crate harnesses.
-                let harnesses = filter_crate_items(tcx, |_, def_id| is_proof_harness(tcx, def_id));
+                let harnesses = queries.target_harnesses();
+                let mut items: HashSet<DefPathHash> = HashSet::with_capacity(harnesses.len());
```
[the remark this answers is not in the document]
celinval: Yes, we could. This is my brain's fault that couldn't ignore the fact that we already know the size of the final set, and preferred avoiding unnecessary reallocations. But I can change it if you prefer. I get that there will be no real performance penalty here.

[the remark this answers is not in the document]
zhassan-aws: I would've thought that `FromIterator` will use the size for the initial capacity, but seems the `HashSet` implementation doesn't:
https://doc.rust-lang.org/src/std/collections/hash/set.rs.html#1023
(unlike the `Vec` one).

In that case, what you have makes sense. Performance might not matter if we have a handful of harnesses, but the number could be large for `assess` and the like.

tests/cargo-kani/stubbing-ws-packages/expected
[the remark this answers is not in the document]
zhassan-aws: I see. We can keep them.

tests/cargo-kani/stubbing-ws-packages/top/Cargo.toml
zhassan-aws: No, just pointing it out :)

zhassan-aws (approved):
Good to go

kani-compiler/src/kani_compiler.rs
```diff
@@ -64,68 +63,262 @@ fn backend(queries: Arc<Mutex<QueryDb>>) -> Box<CodegenBackend> {
...
+    /// actually invoke the rust compiler multiple times.
+    pub fn run(&mut self, orig_args: Vec<String>) -> Result<(), ErrorGuaranteed> {
+        while !self.stage.is_done() {
+            debug!(next=?self.stage, "run");
+            match &self.stage {
+                CompilationStage::Init => {
+                    self.run_compilation_session(&orig_args)?;
+                }
+                CompilationStage::CodegenNoStubs { .. } => {
+                    unreachable!("This stage should always run in the same session an Init");
```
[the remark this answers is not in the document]
celinval:
```suggestion
                    unreachable!("This stage should always run in the same session as Init");
```

PR #2496, opened 2023-05-31
Generate Multiple playback harnesses when multiple crashes exist in a single harness.

### LINKED ISSUE #2461: Concrete playback should be able to generate multiple tests at once
Requested feature: When handling multiple failures / cover checks, concrete playback should be able to generate one test per property.
Use case: Users will likely need to debug multiple failures or they may want to extract tests for multiple covers today. Today, there is no way to control which property Kani generate the test for.

Test case:
```rust
//! Check that Kani correctly adds tests to the cover checks reachable in a harness.
extern crate kani;

#[cfg(kani)]
mod verify {
    use std::num::NonZeroU8;
    use std::convert::TryFrom;
    use kani::cover;

    #[kani::proof]
    fn try_nz_u8() {
        let val: u8 = kani::any();
        let result = NonZeroU8::try_from(val);
        match result {
            Ok(nz_val) => {
                cover!(true, "Ok");      // Cover 1
                assert_eq!(nz_val.get(), val);
            },
            Err(_) => {
                cover!(true, "Not ok");  // Cover 2
                assert_eq!(val, 0);
            }
        }
    }
}
```
I expect Kani to generate two different test cases for this harness. But it instead generates one test, and the following warning:

```
WARNING: Unable to extract concrete values from multiple failing assertions. Skipping property `verify::try_nz_u8.cover.2` with descr
iption `Ok`
```

Note: We should also update the warning, since concrete playback now supports different property type.

YoshikiTakashima:
### Description of changes: 

Allow multiple harnesses to be generated.

### Resolved issues:

Resolves #2461 

### Call-outs:

> Note: We should also update the warning, since concrete playback now supports different property type.

Not sure what is requested here. 

### Testing:

* How is this change tested? `tests/ui/concrete-playback/single-harness-multi-trace`

### Checklist


YoshikiTakashima:
I'll mark this as draft since it is not immidiately clear what ought to be done. Thus fixing the test fails at this point does not make sense.

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -289,35 +292,25 @@ mod concrete_vals_extractor {
...
-        }
+    pub fn extract_harness_values(result_items: &[Property]) -> Vec<Vec<ConcreteVal>> {
+        result_items
+            .iter()
+            .filter(|prop| {
+                (prop.property_class() == "assertion" && prop.status == CheckStatus::Failure)
+                    || (prop.property_class() == "cover" && prop.status == CheckStatus::Satisfied)
+            })
+            .map(|property| {
+                // Extract values for the first assertion that has failed.
```
celinval, line 303: Please update the comment

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -32,13 +32,16 @@ impl KaniSession {
...
 
         if let Ok(result_items) = &verification_result.results {
-            match extract_harness_values(result_items) {
-                None => println!(
+            let harness_values: Vec<Vec<ConcreteVal>> = extract_harness_values(result_items);
+
+            if harness_values.is_empty() {
+                println!(
                     "WARNING: Kani could not produce a concrete playback for `{}` because there \
                     were no failing panic checks.",
```
celinval, line 40: or satisfiable cover statements.

tests/ui/concrete-playback/single-harness-multi-trace/expected
```diff
@@ -0,0 +1,25 @@
...
+}
+```
+
+Concrete playback unit test for `verify::try_nz_u8`:
+```
+#[test]
+fn kani_concrete_playback_try_nz_u8_
+    let concrete_vals: Vec<Vec<u8>> = vec![
+        // 16
+        vec![16],
```
celinval, line 21: This value is not stable. Can you please change the cover condition to ensure that the value is 16 to make this expectation more robust?

celinval (changes requested):
Thanks for doing this. I have 2 requests:
1. Can you please add a test for `inplace` where Kani will add more than one test for more than one harness at the same time? I'm a bit worried about updating the file multiple times in a row since we use original line numbers to calculate the test position.
2. The current implementation is generating a modified version of the file N times. Can we add all the tests in one pass instead?

tests/ui/concrete-playback/single-harness-multi-trace/expected
YoshikiTakashima: 0cae000d356345a82903ef0418a3a96e34503f7d

YoshikiTakashima:
@celinval I think I resolved your comments with the exception of the `inplace` testing, which I think is matter of placing the right files in the right directories. I will flip the PR over to "ready for review"

jaisnan:
I'm a bit worried about a potential proof having a lot of crashes as @YoshikiTakashima has mentioned, and the user clicking the concrete playback button and discovering that they have a lot of test cases pasted into their source. It might be annoying to deal with. Does it make sense to be worried about this corner case? Do we have any plans to deal with that when that happens?

YoshikiTakashima:
@jaisnan I think it's plausible if the harness is large. One possibility is that we inject first N harnesses where N is a user-set parameter in the extension. Not sure if there is a good way to prioritize them. 

As for dealing with it when it happens, I think most IDEs have a history beyond last save, so ctrl-z should still work. Not too familiar except for Emacs and VS Code though.

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -32,48 +32,58 @@ impl KaniSession {
...
+                    .collect();
+                match playback_mode {
+                    ConcretePlaybackMode::Print => {
+                        for generated_unit_test in unit_tests.iter() {
                             println!(
                                 "Concrete playback unit test for `{}`:\n```\n{}\n```",
                                 &harness.pretty_name,
                                 &generated_unit_test.code.join("\n")
                             );
                             println!(
```
celinval, line 59: Please move this INFO out of the loop (if at least one harness was printed).

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -32,48 +32,58 @@ impl KaniSession {
...
+                                 src code, run Kani with `--concrete-playback=inplace`.",
                                 &generated_unit_test.name
                             );
                         }
-                        ConcretePlaybackMode::InPlace => {
-                            if !self.args.common_args.quiet {
+                    }
+                    ConcretePlaybackMode::InPlace => {
+                        if !self.args.common_args.quiet {
+                            for generated_unit_test in unit_tests.iter() {
```
celinval, line 68: Can you print one info with a list of all tests that will be added instead and the name of the file to be modified?

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -84,71 +94,83 @@ impl KaniSession {
...
-        unit_test: &UnitTest,
+        unit_tests: Vec<UnitTest>,
     ) -> Result<()> {
-        let unit_test_already_in_src =
-            self.add_test_inplace(src_path, proof_harness_end_line, unit_test)?;
+        // compute range to run rustfmt on.
+        let concrete_playback_num_lines: usize =
+            unit_tests.iter().map(|unit_test| unit_test.code.len()).sum();
+        let unit_test_start_line = proof_harness_end_line + 1;
+        let unit_test_end_line = unit_test_start_line + concrete_playback_num_lines - 1;
```
celinval, line 103: you should probably take `is_new_injection` into consideration here.

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -84,71 +94,83 @@ impl KaniSession {
...
+                true
             }
+        });
+
+        // Create temp file
+        let mut temp_file = TempFile::try_new("concrete_playback.tmp")?;
+        let mut curr_line_num = 0;
+
+        // Use a buffered reader/writer to generate the unit test line by line
+        let mut is_new_injection = false;
```
celinval, line 155: Is this needed? I'm assuming this will be true if `!unit_tests.is_empty()`.

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -84,71 +94,83 @@ impl KaniSession {
...
                     );
                 }
-                // the drop impl will take care of flushing and resetting
-                return Ok(true);
+
+                false
+            } else {
+                true
             }
+        });
```
celinval, line 148: Can we check if there is any test left before we try to modify the file?

tests/script-based-pre/playback_already_existing/original.rs
```diff
@@ -0,0 +1,35 @@
...
+                assert_eq!(nz_val.get(), val);
+            }
+            Err(_) => {
+                cover!(true, "Not ok");
+                assert_eq!(val, 0);
+            }
+        }
+    }
+    #[test]
+    fn kani_concrete_playback_try_nz_u8_17663051139329126359() {
```
celinval, line 28: I would actually suggest that you run concrete playback twice in a row, instead of relying on the name staying the same.

tests/script-based-pre/playback_already_existing/playback_opts.sh
```diff
@@ -0,0 +1,24 @@
...
+set -o nounset
+
+RS_FILE="modified.rs"
+cp original.rs ${RS_FILE}
+
+echo "[TEST] Generate test..."
+kani ${RS_FILE} -Z concrete-playback --concrete-playback=inplace
+
+echo "[TEST] Only codegen test..."
+kani playback -Z concrete-playback --only-codegen ${RS_FILE} -- kani_concrete_playback
```
celinval, line 15: You probably don't need this for this test

tests/script-based-pre/playback_already_existing/playback_opts.sh
```diff
@@ -0,0 +1,24 @@
...
+
+echo "[TEST] Run test..."
+kani playback -Z concrete-playback ${RS_FILE} -- kani_concrete_playback
+
+echo "[TEST] Json format..."
+kani playback -Z concrete-playback ${RS_FILE} --only-codegen --message-format=json -- kani_concrete_playback
```
celinval, line 21: neither this one

tests/script-based-pre/playback_multi_harness_multi_inject/playback_opts.sh
```diff
@@ -0,0 +1,24 @@
...
+set -o nounset
+
+RS_FILE="modified.rs"
+cp original.rs ${RS_FILE}
+
+echo "[TEST] Generate test..."
+kani ${RS_FILE} -Z concrete-playback --concrete-playback=inplace
+
+echo "[TEST] Only codegen test..."
+kani playback -Z concrete-playback --only-codegen ${RS_FILE} -- kani_concrete_playback
```
celinval, line 15: remove this one

tests/script-based-pre/playback_multi_harness_multi_inject/playback_opts.sh
```diff
@@ -0,0 +1,24 @@
...
+
+echo "[TEST] Run test..."
+kani playback -Z concrete-playback ${RS_FILE} -- kani_concrete_playback
+
+echo "[TEST] Json format..."
+kani playback -Z concrete-playback ${RS_FILE} --only-codegen --message-format=json -- kani_concrete_playback
```
celinval, line 21: remove this one

tests/script-based-pre/playback_multi_harness_multi_inject/playback_opts.sh
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Test that concrete playback -Z concrete-playback executes as expected
+set -o pipefail
+set -o nounset
+
+RS_FILE="modified.rs"
+cp original.rs ${RS_FILE}
+
+echo "[TEST] Generate test..."
+kani ${RS_FILE} -Z concrete-playback --concrete-playback=inplace
```
celinval, line 12: maybe run this twice in a row to make sure the second time doesn't try to re-add the tests.

celinval (commented):
Awesome! Thanks @YoshikiTakashima

adpaco-aws:
> We could add a configurable cap, and even set its default to 1 to keep the same behavior as today.

This is a good starting point: we can remove the technical limitation first and decide on what properties to target later.

> Regarding priorities, we could have different modes, like user assertions, cover statements and UB checks (the last doesn't work yet with concrete playback). I was wondering if by default we should only generate tests for assertion failures.

Still, tests generated for cover statement could be useful for proof debugging. I don't think test cases for UB checks would be that useful though.

tests/script-based-pre/playback_already_existing/original.rs
YoshikiTakashima: Nice idea. Will implement in the script.

celinval:
> > We could add a configurable cap, and even set its default to 1 to keep the same behavior as today.
> 
> This is a good starting point: we can remove the technical limitation first and decide on what properties to target later.
> 
> > Regarding priorities, we could have different modes, like user assertions, cover statements and UB checks (the last doesn't work yet with concrete playback). I was wondering if by default we should only generate tests for assertion failures.
> 
> Still, tests generated for cover statement could be useful for proof debugging. I don't think test cases for UB checks would be that useful though.

I think UB tests would be useful. For example, memory checks could be debugged using valgrind, and other UB checks could be debugged using MIRI.

adpaco-aws:
> I think UB tests would be useful. For example, memory checks could be debugged using valgrind, and other UB checks could be debugged using MIRI.

Right, I hadn't thought of using them with other tools.

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -84,36 +94,37 @@ impl KaniSession {
...
-            line_range: Some((unit_test_start_line, unit_test_end_line)),
-        }];
-        self.run_rustfmt(&file_line_ranges, Some(&path))?;
         Ok(())
     }
 
     /// Writes the new source code to a user's source file using a tempfile as the means.
-    /// Returns whether the unit test was already in the old source code.
+    /// Returns whether new unit test was injected.
     fn add_test_inplace(
```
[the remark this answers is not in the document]
YoshikiTakashima: 930efaf7f4edfaa107774da0e770ce8fe252b931

kani-driver/src/concrete_playback/test_generator.rs
YoshikiTakashima: 930efaf7f4edfaa107774da0e770ce8fe252b931

kani-driver/src/concrete_playback/test_generator.rs
YoshikiTakashima: 930efaf7f4edfaa107774da0e770ce8fe252b931

kani-driver/src/concrete_playback/test_generator.rs
YoshikiTakashima: 930efaf7f4edfaa107774da0e770ce8fe252b931

tests/script-based-pre/playback_multi_harness_multi_inject/playback_opts.sh
YoshikiTakashima: 5979461e053a02b480dd3d5a81cc6af5ad400611

tests/script-based-pre/playback_multi_harness_multi_inject/playback_opts.sh
YoshikiTakashima: 5979461e053a02b480dd3d5a81cc6af5ad400611

tests/script-based-pre/playback_already_existing/playback_opts.sh
YoshikiTakashima: 5979461e053a02b480dd3d5a81cc6af5ad400611

tests/script-based-pre/playback_already_existing/playback_opts.sh
YoshikiTakashima: 5979461e053a02b480dd3d5a81cc6af5ad400611

tests/script-based-pre/playback_multi_harness_multi_inject/playback_opts.sh
YoshikiTakashima: 5979461e053a02b480dd3d5a81cc6af5ad400611

kani-driver/src/concrete_playback/test_generator.rs
YoshikiTakashima: 930efaf7f4edfaa107774da0e770ce8fe252b931

tests/script-based-pre/playback_already_existing/original.rs
YoshikiTakashima: 5979461e053a02b480dd3d5a81cc6af5ad400611

kani-driver/src/concrete_playback/test_generator.rs
YoshikiTakashima: https://github.com/model-checking/kani/commit/930efaf7f4edfaa107774da0e770ce8fe252b931

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -56,21 +56,28 @@ impl KaniSession {
...
-                        if !self.args.common_args.quiet {
-                            for generated_unit_test in unit_tests.iter() {
-                                println!(
-                                    "INFO: Now modifying the source code to include the concrete playback unit test `{}`.",
-                                    &generated_unit_test.name
-                                );
-                            }
+                        if !self.args.common_args.quiet && !unit_tests.is_empty() {
+                            println!(
+                                "INFO: Now modifying the source code to include the concrete playback unit test: [{}].",
```
celinval, line 71: This might become a very long line. Can you print one harness per line instead?

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -119,7 +126,7 @@ impl KaniSession {
     }
 
     /// Writes the new source code to a user's source file using a tempfile as the means.
-    /// Returns whether new unit test was injected.
+    /// Returns whether the unit test was already in the source code.
```
celinval, line 129: That doesn't match the logic implemented

kani-driver/src/concrete_playback/test_generator.rs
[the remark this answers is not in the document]
celinval: The name and description is still related to adding a single test

celinval (commented):
Just some minor comments. Thanks!

kani-driver/src/concrete_playback/test_generator.rs
YoshikiTakashima: Does 919ef57c4dbe9e858593afe73a6f5529d86faea8 work?  I am not what you want here. One INFO and a list of harness paths?

kani-driver/src/concrete_playback/test_generator.rs
YoshikiTakashima: `!unit_tests.is_empty()` so it returns true if some `unit_tests` were left after filtering out those that exists. These are injected, so the return value is "true" when injection occurs, and "false" when injection did no occur b/c all tests were filtered out.

YoshikiTakashima: b4b33363a22a72e4eaf5acb0156e5530d165060c

kani-driver/src/concrete_playback/test_generator.rs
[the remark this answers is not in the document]
YoshikiTakashima: Completely forgot sorry. b4b33363a22a72e4eaf5acb0156e5530d165060c

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -76,7 +76,7 @@ impl KaniSession {
                                         &generated_unit_test.name
                                     ))
                                     .collect::<Vec<String>>()
-                                    .join(", ")
+                                    .join(",\n")
```
celinval, line 79: sorry, I meant as a list:
```
INFO: Now modifying the source code to include the concrete playback unit test:
  - mod::test_blah
  - mod::test_foo
  - another::test_bar
```

celinval (approved):
Just please prune the info. Thanks!

kani-driver/src/concrete_playback/test_generator.rs
YoshikiTakashima: 0cff7804f8383430a82ffaecf8851f9e3740ef79

PR #2506, opened 2023-06-05
Plumbing for Boogie backend

zhassan-aws:
### Description of changes: 

This PR mainly does the plumbing for the Boogie backend:
1. Introduces a `boogie` feature in the `kani-compiler`
2. Introduces a Boogie codegen backend type that currently just prints "Hello, Boogie!"
3. Introduces an unstable option to select the Boogie backend at runtime

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? Added two tests

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_boogie/compiler_interface.rs
+++ b/kani-compiler/src/codegen_boogie/compiler_interface.rs
@@ -1,0 +1,666 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This file contains the code necessary to interface with the boogie backend
+
+use crate::codegen_boogie::BoogieCtx;
+use crate::kani_middle::analysis;
+use crate::kani_middle::attributes::is_test_harness_description;
+use crate::kani_middle::metadata::gen_test_metadata;
+use crate::kani_middle::provide;
+use crate::kani_middle::reachability::{
+    collect_reachable_items, filter_const_crate_items, filter_crate_items,
+};
+use crate::kani_middle::{check_crate_items, check_reachable_items, dump_mir_items};
+use crate::kani_queries::{QueryDb, ReachabilityType};
+use kani_metadata::artifact::convert_type;
+use kani_metadata::CompilerArtifactStub;
+use kani_metadata::{ArtifactType, HarnessMetadata, KaniMetadata};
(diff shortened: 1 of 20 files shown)


kani-compiler/src/codegen_boogie/compiler_interface.rs
```diff
@@ -0,0 +1,666 @@
...
+                    filter_crate_items(tcx, |_, def_id| items.contains(&tcx.def_path_hash(def_id)));
+                for harness in harnesses {
+                    let model_path =
+                        queries.harness_model_path(&tcx.def_path_hash(harness.def_id())).unwrap();
+                    let (gcx, items) =
+                        self.codegen_items(tcx, &[harness], model_path, &results.machine_model);
+                    results.extend(gcx, items, None);
+                }
+            }
+            ReachabilityType::Tests => {
```
celinval, line 230: FYI, this code will change quite a bit soon... I'll see if I can factor it out so we don't have to replicate it around here.

kani-compiler/src/kani_compiler.rs
```diff
@@ -51,16 +53,53 @@ pub fn run(args: Vec<String>) -> ExitCode {
...
+    rustc_session::early_error(
+        ErrorOutputType::default(),
+        "`--backend cprover` requires enabling the `cprover` feature",
+    );
+}
+
+#[cfg(any(feature = "cprover", feature = "boogie"))]
 fn backend(queries: Arc<Mutex<QueryDb>>) -> Box<dyn CodegenBackend> {
-    Box::new(GotocCodegenBackend::new(queries))
+    let backend = queries.lock().unwrap().backend;
```
celinval, line 80: Just be aware that this logic is rather fragile and we might want to improve this in the near future. There is no guarantee on the order that rustc build the backend object and call `Callbacks::config()`. If the order is inverted, the queries may not be initialized. That is the reason why we share the same QueryDb instead of copying the object at this level.

celinval (approved):
I think we should probably think about moving more stuff out of compiler_interface. I am tempted to create a Codegen trait with a smaller interface than rustc CodegenBackend. So we can reduce the amount of logic duplicated logic.

With that, we can also delay picking the right Codegen in a state that we can guarantee that the queries have been initialized.

PR #2508, opened 2023-06-07
Escape Zero-size types in playback

### LINKED ISSUE #2453: Concrete playback not working for `Result<T, E>` when `T` is `()`
I tried this code:

```rust
#[cfg(kani)]
mod verify {
    #[kani::proof]
    fn any_is_ok() {
        let result: Result<(), bool> = kani::any();
        kani::cover!(result.is_ok());
    }
}

```

```
kani --enable-unstable --concrete-playback=inplace
```

then followed wiki to build and run the harness.

with Kani version: 0.27.0

I expected to see this happen: Harness succeed

Instead, this happened: Harness fails with error:
```
thread 'main' panicked at 'Not enough det vals found', /home/ubuntu/kani/library/kani/src/concrete_playback.rs:52:62

```

YoshikiTakashima:
### Description of changes:

This PR adds an escape in the concrete playback for zero-size types
where the previous implementation was crashing for.

### Resolved issues:

Resolves #2453

### Testing:

* How is this change tested? c7f90440382d22935625f2c74b2907315a84a281

### Checklist


library/kani/src/concrete_playback.rs
```diff
@@ -49,7 +51,11 @@ pub(crate) unsafe fn any_raw_internal<T, const SIZE_T: usize>() -> T {
     let mut next_concrete_val: Vec<u8> = Vec::new();
     CONCRETE_VALS.with(|glob_concrete_vals| {
         let mut_ref_glob_concrete_vals = &mut *glob_concrete_vals.borrow_mut();
-        next_concrete_val = mut_ref_glob_concrete_vals.pop().expect("Not enough det vals found");
+        next_concrete_val = if SIZE_T > 0 {
+            mut_ref_glob_concrete_vals.pop().expect("Not enough det vals found")
+        } else {
+            vec![]
```
celinval, line 57: Does an empty array work? Not a big deal though.

tests/script-based-pre/zero-size/config.yml
```diff
@@ -0,0 +1,4 @@
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+script: playback_opts.sh
```
celinval, line 3: nit: playback_zst?

Also, can you please name the folder playback-zero-size?

celinval (approved):
Thanks

library/kani/src/concrete_playback.rs
YoshikiTakashima: I don't think so. If I got your idea correctly, I don't think an array with defined size will type check with the variable-length case for non-zst. For slice, I think the slice's parent dies inside that closure so borrow checker complains. We need to return something that owns the memory.

tests/script-based-pre/zero-size/config.yml
YoshikiTakashima: 277a15967fe578894eda404be4c3b3751145c0c5

YoshikiTakashima:
That's an interesting bug: 2 harnesses came out with exactly the same names... I'll investigate.

YoshikiTakashima:
OK. This is a bug that is separate. It's the interaction between ZST and multi-harness. Tracking: #2509. I will merge this one with a reduced test case, and then fix that issue.

PR #2514, opened 2023-06-09
Compile harnesses with the same set of stubs together

### LINKED ISSUE #2494: Reduce the compilation overhead for harnesses with the same set of stubs.

celinval:
### Description of changes: 

Improve Kani compiler to compile harnesses that have exact the same set of stubs in a single Rust compiler session. This reduces the number of times we compile a crate when it has multiple harnesses use the same set of stubs.

### Resolved issues:

Resolves #2494 

### Related RFC:

Optional #1809 

### Call-outs:

We could optimize further by looking at the transitive rules of the stubbing logic and by looking at matching subset of stubs for cases where some stubs are unreachable.

### Testing:

### Checklist


tests/script-based-pre/stubbing_compiler_sessions/check_compiler_sessions.sh
```diff
@@ -0,0 +1,25 @@
...
+echo "----------------------------"
+
+# We print the reachability analysis results once for each session.
+# Once we unify the stats per Kani compiler run, we should include number of
+# sessions to the result.
+runs=$(/usr/bin/env grep -c "Reachability Analysis Result" ${log_file})
+echo "Rust compiler sessions: ${runs}"
+
+# Cleanup
+rm ${log_file}
```
feliperodri, line 25: Missing new line.

tests/script-based-pre/stubbing_compiler_sessions/check_compiler_sessions.sh
```diff
@@ -0,0 +1,25 @@
...
+
+kani stubbing.rs --enable-unstable --enable-stubbing --verbose >& ${log_file}
+
+echo "------- Raw output ---------"
+cat $log_file
+echo "----------------------------"
+
+# We print the reachability analysis results once for each session.
+# Once we unify the stats per Kani compiler run, we should include number of
+# sessions to the result.
```
feliperodri, line 20: I don't understand this comment. Why is reachability analysis important here? I can see from the code that we grep and print the reachability analysis results, so the comment is a bit redundant. It'd be nice to explain instead why this is important here. Also, is the unification of stats for the Kani compiler being track in an issue? If yes, we should include the link here so we can address it later. Does the issue include the request for the number of sessions?

tests/script-based-pre/stubbing_compiler_sessions/expected
```diff
@@ -0,0 +1 @@
+VERIFICATION:- SUCCESSFUL
```
feliperodri, line 1: It'd be nice to add a small English description of the goal for this test. I don't know if that is the standard for Kani, but [here](https://github.com/diffblue/cbmc/blob/develop/regression/contracts-dfcc/assigns_enforce_01/test.desc) is an example from the CBMC repository.

tests/script-based-pre/stubbing_compiler_sessions/stubbing.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// Check that Kani handle different sets of stubbing correctly.
```
feliperodri, line 4: This description is the one I refer in the `expected` file. If the standard for Kani is to leave in the test header, it's fine. We mean "handle correct" by "compiling all proof harness with same set of stubs in a single run". Is there a way we can check that they run in a single run using the `expected` file?

kani-compiler/src/kani_compiler.rs
```diff
@@ -166,7 +167,9 @@ impl KaniCompiler {
                 CompilationStage::CodegenNoStubs { .. } => {
                     unreachable!("This stage should always run in the same session as Init");
                 }
-                CompilationStage::CodegenWithStubs { target_harness, all_harnesses, .. } => {
+                CompilationStage::CodegenWithStubs { target_harnesses, all_harnesses, .. } => {
+                    assert!(!target_harnesses.is_empty());
```
feliperodri, line 171: Is that an invariant? Should we provide an error message?

kani-compiler/src/kani_compiler.rs
```diff
@@ -364,3 +380,75 @@ impl Callbacks for KaniCompiler {
         self.prepare_codegen()
     }
 }
+
+#[cfg(test)]
+mod tests {
```
feliperodri, line 385: Love this! Is there any tool the measure test coverage that we should be using in CI?

feliperodri (changes requested):

tests/script-based-pre/stubbing_compiler_sessions/check_compiler_sessions.sh
celinval: This is important because it is the only way today to reliably get the number of sessions from the compiler without relying on debug comments.

tests/script-based-pre/stubbing_compiler_sessions/expected
celinval: The `expected` file should be a subset of the Kani output. It doesn't accept comments or special characters. The only special handling we do is `\` at the end of a line.

tests/script-based-pre/stubbing_compiler_sessions/stubbing.rs
celinval: By correctly I mean that the result is correct, i.e., that all harnesses behave as expected given the set of stubs that are specified on them.

kani-compiler/src/kani_compiler.rs
celinval: Yes, it is an invariant. I can add a message... we don't always bother though. Crashes in the compiler usually come with a user friendly message saying that it is a bug and a link to report the issue.

The message is mostly for our consumption, usually used to narrow down the problem. In cases like this, I feel like we're just stating the obvious, something like: "Expected at least one harness to be processed, but found none".

kani-compiler/src/kani_compiler.rs
celinval: There are tools out there that we could and should be using. Unfortunately, we haven't integrated any yet. I think it's important for us to collect data on end-to-end tests, since most of our testing is done at that level.

tests/script-based-pre/stubbing_compiler_sessions/check_compiler_sessions.sh
feliperodri: Awesome! I'd just add this to the comment.

tests/script-based-pre/stubbing_compiler_sessions/expected
feliperodri: Good to know! So I guess we keep the descriptions in the test itself. Got it!

kani-compiler/src/kani_compiler.rs
feliperodri: Do you mean "end-to-end" test by our current regression tests?

feliperodri (approved):

tests/script-based-pre/stubbing_compiler_sessions/check_compiler_sessions.sh
```diff
@@ -0,0 +1,27 @@
...
+echo "------- Raw output ---------"
+cat $log_file
+echo "----------------------------"
+
+# We print the reachability analysis results once for each session.
+# This is the only reliable way to get the number of sessions from the compiler.
+# The other option would be to use debug comments.
+# Ideally, the compiler should only print one set of statistics at the end of its run.
+# In that case, we should include number of sessions to those stats.
+runs=$(/usr/bin/env grep -c "Reachability Analysis Result" ${log_file})
```
zhassan-aws, line 23: What is the purpose of `/usr/bin/env`?

tests/script-based-pre/stubbing_compiler_sessions/expected
zhassan-aws, line 1: Is this `expected` file used? Isn't `check_compiler_sessions.expected` sufficient?

tests/script-based-pre/stubbing_compiler_sessions/stubbing.rs
```diff
@@ -0,0 +1,64 @@
...
```
zhassan-aws, line 4: Nit:
```suggestion
// Check that Kani handles different sets of stubbing correctly.
```

tests/script-based-pre/stubbing_compiler_sessions/stubbing.rs
```diff
@@ -0,0 +1,64 @@
...
+#[kani::stub(increment, identity)]
+fn check_all_identity_2() {
+    let n = kani::any();
+    assert_eq!(decrement(n), n);
+    assert_eq!(increment(n), n);
+}
+
+#[kani::proof]
+#[kani::stub(decrement, increment)]
+#[kani::stub(increment, identity)]
```
zhassan-aws, line 59: I didn't know stubs are applied transitively!

zhassan-aws (approved):

tests/script-based-pre/stubbing_compiler_sessions/check_compiler_sessions.sh
celinval: not needed. I'll remove

tests/script-based-pre/stubbing_compiler_sessions/expected
celinval: yes... this is garbage. Let me remove

tests/script-based-pre/stubbing_compiler_sessions/stubbing.rs
celinval: I don't think this was necessarily planned, but that's how it works today and I wanted to make sure this PR wouldn't have unexpected side effects. :)

Issue #2299, opened 2023-03-13
Option that causes verification to fail for uncoverable cover statements


adpaco-aws:
The [RFC for cover statements](https://model-checking.github.io/kani/rfc/rfcs/0003-cover-statement.html) mentioned:

> We can consider adding an option that would cause verification to fail if a cover property was unsatisfiable or unreachable, e.g. --fail-uncoverable.

We are now aware that some users want this option.

It'd be great if we could include this option as part of the "more granular expectations" discussed [here](https://github.com/model-checking/kani/pull/2272). Let's say we had a language to describe expectations for regular properties and cover statements. Then, this expectation is essentially that all cover statements are satisfied.

adpaco-aws:
We've agreed on supplying the option as it was described in the RFC. In other words, we don't need to include as part of the "more granular expectations" work since it's unknown whether they would be useful or not.

The main problem with this option is how it interacts with `#[kani::should_panic]` and other attributes. There isn't a conflict per se (verification will be successful if both verification panics AND cover statements are covered), but I'm afraid that adding more options that modify the verification result complicates post-processing too much.

PR #2516, opened 2023-06-12
RFC: Global Conditions

adpaco-aws:
### Description of changes: 

This RFC proposes a new section in Kani's output to summarize the status of properties that depend on other properties, referred to as "Global Conditions". Originally, I wanted this change to be part of #2299, but then I realized a small RFC might be appropriate.

### Resolved issues:

N/A

### Testing:

### Checklist

### DIFF
--- a/rfc/src/rfcs/0007-global-conditions.md
+++ b/rfc/src/rfcs/0007-global-conditions.md
@@ -67,7 +67,7 @@ However, we may consider adding a new subsection `Global conditions` to the `Ref
 ## Detailed Design
 
 The only component to be modified is `kani-driver` since that's where verification results are built and determined.
-But given the growing complexity of `kani-driver`, we should consider moving the logic related to global conditions onto a new crate.
+But we should consider moving this logic into another crate.
 
 We don't need new dependencies.
 The corner cases will depend on the specific global conditions to be implemented.
@@ -77,9 +77,6 @@ The corner cases will depend on the specific global conditions to be implemented
 As mentioned earlier, we're proposing this change to help users understand global conditions and how they're determined.
 In many cases, global conditions empower users to write harnesses which weren't possible to write before.
 As an example, the `#[kani::should_panic]` attribute allowed users to write harnesses expecting panic-related failures.
-
-On the other hand, this proposal will add a significant amount of code to `kani-driver`.
-It's possible to move some code into another crate, but even then some code will be added in `kani-driver`.
 
 Also, we don't really know if more global conditions will be requested in the future.
 We may consider discarding this proposal and waiting for the next feature that can be implemented as a global condition to be requested.
--- a/rfc/src/rfcs/0007-global-conditions.md
+++ b/rfc/src/rfcs/0007-global-conditions.md
@@ -25,7 +25,7 @@ This will help users to understand better options that are enabled through globa
 ## User Experience
 
 **The output will refer to properties that depend on other properties as "global conditions"**, which is a simpler term.
-The options to enable different global conditions will depend on a case-by-case basis.
+The options to enable different global conditions will depend on a case-by-case basis[^enable-options].
 
 The main UI change in this proposal is a new `GLOBAL CONDITIONS` section that **won't be printed if no global conditions have been enabled**.
 This section will only appear in Kani's default output after the `RESULTS` section (used for individual checks) and have the format:
@@ -103,3 +103,7 @@ Global conditions' checks aren't part of the SAT, therefore this computation won
 In particular, the description field is the only information printed for properties (such as `cover` statements) without trace locations.
 There are additional improvements we should consider: printing the actual status (for global conditions, this won't always be `FAILED`), avoid the repetition of `Failed Checks: `, etc.
 [This comment](https://github.com/model-checking/kani/pull/2516#issuecomment-1597524184) discusses problems with the current interface on some examples.
+
+[^enable-options]: In other words, global conditions won't force a specific mechanism to be enabled.
+For example, if the `#[kani::should_panic]` attribute is converted into a global condition, it will continue to be enabled through the attribute itself.
+Other global conditions may be enabled through CLI flags only (e.g., `--fail-uncoverable`), or a combination of options in general.
--- a/rfc/src/rfcs/0007-global-conditions.md
+++ b/rfc/src/rfcs/0007-global-conditions.md
@@ -1,7 +1,7 @@
 - **Feature Name:** Global Conditions (`global-conditions`)
 - **Feature Request Issue:** <https://github.com/model-checking/kani/issues/2525>
 - **RFC PR:** <https://github.com/model-checking/kani/pull/2516>
-- **Status:** *Under Review*
+- **Status:** *Unstable*
 - **Version:** 0
 - **Proof-of-concept:** <https://github.com/model-checking/kani/pull/2532>
 
--- a/rfc/src/rfcs/0007-global-conditions.md
+++ b/rfc/src/rfcs/0007-global-conditions.md
@@ -1,35 +1,34 @@
-- **Feature Name:** `global-conditions`
-- **Feature Request Issue:** *N/A*
-- **RFC PR:** *Link to original PR*
+- **Feature Name:** Global Conditions (`global-conditions`)
+- **Feature Request Issue:** <https://github.com/model-checking/kani/issues/2525>
+- **RFC PR:** <https://github.com/model-checking/kani/pull/2516>
 - **Status:** *Under Review*
 - **Version:** 0
 - **Proof-of-concept:** TBD
 
 ## Summary
 
-A new section in Kani's output to summarize the status of hyperproperties.
+A new section in Kani's output to summarize the status of properties that depend on other properties. We use the term *global conditions* to refer to such properties. 
 
 ## User Impact
 
 The addition of new options that affect the overall verification result depending on certain property attributes demands some consideration.
 In particular, the addition of a new option to fail verification if there are uncoverable (i.e., unsatisfiable or unreachable) `cover` properties (requested in [#2299](https://github.com/model-checking/kani/issues/2299)) is posing new challenges to our current architecture and UI.
 
-Hyperproperties are properties that depend on attributes of other properties.
 This concept isn't made explicit in Kani, but exists in some ways.
-For example, the `kani::should_panic` attribute is a hyperproperty because it can be described in terms of other properties (checks).
-The request in [#2299](https://github.com/model-checking/kani/issues/2299) is essentially another hyperproperty, and we may expect more hyperproperties to be requested in the future.
+For example, the `kani::should_panic` attribute is a global condition because it can be described in terms of other properties (checks).
+The request in [#2299](https://github.com/model-checking/kani/issues/2299) is essentially another global conditions, and we may expect more to be requested in the future.
 
-In this RFC, we propose a new section in Kani's output focused on hyperproperty reporting.
+In this RFC, we propose a new section in Kani's output focused on reporting global conditions.
 The goal is for users to receive useful information about hyperproperties without it becoming overwhelming.
-This will help users to understand better hyperproperty-related options and ease the addition of such options to Kani.
+This will help users to understand better options that are enabled through global conditions and ease the addition of such options to Kani.
 
 ## User Experience
 
-**The output will refer to hyperproperties as "global conditions"**, which is a simpler term.
+**The output will refer to properties that depend on other properties as "global conditions"**, which is a simpler term.
 The options to enable different global conditions will depend on a case-by-case basis.
 
 The main UI change in this proposal is a new `GLOBAL CONDITIONS` section that **won't be printed if no global conditions have been enabled**.
-This section will appear after the `RESULTS` section (used for individual checks) and have the format:
+This section will only appear in Kani's default output after the `RESULTS` section (used for individual checks) and have the format:
 
 ```
 GLOBAL CONDITIONS:
@@ -57,6 +56,8 @@ A `FAILED` status in any enabled global condition will cause verification to fai
 VERIFICATION:- FAILURE (one or more global conditions failed)
 ```
 
+This last UI change will also be implemented for the terse output.
+
 **Global conditions which aren't enabled won't appear in the `GLOBAL CONDITIONS` section**.
 Their status will be computed regardless, and we may consider showing this status when the `--verbose` option is passed.
 
@@ -73,20 +74,20 @@ The corner cases will depend on the specific global conditions to be implemented
 
 ## Rationale and alternatives
 
-As mentioned earlier, we're proposing this change to help users understand hyperproperties and how they're determined.
-In many cases, hyperproperties empower users to write harnesses which weren't possible to write before.
+As mentioned earlier, we're proposing this change to help users understand global conditions and how they're determined.
+In many cases, global conditions empower users to write harnesses which weren't possible to write before.
 As an example, the `#[kani::should_panic]` attribute allowed users to write harnesses expecting panic-related failures.
 
 On the other hand, this proposal will add a significant amount of code to `kani-driver`.
 It's possible to move some code into another crate, but even then some code will be added in `kani-driver`.
 
-Also, we don't really know if more hyperproperties will be requested in the future.
-We may consider discarding this proposal and waiting for the next hyperproperty-related feature to be requested.
+Also, we don't really know if more global conditions will be requested in the future.
+We may consider discarding this proposal and waiting for the next feature that can be implemented as a global condition to be requested.
 
 ### Alternative: Global conditions as regular checks
 
-One option we've considered in the past is to enable hyperproperties as a regular checks.
-While it's technically doable, it doesn't feel appropriate for hyperproperties to reported through regular since generally a higher degree of visibility will be appreciated.
+One option we've considered in the past is to enable global conditions as a regular checks.
+While it's technically doable, it doesn't feel appropriate for global conditions to reported through regular since generally a higher degree of visibility will be appreciated.
 
 ## Open questions
 
--- a/rfc/src/rfcs/0007-global-conditions.md
+++ b/rfc/src/rfcs/0007-global-conditions.md
@@ -3,7 +3,7 @@
 - **RFC PR:** <https://github.com/model-checking/kani/pull/2516>
 - **Status:** *Under Review*
 - **Version:** 0
-- **Proof-of-concept:** TBD
+- **Proof-of-concept:** <https://github.com/model-checking/kani/pull/2532>
 
 ## Summary
 
--- a/rfc/src/rfcs/0007-global-conditions.md
+++ b/rfc/src/rfcs/0007-global-conditions.md
@@ -5,6 +5,8 @@
 - **Version:** 0
 - **Proof-of-concept:** <https://github.com/model-checking/kani/pull/2532>
 
+-------------------
+
 ## Summary
 
 A new section in Kani's output to summarize the status of properties that depend on other properties. We use the term *global conditions* to refer to such properties. 
--- a/rfc/src/rfcs/0007-global-conditions.md
+++ b/rfc/src/rfcs/0007-global-conditions.md
@@ -1,0 +1,97 @@
+- **Feature Name:** `global-conditions`
+- **Feature Request Issue:** *N/A*
+- **RFC PR:** *Link to original PR*
+- **Status:** *Under Review*
+- **Version:** 0
+- **Proof-of-concept:** TBD
+
+## Summary
+
+A new section in Kani's output to summarize the status of hyperproperties.
+
+## User Impact
+
+The addition of new options that affect the overall verification result depending on certain property attributes demands some consideration.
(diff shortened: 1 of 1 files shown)


celinval (commented):
It makes sense.

zhassan-aws:
Link to rendered version: https://github.com/adpaco-aws/rmc/blob/update-other-rfcs/rfc/src/rfcs/0007-global-conditions.md

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,97 @@
...
+## Summary
+
+A new section in Kani's output to summarize the status of hyperproperties.
+
+## User Impact
+
+The addition of new options that affect the overall verification result depending on certain property attributes demands some consideration.
+In particular, the addition of a new option to fail verification if there are uncoverable (i.e., unsatisfiable or unreachable) `cover` properties (requested in [#2299](https://github.com/model-checking/kani/issues/2299)) is posing new challenges to our current architecture and UI.
+
+Hyperproperties are properties that depend on attributes of other properties.
```
zhassan-aws, line 17: Is this a standard definition of hyperproperties? The definition I'm aware of is that while properties refer to traces, hyperproperties refer to sets of traces:
https://en.wikipedia.org/wiki/Hyperproperty
The definition used here doesn't seem to match with that.

karkhaz (commented):
It seems to me that there are two commonalities in @adpaco-aws's examples of `should_panic` and `fail_uncoverable`:

* Their statuses are derived entirely from the status of other properties, and not directly a result from the SAT solver
* They are of the form "{panic,coverage} did not happen along any path" rather than "{overflow} happened along some path".

Would you expect these to be true for other global conditions that might be added in the future? I think the first bullet point is how you're defining hyperproperty/global condition, but I think the second bullet point is more relevant to users and might be a good reason to separate hyperproperties out from regular ones.

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,97 @@
...
+But given the growing complexity of `kani-driver`, we should consider moving the logic related to global conditions onto a new crate.
+
+We don't need new dependencies.
+The corner cases will depend on the specific global conditions to be implemented.
+
+## Rationale and alternatives
+
+As mentioned earlier, we're proposing this change to help users understand hyperproperties and how they're determined.
+In many cases, hyperproperties empower users to write harnesses which weren't possible to write before.
+As an example, the `#[kani::should_panic]` attribute allowed users to write harnesses expecting panic-related failures.
```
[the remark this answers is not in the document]
celinval: Will unwinding assertions fall into the global conditions category?

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,97 @@
+- **Feature Name:** `global-conditions`
+- **Feature Request Issue:** *N/A*
```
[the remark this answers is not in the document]
adpaco-aws: Done.

rfc/src/rfcs/0007-global-conditions.md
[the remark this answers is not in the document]
adpaco-aws: Done, see [this comment](https://github.com/model-checking/kani/pull/2516#discussion_r1229770464).

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,97 @@
...
+- **RFC PR:** *Link to original PR*
+- **Status:** *Under Review*
+- **Version:** 0
+- **Proof-of-concept:** TBD
+
+## Summary
+
+A new section in Kani's output to summarize the status of hyperproperties.
```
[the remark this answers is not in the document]
adpaco-aws: Sorry about that, please see [this comment](https://github.com/model-checking/kani/pull/2516#discussion_r1229770464).

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,97 @@
...
+In this RFC, we propose a new section in Kani's output focused on hyperproperty reporting.
+The goal is for users to receive useful information about hyperproperties without it becoming overwhelming.
+This will help users to understand better hyperproperty-related options and ease the addition of such options to Kani.
+
+## User Experience
+
+**The output will refer to hyperproperties as "global conditions"**, which is a simpler term.
+The options to enable different global conditions will depend on a case-by-case basis.
+
+The main UI change in this proposal is a new `GLOBAL CONDITIONS` section that **won't be printed if no global conditions have been enabled**.
```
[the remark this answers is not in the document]
adpaco-aws: The RFC now explicitly mentions the terse output now. My proposal is to only include the last change (i.e., the one that changes overall result).

adpaco-aws:
> I cannot think of other "Global Conditions" that may be introduced in the future. Do you have any in mind?

 Yes, the check for expectations on individual properties mentioned [here](https://model-checking.github.io/kani/rfc/rfcs/0005-should-panic-attr.html#alternative-2-name) could be implemented as a global condition. Another global condition that I think could be interesting for users would check that specific properties (e.g., user-defined assertions) are not unreachable. This can already be manually done by adding `cover` statements, but that isn't always an option (because the user may not want to add them) and users may not realize they are unreachable if they appear in deep places in the call graph.

But one of the cool things about global conditions is that we could also define them for existing properties. For example, we could have a `memory safety` condition that summarizes the status memory-safety checks (and hide them from `RESULTS` if all went well). Or do the same for unwinding assertions, as @celinval mentioned. If we did that, this section could become the main component of Kani's output.

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+A `FAILED` status in any enabled global condition will cause verification to fail, pointing out that one or more global conditions were failed as in:
+
+```
+VERIFICATION:- FAILURE (one or more global conditions failed)
+```
+
+**Global conditions which aren't enabled won't appear in the `GLOBAL CONDITIONS` section**.
+Their status will be computed regardless, and we may consider showing this status when the `--verbose` option is passed.
```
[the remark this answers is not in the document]
adpaco-aws: I've clarified this with a footnote.

celinval:
> > I cannot think of other "Global Conditions" that may be introduced in the future. Do you have any in mind?
> 
> Yes, the check for expectations on individual properties mentioned [here](https://model-checking.github.io/kani/rfc/rfcs/0005-should-panic-attr.html#alternative-2-name) could be implemented as a global condition. Another global condition that I think could be interesting for users would check that specific properties (e.g., user-defined assertions) are not unreachable. This can already be manually done by adding `cover` statements, but that isn't always an option (because the user may not want to add them) and users may not realize they are unreachable if they appear in deep places in the call graph.
> 
> But one of the cool things about global conditions is that we could also define them for existing properties. For example, we could have a `memory safety` condition that summarizes the status memory-safety checks (and hide them from `RESULTS` if all went well). Or do the same for unwinding assertions, as @celinval mentioned. If we did that, this section could become the main component of Kani's output.

Please do replace the memory safety and arithmetic checks by global conditions! :) It would result in a much cleaner output.

One thing I'm missing in this RFC is an explanation on how the user could easily find the root cause of the failing global condition. For example, for the `fail_uncovered`, do we expect users to go inspect all the covers or can we print the uncovered ones the same way we print failed checks at the end?

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,97 @@
...
+On the other hand, this proposal will add a significant amount of code to `kani-driver`.
+It's possible to move some code into another crate, but even then some code will be added in `kani-driver`.
+
+Also, we don't really know if more hyperproperties will be requested in the future.
+We may consider discarding this proposal and waiting for the next hyperproperty-related feature to be requested.
+
+### Alternative: Global conditions as regular checks
+
+One option we've considered in the past is to enable hyperproperties as a regular checks.
+While it's technically doable, it doesn't feel appropriate for hyperproperties to reported through regular since generally a higher degree of visibility will be appreciated.
```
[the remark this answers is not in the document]
adpaco-aws: I've changed this to "may". My take on this is that global condition condense information that involves several checks, so they shouldn't be displayed as regular ones. But I don't have any evidence, it's still a subjective statement.

adpaco-aws:
> They are of the form "{panic,coverage} did not happen along any path" rather than "{overflow} happened along some path".

Actually, the `should_panic` condition expresses "there exists at least one path with an assertion-related failure". I'm defining global conditions as in the first item because that's what I'm trying to generalize here, and I'd expect new ones to follow this rule but I can't guarantee it. But that's OK, we can always go back to this RFC and update it 😄

adpaco-aws:
> One thing I'm missing in this RFC is an explanation on how the user could easily find the root cause of the failing global condition. For example, for the `fail_uncovered`, do we expect users to go inspect all the covers or can we print the uncovered ones the same way we print failed checks at the end?

That's a great point, but I'm concerned about the output getting longer and less comprehensible if we include that change in this RFC. My preference here would be to keep it simple for now, but I'm happy to talk about it.

rfc/src/rfcs/0007-global-conditions.md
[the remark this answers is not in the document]
adpaco-aws: Sorry, missed your comments above. My answers:
 1. No, this wouldn't change how `should_panic` behaves. It would only change its UI.
 2. Yes, a global condition can be used to represent the result of all unwinding conditions.

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,97 @@
...
+We don't need new dependencies.
+The corner cases will depend on the specific global conditions to be implemented.
+
+## Rationale and alternatives
+
+As mentioned earlier, we're proposing this change to help users understand hyperproperties and how they're determined.
+In many cases, hyperproperties empower users to write harnesses which weren't possible to write before.
+As an example, the `#[kani::should_panic]` attribute allowed users to write harnesses expecting panic-related failures.
+
+On the other hand, this proposal will add a significant amount of code to `kani-driver`.
```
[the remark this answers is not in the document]
adpaco-aws: It's actually less than 200 LoC now that I've checked. I'll remove this sentence.

rfc/src/rfcs/0007-global-conditions.md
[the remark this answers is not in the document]
adpaco-aws: Yeah, then I think it's better to not distinguish here and just say that the "global conditions" section will be printed with all failing global conditions for the terse output. That'd be the only difference at this point.

celinval:
> > One thing I'm missing in this RFC is an explanation on how the user could easily find the root cause of the failing global condition. For example, for the `fail_uncovered`, do we expect users to go inspect all the covers or can we print the uncovered ones the same way we print failed checks at the end?
> 
> That's a great point, but I'm concerned about the output getting longer and less comprehensible if we include that change in this RFC. My preference here would be to keep it simple for now, but I'm happy to talk about it.

I agree that our output is pretty long and it needs refactoring ASAP.

However, I think that printing the uncovered statements at the end will alleviate the problem of our super verbose output.  At least the information that the user really needs is readily available at the end of the report, just like failed properties.

adpaco-aws:
> However, I think that printing the uncovered statements at the end will alleviate the problem of our super verbose output. At least the information that the user really needs is readily available at the end of the report, just like failed properties.

Makes sense. I've been extending the current implementation (in #2532) to use the function that reports failed checks for standard verification jobs, but it needs improvements. For example, for a simple harness which contains an unreachable `cover` statement we get this information:

```
RESULTS:
Check 1: harness_uncoverable.cover.1
         - Status: UNREACHABLE
         - Description: "cover location"
         - Location: tests/ui/fail-uncoverable/uncoverable/test.rs:13:9 in function harness_uncoverable

GLOBAL CONDITIONS:
 - fail_uncoverable: FAILURE (encountered one or more cover statements which were not satisfied)
Failed Checks: cover location

SUMMARY:
 ** 0 of 0 failed

 ** 0 of 1 cover properties satisfied (1 unreachable)


VERIFICATION:- FAILED (one or more global conditions failed)
Verification Time: 0.02321334s
```

Note that it only says "Failed Checks: cover location", but it doesn't attach a location despite showing it for the property because the trace doesn't refer to this location, causing us to return earlier. We should make some changes here, also because the "Failed Checks: " gets printed for each item (the results below are for `tests/ui/should-panic-attribute/unexpected-failures/test.rs`) which is causing the output to be cluttered and a little confusing:

```
SUMMARY:
 ** 3 of 18 failed
Failed Checks: panicked on the `1` arm!
 File: "/home/ubuntu/kani/tests/ui/should-panic-attribute/unexpected-failures/test.rs", line 18, in check
Failed Checks: panicked on the `0` arm!
 File: "/home/ubuntu/kani/tests/ui/should-panic-attribute/unexpected-failures/test.rs", line 17, in check
Failed Checks: shift distance too large
```
 In any case, I agree with your point - I'll change the RFC with a new proposal to report these failed checks.

celinval (approved):

celinval:
Since this is an RFC, please wait for a second approval before merging.

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,105 @@
...
+VERIFICATION:- FAILURE (one or more global conditions failed)
+```
+
+This last UI change will also be implemented for the terse output.
+Finally, checks that cause an enabled global condition to fail will be reported using the same interface we use for failed checks[^failed-checks].
+
+**Global conditions which aren't enabled won't appear in the `GLOBAL CONDITIONS` section**.
+Their status will be computed regardless[^status-computation], and we may consider showing this status when the `--verbose` option is passed.
+
+The documentation of global conditions will depend on how they're enabled, which depends on a case-by-case basis.
```
qinheping, line 66: How will users specify and enable global conditions? Can you give an example here?

adpaco-aws: Added a clarification regarding this. As mentioned, global conditions won't force a single mechanism to be enabled. This will rather depend on the specific feature (attributes, CLI options, or combinations of both).

karkhaz (approved):
Awesome. Thanks for working on this!!

jaisnan (approved):
Thanks :) !

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,111 @@
...
+- **RFC PR:** <https://github.com/model-checking/kani/pull/2516>
+- **Status:** *Unstable*
+- **Version:** 0
+- **Proof-of-concept:** <https://github.com/model-checking/kani/pull/2532>
+
+-------------------
+
+## Summary
+
+A new section in Kani's output to summarize the status of properties that depend on other properties. We use the term *global conditions* to refer to such properties.
```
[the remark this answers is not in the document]
remi-delmas-3000: So are global conditions solely defined as conjunctions of assert/cover statement or are there other ways of defining them ?

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,111 @@
...
+
+```
+GLOBAL CONDITIONS:
+ - `fail_uncoverable`: SUCCESS (all cover statements were satisfied as expected)
+```
+
+A `FAILED` status in any enabled global condition will cause verification to fail.
+In that case, the overall verification result will point out that one or more global conditions failed, as in:
+
+```
```
remi-delmas-3000, line 58: Is it possible to list the basic conditions that failed to explain why a global condition failed ?

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,111 @@
...
+
+The addition of new options that affect the overall verification result depending on certain property attributes demands some consideration.
+In particular, the addition of a new option to fail verification if there are uncoverable (i.e., unsatisfiable or unreachable) `cover` properties (requested in [#2299](https://github.com/model-checking/kani/issues/2299)) is posing new challenges to our current architecture and UI.
+
+This concept isn't made explicit in Kani, but exists in some ways.
+For example, the `kani::should_panic` attribute is a global condition because it can be described in terms of other properties (checks).
+The request in [#2299](https://github.com/model-checking/kani/issues/2299) is essentially another global conditions, and we may expect more to be requested in the future.
+
+In this RFC, we propose a new section in Kani's output focused on reporting global conditions.
+The goal is for users to receive useful information about hyperproperties without it becoming overwhelming.
```
remi-delmas-3000, line 24: is "hyperproperties" meant as in "a property that relates two or more executions of the same program?"

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,111 @@
...
+
+For example, let's assume we implement the option requested in [#2299](https://github.com/model-checking/kani/issues/2299).
+A concrete example of this output would be:
+
+```
+GLOBAL CONDITIONS:
+ - `fail_uncoverable`: SUCCESS (all cover statements were satisfied as expected)
+```
+
+A `FAILED` status in any enabled global condition will cause verification to fail.
```
remi-delmas-3000, line 55: Are global conditions only enabled using special flags or can the user also define his own groups of dependent conditions ?

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,111 @@
...
+But we should consider moving this logic into another crate.
+
+We don't need new dependencies.
+The corner cases will depend on the specific global conditions to be implemented.
+
+## Rationale and alternatives
+
+As mentioned earlier, we're proposing this change to help users understand global conditions and how they're determined.
+In many cases, global conditions empower users to write harnesses which weren't possible to write before.
+As an example, the `#[kani::should_panic]` attribute allowed users to write harnesses expecting panic-related failures.
```
remi-delmas-3000, line 83: the documentation of `should_panic` does not really give a precise meaning to the check, i.e. it does not say if at least one execution must trigger a panic or if all executions must trigger a panic. How can it be modelled in terms of dependent conditions  and statuses ?

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,111 @@
...
+
+Also, we don't really know if more global conditions will be requested in the future.
+We may consider discarding this proposal and waiting for the next feature that can be implemented as a global condition to be requested.
+
+### Alternative: Global conditions as regular checks
+
+One option we've considered in the past is to enable global conditions as a regular checks.
```
remi-delmas-3000, line 90: What is a regular check ? kani::assert ? kani::cover ?

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,111 @@
...
+While it's technically doable, it doesn't feel appropriate for global conditions to reported through regular checks since generally a higher degree of visibility may be appreciated.
+
+## Open questions
+
+No open questions.
```
remi-delmas-3000, line 95: is it possible to give a more explicit and rigorous definition of global conditions ? 

e.g. "a global condition is specified by a set of assert or cover statements and an aggregation function that computes a status for the global conditions from the statuses of the basic conditions."

rfc/src/rfcs/0007-global-conditions.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Future possibilities
+
+A redesign of Kani's output is likely to change the style/architecture to report global conditions.
+
+[^status-computation]: The results for global conditions would be computed during postprocessing based on the results of other checks.
```
remi-delmas-3000, line 101: so the status of a global condition could differ when the switch `--fail-uncoverable` is activated vs not activated ?

rfc/src/rfcs/0007-global-conditions.md
adpaco-aws: I don't think we do that for any of the checks we currently produce. Regardless, I think it's well defined in [here](https://model-checking.github.io/kani/reference/attributes.html#kanishould_panic), and if you think otherwise I'm happy to do any editions you consider appropriate.

rfc/src/rfcs/0007-global-conditions.md
adpaco-aws: Check is the simpler term we use to refer to properties. Running Kani on any will output something like this:
```
RESULTS:
Check 1: example.assertion.1
         - Status: <status>
         - Description: <description>
         - Location: <location>
[...]
```

So this refers to those checks/properties.

rfc/src/rfcs/0007-global-conditions.md
adpaco-aws: Maybe? But it's definitely not that.

I'm actually thinking about growing the scope of global conditions to consider things other than other properties. For example, a timeout feature could be implemented a global condition. However, the current definition of global condition does not consider resources like memory or time, only verification properties.

rfc/src/rfcs/0007-global-conditions.md
adpaco-aws: No. The status computed for the condition itself won't differ. But activating the switch will simply cause the status of the global condition to affect the overall verification result.

PR #2527, opened 2023-06-14
Add `--exact` flag

### LINKED ISSUE #2285: Kani doesn't have an `--exact` flag to call for singular harnesses
Requested feature:

Like rust tests, `kani` needs an `--exact` flag that matches against a harness instead of the substring match that happens by default. substring match needs to be the default method, but users should be able to specify harnesses.

jaisnan:
### Description of changes: 

When the user calls kani with a command `cargo kani --harness module::check_blah --exact` and they want to specify which harness to match, using --exact, they can prevent multiple harnesses from being run. The user will need to specify the harness name by using the fully qualified name as well.

### Resolved issues:

Resolves #2285 

### Call-outs:

If the user adds --exact and they want to run `kani` on multiple harnesses, they will need to provide the specific name for every harness.

### Testing:

* How is this change tested? Unit and regression change

### Checklist


kani-driver/src/args/mod.rs
```diff
@@ -177,6 +177,10 @@ pub struct VerificationArgs {
     )]
     pub harnesses: Vec<String>,
 
+    /// If specified, with the fully qualified name of the harness provided, runs that specific harness and nothing more
```
celinval, line 180: What about: "When specified, the harness filter will only match the exact fully qualified name of a harness"

celinval: Can you also please add a note to the `harness` documentation saying that `--exact` will change its behavior?

zhassan-aws (commented):
Thanks @jaisnan! I was expecting to see at least one test with two or more harnesses where the name of one harness is a substring of the other, and `--exact --harness` is used for selecting the one that is a substring while making sure the other one is not selected.

kani-driver/src/metadata.rs
```diff
@@ -191,31 +203,83 @@ mod tests {
...
         assert!(
-            find_proof_harnesses(BTreeSet::from([&"check_one".to_string()]), &ref_harnesses)
+            find_proof_harnesses(BTreeSet::from([&"check_one".to_string()]), &ref_harnesses, false)
                 .first()
                 .unwrap()
                 .mangled_name
                 == "check_one"
         );
+
+        // Check with exact match
```
celinval, line 239: Can you please split this into two tests? From here on, it should be `check_find_proof_harness_with_exact`.

kani-driver/src/metadata.rs
```diff
@@ -191,31 +203,83 @@ mod tests {
...
+                BTreeSet::from([&"check_three".to_string()]),
+                &ref_harnesses,
+                true
+            )
+            .len(),
+            0
+        );
+        assert_eq!(
+            find_proof_harnesses(BTreeSet::from([&"check_two".to_string()]), &ref_harnesses, true)
+                .first()
```
celinval, line 251: Can you change the checks where no harness should match the name as `is_empty()` instead?

kani-driver/src/metadata.rs
```diff
@@ -191,31 +203,83 @@ mod tests {
...
+            true
+        );
+        assert_eq!(
+            find_proof_harnesses(BTreeSet::from([&"check_one".to_string()]), &ref_harnesses, true)
+                .first()
+                .unwrap()
+                .mangled_name,
+            "check_one"
+        );
+        assert_ne!(
```
celinval, line 262: What are you trying to check here? Why `ne()`?

tests/ui/exact-harness/some_matching_harnesses/subset.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: --harness existing --harness check_blah --exact
+//! Check that we just ignore non-matching filters
```
celinval, line 4: Unfortunately, our expected doesn't work for things that shouldn't show up. This test would still succeed if Kani had verified any set of harnesses that include the harness `existing`.

Maybe just use the same trick as you did with `ignored_harness` where the verification would fail if the harness was checked.

tests/ui/exact-harness/incomplete-harness-name/expected
```diff
@@ -0,0 +1 @@
+error: no harnesses matched the harness filter: `check_blah`
```
celinval, line 1: nit: can you change this message for the case where `--exact` is provided? something like:
```
error: no harnesses matched the name: `check_blah`
```

celinval (commented):
It looks good. Just a few minor comments.

I do wonder if we should fail in the case where an exact match was not found. I don't think rust tests fail, but they also don't fail when a filter has no match.

@adpaco-aws might have an opinion about this one. @jaisnan do you have any preference?

celinval:
> Thanks @jaisnan! I was expecting to see at least one test with two or more harnesses where the name of one harness is a substring of the other, and `--exact --harness` is used for selecting the one that is a substring while making sure the other one is not selected.

I think the test [tests/ui/exact-harness/some_matching_harnesses/subset.rs](https://github.com/model-checking/kani/pull/2527/files#diff-dcbce9bea16ddcd3488c3e381d56785ad85f65a7fe6f4045762471c2b3c70954) was design to test that, but I don't think it is quite testing it yet.

tests/ui/exact-harness/some_matching_harnesses/subset.rs
zhassan-aws: Another possibility is to make the test run multiple harnesses, and include the `Complete - <N> successfully verified harnesses, 0 failures, <N> total.` with the correct number in the `expected` file.

@celinval perhaps Kani should always print the number of matched harnesses (even when it's just one)? This would make it easier to write expected files.

jaisnan: I agree with that change, it would be useful for the extension as well.

kani-driver/src/args/mod.rs
jaisnan: By documentation, I'm assuming you mean the doc comment written in the same struct above? Can do

tests/ui/exact-harness/some_matching_harnesses/subset.rs
celinval: I like the idea of always printing the summary.

adpaco-aws: +1 to printing the summary even for one harness.

adpaco-aws:
> I do wonder if we should fail in the case where an exact match was not found. I don't think rust tests fail, but they also don't fail when a filter has no match.
>
> @adpaco-aws might have an opinion about this one.

Yes, I do. We should fail in that case. In addition to "catch" mistakes from users, it'll also be helpful to detect issues with target selection on the VSCode extension.

kani-driver/src/metadata.rs
jaisnan: Trying to check that the harness value returned is `module::not_check_three` and not any substring of that string. Would it be a redundant test?

celinval (approved):
We should consider adding the number of ignored harnesses to the summary. But that's out of the scope of this change. Thanks Jai

jaisnan:
@celinval I've added the error when an exact match is not found.

kani-driver/src/metadata.rs
```diff
@@ -121,11 +121,39 @@ impl KaniSession {
...
+                let harnesses_missing: Vec<&String> =
+                    all_targets.difference(&harness_found_names).cloned().collect();
+                let joined_string = harnesses_missing
+                    .iter()
+                    .map(|&s| (*s).clone())
+                    .collect::<Vec<String>>()
+                    .join("`, `");
+
+                bail!(
+                    "Please provide exact harness name. The harnesses `{}` don't contain the full name.",
```
zhassan-aws, line 151: I suggest rewording the message to:
```
Failed to match the following harness(es):\n{joined_string}\nPlease specify the fully-qualified name of a harness.
```

tests/ui/exact-harness/fail_on_missing/subset.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: --harness existing --harness check_blah --harness check_foo --exact
+//! Check that we just ignore non-matching filters
```
zhassan-aws, line 4: This should say "Check that we error out with non-matching filters when `--exact` is used".

tests/ui/exact-harness/incomplete-harness-name/incomplete_harness.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: --harness ignore_third_harness --exact
+//! Check that we just ignore non-matching filters
```
zhassan-aws, line 4: Same here

zhassan-aws (approved):
Thanks @jaisnan!

Issue #2530, opened 2023-06-15
Failure on cargo-ui/unstable-attr/invalid regression


karkhaz:
I tried running the regression suite just before releasing on an M1 macOS host:

```
RUST_BACKTRACE=1 ./scripts/kani-regression.sh
```

using tag 0.30.0, i.e. tip of [this PR](https://github.com/model-checking/kani/pull/2529).

One test failed:

```
---- [cargo-kani] cargo-ui/unstable-attr/invalid/expected stdout ----

error: test failed: expected output to contain the line(s):
error: failed to parse `#[kani::unstable]`: missing `feature` field
src/lib.rs
|
| #[kani::unstable(
| ^^^^^^^^^^^^^^^^^^
|
= note: expected format: #[kani::unstable(feature="<IDENTIFIER>", issue="<ISSUE>", reason="<DESCRIPTION>")]
= note: this error originates in the attribute macro `kani::unstable` (in Nightly builds, run with -Z macro-backtrace for more info)
status: exit status: 1
command: cd "/private/tmp/kani-0.30.0/tests/cargo-ui/unstable-attr/invalid" && "cargo" "kani" "--target-dir" "/private/tmp/kani-0.30.0/build/tests/cargo-ui/unstable-attr/invalid/expected/target"
stdout:
------------------------------------------
error: failed to parse `#[kani::unstable]`: expected "key = value" pair, but found `feature`
  --> src/lib.rs:17:1
   |
17 | #[kani::unstable(feature, issue)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: expected format: #[kani::unstable(feature="<IDENTIFIER>", issue="<ISSUE>", reason="<DESCRIPTION>")]
   = note: this error originates in the attribute macro `kani::unstable` (in Nightly builds, run with -Z macro-backtrace for more info)

[...]

PR #2531, opened 2023-06-15
Update expected file for unstable attr test

zhassan-aws:
### Description of changes: 

Apparently, the error message that is emitted on M1 for this test is slightly different than on other platforms.

On M1:
```
error: failed to parse `#[kani::unstable]`: expected "key = value" pair, but found `feature`
```
On other platforms:
```
error: failed to parse `#[kani::unstable]`: missing `feature` field
```
To make the test work on both platforms, I'm updating the `expected` file to remove the suffix that is different.

This is a mitigation for the Kani 0.30.0 release till we figure out why the message is different.

### Resolved issues:

Mitigates #2530 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

### Checklist

### DIFF
--- a/tests/cargo-ui/unstable-attr/invalid/expected
+++ b/tests/cargo-ui/unstable-attr/invalid/expected
@@ -1,4 +1,4 @@
-error: failed to parse `#[kani::unstable]`: missing `feature` field\
+error: failed to parse `#[kani::unstable]`:
 src/lib.rs\
 |\
 | #[kani::unstable(\


celinval (approved):
This is a valid mitigation. So it's better than what we have. Thanks

Btw, I think the problem is the order that it get assessed. The long term solution is to break this test into multiple ones or to print all errors again like in the original PR.

PR #2534, opened 2023-06-15
Build the verification libraries using Kani compiler

### LINKED ISSUE #2533: Build Kani library with the kani-compiler
**Proposed change:** Build Kani and the std library using kani-compiler.

**Motivation:** The Kani compiler has extra checks, such as well-formed attributes, that will be useful when compiling our sysroot. For example, if we add an ill-formed `#[kani::unstable]` attribute today, the user crate compilation will fail (we even add a `debug_assert()` since we expect that to be a Kani bug. Instead, we want the `kani` library compilation to fail.

celinval:
### Description of changes: 

This will ensure we run Kani extra validation checks in our libraries.

### Resolved issues:

Resolves #2533 

### Call-outs:

By default, the Kani compiler will not codegen anything, so this change shouldn't generate Goto. Because of that, this change reduced the size of the verification library *rlib files by almost 60MB

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


tools/build-kani/src/sysroot.rs
```diff
@@ -56,35 +56,42 @@ pub fn kani_playback_lib() -> PathBuf {
...
 /// Build the `lib/` folder and `lib-playback/` for the new sysroot.
 /// - The `lib/` folder contains the sysroot for verification.
 /// - The `lib-playback/` folder contains the sysroot used for playback.
-pub fn build_lib() -> Result<()> {
-    build_verification_lib()?;
-    build_playback_lib()
+pub fn build_lib(bin_folder: &Path) -> Result<()> {
+    let compiler = bin_folder.join("kani-compiler");
+    build_verification_lib(&compiler)?;
+    build_playback_lib(&compiler)
```
adpaco-aws, line 69: Do we discuss this in our developer documentation? If not, I think it'd be nice to add an overview there in the future.

tools/build-kani/src/sysroot.rs
```diff
@@ -56,35 +56,42 @@ pub fn kani_playback_lib() -> PathBuf {
...
 }
 
 /// Build the `lib/` folder for the new sysroot used during verification.
 /// This will include Kani's libraries as well as the standard libraries compiled with --emit-mir.
-fn build_verification_lib() -> Result<()> {
+fn build_verification_lib(compiler: &Path) -> Result<()> {
```
adpaco-aws, line 74: Please rename `compiler` to `compiler_path` or any other name that hints at this being a location, not an object.

tools/build-kani/src/sysroot.rs
```diff
@@ -56,35 +56,42 @@ pub fn kani_playback_lib() -> PathBuf {
...
         ["-Z", "build-std=panic_abort,std,test", "--config", "profile.dev.panic=\"abort\""];
-    build_kani_lib(&kani_sysroot_lib(), &extra_args)
+    let compiler_args = ["--kani-compiler", "-Cllvm-args=--ignore-global-asm"];
+    build_kani_lib(compiler, &kani_sysroot_lib(), &extra_args, &compiler_args)
 }
 
 /// Build the `lib-playback/` folder that will be used during counter example playback.
 /// This will include Kani's libraries compiled with `concrete-playback` feature enabled.
-fn build_playback_lib() -> Result<()> {
+fn build_playback_lib(compiler: &Path) -> Result<()> {
```
adpaco-aws, line 83: Same renaming request here.

tools/build-kani/src/sysroot.rs
```diff
@@ -225,18 +232,20 @@ fn build_artifacts(cargo_cmd: &mut Child) -> Vec<Artifact> {
         .collect()
 }
 
+/// Build Kani binaries with the extra arguments provided and return path to the binaries folder.
```
adpaco-aws, line 235:
```suggestion
/// Build Kani binaries with the extra arguments provided and return the path to the binaries folder.
```

adpaco-aws (changes requested):

tools/build-kani/src/sysroot.rs
celinval: Do you mean discuss the build process?

tools/build-kani/src/sysroot.rs
celinval: Sure

tools/build-kani/src/sysroot.rs
adpaco-aws: Not the build process itself, but the libraries which are built and their purpose.

adpaco-aws (approved):

tools/build-kani/src/sysroot.rs
celinval: Sounds good. I want to eventually document Kani's architecture too.

PR #2542, opened 2023-06-20
Add run_command benchcomp visualization

### LINKED ISSUE #2518: Allow visualization to be invoked as executable

karkhaz:
This allows users to write their own custom visualization script to run after running the benchmarks. Prior to this commit, visualizations had to be checked into the Kani repository.

When `run_command` is specified as a visualization, benchcomp runs the specified command and passes the result of the run as a JSON file on stdin. The command can then process the result however it likes.

This resolves #2518.

### Testing:

* How is this change tested?

2 new regression tests

* Is this a refactor change?

no

### Checklist

### DIFF
--- a/tools/benchcomp/benchcomp/visualizers/__init__.py
+++ b/tools/benchcomp/benchcomp/visualizers/__init__.py
@@ -3,6 +3,8 @@
 
 
 import dataclasses
+import json
+import subprocess
 import textwrap
 
 import jinja2
@@ -12,8 +14,42 @@ import benchcomp
 import benchcomp.visualizers.utils as viz_utils
 
 
-# TODO The doc comment should appear in the help output, which should list all
-# available checks.
+
+@dataclasses.dataclass
+class run_command:
+    """Run an executable command, passing the result as a JSON file on stdin.
+
+    This allows you to write your own visualization, which reads a result file
+    on stdin and does something with it, e.g. writing out a graph or other
+    output file.
+
+    Sample configuration:
+
+    ```
+    visualize:
+    - type: run_command
+      command: ./my_visualization.py
+    ```
+    """
+
+    command: str
+
+
+    def __call__(self, results):
+        results = json.dumps(results, indent=2)
+        try:
+            proc = subprocess.Popen(
+                self.command, shell=True, text=True, stdin=subprocess.PIPE)
+            _, _ = proc.communicate(input=results)
+        except subprocess.CalledProcessError as exc:
+            logging.warning(
+                "visualization command '%s' exited with code %d",
+                self.command, exc.returncode)
+        except (OSError, subprocess.SubprocessError) as exe:
+            logging.error(
+                "visualization command '%s' failed: %s", self.command, str(exe))
+
+
 
 @dataclasses.dataclass
 class error_on_regression:
--- a/tools/benchcomp/benchcomp/visualizers/__init__.py
+++ b/tools/benchcomp/benchcomp/visualizers/__init__.py
@@ -41,13 +41,15 @@ class run_command:
             proc = subprocess.Popen(
                 self.command, shell=True, text=True, stdin=subprocess.PIPE)
             _, _ = proc.communicate(input=results)
-        except subprocess.CalledProcessError as exc:
-            logging.warning(
-                "visualization command '%s' exited with code %d",
-                self.command, exc.returncode)
         except (OSError, subprocess.SubprocessError) as exe:
             logging.error(
                 "visualization command '%s' failed: %s", self.command, str(exe))
+            viz_utils.EXIT_CODE = 1
+        if proc.returncode:
+            logging.error(
+                "visualization command '%s' exited with code %d",
+                self.command, proc.returncode)
+            viz_utils.EXIT_CODE = 1
 
 
 
--- a/tools/benchcomp/benchcomp/visualizers/__init__.py
+++ b/tools/benchcomp/benchcomp/visualizers/__init__.py
@@ -17,7 +17,7 @@ import benchcomp.visualizers.utils as viz_utils
 
 @dataclasses.dataclass
 class run_command:
-    """Run an executable command, passing the result as a JSON file on stdin.
+    """Run an executable command, passing the performance metrics as JSON on stdin.
 
     This allows you to write your own visualization, which reads a result file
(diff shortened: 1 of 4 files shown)


karkhaz:
(Note that the doc text for `run_command` will automatically get rendered into the [visualizations section of the Kani book](https://model-checking.github.io/kani/benchcomp-conf.html)

celinval:
> (Note that the doc text for `run_command` will automatically get rendered into the [visualizations section of the Kani book](https://model-checking.github.io/kani/benchcomp-conf.html).)

That's a neat idea. Do you have an example from your repo on how that would look like?

tools/benchcomp/benchcomp/visualizers/__init__.py
```diff
@@ -12,8 +14,42 @@
...
 
 
-# TODO The doc comment should appear in the help output, which should list all
-# available checks.
+
+@dataclasses.dataclass
+class run_command:
+    """Run an executable command, passing the result as a JSON file on stdin.
+
+    This allows you to write your own visualization, which reads a result file
```
celinval, line 22: Where can one find the format of the result file?

I see the following sentence on [this page ](https://model-checking.github.io/kani/benchcomp-conf.html), but I don't see any schema.

> This page describes the file's schema and lists the different parsers and visualizations that are available.

tools/benchcomp/test/test_regression.py
```diff
@@ -737,3 +738,105 @@ def test_command_parser(self):
...
+
+            with open(out_file) as handle:
+                result = yaml.safe_load(handle)
+
+            for item in ["benchmarks", "metrics"]:
+                self.assertIn(item, result)
+
+
+    def test_run_failing_command_visualization(self):
+        """Ensure that benchcomp terminates normally even when run_command visualization doesn't"""
```
celinval, line 795: Does this mean that benchcomp will still exit with exit status of 0 when a visualization fails?

karkhaz:
> That's a neat idea. Do you have an example from your repo on how that would look like?

In [this file](https://github.com/model-checking/kani/blob/main/tools/benchcomp/benchcomp/visualizers/__init__.py#L18) (the same file as in this PR, you can also just expand the diff downwards) there are three classes. Each of them has a doc text, and those three doc texts are what get rendered into the [three visualization documentations in the kani book](https://model-checking.github.io/kani/benchcomp-conf.html)

tools/benchcomp/benchcomp/visualizers/__init__.py
karkhaz: Yeah, not yet merged, sorry :( I have a half-written schema and was going to generate the documentation from it, similar to how the visualization docs are generated.

For now if you'd like to use this feature, I suggest starting off by setting
```yaml
type: run_command
command: cat - > /tmp/result.yaml
```
which just takes stdin and redirects it to a file, so you can then open and read that file to see the structure.

tools/benchcomp/test/test_regression.py
karkhaz: When this particular visualization fails, yes. (There are other visualizations that behave differently, for example `error_on_regression` terminates benchcomp with a return code of 1 if any benchmark regressed).

I don't have a strong opinion about this, I can make benchcomp crash out if this visualization fails, but I think it's sort of a shame to. The visualizations currently run one after another but are supposed to be independent.

tools/benchcomp/benchcomp/visualizers/__init__.py
celinval: Can you please remove the statement I pasted until then?

tools/benchcomp/test/test_regression.py
celinval: I would expect that benchcomp exit status reflects whether it succeeds to run all visualizations. You could implement a delayed error approach since there is no dependency on the visualizations.

The `error_on_regression` is not really a visualization, but I don't think we should revisit this. As the name says, the visualization fails if it finds a regression, so I don't see how its behavior is special.

celinval:
> > That's a neat idea. Do you have an example from your repo on how that would look like?
> 
> In [this file](https://github.com/model-checking/kani/blob/main/tools/benchcomp/benchcomp/visualizers/__init__.py#L18) (the same file as in this PR, you can also just expand the diff downwards) there are three classes. Each of them has a doc text, and those three doc texts are what get rendered into the [three visualization documentations in the kani book](https://model-checking.github.io/kani/benchcomp-conf.html)

I was wondering if you have the link to the rendered version that includes your PR changes. You can enable the documentation to be built in your fork, so it is easy to review documentation changes. For example, here is the link for my fork's documentation: https://celinval.github.io/kani-dev/

The only thing is that those changes have to be made to your forks main branch.

tools/benchcomp/benchcomp/visualizers/__init__.py
```diff
@@ -12,8 +14,42 @@
 import benchcomp.visualizers.utils as viz_utils
 
 
-# TODO The doc comment should appear in the help output, which should list all
-# available checks.
+
+@dataclasses.dataclass
+class run_command:
+    """Run an executable command, passing the result as a JSON file on stdin.
```
qinheping, line 20:
```suggestion
    """Run an executable command, passing the performance metrics as JSON on stdin.
```

tools/benchcomp/benchcomp/visualizers/__init__.py
karkhaz: Done

tools/benchcomp/test/test_regression.py
karkhaz: Done, now benchcomp terminates with 1 if `run_command` exited with non-zero. It doesn't exit straight away, other visualizations get a chance to run, but the return code at the end gets set to 1.

karkhaz: > The `error_on_regression` is not really a visualization, but I don't think we should revisit this. As the name says, the visualization fails if it finds a regression, so I don't see how its behavior is special.

I don't mind renaming it e.g. `post-process` or anything else if you have any suggestions

karkhaz:
> You can enable the documentation to be built in your fork

Awesome, thank you! I'll do that from now on. Here's how it looks:

https://karkhaz.github.io/kani/benchcomp-conf.html#run_command

celinval (approved):

PR #2555, opened 2023-06-22
Support for requires and ensures clauses

JustusAdam:
### Description of changes: 

Adds basic support for `requires` and `ensures` function contract macros.

### Testing:

Some simple test cases have been added

### Checklist


celinval:
Can you please make sure the PR only include changes relevant to contract?

JustusAdam:
Yeah that was because the target branch was out of date.

celinval (commented):
That's great progress! As talked offline, the focus at this point is the CBMC integration. Here is a few notes:

1. I think the bindings code look quite ready. :)
2. Is there any reason why there is no tests for `requires`?
3. For future PRs, I think it would be better if we make the compiler code more modular. We should also figure out what the user flow will be for contracts, so we can design the communication between each component.

JustusAdam:
> 2. Is there any reason why there is no tests for `requires`?
> 3. For future PRs, I think it would be better if we make the compiler code more modular. We should also figure out what the user flow will be for contracts, so we can design the communication between each component.

So they are actually related. I didn't add a test case for `requires` because I didn't implement substitution. But I realize that I can also add a test where the verification would fail if the precondition is violated (e.g. the `assume` did not take hold) so I'll do that.

And wrt. flow we've discussed that in person already but for the benefit of the log: The flow is in the design phase and will be added after history variables and quantifiers.

library/kani_macros/src/lib.rs
```diff
@@ -89,17 +89,67 @@ pub fn derive_arbitrary(item: TokenStream) -> TokenStream {
     derive::expand_derive_arbitrary(item)
 }
 
+/// Add a precondition to this function.
+///
+/// This is part of the function contract API, together with [`ensures`].
+///
+/// The contents of the attribute is a condtition over the input values to the
```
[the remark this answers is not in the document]
JustusAdam: Well for (safe rust) constants it shouldn't matter since they can't change so they either satisfy the conditions always or never.

Static's yes, but I haven't investigated yet if they work with CBMC yet and I'm focusing on safe rust with wouldn't allow that anyway.

library/kani_macros/src/lib.rs
```diff
@@ -89,17 +89,67 @@ pub fn derive_arbitrary(item: TokenStream) -> TokenStream {
...
+    attr_impl::requires(attr, item)
+}
+
+/// Add a postcondition to this function.
+///
+/// This is part of the function contract API, together with [`requires`].
+///
+/// The contents of the attribute is a condtition over the input values to the
+/// annotated function *and* its return value, accessible as a variable called
+/// `result`. All Rust syntax is supported, even calling other functions, but
```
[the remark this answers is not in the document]
JustusAdam: That's TODO. If you think that's important to fix now I can do that.

library/kani_macros/src/lib.rs
```diff
@@ -184,6 +234,110 @@ mod sysroot {
...
+    ///
+    /// ```rs
+    /// #[kani::ensures(x < result)]
+    /// fn foo(x: u32) -> u32 { .. }
+    /// ```
+    ///
+    /// Is rewritten to
+    ///
+    /// ```rs
+    /// fn foo_enusres_<hash of foo>(x: u32, result: u32) {
```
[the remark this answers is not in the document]
JustusAdam: I checked this indirectly in the `vec-map` example, which uses multiple `ensures` on a single function and it does not complain about duplicate functions so I can only assume that the hash changes. (would make sense too, since the `ItemFn` contains the attributes of the function)

library/kani_macros/src/lib.rs
```diff
@@ -184,6 +234,110 @@ mod sysroot {
...
+    ///     x < result
+    /// }
+    ///
+    /// #[kanitook::ensures = "foo_ensures_<hash of foo>"]
+    /// fn foo(x: u32) -> u32 { .. }
+    /// ```
+    ///
+    /// Note that CBMC's contracts always pass the return value (even for
```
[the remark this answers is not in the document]
JustusAdam: Hmm, that would require creating a different signature for the generated functions for `requires` and `ensures`. I can do that but it'll make the code more complex.

JustusAdam:
I need to mention here that this does a small change to the testing infrastructure as well. In particular a directory can now contain multiple "expected" tests. The setup is analogous to regular `.rs` tests but each also has an associated `.expected` file. The way that is implemented is that when looking for test in a directory and we find a candidate file `my_file.rs` then we first look for a file called `my_file.expected` as the expected file. If such a file exists then it will be used as the expected file, otherwise we look for a `expected` file.

library/kani_macros/src/lib.rs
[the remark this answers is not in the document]
celinval: Where is the TODO? Do you have an issue that tracks that?

library/kani_macros/src/lib.rs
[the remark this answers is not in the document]
celinval: Can you please add a second ensures to one of your tests?

library/kani_macros/src/lib.rs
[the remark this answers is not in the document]
celinval: I guess we can live with that for now since it is transparent from the user point of view.

tests/kani/FunctionContract/simple_ensures_pass.rs
```diff
@@ -0,0 +1,16 @@
...
+        x
+    } else {
+        y
+    }
+}
+
+#[kani::proof]
+fn main() {
+    max(7, 6);
+}
```
feliperodri: new line

tests/kani/FunctionContract/simple_ensures_fail.rs
```diff
@@ -0,0 +1,17 @@
...
+        x
+    } else {
+        y
+    }
+}
+
+#[kani::proof]
+fn main() {
+    max(7, 9);
+}
```
feliperodri: new line

tests/kani/FunctionContract/arbitrary_ensures_fail.rs
```diff
@@ -0,0 +1,17 @@
...
+        x
+    } else {
+        y
+    }
+}
+
+#[kani::proof]
+fn main() {
+    max(kani::any(), kani::any());
+}
```
feliperodri: new line

tests/kani/FunctionContract/arbitrary_ensures_pass.rs
```diff
@@ -0,0 +1,16 @@
...
+        x
+    } else {
+        y
+    }
+}
+
+#[kani::proof]
+fn main() {
+    max(kani::any(), kani::any());
+}
```
feliperodri: new line

library/kani_macros/src/lib.rs
```diff
@@ -89,17 +89,48 @@ pub fn derive_arbitrary(item: TokenStream) -> TokenStream {
...
 
+#[proc_macro_attribute]
+pub fn requires(attr: TokenStream, item: TokenStream) -> TokenStream {
+    attr_impl::requires(attr, item)
+}
+
+#[proc_macro_attribute]
+pub fn ensures(attr: TokenStream, item: TokenStream) -> TokenStream {
+    attr_impl::ensures(attr, item)
+}
```
feliperodri: What about assigns? We might don't need it yet, so at least create a GitHub issue to track that we need to add it here.

kani_metadata/src/harness.rs
```diff
@@ -24,6 +24,8 @@ pub struct HarnessMetadata {
     pub goto_file: Option<PathBuf>,
     /// The `#[kani::<>]` attributes added to a harness.
     pub attributes: HarnessAttributes,
+
+    pub contracts: Vec<String>,
```
feliperodri, line 28: Add a comment above as documentation.

kani-compiler/src/kani_middle/contracts.rs
```diff
@@ -0,0 +1,36 @@
+use rustc_hir::def_id::DefId;
```
feliperodri, line 4: header?

kani-compiler/src/kani_middle/contracts.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[derive(Default)]
+pub struct GFnContract<C> {
+    requires: Vec<C>,
+    ensures: Vec<C>,
+    assigns: Vec<C>,
+}
```
feliperodri, line 14: documentation?

kani-compiler/src/kani_middle/contracts.rs
```diff
@@ -0,0 +1,36 @@
...
+        GFnContract {
+            requires: self.requires.iter().map(&mut f).collect(),
+            ensures: self.ensures.iter().map(&mut f).collect(),
+            assigns: self.assigns.iter().map(&mut f).collect(),
+        }
+    }
+
+    pub fn enforceable(&self) -> bool {
+        !self.requires().is_empty() || !self.ensures().is_empty()
+    }
```
feliperodri, line 46: Do we use this at all now?

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -38,7 +46,9 @@ impl KaniAttributeKind {
             | KaniAttributeKind::Solver
             | KaniAttributeKind::Stub
             | KaniAttributeKind::Unwind => true,
-            KaniAttributeKind::Unstable => false,
+            KaniAttributeKind::Unstable
+            | KaniAttributeKind::Ensures
+            | KaniAttributeKind::Requires => false,
```
feliperodri, line 51: The default value is `false`? In CBMC, it's `true` (see https://diffblue.github.io/cbmc/contracts-requires-ensures.html).

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -138,11 +149,68 @@ pub fn extract_harness_attributes(tcx: TyCtxt, def_id: DefId) -> HarnessAttribut
                 // Internal attribute which shouldn't exist here.
                 unreachable!()
             }
+            KaniAttributeKind::Ensures | KaniAttributeKind::Requires => {
+                todo!("Contract attributes are not supported on proofs (yet)")
```
feliperodri, line 153: I don't understand what that means for the user. We may need a better message.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -138,11 +149,68 @@ pub fn extract_harness_attributes(tcx: TyCtxt, def_id: DefId) -> HarnessAttribut
...
+            },
+            Node::Crate(m) => find_in_mod(m),
+            other => panic!("Odd prant node type {other:?}"),
+        }
+        .expect_owner()
+        .def_id
+        .to_def_id()
+    };
+
+    //println!("Searching in {:?}", hir_map.module_items(enclosing_mod).map(|it| hir_map.item(it).ident.name).collect::<Vec<_>>());
```
feliperodri, line 193: debug left over?

cprover_bindings/src/irep/irep_id.rs
```diff
@@ -1462,6 +1463,7 @@ impl ToString for IrepId {
             IrepId::CSpecLoopInvariant => "#spec_loop_invariant",
             IrepId::CSpecRequires => "#spec_requires",
             IrepId::CSpecEnsures => "#spec_ensures",
+            IrepId::CSpecAssigns => "#spec_assigns",
```
feliperodri, line 1466: I'd remove everything related to assigns clauses from this pull request, which is focus only on "simple requires and ensures".

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -222,6 +223,90 @@ impl<'tcx> GotocCtx<'tcx> {
         );
     }
 
+    /// Convert the Kani level contract into a CBMC level contract by creating a
```
feliperodri, line 226:
```suggestion
    /// Convert the Kani-level contract into a CBMC-level contract by creating a
```

feliperodri (changes requested):

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -710,8 +710,8 @@ impl<'tcx> GotocCtx<'tcx> {
             trace!(func=?instance, "codegen_func_symbol");
             let func = self.symbol_name(instance);
             self.symbol_table
-                .lookup(func)
-                .expect("Function `{func}` should've been declared before usage")
+                .lookup(&func)
+                .unwrap_or_else(|| panic!("Function `{func}` should've been declared before usage"))
```
celinval, line 714: Thanks

kani-compiler/src/kani_middle/attributes.rs
celinval: +1... Please add a user friendly message with `sess.error()`. Also, do we even expect contracts to be added to proof harnesses? This message is setting an expectation that hasn't been discussed.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -91,6 +101,7 @@ pub(super) fn check_attributes(tcx: TyCtxt, def_id: DefId) {
             KaniAttributeKind::Unstable => attrs.iter().for_each(|attr| {
                 let _ = UnstableAttribute::try_from(*attr).map_err(|err| err.report(tcx));
             }),
+            KaniAttributeKind::Ensures | KaniAttributeKind::Requires => (),
```
celinval, line 104: The validation logic should go here, not inside `extract_harness_attributes`.

kani-compiler/src/kani_middle/reachability.rs
```diff
@@ -172,8 +177,32 @@ impl<'tcx> MonoItemsCollector<'tcx> {
     /// instruction looking for the items that should be included in the compilation.
     fn reachable_items(&mut self) {
         while let Some(to_visit) = self.queue.pop() {
-            if !self.collected.contains(&to_visit) {
-                self.collected.insert(to_visit);
+            if let Entry::Vacant(e) = self.collected.entry(to_visit) {
```
celinval, line 180: I think we should structure this code a bit better. Please create an issue and add a TODO to do that. Also, for now, can you at least move this code a separate function?

kani-driver/src/args/common.rs
```diff
@@ -42,6 +42,8 @@ pub enum UnstableFeatures {
     CFfi,
     /// Enable concrete playback flow.
     ConcretePlayback,
+    /// Enable function contracts
+    Contracts,
```
celinval, line 46: I don't think you are really using this. Please create an issue to do so and add a TODO.

kani-driver/src/call_goto_instrument.rs
```diff
@@ -160,6 +164,17 @@ impl KaniSession {
         self.call_goto_instrument(args)
     }
 
+    /// Make CBMC enforce a function contract.
+    pub fn enforce_contract(&self, file: &Path, function: &str) -> Result<()> {
+        println!("enforcing {function} contract");
```
celinval, line 169:
```suggestion
        tracing::debug!(?function, "enforce_contract");
```

kani-driver/src/call_goto_instrument.rs
```diff
@@ -37,6 +37,10 @@ impl KaniSession {
             self.goto_sanity_check(output)?;
         }
 
+        for function in &harness.contracts {
```
celinval, line 40: Please add a test (or create an issue) where multiple contracts are reachable from the same harness

kani_metadata/src/harness.rs
celinval, line 28: Can you please rename this to `reachable_contracts` or something similar? I think the name is a bit misleading. As I mentioned offline, we might want to add to the harness which contract to verify.

celinval (commented):
I really think we need to define the UX before most of this code can be pushed to main. But for the feature branch, I think this is in a fairly good shape.

For the comments that reflect code quality / organization, for this PR, I think it's OK if you create an issue to capture the work and add a `TODO` statement with a description and a link to the new issue.

library/kani_macros/src/lib.rs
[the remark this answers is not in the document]
JustusAdam: Because of `old ` the two will end up diverging anyway. I can add the divergence here or it will be merged in once `old` gets merged.

tests/kani/FunctionContract/simple_ensures_pass.rs
JustusAdam: You mean after the function?

kani-compiler/src/kani_middle/attributes.rs
JustusAdam: This is not the default value, this says whether this attribute can only occur on a harness (which is false for contracts).

library/kani_macros/src/lib.rs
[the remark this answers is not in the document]
celinval: Maybe let's not add more stuff in this PR.

tests/kani/FunctionContract/simple_ensures_pass.rs
celinval: He means at the end of the file. It's a good practice to add an empty new line at the end of each file.

celinval: btw, you can configure vscode to do that for you: https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files

library/kani_macros/src/lib.rs
JustusAdam: Has been created https://github.com/model-checking/kani/issues/2594

kani-compiler/src/kani_middle/attributes.rs
JustusAdam: I'm delaying doing this until the attributes are stable(ish) #2595

kani-driver/src/call_goto_instrument.rs
JustusAdam: I've removed this capability, because CBMC can't deal with this anyway

kani-compiler/src/kani_middle/contracts.rs
JustusAdam: Yes.

feliperodri (approved):
LGTM.

feliperodri:
@celinval any idea why [Cargo Audit](https://github.com/model-checking/kani/actions/runs/5512451712/jobs/10049298801?pr=2555) is failing here?

adpaco-aws:
I think you just need to include the commit from https://github.com/model-checking/kani/pull/2581 in the source branch.

JustusAdam:
I did another rebase to try and address the audit issue. Tbh I should have merged instead but oh well. I am going to squash the commits before merging anyway at least that's the plan.

JustusAdam:
I'll bump the base branch next so the PR doesn't include the unrelated commits

library/kani_macros/src/lib.rs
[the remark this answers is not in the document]
JustusAdam: Done

JustusAdam:
I'm having trouble with this regression stuff now. The numbering for checks seems to be platform dependent? When I run on my machine the `expected` test suite completes fine but on the ubuntu CI it doesn't. What's the best practice here to make the expected output stable? Maybe @feliperodri can help?

zhassan-aws:
We don't include the check indices in the expected files to avoid those issues. Even if they're consistent across platforms, they may change with changes to the compiler or new CBMC versions. Note that compiletest checks for a substring and not an exact match for the line.

JustusAdam:
Does it require ordering? Because one thing I now worry about is that the expected success/failure may not be the one with the description I am looking for. E.g. if i expected one property to succeed and another to fail what guarantees me that "Status: FAILURE" actually belongs to my "Description: fail this" test and not my "Description: succeed this"?

zhassan-aws:
> Does it require ordering?

Generally, no: the compiletest tool checks if each line in the `expected` file occurred in the output.

> the expected success/failure may not be the one with the description I am looking for

If a line in the `expected` file ends in backslash, the following line must occur on the consecutive line in the output. This is how we achieve ordering (partially). See https://github.com/model-checking/kani/blob/main/tests/expected/assert-eq/expected for an example.

kani_metadata/src/harness.rs
JustusAdam: Done

JustusAdam:
I've fixed up the test cases but unfortunately benchcomp is still failing. Is that a problem?

qinheping:
> 

There are some interesting founds in the Benchcomp CI.
1. The running time of format/fmt_i8 in the old variant is ~2.62 seconds. However, it the past Benchcomp run, for example https://github.com/model-checking/kani/actions/runs/5558854620/job/15059090659#step:13:71, the running time of format/fmt_i8 is ~1000seconds. 
2. Also, the `number_program_steps` of the old variant of format/fmt_i8 and format_fmt/8 are much smaller than the `number_program_steps` of them in the table https://github.com/model-checking/kani/actions/runs/5627831808

qinheping:
After comparing the Benchomp run between these two commits https://github.com/model-checking/kani/actions/runs/5405839079/job/14637506427 and https://github.com/model-checking/kani/actions/runs/5459160729/job/14779199972.
I confirmed that the regression of all 6 regressed benchmarks
`btreeset/insert_any/main` `btreeset/insert_same/main` `format/fmt_i8`  `s2n-quic/quic/s2n-quic-core/sync::cursor::tests::oracle_test`
`s2n-quic/quic/s2n-quic-core/sync::spsc::tests::alloc_test`
`s2n-quic/tools/xdp/s2n-quic-xdp/task::completion_to_tx::assign::tests::assignment_test`
is due to the huge toolchain upgrade https://github.com/model-checking/kani/commit/713906371b45905ff44f3bc44d919755f61cda11.

PR #2556, opened 2023-06-22
Fix MIR dump to emit one MIR per-harness

celinval:
### Description of changes: 

This was a regression introduced by #2439. We were still writing the result of the reachability algorithm to the same file for every harness. Thus, we could only see the MIR for the last harness that was processed.

Use the file name that is specific for the harness instead and generate one MIR file per harness like we do with other files generated by kani-compiler.

### Resolved issues:

N/A

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? No test. This is only used for debug.

* Is this a refactor change?

### Checklist


jaisnan (approved):
lgtm!

PR #2565, opened 2023-06-23
Create a Boogie AST crate

zhassan-aws:
### Description of changes: 

Introduce a Boogie AST crate and translate a small subset of MIR to Boogie. The current translation can handle simple programs involving integers and booleans and asserts (using `kani::assert`), e.g.:
```rust
#[kani::proof]
fn main() {
    let x = true;
    kani::assert(x, "");
}
```
and
```rust
#[kani::proof]
fn main() {
    let x = 1;
    let y = x;
    kani::assert(y == 1, "");
}
```

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

I've been testing the output by running it with `boogie` manually. The next step would be to install boogie as part of the setup process and run it on the output file via a script.

### Testing:

* How is this change tested? Testing manually for now

### Checklist


boogie_ast/src/boogie_program/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    Label { label: String },
+
+    /// Return statement: `return;`
+    Return,
+
+    /// While statement: `while (condition) { body }`
+    While { condition: Expr, body: Box<Stmt> },
+}
+
+/// Procedure specification
```
celinval, line 152: Contract specification

boogie_ast/src/boogie_program/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// Pre-conditions
+    requires: Vec<Expr>,
+    /// Post-conditions
+    ensures: Vec<Expr>,
+    /// Modifies clauses
+    // TODO: should be symbols and not expressions
+    modifies: Vec<Expr>,
+}
+
+/// Procedure definition
```
celinval, line 163: Can you please document what a procedure is in Boogie?

boogie_ast/src/boogie_program/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+        return_type: Vec<(String, Type)>,
+        contract: Option<Contract>,
+        body: Stmt,
+    ) -> Self {
+        Procedure { name, parameters, return_type, contract, body }
+    }
+}
+
+/// Function definition
+struct Function {}
```
celinval, line 185: Please document what a Function is.

boogie_ast/src/boogie_program/writer.rs
```diff
@@ -0,0 +1,320 @@
...
+        match self {
+            UnaryOp::Not => write!(writer.writer, "!")?,
+            UnaryOp::Neg => write!(writer.writer, "-")?,
+        }
+        Ok(())
+    }
+}
+
+impl Writable for BinaryOp {
+    fn write_to<T: Write>(&self, writer: &mut Writer<T>) -> std::io::Result<()> {
```
celinval, line 302: Wouldn't it make sense to implement Display instead?

kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
```diff
@@ -13,14 +29,303 @@ pub struct BoogieCtx<'tcx> {
...
+            | StatementKind::PlaceMention(..)
+            | StatementKind::AscribeUserType(..)
+            | StatementKind::Coverage(..)
+            | StatementKind::Intrinsic(..)
+            | StatementKind::ConstEvalCounter
+            | StatementKind::Nop => todo!(),
+        }
+    }
+
+    pub fn codegen_rvalue(&self, rvalue: &Rvalue<'tcx>) -> Expr {
```
celinval, line 149: I wonder if we should follow this pattern of returning expression in these functions. In the `goto` translation, we constantly add checks that also generate statements. Does Boogie have anything similar to `goto`'s `StatementExpression`?

Another alternative is to ensure the checks are added before codegen. We could either add these checks as MIR transformation pass or fix https://github.com/model-checking/kani/issues/1740.

kani-compiler/src/codegen_boogie/overrides/hooks.rs
```diff
@@ -0,0 +1,124 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This module contains various codegen hooks for functions that require special handling.
+//!
+//! E.g.: Functions in the Kani library that generate assumptions or symbolic variables.
+//!
+//! It would be too nasty if we spread around these sort of undocumented hooks in place, so
```
celinval, line 8: This comment is a bit odd. I know we have that in the `goto`, but maybe it doesn't need to be replicated.

kani-compiler/src/codegen_boogie/overrides/hooks.rs
```diff
@@ -0,0 +1,124 @@
...
+        Stmt::Block {
+            statements: vec![
+                Stmt::Assert { condition: cond },
+                // handle target
+            ],
+        }
+    }
+}
+
+pub fn fn_hooks<'tcx>() -> BoogieHooks<'tcx> {
```
celinval, line 103: These hooks would probably benefit from a better interface, but not a big deal for this initial implementation. I think the problem with our current hooks is that it does two different things, it implements engine intrinsic and also implements some sort of abstraction.

celinval (commented):
Thanks for working on this. When do you think you'll be able to add a few tests?

boogie_ast/src/boogie_program/mod.rs
```diff
@@ -0,0 +1,279 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
```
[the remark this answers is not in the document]
zhassan-aws: Done

boogie_ast/src/boogie_program/mod.rs
zhassan-aws: Done

boogie_ast/src/boogie_program/mod.rs
zhassan-aws: Done

kani-compiler/src/codegen_boogie/overrides/hooks.rs
zhassan-aws: Agreed. It would be useful to be able to check for abstracted functions independent of the backend as well (e.g. to use in reachability analysis).

zhassan-aws (commented):
> When do you think you'll be able to add a few tests?

I've been testing this manually on a few small programs via running Boogie on them locally.

I looked into adding a script that installs Boogie, but found out that it'll take some effort because of the need to install .NET, so I decided to push this to a future PR.

It should be possible to add tests for the writer though, which I can do in this PR or a follow-up one.

boogie_ast/src/boogie_program/writer.rs
```diff
@@ -0,0 +1,320 @@
...
+impl BoogieProgram {
+    pub fn write_to<T: Write>(&self, writer: &mut T) -> std::io::Result<()> {
+        let mut writer = Writer::new(writer);
+        writer.write(self)
+    }
+}
+
+impl Writable for BoogieProgram {
+    fn write_to<T: Write>(&self, writer: &mut Writer<T>) -> std::io::Result<()> {
+        for _td in &self.type_declarations {}
```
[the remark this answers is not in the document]
zhassan-aws: Done

boogie_ast/src/boogie_program/mod.rs
zhassan-aws: Good catch. Fixed.

kani-compiler/src/codegen_boogie/overrides/hooks.rs
zhassan-aws: Removed.

boogie_ast/src/boogie_program/mod.rs
celinval: Thanks!

boogie_ast/src/boogie_program/writer.rs
```diff
@@ -0,0 +1,320 @@
...
+        self.indentation -= 2;
+    }
+
+    fn indent(&mut self) -> std::io::Result<()> {
+        write!(self.writer, "{:width$}", "", width = self.indentation)
+    }
+}
+
+impl BoogieProgram {
+    pub fn write_to<T: Write>(&self, writer: &mut T) -> std::io::Result<()> {
```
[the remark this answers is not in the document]
zhassan-aws: My intention was to make sure all types implement the `Writable` trait, but I agree this turned out to be too complicated and messy. I simplified the code and removed the `Writable` trait altogether. Hopefully things are clearer now.

boogie_ast/src/boogie_program/writer.rs
```diff
@@ -0,0 +1,320 @@
...
+            writer.write(p)?;
+        }
+        Ok(())
+    }
+}
+
+impl Writable for Procedure {
+    fn write_to<T: Write>(&self, writer: &mut Writer<T>) -> std::io::Result<()> {
+        // signature
+        write!(writer.writer, "procedure {}(", self.name)?;
```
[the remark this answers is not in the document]
zhassan-aws: I removed the `writer.write` one, so things should hopefully be simpler now.

boogie_ast/src/boogie_program/writer.rs
zhassan-aws: Are you referring to implementing it for `BinaryOp` specifically, or for everything? For `BinaryOp`, I think it makes sense to implement it since it's a few characters long, and it could help with debugging.

kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
zhassan-aws: I don't think Boogie has something similar. Adding the checks as a MIR transformation pass would be ideal to make sure there are no discrepancies between backends.

boogie_ast/src/boogie_program/writer.rs
celinval: I think everything

kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
celinval: Ok... if you don't want to rely on the MIR transformation, you might want to change this signature to return statement instead.

boogie_ast/src/boogie_program/writer.rs
```diff
@@ -0,0 +1,474 @@
...
+        write!(self.writer, "{:width$}", "", width = self.indentation)
+    }
+}
+
+impl BoogieProgram {
+    pub fn write_to<T: Write>(&self, writer: &mut T) -> std::io::Result<()> {
+        let mut writer = Writer::new(writer);
+
+        if !self.type_declarations.is_empty() {
+            writeln!(writer.writer, "// Type declarations:")?;
```
celinval, line 86: nit: you could make implement Write for Writer, and avoid a bunch of these `writer.writer`

kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
```diff
@@ -13,14 +29,320 @@ pub struct BoogieCtx<'tcx> {
...
+        Some(Procedure::new(
+            self.tcx.symbol_name(instance).name.to_string(),
+            vec![],
+            vec![],
+            None,
+            Stmt::Block { statements: decl },
+        ))
+    }
+
+    pub fn codegen_declare_variables(&self, instance: Instance<'tcx>) -> Vec<Stmt> {
```
celinval, line 62:
```suggestion
    fn codegen_declare_variables(&self, instance: Instance<'tcx>) -> Vec<Stmt> {
```

kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
```diff
@@ -13,14 +29,320 @@ pub struct BoogieCtx<'tcx> {
...
+    ) -> Vec<Expr> {
+        debug!(?args, "codegen_funcall_args");
+        args.iter()
+            .filter_map(|o| {
+                let ty = self.operand_ty(local_decls, o);
+                // TODO: handle non-primitive types
+                if ty.is_primitive() {
+                    return Some(self.codegen_operand(o));
+                }
+                None
```
celinval, line 234: maybe add a `todo!()`

kani-compiler/src/codegen_boogie/overrides/hooks.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use crate::codegen_boogie::BoogieCtx;
+use boogie_ast::boogie_program::{Expr, Stmt};
+use rustc_middle::mir::{BasicBlock, Place};
+use rustc_middle::ty::{Instance, TyCtxt};
+use rustc_span::Span;
+use std::rc::Rc;
+use tracing::debug;
+
+pub trait BoogieHook<'tcx> {
```
celinval, line 17: 😕 You might want to simplify this a bit later. One would hope that we don't replicate the hooks logic. Our hooks currently implement two different things: kani intrinsics (`kani::assert`, `kani::assume`, `kani::cover`) and function abstraction.

I would suggest that you handle them separately (as we should eventually in the GotoC too). In your current implementation, you only care about intrinsics, so you can only implement the first for now. Hopefully we can rely on stubbing for the second, so you won't need to add anything extra in the backend for that.

For intrinsics, you should basically just check if the function has any diagnostic item, and do a table lookup to see if that is an intrinsic.

celinval (approved):
Mostly some code cleaning suggestions, but not blockers.

boogie_ast/src/boogie_program/writer.rs
zhassan-aws: The reason I avoided using `Display` is that I wanted to maximize the buffering of the output and minimize the creation of intermediate strings that represent components of the Boogie program. The idea is that since everything gets written to a file, we can pass the file buffer to every `write_to` method to allow them to write to it directly as opposed to having to create intermediate strings (which can be large), and then writing them to the file. For instance, if we were to implement `Display` for the `Contract` type, then `Procedure`'s `write_to` method will do:
```rust
write!(writer, "{contract}");
```
which will convert the whole contract into a string, and then write it to the buffer. Whereas, the current approach:
```rust
contract.write_to(writer);
```
allows the components of the contract to be written directly to the buffer without creating an intermediate string that represents the whole contract.

Does this make sense?

kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
zhassan-aws: I gave this some thought, and I'm not clear on how to make the function return a statement given that the rvalue can be in the context of an assignment, a function call, etc.. I think we should keep the current signature that returns an expression, and we can implement `StatementExpression`-like expressions in Boogie using function calls (via inlined procedures). This would allow us to implement any checks in that function, and then return the expression for the rvalue at the end of the function. What do you think?

celinval: You would need to pass the place that would be assigned to and generate the assignment here. Both would work.

celinval: One third option is to return both optional statements and an expr.

zhassan-aws: I went with the third option.

boogie_ast/src/boogie_program/writer.rs
zhassan-aws: Good idea! Done.

kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
zhassan-aws: Removed `pub` from this function and a bunch of other functions as well.

kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
zhassan-aws: Unfortunately, I cannot add a `todo!` here because it'll break the few tests that currently work. This is because they all involve `kani::assert` which has a string slice argument that will result in a crash. I added a `TODO` comment though.

kani-compiler/src/codegen_boogie/overrides/hooks.rs
zhassan-aws: Makes sense. I removed the hooks module, and created a new module for handling Kani intrinsics. I placed it for now under the `codegen_boogie` module, but we should move it up to `kani_middle` at some point.

PR #2571, opened 2023-06-27
RFC: Print Kani version

### LINKED ISSUE #2570: Output - Kani version
Requested feature: Print the version of Kani at the beginning of a run.
Use case: Kani doesn't print its version despite it printing other versions (for example, CBMC's version gets printed once for each harness). This will provide a better experience for both users and developers.

adpaco-aws:
### Description of changes: 

This RFC proposes printing the Kani version at the beginning of a run.

Rendered version available [here](https://github.com/adpaco-aws/rmc/blob/rfc-kani-version/rfc/src/rfcs/0008-kani-version.md).

### Resolved issues:

Resolves #2570 

### Testing:

### Checklist

### DIFF
--- a/rfc/src/rfcs/0008-kani-version.md
+++ b/rfc/src/rfcs/0008-kani-version.md
@@ -1,0 +1,106 @@
+- **Feature Name:** Print Kani version (`kani-version`)*
+- **Feature Request Issue:** <https://github.com/model-checking/kani/issues/2570>
+- **RFC PR:** <https://github.com/model-checking/kani/issues/2571>
+- **Status:** Under Review
+- **Version:** 0
+
+-------------------
+
+## Summary
+
+Print the version of Kani at the beginning of a run.
+
+## User Impact
+
+Many programs print their version at the beginning of a run.
+The version of a program communicates the state of the software at a given point (e.g., features that are available or performance on particular problems).
+At present, Kani does not print its version, but it's something we should strongly consider from now on.
+
+There are many benefits to including the version of Kani in its output.
+However, I think the main ones will be the following:
+ * **Earlier detection of version-related discrepancies**:
+ Users are likely to discuss discrepancies in verification outcomes by looking at Kani's output.
+ These may look exactly the same[^cbmc-version] (except for the discrepant value) on two different versions of Kani.
+ Including the version will help users realize sooner that they're using different versions of Kani.
+ * **Simpler issue triaging**:
+ New issues require users to post the Kani version they used.
+ Getting this information requires another call with `--version`, which wouldn't be needed if we simply printed the version.
+ Also, note that users may need to do more work if they aren't running Kani locally (e.g., Kani running in CI).
+
+In addition, printing the Kani version may be useful for other purposes (automate CI processes, help users realize they're using outdated versions, etc.).
+
+## User Experience
+
+The first line printed in any Kani invocation (either through `kani` or `cargo kani`, and regardless of subcommands) will inform users of the version.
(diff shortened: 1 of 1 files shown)


rfc/src/rfcs/0008-kani-version.md
```diff
@@ -0,0 +1,109 @@
...
+
+It'd be great to discuss any alternatives for the concrete format.
+At some point, I even thought about adding some ASCII art, but wanted to keep it short and simple.
+
+For example, we could:
+ - Replace the word `Launching` with another one.
+ - Prefix the version with `v` (so the version gets printed as `v0.29.0`, for example).
+ - Just print `Kani Rust Verifier <version>`, nothing else.
+
+These are low-level details which I'd love to discuss with you all.
```
JustusAdam, line 96: I don't feel super strongly about `Launching` or `v`, though one might argue they are redundant.
However it might makes sense to print whether it is running the cargo plugin or the standalone Kani. Maybe `Kani Rust Verifier (Cargo Plugin) 1.29.0` and vice versa with `standalone`.

rfc/src/rfcs/0008-kani-version.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Open questions
+
+I'm hoping that we can answer the following questions during the RFC:
+ 1. Do we want to print Kani's version?
```
JustusAdam, line 101: Yes. Especially because of the VSCode extension.

rfc/src/rfcs/0008-kani-version.md
```diff
@@ -0,0 +1,109 @@
...
+### Development versions
+
+The first line to be printed will be:
+
+```
+Launching the Kani Rust Verifier <version> (dev. version - commit: <commit>)
+```
+
+where `<version>` is the version of Kani under use, which follows the semantic versioning format `MAJOR.MINOR.PATCH`,
+and `<commit>` is the short hash (i.e., 7 hexadecimal digits with format `hhhhhhh`) of the `HEAD` commit.
```
JustusAdam, line 62: If we wanted to be super extra we could also print `(dirty)` if your worktree has uncommitted changes or add a hash for all files in `src` :joy:

JustusAdam (commented):
I like this idea. It's especially nice for cases where Kani is run programmatically, e.g. from the extension which might cause a different Kani version to run (because `$PATH` sucks).

JustusAdam:
Also I think the version should be included in the header when printing `--help`

rfc/src/rfcs/0008-kani-version.md
```diff
@@ -0,0 +1,109 @@
...
+
+The first line printed in any Kani invocation (either through `kani` or `cargo kani`, and regardless of subcommands) will inform users of the version.
+The behavior will be extended for development versions, where it'll print the short hash of the HEAD commit in addition to the version.
+
+### Release versions
+
+The first line to be printed will be:
+
+```
+Launching the Kani Rust Verifier <version>
```
celinval, line 42: Do we need the "the" here?

celinval (commented):
IMHO, I don't think we need an RFC for this change. It's a nice improvement to the output Kani. But it's a fairly simple change that shouldn't impact current and future users.

adpaco-aws:
> IMHO, I don't think we need an RFC for this change. It's a nice improvement to the output Kani. But it's a fairly simple change that shouldn't impact current and future users.

Agreed, I'll close this RFC and we can continue the discussion in the (to-be-developed) PR. Thanks @JustusAdam for the suggestions, I'll consider them for the implementation!

PR #2604, opened 2023-07-14
Do not run benchcomp in every PR

celinval:
### Description of changes: 

The performance benchmark PR is taking a lot of our CI time and many times it's not necessary. For example, benchmark job over 2h while all the other runs ran in less than 40min [in this run](https://github.com/model-checking/kani/actions/runs/5533957652)

This PR moves the performance benchmark to a separate workflow that only runs on pushes to main / release tags and PR's that are labeled with 'Z-BenchCI'

### Call-outs:

We still need to fix the release tag runs to compare against previous release. We haven't had this working before, so I hope it's not a blocker for this PR. I could just remove the push to tag trigger for now if people prefer or create an issue to finish supporting release comparison.

I created a label for now, but we can rename it or remove it if this PR doesn't get accepted.

### Testing:

* How is this change tested? I tested this in my fork. Here is [the job run](https://github.com/celinval/kani-dev/actions/runs/5557978872) for [a PR](https://github.com/celinval/kani-dev/pull/6) where I applied the label and here is [the run after I pushed to my fork main branch](https://github.com/celinval/kani-dev/actions/runs/5557992572).

* Is this a refactor change? It's a CI refactoring change

### Checklist


jaisnan (approved):
Thanks! I agree with the change. Should we document the need for the label in our dev-documentation guide?

Issue #2442, opened 2023-05-15
Measure compiler performance and overall execution time as part of our benchmarks


celinval:
**Proposed change:** Our benchmarks today only measure changes to verification time. We should also include overall execution time and the compilation time.

**Motivation:** Overall execution time is probably the most important metric since it reflects the exact user experience. Additionally, it would be valuable to understand how the compilation time is impacted with changes to the compiler.

celinval:
@karkhaz I need your help adding these. All the metrics beforehand were per-harness, while these are per package.

jaisnan:
Thanks for the issue! Should be a matter of adding parsing.

celinval:
We still don't have a reliable way to measure the compiler performance. 😞

PR #2607, opened 2023-07-14
Add a script to measure build performance

celinval:
### Description of changes: 

In order to avoid redundancy and make it more maintainable, I refactored the performance script.
  - I moved the logic into a common script file (kani-perf-setup.sh)
  - I also changed the performance to run in a temporary file
    - To simplify deletion
  - I changed benchcomp to work with the new location

### Resolved issues:

Related to #2442 

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

-  I haven't actually created the CI job or the benchcomp configuration / parser.
- @karkhaz does a benchcomp parser have access to the script output? I guess I could write things the output into a file... Ideally, I would like to get the path from the script output.

### Testing:

* How is this change tested? Manually ran the regression and the parser.

### Checklist


celinval:
@karkhaz can you help me out please? It seems that I cannot rename the output folder because it will stop working for previous versions of Kani. Which `kani-perf.sh` is benchcomp running?? Does this mean that I won't be able to add a new benchmark job that runs on older versions of Kani since the script won't exist there?

scripts/kani-perf-setup.sh
```diff
@@ -0,0 +1,85 @@
...
+
+run_benchmarks() {
+  suite=$(basename "${SUITE}")
+  mode="cargo-kani-test"
+  echo "Check compiletest suite=$suite mode=$mode"
+  cargo run -p compiletest -- --suite $suite --mode $mode --no-fail-fast --report-time "$@"
+}
+
+print_result() {
+  exit_code=$1
```
karkhaz, line 45: Is this intentional?

karkhaz (approved):
(Had a discussion offline about benchcomp).

Thanks very much for this, looking forward to having less duplication between CI and local invocations.

scripts/kani-perf-setup.sh
celinval: Yes. A bunch of these variables should be local, but I don't think it really matters here.

celinval:
Tbh, this solution became too hacky due to benchcomp configuration. I'll create a simple PR without any change to perf.

PR #2608, opened 2023-07-17
Add coverage test suite

jaisnan:
### Description of changes: 

Adds coverage test suite

### Resolved issues:

Resolves #ISSUE-NUMBER

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist
- [ ] Each commit message has a non-empty body, explaining why the change was made
- [ ] Methods or procedures are documented
- [ ] Regression or unit tests are included, or existing tests cover the modified code
- [ ] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #2609, opened 2023-07-19
Line coverage reports

### LINKED ISSUE #2585: Add test coverage for coverage information
**Proposed change:**

Coverage information generated by kani don't have any tests related to them. Test coverage should cover various language features, generate an lcov or json file, and match that in the regression test suite.

### LINKED ISSUE #2610: Coverage option that does not require `cbmc-viewer`
Requested feature: In the past, Kani has relied on `cbmc-viewer` to generate coverage reports. This requires Kani to make additional calls to `cbmc` with the `--cover` flag, then `cbmc-viewer` to essentially perform postprocessing on that output. Unfortunately, the information reported by `cbmc-viewer` isn't tested thoroughly (see #2048). But actually, Kani already has the tools required for generating coverage-oriented information. Having an option that enables this would allow us to remove the `cbmc-viewer` dependency and have end-to-end testing for coverage information.

Use case: Coverage reports.

adpaco-aws:
### Description of changes: 

Adds a coverage report feature to Kani that does not require `cbmc-viewer`.

In particular, it introduces the unstable `--coverage` verification option to `kani-driver` (only usable with the `-Z line-coverage` unstable feature), which in turn enables the `--coverage-checks` option in `kani-compiler`. This results in the injection of `coverage` checks (similar to `cover` statements, but with a fixed condition) before each Rust MIR statement and terminator in a basic block. The post-processing of coverage checks allows us to report coverage information at the line level.

This contribution also includes a new `compiletest` suite developed by @jaisnan for end-to-end testing of the emitted coverage information for different test cases.

The command is:

`kani file.rs --coverage -Z line-coverage`

For instance, for the `assume_false.rs` example
```rust
#[kani::proof]
fn wrong_coverage_2() {
    let a: u8 = kani::any();
    kani::assume(false);
    assert!(a < 5);
}
```
we obtain the verification results:

```
RESULTS:
Check 1: wrong_coverage_2.assertion.1
         - Status: UNREACHABLE
         - Description: "assertion failed: a < 5"
         - Location: assume_false.rs:5:5 in function wrong_coverage_2
```

and the following coverage results:

```
Coverage Results:
/home/ubuntu/kani/library/kani/src/lib.rs, 133, FULL
/home/ubuntu/kani/library/kani/src/lib.rs, 134, FULL
/home/ubuntu/kani/library/kani/src/lib.rs, 176, FULL
/home/ubuntu/kani/library/kani/src/lib.rs, 177, FULL

/home/ubuntu/kani/assume_false.rs, 3, FULL
/home/ubuntu/kani/assume_false.rs, 4, FULL
/home/ubuntu/kani/assume_false.rs, 5, NONE
/home/ubuntu/kani/assume_false.rs, 6, NONE

/home/ubuntu/kani/library/kani/src/arbitrary.rs, 55, FULL
```

### Resolved issues:

Resolves #2585 
Resolves #2608 
Resolves #2610 

### Related RFC:

#2612 

### Testing:

* How is this change tested? Existing tests + new `coverage` suite.

### Checklist


tests/coverage/reachable/assert/expected
adpaco-aws, line 1: I'm seeing empty files like this in many folders. Can you remove them?

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +437,99 @@ pub fn format_result(
...
+        for line in lines.iter() {
+            let satisfiable_statuses: Vec<bool> = files
+                .get(file)
+                .unwrap()
+                .iter()
+                .filter(|(x, _)| *line == *x)
+                .map(|(_, s)| s.contains("SATISFIED"))
+                .collect();
+
+            let covered_status = if satisfiable_statuses.iter().all(|&x| x) {
```
adpaco-aws, line 510: You can avoid the `.map(|(_, s)| s.contains("SATISFIED"))` above and do it on `x` here. Let's use better variable names too, please.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +437,99 @@ pub fn format_result(
...
+pub fn format_result_coverage(properties: &Vec<Property>) -> String {
+    let mut coverage_checks: Vec<CoverageCheck> = Vec::new();
+
+    for prop in properties {
+        let status = &prop.status;
+        let description = &prop.description;
+        let location = &prop.source_location;
+
+        let mut check =
+            CoverageCheck::new("".to_string(), 0, description.to_string(), status.to_string());
```
adpaco-aws, line 449: Not sure why you're creating a new type for coverage checks. You can just filter the properties with `property_class() == "coverage"`.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +437,99 @@ pub fn format_result(
...
+
+        let location_regex = Regex::new(r"(.+):(\d+):").unwrap();
+        let location_msg = format!("{location}\n");
```
adpaco-aws, line 452: Location information should be available in the properties already, no need to use regex right?

scripts/postcov.py
adpaco-aws, line 1: This can be removed now.

tests/coverage/unreachable/wrong_coverage_1/main.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+#[kani::proof]
+fn wrong_coverage_1() {
```
adpaco-aws, line 4: Let's have a more appropriate name for this function now, it's not wrong anymore!

tests/coverage/unreachable/wrong_coverage_2/main.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+#[kani::proof]
+fn wrong_coverage_2() {
```
adpaco-aws, line 4: Same here.

tools/compiletest/src/runtest.rs
```diff
@@ -380,6 +399,13 @@ impl<'test> TestCx<'test> {
         self.verify_output(&proc_res, &expected_path);
     }
 
+    /// Runs Kani in coverage mode on the test file specified by `self.testpaths.file`. An error
```
qinheping, line 402:
```suggestion
    /// Runs Kani in coverage mode on the test file specified by `self.testpaths.file`. 
```

tests/coverage/reachable/bounds/reachable_pass/test.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// This test checks that kani injects a reachability check for
+// index-out-of-bounds checks and that it reports ones that are unreachable.
```
qinheping, line 5: How do we tell whether a check such as index-out-of-bounds check is unreachable from the coverage result?

jaisnan: The comments are redundant. Thanks for pointing that out, shall correct them.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +421,86 @@ pub fn format_result(
...
+        }
+    }
+
+    let mut coverage_results: HashMap<String, Vec<(usize, String)>> = HashMap::new();
+
+    // loop through the files list and create a list of lines accessible in that file
+    for file in files.keys() {
+        let mut lines: HashSet<usize> = HashSet::new();
+        let mut line_results: Vec<(usize, String)> = Vec::new();
+        for check in files.get(file).unwrap() {
```
zhassan-aws, line 462: You can change the loop header to:
```rust
for (file, val) in files
```
to avoid doing a `get(file)` to get the value.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +421,86 @@ pub fn format_result(
...
+    formatted_output.push_str("\nCoverage Results:\n");
+
+    let coverage_checks: Vec<Property> =
+        properties.iter().filter(|&x| x.property_class() == "coverage").cloned().collect();
+
+    let mut sorted_checks: Vec<&Property> = coverage_checks.iter().collect();
+
+    sorted_checks.sort_by_key(|check| (&check.source_location.file, &check.source_location.line));
+
+    let mut files: HashMap<String, Vec<(usize, String)>> = HashMap::new();
```
zhassan-aws, line 439: Why not use `CheckStatus` instead of `String`?

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +421,86 @@ pub fn format_result(
...
+        // example - {3 -> ["SAT", "UNSAT"], 4 -> ["UNSAT"] ...}
+        for line in lines.iter() {
+            let is_line_satisfied: Vec<_> = files
+                .get(file)
+                .unwrap()
+                .iter()
+                .filter(|(line_number_accumulated, _)| *line == *line_number_accumulated)
+                .collect();
+
+            // If any of the statuses is UNSAT, we report that line as UNCOVERED
```
zhassan-aws, line 476: Should we introduce the notion of "Partially covered" like CBMC viewer does for lines with "mixed" covered and uncovered statements?

tests/coverage/reachable/assert-false/expected
```diff
@@ -0,0 +1,21 @@
+library/kani/src/arbitrary.rs, 55, COVERED
```
zhassan-aws, line 1: Code coverage should exclude dependencies and only report coverage for user crates.

tests/coverage/unreachable/tutorial_unreachable/expected
```diff
@@ -0,0 +1,5 @@
+tests/coverage/unreachable/tutorial_unreachable/main.rs, 5, COVERED
+tests/coverage/unreachable/tutorial_unreachable/main.rs, 6, UNCOVERED
```
zhassan-aws, line 2: Why are line 6 and 7 uncovered?

zhassan-aws (commented):
As mentioned in my comments on the RFC, I think the approach for injecting coverage checks should rely on coverage information from MIR.

For tests, I suggest creating new tests that are specifically targeting the feature rather than copying tests that are targeting a different feature.

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
```diff
@@ -18,23 +18,54 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    self.current_fn_mut().push_onto_block(scode);
+                } else {
+                    let scode = self.codegen_statement(stmt);
+                    self.current_fn_mut().push_onto_block(scode.with_label(label));
+                }
 
                 for s in &bbd.statements[1..] {
+                    if check_coverage {
+                        let span = s.source_info.span;
+                        // TODO: Push cover statement based on some TBD condition
```
celinval, line 52: ?

tests/coverage/reachable/rem-zero/reachable_pass/expected
```diff
@@ -0,0 +1,4 @@
+/home/ubuntu/kani/tests/coverage/reachable/rem-zero/reachable_pass/test.rs, 5, UNCOVERED
```
celinval: Shouldn't use full path

kani-driver/src/cbmc_property_renderer.rs
celinval: +1

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
```diff
@@ -18,23 +18,54 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        let cover = self.codegen_coverage(span);
+                        self.current_fn_mut().push_onto_block(cover);
+                    }
                     let stmt = self.codegen_statement(s);
                     self.current_fn_mut().push_onto_block(stmt);
                 }
-                let term = self.codegen_terminator(bbd.terminator());
-                self.current_fn_mut().push_onto_block(term);
+                // TODO: test with division by zero or overflow that false
```
celinval, line 59: please address this and remove the TODO

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -45,6 +45,10 @@ pub enum PropertyClass {
     ///
     /// SPECIAL BEHAVIOR: "Errors" for this type of assertion just mean "reachable" not failure.
     Cover,
+    /// See [GotocCtx::codegen_coverage] below. Always an `assert(false)` that's not an error.
+    ///
+    /// SPECIAL BEHAVIOR: "Errors" for this type of assertion just mean "reachable" not failure.
+    Coverage,
```
celinval, line 51: Can you please document the difference between `Cover` and `Coverage`? How is someone supposed to know when to use one over the other?

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
```diff
@@ -18,23 +18,54 @@ impl<'tcx> GotocCtx<'tcx> {
         let label: String = self.current_fn().find_label(&bb);
         // the first statement should be labelled. if there is no statements, then the
         // terminator should be labelled.
+        let check_coverage = self.queries.check_coverage;
         match bbd.statements.len() {
             0 => {
                 let term = bbd.terminator();
-                let tcode = self.codegen_terminator(term);
-                self.current_fn_mut().push_onto_block(tcode.with_label(label));
+                if check_coverage {
```
celinval, line 25: I won't lie... not a big fan of all these `if check_coverage` in this function. It can be in a follow up PR, but I think wrapping the coverage handling logic will help a lot with maintenance and readability here.

celinval (commented):
I think we might need to rethink these tests. They may become a maintenance pain.

jaisnan:
> I think we might need to rethink these tests. They may become a maintenance pain.

Could you clarify how they might become a maintenance pain? Is it because they directly use line numbers, and are not very robust?

kani-driver/src/cbmc_property_renderer.rs
jaisnan: Done

tests/coverage/reachable/assert-false/expected
adpaco-aws: I'm not sure I agree with this. If we exclude dependencies, won't it be more confusing for users who are debugging?

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -101,7 +105,11 @@ impl<'tcx> GotocCtx<'tcx> {
         loc: Location,
     ) -> Stmt {
         let property_name = property_class.as_str();
-        Stmt::assert(cond, property_name, message, loc)
+        if self.queries.check_coverage && property_class != PropertyClass::Coverage {
+            Stmt::assume(cond, loc)
```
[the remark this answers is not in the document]
zhassan-aws: Right, but is this function used for instrumenting checks? I checked the code and found that it's mainly used for cover and reachability checks (which we should **not** use an assumption for). The only case it's used for checks is https://github.com/model-checking/kani/blob/319d859bf2fc32e6025b9844538615f612b9788e/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs#L193
but this should probably be fixed to use `codegen_assert_assume` like all other checks.

tests/coverage/reachable/assert-false/expected
zhassan-aws: AFAIK, code coverage tools exclude dependencies, because what users care about in testing is that their own code is covered. The target is to get to 100% code coverage with tests, which is likely going to be impossible if code in dependencies needs to be covered as well.

jaisnan: I agree with @zhassan-aws's comments here, if we are calling it line coverage per harness, including coverage information about the library code makes it confusing.

tools/compiletest/src/runtest.rs
```diff
@@ -75,6 +77,7 @@ impl<'test> TestCx<'test> {
             KaniFixme => self.run_kani_test(),
             CargoKani => self.run_cargo_kani_test(false),
             CargoKaniTest => self.run_cargo_kani_test(true),
+            CoverageBased => self.run_expected_coverage_test(),
```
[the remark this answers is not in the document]
jaisnan: We could, but we thought it would be better to strongly enforce the flag for this suite, since the testing is sensitive to the flag being present throughout the suite. If it's not a problem, I think it makes sense to add a new mode for now and change the logic to not use expected files in a future PR.

tests/coverage/reachable/assert-false/expected
adpaco-aws: Whether coverage in dependencies counts towards certain coverage metrics or not is a different discussion. Note that users doing harness debugging may be interested in the coverage reported on external dependencies, and if we filter these results out they'll have no indication of where the trace went.

jaisnan: Resolved offline. Left as an open question in the [RFC](https://github.com/model-checking/kani/blob/main/rfc/src/rfcs/0008-line-coverage.md#open-questions) but for now we are going to display dependency coverage information in the post-processed output as well.

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
adpaco-aws: I'd like to know how you'd do this and take care of it now. I've [refactored this code a little bit](https://github.com/model-checking/kani/pull/2609/commits/823d72a42896a6d34e909710982448384d598e4f), but I'm not a fan of the code neither.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -134,6 +142,16 @@ impl<'tcx> GotocCtx<'tcx> {
         self.codegen_assert(cond.not(), PropertyClass::Cover, msg, loc)
     }
 
+    /// Generate code for coverage reports
```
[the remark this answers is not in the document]
adpaco-aws: Done!

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
adpaco-aws: I've added the main differences here and made clear the use case.

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
celinval: BTW, I see `div-zero` tests.So I'm assuming you can remove this todo already.

adpaco-aws: Included in the latest changes.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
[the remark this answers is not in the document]
adpaco-aws: The change was undone recently (i.e., we don't replace the `assert` anymore). I also think we should be using `codegen_assert_assume` in the intrinsic (otherwise there should be a comment explaining the exception), so I've fixed it.

kani-driver/src/args/mod.rs
```diff
@@ -300,6 +300,9 @@ pub struct VerificationArgs {
     )]
     pub enable_stubbing: bool,
 
+    #[arg(long, hide_short_help = true, requires("enable_unstable"))]
```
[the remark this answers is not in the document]
adpaco-aws: Done.

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
celinval: Sure. Let's take this offline since I don't think it's a blocker for this PR.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -45,6 +45,19 @@ pub enum PropertyClass {
     ///
     /// SPECIAL BEHAVIOR: "Errors" for this type of assertion just mean "reachable" not failure.
     Cover,
+    /// A specialization of the `Cover` class used for collecting coverage information.
```
celinval, line 48: What about calling it CodeCoverage instead?

I would also say something like used for "code coverage instrumentation". I don't think you should mention specialization of "Cover", since they have very different purpose as you mentioned.

kani-driver/src/call_cbmc.rs
```diff
@@ -303,12 +303,31 @@ impl VerificationResult {
...
+                        should_panic,
+                        failed_properties,
+                        show_checks,
+                    );
+                    writeln!(result, "Verification Time: {}s", self.runtime.as_secs_f32()).unwrap();
+                    return result;
+                }
+
                 let mut result =
                     format_result(results, status, should_panic, failed_properties, show_checks);
```
celinval, line 332:
```suggestion
                let result = if coverage_mode {
                    formatter_coverage(
                        results,
                        status,
                        should_panic,
                        failed_properties,
                        show_checks,
                    )
                } else {
                    format_result(results, status, should_panic, failed_properties, show_checks)
                };
```

Ps.: Maybe apply the patch locally and ensure it compiles. :)

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +421,103 @@ pub fn format_result(
     result_str
 }
 
+/// Seperate checks into coverage and non-coverage based on property class and format them seperately for --coverage. We report both verification and processed coverage
+/// results
+pub fn formatter_coverage(
```
celinval, line 426: this should be `format_coverage`, since we have `format_results`

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +421,103 @@ pub fn format_result(
...
+    let result = format!("{}\n{}", verification_output, coverage_output);
+
+    result
+}
+
+/// Formats a coverage result item (i.e., the subset of verification checks with coverage property class).
+/// To be used when the user requests coverage information with --coverage. The output is tested through the coverage-based testing suite, not the regular expected suite.
+/// Loops through each of the check with a coverage property class and gives a status of FULL if all checks pertaining
+/// to a line number are SATISFIED. Similarly, it gives a status of NONE if all checks related to a line are UNSAT. If a line has both, it reports PARTIAL coverage.
+pub fn format_result_coverage(properties: &[Property]) -> String {
```
celinval, line 450:
```suggestion
fn format_result_coverage(properties: &[Property]) -> String {
```

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +421,103 @@ pub fn format_result(
...
+
+    let verification_output =
+        format_result(&non_coverage_checks, status, should_panic, failed_properties, show_checks);
+    let coverage_output = format_result_coverage(&coverage_checks);
+    let result = format!("{}\n{}", verification_output, coverage_output);
+
+    result
+}
+
+/// Formats a coverage result item (i.e., the subset of verification checks with coverage property class).
```
celinval, line 446:
```suggestion
/// Generate coverage result from all coverage properties (i.e., the checks with "coverage" property class).
```

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +421,103 @@ pub fn format_result(
...
+            } else {
+                "PARTIAL".to_string()
+            };
+
+            line_results.push((*line, covered_status));
+        }
+
+        line_results.sort_by_key(|&(line, _)| line);
+        coverage_results.insert(file.clone(), line_results);
+    }
```
celinval, line 508: I was wondering if we could improve this a bit. I think you could actually process everything using one for loop.

Something like:

```rust
    #[derive(PartialEq, Eq, AsRefStr)]
    #[strum(serialize_all = "UPPERCASE")]
    enum CoverStatus {
        Full,
        Partial,
        None
    }
    let mut coverage_results : BTreeMap<String, BTreeMap<usize, CoverStatus>> = BTreeMap::default();
    for prop in properties {
        let src = prop.source_location;
        let file_entries = coverage_results.entry(src.file).or_default();
        let check_status = if prop.status == CheckStatus:: Satisfied { CoverStatus::Full } else { CoverStatus::None };
        file_entries.entry(src.line).and_modify(|line_status| {
            if line_status == check_status {
                line_status
            } else {
                CheckStatus::Partial
            }
        }.or_insert(check_status);
    }
```

celinval (commented):
It's looking good. Mostly minor changes.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
adpaco-aws: Makes sense, done!

zhassan-aws (commented):
Re-iterating my request to remove all tests copied from `tests/expected/reach` as they don't add much of a value for testing the code coverage feature. If a test is useful for checking the code coverage feature, then the test description should clearly mention what it is testing.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +421,102 @@ pub fn format_result(
...
+/// to a line number are SATISFIED. Similarly, it gives a status of NONE if all checks related to a line are UNSAT. If a line has both, it reports PARTIAL coverage.
+fn format_result_coverage(properties: &[Property]) -> String {
+    let mut formatted_output = String::new();
+    formatted_output.push_str("\nCoverage Results:\n");
+
+    let mut coverage_checks: Vec<&Property> = properties.iter().collect();
+
+    coverage_checks.sort_by_key(|check| (&check.source_location.file, &check.source_location.line));
+
+    let mut files: HashMap<String, Vec<(usize, CheckStatus)>> = HashMap::new();
```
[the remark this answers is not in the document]
jaisnan: As Celina has suggested, I've changed this to use a BTree instead and have simplified the function.

kani-driver/src/cbmc_property_renderer.rs
jaisnan: Done! Thanks for the suggestion

celinval (commented):
I haven't checked the tests yet, but the code looks good to me (modulo @zhassan-aws comments). I'll check the tests tomorrow.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -431,13 +431,14 @@ pub fn format_coverage(
...
-        properties.iter().filter(|&x| x.property_class() != "coverage").cloned().collect();
+        properties.iter().filter(|&x| x.property_class() != "code_coverage").cloned().collect();
     let coverage_checks: Vec<Property> =
-        properties.iter().filter(|&x| x.property_class() == "coverage").cloned().collect();
+        properties.iter().filter(|&x| x.property_class() == "code_coverage").cloned().collect();
 
     let verification_output =
         format_result(&non_coverage_checks, status, should_panic, failed_properties, show_checks);
     let coverage_output = format_result_coverage(&coverage_checks);
+    print!("printing coverage output = {}", coverage_output);
```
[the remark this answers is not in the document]
jaisnan: Yes, removed by a previous commit.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -149,6 +151,25 @@ static CBMC_ALT_DESCRIPTIONS: Lazy<CbmcAltDescriptions> = Lazy::new(|| {
...
+#[derive(PartialEq, Eq, AsRefStr, Clone, Copy)]
+#[strum(serialize_all = "UPPERCASE")]
+// The status of coverage reported by Kani
+enum CoverStatus {
+    Full,
+    Partial,
+    None,
+}
+
+impl fmt::Display for CoverStatus {
```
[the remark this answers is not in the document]
jaisnan: Done!

tests/coverage/reachable/assert/reachable_fail/test.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright Kani Contributors
```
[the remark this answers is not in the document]
jaisnan: Removed. Other tests from `expected/reach` have also been removed.

tests/coverage/unreachable/contradiction/expected
```diff
@@ -0,0 +1,6 @@
+coverage/unreachable/contradiction/main.rs, 5, FULL
+coverage/unreachable/contradiction/main.rs, 6, FULL
+coverage/unreachable/contradiction/main.rs, 7, PARTIAL
```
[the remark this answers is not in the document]
jaisnan: The two checks correspond to the `if` (which is SUCCESS) and the condition (which is FAILURE). Since we aggregate checks by line and not by column, the line is being reported as PARTIAL.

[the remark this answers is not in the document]
adpaco-aws: This case is confusing due to compiler optimizations: the entire `if x < 2 { ... }` statement can be optimized away because the compiler knows the condition will always be `false`. The MIR code obtained using the command shown [here](https://model-checking.github.io/kani/cheat-sheets.html#debug) doesn't show the optimized version, but the Rust Playground can if we replace the `kani::any()` call with a concrete value. This is what I got when replacing the call with `20`:
```
bb0: {
        _1 = const 20_u8;                // scope 0 at [src/main.rs:2:17: 2:19](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=330a3113cd82d916e617018557be49fc#)
        _2 = Gt(_1, const 5_u8);         // scope 1 at [src/main.rs:3:8: 3:13](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=330a3113cd82d916e617018557be49fc#)
        switchInt(move _2) -> [0: bb2, otherwise: bb1]; // scope 1 at [src/main.rs:3:8: 3:13](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=330a3113cd82d916e617018557be49fc#)
    }

    bb1: {
        goto -> bb4;                     // scope 1 at [src/main.rs:4:12: 4:17](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=330a3113cd82d916e617018557be49fc#)
    }

    bb2: {
        _4 = Lt(_1, const 10_u8);        // scope 1 at [src/main.rs:8:17: 8:23](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=330a3113cd82d916e617018557be49fc#)
        _3 = Not(move _4);               // scope 1 at [src/main.rs:8:9: 8:24](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=330a3113cd82d916e617018557be49fc#)
        switchInt(move _3) -> [0: bb4, otherwise: bb3]; // scope 1 at [src/main.rs:8:9: 8:24](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=330a3113cd82d916e617018557be49fc#)
    }

    bb3: {
        _5 = core::panicking::panic(const "assertion failed: x < 10"); // scope 1 at [src/main.rs:8:9: 8:24](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=330a3113cd82d916e617018557be49fc#)
                                         // mir::Constant
                                         // + span: [src/main.rs:8:9: 8:24](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=330a3113cd82d916e617018557be49fc#)
                                         // + literal: Const { ty: fn(&'static str) -> ! {core::panicking::panic}, val: Value(<ZST>) }
                                         // mir::Constant
                                         // + span: no-location
                                         // + literal: Const { ty: &str, val: Value(Slice(..)) }
    }

    bb4: {
        return;                          // scope 0 at [src/main.rs:10:2: 10:2](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=330a3113cd82d916e617018557be49fc#)
    }
```

Note that `bb1` has been replaced with an unconditional `goto` statement. The `goto` statement itself is reachable and reported as such. But the checks that come after it in the same basic block won't be. Because of that, we report `PARTIAL` instead of `FULL` (which is what most people would expect). This can be considered a spurious "failure" for coverage: not desirable but still sound.

tests/coverage/reachable/assert-false/expected
```diff
@@ -0,0 +1,8 @@
+coverage/reachable/assert-false/main.rs, 6, FULL
+coverage/reachable/assert-false/main.rs, 7, FULL
+coverage/reachable/assert-false/main.rs, 11, FULL
+coverage/reachable/assert-false/main.rs, 12, FULL
+coverage/reachable/assert-false/main.rs, 15, FULL
+coverage/reachable/assert-false/main.rs, 16, FULL
+coverage/reachable/assert-false/main.rs, 17, PARTIAL
```
celinval, line 7: Why is this PARTIAL?

tests/coverage/reachable/assert/reachable_pass/test.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#[kani::proof]
+fn main() {
+    let x = 5;
```
celinval, line 6: Will constant propagation eventually break this?

You could use instead:
```rust
let x = kani::any_where(|val| val == 5);
```

tests/coverage/reachable/bounds/reachable_fail/test.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+fn get(s: &[i16], index: usize) -> i16 {
+    s[index] // PARTIAL: `index = 15` causes failure, but `s[index]` is `COVERED`
```
celinval, line 5: This is rather confusing, eh? I guess the return statement is unreachable.

tools/compiletest/src/common.rs
```diff
@@ -19,6 +19,7 @@ pub enum Mode {
     KaniFixme,
     CargoKani,
     CargoKaniTest, // `cargo kani --tests`. This is temporary and should be removed when s2n-quic moves --tests to `Cargo.toml`.
+    CoverageBased,
```
celinval, line 22: I still don't understand why we are adding this here... But since we still need to improve how we test coverage, we  can handle this later.

tests/coverage/unreachable/while-loop-break/main.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: Please add a description

tests/coverage/unreachable/vectors/main.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: Please add a description

tests/coverage/unreachable/variant/main.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: Please add a description, e.g.: This test checks that code coverage on a match statement with an enum behaves correctly

celinval (approved):
I think we need to polish the coverage output to make it more intuitive, but this is a good start.

Please create issues to:
 1. Improve the tests
 2. Create a user friendly output
 3. Export results to a well-supported output
 4. Investigate if we can make some results more intuitive (maybe compare against Rust code coverage results).
 5. All other open questions in the RFC

tests/coverage/unreachable/turned-off/test.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
```
zhassan-aws, line 2: This test is meant to check when assertion reachability checks are turned off (as indicated by its name). What is it testing for code coverage?

tests/coverage/unreachable/rem-zero/test.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: This test was introduced to check that the assertion associated with a remainder operation is instrumented with a reachability check, and the result of the reachability check is reflected in the overall result. Is it serving a purpose for code coverage?

tests/coverage/unreachable/overflow/test.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: Redundant test?

tests/coverage/unreachable/overflow-neg/test.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: Redundant test?

tests/coverage/unreachable/div-zero/test.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: Is this serving a purpose for code coverage?

tests/coverage/unreachable/debug-assert-ne/test.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: Redundant test?

zhassan-aws (commented):
There are a few redundant tests remaining. Otherwise, looks good.

tests/coverage/unreachable/turned-off/test.rs
adpaco-aws: Removed.

tests/coverage/unreachable/rem-zero/test.rs
adpaco-aws: Removed.

tests/coverage/unreachable/overflow/test.rs
adpaco-aws: Removed.

tests/coverage/unreachable/overflow-neg/test.rs
adpaco-aws: Removed.

kani-driver/src/cbmc_output_parser.rs
```diff
@@ -322,18 +328,22 @@ impl std::fmt::Display for TraceData {
...
+    Unsatisfiable, // for `cover` properties only
 }
 
 impl std::fmt::Display for CheckStatus {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let check_str = match self {
             CheckStatus::Satisfied => style("SATISFIED").green(),
             CheckStatus::Success => style("SUCCESS").green(),
+            CheckStatus::Covered => style("COVERED").green(),
+            CheckStatus::Uncovered => style("UNCOVERED").yellow(),
```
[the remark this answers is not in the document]
adpaco-aws: Changed to red.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -421,6 +431,70 @@ pub fn format_result(
...
+
+        // Create Map<file, Map<line, status>>
+        file_entries
+            .entry(src.line.unwrap().parse().unwrap())
+            .and_modify(|line_status| {
+                if *line_status == check_status {
+                    *line_status = check_status
+                } else {
+                    *line_status = CoverStatus::Partial
+                }
```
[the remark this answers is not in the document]
adpaco-aws: Thanks.

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -613,6 +688,25 @@ fn update_properties_with_reach_status(
...
+/// users but rather internally consumed and reported finally as `PARTIAL`, `FULL`
+/// or `NONE` based on aggregated line coverage results.
+fn update_results_of_code_covererage_checks(mut properties: Vec<Property>) -> Vec<Property> {
+    for prop in properties.iter_mut() {
+        if prop.is_code_coverage_property() {
+            if prop.status == CheckStatus::Success {
+                prop.status = CheckStatus::Uncovered;
+            } else if prop.status == CheckStatus::Failure {
+                prop.status = CheckStatus::Covered;
+            }
```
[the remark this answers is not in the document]
adpaco-aws: Done, and changed into `match`.

tests/coverage/reachable/assert-false/expected
```diff
@@ -0,0 +1,8 @@
+coverage/reachable/assert-false/main.rs, 6, FULL
```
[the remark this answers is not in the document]
adpaco-aws: Order isn't a requirement, just more comfortable for humans to read.

tests/coverage/reachable/assert/reachable_pass/test.rs
adpaco-aws: Using that now.

tests/coverage/reachable/bounds/reachable_fail/test.rs
adpaco-aws: Yeah, sorry... you're right.

tools/compiletest/src/common.rs
adpaco-aws: Having a dedicated mode for coverage seems OK to me. Or am I missing something here?

Last night I was thinking of some improvements that we could make for this mode (e.g., annotate lines with a comment that includes the status), but it doesn't make sense to invest a lot on it since we'll change how we represent results soon.

tests/coverage/reachable/assert-false/expected
adpaco-aws: Because we override and expand these `assert`s as:
```
    ($cond:expr, $($arg:tt)+) => {{
        kani::assert(!!$cond, concat!(stringify!($($arg)+)));
        // [...]
        if false {
            __kani__workaround_core_assert!(true, $($arg)+);
        }
    }};
```
The code there includes some checks that won't be covered.

adpaco-aws:
Created the following issues:
 * #2635 
 * #2636 
 * #2638 
 * #2639 
 * #2640 
 * #2641 
 * #2642 

which I'm hoping take care of the remaining work here and in the RFC (#2618) 

> There are a few redundant tests remaining

Indeed, there were many tests that were redundant or didn't have a clear purpose. I removed a bunch of them and have #2635 for another pass.

celinval:
@adpaco-aws should we create a milestone for those?

adpaco-aws:
Yes, just created a [Coverage reports](https://github.com/model-checking/kani/milestone/32) milestone to keep track of all of these. I create a new one instead of using the [Source-based coverage report generation](https://github.com/model-checking/kani/milestone/2) milestone because that one contains issues with the `visualize` feature, and it's not clear to me that we should mix both.

Issue #1777, opened 2022-10-12
Generate LCOV coverage


JonathanWoollett-Light:
This is like: https://github.com/model-checking/kani/issues/1706

### Requested feature

Provide argument to generate LCOV coverage output.

### Use case

For checking coverage in CI it would be useful to combine Kani coverage with coverage generated from typical tests (by tools like [grcov](https://github.com/mozilla/grcov).

It would also be useful to include documentation on this joint coverage in CI (this could be here https://model-checking.github.io/kani/install-github-ci.html)

Issue #1706, opened 2022-09-21
Generate GCOV coverage output from `cargo kani`


allenste-aws:
Requested feature: Generate GCOV coverage output from `cargo kani --visualize`
Use case: GCOV coverage reports are integrated with our CI framework and Code review tools. We can fail builds if coverage drops. Reviewers can see what lines have not been covered by tests and proofs.

tedinski:
Absolutely good idea.

Just some additional background:

The rust book has a section on coverage (https://doc.rust-lang.org/rustc/instrument-coverage.html) which states:

> The Rust compiler includes two code coverage implementations:
> * A GCC-compatible, gcov-based coverage implementation, enabled with -Z profile, which derives coverage data based on DebugInfo.
> * A source-based code coverage implementation, enabled with -C instrument-coverage, which uses LLVM's native, efficient coverage instrumentation to generate very precise coverage data.

So between it seeming to be a Rust thing and user request, gcov seems like a good option. But we should consider whether the other format should be supported as well. So far I haven't found good documentation about what's "typical" out there for Rust projects, in terms of measuring test coverage...

carolynzech:
Closing this since we are now deprecating --visualize option.

carolynzech:
Reopening with `--visualize` removed from title instead.

PR #2612, opened 2023-07-20
RFC: Line Coverage

adpaco-aws:
### Description of changes: 

An RFC to discuss an option in Kani to generate verification-based line coverage reports.

Rendered version is available [here](https://github.com/adpaco-aws/rmc/blob/rfc-coverage/rfc/src/rfcs/0008-line-coverage.md).

Proof-of-concept is available in #2609 

### Resolved issues:

Resolves (partially) #2610 

### Testing:

### Checklist


rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,89 @@
...
+
+![Coverage reported on the Kani VS Code Extension](../images/0008/coverage-kani-vscode.png)
+
+How to activate and display coverage information in the extension is out of scope for this RFC.
+We will need to change our documentation (including the tutorial) to use this option if we ever remove `--visualize` in favor of this one.
+
+## Detailed Design
+
+We will add a new unstable `--coverage` verification option to Kani.
+This will prevent `kani-driver` from reporting verification results as usual and instead output the coverage information described above.
```
zhassan-aws, line 48: Why not report both (coverage information and verification results) with this option?

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,89 @@
...
+It's likely that this will be done through a new output format that's not exposed to users.
+
+We will also add a new `--coverage-checks` option to `kani-compiler`, which will result in the injection of coverage checks before each Rust statement and terminator[^coverage-experiments].
```
zhassan-aws, line 51: How does injecting coverage checks before each statement and terminator compare to Rust's instrumentation-based code coverage? One of the main requests in #1706 and #1777 is to generate coverage information that is consistent with what the compiler generates.

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,89 @@
...
+This feature is expected to have a substantial impact for Kani.
+Users frequently ask for coverage-related options but Kani doesn't have one.
+That's because, some months ago, we noticed incorrect coverage results were being reported with the `--visualize` option, prompting us to disable the coverage reports emitted by `cbmc-viewer`.
+
+## Open questions
+
+Open questions:
+ * Do we want a `--coverage` option or a `coverage` subcommand? Coverage can be seen as a different workflow but also a verification option.
+
+We expect to remove the `--visualize` before stabilizing this feature.
```
zhassan-aws, line 77: There's no need to make this feature dependent on the removal of the `--visualize` option.

zhassan-aws (commented):
I like the proposal and think that implementing coverage in Kani is the way to go. However, I think we should rely on existing compiler APIs (e.g. [CodeRegion](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/coverage/struct.CodeRegion.html)) to compute coverage that is consistent with compiler-instrumented coverage.

zhassan-aws:
In particular, can we implement a MIR pass similar to [rustc's `InstrumentCoverage`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_mir_transform/coverage/struct.InstrumentCoverage.html)? This is described in some details in the coverage instrumentation dev guide: 
https://github.com/rust-lang/rustc-dev-guide/blob/master/src/llvm-coverage-instrumentation.md#mir-pass-instrumentcoverage

rfc/src/rfcs/0008-viewerless-coverage.md
jaisnan: One of the reasons is that we post-process the regular output to give the coverage information. If we were to include verification results, we'd need to post-process verification results in a way that is consistent with the request for coverage information, and we've seen that the resulting combined output can be misleading/confusing to the user. So to keep it easy for the user to interpret, --coverage returns only coverage output.

The second is that it, it is easy for the extension to consume this standardized coverage output as well.

adpaco-aws:
> However, I think we should rely on existing compiler APIs (e.g. [CodeRegion](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/coverage/struct.CodeRegion.html)) to compute coverage that is consistent with compiler-instrumented coverage.

We both agree on this, and the RFC should discuss these references in more detail.

That said, my main concern as of now with the compiler APIs for coverage is that they heavily rely on LLVM intrinsics and tools to do most of the profiling work. We can definitely plug in the [InstrumentCoverage](https://doc.rust-lang.org/stable/nightly-rustc/rustc_mir_transform/coverage/struct.InstrumentCoverage.html) MIR pass to our compiler, but our backend is not LLVM-based so it's not clear to me what to do then (note that LLVM intrinsics are used to increment the region-based counters, but then you also need the [LLVM runtime](https://github.com/rust-lang/rustc-dev-guide/blob/master/src/llvm-coverage-instrumentation.md#llvm-runtime-dependency) to extract their values). So I don't think the compiler APIs are directly usable for us unless some work happens in our backend first.

In summary, I don't think we're capable of adopting these APIs at the moment, but I'm looking forward to learning more about them and eventually replace the proposed mechanism. But that's out of scope for this version, which aims at line-level coverage information.

rfc/src/rfcs/0008-viewerless-coverage.md
adpaco-aws: True, let me remove that.

rfc/src/rfcs/0008-viewerless-coverage.md
zhassan-aws: > we'd need to post-process verification results in a way that is consistent with the request for coverage information, and we've seen that the resulting combined output can be misleading/confusing to the user

Can you clarify what you mean by that?

> it is easy for the VS Code extension to consume this separated coverage output as well.

It shouldn't be difficult to consume with verification results included.

adpaco-aws: This is mainly because we're replacing assertions with assumptions when doing a coverage run, so we cannot report them as verification results. Added a footnote regarding this.

zhassan-aws:
> We can definitely plug in the [InstrumentCoverage](https://doc.rust-lang.org/stable/nightly-rustc/rustc_mir_transform/coverage/struct.InstrumentCoverage.html) MIR pass to our compiler, but our backend is not LLVM-based so it's not clear to me what to do then (note that LLVM intrinsics are used to increment the region-based counters

True: the `InstrumentCoverage` pass cannot be used as is. But what I meant is that we can implement our own version of it that does not rely on any LLVM intrinsics. In particular, from [the description of the pass ](https://github.com/rust-lang/rustc-dev-guide/blob/master/src/llvm-coverage-instrumentation.md#implementation-details-of-the-instrumentcoverage-mir-pass), it performs multiple steps, one of which is to determine a minimum set of spans for computing coverage. Next, for each span, it injects a counter. Our version can use the step that computes the minimum set of spans, but replace the counter injection step with the injection of a cover statement (`assert(false)`). Does this make sense?

adpaco-aws:
Yes, that makes sense. Thank you!

Do you have any feedback on the UX side?

zhassan-aws:
> Do you have any feedback on the UX side?

One question is: how does this work with multiple harnesses? Ideally, one would want a single coverage report where a line is considered covered if it is reachable from one or more harnesses. That way, one can focus on writing new harnesses that cover lines that are currently uncovered by any of the existing harnesses.

zhassan-aws:
Another question: how would we report coverage for functions in user code that are not reachable from a harness? As a user, I would expect the lines in those functions to be reported as uncovered, but the pruning done by the MIR linker may remove those functions altogether, and thus no checks would be generated for them at all.

adpaco-aws:
> We can implement our own version of it that does not rely on any LLVM intrinsics. In particular, from [the description of the pass ](https://github.com/rust-lang/rustc-dev-guide/blob/master/src/llvm-coverage-instrumentation.md#implementation-details-of-the-instrumentcoverage-mir-pass), it performs multiple steps, one of which is to determine a minimum set of spans for computing coverage. Next, for each span, it injects a counter. Our version can use the step that computes the minimum set of spans, but replace the counter injection step with the injection of a cover statement (assert(false)).

Today I tried implementing a MIR pass similar to [`InstrumentCoverage`](https://github.com/rust-lang/rustc-dev-guide/blob/master/src/llvm-coverage-instrumentation.md#implementation-details-of-the-instrumentcoverage-mir-pass) but I haven't been successful. The main problem is that **the `coverage` module in `rustc_mir_transform` is private**, so all types (`BasicCoverageBlocks`, `CoverageSpan`, etc.) and APIs (`inject_counters`, `generate_coverage_spans`, etc.) used for the transformation pass need to copied into `kani-compiler`. I tried bringing them into `kani-compiler` so I could experiment with this approach but at some point the number of compiler errors was too high for me to resolve in a single day.

But I learned something important along the way: **This instrumentation is quite similar to the one proposed in the RFC**, but with many steps dedicated to optimizing runtime. In particular, I'd like to point you to these comments in [`CoverageSpan`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir_transform/coverage/spans/struct.CoverageSpan.html) and [`CoverageSpans`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir_transform/coverage/spans/struct.CoverageSpans.html#method.generate_coverage_spans):

> Each `Span` maps to a `CoverageSpan` that references the originating BCB and one or more MIR `Statement`s and/or `Terminator`s. Initially, the `Span`s come from the `Statement`s and `Terminator`s, but subsequent transforms can combine adjacent `Span`s and `CoverageSpan` from the same BCB [...]

> Converts the initial set of `CoverageSpan`s (one per MIR `Statement` or `Terminator`) into a minimal set of `CoverageSpan`s [...]

Because of that, I'm confident that our proposed implementation, which injects coverage statements for each statement and terminator, is essentially a naive version of the [`InstrumentCoverage`](https://github.com/rust-lang/rustc-dev-guide/blob/master/src/llvm-coverage-instrumentation.md#implementation-details-of-the-instrumentcoverage-mir-pass) algorithm. Steps like [`make_bcb_counters`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir_transform/coverage/counters/struct.BcbCounters.html#method.make_bcb_counters) wouldn't add much in our case since we don't need to distinguish between different kinds of counters. I'm hoping this makes clearer the relation between the proposed approach and Rust's coverage instrumentation.

Another thing that I've tried is passing `-C instrument-coverage` when calling the Rust compiler from Kani. This resulted in the error:

```
error[E0463]: can't find crate for `profiler_builtins`
  |
  = note: the compiler may have been built without the profiler runtime
```

I was able to work around this by also adding the `-Z no-profiler-runtime`, but then all I got was a mysterious assertion:

```
Check 1: wrong_coverage_2.missing_definition.1
         - Status: FAILURE
         - Description: "assertion"
         - Location: assume_false.rs:3:17 in function wrong_coverage_2

Check 2: wrong_coverage_2.assertion.1
         - Status: UNREACHABLE
         - Description: "assertion failed: a < 5"
         - Location: assume_false.rs:5:5 in function wrong_coverage_2
```

Debug logs showed a single counter being injected for the whole block:

```
2023-07-21T18:22:33.702508Z DEBUG CodegenFunction{name="wrong_coverage_2"}: kani_compiler::codegen_cprover_gotoc::utils::debug: Coverage::Counter(1) for assume_false.rs:2:1 - 6:2
[...]
2023-07-21T18:22:33.703152Z DEBUG CodegenFunction{name="wrong_coverage_2"}:CodegenStatement{statement=Coverage::Counter(1) for assume_false.rs:2:1 - 6:2}: kani_compiler::codegen_cprover_gotoc::codegen::statement: handling_statement stmt=Coverage::Counter(1) for assume_false.rs:2:1 - 6:2 kind=Coverage(Coverage { kind: Counter(1), code_region: Some(assume_false.rs:2:1 - 6:2) })
```

At that point, Kani APIs may not have been taken into account in the computation for coverage basic blocks, resulting in a single `CoverageStatement` for the whole function (this is the example with `assume(false)` followed by `assert!(...)`). But note that the check introduced here isn't the coverage statement we discussed earlier. It's possible that this `missing_definition` check is related to the profile missing from the compilation process in Kani.

In summary, some of these APIs aren't public which makes it difficult to use them, and it isn't clear to me what code we'd need to trim even if we do. The basic idea seems to be the same, but many steps seem focused on optimizing runtime. Another attempt to use the coverage instrumentation has failed. Do you still think it's worth exploring this for the first version?

zhassan-aws:
I see. Thanks for exploring this option @adpaco-aws. If we have no reason to believe that the proposed approach is different from the way rustc instruments coverage, then I'm OK with proceeding with the current plan. Once we implement it, we can conduct experiments that check for discrepancies between rustc's coverage and ours (e.g. by checking that lines that are reported as covered through rustc's instrumentation are reported as covered by Kani).

qinheping:
How do we report if a line is partially covered? That is, there are multiple instructions in the same line and only some of them are reachable.

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+
+**Users are not expected to consume this output directly.**
+Instead, coverage data is to be consumed by the [Kani VS Code extension](https://github.com/model-checking/kani-vscode-extension) and displayed as in the following picture:
+
+![Coverage reported on the Kani VS Code Extension](../images/0008/coverage-kani-vscode.png)
+
+How to activate and display coverage information in the extension is out of scope for this RFC.
+We will need to change our documentation (including the tutorial) to use this option if we ever remove `--visualize` in favor of this one.
+
+## Detailed Design
```
celinval, line 45: Will there be any chance on how we invoke CBMC? Are we running a full verification + coverage checks?

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+This option will be supplied by `kani-driver` when the `--coverage` option is selected.
+These coverage checks can be processed after regular verification postprocessing to obtain the coverage information, which can directly be passed on to the Kani VS Code Extension.
+
+Eventually, this could allow us to remove `cbmc-viewer` as a dependency.
+
+We anticipate missing locations being the main corner case for the coverage feature.
+However, coverage reported through `cbmc-viewer` also suffers from this issue.
+This will be mitigated by ignoring checks coming from unknown locations and propagating locations if possible whenever they're missing (as suggested in [this comment](https://github.com/model-checking/kani/issues/2048#issuecomment-1599680694)).
+
+## Rationale and alternatives
```
celinval, line 61: I think it would be nice to explore a bit more the why not fix the issues we've found so far with CBMC coverage. Another thing I believe you will have to include here eventually is the output format. I do think we could already add that the goal is to provide UX that is already familiar to users, instead of having our own (like CBMC viewer HTML report).

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+The main advantage of this design is that it'll give Kani full control over its coverage-related options.
+This implies that not only it'll be able to select the mechanism to collect coverage information, but also have different options to output coverage data.
+All of that fully integrated in Kani and tested through `compiletest`.
+In contrast, this means that Kani will become more complex, so the development and maintainance burden will be higher.
+
+This feature is expected to have a substantial impact for Kani.
+Users frequently ask for coverage-related options but Kani doesn't have one.
+That's because, some months ago, we noticed incorrect coverage results were being reported with the `--visualize` option, prompting us to disable the coverage reports emitted by `cbmc-viewer`.
+
+## Open questions
```
celinval, line 72: I think you are missing a big one here: Output format

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+
+Open questions:
+ * Do we want a `--coverage` option or a `coverage` subcommand? Coverage can be seen as a different workflow but also a verification option.
```
celinval, line 75: I think it depends whether verification will be performed at the same time.

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+ 2. Displaying region-based coverage information similar to [Rust's instrument-based code coverage](https://doc.rust-lang.org/rustc/instrument-coverage.html), which allows users to see coverage information for more granular parts of code (e.g., subconditions hit in a disjunctive condition).
+ 3. Adding new user-requested coverage formats such as GCOV [#1706](https://github.com/model-checking/kani/issues/1706) or LCOV [#1777](https://github.com/model-checking/kani/issues/1777).
+ 4. Enabling an option to run verification and coverage at the same time, so users can obtain both results at the same time.
+ 5. Performing optimization improvements to `kani-compiler` and its engines to speed up the data collected for coverage information.
+
+
+[^coverage-assertions]: Currently, we replace non-coverage assertions with assumptions so the execution is blocked instead of reporting a failure.
```
celinval, line 87: This should be part of the design section. If it's not clear whether this is something we want, please also add it to the open questions.

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+Because of that, we cannot report coverage runs as verification results.
+
+[^coverage-experiments]: We have experimented with different options for injecting coverage checks.
```
celinval, line 90: This should be part of the rationale and alternative

celinval (commented):
I think the goal of this RFC is not quite clear to me (neither, the path to stabilization).

BTW, I think it's OK for us to have a lot of open questions, it's just not clear to me what is an open question and what is a design decision at this point. For example, what is the coverage granularity that we are planning to provide? The design talks about statements, but the output is line based.

adpaco-aws:
> One question is: how does this work with multiple harnesses? Ideally, one would want a single coverage report where a line is considered covered if it is reachable from one or more harnesses. That way, one can focus on writing new harnesses that cover lines that are currently uncovered by any of the existing harnesses.

Good point! Yes, we should merge coverage results from all harnesses to produce a combined report. I've discussed it with @jaisnan and I think we're going to go with the following approach:
 1. New section for coverage results for each individual harness.
 2. Final section that combines the previous coverage results into one if we're running more than one harness.

adpaco-aws:
Update: I've tried to address most comments in the latest changes, but tomorrow I'll post individual replies to each of them.

Thank you all for the feedback!

adpaco-aws:
> > Do you have any feedback on the UX side?
> 
> One question is: how does this work with multiple harnesses? Ideally, one would want a single coverage report where a line is considered covered if it is reachable from one or more harnesses. That way, one can focus on writing new harnesses that cover lines that are currently uncovered by any of the existing harnesses.

This is now considered in the future possibilities section. Some coverage frameworks have a dedicated tool to combine coverage results from individual harnesses and I'd like Kani to do the same. However, I'm not sure how to distribute and in any case it's a nice-to-have which is why I'm adding it to that section (actually, @jaisnan has a prototype if I'm not mistaken).

adpaco-aws:
> Another question: how would we report coverage for functions in user code that are not reachable from a harness? As a user, I would expect the lines in those functions to be reported as uncovered, but the pruning done by the MIR linker may remove those functions altogether, and thus no checks would be generated for them at all.

We don't provide coverage results for unreachable code. In other words, we only report coverage for lines which contain at least one coverage check. The rest are essentially considered not covered.

adpaco-aws:
> How do we report if a line is partially covered? That is, there are multiple instructions in the same line and only some of them are reachable.

This is now addressed in the design section (postprocessing of coverage checks). We'll report three different types of coverage: `FULL` , `PARTIAL` and `NONE`.

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
+- **Feature Name:** Viewerless coverage information (`viewerless-coverage`)
```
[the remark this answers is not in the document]
adpaco-aws: Renamed to line coverage.

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+- **RFC PR:** <https://github.com/model-checking/kani/pull/2609>
+- **Status:** Under Review
+- **Version:** 0
+- **Proof-of-concept:** <https://github.com/model-checking/kani/pull/2609>
+
+-------------------
+
+## Summary
+
+An option in Kani to generate coverage reports without `cbmc-viewer`.
```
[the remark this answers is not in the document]
adpaco-aws: Done.

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+
+## User Impact
```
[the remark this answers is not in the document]
adpaco-aws: We now discuss the impact of a coverage feature across four different stages of a verification effort: learning, development, debugging and evaluation.

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+ * **The instrumentation performed** on the program. Eventually, this would allow us to report more precise coverage information, similar to [Rust's instrument-based code coverage](https://doc.rust-lang.org/rustc/instrument-coverage.html).
+ * **The format of the coverage report** to be generated. Similarly, this would allow us to generate coverage data in different formats (see [#1706](https://github.com/model-checking/kani/issues/1706) for GCOV, or [#1777](https://github.com/model-checking/kani/issues/1777) for LCOV). While technically this is also doable from `cbmc-viewer`'s output, development is likely to be faster this way.
+
+Moreover, producing coverage information directly from Kani is likely to increase speed of development and improve testing for coverage features, which translates into faster and more reliable coverage options for users.
+
+## User Experience
+
+For the first version, this experimental feature will be limited to produce a sequence of text lines as follows:
+```
+<file>, <line>, <status>
```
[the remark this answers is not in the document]
adpaco-aws: We're only including the data required for line coverage for now. I'm confident the format can be easily extended to include span information (i.e., start/end line/column).

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+![Coverage reported on the Kani VS Code Extension](../images/0008/coverage-kani-vscode.png)
+
+How to activate and display coverage information in the extension is out of scope for this RFC.
+We will need to change our documentation (including the tutorial) to use this option if we ever remove `--visualize` in favor of this one.
+
+## Detailed Design
+
+We will add a new unstable `--coverage` verification option to Kani.
+This will prevent `kani-driver` from reporting verification results as usual and instead output the coverage information described above[^coverage-assertions].
+It's likely that this will be done through a new output format that's not exposed to users.
```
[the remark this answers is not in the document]
adpaco-aws: It's gone.

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+Instead, coverage data is to be consumed by the [Kani VS Code extension](https://github.com/model-checking/kani-vscode-extension) and displayed as in the following picture:
+
+![Coverage reported on the Kani VS Code Extension](../images/0008/coverage-kani-vscode.png)
+
+How to activate and display coverage information in the extension is out of scope for this RFC.
+We will need to change our documentation (including the tutorial) to use this option if we ever remove `--visualize` in favor of this one.
+
+## Detailed Design
+
+We will add a new unstable `--coverage` verification option to Kani.
```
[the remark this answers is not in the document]
adpaco-aws: Added, thanks!

rfc/src/rfcs/0008-viewerless-coverage.md
adpaco-aws: I'd like to avoid an option with `cover` and `statement` since that's how we name the macro and it could be confusing at some point. So the option is still `--coverage-checks` in this version, but I'm open to other suggestions.

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+This will prevent `kani-driver` from reporting verification results as usual and instead output the coverage information described above[^coverage-assertions].
+It's likely that this will be done through a new output format that's not exposed to users.
+
+We will also add a new `--coverage-checks` option to `kani-compiler`, which will result in the injection of coverage checks before each Rust statement and terminator[^coverage-experiments].
+This option will be supplied by `kani-driver` when the `--coverage` option is selected.
+These coverage checks can be processed after regular verification postprocessing to obtain the coverage information, which can directly be passed on to the Kani VS Code Extension.
+
+Eventually, this could allow us to remove `cbmc-viewer` as a dependency.
+
+We anticipate missing locations being the main corner case for the coverage feature.
```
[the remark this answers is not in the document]
adpaco-aws: Moved. Also, we now enumerate and briefly discuss issues with `cbmc-viewer`.

rfc/src/rfcs/0008-viewerless-coverage.md
adpaco-aws: The last version includes a brief discussion on this, but I'm not sure how deep you want it to be. Can you please take a look?

rfc/src/rfcs/0008-viewerless-coverage.md
adpaco-aws: Done.

rfc/src/rfcs/0008-viewerless-coverage.md
adpaco-aws: The RFC now proposes doing verification and coverage at the same time. My intuition says this will too much for harder problems, but we need to perform a proper evaluation (included in "feedback to gather before stabilization").

rfc/src/rfcs/0008-line-coverage.md
```diff
@@ -0,0 +1,156 @@
...
+ * **Development:** Some users prefer coverage results to property-based results since they are easier to interpret.
+ * **Debugging:** Users may find coverage reports particularly helpful when inputs are over-constrained (missing some corner cases).
+ * **Evaluation:** Users can easily evaluate where and when more verification work is needed (some projects aim for 100% coverage).
+
+Moreover, adding this option directly to Kani, instead of relying on another tools, is likely to:
+ 1. Increase the speed of development
+ 2. Improve testing for coverage features
+Which translates into faster and more reliable coverage options for users.
+
+## User Experience
```
celinval, line 35: Can you add an initial paragraph explaining that this is still a work in progress? What about something like:

```suggestion
## User Experience

The goal is for Kani to generate code coverage report per harness in a well established format, such as LCOV, and possibly a summary in the output.
For now, we will focus on an interim solution that will enable us to assess our instrumentation and enable VSCode plugin integration.
```

celinval (commented):
This looks good! Thanks for all the changes. My main request is to make it clear that we still need to define a final UX as part of this RFC. Thanks!

rfc/src/rfcs/0008-viewerless-coverage.md
adpaco-aws: No changes, this is now said explicitly (verification + coverage). We'll find out how this affects performance (especially for hard benchmarks) later.

rfc/src/rfcs/0008-line-coverage.md
adpaco-aws: Thanks for the suggestion, I've added this with some minor changes.

rfc/src/rfcs/0008-line-coverage.md
```diff
@@ -0,0 +1,156 @@
...
+
+Nowadays, users can't easily obtain verification-based coverage reports in Kani.
+Generally speaking, coverage reports show which parts of the code under verification are covered and which are not.
+Because of that, coverage is often seen as a great metric to determine the quality of a verification effort.
+
+Moreover, some users prefer using coverage information for harness development and debugging.
+That's because coverage information provides users with more familiar way to interpret verification results.
+
+This RFC proposes adding a new option for verification-based line coverage reports to Kani.
+As mentioned earlier, we expect users to employ this coverage-related option on several stages of a verification effort:
```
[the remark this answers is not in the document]
adpaco-aws: Definitely! Done.

rfc/src/rfcs/0008-line-coverage.md
```diff
@@ -0,0 +1,156 @@
...
+One of the long-term goals for this feature is to provide a UX that is familiar for users.
+This is particularly relevant when talking about output formats.
+Some services and frameworks working with certain coverage output formats have become quite popular.
+
+However, this version doesn't consider common output formats (i.e., GCOV or LCOV) since coverage results will only be consumed by the Kani VS Code Extension at first.
+But other output formats will be considered in the future.
+
+## Open questions
+
+Open questions:
```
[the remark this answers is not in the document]
adpaco-aws: Done.

rfc/src/rfcs/0008-line-coverage.md
```diff
@@ -0,0 +1,156 @@
...
+ * Hidden from verification results.
+ * Postprocessed to produce coverage results.
+
+In the following, we describe the injection and postprocessing procedures to generate coverage results.
+
+#### Injection of Coverage Checks
+
+The injection of coverage checks will be done while generating code for basic blocks.
+This allows us to add one coverage check before each statement and terminator, which provides the most accurate results[^coverage-experiments].
+It's not completely clear how this compares to the coverage instrumentation done in the Rust compiler, but an exploration to use the compiler APIs revealed that they're quite similar[^coverage-api].
```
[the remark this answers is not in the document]
adpaco-aws: Just double-checked and constant propagation wasn't actually mentioned, just compiler optimizations. So I rewrote this bit in "Feedback to gather before stabilization":
> * Determine if there are cases for which coverage information is confusing for users (due to, e.g., constant propagation or other compiler optimizations).

rfc/src/rfcs/0008-line-coverage.md
```diff
@@ -0,0 +1,156 @@
...
+### Arquitecture
+
+We will add a new unstable `--coverage` verification option to Kani which will require `-Z line-coverage` until this feature is stabilized.
+We will also add a new `--coverage-checks` option to `kani-compiler`, which will result in the injection of coverage checks before each Rust statement and terminator[^coverage-experiments].
+This option will be supplied by `kani-driver` when the `--coverage` option is selected.
+These options will cause Kani to inject coverage checks during compilation and postprocess them to produce the coverage results sections described earlier.
+
+### Coverage Checks
+
+Coverage checks are a new class of checks similar to `cover` checks.
```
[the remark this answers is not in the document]
adpaco-aws: That's the least I could do! (done)

rfc/src/rfcs/0008-line-coverage.md
```diff
@@ -0,0 +1,156 @@
...
+Moreover, adding this option directly to Kani, instead of relying on another tools, is likely to:
+ 1. Increase the speed of development
+ 2. Improve testing for coverage features
+Which translates into faster and more reliable coverage options for users.
+
+## User Experience
+
+### High-level changes
+
+For the first version, this experimental feature will report verification results along coverage reports.
```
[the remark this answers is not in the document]
adpaco-aws: We now describe the "naive" output format as experimental and mention that the final format + UX is to be determined (as in other suggestions).

rfc/src/rfcs/0008-line-coverage.md
```diff
@@ -0,0 +1,156 @@
...
+ * **Evaluation:** Users can easily evaluate where and when more verification work is needed (some projects aim for 100% coverage).
+
+Moreover, adding this option directly to Kani, instead of relying on another tools, is likely to:
+ 1. Increase the speed of development
+ 2. Improve testing for coverage features
+Which translates into faster and more reliable coverage options for users.
+
+## User Experience
+
+### High-level changes
```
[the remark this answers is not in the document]
adpaco-aws: Good question! I feel like this particular subsection briefly describes the high-level changes regarding "the plain Kani experience" (from the POV of a users), so in my opinion that's OK. But it's not set in stone and we can definitely move to another place (not sure where).

rfc/src/rfcs/0008-viewerless-coverage.md
adpaco-aws: I think this comment hasn't been addressed. Do you want me to go in detail about these?

rfc/src/rfcs/0008-viewerless-coverage.md
```diff
@@ -0,0 +1,92 @@
...
+The status reported by the injected checks is sufficient to construct a coverage report.
+Note that this gives Kani control on both ends:
+ * **The instrumentation performed** on the program. Eventually, this would allow us to report more precise coverage information, similar to [Rust's instrument-based code coverage](https://doc.rust-lang.org/rustc/instrument-coverage.html).
+ * **The format of the coverage report** to be generated. Similarly, this would allow us to generate coverage data in different formats (see [#1706](https://github.com/model-checking/kani/issues/1706) for GCOV, or [#1777](https://github.com/model-checking/kani/issues/1777) for LCOV). While technically this is also doable from `cbmc-viewer`'s output, development is likely to be faster this way.
+
+Moreover, producing coverage information directly from Kani is likely to increase speed of development and improve testing for coverage features, which translates into faster and more reliable coverage options for users.
+
+## User Experience
```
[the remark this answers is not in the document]
adpaco-aws: This appears to have been addressed in the latest version (including your suggestions), but please let me know if you want it to be emphasized.

rfc/src/rfcs/0008-viewerless-coverage.md
celinval: Sure. This is internal to Kani anyways. Not a biggie

celinval (approved):

qinheping (approved):
> > How do we report if a line is partially covered? That is, there are multiple instructions in the same line and only some of them are reachable.
> 
> This is now addressed in the design section (postprocessing of coverage checks). We'll report three different types of coverage: `FULL` , `PARTIAL` and `NONE`.

LGTM. Thanks.

Issue #2617, opened 2023-07-23
Print Kani development version info


adpaco:
Requested feature: Print development version info (i.e., `git` commit, branch or status) in development builds.
Use case: Development and debugging.
Link to relevant documentation (Rust reference, Nomicon, RFC): N/A

PR #2619, opened 2023-07-23
Print Kani version

### LINKED ISSUE #2570: Output - Kani version

adpaco-aws:
### Description of changes: 

Dusting this off since it'd be great to have for next release.

This PR changes Kani so it prints its version and the invocation type/mode as
`Kani Rust Verifier <version> (<invocation>)` where `<invocation>` is either `standalone` or `cargo plugin`.

Please note that I'm leaving development version information for later in #2617. We can continue the discussion we started in #2571 here.

### Resolved issues:

Resolves #2570 

### Testing:

* How is this change tested? Two new tests.

### Checklist


kani-driver/build.rs
```diff
@@ -3,11 +3,28 @@
...
+    // Force this script to re-run if `kani-verifier/Cargo.toml` changed.
+    // Otherwise, we may not get the up-to-date version in `kani-driver`.
+    println!("cargo:rerun-if-changed=../Cargo.toml");
+
+    // Pass the `kani-verifier` version through the build environment.
+    println!("cargo:rustc-env=KANI_VERIFIER_VERSION={}", kani_verifier_version());
+}
+
+/// Retrieve the `kani-verifier` version using `cargo metadata`
+fn kani_verifier_version() -> String {
```
celinval, line 25: Shouldn't the `kani-driver` version always match `kani-verifier`? I don't understand why we are doing this.

kani-driver/src/main.rs
```diff
@@ -77,6 +82,11 @@ fn cargokani_main(input_args: Vec<OsString>) -> Result<()> {
         return assess::run_assess(session, assess::AssessArgs::default());
     }
 
+    if !session.args.common_args.quiet {
+        let kani_version = print_kani_version();
```
celinval, line 86: Shouldn't this method print the version? It has `print` in the name. :)

kani-driver/src/version.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use crate::{determine_invocation_type, InvocationType};
+
+const KANI_RUST_VERIFIER: &str = "Kani Rust Verifier";
+const KANI_VERIFIER_VERSION: &str = env!("KANI_VERIFIER_VERSION");
+
+/// Print Kani version. At present, this is only release version information.
```
celinval, line 9: This doesn't print though

kani-driver/src/version.rs
```diff
@@ -0,0 +1,30 @@
...
+/// Print Kani release version as `Kani Rust Verifier <version> (<invocation>)`
+/// where:
+///  - `<version>` is the `kani-verifier` version
+///  - `<invocation>` is `cargo plugin` if Kani was invoked with `cargo kani` or
+///    `standalone` if it was invoked with `kani`.
+fn kani_version_release() -> String {
+    let mut version_str = "Kani Rust Verifier ".to_string();
+    version_str.push_str(KANI_VERIFIER_VERSION);
+
+    let invocation_str = match determine_invocation_type(Vec::from_iter(std::env::args_os())) {
```
celinval, line 25: nit: you could get this as an argument.

celinval (commented):
Shouldn't we print the same blob for all sub-commands as well?

kani-driver/src/main.rs
adpaco-aws: True, it prints now.

kani-driver/src/version.rs
adpaco-aws: Done.

adpaco-aws:
Addressed all comments.

> Shouldn't we print the same blob for all sub-commands as well?

To be honest, this wasn't clear to me. Guessing it's because I'd really like `assess` to be a tool rather than a sub-command...

kani-driver/build.rs
adpaco-aws: I've opened #2626 to ensure these are the same versions. We just use the CARGO_PKG_VERSION now.

celinval (approved):
Thanks

kani-driver/src/version.rs
```diff
@@ -0,0 +1,34 @@
...
+    println!("{kani_version}");
+}
+
+/// Print Kani release version as `Kani Rust Verifier <version> (<invocation>)`
+/// where:
+///  - `<version>` is the `kani-verifier` version
+///  - `<invocation>` is `cargo plugin` if Kani was invoked with `cargo kani` or
+///    `standalone` if it was invoked with `kani`.
+fn kani_version_release(invocation_type: InvocationType) -> String {
+    let mut version_str = "Kani Rust Verifier ".to_string();
```
celinval, line 26: do you ever use this?

adpaco-aws: Good catch. I'm surprised the compiler/`clippy` didn't complain about this.

PR #2628, opened 2023-07-27
Delay writing metadata file (no UX impact)

celinval:
### Description of changes: 

Kani compiler will now only store KaniMetadata after compiling all harnesses. Before, we were storing before codegen in the first iteration of the compiler.

This will still allow us to generate metadata without actually performing codegen, if we ever implement a `kani list` subcommand. The metadata won't be stored though if Kani fails to codegen. However, we don't do anything extra with that file if the compilation fails.

This change is required for #2493 and contracts work. This will allow us to store information collected during code generation.

### Resolved issues:

Related to #2493

### Related RFC:
N/A

### Testing:

### Checklist


zhassan-aws:
Can you clarify in the description which compilation stage the writing of the metadata file was delayed to (and which stage it was currently written in)?

kani-compiler/src/kani_compiler.rs
```diff
@@ -386,6 +410,31 @@ impl Callbacks for KaniCompiler {
...
+        test_harnesses,
+    }
+}
+
+/// Extract the filename for the metadata file.
+fn metadata_output_path(tcx: TyCtxt) -> PathBuf {
+    let mut filename = tcx.output_filenames(()).output_path(OutputType::Object);
+    filename.set_extension(ArtifactType::Metadata);
+    filename
+}
```
feliperodri, line 436: No tests required for the these new functions?

kani-compiler/src/kani_compiler.rs
```diff
@@ -77,6 +78,12 @@ struct HarnessInfo {
     pub stub_map: Stubs,
 }
 
+#[derive(Clone, Debug)]
+struct CrateInfo {
+    pub name: String,
+    pub output_path: PathBuf,
+}
```
feliperodri, line 85: I'd add a small documentation/comment for `CrateInfo`.

kani-compiler/src/kani_compiler.rs
```diff
@@ -121,18 +129,17 @@ enum CompilationStage {
...
 }
 
 impl CompilationStage {
     pub fn is_init(&self) -> bool {
         matches!(self, CompilationStage::Init)
     }
-
-    pub fn is_done(&self) -> bool {
-        matches!(self, CompilationStage::Done)
-    }
```
feliperodri, line 135: Do we even use `is_init` too? I'd remove it.

feliperodri (approved):
LGTM.

kani-compiler/src/kani_compiler.rs
celinval: I could add a test for `generate_metadata`, but the second function depends on TyCtxt, which we can't mock

kani-compiler/src/kani_compiler.rs
celinval: We do

kani-compiler/src/kani_compiler.rs
```diff
@@ -121,18 +129,17 @@ enum CompilationStage {
         target_harnesses: Vec<HarnessId>,
         next_harnesses: Vec<Vec<HarnessId>>,
         all_harnesses: HashMap<HarnessId, HarnessInfo>,
+        crate_info: CrateInfo,
+    },
+    Done {
+        metadata: Option<(KaniMetadata, CrateInfo)>,
```
JustusAdam, line 135: Why the encoding with an optional metadata? From the other comments it seems this only occurs in like `--version`, e.g. if no actual compilation happens, so why not make it an explicit `CompilationStage::CompilationSkipped` variant?

celinval: Yeah, I thought about that, but we still two other modes where we do run codegen and we don't have a metadata (tests and pubfn). So this would still need to be optional.

Once we migrate them, I agree that we should just have a different state.

kani-compiler/src/kani_compiler.rs
celinval: I kept `is_init` for now. It is currently used in two different `if` conditions. Let me know if you have any strong reason for us to remove it at this point.

kani-compiler/src/kani_compiler.rs
celinval: Actually, I think we can add this now. Let me do that. Thanks

JustusAdam:
Also, not trying to rush (well maybe a bit) but I'm sorta waiting for this to get merged so I can finish the contracts rewrite that uses stubbing.

JustusAdam:
Nice

JustusAdam (approved):

PR #2643, opened 2023-07-30
Propogate backend options into goto-synthesizer

qinheping:
### Description of changes: 

`goto-synthesizer` should take the same backend options as `cbmc` to make sure that the checker we used to check loop-contracts candidates is the same as the checker we verify the final goto-binary.

CBMC backend options: https://github.com/diffblue/cbmc/blob/develop/src/cbmc/cbmc_parse_options.cpp#L993C1-L1000C58 including
1. solver options (already propagated)
2. `object-bits` option (propagated in this PR)
3. `array-uf` options (propagated in this PR)
4. string refine options (TODO)

### Resolved issues:

Resolves #ISSUE-NUMBER

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


kani-driver/src/call_goto_synthesizer.rs
```diff
@@ -35,7 +35,21 @@ impl KaniSession {
...
+        // goto-synthesizer should take the same backend options as cbmc.
+        // Backend options include
+        // 1. solver options
         self.handle_solver_args(&harness_metadata.attributes.solver, &mut args)?;
+        // 2. object-bits option
+        if let Some(object_bits) = self.args.cbmc_object_bits() {
+            args.push("--object-bits".into());
+            args.push(object_bits.to_string().into());
+        }
+        // 3. array-as-uninterpreted-functions options
```
zhassan-aws, line 47: Rather than handling every single backend option, when we call `cbmc`, we rely on `--cbmc-args` to pass other less-common options. Can we do something similar here?

zhassan-aws: What I meant is that `kani`/`cargo kani` currently have a `--cbmc-args` option, where anything appearing after this option gets passed to the `cbmc` command as is. This is to avoid adding a Kani option for each one of the possible options.

Can we rely on a similar mechanism, and pass options after `--cbmc-args` to the goto synthesizer as well?

qinheping: We don't need something like `synthesizer-args` for back-end options. If users want to specify some backend options for CBMC using `--cbmc-args`, this PR propagate the options to make sure those options are also specified for the synthesizer. Users should not specify backend options only for the synthesizer.

zhassan-aws: I understand, but if we'd like to only propagate backend options, I think we should create a new Kani option, e.g. `--cbmc-backend-args` whose arguments get passed to both `cbmc` and the goto synthesizer. Otherwise, we will have to handle all other backend options on the Kani side (this PR only handles `object-bits`, `--arrays-uf-always`, and `--arrays-uf-never`, but there are many other backend options, e.g. `--smt2`, `--bitwuzla`, `--refine-arrays`, `--refine-arithmetic`, etc.), so it might get messy.

The issue with adding a new option though is that we can't have two options with positional arguments (`--cbmc-args` and `--cbmc-backend-args`). So we'll need to figure out a way to address that.

qinheping: Alternatively, we can also let `goto-synthesizer` take `--cbmc-args`, pass all `cbmc-args` from Kani to `goto-synthesizer`, and let it to decide which are backend options. Does this approach sounds better to you?

zhassan-aws: Yes, that could work too. Would it require changing the `goto-synthesizer` code significantly?

qinheping: @zhassan-aws , `goto-synthesizer` can now take CBMC options as arguments as this PR merged. I also updated this PR to let goto-synthesizer accept all CBMC arguments.

zhassan-aws:
Looks good. We'll need to wait for a new CBMC release that includes https://github.com/diffblue/cbmc/pull/7900 before merging though, right?

qinheping:
> Looks good. We'll need to wait for a new CBMC release that includes [diffblue/cbmc#7900](https://github.com/diffblue/cbmc/pull/7900) before merging though, right?

Yes, we should merge this PR after CBMC 5.94.0 published.

zhassan-aws (approved):

celinval (dismissed):
Just blocking this until we release 0.39.0. Thank you for your patience. :smiley:

Issue #2595, opened 2023-07-10
Refactor contract attribute validation into `check_attributes`


JustusAdam:
In the spirit of getting features working first parsing contract-related attributes (`requires`, `ensures`, `assigns`, etc) validates as they are parsed and resolved. 

However to provide better (early) feedback to users it should validate them first in `check_attribute`.

PR #2655, opened 2023-08-03
Everything needed for requires/ensures checking

### LINKED ISSUE #2596: Unstable gating for function contracts
Function contracts are considered unstable and should be appropriately gated

JustusAdam:
### Description of changes: 

Adds simple function contracts in the form of `requires` and `ensures` clauses to Kani. 

A contract for e.g. a function `f` can be checked by creating a harness and adding the `proof_for_contract(f)` annotation.

### Resolved issues:

Resolves #2596, #2595 

### Related RFC:

#2620.

### Call-outs:

There are some auxiliary changes here that are not strictly necessary.

* The testing framework `compiletest` has been changed so that multiple "expected" tests can be located in the same directory. In this case the runner expects there to be for each `testfile.rs` a `testfile.expected` file
* `KaniAttributes` have been encapsulated into a struct that is exported but opaque. This is not strictly necessary for this PR, but in later contract work the attributes will need to be requeried later in the compiler which would be inefficient with the way they were previously used. This leaves two options: Expose the internal attribute map, or expose the attribute map in encapsulated fashion (which is what is done here).
* The `sysroot` module in `kani_macros` has been refactored into a new file. This is because it is significantly larger now and mostly because my editor chokes on scrolling through it otherwise (because it grays it out which I guess is bad for the rendering engine).

### Testing:

* How is this change tested? A suite of "expected" test cases is provided in the "function-contract" directory

### Checklist


celinval (commented):
Thanks for doing this!

kani-compiler/src/kani_compiler.rs
```diff
@@ -419,7 +422,11 @@ impl Callbacks for KaniCompiler {
     ) -> Compilation {
         if self.stage.is_init() {
             self.stage = rustc_queries.global_ctxt().unwrap().enter(|tcx| {
-                check_crate_items(tcx, self.queries.lock().unwrap().ignore_global_asm);
+                // Extra block necessary to unlock queries
+                {
+                    let queries = self.queries.lock().unwrap();
+                    check_crate_items(tcx, &queries);
```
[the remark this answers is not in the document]
JustusAdam: Absolutely. Happy to change it.

kani-driver/src/args/mod.rs
```diff
@@ -344,6 +346,21 @@ impl VerificationArgs {
             Some(Some(x)) => Some(x), // -j=x
         }
     }
+
+    /// Public access to the value set for the `--exact` flag
+    pub fn exact(&self) -> bool {
```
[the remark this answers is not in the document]
JustusAdam: Originally the contracts cli had some implications on `exact`. Happy to revert.

library/kani_macros/src/lib.rs
```diff
@@ -7,6 +7,7 @@
 // downstream crates to enable these features as well.
 // So we have to enable this on the commandline (see kani-rustc) with:
 //   RUSTFLAGS="-Zcrate-attr=feature(register_tool) -Zcrate-attr=register_tool(kanitool)"
+#![feature(let_chains, proc_macro_diagnostic, box_patterns)]
```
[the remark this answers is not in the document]
JustusAdam: 😞 okay. They are just so useful. The only one I think we should keep for sure is `proc_macro_diagnostic`, because it allows us to throw clean errors in the macro (e.g. the `thing.span();unwrap().err()` stuff)

library/kani_macros/src/sysroot.rs
```diff
@@ -0,0 +1,457 @@
...
+                qself: None,
+                path: syn::Path {
+                    leading_colon: None,
+                    segments: [PathSegment {
+                        ident: match arg {
+                            FnArg::Receiver(_) => Ident::new("self", Span::call_site()),
+                            FnArg::Typed(PatType { pat, .. }) => match &**pat {
+                                Pat::Ident(PatIdent { ident, .. }) => ident.clone(),
+                                _ => {
+                                    pat.span().unwrap().error("Unexpected pattern").emit();
```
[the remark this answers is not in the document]
JustusAdam: Yes exactly. I'll change this to a more descriptive error.

JustusAdam:
Okay the pivot has happened. The generation algorithm has been changed and should now not exhibit the weird drop issue.

There is still a lingering drop issue where if you move a value with heap data such that it's `drop` gets called early it causes a double-free.

library/kani_macros/src/sysroot.rs
```diff
@@ -0,0 +1,457 @@
...
+/// "check_<fn_name>_<fn_hash>"]
+///
+/// Each clause (requires or ensures) creates its own check function that calls
+/// the prior check function inside. The innermost check function calls a copy
+/// of the originally decorated function. It is a copy, because the compiler
+/// later replaces all invocations of the original function with this check and
+/// that would also apply to the inner check. We need that to be the untouched
+/// function though so we make a copy that will survive the replacement from the
+/// compiler.
+fn requires_ensures_alt(attr: TokenStream, item: TokenStream, is_requires: bool) -> TokenStream {
```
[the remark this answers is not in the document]
JustusAdam: After the latest major changes this function got a lot smaller (in terms of code) and I added much more comments so let me know if you still think it should be changed.

library/kani_macros/src/sysroot.rs
```diff
@@ -0,0 +1,457 @@
...
+    /// occurrences, but not the binding location. This is because our
+    /// [`visit_expr_path_mut`] is scope-unaware.
+    fn visit_pat_ident_mut(&mut self, i: &mut syn::PatIdent) {
+        if let Some(new) = self.0.get(&i.ident) {
+            i.ident = new.clone();
+        }
+    }
+}
+
+/// Does the provided path have the same chain of identifiers as `mtch` (match)
```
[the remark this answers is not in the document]
JustusAdam: It's a convenient way to find out if attributes have the shape of the one's we emit, e.g. `kanitool::checked_with`. Now I only need to call `matches_path(attribute.path, &["kanittool", "checked_with"])`

library/kani_macros/src/sysroot.rs
```diff
@@ -0,0 +1,457 @@
...
+        self.0.insert(i.ident.clone());
+        syn::visit::visit_pat_ident(self, i)
+    }
+    fn visit_receiver(&mut self, _: &'ast syn::Receiver) {
+        self.0.insert(Ident::new("self", proc_macro2::Span::call_site()));
+    }
+}
+
+/// Applies the contained renaming to everything visited.
+struct Renamer<'a>(&'a HashMap<Ident, Ident>);
```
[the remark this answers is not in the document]
JustusAdam: better?

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -325,6 +325,40 @@ impl<'tcx> GotocHook<'tcx> for MemCmp {
     }
 }
 
+/// A builtin that is essentially a C-style dereference operation. Takes in a
+/// `&T` reference and returns a `T` (like clone would but without cloning).
```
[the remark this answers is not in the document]
JustusAdam: Added. Also added that this necessitated `mem::forget`

library/kani_macros/src/sysroot.rs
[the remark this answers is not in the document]
celinval: I got that after I checked how this is being used. An example would've been helpful, but not a big deal. I guess there is no easy way to do this in `syn` today? Is it hard to build a Path from segments and compare the two paths?

library/kani/src/lib.rs
```diff
@@ -65,6 +65,20 @@ pub fn assume(cond: bool) {
     assert!(cond, "`kani::assume` should always hold");
 }
 
+/// If the `premise` is true, so must be the `conclusion`
```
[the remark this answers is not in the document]
celinval: This will be user visible, right? So I really recommend a bit more documentation.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -37,66 +54,379 @@ impl KaniAttributeKind {
...
+                self.tcx,
+                self.tcx.parent_module_from_def_id(self.item.expect_local()),
+                name.as_str(),
+            );
+            match resolved {
+                Err(e) => {
+                    self.tcx.sess.span_err(
+                        target.span,
+                        format!(
+                            "Sould not resolve replacement function {} because {e}",
```
celinval, line 152:
```suggestion
                            "Failed to resolve replacement function {} because {e}",
```

kani-compiler/src/kani_middle/attributes.rs
celinval, line 1: Can you please create a separate PR for this refactoring? Thanks

tools/compiletest/src/runtest.rs
celinval, line 1: This is another one that should be a separate PR. It also makes it easier for us to split the reviewer load, since it's much easier to have different reviewer per PR.

celinval (commented):
@JustusAdam can you please revert the changes to `kani_macro/lib.rs` and move the refactoring of attributes to a separate PR. It will make reviewing this PR much easier.

Even though the compiletest changes should've be a different PR, at least the changes are fairly independent from everything else, and small enough. I've reviewed them already, so I'm OK keeping them here.

library/kani_macros/src/sysroot.rs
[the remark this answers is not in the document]
celinval: Just as a matter of organization. I would say it is more readable if you create a composite of functions, instead of having an argument that changes what the function produces. Although there is a bit more redundancy, I believe it's much easier to read something like:
```rust
fn requires(/* args */) {
  let Some(check_fn) = get_check_fn() else { /* original */ return };
  let new_body = /* requires logic */;
  check_fn.finalize(new_body);
}

fn ensures(/* args */) {
  let Some(check_fn) = get_check_fn() } else { /* original */ return };
  let new_body = /* ensures logic */;
  check_fn.finalize(new_body);
}
```

celinval (commented):
I was reviewing the tests. They are looking good, thanks for putting the effort.

I think there are a few things that we are not covering yet, but we can leave them for a follow up PR. Things like:
- Multiple requires and ensures
- Contract doesn't get checked in regular harnesses
- Only target contract is checked in the `proof_for_contract` harnesses.

library/kani_macros/src/sysroot.rs
```diff
@@ -0,0 +1,550 @@
...
+///
+/// ```rs
+/// #[kanitool::checked_with = "div_check_965916"]
+/// fn div(dividend: u32, divisor: u32) -> u32 { dividend / divisor }
+///
+/// #[allow(dead_code)]
+/// #[allow(unused_variables)]
+/// #[kanitool::is_contract_generated(check)]
+/// fn div_check_965916(dividend: u32, divisor: u32) -> u32 {
+///     let dividend_renamed = kani::untracked_deref(&dividend);
```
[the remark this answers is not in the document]
JustusAdam: Neither. It's done for every argument. The documentation should adequately expand on that now hopefully.

library/kani_macros/src/sysroot.rs
```diff
@@ -0,0 +1,550 @@
...
+        .collect::<HashMap<_, _>>();
+
+    let mut ident_rewriter = Renamer(&arg_idents);
+    ident_rewriter.visit_expr_mut(attr);
+    arg_idents
+}
+
+/// The main meat of handling requires/ensures contracts.
+///
+/// Generates a `check_<fn_name>_<fn_hash>` function that assumes preconditions
```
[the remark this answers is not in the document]
JustusAdam: Yes. I've expanded on this a bit more in the documentation now.

tools/compiletest/src/runtest.rs
JustusAdam: Done #2671

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -99,6 +179,20 @@ impl<'tcx> KaniAttributes<'tcx> {
                     ),
                 );
             }
+            if kind.is_function_contract_api() && !queries.function_contracts_enabled() {
```
celinval, line 182: I have mixed feelings about failing the compilation if a function has a contract annotation and contracts is disabled.

I believe we should only fail if the contract is being used for verification (i.e.: either there's a `[kani::proof_for_contracts]` or `[kani::stub_verified]`.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -83,10 +113,60 @@ impl<'tcx> KaniAttributes<'tcx> {
...
+    pub fn checked_with(&self) -> Option<Symbol> {
+        self.expect_maybe_one(KaniAttributeKind::CheckedWith)
+            .map(|target| expect_key_string_value(self.tcx.sess, target))
+    }
+
     /// Check that all attributes assigned to an item is valid.
     /// Errors will be added to the session. Invoke self.tcx.sess.abort_if_errors() to terminate
     /// the session and emit all errors found.
-    pub(super) fn check_attributes(&self) {
+    pub(super) fn check_attributes(&self, queries: &QueryDb) {
```
celinval, line 169: I was wondering if this check should be applied to `check_unstable_features` instead.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -185,29 +294,65 @@ impl<'tcx> KaniAttributes<'tcx> {
...
+                    KaniAttributeKind::CheckedWith
+                    | KaniAttributeKind::ReplacedWith
+                    | KaniAttributeKind::IsContractGenerated => {
+                        todo!("Contract attributes are not supported on proofs")
+                    }
+                };
+                harness
+            },
+        );
+        let current_module = self.tcx.parent_module_from_def_id(local_id);
```
celinval, line 332: Can we do this when we match the `KaniAttributeKind::ProofForContract` instead? Can you please also export this to a function that returns the stub pair?

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -258,6 +473,27 @@ pub fn test_harness_name(tcx: TyCtxt, def_id: DefId) -> String {
     parse_str_value(&marker).unwrap()
 }
 
+/// Expect the contents of this attribute to be of the format #[attribute =
+/// "value"] and return the `"value"`
+pub fn expect_key_string_value(sess: &Session, attr: &Attribute) -> rustc_span::Symbol {
```
celinval, line 478:
```suggestion
fn expect_key_string_value(sess: &Session, attr: &Attribute) -> rustc_span::Symbol {
```

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -258,6 +473,27 @@ pub fn test_harness_name(tcx: TyCtxt, def_id: DefId) -> String {
...
+    let span = attr.span;
+    let AttrArgs::Eq(_, it) = &attr.get_normal_item().args else {
+        sess.span_fatal(span, "Expected attribute of the form #[attr = \"value\"]")
```
celinval, line 481: Please delay aborting the session while validating the attributes. You should return `Result<Symbol, ErrorGuaranteed>` instead.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -228,6 +373,74 @@ impl<'tcx> KaniAttributes<'tcx> {
...
+        is_prohibited: fn(ty::Ty<'tcx>) -> bool,
+        r#where: &'static str,
+        what: &'static str,
+    }
+
+    impl<'tcx> TypeVisitor<TyCtxt<'tcx>> for NoMutPtr<'tcx> {
+        fn visit_ty(&mut self, t: ty::Ty<'tcx>) -> std::ops::ControlFlow<Self::BreakTy> {
+            use ty::TypeSuperVisitable;
+            if (self.is_prohibited)(t) {
+                self.tcx.sess.span_err(self.span, format!("{} contains a {}pointer ({t:?}). This is prohibited for functions with contracts, as they cannot yet reason about the pointer behavior.", self.r#where, self.what));
```
celinval, line 395: you can add a note to make this error more user friendly.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -228,6 +373,74 @@ impl<'tcx> KaniAttributes<'tcx> {
     }
 }
 
+/// A basic check that ensures a function with a contract does not receive
+/// mutable pointers in its input and does not return raw pointers of any kind.
+///
+/// This is a temporary safety measure because contracts cannot yet reason
+/// about the heap.
+fn check_is_contract_safe(tcx: TyCtxt, item: DefId) {
```
celinval, line 381: How do you know if a generic parameter is not specialized into a pointer?

celinval: What if we move this function to either `contracts.rs` or to `kani_middle/mod.rs` and change the input to be a `MonoItem`? Let's say something like:
```rust
fn check_is_contract_safe(tcx: TyCtxt, item: MonoItem)
```

you could then invoke this function from `check_reachable_item` [here](https://github.com/model-checking/kani/blob/596bc5085ac036f02459dd77823f8b6df27fbea1/kani-compiler/src/kani_middle/mod.rs#L68C11-L68C11).

library/kani/src/lib.rs
```diff
@@ -65,6 +65,29 @@ pub fn assume(cond: bool) {
...
+///
+/// Note that boolean operators (such as `||`) are evaluated lazily by Rust.
+/// This function is not and both conditions will be evaluated always. As a
+/// reult this function is not intended to be used in regular code. Instead it
```
celinval, line 72:
```suggestion
/// result this function is not intended to be used in regular code. Instead it
```

library/kani/src/lib.rs
```diff
@@ -65,6 +65,29 @@ pub fn assume(cond: bool) {
...
+/// is intended to make implications in a function function contract
```
celinval, line 73:
```suggestion
/// is intended to make implications in a function contract
```

library/kani/src/lib.rs
celinval: Wouldn't the performance be worse for verification too?

tests/expected/function-contract/arbitrary_requires_pass.expected
```diff
@@ -0,0 +1 @@
+VERIFICATION:- SUCCESSFUL
```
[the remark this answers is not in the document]
celinval: Sure. Thanks

celinval (commented):
I still have do another pass on the macro logic. Thanks @JustusAdam

library/kani_macros/src/sysroot.rs
```diff
@@ -0,0 +1,550 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+use std::collections::{HashMap, HashSet};
+
+use super::*;
```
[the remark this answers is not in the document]
JustusAdam: Has been removed

kani-compiler/src/kani_middle/attributes.rs
JustusAdam: We talked about this today and if I recall we decided this would be an error, right?

kani-compiler/src/kani_middle/attributes.rs
JustusAdam: I could but then I need to change the handling of `stub`, because that overrides `attrs.stubs`. If you're ok with that I can move it. 

Beyond that the main reason this is extra is because we will need to treat replacement in essentially the same way. In particular the second `map` is common. If I refactor it into a function for `proof_for_contract` it won't be obvious that the code is the same. 

I could factor the whole "get stubbing for contracts stuff" into a separate function though if you like without obfuscating the shared code?

library/kani/src/lib.rs
JustusAdam: I don't think so? I don't think the verification performs any sort of short circuiting of boolean expressions. Or at least not in such a way that using control flow would make a difference. I'm not sure though. I can remove it if you like.

kani-compiler/src/kani_middle/attributes.rs
JustusAdam: This should also mean it automatically only applies if the contracts are reachable I think.

library/kani/src/lib.rs
celinval: It does. Here is a dummy example of using eager imply:
```rust
pub fn implies(premise: bool, conclusion: bool) -> bool {
    !premise || conclusion
}

#[kani::proof]
fn check_eager() {
    let arg1 = false;
    let arg2 = || {
        let mut heads = false;
        loop {
            heads = !heads;
            if kani::any() {
                break
            }
        }
        heads
    };
    assert!(implies(arg1, arg2()));
}
```

and the same example but using a lazy implies:
```rust
pub fn lazy_implies<F>(premise: bool, mut conclusion: F) -> bool where F: FnMut() -> bool {
    !premise || conclusion()
}


#[kani::proof]
fn check_lazy() {
    let arg1 = false;
    let arg2 = || {
        let mut heads = false;
        loop {
            heads = !heads;
            if kani::any() {
                break
            }
        }
        heads
    };
    assert!(lazy_implies(arg1, arg2));
}
```
The first harness won't terminate while the second verifies in less than a second.

JustusAdam: That surprises me. What should we do about it? I could remove `implies` and force people to use the expanded version instead. I can also add syntactic sugar for it that expands it out.

JustusAdam:
That last test is an out-of-disk error that I don't believe these changes are responsible for.

library/kani/src/lib.rs
celinval: I would remove it for now and create a separate PR with a more thorough performance analysis to see which one performed better in different scenarios

celinval:
> That last failing check is an out-of-disk error that I don't believe these changes are responsible for.

I'm looking into that. I'll probably fix that as part of #2703

library/kani_macros/src/sysroot.rs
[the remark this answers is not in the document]
JustusAdam: Hmm, I guess I could try building them as `lazy_static`s  instead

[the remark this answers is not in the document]
JustusAdam: Actually I don't know if comparing two `Path`s does the right thing. They have a bunch of `Span`s inside of them and I don't know how they interact with `Eq`.

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -0,0 +1,437 @@
...
+    if matches!(function_state, ContractFunctionState::Untouched) {
+        // We are the first time a contract is handled on this function, so
+        // we're responsible for
+        //
+        // 1. Generating a name for the check function
+        // 2. Emitting the original, unchanged item and register the check
+        //    function on it via attribute
+        // 3. Renaming our item to the new name
+        // 4. And (minor point) adding #[allow(dead_code)] and
+        //    #[allow(unused_variables)] to the check function attributes
```
feliperodri, line 330:
```suggestion
        // we're responsible for:
        // 1. Generating a name for the check function;
        // 2. Emitting the original, unchanged item and register the check
        //    function on it via attribute;
        // 3. Renaming our item to the new name;
        // 4. And (minor point) adding #[allow(dead_code)] and
        //    #[allow(unused_variables)] to the check function attributes.
```

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        let check_fn_name = identifier_for_generated_function(item_fn, "check", a_short_hash);
+
+        // Constructing string literals explicitly here, because if we call
+        // `stringify!` in the generated code that is passed on as that
+        // expression to the next expansion of a contract, not as the
+        // literal.
```
feliperodri, line 337: I don't understand that comment. Could you elaborate a bit more?

tests/expected/function-contract/arbitrary_requires_fail.expected
```diff
@@ -0,0 +1,4 @@
+assertion\
+- Status: FAILURE\
+- Description: "attempt to divide by zero"\
+arbitrary_requires_fail.rs:7:5
```
feliperodri, line 4: There are several missing lines, could you do a pass and fix all of them?

tests/expected/function-contract/attribute_no_complain.expected
```diff
@@ -0,0 +1 @@
+VERIFICATION:- SUCCESSFUL
```
feliperodri, line 1: Maybe you should combined these little examples that should succeed into a single file. Also, fix the missing lines.

tests/expected/function-contract/gcd_failure_code.expected
```diff
@@ -0,0 +1,9 @@
+gcd.assertion\
+- Status: FAILURE\
+- Description: "result != 0 && x % result == 0 && y % result == 0"\
+in function gcd
+
+Failed Checks: result != 0 && x % result == 0 && y % result == 0
+
+VERIFICATION:- FAILED
+
```
feliperodri, line 9: extra line, could you do a pass to clean up the tests?

tests/expected/function-contract/gcd_failure_code.rs
```diff
@@ -0,0 +1,72 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
+type T = u8;
+
+/// Euclid's algorithm for calculating the GCD of two numbers
```
feliperodri, line 6: Could you add a link where did you extra it from? This might not be necessary here because it probably comes from the past internship. If that's true, ignore it and keep it as note for future tests.

tests/expected/function-contract/out.txt
```diff
@@ -0,0 +1,258 @@
+Kani Rust Verifier 0.34.0 (standalone)
```
feliperodri, line 1: you should delete this file.

tests/expected/function-contract/prohibit-pointers/hidden.expected
```diff
@@ -0,0 +1 @@
+error: This argument contains a mutable pointer (*mut u32). This is prohibited for functions with contracts, as they cannot yet reason about the pointer behavior.
```
feliperodri, line 1: Beautiful.

feliperodri (changes requested):
Well done! Only minor comments over documentation and clean ups.

tests/expected/function-contract/attribute_no_complain.expected
JustusAdam: I'm not sure that's a good idea. I don't think we have a good way to distinguish to which example each output line corresponds to if we combine them and I don't really see the benefit in combining it either.

tests/expected/function-contract/prohibit-pointers/hidden.expected
JustusAdam: Thanks :)

JustusAdam (commented):
@feliperodri Thanks for the review. I mostly just adopted your changes but a few I voiced objections to in the comments.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -37,10 +51,31 @@ impl KaniAttributeKind {
...
         }
     }
+
+    /// Is this an "active" function contract attribute? This means it is is
+    /// part of the function contract interface *and* it implies that a contract
+    /// will be used (stubbed or checked) in some way, thus requiring that the
+    /// user activate the unstable feature.
+    ///
+    /// If we find an "inactive" contract attribute we chose not to error,
+    /// because it wouldn't have any effect anyway.
```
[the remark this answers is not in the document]
feliperodri: Let's track this on the issue and tackle it on a separate PR if necessary.

tests/expected/function-contract/attribute_no_complain.expected
feliperodri: We could track successful property from different harness using the Kani output. For instance,
```
Check 2: dig_deeper.assertion.1
         - Status: SUCCESS
         - Description: "assertion failed: ptr(input) as usize == 10"
         - Location: foreign_functions.rs:34:9 in function dig_deeper
```
Here, we have location and property. That should be enough to differ them using regex in the `*.expected` files. Am I missing something? In any case, we don't need to block this PR on this. It's just a clean up.

JustusAdam: `*.expected` files don't support regex. 

Also @celinval specifically requested not to use locations in `*.expected` files. They used to be in there and she requested I remove them which I think was sensible because at least the file path would come out slightly differently depending on how you called the test binary.

Similarly relying on the check number or property type is, in my opinion, brittle.

feliperodri (approved):

Issue #2587, opened 2023-07-07
Stub of extern functions including C-FFI


celinval:
Requested feature: Users should be able to stub an extern function.
Use case: Enable users to verify their code when the body of a function is missing.
Link to relevant documentation (Rust reference, Nomicon, RFC):

Test case: See https://github.com/model-checking/kani/issues/2423#issuecomment-1625496018 for an example.

PR #2658, opened 2023-08-03
Support for stubbing out foreign functions

feliperodri:
### Description of changes: 

Support for stubbing out foreign functions.

### Resolved issues:

Resolves https://github.com/model-checking/kani/issues/2587.

### Call-outs:

N/A.

### Testing:

* How is this change tested? Regression tests.

### Checklist

### DIFF
--- a/kani-compiler/src/kani_middle/provide.rs
+++ b/kani-compiler/src/kani_middle/provide.rs
@@ -55,7 +55,9 @@ fn run_kani_mir_passes<'tcx>(
     body: &'tcx Body<'tcx>,
 ) -> &'tcx Body<'tcx> {
     tracing::debug!(?def_id, "Run Kani transformation passes");
-    stubbing::transform(tcx, def_id, body)
+    let mut transformed_body = stubbing::transform(tcx, def_id, body);
+    stubbing::transform_extern_functions(tcx, &mut transformed_body);
+    tcx.arena.alloc(transformed_body)
 }
 
 /// Runs a reachability analysis before running the default
--- a/kani-compiler/src/kani_middle/provide.rs
+++ b/kani-compiler/src/kani_middle/provide.rs
@@ -56,7 +56,7 @@ fn run_kani_mir_passes<'tcx>(
 ) -> &'tcx Body<'tcx> {
     tracing::debug!(?def_id, "Run Kani transformation passes");
     let mut transformed_body = stubbing::transform(tcx, def_id, body);
-    stubbing::transform_extern_functions(tcx, &mut transformed_body);
+    stubbing::transform_foreign_functions(tcx, &mut transformed_body);
     tcx.arena.alloc(transformed_body)
 }
 
--- a/kani-compiler/src/kani_middle/stubbing/transform.rs
+++ b/kani-compiler/src/kani_middle/stubbing/transform.rs
@@ -12,7 +12,10 @@ use lazy_static::lazy_static;
 use regex::Regex;
 use rustc_data_structures::fingerprint::Fingerprint;
 use rustc_hir::{def_id::DefId, definitions::DefPathHash};
-use rustc_middle::{mir::Body, ty::TyCtxt};
+use rustc_middle::mir::{
+    interpret::ConstValue, visit::MutVisitor, Body, ConstantKind, Location, Operand,
+};
+use rustc_middle::ty::{self, TyCtxt};
 
 /// Returns the `DefId` of the stub for the function/method identified by the
 /// parameter `def_id`, and `None` if the function/method is not stubbed.
@@ -23,18 +26,43 @@ pub fn get_stub(tcx: TyCtxt, def_id: DefId) -> Option<DefId> {
 
 /// Returns the new body of a function/method if it has been stubbed out;
 /// otherwise, returns the old body.
-pub fn transform<'tcx>(
-    tcx: TyCtxt<'tcx>,
-    def_id: DefId,
-    old_body: &'tcx Body<'tcx>,
-) -> &'tcx Body<'tcx> {
+pub fn transform<'tcx>(tcx: TyCtxt<'tcx>, def_id: DefId, old_body: &'tcx Body<'tcx>) -> Body<'tcx> {
     if let Some(replacement) = get_stub(tcx, def_id) {
         let new_body = tcx.optimized_mir(replacement).clone();
         if check_compatibility(tcx, def_id, old_body, replacement, &new_body) {
-            return tcx.arena.alloc(new_body);
+            return new_body;
         }
     }
-    old_body
+    old_body.clone()
+}
+
+struct ExternFunctionTransformer<'tcx> {
+    tcx: TyCtxt<'tcx>,
+    body: Body<'tcx>,
+}
+
+impl<'tcx> MutVisitor<'tcx> for ExternFunctionTransformer<'tcx> {
+    fn tcx(&self) -> TyCtxt<'tcx> {
+        self.tcx
+    }
+
+    fn visit_operand(&mut self, operand: &mut Operand<'tcx>, _location: Location) {
+        let func_ty = operand.ty(&self.body, self.tcx);
+        if let ty::FnDef(reachable_function, generics) = *func_ty.kind() {
+            if let Some(stub) = get_stub(self.tcx, reachable_function) {
+                let Operand::Constant(fn_def) = operand else { unreachable!() };
+                fn_def.literal = ConstantKind::from_value(
+                    ConstValue::ZeroSized,
+                    self.tcx.type_of(stub).subst(self.tcx, generics),
+                );
+            }
+        }
+    }
+}
+
+pub fn transform_extern_functions<'tcx>(tcx: TyCtxt<'tcx>, body: &mut Body<'tcx>) {
+    let mut visitor = ExternFunctionTransformer { tcx, body: body.clone() };
+    visitor.visit_body(body);
 }
 
 /// Checks whether the stub is compatible with the original function/method: do
--- a/kani-compiler/src/kani_middle/stubbing/transform.rs
+++ b/kani-compiler/src/kani_middle/stubbing/transform.rs
@@ -36,17 +36,15 @@ pub fn transform<'tcx>(tcx: TyCtxt<'tcx>, def_id: DefId, old_body: &'tcx Body<'t
     old_body.clone()
 }
 
-/// Tranverse `body` searching for calls to foreing functions and, whevever there is
+/// Traverse `body` searching for calls to foreing functions and, whevever there is
 /// a stub available, replace the call to the foreign function with a call
 /// to its correspondent stub. This happens as a separate step because there is no
 /// body available to foreign functions at this stage.
 pub fn transform_foreign_functions<'tcx>(tcx: TyCtxt<'tcx>, body: &mut Body<'tcx>) {
-    let mut visitor = if let Some(stub_map) = get_stub_mapping(tcx) {
-        ForeignFunctionTransformer { tcx, body: body.clone(), stub_map }
-    } else {
-        ForeignFunctionTransformer { tcx, body: body.clone(), stub_map: HashMap::default() }
-    };
-    visitor.visit_body(body);
+    if let Some(stub_map) = get_stub_mapping(tcx) {
+        let mut visitor = ForeignFunctionTransformer { tcx, body: body.clone(), stub_map };
+        visitor.visit_body(body);
+    }
 }
 
 struct ForeignFunctionTransformer<'tcx> {
@@ -68,8 +66,8 @@ impl<'tcx> MutVisitor<'tcx> for ForeignFunctionTransformer<'tcx> {
         if let ty::FnDef(reachable_function, generics) = *func_ty.kind() {
             if self.tcx.is_foreign_item(reachable_function) {
                 if let Some(stub) = self.stub_map.get(&reachable_function) {
-                    let Operand::Constant(fn_def) = operand else { unreachable!() };
-                    fn_def.literal = ConstantKind::from_value(
+                    let Operand::Constant(function_definition) = operand else { return; };
+                    function_definition.literal = ConstantKind::from_value(
                         ConstValue::ZeroSized,
                         self.tcx.type_of(stub).subst(self.tcx, generics),
                     );
--- a/kani-compiler/src/kani_middle/stubbing/transform.rs
+++ b/kani-compiler/src/kani_middle/stubbing/transform.rs
@@ -12,8 +12,10 @@ use lazy_static::lazy_static;
 use regex::Regex;
 use rustc_data_structures::fingerprint::Fingerprint;
 use rustc_hir::{def_id::DefId, definitions::DefPathHash};
+use rustc_index::IndexVec;
 use rustc_middle::mir::{
-    interpret::ConstValue, visit::MutVisitor, Body, ConstantKind, Location, Operand,
+    interpret::ConstValue, visit::MutVisitor, Body, ConstantKind, Local, LocalDecl, Location,
+    Operand,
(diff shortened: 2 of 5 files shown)


kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -23,18 +26,44 @@ pub fn get_stub(tcx: TyCtxt, def_id: DefId) -> Option<DefId> {
...
         let new_body = tcx.optimized_mir(replacement).clone();
         if check_compatibility(tcx, def_id, old_body, replacement, &new_body) {
-            return tcx.arena.alloc(new_body);
+            return new_body;
+        }
+    }
+    old_body.clone()
+}
+
+struct ExternFunctionTransformer<'tcx> {
```
celinval, line 39: please add some comments. :)

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -23,18 +26,44 @@ pub fn get_stub(tcx: TyCtxt, def_id: DefId) -> Option<DefId> {
...
+                    ConstValue::ZeroSized,
+                    self.tcx.type_of(stub).subst(self.tcx, &*generics),
+                );
+            }
         }
     }
-    old_body
+}
+
+pub fn transform_extern_functions<'tcx>(tcx: TyCtxt<'tcx>, callee: DefId, body: &mut Body<'tcx>) {
```
celinval: maybe move this to the top (right after transform) and document it.

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -23,18 +26,43 @@ pub fn get_stub(tcx: TyCtxt, def_id: DefId) -> Option<DefId> {
...
+
+impl<'tcx> MutVisitor<'tcx> for ExternFunctionTransformer<'tcx> {
+    fn tcx(&self) -> TyCtxt<'tcx> {
+        self.tcx
+    }
+
+    fn visit_operand(&mut self, operand: &mut Operand<'tcx>, _location: Location) {
+        let func_ty = operand.ty(&self.body, self.tcx);
+        if let ty::FnDef(reachable_function, generics) = *func_ty.kind() {
+            if let Some(stub) = get_stub(self.tcx, reachable_function) {
```
celinval, line 52: Maybe avoid calling `get_stub` for every function definitions you find. This function is actually parsing the `llvm_args` and creating the stub mapping.

You could instead create the map first and store it in your visitor, and check the map directly.

tests/kani/Stubbing/extern_c.rs
JustusAdam, line 1: Might be nice to have one test case where the function-to-stub is buried a bit deeper, e.g. called transitively.

Also I guess this is a more general stubbing question: but do we detect if the stub calls the function that is supposed to be stubbed?

kani-compiler/src/kani_middle/provide.rs
```diff
@@ -55,7 +55,9 @@ fn run_kani_mir_passes<'tcx>(
     body: &'tcx Body<'tcx>,
 ) -> &'tcx Body<'tcx> {
     tracing::debug!(?def_id, "Run Kani transformation passes");
-    stubbing::transform(tcx, def_id, body)
+    let mut transformed_body = stubbing::transform(tcx, def_id, body);
+    stubbing::transform_extern_functions(tcx, &mut transformed_body);
+    tcx.arena.alloc(transformed_body)
```
JustusAdam, line 60: Hmm, I'm not sure we need this allocation every time. Both `transform` and `transform_extern_function` could return a `Cow`. Only if at the end we have a `Cow::Owned` we would need to allocate a new body.

What I'm imagining is something like this

```rs
let transformed_body = /* in stubbing::transform */ |old_body| {
  if let Some(replacement) {
    Cow::Owned(/* replacement transformed */)
  } else {
    Cow::Borrowed(old_body)
  };
  let extern_transformed = /* in transform_extern_functions */ {
    let mut body = transformed_body.clone(); // We a local copy to mutate
    let body_was_borrowed = transformed_body.is_borrowed();
    vis.visit_body(&mut body);
    if body == old_body.borrow() {
      // No change happened, we can just use the old body
      transformed_body
    } else {
      Cow::Owned(body)
    };
    match extern_transformed {
      Cow::Owned(new_body) => tcx.arena.alloc(body),
      Cow::Borrowed(old_body) => old_body,
    }
```

Hope this makes sense.

This way we avoid the additional allocation in the arena which otherwise would happen for every single function.

Also it means that the body handed to `transform_extern_functions` is used borrowed. That means while it still needs to make a copy to mutate you can elide the copy you currently store in the visitor and just store a `&LocalDecls` in there instead which is all you need for `operand.ty` actually.

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -23,18 +26,43 @@ pub fn get_stub(tcx: TyCtxt, def_id: DefId) -> Option<DefId> {
...
+                let Operand::Constant(fn_def) = operand else { unreachable!() };
```
JustusAdam, line 53: This can actually fail if the user uses function pointers so I'd recommend to do one of the following:

1. *We disallow function pointers:* Then this should be a graceful `self.tcx.span_err`
2. *We allow function pointers:* Then in this case you just ignore this operand, since you know the visitor will have performed the stub swap earlier where the place in the operand was assigned.

JustusAdam (changes requested):
I think we should add a test case for the use of function pointers ad make sure it is handled, irrespective of whether we want to allow use of function pointers or not.

kani-compiler/src/kani_middle/stubbing/transform.rs
JustusAdam: Even better: Add documentation :P

celinval: That's what I meant. 😝

tests/kani/Stubbing/extern_c.rs
celinval: No. We do not.

kani-compiler/src/kani_middle/provide.rs
celinval: We could do that, but wouldn't it be easier to do that outside of the transformation. I.e.: 
```rust
If *orig_body != body {
    tcx.arena.alloc(body)
} else {
    old.body()
}
```

kani-compiler/src/kani_middle/stubbing/transform.rs
celinval: I believe @feliperodri is still working on enabling function pointers

kani-compiler/src/kani_middle/provide.rs
JustusAdam: Oh yeah I meant outside of the transformation. I just inlined the code to show the pipeline.
I have to say though your method is a lot simpler 😂

tests/kani/Stubbing/extern_c.rs
JustusAdam: Maybe we should?

kani-compiler/src/kani_middle/stubbing/transform.rs
JustusAdam: I think with this it could just work, since the visitor checks every kind of operand, including in assignments so it should swap them out even there.

kani-compiler/src/kani_middle/stubbing/transform.rs
feliperodri: Implemented as suggested.

kani-compiler/src/kani_middle/stubbing/transform.rs
feliperodri: Implemented as suggested.

kani-compiler/src/kani_middle/stubbing/transform.rs
feliperodri: Implemented as suggested.

tests/kani/Stubbing/extern_c.rs
feliperodri: Implemented as suggested.

kani-compiler/src/kani_middle/stubbing/transform.rs
feliperodri: The current implementation already takes care of function pointers. I added a test case on this. Am I missing something?

tests/kani/Stubbing/extern_c.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub unsafe extern "C" fn strlen(cs: *const c_char) -> size_t {
+        4
+    }
+
+    pub unsafe extern "C" fn sysconf(_input: c_int) -> c_longlong {
+        10
+    }
+}
+
+fn dig_depper(input : c_int) {
```
JustusAdam, line 30: I think you mean `dig_deeper`?

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -23,18 +26,57 @@ pub fn get_stub(tcx: TyCtxt, def_id: DefId) -> Option<DefId> {
...
+        ForeignFunctionTransformer { tcx, body: body.clone(), stub_map: HashMap::default() }
+    };
+    visitor.visit_body(body);
+}
+
+struct ForeignFunctionTransformer<'tcx> {
+    /// The compiler context.
+    tcx: TyCtxt<'tcx>,
+    /// Body of the callee function. Kani searches here for foreign functions.
+    body: Body<'tcx>,
```
JustusAdam, line 56: As far as I can tell you're only using this for `operand.ty` so it ould be enough if you cloned and stored just `LocalDecls` here.

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -23,18 +26,57 @@ pub fn get_stub(tcx: TyCtxt, def_id: DefId) -> Option<DefId> {
...
+    fn tcx(&self) -> TyCtxt<'tcx> {
+        self.tcx
+    }
+
+    fn visit_operand(&mut self, operand: &mut Operand<'tcx>, _location: Location) {
+        let func_ty = operand.ty(&self.body, self.tcx);
+        if let ty::FnDef(reachable_function, generics) = *func_ty.kind() {
+            if self.tcx.is_foreign_item(reachable_function) {
+                if let Some(stub) = self.stub_map.get(&reachable_function) {
+                    let Operand::Constant(fn_def) = operand else { unreachable!() };
```
JustusAdam, line 71: I still think this would get triggered if you do something like this
```rs
let ptr = function_name;
ptr(input)
```

What I mean to say is that I think this doesn't get triggered in your test case because the type of `ptr` is something generic you set it to but in the example I provided the type would be exactly type of `function_name` and trigger this code path. 
I don't think there is any expectation that this is a common path, but there is also no reason why we shouldn't handle that case.

I'd propose we add another test case where the function pointer variable is not a generic type and see if it crashes. It might be that constant propagation actually takes care of this. 
If it does crash you could simply do `return` instead of `unreachable!()` here and everything should work just fine.

kani-compiler/src/kani_middle/provide.rs
celinval: It was a team effort. :)

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -23,18 +26,57 @@ pub fn get_stub(tcx: TyCtxt, def_id: DefId) -> Option<DefId> {
...
         let new_body = tcx.optimized_mir(replacement).clone();
         if check_compatibility(tcx, def_id, old_body, replacement, &new_body) {
-            return tcx.arena.alloc(new_body);
+            return new_body;
+        }
+    }
+    old_body.clone()
+}
+
+/// Tranverse `body` searching for calls to foreing functions and, whevever there is
```
celinval, line 39: Traverse

kani-compiler/src/kani_middle/stubbing/transform.rs
feliperodri: I added a new test case without generic type and replaced `unreachable!()` with `return`, even tho it made no difference in the test case.

kani-compiler/src/kani_middle/stubbing/transform.rs
feliperodri: Why would that be better? It looks clean and we might want to reuse this for future transformations and having access to the body might still be useful. Is there any huge drawback I might be missing?

JustusAdam: Well you're creating a deep copy of quite a large structure without specific reason to do so.

kani-compiler/src/kani_middle/stubbing/transform.rs
JustusAdam: Great. Then my suspicions were unwarranted it seems. Now we know :)

kani-compiler/src/kani_middle/provide.rs
feliperodri: Body is not comparable, so we would either have to implement our own equality, use a `StableHash`, or implement the solution using the COW pattern as suggested by @JustusAdam. Since this change is somewhat orthogonal to this PR, I'm gonna track on this separate issue https://github.com/model-checking/kani/issues/2664.

JustusAdam (approved):

tests/kani/Stubbing/extern_c.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    pub unsafe extern "C" fn sysconf(_input: c_int) -> c_longlong {
+        10
+    }
+}
+
+fn dig_deeper(input: c_int) {
+    unsafe {
+        type FunctionPointerType = unsafe extern "C" fn(c_int) -> c_longlong;
+        let ptr: FunctionPointerType = libc::sysconf;
```
celinval, line 33: Can you please wrap the function call into a function? I.e.: Pass the function pointer as an argument.

Just to make sure the compiler won't optimize this away. Thanks!

tests/kani/Stubbing/extern_c.rs
```diff
@@ -0,0 +1,51 @@
...
+fn deeper_call() {
+    dig_depper(libc::_SC_PAGESIZE)
+}
+
+#[kani::proof]
+#[kani::stub(libc::strlen, stubs::strlen)]
+#[kani::stub(libc::sysconf, stubs::sysconf)]
+fn harness() {
+    let str: Box<i8> = Box::new(4);
+    let str_ptr: *const i8 = &*str;
```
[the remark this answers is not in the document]
feliperodri: Implemented as suggested.

tests/kani/Stubbing/extern_c.rs
feliperodri: Implemented as suggested.

kani-compiler/src/kani_middle/stubbing/transform.rs
```diff
@@ -23,18 +28,58 @@ pub fn get_stub(tcx: TyCtxt, def_id: DefId) -> Option<DefId> {
...
+}
+
+/// Traverse `body` searching for calls to foreing functions and, whevever there is
+/// a stub available, replace the call to the foreign function with a call
+/// to its correspondent stub. This happens as a separate step because there is no
+/// body available to foreign functions at this stage.
+pub fn transform_foreign_functions<'tcx>(tcx: TyCtxt<'tcx>, body: &mut Body<'tcx>) {
+    if let Some(stub_map) = get_stub_mapping(tcx) {
+        let mut visitor =
+            ForeignFunctionTransformer { tcx, local_decls: body.clone().local_decls, stub_map };
```
celinval, line 48: you should do `body.local_decls.clone()` instead.

PR #2668, opened 2023-08-09
Encapsulate KaniAttributes

JustusAdam:
### Description of changes: 

Creates a `KaniAttributes` struct with a low-level API for interacting with all the attributes found on an item.

The old `attributes` API exposed functions of the form of `extract_<features>_attribute(tcx: TyCtxt, def_id: DefId)`. These functions would parse the attributes on the item, create an intermediate map of all attributes found and then extract whichever one was relevant. The most used extraction was the `extract_harness_attributes` which builds a whole struct describing all harness attributes for an item. 

For the time this API was sufficient but in the future we may have to (hint: contracts) interact multiple times with the attributes on a single item. Ideally in this case we would not parse all attributes again and again, but we also would not want to expose the internals of the `attributes` module, because that makes maintenance hard.

This PR proposes to encapsulate the intermediate parse result as an opaque `KaniAttributes` structure. This can subsequently be efficiently queried to extract values or check them.

### Testing:

This is a refactoring change

### Checklist


kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -43,60 +43,170 @@ impl KaniAttributeKind {
...
-                attrs.iter().for_each(|attr| check_proof_attribute(tcx, def_id, attr))
+        }
+    }
+
+    /// Check that any unstable API has been enabled. Otherwise, emit an error.
+    ///
+    /// TODO: Improve error message by printing the span of the harness instead of the definition.
+    pub fn check_unstable_features(&self, enabled_features: &[String]) {
+        if !matches!(self.tcx.type_of(self.item).skip_binder().kind(), TyKind::FnDef(..)) {
+            // skip closures due to an issue with rustc.
```
feliperodri, line 140: Nitpick: maybe we could add a warning here instead of just returning with no message. No if that would cause too much noise for the user.

feliperodri (approved):
LGTM.

kani-compiler/src/kani_middle/attributes.rs
celinval: Too much noise. We could improve the warning and say that today you can't really attach a unstable attribute to a closure anyways.

celinval (approved):
Looks good. Thanks for moving this to a separate PR

kani-compiler/src/kani_middle/attributes.rs
JustusAdam: Hmm, I could have `KaniAttributes` store the span to the item it is on, that way I could also report that span here?

celinval: Sorry, improve the comment

JustusAdam: I'm confused what do you mean?

celinval: Something like this:
```suggestion
            // Skip closures since it shouldn't be possible to add an unstable attribute to them.
            // We have to explicitly skip them though due to an issue with rustc:
```

PR #2674, opened 2023-08-11
Free up additional disk space on GitHub runner

tautschnig:
### Description of changes: 

Our releasebundle-ubuntu action occasionally failed for it ran out of disk space. (See
https://github.com/model-checking/kani/actions/runs/5765091278/job/15630375703 for one such example.

This additional build step cleans out binaries that we do not need, such as Haskell, .NET, or Android SDKs. This frees up 28 GB of memory (out of a total of 84 GB).

### Resolved issues:

n/a

### Related RFC:

n/a

### Call-outs:

n/a

### Testing:

### Checklist

### DIFF
--- a/kani-compiler/src/kani_middle/attributes.rs
+++ b/kani-compiler/src/kani_middle/attributes.rs
@@ -374,7 +374,7 @@ fn parse_unwind(tcx: TyCtxt, attr: &Attribute) -> Option<u32> {
 fn parse_stubs(tcx: TyCtxt, harness: DefId, attributes: &[&Attribute]) -> Vec<Stub> {
     let current_module = tcx.parent_module_from_def_id(harness.expect_local());
     let check_resolve = |attr: &Attribute, name: &str| {
-        let result = resolve::resolve_fn(tcx, current_module, name);
+        let result = resolve::resolve_fn(tcx, current_module.to_local_def_id(), name);
         if let Err(err) = result {
             tcx.sess.span_err(attr.span, format!("failed to resolve `{name}`: {err}"));
         }
--- a/kani-compiler/src/kani_middle/resolve.rs
+++ b/kani-compiler/src/kani_middle/resolve.rs
@@ -14,7 +14,7 @@ use std::fmt;
 use std::iter::Peekable;
 
 use rustc_hir::def::{DefKind, Res};
-use rustc_hir::def_id::{DefId, LocalDefId, CRATE_DEF_INDEX, LOCAL_CRATE};
+use rustc_hir::def_id::{DefId, LocalDefId, LocalModDefId, CRATE_DEF_INDEX, LOCAL_CRATE};
 use rustc_hir::{ItemKind, UseKind};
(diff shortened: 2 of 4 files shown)


.github/workflows/kani.yml
```diff
@@ -192,6 +192,12 @@ jobs:
     container:
       image: ubuntu:18.04
     steps:
+      - name: Remove unnecessary software to free up disk space
+        run: |
+          # inspired by https://github.com/easimon/maximize-build-space/blob/master/action.yml
+          df -h
+          rm -rf /usr/share/dotnet /usr/local/lib/* /opt/*
```
zhassan-aws, line 199: Isn't this a bit too aggressive and could result in deleting binaries we need? Should we make it a bit more specific, e.g. `/usr/local/lib/android` and `/opt/ghc`?

zhassan-aws: We can also using an action: https://github.com/marketplace/actions/maximize-build-disk-space but this might be an overkill.

celinval (commented):
I was wondering if we should add this new step to [the setup action](https://github.com/model-checking/kani/blob/7eefefc880a5e60e3b0192a4568b511a807e3244/.github/actions/setup/action.yml) instead, so most of our jobs would benefit from this cleanup.

.github/workflows/kani.yml
tautschnig: I think if we're deleting binaries that we need then we'll notice in subsequent CI steps? I opted for the immediate commands over the existing action for improved transparency. The commands themselves are inferred from that action's code as indicated in my comment.

tautschnig:
> I was wondering if we should add this new step to [the setup action](https://github.com/model-checking/kani/blob/7eefefc880a5e60e3b0192a4568b511a807e3244/.github/actions/setup/action.yml) instead, so most of our jobs would benefit from this cleanup.

Done.

adpaco-aws:
I agree with @zhassan-aws that this is too aggressive. We shouldn't need 28GB ever, and it's good to have less space than that so we can notice performance regressions in the future.

tautschnig:
> I agree with @zhassan-aws that this is too aggressive. We shouldn't need 28GB ever, and it's good to have less space than that so we can notice performance regressions in the future.

I'm now more conservative in what is being removed.

zhassan-aws (approved):

.github/actions/setup/action.yml
```diff
@@ -12,6 +12,14 @@ inputs:
 runs:
   using: composite
   steps:
+      - name: Remove unnecessary software to free up disk space
+        shell: bash
+        run: |
+          # inspired by https://github.com/easimon/maximize-build-space/blob/master/action.yml
+          df -h
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android and /opt/ghc
```
adpaco-aws, line 20: I'm curious: What's `and` doing here?

tautschnig: Argh, copy&paste ... Fixed!

PR #2676, opened 2023-08-11
Add flag to the compiler to build std library (no user visible change)

celinval:
### Description of changes: 

Parts of the compiler relies on the existence of some constructs that are available in the standard library and Kani library. However, they aren't available when we are actually building the standard library (`std-lib-regression.sh`) and the Kani libraries (when building the sysroot with `cargo build-dev`).

Thus, we introduce a flag to inform the compiler that we are currently building the standard library. This flag can then be used to properly enable / disable sysroot dependent code.

### Resolved issues:
N/A

### Call-outs:

This is related to the bitmask work. 

### Testing:

* Is this a refactor change?

### Checklist

### DIFF
--- a/kani-compiler/src/kani_compiler.rs
+++ b/kani-compiler/src/kani_compiler.rs
@@ -395,6 +395,7 @@ impl Callbacks for KaniCompiler {
             queries.write_json_symtab =
                 cfg!(feature = "write_json_symtab") || matches.get_flag(parser::WRITE_JSON_SYMTAB);
             queries.reachability_analysis = matches.reachability_type();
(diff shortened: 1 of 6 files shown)


zhassan-aws:
I'm not sure I see what this change is doing. Can you expand the description and/or the comments in the code to clarify?

celinval:
> I'm not sure I see what this change is doing. Can you expand the description and/or the comments in the code to clarify?

I updated the description. Let me know if that makes sense now.

zhassan-aws:
Yes, it does. Thanks!

zhassan-aws (approved):

PR #2677, opened 2023-08-11
Add support to `simd_bitmask`

### LINKED ISSUE #2131: Add support to `simd_bitmask` intrinsic
Requested feature: Add support to `simd_bitmask` intrinsic
Use case: This intrinsic is commonly found during the code generation of the top 100 crates. Harnesses that exercise this intrinsic will fail due to unimplemented feature.

celinval:
### Description of changes: 
    
Create a sequential model of `simd_bitmask` and create a transformation pass that replace calls to the simd_bitmask intrinsic by the new model.

We will only replace those usages that we can safely get information about the simd representation.

### Resolved issues:

Resolves #2131

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

1. I still need to add some tests for the failure scenarios.
2. I created https://github.com/model-checking/kani/pull/2676 to add the `--build-std` flag which is needed to build the standard library. Once that gets merged into main, all the related changes should go away.

### Testing:

* How is this change tested? new tests

* Is this a refactor change?

### Checklist


tests/kani/SIMD/portable_simd.rs
```diff
@@ -13,3 +13,13 @@ fn check_sum_any() {
     // Cannot compare them directly: https://github.com/model-checking/kani/issues/2632
     assert_eq!((a + b).as_array(), b.as_array());
 }
+
+#[kani::proof]
+fn check_mask() {
+    // From array doesn't work either. Manually build [false, true, false, true].t st
```
zhassan-aws, line 19:
```suggestion
    // From array doesn't work either. Manually build [false, true, false, true]
```
?

kani-compiler/src/kani_middle/intrinsics.rs
```diff
@@ -0,0 +1,104 @@
...
+#![allow(dead_code)]
+use rustc_index::IndexVec;
+use rustc_middle::mir::{interpret::ConstValue, Body, ConstantKind, Operand, TerminatorKind};
+use rustc_middle::mir::{Local, LocalDecl};
+use rustc_middle::ty::{self, Ty, TyCtxt};
+use rustc_middle::ty::{Const, GenericArgsRef};
+use rustc_span::symbol::{sym, Symbol};
+use tracing::{debug, trace};
+
+pub struct AbstractIntrinsics<'tcx> {
```
zhassan-aws, line 14: Would `ModelIntrinsics` be a more suitable name?

kani-compiler/src/kani_middle/provide.rs
```diff
@@ -57,6 +65,8 @@ fn run_kani_mir_passes<'tcx>(
     tracing::debug!(?def_id, "Run Kani transformation passes");
     let mut transformed_body = stubbing::transform(tcx, def_id, body);
     stubbing::transform_foreign_functions(tcx, &mut transformed_body);
+    // This should be appled after stubbing so user stubs take precedence.
```
zhassan-aws, line 68:
```suggestion
    // This should be applied after stubbing so user stubs take precedence.
```

library/kani/src/models/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+                const TRUE: Self = -1;
+                const FALSE: Self = 0;
+            }
+        }
+    }
+
+    macro_rules! impl_unsigned_element {
+        { $ty:ty } => {
+            impl MaskElement for $ty {
+                const TRUE: Self = <$ty>::MAX;
```
zhassan-aws, line 36: Any reference for the value used here?

library/kani/src/models/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    impl_unsigned_element! { u8 }
+    impl_unsigned_element! { u16 }
+    impl_unsigned_element! { u32 }
+    impl_unsigned_element! { u64 }
+    impl_unsigned_element! { u128 }
+    impl_unsigned_element! { usize }
+
+    pub(super) const fn mask_len(len: usize) -> usize {
+        (len + 7) / 8
```
zhassan-aws, line 57: Reference?

library/kani/src/models/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+        (&mask as *const [u8; mask_len(LANES)] as *const U).read()
+    }
+
+    /// Structure used for sanity check our parameters.
+    #[repr(simd)]
+    struct Simd<T, const LANES: usize>([T; LANES]);
+}
+
+#[cfg(test)]
+mod test {
```
zhassan-aws, line 117: Can you add a description to each test? It's difficult to see what they're doing.

zhassan-aws:
Thanks for implementing this, @celinval!

kani-compiler/src/kani_middle/intrinsics.rs
```diff
@@ -0,0 +1,104 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! This module contains a MIR pass that replaces some intrinsics by rust intrinsics models as
+//! well as validation logic that can only be added during monomorphization.
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
celinval: Nope... that's leftover from development. I'll remove it.

library/kani/src/models/mod.rs
celinval: Sure

kani-compiler/src/kani_middle/intrinsics.rs
celinval, line 5:
```suggestion
```

library/kani/src/models/mod.rs
celinval: I was actually debating if we should accept unsigned input, but the rust compiler and MIRI seems to be OK with that. Anyway, there isn't much when it comes to references to intrinsics and the portable_simd library restricts the usage of Masks to just a few options, and none is unsigned.

celinval:
Sorry, I merged before pulling from my remote branch, so I force pushed over [4b6858a](https://github.com/model-checking/kani/pull/2677/commits/4b6858a7ff61fa278194b977e2542052910ec8db).

kani-compiler/src/kani_middle/intrinsics.rs
```diff
@@ -0,0 +1,103 @@
...
+            let mut new_gen_args = Vec::from_iter(gen_args.iter());
+            new_gen_args.push(elem_ty.into());
+            new_gen_args.push(len.into());
+
+            fn_def.literal = ConstantKind::from_value(
+                ConstValue::ZeroSized,
+                tcx.type_of(stub_id).instantiate(tcx, &new_gen_args),
+            );
+        } else {
+            debug!(?arg_ty, "replace_simd_bitmask failed");
```
JustusAdam, line 72: I'm curious, why are we not erroring out here?

kani-compiler/src/kani_middle/intrinsics.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+fn resolve_rust_intrinsic<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    func_ty: Ty<'tcx>,
+) -> Option<(Symbol, GenericArgsRef<'tcx>)> {
+    if let ty::FnDef(def_id, args) = *func_ty.kind() {
+        if tcx.is_intrinsic(def_id) {
+            return Some((tcx.item_name(def_id), args));
```
JustusAdam, line 99: Why does this return `args`, if we are just ignoring it later and doing the same `FnDef` pattern match again manually in `replace_simd_bitmask`?

kani-compiler/src/kani_middle/intrinsics.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// However, this pass runs after lowering intrinsics, which may replace the terminator by
+    /// an intrinsic statement (non-diverging intrinsic).
+    pub fn run_pass(tcx: TyCtxt<'tcx>, body: &mut Body<'tcx>) {
+        ModelIntrinsics { tcx, local_decls: body.local_decls.clone() }.transform(body)
+    }
+
+    pub fn transform(&self, body: &mut Body<'tcx>) {
+        for block in body.basic_blocks.as_mut() {
+            let terminator = block.terminator.as_mut().unwrap();
+            if let TerminatorKind::Call { func, args, .. } = &mut terminator.kind {
```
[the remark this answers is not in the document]
celinval: I could... but I also don't think this loop is that bad since we just care about the terminators. The rust compiler has the same loop to handle lowering intrinsics.

kani-compiler/src/kani_middle/intrinsics.rs
celinval: We prefer handling unsupported features as verification failure, instead of compilation failure. This enables the user to verify part of the code, as long as the unsupported function is not reachable.

kani-compiler/src/kani_middle/intrinsics.rs
celinval: I totally forgot that we had that already. Let me change that.

library/kani/src/models/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+    use super::intrinsics as kani_intrinsic;
+    use std::{fmt::Debug, simd::*};
+
+    extern "platform-intrinsic" {
+        fn simd_bitmask<T, U>(x: T) -> U;
+    }
+
+    /// Test that the simd_bitmask model is equivalent to the intrinsic for all true and all false
```
zhassan-aws, line 136:
```suggestion
    /// Test that the `simd_bitmask` model is equivalent to the intrinsic for all true and all false
```

library/kani/src/models/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// Tests that the model correctly fails if the size parameter of the mask doesn't match the
+    /// expected number of bytes in the representation.
+    #[test]
+    #[should_panic(expected = "Expected size of return type and mask lanes to match")]
+    fn test_invalid_generics() {
+        let mask = unsafe { mask32x16::splat(false) };
+        assert_eq!(unsafe { kani_intrinsic::simd_bitmask::<_, u16, i32, 2>(mask) }, u16::MAX);
+    }
+
+    /// Test that the simd_bitmask model is equivalent to the intrinsic for a few random values.
```
zhassan-aws, line 164:
```suggestion
    /// Test that the `simd_bitmask` model is equivalent to the intrinsic for a few random values.
```

library/kani/src/models/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// These values shouldn't be symmetric and ensure that we also handle endianess correctly.
```
zhassan-aws, line 165:
```suggestion
    /// These values shouldn't be symmetric and ensure that we also handle endianness correctly.
```

zhassan-aws (approved):

kani-compiler/src/kani_middle/intrinsics.rs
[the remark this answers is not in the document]
celinval: I think it would make sense to use the visitor if we had to handle function pointers, but I don't think you can really do that.

kani-compiler/src/kani_middle/intrinsics.rs
JustusAdam: Let me be more specific, how is this situation (a non-simd argument) legal? Do we expect this to happen? And if so, if we can't handle the intrinsic normally, shouldn't we at least issue a warning?

kani-compiler/src/kani_middle/intrinsics.rs
```diff
@@ -0,0 +1,103 @@
...
+    ///
+    /// However, this pass runs after lowering intrinsics, which may replace the terminator by
+    /// an intrinsic statement (non-diverging intrinsic).
+    pub fn run_pass(tcx: TyCtxt<'tcx>, body: &mut Body<'tcx>) {
+        ModelIntrinsics { tcx, local_decls: body.local_decls.clone() }.transform(body)
+    }
+
+    pub fn transform(&self, body: &mut Body<'tcx>) {
+        for block in body.basic_blocks.as_mut() {
+            let terminator = block.terminator.as_mut().unwrap();
```
JustusAdam, line 32: Btw there is `BasicBlockData::terminator_mut` for precisely this situation 😉

celinval: Even better. Thanks

PR #2678, opened 2023-08-12
Enable `-Z stubbing` and error out instead of ignore stub

celinval:
### Description of changes: 

Enable `-Z stubbing`, deprecate `--enable-stubbing`, and emit an error if the user tries to verify a harness that has stubs without explicitly enabling stubs. This used to be a warning before which was easy to miss. It was also hard to debug when the harness would take a while to run.

### Resolved issues:

Resolves #ISSUE-NUMBER

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

* Is this a refactor change?

### Checklist


tautschnig:
There are a number of tests that have `enable-stubbing` in their `Cargo.toml` files. Will these tests still work (without issuing a warning)?

celinval:
> There are a number of tests that have `enable-stubbing` in their `Cargo.toml` files. Will these tests still work (without issuing a warning)?

They will emit a warning but the test shouldn't fail because of that.  I can modify them in this PR If you prefer

tautschnig:
> > There are a number of tests that have `enable-stubbing` in their `Cargo.toml` files. Will these tests still work (without issuing a warning)?
> 
> They will emit a warning but the test shouldn't fail because of that. I can modify them in this PR If you prefer

If we are deprecating a feature then, yes, I think we should completely remove its use from our repository.

kani-driver/src/harness_runner.rs
```diff
@@ -70,6 +71,33 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
...
+                .collect();
+            match with_stubs.as_slice() {
+                [] => { /* do nothing */ }
+                [harness] => bail!(
+                    "Harness `{}` contains stubs which are unstable.\n\
+                    To enable stubbing, pass options `-Z stubbing`",
+                    harness
+                ),
+                harnesses => bail!(
+                    "Harness `{}` contains stubs which are unstable.\n\
```
JustusAdam, line 93: "Harnesses" and "contain"

JustusAdam: Also I'm generally a bit unhappy with the phrasing "stubs which are unstable", it sounds as if *these particular* stubs are unstable, not stubbing in general.

JustusAdam (commented):
We should probably do a similar thing for the compiler. It also tracks stubbing differently from unstable features for some reason and it's quite confusing.

Also we could consider factoring the whole "unstable features" stuff out into a crate that is common between driver and compiler so we don't have to reimplement it twice for every feature.

celinval:
> We should probably do a similar thing for the compiler. It also tracks stubbing differently from unstable features for some reason and it's quite confusing.

Ah, do you mean replace the `--enable-stubbing` from the compiler as well and just use the unstable flag that is already passed? I can do that. The reason is legacy code. The [stubbing RFC](https://model-checking.github.io/kani/rfc/rfcs/0002-function-stubbing.html) came before the [unstable API RFC](https://model-checking.github.io/kani/rfc/rfcs/0006-unstable-api.html). =)

> 
> Also we could consider factoring the whole "unstable features" stuff out into a crate that is common between driver and compiler so we don't have to reimplement it twice for every feature.

I do think having a crate just for the "unstable features" stuff is a bit overkill, but we have been talking about creating a `kani_common` crate. Just didn't have the time to do that.

JustusAdam:
> > We should probably do a similar thing for the compiler. It also tracks stubbing differently from unstable features for some reason and it's quite confusing.
> 
> Ah, do you mean replace the `--enable-stubbing` from the compiler as well and just use the unstable flag that is already passed? I can do that. The reason is legacy code. The [stubbing RFC](https://model-checking.github.io/kani/rfc/rfcs/0002-function-stubbing.html) came before the [unstable API RFC](https://model-checking.github.io/kani/rfc/rfcs/0006-unstable-api.html). =)
> 
> > Also we could consider factoring the whole "unstable features" stuff out into a crate that is common between driver and compiler so we don't have to reimplement it twice for every feature.
> 
> I do think having a crate just for the "unstable features" stuff is a bit overkill, but we have been talking about creating a `kani_common` crate. Just didn't have the time to do that.

I actually already wrote the code for that and then realized it'd work best if we refactored the compiler args to use clap deriving also so I opened #2689

celinval:
> > > We should probably do a similar thing for the compiler. It also tracks stubbing differently from unstable features for some reason and it's quite confusing.
> > 
> > 
> > Ah, do you mean replace the `--enable-stubbing` from the compiler as well and just use the unstable flag that is already passed? I can do that. The reason is legacy code. The [stubbing RFC](https://model-checking.github.io/kani/rfc/rfcs/0002-function-stubbing.html) came before the [unstable API RFC](https://model-checking.github.io/kani/rfc/rfcs/0006-unstable-api.html). =)
> > > Also we could consider factoring the whole "unstable features" stuff out into a crate that is common between driver and compiler so we don't have to reimplement it twice for every feature.
> > 
> > 
> > I do think having a crate just for the "unstable features" stuff is a bit overkill, but we have been talking about creating a `kani_common` crate. Just didn't have the time to do that.
> 
> I actually already wrote the code for that and then realized it'd work best if we refactored the compiler args to use clap deriving also so I opened #2689

Ok. So I won't do that in this PR. Thanks @JustusAdam

JustusAdam:
> Ok. So I won't do that in this PR. Thanks @JustusAdam

Thanks. I work fast when I get annoyed by something 😂 

Looks good to me otherwise

JustusAdam (approved):

tests/ui/stub-attribute/attribute.rs
```diff
@@ -1,6 +1,6 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-//
+// kani-flags: -Z stubbing
```
adpaco-aws, line 3: Do we want to make a copy of this test, use `--enable-stubbing` and just check that the warning is there?

I know this goes against @tautschnig advice of removing all uses of `--enable-stubbing`. But if we have an option, even a deprecated one, I don't think it's a good idea to leave it untested.

kani-driver/src/args/mod.rs
```diff
@@ -345,6 +345,12 @@ impl VerificationArgs {
             Some(Some(x)) => Some(x), // -j=x
         }
     }
+
+    /// Is experimental stubing enabled?
```
adpaco-aws, line 349:
```suggestion
    /// Is experimental stubbing enabled?
```

adpaco-aws (approved):
This makes sense to me, thanks @celinval !

One question: Are we tracking the removal of this feature somewhere? I don't think we have a mechanism in place to keep track of deprecated feature and when their removal is estimated to occur.

kani-driver/src/harness_runner.rs
```diff
@@ -70,6 +71,33 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
...
+            let with_stubs: Vec<_> = harnesses
+                .iter()
+                .filter_map(|harness| {
+                    (!harness.attributes.stubs.is_empty()).then_some(harness.pretty_name.as_str())
+                })
+                .collect();
+            match with_stubs.as_slice() {
+                [] => { /* do nothing */ }
+                [harness] => bail!(
+                    "Use of unstable feature 'stubbing' in harness `{}`.\n\
```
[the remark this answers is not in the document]
celinval: That was my initial thought too, but stubbing in this case is the name of the feature. That's consistent with rustc's [error message for unstable features](https://doc.rust-lang.org/stable/error_codes/E0658.html):

```rust
fn main() {
    // error[E0658]: use of unstable library feature 'option_as_slice'
    None.as_slice();
}
```

kani-driver/src/harness_runner.rs
```diff
@@ -70,6 +71,33 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
...
+                    To enable stubbing, pass options `-Z stubbing`",
```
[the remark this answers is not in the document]
celinval: I'll fix the message as soon as you reply to my comment on the single quotes.

tests/ui/stub-attribute/attribute.rs
celinval: I agree with you. @tautschnig would that be OK?

kani-driver/src/harness_runner.rs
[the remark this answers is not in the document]
adpaco-aws: I didn't know this, thanks! It's fine with me to keep `'stubbing'` then.

kani-driver/src/harness_runner.rs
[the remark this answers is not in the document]
adpaco-aws: Replied.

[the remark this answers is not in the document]
celinval:
```suggestion
                    To enable stubbing, pass option `-Z stubbing`",
```

kani-driver/src/harness_runner.rs
```diff
@@ -70,6 +71,33 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
...
+                    harness
+                ),
+                harnesses => bail!(
+                    "Use of unstable feature 'stubbing' in harnesses `{}`.\n\
+                    To enable stubbing, pass options `-Z stubbing`",
```
celinval, line 94:
```suggestion
                    To enable stubbing, pass option `-Z stubbing`",
```

tests/expected/function-stubbing-error/expected
```diff
@@ -0,0 +1,2 @@
+error: Use of unstable feature 'stubbing' in harness `main`.
+To enable stubbing, pass options `-Z stubbing`
```
celinval, line 2:
```suggestion
To enable stubbing, pass option `-Z stubbing`
```

tests/ui/stub-attribute/attribute.rs
tautschnig: If the test is all about testing the warning, then, sure, I have no objection at all! My concern was solely about leaving deprecated options in there that will continue to work for a while, and then (once we actually remove `--enable-stubbing`) having a very large cleanup commit.

Issue #2576, opened 2023-06-28
Formatting an `i8` takes a long time after nightly-2023-06-20 upgrade


celinval:
I tried this code:

```rust
// fmt.rs
#[kani::proof]
fn fmt_i8() {
    let num: i8 = kani::any();
    let s = format!("{num}");
    assert!(s.len() <= 4);
}
```

using the following command line invocation:

```
kani fmt.rs
```

with Kani version: #2551 

I expected to see this happen: Verification succeeds in a matter of seconds (before upgrade this harness used to take 6s)

Instead, this happened: Verification took about 10 minutes.

celinval:
Comparing Kani before and after the upgrade. I noticed the following:

1. There was no significant changes to the reachability analysis. The number of intrinsics decreased  by 1/3 due to the new lowering logic.
<details>
<summary>MIR Reachability Results: Before Upgrade</summary>

```
====== Reachability Analysis Result =======
Total # items: 693
Total # statements: 5788
Total # expressions: 2832

Reachable Items:
  - function: 691
  - static item: 2

Statements:
  - Assign: 2832
  - Call: 1343
  - Return: 669
  - SwitchInt: 399
  - Goto: 302
  - Assert: 128
  - Unreachable: 56
  - Intrinsic: 32
  - Drop: 27

Expressions:
  - Use: 1083
  - Aggregate: 375
  - BinaryOp: 309
  - Cast: 304
  - Ref: 259
  - UnaryOp: 117
  - Discriminant: 105
  - CheckedBinaryOp: 78
  - CopyForDeref: 74
  - Len: 67
  - AddressOf: 46
  - NullaryOp: 10
  - Repeat: 5

-------------------------------------------
```
</details>

<details>
<summary>MIR Reachability Results: After Upgrade</summary>

```
====== Reachability Analysis Result =======
Total # items: 624
Total # statements: 5268
Total # expressions: 2662

Reachable Items:
  - function: 622
  - static item: 2

Statements:
  - Assign: 2662
  - Call: 1142
  - Return: 600
  - SwitchInt: 376
  - Goto: 278
  - Assert: 118
  - Unreachable: 56
  - Drop: 25
  - Intrinsic: 11

Expressions:
  - Use: 1027
  - Aggregate: 347
  - BinaryOp: 309
  - Cast: 283
  - Ref: 245
  - Discriminant: 103
  - UnaryOp: 92
  - CheckedBinaryOp: 76
  - CopyForDeref: 74
[...]

PR #2688, opened 2023-08-16
Upgrade Rust toolchain to 2023-08-14

zhassan-aws:
### Description of changes: 

Relevant upstream commits/PRs:

https://github.com/rust-lang/rust/pull/114382: this required handling the new `compare_bytes` intrinsic

https://github.com/rust-lang/rust/pull/114622: `crate_types` is no longer a method in `Session`

### Resolved issues:

Resolves #ISSUE-NUMBER

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? Current regressions

### Checklist

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
@@ -4,6 +4,7 @@
 use super::typ::{self, pointee_type};
 use super::PropertyClass;
 use crate::codegen_cprover_gotoc::GotocCtx;
+use crate::unwrap_or_return_codegen_unimplemented_stmt;
 use cbmc::goto_program::{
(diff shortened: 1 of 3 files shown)


zhassan-aws:
I'm a bit confused about the `fmt_i8` test because in #2576, it was reported that its verification time increased from ~6s to 10 minutes. Looks like the time went down at some point, and then went up again with this toolchain upgrade? The increase is not as substantial as before though as it only went up to ~44 seconds (and solver runtime went up to ~19 seconds).

@celinval do you have any more information on the performance trend of this test?

zhassan-aws:
Ah, looks like changing the default solver to cadical brought down the time from ~1059 seconds to ~42 seconds:

https://github.com/model-checking/kani/actions/runs/5744858062

But based on the `perf-benchcomp` action runs, there is some variance in the runtime of this benchmark, so we can safely ignore the increase in runtime seen on this PR.

PR #2722, opened 2023-08-31
Automatic toolchain upgrade improvements

tautschnig:
### Description of changes: 

1. Build and run regression tests.
2. File an issue instead of creating a PR when either of the checks fail.

### Resolved issues:

N/a

### Related RFC:

n/a

### Call-outs:

Tested in my fork, which doesn’t support issues, so it’s still to be tested in full over here.

### Testing:

* How is this change tested? Ran in my fork.

### Checklist
- n/a Regression or unit tests are included, or existing tests cover the modified code

### DIFF
--- a/.github/workflows/toolchain-upgrade.yml
+++ b/.github/workflows/toolchain-upgrade.yml
@@ -66,6 +66,7 @@ jobs:
           branch: toolchain-${{ env.next_toolchain_date }}
           delete-branch: true
           title: 'Automatic toolchain upgrade to nightly-${{ env.next_toolchain_date }}'
+          labels: Z-BenchCI
           body: >
             Update Rust toolchain from nightly-${{ env.current_toolchain_date }} to
             nightly-${{ env.next_toolchain_date }} without any other source changes.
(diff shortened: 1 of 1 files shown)


zhassan-aws:
Will the PR still get created if the build or regressions fail?

celinval:
> Will the PR still get created if the build or regressions fail?

It doesn't look like it:

```yml
  if: ${{ env.next_step == 'create_pr' }}
```

celinval (approved):
Thanks

zhassan-aws:
> It doesn't look like it

In that case, would that result in the bot creating the same issue multiple times (every time the cronjob runs)?

celinval:
> > It doesn't look like it
> 
> In that case, would that result in the bot creating the same issue multiple times (every time the cronjob runs)?

That's a good point

tautschnig:
> > > It doesn't look like it
> > 
> > 
> > In that case, would that result in the bot creating the same issue multiple times (every time the cronjob runs)?
> 
> That's a good point

Reading the action's source code: I think that could happen. May I postpone addressing this limitation when we bump into it?

tautschnig:
> > > > It doesn't look like it
> > > 
> > > 
> > > In that case, would that result in the bot creating the same issue multiple times (every time the cronjob runs)?
> > 
> > 
> > That's a good point
> 
> Reading the action's source code: I think that could happen. May I postpone addressing this limitation when we bump into it?

Nevermind, I now added code so as not to create issues when one already exists.

Issue #2652, opened 2023-08-01
Feature Request: Function Contracts


JustusAdam:
By adding contracts support (requires/ensures/assigns/frees) to Kani we can unlock modular verification.

The RFC for Function Contracts is now [live](https://model-checking.github.io/kani/rfc/rfcs/0009-function-contracts.html).

celinval:
Let's keep this open until stabilization.

BusyBeaver-42:
Hello, I have a couple of issues with function contracts. This is not about incorrect behavior, but about features that could improve Kani's integration into a codebase.

# Multiple contracts for the same function

Let's say I have a *safe* function `foo` and I want to prove both its safety and its correctness.

```Rust
fn foo<F: Fn(i32) -> i32>(f: &F) -> i32 {
    f(0)
}
```

The function foo assumes that f(0) == 0. However, since it is a safe function, it cannot rely on this assumption to guarantee safety. Therefore, to check safety, I would use the following over-approximation.

```Rust
fn over_approximation(_:i32) -> i32 {
    kani::any()
}
```

However, I also would like to express the correctness in the function contract.

```Rust
#[kani::ensures(|res: &i32| kani::implies!(f(0) == 0 => *res == 0)]
fn foo<F: Fn(i32) -> i32>(f: &F) -> i32 {
    f(0)
}
```

But in this case, the harness with `over_approximation` would fail. One solution is to check safety only, and then create another function to wrap foo and verify correctness. However, this is not ideal, as I would also need to re-export the wrappers if users want to reuse the correctness contracts.

# Pre-conditions on function with an `FnMut` argument (which might cause soundness issues)

Let's say `foo` now takes an `FnMut`.

```Rust
fn foo<F: FnMut(i32) -> i32>(f: &mut F) -> i32 {
    f(0)
}
```

Now I cannot impose pre-conditions on `f`. If I want to verify partial correctness for f: Fn, I must also add a wrapper function.

# Code expansion of the `modifies` macro

Let's say I want to express in a contract that a function modifies a range of pointers. Currently, I express this as follows.

```Rust
#[kani::modifies(slice::from_raw_parts_mut(start, tail.offset_from(end) as usize))]
fn f(start: *mut T, end: *mut T) { ... }
```

I am unsure whether the [safety condition](https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html#safety) for the lifetime is satisfied during macro expansion. Or maybe directly accept pointer ranges?

[...]

Issue #2597, opened 2023-07-10
Avoiding naming clashes for `result` in function contracts


JustusAdam:
Function contracts provide the result of the call as a variable called `result` which can cause name clashes if an input is named the same.

We should handle this more gracefully and allow the user to rename the binding to something else, avoiding the clash.

JustusAdam:
Hey @pi314mm, have you tested whether this name clash actually causes compile/semantic failures? The reason I ask is there are theoretically hygiene protections in macros which (I would hope) would mean that the `arg1` and `arg1_renamed` cases should actually not conflict, but I could be wrong.

PR #2746, opened 2023-09-07
Simple Stubbing with Contracts

### LINKED ISSUE #2621: Stubbing with Contracts

JustusAdam:
### Description of changes: 

Expands the contract features introduced in #2655 by completing the checking with stubbing/replacement capabilities.

### Resolved issues:

Resolves #2621 

### Related RFC: [0009 Function Contracts](https://model-checking.github.io/kani/rfc/rfcs/0009-function-contracts.html)

Tracking issue: #2652

Optional #ISSUE-NUMBER.

### Call-outs:

This makes the internal state machine in the macro expansion more explicit. Previously it was a single, long function it is now a struct and an enum that hold the state and the generation is split into multiple methods on those. While technically a refactoring change to some degree it really only made sense to do this now, as now there are two dimensions along which the macro codegen operates.

1. Type of attribute (`requires`/`ensures`). This one was present before also.
2. Type of generated function (checking/replacement). This is new.

### Testing:

* How is this change tested? New `expected` test cases

### Checklist


kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -345,12 +404,31 @@ impl<'tcx> KaniAttributes<'tcx> {
...
                 }
+                KaniAttributeKind::StubVerified => {
+                    for contract in self.interpret_stub_verified_attribute() {
+                        let Ok((name, def_id, _span)) = contract else {
+                            continue;
+                        };
+                        let Some(Ok(replacement_name)) =
+                            KaniAttributes::for_item(self.tcx, def_id).replaced_with()
+                        else {
+                            self.tcx.sess.span_err(
```
celinval, line 415: This is returning the wrong span and the message is very generic. For the following code:

```rust
fn no_contract() {}

#[kani::proof]
#[kani::stub_verified(no_contract)]
fn harness() {
    no_contract();
}
```

I get the error:
```
error: Invalid `stub_verified` attribute format
 --> invalid_stub.rs:5:1
  |
5 | fn harness() {
  | ^^^^^^^^^^^^

```
The message should be pointing to the attribute definition. Also, in this case the error is that the function has no contract.

JustusAdam: Done. It's now pointing to both ;)

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -120,7 +120,18 @@ where
         && path.segments.iter().zip(mtch).all(|(actual, expected)| actual.ident == *expected)
 }
 
-/// Classifies the state a function is in the contract handling pipeline.
+/// Temporarily swap `$src` and `$target` using `std::mem::swap` for the
+/// execution of `$code`, then swap them back.
+macro_rules! swapped {
+    ($src:expr, $target:expr, $code:expr) => {{
+        std::mem::swap($src, $target);
+        let result = $code;
```
celinval, line 128: Can you please add the type here? I got the following error:
```rust
  --> external_contract.rs:28:17
   |
28 | #[kani::ensures(result.is_some() && rectangle.can_hold(result.unwrap()))]
   |                 ^^^^^^ cannot infer type

```

JustusAdam: Can you provide a full example I use to test whether it works? I tried a simple example but that didn't cause the error

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -340,17 +399,37 @@ impl<'tcx> KaniAttributes<'tcx> {
...
                     harness.stubs.push(self.stub_for_relative_item(name, replacement_name));
                 }
+                KaniAttributeKind::StubVerified => {
+                    for contract in self.interpret_stub_verified_attribute() {
+                        let Ok((name, def_id, span)) = contract else {
+                            continue;
+                        };
+                        let replacement_name = match KaniAttributes::for_item(self.tcx, def_id).replaced_with() {
+                            None => {
+                                self.tcx.sess.struct_span_err(span, format!("Could not find a contract for stubbing on target of `{}` attribute.", kind.as_ref())).span_note(self.tcx.def_span(def_id), "Expected a contract on this function.").emit();
```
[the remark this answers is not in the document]
celinval: Also, I'm not sure what error you are describing here

library/kani_macros/src/lib.rs
```diff
@@ -162,6 +162,26 @@ pub fn proof_for_contract(attr: TokenStream, item: TokenStream) -> TokenStream {
     attr_impl::proof_for_contract(attr, item)
 }
 
+/// `stub_verified(TARGET)` is a harness attribute (to be used on [`proof`] or
+/// [`proof_for_contract`] function) that replaces all occurrences of `TARGET`
+/// reachable from this harness with a stub generated from the contract on
+/// `TARGET`.
+///
+/// To create a contract for `TARGET` you must decorate it with at least one
```
celinval, line 170: maybe say something more generic and add for example [`requires`] or [`ensures`]. I don't want us to update this comment when we add `assigns` or any other contract attribute.

tests/expected/function-contract/missing_contract_for_replace.expected
```diff
@@ -0,0 +1,8 @@
+error: Could not find a contract for stubbing on target of `stub_verified` attribute.
```
celinval, line 1: What about something like:
```
Failed to generate verified stub: Function `no_contract` has no contract.

note: `stub_verified` can only be applied to functions with contracts. Try adding a contract to your function or use `stub` instead.
```

celinval (commented):
It looks good so far! I still need to review the proc_macro logic.

library/kani_macros/src/sysroot/contracts.rs
celinval: Here is an example...

```rust
#[kani::ensures(result.is_some())]
fn or_default<T: Default>(opt: Option<T>) -> Option<T> {
    opt.or(Some(T::default()))
}

#[kani::proof_for_contract(or_default)]
fn harness() {
    let input: Option<i32> = kani::any();
    or_default(input);
}
```

library/kani_macros/src/lib.rs
celinval: Sure. That works for me.

tests/expected/function-contract/missing_contract_for_check.expected
```diff
@@ -0,0 +1,10 @@
+error: Failed to check contract: Function `harness` has no contract.
+  |
+7 | #[kani::proof_for_contract(no_contract)]
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  |
+note: Try adding a contract to this function or use the unsound `stub` attribute instead.
```
celinval, line 6: I think you got your notes switched... I believe this note is for misusing stub_verified. The one for stub_verified seems to be the one you want here.

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: Thanks. Has been implemented. It does require a `kani::Arbitrary` bound on `T` though.

JustusAdam:
I've made a bit of a major change and added a "module-level" contract documentation as `kani::contracts` that provides a general overview and a central place for information about contracts and also how the various attributes interact.

I would appreciate any feedback you ahve on this new "module-level" documentation. I think at this point it doesn't have to be perfect as it will expand with future contract PRs anyway but whether it's good in broad strokes. Also if you have the time, download it and build the docs for the `kani` library and let me know if it works.

This now offers also the possibility of grouping all contract attributes in this module. I haven't done that here, I just reexport them, but I could if people want.

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -237,30 +356,206 @@ fn rename_argument_occurences(sig: &syn::Signature, attr: &mut Expr) -> HashMap<
...
+        let call_to_prior = &self.body;
+        let return_type = return_type_to_type(&sig.output);
+        match &self.condition_type {
+            ContractConditionsType::Requires => quote!(
+                kani::assume(#attr);
+                #call_to_prior
+            ),
+            ContractConditionsType::Ensures { argument_names } => {
+                let (arg_copies, copy_clean) = make_unsafe_argument_copies(&argument_names);
+                let attr = &self.attr;
```
celinval, line 430: this is duplicated

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -218,7 +337,7 @@ impl VisitMut for PostconditionInjector {
 /// - Creates new names for them;
 /// - Replaces all occurrences of those idents in `attrs` with the new names and;
 /// - Returns the mapping of old names to new names.
-fn rename_argument_occurences(sig: &syn::Signature, attr: &mut Expr) -> HashMap<Ident, Ident> {
+fn rename_argument_occurrences(sig: &syn::Signature, attr: &mut Expr) -> HashMap<Ident, Ident> {
```
celinval, line 340: Can you please create an issue to rename arguments (or trigger an error) if there's a name conflict with `result` and any other special variable names required by contracts?

tests/expected/function-contract/type_annotation_needed.rs
```diff
@@ -0,0 +1,14 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
+
+#[kani::ensures(result.is_some())]
+fn or_default<T: Default + kani::Arbitrary>(opt: Option<T>) -> Option<T> {
```
celinval, line 6: Can you please add this condition as part of the macro? I.e.: The return type has to implement Arbitrary.

celinval (commented):
I'm impressed with all this macro magic. Thanks!

library/kani/src/contracts.rs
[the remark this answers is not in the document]
JustusAdam: Thanks.

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -276,6 +571,7 @@ fn rename_argument_occurences(sig: &syn::Signature, attr: &mut Expr) -> HashMap<
 ///
 /// ```rs
 /// #[kanitool::checked_with = "div_check_965916"]
+/// #[kanitool::replaced_with = "div_replace_965916"]
```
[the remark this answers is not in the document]
JustusAdam: It's just easier that way, because the infrastructure to parse these already exists in the compiler. This way it's also easier to add or remove them without having to change the parser every time. 

Once this is more stabilized we can make it prettier but these are also internal so idk if it's necessary.

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: What do you mean?

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: Already exists 😛  https://github.com/model-checking/kani/issues/2597

JustusAdam: Also this renaming is safe, because we use `Span::mixed_site` which is subject to hygiene rules.

library/kani_macros/src/sysroot/contracts.rs
celinval: Line 419:
```rust
        let attr = &self.attr;
```

library/kani_macros/src/sysroot/contracts.rs
celinval: Oh, what does `mixed_site` do? I don't know this one.

JustusAdam: Yeah it's a bad naming. If you use `mixed_site` on an ident for a variable, then that variable is only visible to other code generated by this macro. If you use `call_site` it is also visible to user-written code. 

The reason it's called "mixed site" is that those rules only apply to variables (and maybe a few more things) but not to everything, so i think if you define new types for instance, they are always visible like with `call_site`.

JustusAdam: Yeah so according to https://docs.rs/proc-macro2/1.0.67/proc_macro2/struct.Span.html#method.mixed_site it basically uses `macro_rules` rules, where variable identifiers are specific to the macro invocation site and only visible there.

library/kani_macros/src/lib.rs
```diff
@@ -97,29 +97,29 @@ pub fn derive_arbitrary(item: TokenStream) -> TokenStream {
     derive::expand_derive_arbitrary(item)
 }
 
-///  Add a precondition to this function.
+/// Add a precondition to this function.
 ///
-/// This is part of the function contract API, together with [`ensures`].
+/// This is part of the function contract API, for more general information see
+/// the [module-level documentation](../kani/contracts/index.html).
```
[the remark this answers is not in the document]
JustusAdam: It works for the docs of `kani` as generated by our script. 🤷

library/kani_macros/src/sysroot/contracts.rs
celinval: I get that, but you already did that on line 419.

JustusAdam: Oh you're right.

JustusAdam:
Resolving the last conversation (the obscure paragraph) was quite tricky and lead to a major expansion in the docs. You may want to give them a read and see if they make sense @celinval and also it now has the diagram you wanted @feliperodri.

You may want to read the rendered version which you can get with `cargo rustdoc -p kani_macros -- --cfg kani --document-private-items`

library/kani/src/contracts.rs
```diff
@@ -0,0 +1,173 @@
...
+//!
+//! During verified stubbing the return value of a function with a contract is
+//! replaced by a call to `kani::any`. As such the return value must implement
+//! the `kani::Arbitrary` trait.
+//!
+//! In Kani, function contracts are optional. As such a function with at least
+//! one specification attribute is considered to "have a contract" and any
+//! absent specification type defaults to its most general interpretation
+//! (`true`). All functions with not a single specification attribute are
+//! considered " not to "have a contract" and are ineligible for use as the
```
feliperodri, line 137:
```suggestion
//! considered "not to have a contract" and are ineligible for use as the
```

library/kani/src/contracts.rs
```diff
@@ -0,0 +1,173 @@
...
+//!
+//! ```
+//! #[kani::requires(divisor != 0)]
+//! #[kani::ensures(result <= dividend)]
+//! fn my_div(dividend: u32, divisor: u32) -> u32 {
+//!   dividend / divisor
+//! }
+//! ```
+//!
+//! Once we are finished specifying our contract we can ask kani to check it's
```
feliperodri, line 76:
```suggestion
//! Once we are finished specifying our contract we can ask Kani to check it's
```

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -237,179 +430,382 @@ fn rename_argument_occurences(sig: &syn::Signature, attr: &mut Expr) -> HashMap<
...
+    function_state: ContractFunctionState,
+    /// Information specific to the type of contract attribute we're expanding.
+    condition_type: ContractConditionsType,
+    /// The contents of the attribute.
+    attr: Expr,
+    /// Body of the function this attribute was found on.
+    annotated_fn: &'a ItemFn,
+    /// An unparsed, unmodified copy of `attr`, used in the error messages.
+    attr_copy: TokenStream2,
+    /// The stream to which we should write the generated code
```
feliperodri, line 445:
```suggestion
    /// The stream to which we should write the generated code.
```

feliperodri (approved):
Thank you, Justus! Only small nitpicks (feel free to merge) and the state machine looks very clear!

PR #2753, opened 2023-09-09
Do not label auto-generated toolchain PR

tautschnig:
### Description of changes: 

Creating the PR with the Z-BenchCI label will not actually trigger the performance benchmarking workflow: that workflow is conditional on the label newly being attached to an existing PR. Consequently, one had to remove and then re-add the label for the workflow to be triggered (see
 #2752 for an example).

### Resolved issues:

n/a

### Related RFC:

n/a

### Call-outs:

n/a

### Testing:

### Checklist


celinval:
What if instead of removing the label here, we change the benchCI to run if the PR has the label...

tautschnig:
> What if instead of removing the label here, we change the benchCI to run if the PR has the label...

I’m afraid I haven’t found a good way to put this in a condition for a workflow: AFAIK this requires using the API and cannot be read from a variable directly. But I’m happy to be told otherwise!

celinval:
Yeah, I think the only way it works is that you trigger the workflow for every PR but you skip the job if the label is not available

adpaco-aws:
Another option is to add label after the PR has been created, no?

celinval:
> Another option is to add label after the PR has been created, no?

Indeed. I don't know if the bot has permission to label our PRs though.

tautschnig:
> Yeah, I think the only way it works is that you trigger the workflow for every PR but you skip the job if the label is not available

I'll try to do something to this effect, the current approach seems dangerous: Updating a branch to the latest changes from main will also quietly skip the Performance workflow even when the "Z-BenchCI" label is present.

zhassan-aws:
Another option is to rely on comments instead of a label:

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment 

similar to what is done in the rust compiler, e.g. https://github.com/rust-lang/rust/pull/115684#issuecomment-1713544516.

celinval:
I created https://github.com/model-checking/kani/pull/2785 as an alternative. It incorporates these changes and also auto-label our PRs depending on the files that were touched in the PR.

PR #2770, opened 2023-09-15
Delete obsolete stubs for Vec and related options

### LINKED ISSUE #697: Abstraction options not working
I tried this code:

```rust
fn main() {
    let mut v = Vec::<u8>::new();
    v.push(5);
    assert!(v.len() == 1);
    let x = v.pop().unwrap();
    assert!(x == 5);
}

```

```
rmc test.rs --use-abs --abs-type no-back
```

with RMC version: `8de5cd67609`

I expected to see this happen: The `assert!(x == 5)` should fails since the `no-back` abstraction doesn't store elements in the vector and returns a non-det value on `pop` (`library/rmc/stubs/Rust/vec/noback_vec.rs`):
```rust
   pub fn pop(&mut self) -> Option<T> {
        if self.len == 0 {
            None
        } else {
            self.len -= 1;
            Some(rmc::nondet::<T>())
        }
    }
```

Instead, this happened: The assertion passed and verification was successful.

If I explicitly include the `noback_vec.rs` file:
```rust
include!("/home/ubuntu/git/rmc/library/rmc/stubs/Rust/vec/noback_vec.rs");
```
The assertion fails.

### LINKED ISSUE #2766: Problem with `NoBackIter<T>::next` in  `noback_vec`
I think that the mock for `next` defined in `noback_vec` class might be unsound (https://github.com/model-checking/kani/blob/main/library/kani/stubs/Rust/vec/noback_vec.rs#L190)

`self.start` must be incremented by one before the functions returns. Something like:

```
fn next(&mut self) -> Option<T> {
        let res = if self.start == self.end { None } else { Some(__nondet::<T>())};
        self.start = self.start + 1;
        return res
    }
```

A common situation where the current mock implementation might produce unsound results:

1. `let iter = NoBackIter::new(len)` where `len > 0` 
2. Any call to `iter.next()` will NEVER returns `None` since `self.start  < self.end`

zhassan-aws:
### Description of changes: 

The `Vec` stubs have been obsolete for 2+ years, are not getting tested in our CI, and the related options (`--use-abs` and `--abs-type`) don't work. The stubs code has bugs (e.g. the one pointed out in #2766). Since we're currently taking a different approach for stubbing (stubbing functions and methods as opposed to types), this PR removes all the `Vec` stubs and related functionality.

### Resolved issues:

Resolves #697
Resolves #2766 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested?

### Checklist


adpaco-aws (approved):

celinval (approved):
Thank you! Can you please document this as a major change to the release notes? I don't think we need to add them to breaking changes, since this was an unstable option.

zhassan-aws:
> Can you please document this as a major change to the release notes? 

Done

Issue #2606, opened 2023-07-14
Automate running benchmarks in PRs that are more likely to impact performance


celinval:
**Proposed change:** 

- [ ] Automatically run performance benchmark workflow if the PR code has changed and the PR has the "Z-benchCI" label.
- [x] Automatically tag PRs with "Z-benchCI" if they touch files are likely to impact Kani's performance. For example:
    - `kani-compiler/**`
    - `rust-toolchain.toml`
    - `kani-dependencies`
    - `kani-driver/src/call-*`

**Motivation:** After #2604, running benchmark workflow for a PR will be triggered manually. In order to reduce the risks of forgetting to do so, we should automate running the workflow in riskier scenarios.

My suggestion is to use the `Z-benchCI` as the source of truth for whether the workflow runs or not. This will allow us to override the automation. For example, if the changes only touches comments of a compiler file.

celinval:
See https://github.com/marketplace/actions/labeler for that

celinval:
It looks like the auto label logic doesn't work to trigger other github actions. See: https://github.com/orgs/community/discussions/33804

celinval:
Just to clarify what's going on. The [github documentation](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow) has a good explanation on why the auto label workflow does not trigger a 'labeled' action:

> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, [redacted] , will not create a new workflow run.

To fix this issue, we have the following options:

- Create a new token and use that instead.
- Explicitly trigger the workflow when the label is added.

PR #2785, opened 2023-09-21
Auto label PRs that may require extra checks 

celinval:
### Description of changes: 

Add a new workflow that automatically adds the label "Z-BenchCI" to PRs that touch compiler, toolchain, dependencies and driver files that invoke our dependencies.

I also removed the label from the toolchain update job since this workflow will add the label once the PR is created.

### Resolved issues:

#2606

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? I manually tested in my fork: https://github.com/celinval/kani-dev/pull/12

### Checklist

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -11,4 +11,6 @@ Z-BenchCI:
   - kani-dependencies
   - kani-driver/src/call-*
   - Cargo.lock
+  - cprover_bindings/**
+  - library/**
 
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,14 @@
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+# Configuration for auto-labeling PRs
+#
+# Note that we enable dot, so "**" matches all files in a folder
+
+Z-BenchCI:
+  - kani-compiler/**
+  - rust-toolchain.toml
+  - kani-dependencies
+  - kani-driver/src/call-*
+  - Cargo.lock
+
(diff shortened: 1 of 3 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,14 @@
...
+#
+# Configuration for auto-labeling PRs
+#
+# Note that we enable dot, so "**" matches all files in a folder
+
+Z-BenchCI:
+  - kani-compiler/**
+  - rust-toolchain.toml
+  - kani-dependencies
+  - kani-driver/src/call-*
```
zhassan-aws, line 12: Why are `call-*` files included?

.github/labeler.yml
```diff
@@ -0,0 +1,14 @@
...
+  - Cargo.lock
+
```
zhassan-aws, line 14: I think we should add `cprover_bindings/**` as well.

zhassan-aws: Perhaps also `library/**`?

.github/labeler.yml
celinval: Because they may change how we invoke our dependencies.

celinval: Like a new cbmc flag

.github/labeler.yml
celinval: Sure. I thought about adding library, but I'm assuming most changes are new APIs or documentation. Anyway, we can manually add or remove the label if we feel like it's not needed

zhassan-aws (approved):

PR #2791, opened 2023-09-26
Update PR template

celinval:
### Change Description

Change the template to be more concise to address the team's concern that the template was too verbose.

### Call-outs

1. I don't think we need the disclaimer at the bottom, but I'm keeping it for now.
2. BTW, I do like the call-outs to provide context to the PR, but I believe we can leave it to the author's discretion.

Note: By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,35 +1,9 @@
+### Change Description
```
tautschnig, line 1: Can we please remove this header? Upon creating a PR, GitHub will place the commit message's body at the top. So the first manual step is to move this header.

JustusAdam: I think the template should either not have any headers or use multiple to bring structure to the PR/the commit message. I don't think there would be anything wrong with having sections in the commit message if that makes it more readable.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,35 +1,9 @@
...
 
-### Description of changes: 
-
-Describe Kani's current behavior and how your code changes that behavior. If there are no issues this PR is resolving, explain why this change is necessary.
-
-### Resolved issues:
+<!-- Describe how your changes improve Kani. Please provide some context on the problem you are solving. -->
```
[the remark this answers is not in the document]
celinval: Done, I used a block quote instead.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,35 +1,11 @@
-
-### Description of changes: 
-
-Describe Kani's current behavior and how your code changes that behavior. If there are no issues this PR is resolving, explain why this change is necessary.
-
-### Resolved issues:
+> Please ensure your PR description includes the following:
+> 1. Description on how your changes improve Kani.
```
zhassan-aws, line 2:
```suggestion
> 1. A description of how your changes improve Kani.
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,35 +1,11 @@
...
+> 2. Provide some context on the problem you are solving.
```
zhassan-aws, line 3:
```suggestion
> 2. Some context on the problem you are solving.
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,35 +1,11 @@
...
+> 3. List issues that are resolved by this PR.
```
zhassan-aws, line 4:
```suggestion
> 3. A list of issues that are resolved by this PR.
```

zhassan-aws (commented):
IMHO, the current template is clearer, and is more convenient to fill out because of the structure it provides. It could use some cleanup though (e.g. removing the checkboxes list, and perhaps the testing section). Would that make the commit message too cumbersome?

adpaco-aws:
> Would that make the commit message too cumbersome?

In my opinion, they'll be a bit harder to read. But that's not the only problem: including structure will make them more difficult to distinguish because they'll include repeated content (for example, think about how `git log` would look like).

Should we update our development documentation to clearly state that we'll be using the PR description as the commit message? And what parts to remove before doing so, too.

.github/PULL_REQUEST_TEMPLATE.md
adpaco-aws: Do we even need this item? We already have `Resolves #ISSUE-NUMBER` below.

tautschnig (approved):

JustusAdam (approved):
Seems reasonable

celinval:
I don't think there's a perfect solution here, but this is also a small problem that I don't think it's worth spending too much time. The goal is to reduce unnecessary burden and ensure that our git history captures the right amount of information.

Thank you for all the feedback! I'm going ahead with the current proposal, and I think we can adjust later.

PR #2800, opened 2023-10-02
`modifies` Clauses for Function Contracts

### LINKED ISSUE #2594: Basic `assigns` clause implementation for function contracts
CBMC's default assumption for a function that receives a pointer is that no modification of the pointer or its target is performed. This means that for functions that receive (and mutate) `&mut T` we need to specify an `assigns` clause in addition to `requires` and `ensures`.

In the first iteration we propose to add a simple `kani::assigns` macro which takes a list of lvalues that are lowered the usual way and passed onto CBMC as a `__CPROVER_assigns` clause.

In addition to add some basic support for arrays we propose using slice syntax `[..]` to denote that an entire array or the pointed-to range of a pointer might be assigned to.

Here is an example of what this looks like 

```
#[post(self.len() == 0)]
#[kani::assigns((*self).keys.buf.ptr.pointer.pointer[..], (*self).keys.buf.cap, (*self).keys.len)]
#[kani::assigns((*self).values.buf.ptr.pointer.pointer[..], (*self).values.buf.cap, (*self).values.len)]
pub fn clear(&mut self) {
    self.keys.clear();
    self.values.clear();
}
```

Note this is very basic and leaks private detail, e.g. the lvalues `assigns` can refer to private fields

JustusAdam:
Extends the function contract functionality with a `modifies` clause. 

The design is different from #2594 but serves a similar purpose. The `modifies` clause allows the user to specify which parts of a structure a function may assign to. Essentially refining the `mut` annotation.

We allow arbitrary (side-effect free) expressions in the `modifies` clause. The expressions are evaluated as part of the preconditions and passed to the function-under-verification as additional arguments. CBMC is then instructed to check that those locations are assigned. Aliasing means that this also adds the location in the original structure to the write set.

Each expression must return a pointer to a value that implements `Arbitrary`. On replacement we then simply assign `*ptr = kani::any()`, relying again on aliasing to update the original structure.

Additional tests for the new functionality are provided.

Resolves #2594 

### API divergence from CBMC (accepted)

The current design goes roughly as follows: We start with a `modifies` annotation on a function

```rs
#[modifies(obj.some_expr())]
fn target(obj: ...) { ... }
```

And from this we generate code to the effect of (simplified here)

```rs
fn target_check(obj: ...) {
    // Undo the lifetime entanglements
    let modified_1 = std::mem::transmute::<&'a _, &'b _>(obj.some_expr());
    target_wrapper(obj, modified_1);
}

#[cbmc::assigns(*modified_1)]
fn target_wrapper(obj: ..., modified_1: &impl kani::Arbitrary) { ... }
```

Unlike CBMC we expect `obj.some_expr()` to be of a **pointer type** (`*const`, `*mut`, `&mut` or `&`) that points to the object which is target of the modification. So if we had a `t : &mut T` that was modified, CBMC would expect its assigns clause to say `*t`, but we expect `t` (no dereference). 

The reason is that the code we generate uses the workaround of creating an alias to whichever part of `obj` is modified and registers the alias with CBMC (thereby registering the original also). If we generated code where the right side of `let modified_1 =` is not of pointer type, then the object is moved to the stack and the aliasing destroyed.

The open questions is whether we are happy with this change in API. (Yes)

### Test cases when expressions are used in the clause.

With more complex expressions in the modifies clause it becomes hard to define good test cases because they reference generated code as in this case:

```rs
#[kani::requires(**ptr < 100)]
#[kani::modifies(ptr.as_ref())]
fn modify(ptr: &mut Box<u32>) {
    *ptr.as_mut() += 1;
}
```

This passes (as it should) and when commenting out the `modifies` clause we get this error:

```
Check 56: modify_wrapper_895c4e.assigns.2
	 - Status: FAILURE
	 - Description: "Check that *var_2 is assignable"
	 - Location: assigns_expr_pass.rs:8:5 in function modify_wrapper_895c4e
```

The information in this error is very non-specific, hard to read and brittle. How should we define robust "expected" test cases for such errors?

### Corner Cases / Future Improvements

- #2907 
- #2908 
- #2909 

## TODOs

- [ ] Test Cases where the clause contains
  - [x] `Rc` + (`RefCell` or `unsafe`) (see #2907)
  - [x] Fields
  - [x] Statement expressions
  - [x] `Vec` (see #2909)
  - [ ] Fat pointers
- [ ] update contracts documentation
- [x] Make sure the wrapper arguments are unique.
- [x] Ensure `nondet-static-exclude` always uses the correct filepath (relative or absolute)
- [ ] Test case for multiple `modifies` clauses.


tests/expected/function-contract/modifies/havoc_pass.expected
```diff
@@ -0,0 +1,38 @@
+copy_replace.assertion\
+- Status: SUCCESS\
+- Description: "equality"\
+in function copy_replace
+
+VERIFICATION:- SUCCESSFUL
+
+copy.assigns\
+- Status: SUCCESS\
+- Description: "Check that var_4 is assignable"\
```
zhassan-aws, line 10: Checking the MIR variable might be brittle.

tests/expected/function-contract/modifies/havoc_pass.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
+
+// These two are reordered in comparison to `havoc_pass` and we expect the test case to pass still
```
zhassan-aws, line 5: I think this comment is not relevant to this file.

tests/expected/function-contract/modifies/havoc_pass_reordered.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
+
+// These two are reordered in comparison to `havoc_pass` and we expect the test case to pass still
```
zhassan-aws, line 5: This test seems identical to `havoc_pass.rs`. Is there some order that meant to be changed?

tests/expected/function-contract/modifies/simple_pass.rs
```diff
@@ -0,0 +1,16 @@
...
+
+#[kani::requires(*ptr < 100)]
+#[kani::modifies(ptr)]
+fn modify(ptr: &mut u32) {
+    *ptr += 1;
+}
+
+#[kani::proof_for_contract(modify)]
+fn main() {
+    let _ = Box::new(());
```
zhassan-aws, line 13: Same question here

tests/expected/function-contract/modifies/vec_pass.fixme-2909.rs
```diff
@@ -0,0 +1,46 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
+
```
zhassan-aws, line 4: Can you add a description of why this is a fixme?

zhassan-aws (commented):
First round of comments for the tests only. A nice set of tests!.

zhassan-aws (commented):
Looks great!

One question I have is: what happens if a function modifies a static variable?

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -48,6 +58,8 @@ enum KaniAttributeKind {
     /// Attribute on a function that was auto-generated from expanding a
     /// function contract.
     IsContractGenerated,
+    Modifies,
+    InnerCheck,
```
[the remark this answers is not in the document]
JustusAdam: Documentation has been added

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -222,6 +222,74 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    /// This merges with any previously attached contracts.
+    pub fn attach_contract(&mut self, instance: Instance<'tcx>, contract: Vec<Local>) {
+        // This should be safe, since the contract is pretty much evaluated as
+        // though it was the first (or last) assertion in the function.
+        self.set_current_fn(instance);
+        let goto_contract = self.as_goto_contract(contract);
+        let name = self.current_fn().name();
+
+        // CBMC has two ways of attaching the contract and it seems the
+        // difference is whether dfcc is used or not. With dfcc it's stored in
```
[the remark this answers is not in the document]
JustusAdam: Has been updated. We always use dfcc

tests/expected/function-contract/modifies/expr_replace_pass.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
+
+#[kani::requires(**ptr < 100)]
+#[kani::modifies(ptr.as_ref())]
+#[kani::ensures(*ptr.as_ref() == prior + 1)]
```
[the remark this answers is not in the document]
JustusAdam: Yes

tests/expected/function-contract/modifies/havoc_pass.expected
JustusAdam: Sadly, as  mentioned in the description of the PR, there's not really a good way to specify these checks in `.expected` files. If you have suggestions how to work around this I'm all ears.

cprover_bindings/src/irep/to_irep.rs
```diff
@@ -499,10 +503,44 @@ impl ToIrep for SwitchCase {
...
 
+impl ToIrep for Lambda {
+    fn to_irep(&self, mm: &MachineModel) -> Irep {
+        let (ops_ireps, types) = self
+            .arguments
+            .iter()
+            .map(|param| {
+                let ty_rep = param.typ().to_irep(mm);
+                (
+                    Irep::symbol(param.identifier().unwrap_or("_".into()))
```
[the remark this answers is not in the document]
JustusAdam: Would you prefer a different default name for unnamed arguments? I could make it `_n` where `n` is the argument index.

JustusAdam:
> Looks great!
> 
> One question I have is: what happens if a function modifies a static variable?

Good point. I think it should work but let me add a test case for it.

tests/expected/function-contract/modifies/havoc_pass.expected
feliperodri: could you at least use a regular expression here like `var_*`?

library/kani/src/contracts.rs
```diff
@@ -118,14 +118,22 @@
 //!
 //! ## Specification Attributes Overview
 //!
-//! There are currently two specification attributes available for describing
-//! function behavior: [`requires`][macro@requires] for preconditions and
+//! The basic two two specification attributes available for describing
```
feliperodri, line 121:
```suggestion
//! The basic two specification attributes available for describing
```

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -490,6 +563,79 @@ impl<'tcx> KaniAttributes<'tcx> {
...
+}
+
+/// Pattern macro for the comma token used in attributes.
+macro_rules! comma_tok {
+    () => {
+        TokenTree::Token(Token { kind: TokenKind::Comma, .. }, _)
+    };
+}
+
+/// Parse the a token stream inside an attribute (like `kanitool::modifies`) as a comma separated
```
feliperodri, line 591:
```suggestion
/// Parse the token stream inside an attribute (like `kanitool::modifies`) as a comma separated
```

cprover_bindings/src/goto_program/symbol_table.rs
```diff
@@ -79,6 +79,15 @@ impl SymbolTable {
...
     }
+
+    pub fn attach_contract<T: Into<InternedString>>(
+        &mut self,
+        name: T,
+        contract: FunctionContract,
+    ) {
+        let sym = self.symbol_table.get_mut(&name.into()).unwrap();
+        sym.attach_contract(contract);
+    }
```
feliperodri, line 90: Doc?

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -181,10 +191,80 @@ impl GotocCodegenBackend {
...
+        let span_of_recursion_wrapper = tcx.def_span(recursion_wrapper_id);
+        let location_of_recursion_wrapper = self.codegen_span(&span_of_recursion_wrapper);
+
+        let full_name = format!(
+            "{}:{}::REENTRY",
+            location_of_recursion_wrapper
+                .filename()
+                .expect("recursion location wrapper should have a file name"),
+            tcx.item_name(recursion_wrapper_id),
+        );
```
feliperodri, line 254: In which case we would ended up without a file name at this point?

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -996,31 +1298,159 @@ fn is_probably_impl_fn(fun: &ItemFn) -> bool {
...
+}
+
+/// Hash this `TokenStream` and return an integer that is at most digits
+/// long when hex formatted.
+fn short_hash_of_token_stream(stream: &proc_macro::TokenStream) -> u64 {
+    use std::hash::Hasher;
+    let mut hasher = std::collections::hash_map::DefaultHasher::default();
+    hash_of_token_stream(&mut hasher, proc_macro2::TokenStream::from(stream.clone()));
+    let long_hash = hasher.finish();
+    long_hash % 0x1_000_000 // six hex digits
```
feliperodri, line 1326: I'd replace this magic number with a self-explained constant.

feliperodri (approved):
Thanks for this work! I didn't find any major issues that would block this merge.

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
JustusAdam: Are you asking in which case this can fail? I'm honestly not sure. I assume only Kani or CBMC builtin symbols might not have a filename but those you couldn't attach a contract to anyway.

tests/expected/function-contract/modifies/havoc_pass.expected
JustusAdam: I checked and the test runner does not (currently) support regular expressions. I could add this capability but that would perhaps need to be a separate PR. Would you like me to try?

feliperodri: Nope. We can track this in a github issue and merge it on a separate PR. Not a blocker for this one.

library/kani/src/lib.rs
```diff
@@ -92,6 +92,73 @@ macro_rules! implies {
     };
 }
 
+/// Helper trait for code generation for `modifies` contracts.
+///
+/// We allow the user to provide us with a pointer-like object that we convert as needed.
+#[doc(hidden)]
+pub trait Pointer<'a> {
```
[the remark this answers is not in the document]
celinval: Can you please move this a different module? Maybe something like "internal" or even contracts.

tests/expected/function-contract/.unsafe_assigns_rc
```diff
@@ -0,0 +1,33 @@
+// Copyright Kani Contributors
```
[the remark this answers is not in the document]
celinval: What are you trying to achieve here? Is this supposed to be a fixme test?

kani-compiler/src/kani_middle/attributes.rs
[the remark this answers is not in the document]
celinval: where?

celinval (commented):
From a design perspective, I was wondering why we need the `ContractInfoChannel`. Can we compute the information upfront at the same time we generate the HarnessMetadata?

celinval:
BTW, can you please create a second PR that updates the RFC with the new syntax and semantics of modifies?

JustusAdam:
Thanks for the review. I'll get to your individual changes in a bit but I thought I'd address this question first.

> From a design perspective, I was wondering why we need the `ContractInfoChannel`. Can we compute the information upfront at the same time we generate the HarnessMetadata?

Technically yes, we could compute it then, however the information relies on the output of the reachability analysis, so we'd have to run a separate reachability pass. If you're happy with that overhead, and also the complexity of having two separate passes or a generic, visitor-like, reusable pass then yes, we can precompute.

celinval:
> Thanks for the review. I'll get to your individual changes in a bit but I thought I'd address this question first.
> 
> > From a design perspective, I was wondering why we need the `ContractInfoChannel`. Can we compute the information upfront at the same time we generate the HarnessMetadata?
> 
> Technically yes, we could compute it then, however the information relies on the output of the reachability analysis, so we'd have to run a separate reachability pass. If you're happy with that overhead, and also the complexity of having two separate passes or a generic, visitor-like, reusable pass then yes, we can precompute.

That's because you need the mangled name for the function, right? If you need the result of the reachability analysis, the upfront analysis won't work, since the harness could have stubs that influence the reachability result.

Another possibility is to use the QueryDb which is already shared between KaniCompiler and GotocCodegenBackend

tests/expected/function-contract/.unsafe_assigns_rc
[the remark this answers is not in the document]
JustusAdam: Yes, pretty much. Do I just rename it "fixme..." and that would be fine?

JustusAdam:
> That's because you need the mangled name for the function, right? If you need the result of the reachability analysis, the upfront analysis won't work, since the harness could have stubs that influence the reachability result.

Not just the mangled name of the function, but the mangled name of the *instance* which we only know post-reachability.

JustusAdam:
> Another possibility is to use the QueryDb which is already shared between KaniCompiler and GotocCodegenBackend

I can do that, however there is actually a comment in the code gen "Queries shouldn't change today once codegen starts." which I would be violating. Just so you know

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -178,10 +188,82 @@ impl GotocCodegenBackend {
...
+        self.attach_contract(instance_of_check, assigns_contract);
+
+        let wrapper_name = self.symbol_name_stable(instance_of_check);
+
+        let recursion_wrapper_id =
+            function_under_contract_attrs.checked_with_id().unwrap().unwrap();
+        let span_of_recursion_wrapper = tcx.def_span(recursion_wrapper_id);
+        let location_of_recursion_wrapper = self.codegen_span(&span_of_recursion_wrapper);
+
+        let full_name = format!(
```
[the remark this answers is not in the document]
JustusAdam: I don't have control over this. This is the name that CBMC expects.

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -77,13 +78,16 @@ impl GotocCodegenBackend {
...
+    /// Invariant: iff `check_contract.is_some()` then `return.2.is_some()`
     fn codegen_items<'tcx>(
         &self,
         tcx: TyCtxt<'tcx>,
         starting_items: &[MonoItem],
         symtab_goto: &Path,
         machine_model: &MachineModel,
-    ) -> (GotocCtx<'tcx>, Vec<MonoItem>) {
+        check_contract: Option<InternalDefId>,
+    ) -> (GotocCtx<'tcx>, Vec<MonoItem>, Option<AssignsContract>) {
```
celinval, line 90: I was wondering if there's any reason why you didn't add the AssignsContract to GotocCtx. Not a blocker.

kani-compiler/src/kani_queries/mod.rs
```diff
@@ -41,4 +44,40 @@ impl QueryDb {
     pub fn args(&self) -> &Arguments {
         self.args.as_ref().expect("Arguments have not been initialized")
     }
+
+    pub fn add_modifies_contract(&mut self, name: InternedString, contract: AssignsContract) {
+        assert!(
+            self.modifies_contracts.insert(name, contract).is_none(),
+            "Invariant broken, tried adding second modifies contracts to: {name}",
+        )
```
celinval, line 52: I prefer not having functions with side effects inside assertions, except for tests. It's more readable to separate the action from the validation.

```suggestion
        let replaced = self.modifies_contracts.insert(name, contract);
        assert_eq!(replaced, None, "Tried adding second modifies contracts to: {name}")
```

Note: The suggestion might need tweaking.

kani-compiler/src/kani_queries/mod.rs
```diff
@@ -41,4 +44,40 @@ impl QueryDb {
...
+    }
+
+    pub fn get_modifies_contracts(&mut self, key: &str) -> Option<AssignsContract> {
+        self.modifies_contracts.remove(&key.intern())
+    }
+
+    pub fn assert_modifies_contracts_received(&self) {
```
celinval, line 59: Why do we need this? I don't understand why users of this class should be invoking this function.

kani-compiler/src/kani_queries/mod.rs
```diff
@@ -41,4 +44,40 @@ impl QueryDb {
...
+
+    pub fn add_modifies_contract(&mut self, name: InternedString, contract: AssignsContract) {
+        assert!(
+            self.modifies_contracts.insert(name, contract).is_none(),
+            "Invariant broken, tried adding second modifies contracts to: {name}",
+        )
+    }
+
+    pub fn get_modifies_contracts(&mut self, key: &str) -> Option<AssignsContract> {
+        self.modifies_contracts.remove(&key.intern())
```
celinval, line 56: Why are you removing the contract here?

nit: The rust naming convention is to omit the "get_" prefix. Also, shouldn't this be singular. `modifies_contract`.

kani-driver/src/call_goto_instrument.rs
```diff
@@ -160,6 +162,30 @@ impl KaniSession {
         self.call_goto_instrument(args)
     }
 
+    /// Make CBMC enforce a function contract.
+    pub fn instrument_contracts(&self, harness: &HarnessMetadata, file: &Path) -> Result<()> {
```
celinval, line 166: Can you please improve the structure of this function? You have an early return if `check` is `None`. Then there's some code that runs inside a `if let Some(assigns) = check`, but not others, such as the call to `goto_instrument`.

Instead, you should wrap the entire code after line 167 inside the `if let Some(assigns) = check {}` block.

celinval (changes requested):
Just a few more changes, and I think we'll be ready to go. Thanks

kani-driver/src/call_goto_instrument.rs
JustusAdam: Good point. I kept the early return but simplified the latter half significantly

kani-compiler/src/kani_queries/mod.rs
JustusAdam: The naming might not be ideal. This works in conjunction with the fact that `get_modifies_contract` removes the contract to ensure that all contracts were handled. 

I've added documentation explaining this

celinval: Can you retrieve the map of contracts instead inside `kani-compiler`? Removing an item as part of the `get` method is fragile and the interface of this object is trying to reflect the one specific use case you have right now.

kani-driver/src/call_goto_instrument.rs
celinval: That looks great. Thanks

celinval (approved):

PR #2802, opened 2023-10-03
Update Rust toolchain to 2023-10-05

tautschnig:
Source changes required by the following upstream commits:

* rust-lang/rust@44ac8dcc71 Remove GeneratorWitness and rename GeneratorWitnessMIR.
* rust-lang/rust@4ed4913e67 Merge ExternProviders into the general Providers struct
* rust-lang/rust@3c52a3e280 subst -> instantiate
* rust-lang/rust@6a02baaa3d Partially outline code inside the panic! macro
* rust-lang/rust@3148e6a993 subtyping_projections

Resolves: #2811


tautschnig:
The changes (seemingly) required by https://github.com/rust-lang/rust/commit/4ed4913e67cab329bd3cca92dff1cf558cc3575e break stubbing. I have now issued #2806, which is as far as we can go until we have to face this change.

tautschnig:
It seems that three of the already-slow tests got even slower. Not sure what to do about this.

zhassan-aws:
Sometimes, this is due to noise, so I suggest running them locally multiple times to check whether the slow-downs are real.

celinval (approved):
I think we should merge this despite the performance benchmark single test failure. The failing test is fairly flaky, and the only difference is in solving time. Number of VCCs and program steps didn't change at all.

PR #2809, opened 2023-10-05
Support for inductively verifying contracts of recursive functions

### LINKED ISSUE #2724: Contract Verification of Recursive Functions

JustusAdam:
This adds support for verifying the contracts of recursive functions inductively. 

The idea is quite simple. We use a global variable to track whether we're reentering a function and depending on that value we either continue with checking the contract or assume the hypothesis by using its replacement.

The precise mechanism that I implemented is explained in the documentation [here](https://github.com/JustusAdam/kani/blob/a31176c79098205ca0b2c2b199cdf06653754551/library/kani_macros/src/sysroot/contracts.rs#L126) 

Resolves #2724 

**Open Question:** to facilitate induction the return type of the function must implement `kani::Arbitrary`. Since we can't discriminate between recursive and non-recursive functions in the proc-macro this means that this applies to all functions with contracts, not just recursive ones. This may be an unacceptable restriction. We could consider making inductive verification explicit with an annotation like `#[kani::inductive]` or conversely make inductive the default and let users opt-out with `#[kani::no_inductive]`. This is now tracked in #2823 

I had to remove the names of the functions in which the messages occur in the test cases, since they are now generated, hard to read names


celinval:
BTW, I have the following pre-commit hook to avoid all the whitespace issues:

```bash
#!/usr/bin/env bash
# Copyright Kani Contributors
set -e

# source: https://github.com/awslabs/git-secrets
git secrets --pre_commit_hook -- "$@"

TOP_DIR=$(git rev-parse --show-toplevel)

check_rust_style() {
    pushd ${TOP_DIR}
    ./scripts/kani-fmt.sh --check
    cargo clippy --all -- -D warnings
    popd
}

check_rust_style
```

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -809,6 +904,40 @@ fn requires_ensures_main(attr: TokenStream, item: TokenStream, is_requires: bool
     output.into()
 }
 
+fn pats_to_idents<P>(
+    sig: &mut syn::punctuated::Punctuated<syn::FnArg, P>,
+) -> impl Iterator<Item = Ident> + '_ {
+    sig.iter_mut().enumerate().map(|(i, arg)| match arg {
+        syn::FnArg::Receiver(_) => Ident::new("self", Span::call_site()),
```
zhassan-aws, line 911: Should `"self"` be replaced with [`SelfValue`](https://docs.rs/syn/latest/syn/token/struct.SelfValue.html)?

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -794,13 +861,41 @@ fn requires_ensures_main(attr: TokenStream, item: TokenStream, is_requires: bool
...
+                (quote!(Self::#check_fn_name), quote!(Self::#replace_fn_name))
+            } else {
+                (quote!(#check_fn_name), quote!(#replace_fn_name))
+            };
+
+            handler.output.extend(quote!(
+                #[allow(dead_code, unused_variables)]
+                #[kanitool::is_contract_generated(recursion_wrapper)]
+                #wrapper_sig {
+                    static mut REENTRY: bool = false;
```
zhassan-aws, line 887: Would it be safer/cleaner to pass the bool to the functions instead of using a static?

zhassan-aws (commented):
Nice! A few comments/questions.

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: Hmm, well then the function would have to return something like `Either<Ident, SelfValue>` because I don't think there's a canonical way to turn a `SelfValue` into an `Ident`

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: That doesn't work unfortunately, because you could have mutually recursive call chains where `foo` calls `bar` which then calls `foo` so to thread the boolean you'd have to thread it through `bar` also.

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -109,18 +111,64 @@
...
+//! up upon entry into the body of the function under verification. We use a
+//! global variable that tracks whether we are re-entering the function
+//! recursively and starts off as `false`. On entry to the function we flip the
+//! variable to `true` and dispatch to the check (induction step). If the check
+//! recursively calls our function our re-entry tracker now reads `true` and we
+//! dispatch to the replacement (application of induction hypothesis). Because
+//! the replacement function only checks the conditions and does not perform
+//! other computation we will only ever go "one recursion level deep", making
+//! inductive verification very efficient. Once the check function returns we
+//! flip the tracker variable back to `false` in case the function is called
```
zhassan-aws, line 148: Do we have a test for this scenario (a function called more than once)?

library/kani_macros/src/sysroot/contracts.rs
zhassan-aws: I see.

library/kani_macros/src/sysroot/contracts.rs
zhassan-aws: I see. It might make sense to use it in `visit_path` below (see my next comment).

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -809,6 +904,40 @@ fn requires_ensures_main(attr: TokenStream, item: TokenStream, is_requires: bool
...
+            ident
+        }
+    })
+}
+
+struct SelfDetector(bool);
+
+impl<'ast> Visit<'ast> for SelfDetector {
+    fn visit_path(&mut self, i: &'ast syn::Path) {
+        self.0 |= i.get_ident().map_or(false, |i| i == "self")
```
zhassan-aws, line 930: Nit: use `SelfValue` and `SelfType`?

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -794,13 +861,41 @@ fn requires_ensures_main(attr: TokenStream, item: TokenStream, is_requires: bool
...
-                #[kanitool::checked_with = #check_fn_name_str]
+                #[kanitool::checked_with = #recursion_wrapper_name_str]
                 #[kanitool::replaced_with = #replace_fn_name_str]
                 #vis #sig {
                     #block
                 }
             ));
 
+            let mut wrapper_sig = sig.clone();
+            attach_require_kani_any(&mut wrapper_sig);
```
zhassan-aws, line 872: This is mentioned in the callouts, but I'm not clear on how this constraints things beyond the already existing call to `attach_require_kani_any` in `emit_replace_function`.

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: Not possible (I'm fairly sure, I double checked). This is the canonical way to match for `self` as an expression, because `self` as an expression is (simplified) `Expr(PathExpr { path: Path { path: [Ident("self")] } })`. `SelfType` and `SelfValue` is only used in function signatures.

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: I'm using it now via `Token!`, maybe that's what you meant?

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: This actually made me realize that we don't have to match for `self`, but can just check if there's a `Receiver`

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: Well, the wrapper calls both check and replace, and since replace requires in `impl kani::Arbitary` so must the wrapper

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: I've added one that I think tests what we need, but I'm not 100% certain 😂

JustusAdam:
Hmm, my formatting was wrong *even though* I ran `kani-fmt.sh`

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: Your comments sent me down a bit of a rabbit hole wo I made some changes and actually added some tests.
I'm not super happy with the way I match for `Self`, the fact that the `Ident`s only compare equal with `Span::mixed_site()` makes me feel uneasy somehow. But I guess it's fine

library/kani_macros/src/sysroot/contracts.rs
zhassan-aws: Yes, that's what I meant.

library/kani_macros/src/sysroot/contracts.rs
zhassan-aws: Can you give an example of a function that would be rejected, that was accepted before this PR?

tests/expected/function-contract/fail_on_two.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
+
```
zhassan-aws, line 4: Can you add a description of what this test is checking?

tests/expected/function-contract/fail_on_two.rs
```diff
@@ -0,0 +1,18 @@
...
+#[kani::ensures(result < 3)]
+fn fail_on_two(i: i32) -> i32 {
+    match i {
+        0 => fail_on_two(i + 1),
+        1 => 2,
+        _ => unreachable!("fail on two"),
```
zhassan-aws, line 10: Would it make more sense to make this test fail the postcondition instead via a recursive call, e.g.:
```rust
        2 => fail_on_two(i + 1),
        _ => 3,
```

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -966,3 +967,58 @@ macro_rules! passthrough {
...
+
+        detect_impl_fn!(true, fn self_by_other_wrapped_arg(u: usize, slf: Vec<Self>) {})
+    }
+
+    #[test]
+    fn detect_impl_fn_by_qself() {
+        detect_impl_fn!(
+            true,
+            fn self_by_mention(u: usize) {
+                Self::other(u)
```
zhassan-aws, line 1008: Is this possible given that the function doesn't have a return type?

JustusAdam: Of course. The whole issue is that you can have a function in an `impl` that possibly never mentions `self` or `Self`. But if you want to call that function (which we need to do in our generated code) you need to call it `Self::fn_name`.

for instance it's perfectly legal to do

```rs
struct S;
impl S {
  fn helper(u: usize) -> bool {
    u < 8
  }
}
```

But if you want to call it you have to do `S::foo(8)`.

That's why the whole self detector is just a heuristic and can actually lead to spurious compile errors. It's not very likely to have such a helper function but in can happen. What's worse actually is not using self when you should. For instance

```rs
impl<T> Vec<T> {
  fn new() -> Vec<T> {
    Vec { cap: 0, buf: NonNull::dangling() }
  }
}
```

This is perfectly valid Rust code , but wouldn't be detected by this as being in an `impl` because it uses `Vec<T>` as return type not `Self`. In the proc macro it's actually not possible to detect this case. And don't get fooled by the existence of the `ImplItem` enum and `ImplItemFn` struct, those you'll only get if you stick the macro on top of a whole `impl` block, then it gives you an `ItemImpl`, which contains them but if your macro is places on a function you can only parse an `ItemFn` (or possibly an `Item` but that will always be the `ItemFn` variant).

The basic problem with rust macros is that they only see the syntax that's given to them and no other context which is what you'd need to determine whether you're in an `impl`.

tests/expected/function-contract/fail_on_two.rs
JustusAdam: I've added one so now we do both

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: Oh yes, so if you had `Option::unwrap` for instance

```rs
impl<T> Option<T> {
  #[kani::ensures(...doesnt matter)]
  fn unwrap(self) -> T {
    if let Some(t) = self {
      t
    } else { panic!() }
  }
}
```

Then before the signature of the check function would have been `for<T> fn(Option<T>) -> T` and of replace `for<T: Arbitrary> fn(Option<T>) -> T`.
In other words you could have done `proof_for_contract` for your custom `struct MyThing { .. }` and if you wanted to use `stub_verified` you'd have to implement `Arbitrary for MyThing` (makes sense how else to would you do it).

But now "check" (or in this case "wrapper" which is now the new entry to the checking logic) calls "replace" and so the signature is `for<T: Arbitrary> fn(Option<T>) -> T` even for the check so if you want to do `proof_for_contract` you need `Arbitrary for MyThing`.

library/kani_macros/src/sysroot/contracts.rs
zhassan-aws: OK, thanks for the explanation. I think it might be worth incorporating some of that explanation in the doc comments to make this clearer.

library/kani_macros/src/sysroot/contracts.rs
zhassan-aws: I see. In that case, I think it makes sense to require `Arbitrary` anyway, since there one is likely to want to use `stub_verified`, and not just `proof_for_contract`.

zhassan-aws (approved):
Thanks @JustusAdam! It's great to be able to support recursive functions!

JustusAdam:
> Thanks @JustusAdam! It's great to be able to support recursive functions!

Thanks. Yeah I'll do that.

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: I added the documentation I think you were referring to (and more). Let me know if anything else is missing.

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: Probably true but I've opened #2823 to maybe relax this restriction.

library/kani/src/contracts.rs
```diff
@@ -170,4 +170,23 @@
 //!
 //! Unlike `proof_for_contract` multiple `stub_verified` attributes are allowed
 //! on the same proof harness though they must target different functions.
+//!
+//! ## Inductive Verification
+//!
+//! Function contracts by default use inductive verification to efficiently
+//! verify recursive functions. In inductive verification a recursive function
+//! is executed once and every recursive call is instead uses the contract
```
zhassan-aws, line 178:
```suggestion
//! is executed once and every recursive call instead uses the contract
```

library/kani/src/contracts.rs
```diff
@@ -170,4 +170,23 @@
...
+//! replacement. In this way a great many recursive calls can be checked with a
```
zhassan-aws, line 179:
```suggestion
//! replacement. In this way many recursive calls can be checked with a
```
?

library/kani/src/contracts.rs
```diff
@@ -170,4 +170,23 @@
...
+//! single verification pass.
+//!
+//! The downside of inductive verification is that the return value of a
+//! contracted function must implement `kani::Arbitrary`. Due to restrictions to
+//! code generation in proc macros the contract macros cannot determine reliably
```
zhassan-aws, line 184:
```suggestion
//! code generation in proc macros, the contract macros cannot determine reliably
```

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -809,6 +905,97 @@ fn requires_ensures_main(attr: TokenStream, item: TokenStream, is_requires: bool
...
+        self.0 = true;
+    }
+}
+
+/// Try to determine if this function is part of an `impl`.
+///
+/// Detects *methods* by the presence of a receiver argument. Heuristically
+/// detects *associated functions* buy the use of `Self` anywhere.
+///
+/// Why do we need this? It's because if we want to call this `fn`, or any other
```
zhassan-aws, line 948: Nice explanation!

library/kani_macros/src/sysroot/contracts.rs
```diff
@@ -809,6 +905,97 @@ fn requires_ensures_main(attr: TokenStream, item: TokenStream, is_requires: bool
...
+
+    fn visit_receiver(&mut self, _node: &'ast syn::Receiver) {
+        self.0 = true;
+    }
+}
+
+/// Try to determine if this function is part of an `impl`.
+///
+/// Detects *methods* by the presence of a receiver argument. Heuristically
+/// detects *associated functions* buy the use of `Self` anywhere.
```
zhassan-aws, line 946:
```suggestion
/// detects *associated functions* by the use of `Self` anywhere.
```

library/kani_macros/src/sysroot/contracts.rs
zhassan-aws: Looks great, thanks!

library/kani_macros/src/sysroot/contracts.rs
JustusAdam: Thanks 🙏

zhassan-aws (approved):

PR #2810, opened 2023-10-05
Run benchcomp if PR is labeled 'Z-BenchCI' 

### LINKED ISSUE #2606: Automate running benchmarks in PRs that are more likely to impact performance

celinval:
The "Auto Label" workflow correctly label our PRs; however, this action does not trigger a PR 'labeled' workflow. To automatically do that, we now trigger the benchmark workflow from the labeler workflow.

In this PR, we also change the workflow to run after any update to a PR that has the 'Z-BenchCI' label.

Fixes #2606 

**Call-Out**: I still want to cleanup how the jobs will be displayed. In my tests, when the labeler triggers the workflow, it displays as part of the labeler workflow. Need to figure out the best way forward here. Suggestions are welcome.

Here is how it looks today: https://github.com/celinval/kani-dev/actions/runs/6425630976

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -9,6 +9,7 @@
 #  - Invoked by the labeler workflow after a PR update.
 #    - Adding 'Z-BenchCI' is not enough. For more details see:
 #      https://github.com/model-checking/kani/issues/2606#issuecomment-1749706887
+#    - In this case, payload is the same as the callee workflow.
 name: Kani Performance Benchmarks
 on:
   push:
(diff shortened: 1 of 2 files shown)


celinval:
I created this PR from the wrong branch. I'll recreate it later

PR #2820, opened 2023-10-12
Fix panic warning and add arbitrary Duration

celinval:
Add an Arbitrary implementation for Duration and also fix unused variable warnings when panic message uses a variable in its format string.

### DIFF
--- a/library/kani/src/arbitrary.rs
+++ b/library/kani/src/arbitrary.rs
@@ -174,7 +174,8 @@ where
 impl Arbitrary for std::time::Duration {
     fn any() -> Self {
         const NANOS_PER_SEC: u32 = 1_000_000_000;
-        let nanos = u32::any() % NANOS_PER_SEC;
+        let nanos = u32::any();
+        crate::assume(nanos < NANOS_PER_SEC);
         std::time::Duration::new(u64::any(), nanos)
     }
 }
--- a/library/kani/src/arbitrary.rs
+++ b/library/kani/src/arbitrary.rs
@@ -170,3 +170,11 @@ where
         Box::new(T::any())
     }
 }
+
+impl Arbitrary for std::time::Duration {
+    fn any() -> Self {
+        const NANOS_PER_SEC: u32 = 1_000_000_000;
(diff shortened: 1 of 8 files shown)


library/kani/src/arbitrary.rs
```diff
@@ -170,3 +170,11 @@ where
         Box::new(T::any())
     }
 }
+
+impl Arbitrary for std::time::Duration {
+    fn any() -> Self {
+        const NANOS_PER_SEC: u32 = 1_000_000_000;
+        let nanos = u32::any() % NANOS_PER_SEC;
```
zhassan-aws, line 177: How does this perform compared to using an assumption (`nanos < NANOS_PER_SEC`)?

tests/ui/should-panic-attribute/expected-panics/test.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-
+//
+// compile-flags: -Copt-level=1
```
zhassan-aws, line 4: Why was this added?

celinval: Because the expected file checks the number of existing assertions which went up from 2 to 6. However, those are dead assumption.

library/kani/src/arbitrary.rs
celinval: great question... I didn't spend time optimizing this implementation since my use case was fairly simple.

zhassan-aws: I see. I'm worried that a modulo operation with a number that is not 2^n might be expensive.

celinval: I added a performance test. The original version took around 14s and the assume one took close to 12s. So not a huge difference, but assume seems better. I also tried using a conditional assignment instead of a global assumption, but there wasn't any difference, so I'm keeping it simple for now and using the assume version.

zhassan-aws (approved):

PR #2821, opened 2023-10-16
Update CBMC version to 5.94

celinval:
Update CBMC version to 5.94 and change our regression script to read the values from kani-dependencies instead of requiring us to keep two sources of truth.

### DIFF
--- a/kani-dependencies
+++ b/kani-dependencies
@@ -1,4 +1,10 @@
-CBMC_VERSION="5.93.0"
+CBMC_MAJOR="5"
+CBMC_MINOR="94"
+CBMC_VERSION="5.94.0"
+
 # If you update this version number, remember to bump it in `src/setup.rs` too
+CBMC_VIEWER_MAJOR="3"
+CBMC_VIEWER_MINOR="8"
 CBMC_VIEWER_VERSION="3.8"
+
 KISSAT_VERSION="3.1.1"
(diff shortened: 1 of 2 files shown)


kani-dependencies
```diff
@@ -1,4 +1,10 @@
-CBMC_VERSION="5.93.0"
+CBMC_MAJOR="5"
+CBMC_MINOR="94"
+CBMC_VERSION="5.94.0"
```
zhassan-aws, line 3: Can we set this to:
```
CBMC_VERSION="${CBMC_MAJOR}.${CBMC_MINOR}.0"
```
or add `CBMC_PATCH` as well:
```
CBMC_VERSION="${CBMC_MAJOR}.${CBMC_MINOR}.${CBMC_PATCH}"
```
?

kani-dependencies
```diff
@@ -1,4 +1,10 @@
...
+
 # If you update this version number, remember to bump it in `src/setup.rs` too
+CBMC_VIEWER_MAJOR="3"
+CBMC_VIEWER_MINOR="8"
 CBMC_VIEWER_VERSION="3.8"
```
zhassan-aws, line 8: Same here

zhassan-aws (commented):
Thanks for doing this!

kani-dependencies
celinval: I thought about it, but this file is not necessarily a bash script. It can be imported by python scripts as well

kani-dependencies
zhassan-aws: I see.

zhassan-aws (approved):

PR #2828, opened 2023-10-18
Update our CI to run on MacOS 12 instead of 11

### LINKED ISSUE #2827: Replace macos-11 with macos-12 in our CI
**Proposed change:** Use macos-12 (Monterey) instead of macos-11 (Big Sur) in our CI

**Motivation:** Homebrew stopped creating bottles for Big Sur, which means that pre-built CBMC binaries for Big Sur are no longer available on https://github.com/diffblue/homebrew-cbmc/. This is resulting in a significant increase in the CI time for macos-11, because CBMC has to be built from source:

https://github.com/model-checking/kani/actions/runs/6560653897/job/17818740593#step:3:232

The result is that the CI's "Setup Kani Dependencies" step takes ~34 minutes on macos (vs ~6 minutes on Ubuntu).

celinval:
Update the CI to use MacOS 12 instead of MacOS 11. I'm leaving this as a draft until I can validate that the bundle works on MacOS 11.

Resolves #2827 


scripts/setup/macos/install_viewer.sh
```diff
@@ -17,4 +17,6 @@ fi
 set -x
 
 # brew doesn't recognize specific versions of viewer
```
adpaco-aws, line 19: Do we want to keep this comment?

adpaco-aws (approved):
Thanks!

scripts/setup/macos/install_viewer.sh
celinval: I don't see why not.

PR #2837, opened 2023-10-25
Create workflow Kani Extra for checks that run conditionally on labels

### LINKED ISSUE #2606: Automate running benchmarks in PRs that are more likely to impact performance

celinval:
The "Auto Label" workflow correctly label our PRs; however, this action do not trigger a PR 'labeled' workflow. To automatically do that, we now changed the CI structure to have a new workflow with extra jobs that are conditional to labels and files being modified.

This new workflow starts by triggering the auto-label job. The benchmark workflow is only triggered if after the auto-label job, the PR has a Z-BenchCI label.

This also means that the benchmark workflow will now run after any update to a PR that has the 'Z-BenchCI' label.

Resolves #2606

### Call-out

I manually tested these changes in my branch with this dummy PR: https://github.com/celinval/kani-dev/pull/19. The new workflow can be seen [here](https://github.com/celinval/kani-dev/actions/runs/6633993873?pr=19) and the benchcomp run shows up in the list of job as "Kani Extra / Verification Benchmarks / perf-benchcomp".


jaisnan (approved):
Awesome change! Thank you

jaisnan:
> This also means that the benchmark workflow will now run after any update to a PR that has the 'Z-BenchCI' label.

Is this a unwanted side-effect? I thought this was the intention behind this change?

celinval:
> > This also means that the benchmark workflow will now run after any update to a PR that has the 'Z-BenchCI' label.
> 
> Is this a unwanted side-effect? I thought this was the intention behind this change?

Not a side effect. I just wanted to call out that this is more than just refactoring.

PR #2841, opened 2023-10-26
Unify Kani bundle CI jobs

### LINKED ISSUE #2703: Unify release bundle tests
**Proposed change:** We should run the same release bundle workflow for PR and release. The only difference is that during release, at the end of the workflow, we upload the artifacts.

**Motivation:** Having different workflows is harder to maintain and more error prone. I'm not aware of any reason why we should keep them different.

We should also consider restructuring the workflow into separate jobs. One to build the bundle, one that download and test the bundle and a third to publish.

celinval:
We have two different ways today of building release bundles, one in `kani.yml` and another one in `release.yml`, and only the `kani.yml` actually run tests. This is error prone, since the release workflow only runs during a release without validation tests, and the build can diverge from what's being tested in CI.

Instead, we will always run the same workflow, except for the steps that create the release.

Resolves #2703

### Call-outs

1. I've tested this in my fork:
    - CI run when push to main: https://github.com/celinval/kani-dev/actions/runs/6646601378
    - CI run when PR is created: https://github.com/celinval/kani-dev/actions/runs/6657614151
    - CI run when label is created: https://github.com/celinval/kani-dev/actions/runs/6657946688
      - This one failed since I can't publish things to github container registry.
2. Should we keep publishing the docker image? I'm tempted to at least mark this step with `continue-on-error: true` so it doesn't fail the release workflow.


celinval:
Note that to merge this PR, we will need to remove the "Required" check from the jobs that were removed

scripts/ci/Dockerfile.bundle-test-ubuntu-18-04
```diff
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
 # Note: this file is intended only for testing the kani release bundle
+# This docker assumes the following locations:
+# - ./kani_src/: Kani source code
+# - ./kani-*tar.bz/: The Kani release bundle. E.g. `kani-0.39.0-x86_64-unknown-linux-gnu.tar.gz`
```
zhassan-aws, line 7:
```suggestion
# - ./kani-*tar.gz/: The Kani release bundle. E.g. `kani-0.39.0-x86_64-unknown-linux-gnu.tar.gz`
```

zhassan-aws (approved):
LGTM. Thanks.

PR #2849, opened 2023-11-01
Fix path-dependent failures in `assess-artifacts` test

adpaco-aws:
Shorten the tables frames in the `expected` file for the `cargo-kani/assess-artifacts` test. Otherwise, the test will fail when running the regression suite in a shorter-than-usual directory path (e.g., `/tmp/kani`). Came across this problem this morning when preparing for the release.

### Callouts
 * Could also remove the frames/lines altogether.


jaisnan (approved):
Thanks for catching this! 😆

PR #2852, opened 2023-11-01
Do not override `std` library during playback

### LINKED ISSUE #2850: Do not override `std` functions when running concrete playback
For example, when running a concrete playback to figure out why an `assert_eq!()` failed, it is super useful for the user to see the proper assertion message where the values get printed.

celinval:
We override a few std functions to improve verification time or verification results, however, those functions are quite helpful during concrete execution. Thus, avoid overriding functions when executing concrete playback.

Resolves #2850


tests/script-based-pre/playback_print/print_vars.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// compile-flags: --edition 2021
+// kani-flags: playback
```
adpaco-aws, line 5: I don't think these are being used, but if you wanted to leave them for documentation purposes we should add `-Z concrete-playback` to `kani-flags`.

tests/script-based-pre/playback_print/playback_print.sh
```diff
@@ -0,0 +1,40 @@
...
+
+echo "[TEST] Generate test..."
+kani ${RS_FILE} -Z concrete-playback --concrete-playback=inplace
+
+echo "[TEST] Run test..."
+kani playback -Z concrete-playback ${RS_FILE} 2>&1 | tee output.log
+
+echo "------ Check output ----------"
+
+set -e
```
adpaco-aws, line 29: Why not `set -e` earlier?

tests/script-based-pre/playback_print/playback_print.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# Test that concrete playback -Z concrete-playback does not override std print
+# functions
```
adpaco-aws, line 5:
```suggestion

# Test that concrete playback does not override std print functions
```
or
```suggestion

# Test that `kani playback -Z concrete-playback` does not override std print functions
```

zhassan-aws (approved):
Thanks

tests/script-based-pre/playback_print/playback_print.sh
celinval: Because Kani exit status is non-zero when the verification fail.

tests/script-based-pre/playback_print/print_vars.rs
celinval, line 5:
```suggestion
// Harness used to test that playback do not override assertion and panic functions.
```

PR #2855, opened 2023-11-03
update in visualizing trace option

muqsit-azeem:
`--visualize` option does not work, it needs to include `--enable-unstable`  as well. 


zhassan-aws (approved):
Thanks!

PR #2789, opened 2023-09-25
Deprecate any_slice

zhassan-aws:
### Description of changes: 

See #2788 for details.

### Resolved issues:

Resolves #2788 
Resolves #2634 

### Related RFC:

Optional #ISSUE-NUMBER.

### Testing:

* How is this change tested? Added one test

* Is this a refactor change? No

### Checklist
- [X] Each commit message has a non-empty body, explaining why the change was made
- [X] Methods or procedures are documented
- [X] Regression or unit tests are included, or existing tests cover the modified code
- [X] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #2860, opened 2023-11-06
Delete `any_slice` which has been deprecated since Kani 0.38.0.

zhassan-aws:
Kani's `any_slice` function (and related code) has been deprecated since Kani 0.38.0 (see #2789). This PR officially obsoletes it by deleting the code.


library/kani/src/slice.rs
```diff
@@ -1,8 +1,6 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-use crate::{any, assume, Arbitrary};
-use std::alloc::{alloc, dealloc, Layout};
-use std::ops::{Deref, DerefMut};
+use crate::{any, assume};
```
feliperodri, line 3: Are there any users relying on `any_slice_of_array`, `any_slice_of_array_mut`, or `any_range`? Should we deprecate these functions as well?

zhassan-aws: I'm not sure if there are any users that use them, but these functions do not suffer from the same issues that `any_slice` had (being too conservative at a severe expense of performance), and they might be convenient to use. So I prefer if we keep them.

feliperodri (approved):

PR #2865, opened 2023-11-07
expect() does not take format strings

matthiaskrgr:
This would just print 
````
Function `{func}` should've been declared before usage'
````
on panic, it just now occurred to me that `func` is supposed to be content of a var :smile:


zhassan-aws (approved):
Good catch! Thanks @matthiaskrgr!

PR #2871, opened 2023-11-11
Use stable_mir api instead of internal api for analysis

ouz-a:
This update introduces several `stable_mir` APIs to `Kani`. Previously, we relied on internal APIs to accomplish our goals. However, with the recent enhancements to `stable_mir`, we no longer need these internal APIs. This pull request hopefully marks the beginning of a migration towards `stable_mir` for `Kani`.

### DIFF
--- a/kani-compiler/Cargo.toml
+++ b/kani-compiler/Cargo.toml
@@ -30,6 +30,7 @@ tracing-subscriber = {version = "0.3.8", features = ["env-filter", "json", "fmt"
 default = ['cprover']
 cprover = ['cbmc', 'num', 'serde']
 write_json_symtab = []
+stable_mir = []
 
 [package.metadata.rust-analyzer]
 # This package uses rustc crates.
--- a/kani-compiler/Cargo.toml
+++ b/kani-compiler/Cargo.toml
@@ -30,7 +30,6 @@ tracing-subscriber = {version = "0.3.8", features = ["env-filter", "json", "fmt"
 default = ['cprover']
 cprover = ['cbmc', 'num', 'serde']
 write_json_symtab = []
-stable_mir = []
 
 [package.metadata.rust-analyzer]
 # This package uses rustc crates.
--- a/kani-compiler/src/kani_middle/analysis.rs
+++ b/kani-compiler/src/kani_middle/analysis.rs
@@ -10,9 +10,9 @@
 use rustc_middle::mir::mono::MonoItem as InternalMonoItem;
 use rustc_middle::ty::TyCtxt;
 use rustc_smir::rustc_internal;
+use stable_mir::mir::mono::MonoItem;
 use stable_mir::mir::{
-    visit::Location, MirVisitor, Rvalue, Statement,
-    StatementKind, Terminator, TerminatorKind,
+    visit::Location, MirVisitor, Rvalue, Statement, StatementKind, Terminator, TerminatorKind,
 };
 use std::collections::HashMap;
 use std::fmt::Display;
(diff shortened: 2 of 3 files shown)


kani-compiler/Cargo.toml
```diff
@@ -30,6 +30,7 @@ tracing-subscriber = {version = "0.3.8", features = ["env-filter", "json", "fmt"
 default = ['cprover']
 cprover = ['cbmc', 'num', 'serde']
 write_json_symtab = []
+stable_mir = []
```
celinval, line 33: I don't think we need this config for this PR.

kani-compiler/src/kani_middle/analysis.rs
```diff
@@ -20,19 +23,20 @@ use std::fmt::Display;
...
 ///  - Number of instructions per type.
 ///  - Total number of MIR instructions.
-pub fn print_stats<'tcx>(tcx: TyCtxt<'tcx>, items: &[MonoItem<'tcx>]) {
+pub fn print_stats<'tcx>(tcx: TyCtxt<'tcx>, items: &[InternalMonoItem<'tcx>]) {
     let item_types = items.iter().collect::<Counter>();
     let visitor = items
         .iter()
         .filter_map(|&mono| {
-            if let MonoItem::Fn(instance) = mono {
+            if let InternalMonoItem::Fn(instance) = mono {
```
celinval, line 31: Can you call stable on the internal MonoItem here instead of the body?

celinval (commented):
That's awesome and quick! Thanks!

I think you'll need to initialize stable mir for this to work. For now, you can wrap the logic of the`print_stats` in a closure, and invoke `rustc_internal::run`. Also, I forgot to mention, and I don't know if you've figured out already, but one way to test this code is to run Kani with `--verbose`.

ouz-a:
Well I tried few things but It wasn't really clear to me which things supposed to fail and which things supposed to not fail, I run `cargo kani --verbose` from that all I got was reachability report.

celinval:
@ouz-a  No worries. Feel free to ping me offline if you want to know more. The important thing was to see the report when you ran with --verbose. Thanks!

kani-compiler/src/kani_middle/analysis.rs
```diff
@@ -115,28 +120,28 @@ impl<T: Into<Key>> FromIterator<T> for Counter {
 #[derive(Debug, Eq, Hash, PartialEq)]
 struct Key(pub &'static str);
 
-impl<'tcx> From<&MonoItem<'tcx>> for Key {
-    fn from(value: &MonoItem) -> Self {
+impl<'tcx> From<&InternalMonoItem<'tcx>> for Key {
```
celinval, line 123: Here we should also be using stable MonoItem

celinval (approved):

kani-compiler/src/kani_middle/analysis.rs
```diff
@@ -20,29 +23,32 @@ use std::fmt::Display;
...
-    eprintln!("Expressions:\n{}", visitor.exprs);
-    eprintln!("-------------------------------------------")
+pub fn print_stats<'tcx>(tcx: TyCtxt<'tcx>, items: &[InternalMonoItem<'tcx>]) {
+    rustc_internal::run(tcx, || {
+        let item_types: Vec<MonoItem> = items.iter().map(rustc_internal::stable).collect();
+        let item_types = item_types.iter().collect::<Counter>();
+        let visitor = items
+            .iter()
+            .filter_map(|&mono| {
+                if let MonoItem::Fn(instance) = rustc_internal::stable(&mono) {
```
celinval, line 33:
```suggestion
        let items: Vec<MonoItem> = items.iter().map(rustc_internal::stable).collect();
        let item_types = items.iter().collect::<Counter>();
        let visitor = items
            .iter()
            .filter_map(|mono| {
                if let MonoItem::Fn(instance) = mono {
```

celinval (approved):
Thanks again!

PR #2877, opened 2023-11-14
install_cbmc.sh: build from source on non-x86_64

bennofs:
CBMC only provides binaries for x86_64, but it runs fine on other arches (like aarch64) Build from source in that case

### DIFF
--- a/scripts/setup/ubuntu/install_cbmc.sh
+++ b/scripts/setup/ubuntu/install_cbmc.sh
@@ -15,7 +15,7 @@ fi
 UBUNTU_VERSION=$(lsb_release -rs)
 MAJOR=${UBUNTU_VERSION%.*}
 
-if [[ "${MAJOR}" -gt "18" ]]
+if [[ "${MAJOR}" -gt "18" ]] && [[ $(uname -m) = "x86_64" ]]
 then
   FILE="ubuntu-${UBUNTU_VERSION}-cbmc-${CBMC_VERSION}-Linux.deb"
   URL="https://github.com/diffblue/cbmc/releases/download/cbmc-${CBMC_VERSION}/$FILE"
@@ -29,7 +29,7 @@ then
   exit 0
 fi
 
-# Binaries are no longer released for 18.04, so build from source
+# There are no binaries for 18.04 or for non-x86_64, so build from source
 
 WORK_DIR=$(mktemp -d)
 git clone \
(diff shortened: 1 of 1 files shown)


zhassan-aws:
Thanks @bennofs for the contribution!

@adpaco-aws can you review this PR since you're the most familiar with this code?

The failure is unrelated to the PR, and is due to a CBMC viewer/python version issue.

scripts/setup/ubuntu/install_cbmc.sh
```diff
@@ -15,7 +15,7 @@ fi
 UBUNTU_VERSION=$(lsb_release -rs)
 MAJOR=${UBUNTU_VERSION%.*}
 
-if [[ "${MAJOR}" -gt "18" ]]
+if [[ "${MAJOR}" -gt "18" ]] && [[ $(uname -m) = "x86_64" ]]
```
celinval, line 18: I have to say... my ubuntu prints an empty line when I execute `uname -m`

adpaco-aws (approved):
Thanks @bennofs ! Makes sense, we should build from source in that case.

scripts/setup/ubuntu/install_cbmc.sh
adpaco-aws: Well, that's surprising. Do you think `uname` is not reliable enough?
AFAIK `arch` is an alternative, but funnily enough it produces `i386` on my 2019 Macbook Pro 😬

celinval: Yes, that's what I've been thinking. It will install from source at least on my machine every time I run this script. :disappointed: Anyway, it's likely that my machine is a bit messed up, so I can create a little workaround for me.

bennofs: Since we're on ubuntu anyway, perhaps we should use `dpkg --print-architecture`?

bennofs: Since we're on ubuntu anyway, perhaps we should use `dpkg --print-architecture`?

bennofs: I tested on both aarch64 and amd64, and `dpkg --print-architecture` works fine. Changed to that, hopefully this is reliable (I think it should be, as that's what we care about anyway: we want to install a amd64 deb so dpkg arch should be amd64)

adpaco-aws (approved):

scripts/setup/ubuntu/install_cbmc.sh
adpaco-aws: Thank you @bennofs for finding and testing an alternative!

PR #2879, opened 2023-11-14
set python to 3.7 in docker container and release action

remi-delmas-3000:
An attempt to fix a release build issue due to cbmc-viewer depending on voluptuous which now depends on python 3.7 for type annotations. 

This adds a ppa in the dockerfile and github release action for ubuntu 18.04 and installs python 3.7. 

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -68,9 +68,9 @@ jobs:
           df -h
 
       # This is required before checkout because the container does not
-      # have Git installed, so cannot run checkout action. T
+      # have Git installed, so cannot run checkout action.
       # The checkout action requires Git >=2.18 and python 3.7, so use the Git maintainers' PPA.
-      # and the "deadsnakes" PPA, as the default version of python on ubuntu is 3.11.
+      # and the "deadsnakes" PPA, as the default version of python on ubuntu 22.04 is Python 3.10
       - name: Install system dependencies
         run: |
           apt-get update
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,6 +80,7 @@ jobs:
           apt-get install -y \
             build-essential bash-completion curl lsb-release sudo g++ gcc flex \
             bison make patch git python3.7 python3.7-dev python3.7-distutils
+          update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.7 1
           curl -s https://bootstrap.pypa.io/get-pip.py -o get-pip.py
           python3 get-pip.py --force-reinstall
           rm get-pip.py
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,10 +75,14 @@ jobs:
           apt-get update
           apt-get install -y software-properties-common apt-utils
           add-apt-repository ppa:git-core/ppa
+          add-apt-repository ppa:deadsnakes/ppa
           apt-get update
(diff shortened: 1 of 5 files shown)


feliperodri:
Is there a way to also upgrade the "workaround" for Ubuntu 18.04.

https://github.com/model-checking/kani/blob/d1a038fcf29b6f5caa92f7cf8cb4b40edd811733/src/os_hacks.rs#L53

feliperodri:
We also need to update our dependencies list in the Installation guidelines.

https://github.com/model-checking/kani/blob/d1a038fcf29b6f5caa92f7cf8cb4b40edd811733/docs/src/install-guide.md?plain=1#L17

remi-delmas-3000:
> Is there a way to also upgrade the "workaround" for Ubuntu 18.04.

I am not modifying the OS hacks 18.04 because they are meant to be applied only when the default pip 9.x and python 3.6.x versions are detected on the host. Since we now require python 3.7 they should not have to be applied anymore.

adpaco-aws (commented):
> Since we now require python 3.7 they should not have to be applied anymore.

Why not remove the hacks then?

remi-delmas-3000:
> Why not remove the hacks then?
we would need to add a python version check too

jaisnan:
> > Why not remove the hacks then?
> > we would need to add a python version check too

As discussed offline, I am adding a check for the python version directly, and removing the hacks. @remi-delmas-3000 can you confirm if this is the right plan?

adpaco-aws (commented):
The Python version checks will abort the installation if the minimum version isn't met. But the setup code isn't expected to do this, what it should do in any case is attempt to solve these problems for you.

jaisnan:
> The Python version checks will abort the installation if the minimum version isn't met. But the setup code isn't expected to do this, what it should do in any case is attempt to solve these problems for you.

I did attempt to solve the problem, but even upgrading to 3.7 is a tough job to accomplish through rust's interface since even python 3.7's [reached end-of-life support](https://devguide.python.org/versions/) ☹️. Should we skip 3.7 and directly upgrade the user's python to 3.8?

adpaco-aws:
> I assumed the answer to be no, since anyone using python 3.6 or older is likely doing it for very specific reasons, and just informing them seemed like a good idea.

I agree that we shouldn't upgrade python in that case. But here you're not "just informing" the users, you're failing the installation...

Again, why not *just* remove the hacks? You can leave improvements for later.

.github/workflows/release.yml
```diff
@@ -68,17 +68,23 @@ jobs:
           df -h
 
       # This is required before checkout because the container does not
-      # have Git installed, so cannot run checkout action. The checkout
-      # action requires Git >=2.18, so use the Git maintainers' PPA.
+      # have Git installed, so cannot run checkout action. T
```
adpaco-aws, line 71:
```suggestion
      # have Git installed, so cannot run checkout action.
```

.github/workflows/release.yml
```diff
@@ -68,17 +68,23 @@ jobs:
...
+      # The checkout action requires Git >=2.18 and python 3.7, so use the Git maintainers' PPA.
+      # and the "deadsnakes" PPA, as the default version of python on ubuntu is 3.11.
```
adpaco-aws, line 73: Which ubuntu?

adpaco-aws (approved):

PR #2882, opened 2023-11-16
Make `codegen_span` stable

ouz-a:
Follow up pr to https://github.com/model-checking/kani/pull/2871 as `stable_mir::ty::Span` have ability to `get_filename` and `get_lineinfo` we could start adopting it.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/span.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/span.rs
@@ -3,16 +3,21 @@
 
 //! MIR Span related functions
 
-use crate::{codegen_cprover_gotoc::GotocCtx, kani_middle::SourceLocation};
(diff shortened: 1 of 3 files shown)


celinval (approved):

celinval:
@ouz-a, I believe you need to initialize the StableMIR components. In this case, I suggest that you move the initial initialization to be inside [`compiler_interface::codegen_crate()`](https://github.com/model-checking/kani/blob/b0301a3f4b48b95381de290f370e45b137e3fec5/kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs#L205). That case, the entire codegen step can be used with the same SMIR session.

Note that you will have to remove the initialization inside analysis. Thanks!

ouz-a:
**Update**: We are waiting for `Opaque` fix in this https://github.com/rust-lang/rust/pull/118172 pr to land, before continuing the work on this as this is currently blocked by how we store strings.

Issue #2589, opened 2023-07-07
Invalid stub cause different ICE while monomorphizing the code


celinval:
I tried this code:

```rust
fn original<T>() {}

trait Dummy {
    const TRUE: bool = true;
}

fn stub<T: Dummy>() {
    // Do nothing.
    assert!(T::TRUE)
}

#[kani::proof]
#[kani::stub(original, stub)]
fn check_mismatch() {
    original::<String>();
}
```

using the following command line invocation:

```
kani --enable-unstable --enable-stubbing
```

with Kani version: 0.31.0 (dev)

I expected to see this happen: Kani should print a trait mismatch error

Instead, this happened: Kani compiler crashed with the following error:

```
error: internal compiler error: kani-compiler/src/kani_middle/reachability.rs:443:54: Unexpected polymorphic constant: Unevaluated(UnevaluatedConst { def: DefId(0:16 ~ trait_mismatch[1bfd]::Dummy::TRUE), substs: [std::string::String], promoted: None }, bool)
  --> trait_mismatch.rs:36:5
   |
36 |     assert!(T::TRUE)
   |     ^^^^^^^^^^^^^^^^
   |
   = note: this error: internal compiler error originates in the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)

thread 'rustc' panicked at 'Box<dyn Any>', /rustc/fe7454bf439c93cbe9ac8a8f7fcfacd5a40244c2/compiler/rustc_errors/src/lib.rs:994:33
```

<details><summary> Stack backtrace </summary>

```
stack backtrace:
   0: std::panicking::begin_panic
             at /rustc/fe7454bf439c93cbe9ac8a8f7fcfacd5a40244c2/library/std/src/panicking.rs:625:12
   1: std::panic::panic_any
             at /rustc/fe7454bf439c93cbe9ac8a8f7fcfacd5a40244c2/library/std/src/panic.rs:63:5
   2: rustc_errors::HandlerInner::span_bug
             at /rustc/fe7454bf439c93cbe9ac8a8f7fcfacd5a40244c2/compiler/rustc_errors/src/lib.rs:1588:9
   3: rustc_errors::Handler::span_bug
             at /rustc/fe7454bf439c93cbe9ac8a8f7fcfacd5a40244c2/compiler/rustc_errors/src/lib.rs:994:9
   4: rustc_middle::util::bug::opt_span_bug_fmt::{{closure}}
[...]

PR #2894, opened 2023-11-25
Change reachability module to use StableMIR

celinval:
Rewrite the reachability module to use Stable APIs wherever possible. Note that in StableMIR the instance body is already monomorphized and constants are already evaluated, which simplifies the code for most of it, except to handle stubbing issue #2589.

For the stubbing issue, we still use a mix of stable and internal APIs to detect an invalid monomorphization.

### Call-outs

1. ~~This change requires the newest version of the Rust toolchain. Thus, it is upgrading the version as well.~~
2. ~~Creating this as a draft until https://github.com/rust-lang/rust/pull/118274 is merged.~~
3. I ran s2n-quic manually to assess performance and no significant difference.
  - s2n-quic-core:  36.86 vs 37.19
  - s2n-quic-platform: 16.17 vs 16.19


tests/cargo-kani/stubbing-double-extern-path/harness/expected
```diff
@@ -1 +1 @@
-error[E0391]: cycle detected when optimizing MIR for `crate_a::assert_true`
```
celinval, line 1: FYI, this test checks that there is a cycle so the error printed here depends on which method we visit first. But the cycle is correctly detected no matter what.

kani-compiler/src/kani_middle/reachability.rs
```diff
@@ -13,146 +13,135 @@
...
+        // Filter regular items.
+        crate_items
+            .iter()
+            .filter_map(|item| {
+                // Only collect monomorphic items.
+                Instance::try_from(*item).ok().and_then(|instance| {
+                    let def_id = rustc_internal::internal(item);
+                    predicate(tcx, def_id)
+                        .then_some(InternalMonoItem::Fn(rustc_internal::internal(&instance)))
+                })
```
adpaco-aws, line 92: I'm not able to establish how this corresponds to what we had. What does `internal` do?

kani-compiler/src/kani_middle/reachability.rs
```diff
@@ -13,146 +13,135 @@
...
+    collected: FxHashSet<MonoItem>,
     /// Items enqueued for visiting.
-    queue: Vec<MonoItem<'tcx>>,
+    queue: Vec<MonoItem>,
     #[cfg(debug_assertions)]
-    call_graph: debug::CallGraph<'tcx>,
+    call_graph: debug::CallGraph,
 }
 
+// TODO: Need to figure out how to handle stubbing errors.
```
adpaco-aws, line 144: What's this referring to? Please open an issue (unless we have it) and refer to it here.

kani-compiler/src/kani_middle/stubbing/mod.rs
```diff
@@ -28,3 +35,83 @@ pub fn harness_stub_map(
...
     stub_pairs
 }
+
+/// Validate that an instance body can be instantiated.
+///
+/// Stubbing may cause an instance to not be correctly instantiated since we delay checking its
+/// generic bounds.
+///
+/// In stable mir, trying to retrieve an Instance::body() will ICE if we cannot evaluate a
+/// constant as expected. For now, use internal APIs to anticipate this issue.
```
adpaco-aws, line 45:
```suggestion
/// In stable MIR, trying to retrieve an `Instance::body()` will ICE if we cannot evaluate a
/// constant as expected. For now, use internal APIs to anticipate this issue.
```

kani-compiler/src/kani_middle/reachability.rs
```diff
@@ -194,146 +184,117 @@ impl<'tcx> MonoItemsCollector<'tcx> {
...
-    fn collect_vtable_methods(&mut self, concrete_ty: Ty<'tcx>, trait_ty: Ty<'tcx>) {
+    fn collect_vtable_methods(&mut self, concrete_ty: Ty, trait_ty: Ty) {
         trace!(?concrete_ty, ?trait_ty, "collect_vtable_methods");
-        assert!(!concrete_ty.is_trait(), "Expected a concrete type, but found: {concrete_ty:?}");
-        assert!(trait_ty.is_trait(), "Expected a trait: {trait_ty:?}");
-        if let TyKind::Dynamic(trait_list, ..) = trait_ty.kind() {
+        let concrete_kind = concrete_ty.kind();
+        let trait_kind = trait_ty.kind();
+
+        assert!(!concrete_kind.is_trait(), "Expected a concrete type, but found: {concrete_ty:?}");
```
adpaco-aws, line 244:
```suggestion
        assert!(!concrete_kind.is_trait(), "expected a concrete type, but found `{concrete_ty:?}`");
```

kani-compiler/src/kani_middle/reachability.rs
```diff
@@ -194,146 +184,117 @@ impl<'tcx> MonoItemsCollector<'tcx> {
...
+        assert!(trait_kind.is_trait(), "Expected a trait: {trait_ty:?}");
```
adpaco-aws, line 245:
```suggestion
        assert!(trait_kind.is_trait(), "expected a trait `{trait_ty:?}`");
```

kani-compiler/src/kani_middle/reachability.rs
```diff
@@ -518,22 +422,22 @@ impl<'a, 'tcx> MirVisitor<'tcx> for MonoItemsFnCollector<'a, 'tcx> {
...
                                 // even the return type, for instance for a
                                 // trait like `FromIterator`.
-                                let generic_ty = outer_args[0].ty(self.body, tcx).peel_refs();
-                                let receiver_ty = tcx.instantiate_and_normalize_erasing_regions(
-                                    substs,
-                                    ParamEnv::reveal_all(),
-                                    EarlyBinder::bind(generic_ty),
-                                );
+                                let receiver_ty = args.0[0].expect_ty();
+                                trace!(?receiver_ty, ?callee, "**** receiver:");
```
adpaco-aws, line 426: Looks much simpler than what we used to do. Do we want to keep the `trace!` call?

adpaco-aws (commented):
It's looking really good. I wonder if we could add links to the stable MIR project or its APIs so it's clear what's going on here. But I'm not sure how much you'll use these APIs in other parts of the compiler, maybe it's better to add that elsewhere.

kani-compiler/src/kani_middle/reachability.rs
celinval: No. Also a debug artifact that I need to remove. :)

zhassan-aws:
> I wonder if we could add links to the stable MIR project or its APIs so it's clear what's going on here.

I second this request. It's a bit difficult for me to understand the changes given my lack of knowledge of how the stable/internal MIR APIs compare.

celinval:
> > I wonder if we could add links to the stable MIR project or its APIs so it's clear what's going on here.
> 
> I second this request. It's a bit difficult for me to understand the changes given my lack of knowledge of how the stable/internal MIR APIs compare.

Sure. Right now `stable_mir` and `rustc_smir` are new crates in the rust compiler itself. So there's nothing new, and you can apply the same logic as you currently use to check APIs from `rustc_span` and `rustc_middle`.

I will add more information to the code and to [`main.rs`](https://github.com/model-checking/kani/blob/main/kani-compiler/src/main.rs), since this code will start showing up in different parts of our compiler.

kani-compiler/src/kani_middle/reachability.rs
celinval: This code collects all items in the current crate (`all_local_items`), and tries to instantiate them (via `Instance::try_from()`). Only the non-generic code will be successfully converted.

For those that were non-generic, we convert the stable Instance to the internal version, so we can invoke the closure that was passed as an argument which still expects the internal instance.

celinval:
@zhassan-aws and @adpaco-aws, I added some documentation to our wiki instead. I hope that's what you were looking for. Cheers!

docs/src/stable_mir.md
```diff
@@ -0,0 +1,94 @@
+# Transition to StableMIR
```
zhassan-aws, line 1: Very detailed and informative! Thanks for writing this.

zhassan-aws (approved):

PR #2902, opened 2023-11-29
[Boogie Backend] Use the original names of variables when generating Boogie

zhassan-aws:
For MIR variables that map to a user variable in the source code, use their name in the source code in the generated Boogie.

If there are multiple occurrences (due to SSA), append a `_i` to the variable name, where `i` is 0, 1, ...

This change allows one to refer to the original variable name when adding contracts (e.g. loop invariants) in the output Boogie.

Example:

For the following Rust program:

```rust
#[kani::proof]
fn main() {
    let mut x = 41;
    let mut y = 43;
    x = 42;
    y = 42;
    kani::assert(x == y, "");
}
```
before this change, the generated Boogie was:
```
// Procedures:
procedure _RNvCsjrUp1z5o1KH_5test64main() 
{
  var _1: int;
  var _2: int;
  var _4: bool;
  var _5: int;
  var _6: int;
  _1 := 41;
  _2 := 43;
  _1 := 42;
  _2 := 42;
  _5 := _1;
  _6 := _2;
  _4 := (_5 == _6);
  assert _4;
  return;
}
```
but with this change, the generated Boogie is:
```
// Procedures:
procedure _RNvCsjrUp1z5o1KH_5test64main() 
{
  var x: int;
  var y: int;
  var _4: bool;
  var _5: int;
  var _6: int;
  x := 41;
  y := 43;
  x := 42;
  y := 42;
  _5 := x;
  _6 := y;
  _4 := (_5 == _6);
  assert _4;
  return;
}
```

### DIFF
--- a/kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
+++ b/kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
@@ -5,12 +5,13 @@ use std::io::Write;
 
 use crate::kani_queries::QueryDb;
 use boogie_ast::boogie_program::{BinaryOp, BoogieProgram, Expr, Literal, Procedure, Stmt, Type};
+use rustc_data_structures::fx::FxHashMap;
 use rustc_middle::mir::interpret::Scalar;
 use rustc_middle::mir::traversal::reverse_postorder;
 use rustc_middle::mir::{
     BasicBlock, BasicBlockData, BinOp, Body, Const as mirConst, ConstOperand, ConstValue,
     HasLocalDecls, Local, Operand, Place, Rvalue, Statement, StatementKind, Terminator,
-    TerminatorKind,
+    TerminatorKind, VarDebugInfoContents,
 };
 use rustc_middle::span_bug;
 use rustc_middle::ty::layout::{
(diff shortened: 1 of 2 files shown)


kani-compiler/src/codegen_boogie/context/boogie_ctx.rs
```diff
@@ -74,26 +76,65 @@ impl<'tcx> BoogieCtx<'tcx> {
...
+        let mut name_occurrences: FxHashMap<String, usize> = FxHashMap::default();
+        let mir = bcx.tcx.instance_mir(instance.def);
+        let ldecls = mir.local_decls();
+        for local in ldecls.indices() {
+            let debug_info = mir.var_debug_info.iter().find(|info| match info.value {
+                VarDebugInfoContents::Place(p) => p.local == local && p.projection.len() == 0,
+                VarDebugInfoContents::Const(_) => false,
+            });
+            let name = if let Some(debug_info) = debug_info {
+                //let base_name = format!("{}", debug_info.name);
```
jaisnan, line 98: Should this line be removed?

zhassan-aws: Yes, thanks for catching that.

jaisnan (approved):
LGTM from reading the code, but if you could include an example in the description of the PR or a test case, it would be great.

zhassan-aws:
> LGTM from reading the code, but if you could include an example in the description of the PR or a test case, it would be great.

Added an example to the PR description.

PR #2905, opened 2023-11-30
Add disk space output to debug occasional test failures

tautschnig:
Despite freeing up disk space early on, we saw occasional failures like https://github.com/model-checking/kani/actions/runs/7030685174/job/19130721957 that look like running out of disk space.

This change should be reverted once we have diagnosed the problem.


tautschnig:
The CI run here suggests that we do have 35 GB of free disk space, so should be fine. Given that we have these sporadic failures, I'd now ask for reviews to get this merged so that we have more diagnostics upon the next failure that we see.

jaisnan (approved):
Thanks!

Issue #45, opened 2021-04-13
Cleanup: consistent names for typing contexts


danielsn:
Rename typing contexts to be consistent with:
- tcx = MIR typing context
- gcx = GOTO typing context

PR #2910, opened 2023-12-02
Migrate hooks and place modules to use mostly StableMIR APIs

celinval:
The main changes needed to make this migration besides a few method call tweaks were:
1. Adapt `FunctionCtx` to cache information about the StableMIR version of instance and its body.
    - I also cleaned up how we were handling basic blocks which were unnecessary.
2. Created a new ty_stable module that provide stable versions to retrieve type information from StableMIR.
    - I decided to keep these separate so it is cleaner for now. I foresee that the type module will still rely on internal APIs for the next little while, so separating them here made sense to me.
3. Since `Place` when retrieved from StableMIR body already comes monomorphized, I modified the existing `codegen_place` to preemptively monomorphize Place before converting it to a Stable version and invoking `codegen_place_stable`.

### Call-outs

Leaving this as a draft for now since this still depends on the following PRs to be merged into the Rust compiler:
  - https://github.com/rust-lang/rust/pull/118524
  - https://github.com/rust-lang/rust/pull/118516


kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -242,54 +240,61 @@ impl<'tcx> GotocCtx<'tcx> {
     fn codegen_field(
         &mut self,
         parent_expr: Expr,
-        parent_ty_or_var: TypeOrVariant<'tcx>,
-        field: &FieldIdx,
-        field_ty_or_var: TypeOrVariant<'tcx>,
+        parent_ty_or_var: TypeOrVariant,
+        field_idx: FieldIdx,
```
adpaco-aws, line 244: Is passing `field_idx` by value here intended?

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -242,54 +240,61 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                    | ty::Uint(_)
-                    | ty::Float(_)
-                    | ty::FnPtr(_)
-                    | ty::Never
-                    | ty::FnDef(..)
-                    | ty::CoroutineWitness(..)
-                    | ty::Foreign(..)
-                    | ty::Dynamic(..)
-                    | ty::Bound(..)
-                    | ty::Placeholder(..)
```
adpaco-aws, line 265: What happened to `ty::Placeholder`?

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -334,21 +340,15 @@ impl<'tcx> GotocCtx<'tcx> {
...
     /// what to return.
-    fn codegen_simd_field(
-        &mut self,
-        parent_expr: Expr,
-        field: FieldIdx,
-        field_ty: Ty<'tcx>,
-    ) -> Expr {
-        let field_ty = self.monomorphize(field_ty);
-        if matches!(field_ty.kind(), ty::Array { .. }) {
+    fn codegen_simd_field(&mut self, parent_expr: Expr, field: FieldIdx, field_ty: Ty) -> Expr {
```
adpaco-aws, line 343: Please let us `field_idx` for value of the `FieldIdx` type.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -365,18 +365,20 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            TyKind::RigidTy(RigidTy::RawPtr(inner, _)) => self
+                .codegen_local_fndef(inner)
                 .map(|f| if f.can_take_address_of() { f.address_of() } else { f }),
             // A local can be a boxed function pointer
-            ty::Adt(def, _) if def.is_box() => {
-                let boxed_ty = self.codegen_ty(ty);
-                self.codegen_local_fndef(ty.boxed_ty())
+            TyKind::RigidTy(RigidTy::Adt(def, args)) if def.is_box() => {
+                let boxed_ty = self.codegen_ty_stable(ty);
+                self.codegen_local_fndef(*args.0[0].ty().unwrap())
```
adpaco-aws, line 381: I don't know what `*args.0[0].ty().unwrap()` is... can we add an intermediate step?

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -652,40 +673,47 @@ impl<'tcx> GotocCtx<'tcx> {
...
             _ => result,
         }
     }
 
+    pub fn codegen_place(
+        &mut self,
+        place: &PlaceInternal<'tcx>,
+    ) -> Result<ProjectedPlace, UnimplementedData> {
+        self.codegen_place_stable(&StableConverter::convert_place(self, *place))
+    }
```
adpaco-aws, line 706: I don't think you need this anymore?

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -752,6 +780,10 @@ impl<'tcx> GotocCtx<'tcx> {
     }
 }
 
+fn is_box(ty: Ty) -> bool {
+    matches!(ty.kind(), TyKind::RigidTy(RigidTy::Adt(def, _)) if def.is_box())
+}
+
```
adpaco-aws, line 786: Shouldn't these wrappers live in the compiler?

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -562,32 +577,38 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        &ty.kind()
                     ),
                 };
-                let layout = self.layout_of(t);
+                let layout = self.layout_of(rustc_internal::internal(ty));
                 let expr = match &layout.variants {
                     Variants::Single { .. } => before.goto_expr,
                     Variants::Multiple { tag_encoding, .. } => match tag_encoding {
                         TagEncoding::Direct => {
-                            let cases = if t.is_coroutine() {
```
adpaco-aws, line 586: Let's have `is_coroutine` like we do `is_box`.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
adpaco-aws: Not saying "don't add them here" just that we should eventually migrate them.

kani-compiler/src/codegen_cprover_gotoc/codegen/ty_stable.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+/// Convert internal rustc's structs into StableMIR ones.
+///
+/// The body of a StableMIR instance already comes monomorphized, which is different from rustc's
+/// internal representation. To allow us to migrate parts of the code generation stage with
+/// smaller PRs, we have to instantiate rustc's components when converting them to stable.
+///
+/// Once we finish migrating the entire function code generation, we can remove this code.
```
adpaco-aws, line 47: Do we have a tracking issue/milestone we could link here?

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
```diff
@@ -3,26 +3,26 @@
...
+use stable_mir::mir::Body;
+use stable_mir::CrateDef;
 
 /// This structure represents useful data about the function we are currently compiling.
 #[derive(Debug)]
 pub struct CurrentFnCtx<'tcx> {
     /// The GOTO block we are compiling into
     block: Vec<Stmt>,
-    /// The current MIR basic block
-    current_bb: Option<BasicBlock>,
```
adpaco-aws, line 17: How do we keep track of `current_bb` now? Is it kept by the `tcx` in stableMIR?

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
```diff
@@ -3,26 +3,26 @@
...
     /// The codegen instance for the current function
-    instance: Instance<'tcx>,
+    instance: Instance,
     /// The crate this function is from
     krate: String,
-    /// The goto labels for all blocks
-    labels: Vec<String>,
-    /// The mir for the current instance
-    mir: &'tcx Body<'tcx>,
+    /// The mir for the current instance. This is using the internal representation.
```
adpaco-aws, line 22:
```suggestion
    /// The MIR for the current instance. This is using the internal representation.
```

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
```diff
@@ -3,26 +3,26 @@
...
+    mir: &'tcx InternalBody<'tcx>,
+    /// The mir for the current instance.
```
adpaco-aws, line 24:
```suggestion
    /// The MIR for the current instance.
```

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
```diff
@@ -3,26 +3,26 @@
...
+    body: Body,
```
adpaco-aws, line 25: So `mir` is `mir_internal` and `body` is `mir_stable`, right?

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
```diff
@@ -108,6 +104,11 @@ impl<'tcx> CurrentFnCtx<'tcx> {
     pub fn sig(&self) -> PolyFnSig<'tcx> {
         self.sig
     }
+
+    /// The body of the function.
+    pub fn body(&self) -> &Body {
+        &self.body
+    }
```
adpaco-aws, line 111: If I'm not mistaken, this should be either `mir_stable` or `body_stable`, but for some reason you've preferred to keep the variables `mir` and `body` separate instead of adding the suffix. What's the reason?

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
```diff
@@ -118,6 +119,6 @@ impl CurrentFnCtx<'_> {
     }
 
     pub fn find_label(&self, bb: &BasicBlock) -> String {
-        self.labels[bb.index()].clone()
+        format!("{bb:?}")
```
adpaco-aws, line 122: Wait, this could be an associated function.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -8,36 +8,37 @@
...
-        instance: Instance<'tcx>,
+        gcx: &mut GotocCtx,
+        instance: Instance,
         fargs: Vec<Expr>,
-        assign_to: Place<'tcx>,
-        target: Option<BasicBlock>,
-        span: Option<Span>,
+        assign_to: Place,
+        target: Option<BasicBlockIdx>,
+        span: Span,
```
adpaco-aws, line 34: Again, I'd prefer to keep `span` as `Option<Span>` until we're 100% sure we can remove it from everywhere.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -54,104 +55,98 @@ fn matches_function(tcx: TyCtxt, instance: Instance, attr_name: &str) -> bool {
...
-        span: Option<Span>,
+        _assign_to: Place,
+        target: Option<BasicBlockIdx>,
+        span: Span,
     ) -> Stmt {
         assert_eq!(fargs.len(), 2);
         let cond = fargs.remove(0).cast_to(Type::bool());
         let msg = fargs.remove(0);
-        let msg = tcx.extract_const_message(&msg).unwrap();
+        let msg = gcx.extract_const_message(&msg).unwrap();
```
adpaco-aws, line 75: We'll have to close https://github.com/model-checking/kani/issues/45 if you keep doing that 😆

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -160,34 +155,36 @@ impl<'tcx> GotocHook<'tcx> for Assert {
...
         assert!(fargs.is_empty());
-        let loc = tcx.codegen_span_option(span);
+        let loc = gcx.codegen_span_stable(span);
         let target = target.unwrap();
-        let pt = tcx.place_ty(&assign_to);
-        if pt.is_unit() {
-            Stmt::goto(tcx.current_fn().find_label(&target), loc)
+        let pt = gcx.place_ty_stable(&assign_to);
+        if pt.kind().is_unit() {
+            Stmt::goto(bb_label(target), loc)
```
adpaco-aws, line 177: May need changes depending on what you do with `bb_label`/`find_labels`.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -209,50 +206,53 @@ impl<'tcx> GotocHook<'tcx> for Panic {
...
 
 struct RustAlloc;
 // Removing this hook causes regression failures.
 // https://github.com/model-checking/kani/issues/1170
-impl<'tcx> GotocHook<'tcx> for RustAlloc {
-    fn hook_applies(&self, tcx: TyCtxt<'tcx>, instance: Instance<'tcx>) -> bool {
-        let full_name = with_no_trimmed_paths!(tcx.def_path_str(instance.def_id()));
+impl GotocHook for RustAlloc {
+    fn hook_applies(&self, _tcx: TyCtxt, instance: Instance) -> bool {
+        let full_name = instance.name();
```
adpaco-aws, line 225: Very nice!

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -224,15 +221,16 @@ impl<'tcx> GotocCtx<'tcx> {
...
         // CBMC requires that the argument to the assertion must be a string constant.
         // If there is one in the MIR, use it; otherwise, explain that we can't.
         assert!(!fargs.is_empty(), "Panic requires a string message");
         let msg = self.extract_const_message(&fargs[0]).unwrap_or(String::from(
             "This is a placeholder message; Kani doesn't support message formatted at runtime",
         ));
 
-        self.codegen_fatal_error(PropertyClass::Assertion, &msg, span)
+        let loc = self.codegen_caller_span_stable(span);
+        self.codegen_assert_assume_false(PropertyClass::Assertion, &msg, loc)
```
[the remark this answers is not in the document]
celinval: Yes. I could either create a codegen_fatal_error_stable or add the `codegen_assert_assume_false` directly here.

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
```diff
@@ -3,8 +3,13 @@
 
 use crate::codegen_cprover_gotoc::GotocCtx;
 use rustc_middle::mir::{BasicBlock, BasicBlockData};
+use stable_mir::mir::BasicBlockIdx;
 use tracing::debug;
 
+pub fn bb_label(bb: BasicBlockIdx) -> String {
+    format!("bb{bb}")
+}
+
```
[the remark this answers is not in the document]
celinval: We still have a few usages of `find_label()` which provides the internal version of `BasicBlockIdx`. Following the pattern that I've been trying to follow would be to create another function named `find_label_stable()` which takes the new type. However, I decided to create a function that is much simpler and that has a more meaningful name. The original `find_label` doesn't find any label and it doesn't need access to the function context. So the new usages are much  simpler, and instead of rewriting them to be something like:

```rust
self.function_ctx().find_label_stable(bb)
```
we just use:
```rust
bb_label(bb)
```

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -700,6 +703,19 @@ impl<'tcx> GotocCtx<'tcx> {
         self.codegen_fn_item(instance, span)
     }
 
+    pub fn codegen_fndef_stable(
+        &mut self,
+        def: FnDef,
+        args: &GenericArgs,
+        span: Option<SpanStable>,
```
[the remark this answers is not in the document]
celinval: Yes... so I've looked into this as part of this rewrite. We need to use Option<Span> if a codegen function is ever invoked as part of codegen_type() flow. We don't carry a span while processing types.  Almost everything else can just be straight `Span`.

This lack of consistency is the main reason why we cannot rely on CBMC's code coverage.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -172,12 +173,8 @@ impl<'tcx> GotocCtx<'tcx> {
     /// reachability check.
     /// If reachability checks are disabled, the function returns the message
     /// unmodified and an empty (skip) statement.
-    pub fn codegen_reachability_check(
-        &mut self,
-        msg: String,
-        span: Option<Span>,
-    ) -> (String, Stmt) {
-        let loc = self.codegen_caller_span(&span);
+    pub fn codegen_reachability_check(&mut self, msg: String, span: SpanStable) -> (String, Stmt) {
```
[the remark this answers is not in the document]
celinval: If you have a preference, I can move it back. I don't feel strong about any of them.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -700,6 +703,19 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    ) -> Expr {
+        let instance = InstanceStable::resolve(def, args).unwrap();
```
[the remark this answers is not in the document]
celinval: Handle how exactly? It is a bug not being able to resolve an instance here.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -733,6 +749,12 @@ impl<'tcx> GotocCtx<'tcx> {
             .with_location(self.codegen_span_option(span.cloned()))
     }
 
+    pub fn codegen_func_expr_stable(&mut self, instance: InstanceStable, span: SpanStable) -> Expr {
+        let (func_symbol, func_typ) = self.codegen_func_symbol(rustc_internal::internal(instance));
+        Expr::symbol_expression(func_symbol.name, func_typ)
+            .with_location(self.codegen_span_stable(span))
```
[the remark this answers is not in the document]
celinval: I'm assuming you mean moving the function call out. I can do that.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
```diff
@@ -138,26 +141,33 @@ impl<'tcx> ProjectedPlace<'tcx> {
...
         }
     }
 
+    pub fn try_new_internal<'tcx>(
+        goto_expr: Expr,
+        ty: TyInternal<'tcx>,
+        ctx: &mut GotocCtx<'tcx>,
+    ) -> Result<Self, UnimplementedData> {
+        let ty = ctx.monomorphize(ty);
+        Self::try_new(goto_expr, TypeOrVariant::Type(rustc_internal::stable(ty)), None, None, ctx)
```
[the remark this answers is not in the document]
celinval: I named things with `internal` if the function signature takes internal items. Here, I tried to explain this here:
https://model-checking.github.io/kani/stable-mir.html#naming-conventions-in-kani

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
celinval: Yes. It's just an usize

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
celinval: Great question, it doesn't exist today. This type cannot appear in the MIR level that StableMIR represents, so we just didn't add it.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
celinval: sure

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
celinval: ok

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
adpaco-aws: Thanks for the explanation.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
celinval: Is it OK if I create an util method for now in Kani. StableMIR doesn't have all the is_<type> methods yet, and I don't think its high priority right now, since users can still check that, like we do.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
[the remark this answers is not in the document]
adpaco-aws: Having a user-friendly error, not the usual `failed to unwrap ...`. But I don't know if the "nice error" APIs are accessible from here.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
adpaco-aws: Of course, you can add it next to `is_box`.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
celinval: Yes, they should and will. It's just not high priority at this point. We're trying to focus on adding information that cannot be retrieved with stable APIs at all.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
adpaco-aws: Right, I forgot about callers outside this PR.

kani-compiler/src/codegen_cprover_gotoc/codegen/ty_stable.rs
celinval: No. I don't think we need though, since once we migrate things, these will become dead code, and we will have a lint to complain about that.

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
celinval: We don't need to. That information wasn't used at all

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
celinval: I think body is a better name and I didn't want to bother renaming `mir` to `mir_internal` to reduce the amount of changes here. But I can if you prefer. :)

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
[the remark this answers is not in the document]
adpaco-aws: Yeah, I read that before reviewing the PR. But I think this was one of the first I reviewed and was confused about the stable API usage. Anyway, nothing to do here.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
celinval: Why? We should minimize the amount of places with `Option<Span>`. See my comment above on why we need them in the first place.

All callers of this function have access to a valid Span.

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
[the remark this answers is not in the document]
adpaco-aws: Ok, so you need both for now but won't need `find_label` once you've completed the migration. Sounds good!

kani-compiler/src/codegen_cprover_gotoc/codegen/ty_stable.rs
adpaco-aws: That's true.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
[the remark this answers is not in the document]
adpaco-aws: No problem, let's keep it as is. Removing `Option<...>` from wherever we can now is actually a good idea.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
adpaco-aws: Ok, you convinced me to minimize `Option<Span>` here.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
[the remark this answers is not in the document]
adpaco-aws: It's fine, I don't want another function.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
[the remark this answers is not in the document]
celinval: Ok. I'll replace the unwrap here by an expect(). I think we should still crash, since this would be a compilation bug not an error that user should act on.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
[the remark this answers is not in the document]
celinval: Let me know if I can improve the doc though.

kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
celinval: Yeah, even I had to spend 5min remembering what this meant. I also added a comment to help. LOL

adpaco-aws (approved):
Thanks, @celinval !

PR #2925, opened 2023-12-08
Remove doc build from regular regression

celinval:
This check is redundant in our CI since the bookrunner job already builds the documentation. It is also often breaking the regression job, maybe due to some disk requirement.

If we decide that we want this to be part of the regular developer workflow before it pushes a PR, I suggest that we create a separate script that is meant for that.


jaisnan (approved):
Ty!

PR #2895, opened 2023-11-27
Automate CBMC update

tautschnig:
Automatically create pull requests from the result of updating the CBMC release version checks every Monday morning.

Resolves #2243.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #2942, opened 2023-12-13
Automate cargo update without dependabot

tautschnig:
Automatically create pull requests from the result of running `cargo update` every Monday morning. This should avoid the need for manual PRs to update Cargo.lock, which seemingly dependabot wouldn't take care of. We now only use dependabot to update github actions.

This revives what I had initially proposed in #2895 in light of #2940.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,13 +9,3 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-
-  - package-ecosystem: "cargo"
-    directory: "/"
-    schedule:
-      interval: "weekly"
-    groups:
-      cargo:
-        update-types:
-          - "minor"
-          - "patch"
--- a/.github/workflows/cargo-update.yml
+++ b/.github/workflows/cargo-update.yml
@@ -1,0 +1,47 @@
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+name: Attempt cargo update
+
+on:
+  schedule:
+    - cron: "30 3 * * Mon" # Run this every Monday at 03:30 UTC
+  workflow_dispatch:     # Allow manual dispatching for a custom branch / tag.
+
+permissions:
+  checks: write
+  contents: write
+  pull-requests: write
+
+jobs:
+  create-cargo-update-pr:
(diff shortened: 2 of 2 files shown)


zhassan-aws:
> This should avoid the need for manual PRs to update Cargo.lock, which seemingly dependabot wouldn't take care of.

Can you clarify what you mean? Both PRs related to cargo dependencies opened by dependabot so far updated `Cargo.lock`:

https://github.com/model-checking/kani/pull/2920
https://github.com/model-checking/kani/pull/2933

celinval:
I'm assuming that depandabot doesn't update transitive dependencies.

tautschnig:
> I'm assuming that depandabot doesn't update transitive dependencies.

It would seem so, but this doesn't seem to be documented anywhere (`lockfile-only` does the opposite: it restricts updates to just those affecting `Cargo.lock`). Our release process, however, does mandate this kind of update. So we end up having to do human-made PRs (like #2940) despite using dependabot.

I don't mind much about dependabot, we could keep it. But I'd really like to reduce the number of manual steps where they are straightforward to automate.

zhassan-aws:
According to https://github.com/dependabot/feedback/issues/394#issuecomment-471183054, dependabot runs `cargo update` under the hood, so I'm not sure why the behavior is different. Perhaps a configuration change is needed?

Anyway, if we have a working solution, we don't need to dwell too much on getting dependabot to do what we need. I was just trying to avoid introducing new code for this purpose, to avoid increasing our maintenance burden.

zhassan-aws (approved):

tautschnig:
> According to [dependabot/feedback#394 (comment)](https://github.com/dependabot/feedback/issues/394#issuecomment-471183054), dependabot runs `cargo update` under the hood, so I'm not sure why the behavior is different. Perhaps a configuration change is needed?
> 

I'd love to learn about this - I have tried to find what configuration it would take, but wasn't able to come up with anything that worked.

> Anyway, if we have a working solution, we don't need to dwell too much on getting dependabot to do what we need. I was just trying to avoid introducing new code for this purpose, to avoid increasing our maintenance burden.

Equally I am for keeping our own code to a minimum. I suggest that we revert this PR if we ever come across a way to make dependabot do all the work for us.

PR #2954, opened 2023-12-18
Fix nightly jobs running CBMC latest on the `perf` suite 

### LINKED ISSUE #2952: Nightly CBMC jobs fail due to `goto-cc` version mismatch
The nightly CBMC jobs attempt to build and run the `perf` suite with the latest CBMC versions. Unfortunately, it looks like these jobs have been failing over the weekend as in [this run](https://github.com/model-checking/kani/actions/runs/7245792276/job/19736492757):

```
The input was compiled with an old version of goto-cc; please recompile
error: goto-cc exited with status exit status: 1
```

adpaco-aws:
As explained in #2952 , the nightly `perf` runs started failing over the weekend with this error:

```
The input was compiled with an old version of goto-cc; please recompile
error: goto-cc exited with status exit status: 1
```

Changing the build process for CBMC to use `make` instead of `cmake` is enough for the error to go away. [This action run](https://github.com/adpaco-aws/rmc/actions/runs/7254410145) shows how a successful `perf` job with the change in this PR.

Resolves #2952

### DIFF
--- a/.github/workflows/cbmc-latest.yml
+++ b/.github/workflows/cbmc-latest.yml
@@ -81,8 +81,8 @@ jobs:
       - name: Build CBMC
         working-directory: ./cbmc
         run: |
-          cmake -S . -Bbuild -DWITH_JBMC=OFF
-          cmake --build build -- -j 4
+          make -C src minisat2-download cadical-download
+          make -C src -j4 MINISAT2=../../minisat-2.2.1 CADICAL=../../cadical
           # Prepend the bin directory to $PATH
           echo "${GITHUB_WORKSPACE}/cbmc/build/bin" >> $GITHUB_PATH
 


jaisnan (approved):
Tyvm!

zhassan-aws (approved):

tautschnig (changes requested):
I’m sorry, but we’ll need to dig deeper to see how the build system should relate to that error.

adpaco-aws:
> I’m sorry, but we’ll need to dig deeper to see how the build system should relate to that error.

Fair enough, I've started bisecting on the Dec 15 commits.

adpaco-aws:
The problem comes from PR https://github.com/diffblue/cbmc/pull/5815, where `GOTO_BINARY_VERSION` is redefined. 

Now, I'm not sure where to go from here. The first question that comes to mind is: why is this error not reproducing with `make`?

adpaco-aws:
When comparing the outputs of both build systems, I've realized that `make` doesn't actually produce a `build` folder with these commands.

To confirm this, I [made this change to the workflow](https://github.com/adpaco-aws/rmc/commit/a2a23378b96a15f56519f90d6447c36b2c46c865), and then triggered [the CI run](https://github.com/adpaco-aws/rmc/actions/runs/7301534911). Both the `perf` and `regresssion (ubuntu-22.04)` jobs print the same at the end of the `Build CBMC` step:

```
/usr/bin/cbmc
5.95.1 (cbmc-5.95.1)
```

Therefore, it's likely that these jobs aren't testing with the CBMC we're building, but the one that we install through the `Setup Kani Dependencies` step.

Still, this doesn't explain why the `perf` job is failing. If it was picking up version 5.91.1 as the output of the `which cbmc` and `cbmc --version` suggest, then it shouldn't be failing. But it's possible that building Kani with `--release` is helping it pick-up the newest version? That's what I'm testing in [this other CI run](https://github.com/adpaco-aws/rmc/actions/runs/7302158947).

adpaco-aws:
> Still, this doesn't explain why the perf job is failing. [...] But it's possible that building Kani with --release is helping it pick-up the newest version? That's what I'm testing in [this other CI run](https://github.com/adpaco-aws/rmc/actions/runs/7302158947).

The `perf` job is picking up the right version because the path to the CBMC binaries is being added to `$GITHUB_PATH`. The GA runners will use that variable instead of the usual `$PATH`, so `which` is not the most reliable way to determine the `cbmc` location.

Regardless, I've been able to determine that the CBMC being executed in the `regression` is not the one we're building, but the one we install with the "setup" action earlier. This becomes clear in [this CI run](https://github.com/adpaco-aws/rmc/actions/runs/7303701081), which is running [these changes](https://github.com/model-checking/kani/compare/main...adpaco-aws:rmc:issue-2952-5). In that run, the `regression` fails because it can't find `cbmc`, while `perf` also fails because of the version mismatch problem that we're aware of.

Unfortunately, I haven't been able to make the `setup` action work with two steps using an exclusive condition (i.e., based on `inputs.install_cbmc`) to determine if the `install_deps.sh` should be called with `--no-cbmc`. This would add some complexity here and there, but I don't think we should be installing `cbmc` in workflows where we want to test a version built from source. The next step is to get that working.

tautschnig:
Apologies for having been very quiet on this. Failing with current CBMC HEAD should be expected at this time as the new major release being prepared for CBMC includes several breaking changes. We should either ignore the job’s failure, or disable it, or prepare a Kani branch and use that branch for our testing.

adpaco-aws:
> Failing with current CBMC HEAD should be expected at this time as the new major release being prepared for CBMC includes several breaking changes. We should either ignore the job’s failure, or disable it, or prepare a Kani branch and use that branch for our testing.

Agree, #2952 just seems to be the first breaking change to be noticed.

However, I think we need to at least change the `regression` jobs to use the CBMC built from source. Closing this PR in favor of #2965 which does exacty that.

Issue #2129, opened 2023-01-17
Remove `--function` option


celinval:
**Proposed change:** We should include information about the public functions as part of the KaniMetadata when the reachability mode is `pub_fns`. We should stop merging all the metadata and goto binaries together.

**Motivation:** We have a special handling for `--function` today inside `kani-driver` because we cannot tell which artifacts are relevant to some function. So we just merge everything. If we treat functions as potential harnesses, we can get rid of the hack described above. We should also restrict the metadata to include functions that don't take any parameter.

tedinski:
We could also just search the symtab files directly, it might be simpler that way.

celinval:
Do we mean parse the symtab file or invoke some goto utility to print the functions available?

celinval:
I am leaning towards adding information to the Metadata because is more reliable and it will work well even with mangled names. Since this will only be done for the target crate, the cost shouldn't be too high.

celinval:
We no longer have a use for this flag. We used to keep it for the book runner, but we no longer use that. Users should use harnesses instead.

PR #2959, opened 2023-12-20
Migrate foreign function, compiler-interface and kani-middle modules to use StableMIR

celinval:
We still need internal APIs to handle things like:

- **FnAbi:** Support is pending this PR to be merged: https://github.com/rust-lang/rust/pull/119094
- **Filter non-function items:** Also pending a fix on the rust side: https://github.com/rust-lang/rust/pull/119135
- **Checking for reachable functions:** This is not needed in theory, but std lib build fails since it unncovers an existing issue in Kani with a function that was not previously included. I'll create an issue for that. It's unrelated to this change though. 
- **Function attribute handling**
- **Span error**

## Call-out

- ~~This PR requires nightly-2023-12-20. Creating it as a draft until it gets published~~
- This PR currently includes one commit with this toolchain upgrade: https://github.com/model-checking/kani/pull/2961.


kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -289,11 +282,15 @@ impl CodegenBackend for GotocCodegenBackend {
...
-                            || entry_fn == Some(def_id)
-                    });
+                    let main_instance =
+                        stable_mir::entry_fn().map(|main_fn| Instance::try_from(main_fn).unwrap());
+                    let local_reachable = filter_crate_items(tcx, |_, instance| {
+                        let def_id = rustc_internal::internal(instance.def.def_id());
+                        Some(instance) == main_instance || tcx.is_reachable_non_generic(def_id)
+                    })
+                    .into_iter()
+                    .map(MonoItem::Fn)
```
adpaco-aws, line 292: Oh, looks like it's not only the harnesses, but any reachable function. Now I really want to know why 😅

kani-compiler/src/codegen_cprover_gotoc/utils/names.rs
```diff
@@ -52,34 +50,11 @@ impl<'tcx> GotocCtx<'tcx> {
...
-            instance,
-            instance,
-            self.readable_instance_name(instance),
-            llvm_mangled,
-        );
-
-        let pretty = self.readable_instance_name(instance);
-
-        // Make main function a special case in order to support `--function main`
-        // TODO: Get rid of this: https://github.com/model-checking/kani/issues/2129
```
adpaco-aws, line 74: Can #2129 be closed then?

kani-compiler/src/kani_compiler.rs
```diff
@@ -25,6 +25,7 @@ use crate::kani_middle::reachability::filter_crate_items;
 use crate::kani_middle::stubbing::{self, harness_stub_map};
 use crate::kani_queries::QueryDb;
 use crate::session::init_session;
+use cbmc::{InternString, InternedString};
```
adpaco-aws, line 28: Why is the `InternString` import needed?

kani-compiler/src/kani_compiler.rs
```diff
@@ -446,7 +446,14 @@ mod tests {
     use rustc_hir::definitions::DefPathHash;
     use std::collections::HashMap;
 
-    fn mock_next_id() -> HarnessId {
+    fn mock_next_harness() -> HarnessId {
```
adpaco-aws, line 449: Still, we're mocking IDs so I think the names should be `mock_next_harness_id` and `mock_next_stub_id`.

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
celinval: No, it's just the harness

kani-compiler/src/codegen_cprover_gotoc/utils/names.rs
celinval: not yet... we have the same comment on the stable version. The problem here is with the communication between kani-driver and the compiler.

kani-compiler/src/kani_compiler.rs
celinval: Because I'm using the interned string id as the hash now for the harnesses. I could just use the harness names directly, but we usually store the intern version in our data structures.

kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
```diff
@@ -237,28 +227,31 @@ impl CodegenBackend for GotocCodegenBackend {
...
-                            .harness_model_path(&tcx.def_path_hash(harness.def_id()))
-                            .unwrap();
-                        let (gcx, items) =
-                            self.codegen_items(tcx, &[harness], model_path, &results.machine_model);
-                        results.extend(gcx, items, None);
+                        let model_path =
+                            queries.harness_model_path(&harness.mangled_name()).unwrap();
+                        let (gcx, mono_items) = self.codegen_items(
+                            tcx,
+                            &[MonoItem::Fn(harness)],
```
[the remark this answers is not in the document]
celinval: We need the instance to generate the metadata, not the mono item. The conversion from instance to mono item cannot fail, while the opposite can fail in theory. So I believe this is cleaner.

zhassan-aws (approved):
LGTM

adpaco-aws (approved):

PR #2985, opened 2024-01-26
Don't crash when benchmark contains unknown metric

karkhaz:
Previously, the code assumed that all metrics of the benchmark results would be 'declared' in the "metrics" key of the parser output. This commit ensures that if a parser emits a benchmark containing a metric that hasn't been declared in the "metrics" key, that result is ignored and benchcomp does not crash.

Here's an example of a run where benchcomp crashes: https://github.com/diffblue/cbmc/actions/runs/7664661164/job/20889308681?pr=8171

Resolves N/A


tautschnig (approved):
Experiments confirmed that this works as intended.

PR #2983, opened 2024-01-24
Upgrade Rust toolchain to nightly-2024-01-23

celinval:
Try upgrading Rust toolchain to 2024-01-22, however, regression isn't working yet due to an issue with our assertion override hack.  Related PRs so far:

- https://github.com/rust-lang/rust/pull/119869
- https://github.com/rust-lang/rust/pull/120080
- https://github.com/rust-lang/rust/pull/120128
- https://github.com/rust-lang/rust/pull/119369
- https://github.com/rust-lang/rust/pull/116672

Resolves #ISSUE-NUMBER

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #2989, opened 2024-02-03
Fix std overrides when crate has extern std

### LINKED ISSUE #2187: Issue finding `debug_assert` in `no_std` crates when std enabled

celinval:
I tried applying https://github.com/model-checking/kani/pull/2983 fix, however, this would require user to import
`__kani_workaround_core_assert`. To fix that, I moved the definition to be under `kani` crate.

I replaced the existing fixme test. Initially I didn't check we had one, and I created a second one which is simpler (no cargo needed) but that also includes other cases.

Resolves #2187


zhassan-aws (approved):
Awesome, thanks @celinval!

feliperodri (approved):

PR #2992, opened 2024-02-05
Refactoring contracts macro logic into separate modules

JustusAdam:
Refactors the `contracts` module into several smaller modules grouped by task. The previous modules was very long and interleaved a lot of logic for various parts of the contracts state machine. This refactoring groups functions by specificity, task correlation and reusability.

All code changes here are exclusively moves with no changes to functionality.

The split is as follows:

- `mod.rs`  contains definitions of the data structures used to handle contracts, the entry point for contract handling, as well as a documentation overview.
- `bootstrap.rs` is the logic that initialized the central data structures, such as the handler and the functions state
- `check.rs` is all the logic required to create the function that checks a contract.
- `replace.rs` is all the logic required to create a function that abstracts via contract
- `shared.rs` are Kani/contract-specific helper functions that are used in more than one of `bootstrap`, `check` or `replace`
- `helpers.rs` are Kani/contract-independent helper functions

I was going to make the `old` PR but I shudder at the though of appending more logic to the extremely long `contracts.rs` file so I figured I'd split the logic real quick before adding more.


feliperodri (approved):
LGTM.

celinval (approved):
Awesome! This looks much cleaner. Thank you, @JustusAdam!

I'm assuming there was no change in the logic, just moving stuff around, right?

PR #2993, opened 2024-02-05
History Expressions

### LINKED ISSUE #2803: History Expression Support for Function Contracts

JustusAdam:
Add support for History Expressions. This allows an `ensures` clause to evaluate expressions before the function runs, bing them to anonymous variables and use them after calculating the result to be able to compare to the state before the function evaluation.

## TODO

- [ ] Update external documentation with information about how history expressions are used
- [ ] Update internal documentation with expanded example of how `old` is implemented
- [ ] Test case. I could use some help here both with implementing test cases as well as with suggestions for other things to test. Here is the list so far
  - [x] Basic pass/fail
  - [x] pass/fail on using `clone` and not using `clone` respectively for a reference
  - [ ] Using a function call in `old`
  - [ ] Using a block expression (e.g. sequence of statements) as argument to `old`
  - [ ] Denying nesting of `old`, e.g. `old(... old( ) ... )`

Resolves #2803 


library/kani_macros/src/sysroot/contracts/shared.rs
```diff
@@ -197,3 +197,15 @@ pub fn try_as_result_assign(stmt: &syn::Stmt) -> Option<&syn::LocalInit> {
 pub fn try_as_result_assign_mut(stmt: &mut syn::Stmt) -> Option<&mut syn::LocalInit> {
     try_as_result_assign_pat!(stmt, as_mut)
 }
+
+/// Emits a sequence of `let id = expr;` from a vector of `[(id, expr), ...]`.
+///
+/// This doesn't do anything fancy, it's just to make sure the way that history
+/// expressions are treated in contract checks and replaces is uniform.
+pub fn create_history_expr_bindings(history_expressions: &[(Ident, Expr)]) -> TokenStream2 {
+    let history_variables = history_expressions.iter().map(|(v, _)| v);
```
pi314mm, line 206: Maybe this should be
```
let (history_variables, history_exprs) = history_expressions.iter().unzip();
```
instead of creating two iters

JustusAdam: `unzip` requires creation of two intermediate vectors though because it doesn't split into an iterator but has to be collected.

pi314mm:
This seems like it would cause problems with the result ident since all idents are lifted, or in the event of someone making a locally bound variable within a closure, so something like `ensures((|a| old(a)) (argument))`. Wouldn't it be easier to implement if we require that the argument for old be an ident instead of an expression?

Here is an example of something reasonable that breaks
```
#[kani::ensures((|a| old(*ptr+a))(1) == *ptr)]
fn modify(ptr: &mut u32) {
    *ptr += 1;
}
```

Here is another example of something reasonable that breaks
```
#[kani::ensures(old(*ptr+result) == *ptr)]
fn modify(ptr: &mut u32) -> u32 {
    *ptr += 1;
    1
}
```

Issue #2972, opened 2024-01-11
Prepare for CBMC version 6


tautschnig:
We will need to make changes in Kani to support CBMC version 6 (see https://github.com/diffblue/cbmc/issues/7743), once that's released. See also #2952 for what's already failing.

The following changes will be required:
* Adjust the irep strings in line with https://github.com/diffblue/cbmc/pull/5815
* Bump the goto binary version

PR #2995, opened 2024-02-06
Update to CBMC version 6.1.1

tautschnig:
Updates to match changes to the GOTO binary format.

Resolves: #2952, #2972, #3287, #3391

This includes changes our handling of storage markers to be marking is-alive only rather than treating StorageLive as creating a new object. That is, object instances are now tied to their Mir-provided declarations (which, at present, only appear once per function). To still account for when Rust scopes deem an object to be alive, we use StorageLive and StorageDead to update __CPROVER_dead_object. This (global) variable is used by CBMC's pointer checks to track when a pointer may not be safe to dereference for it could be pointing to an object that no longer is in scope.

Resolves: #3099


kani-driver/src/call_cbmc.rs
```diff
@@ -160,30 +160,39 @@ impl KaniSession {
     pub fn cbmc_check_flags(&self) -> Vec<OsString> {
         let mut args = Vec::new();
 
-        if self.args.checks.memory_safety_on() {
-            args.push("--bounds-check".into());
-            args.push("--pointer-check".into());
+        // We assume that malloc cannot fail
+        args.push("--no-malloc-may-fail".into());
```
remi-delmas-3000, line 164: Should we consider to switch this on at some point ?

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -76,17 +76,38 @@ impl<'tcx> GotocCtx<'tcx> {
                 self.codegen_set_discriminant(dest_ty, dest_expr, *variant_index, location)
             }
             StatementKind::StorageLive(var_id) => {
-                if self.queries.args().ignore_storage_markers {
+                if !self.current_fn().is_address_taken_local(*var_id) {
                     Stmt::skip(location)
                 } else {
-                    Stmt::decl(self.codegen_local(*var_id, location), None, location)
+                    let global_dead_object = cbmc::global_dead_object(&self.symbol_table);
```
remi-delmas-3000, line 82: Is the high-level translation approach documented somewhere ?

Maybe just add a comment and example explaining how Mir declarations are mapped to DECL/DEAD and how storage markers are mapped to dynamic updates of dead_object ? Also does it solve any of the problems related to multiple occurrences of StorageLive/StorageDead or the problems with StorageDead not being dominated by StorageLive ?

tests/expected/shadow/unsupported_num_objects/test.rs
```diff
@@ -15,14 +15,14 @@ fn check_max_objects<const N: usize>() {
     // - the NULL pointer whose object ID is 0, and
     // - the object ID for `i`
     while i < N {
-        let x = i;
-        assert_eq!(kani::mem::pointer_object(&x as *const usize), i + 2);
+        let x: Box<usize> = Box::new(i);
```
remi-delmas-3000, line 18: Why this change?

tautschnig: MIR doesn't actually introduce new objects for `x` in this loop, it allocates just one object at function scope. The reason we previously considered this as producing multiple objects was through our interpretation of StorageLive as yielding a declaration. With that gone I instead had to go for an approach that would also produce fresh objects under MIR semantics.

kani-driver/src/call_cbmc.rs
tautschnig: @celinval @zhassan-aws any ideas on this? I was quite surprised to see that this was necessary as otherwise a lot of tests newly fail. How does safe Rust avoid UB with code like `let x: Box<usize> = Box::new(i); let y = *x;`?

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
tautschnig: I have added a comment explaining the approach and also stating that this does address the risk of multiple/mismatching occurrences of StorageLive/StorageDead.

tautschnig:
Needs https://github.com/diffblue/cbmc/pull/8363 to be merged (and a new release containing that PR to be produced) to address remaining performance issues.

tautschnig:
> Needs [diffblue/cbmc#8363](https://github.com/diffblue/cbmc/pull/8363) to be merged (and a new release containing that PR to be produced) to address remaining performance issues.

Done, now waiting for [CBMC 6.1.0 to be available via brew](https://github.com/Homebrew/homebrew-core/pull/178134).

scripts/kani-perf.sh
```diff
@@ -27,7 +27,8 @@ done
 suite="perf"
 mode="cargo-kani-test"
 echo "Check compiletest suite=$suite mode=$mode"
-cargo run -p compiletest -- --suite $suite --mode $mode --no-fail-fast
+cargo run -p compiletest -- --suite $suite --mode $mode --no-fail-fast \
+  --kani-flag="--enable-unstable" --kani-flag="--cbmc-args" --kani-flag="--verbosity" --kani-flag="9"
```
zhassan-aws, line 31: Is `--verbosity 9` now necessary to report performance statistics?

tests/expected/dead-invalid-access-via-raw/main.expected
```diff
@@ -1,7 +1,5 @@
 SUCCESS\
 address must be a multiple of its type's alignment
-FAILURE\
```
zhassan-aws, line 3: Does a dereference failure now block further checks in the execution path?

tools/benchcomp/configs/perf-regression.yaml
```diff
@@ -10,13 +10,13 @@ variants:
   kani_new:
     config:
       directory: new
-      command_line: scripts/kani-perf.sh
+      command_line: "scripts/setup/ubuntu/install_deps.sh && scripts/kani-perf.sh"
```
zhassan-aws, line 13: Should we just run `install_cbmc.sh` instead? The `bench` workflow runs `install_deps.sh` as part of the `setup` action.

zhassan-aws (commented):
Thanks for making this happen!

How does the change in CBMC's default verbosity affect the Kani output? It the change worth pointing out in the CHANGELOG.md?

tools/benchcomp/configs/perf-regression.yaml
tautschnig: Possibly - I guess it is the only dependency that could affect performance?

tests/expected/dead-invalid-access-via-raw/main.expected
tautschnig: Correct - all other checks will be marked "UNKNOWN".

scripts/kani-perf.sh
tautschnig: Yes, CBMC no longer prints timing stats by default.

tautschnig:
> How does the change in CBMC's default verbosity affect the Kani output? Is the change worth pointing out in the CHANGELOG.md?

I think we used to see loop-unwinding information while running Kani, that is now no longer the case. Other than this, I don't think Kani is impacted.

kani-compiler/src/codegen_cprover_gotoc/context/current_fn.rs
```diff
@@ -26,6 +26,8 @@ pub struct CurrentFnCtx<'tcx> {
     locals: Vec<LocalDecl>,
     /// A list of pretty names for locals that corrspond to user variables.
     local_names: HashMap<Local, InternedString>,
+    /// Collection of variables that are used in a reference or address-of expression.
+    address_taken_locals: HashSet<usize>,
```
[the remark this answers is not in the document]
tautschnig: Thanks, done.

tools/benchcomp/configs/perf-regression.yaml
tautschnig: We also install kissat as part of this, and a PR changing kissat versions could also change performance characteristics. So I'd rather not make this proposed change.

zhassan-aws:
> I think we used to see loop-unwinding information while running Kani, that is now no longer the case.

While I'm happy to see that, users may have depended on this output to see when CBMC is unwinding forever. So, I think we should at least point this change out in the CHANGELOG.md.

I would also suggest adding a Kani option that would turn on `--verbosity 9` in case users want to restore the old behavior. What do you think?

tests/expected/shadow/unsupported_num_objects/test.rs
```diff
@@ -32,10 +32,10 @@ fn check_max_objects<const N: usize>() {
 
 #[kani::proof]
 fn check_max_objects_pass() {
-    check_max_objects::<1021>();
+    check_max_objects::<510>();
```
feliperodri, line 35: Quite a drop here.

tests/expected/shadow/slices/slice_of_array/test.rs
```diff
@@ -13,7 +13,7 @@ const N: usize = 16;
 static mut SM: kani::shadow::ShadowMem<bool> = kani::shadow::ShadowMem::new(false);
 
 #[kani::proof]
-#[kani::unwind(31)]
+#[kani::unwind(17)]
```
feliperodri, line 16: Quite a drop too.

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -75,18 +75,68 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            //    object). We update __CPROVER_dead_object as follows:
+            //    * StorageLive is set to NULL when __CPROVER_dead_object pointed to the object
+            //    (re-)entering scope, or else is left unchanged.
+            //    * StorageDead non-deterministically updates (or leaves unchanged)
+            //    __CPROVER_dead_object to point to the object going out of scope. (This is the
+            //    same update approach as used within CBMC.)
+            //
+            // This approach will also work when there are multiple occurrences of StorageLive (or
+            // StorageDead) on a path, or across control-flow branches, and even when StorageDead
+            // occurs without a preceding StorageLive.
```
feliperodri, line 106: Thank you for documenting this thoroughly.

feliperodri (approved):
🚢 it! (after @zhassan-aws comments)

tests/expected/shadow/unsupported_num_objects/test.rs
tautschnig: Ack, but not one related to performance: it's just a different way of interpreting what is/is not an object.

tests/expected/shadow/slices/slice_of_array/test.rs
tautschnig: I'll see whether we can avoid this when moving to CBMC 6.1.1.

tautschnig: Reverted with CBMC 6.1.1

scripts/kani-perf.sh
tautschnig: I'll now make `--verbosity 9` the default with Kani.

PR #3015, opened 2024-02-09
Upgrade Rust toolchain to nightly-2024-02-09

feliperodri:
Attempt to update Rust toolchain from nightly-2024-01-25 to nightly-2024-02-09.

Resolves https://github.com/model-checking/kani/issues/3014.


kani-driver/src/args/common.rs
```diff
@@ -50,8 +50,6 @@ pub trait Verbosity {
     /// Whether we should be verbose.
     /// Note that `debug() == true` must imply `verbose() == true`.
     fn verbose(&self) -> bool;
-    /// Whether we should emit debug messages.
-    fn debug(&self) -> bool;
```
zhassan-aws, line 54: Was it necessary to remove this?

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -108,6 +109,7 @@ impl TypeExt for Type {
     }
 }
 
+#[allow(unused)]
```
zhassan-aws, line 112: Is this necessary? I see that the trait is used below.

tests/ui/derive-arbitrary/non_arbitrary_param/expected
```diff
@@ -1,5 +1 @@
 error[E0277]: the trait bound `Void: kani::Arbitrary` is not satisfied
```
zhassan-aws, line 1: I think part of the test is making sure the diagnostic message correctly points to the problematic line.

Can you try updating the `expected` file to:
```
error[E0277]: the trait bound `Void: kani::Arbitrary` is not satisfied
    |\
14  |     let _wrapper: Wrapper<Void> = kani::any();\
    |                                   ^^^^^^^^^ the trait `kani::Arbitrary` is not implemented for `Void`\
    |
```
?

kani-driver/src/args/common.rs
feliperodri: It wasn't necessary to remove it, it was just unused. I can add it back with `#[allow(unused)]` to make clippy happy.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
feliperodri: I get the following warnings without it:
```
warning: method `is_unit_pointer` is never used
  --> kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs:55:8
   |
50 | pub trait TypeExt {
   |           ------- method in this trait
...
55 |     fn is_unit_pointer(&self) -> bool;
   |        ^^^^^^^^^^^^^^^
   |

warning: trait `ExprExt` is never used
   --> kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs:112:7
    |
112 | trait ExprExt {
    |       ^^^^^^^

warning: `kani-compiler` (bin "kani-compiler") generated 2 warnings
```

tests/ui/derive-arbitrary/non_arbitrary_param/expected
feliperodri: I added the message back.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
zhassan-aws: I see. In that case, I would suggest deleting the unused code (e.g. `ExprExt` trait and the `impl` block).

zhassan-aws (approved):

Cargo.lock
```diff
@@ -14,13 +14,14 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.7.8"
+version = "0.8.7"
```
adpaco-aws, line 17: Why are these Cargo updates needed?

feliperodri: Some of our current dependencies would break the build with this toolchain update. The fix is updating to a newer version.

Issue #3035, opened 2024-02-20
Regression in contract verification from 0.45.0 to 0.46.0


roypat:
While upgrading the kani version used in Firecracker's CI to 0.46.0 I started noticing one of our [harnesses](https://github.com/firecracker-microvm/firecracker/blob/8bb883111ff608dc2ca1a08d5e4eb2ddbd5c1c8b/src/vmm/src/rate_limiter/mod.rs#L659-L672) started to fail with a failure that should have been ruled out by the contract's `requires` clause ("Failed Checks: x > 0 && y > 0", see https://buildkite.com/firecracker/firecracker-pr/builds/9140#018dc0ef-ba26-42b3-8b0f-124c00dad946 for the verification failure). 

I was able to minimize the example to the following:

```rust
#[kani::requires(x != 0)]
fn function(x: u32) {
    assert_ne!(x, 0);
}

#[kani::proof_for_contract(function)]
fn function_harness() {
    function(kani::any());
}
```

using the following command line invocation:

```
cargo kani -p vmm --harness function_harness --enable-unstable -Zfunction-contracts
```

with Kani version:

I expected to see this happen: The verification passes with kani 0.45.0 (both the firecracker example, and the minimal reproducer), so I was expecting it to also pass in 0.46.0 (although the interesting thing is that 0.45.0 only has 20 checks, while 0.46.0 has over 1000, so maybe there was something wrong in 0.45.0 that got fixed in 0.46.0?)

Instead, this happened: Verification failed (output: https://pastebin.com/EDUN0TVq, sorry it was too long for a github comment :( )

PR #3045, opened 2024-02-28
Retrieve info for recursion tracker reliably

feliperodri:
Fixes https://github.com/model-checking/kani/issues/3035.


kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -37,6 +37,23 @@ impl<'tcx> GotocCtx<'tcx> {
         let function_under_contract_attrs = KaniAttributes::for_item(tcx, function_under_contract);
         let wrapped_fn = function_under_contract_attrs.inner_check().unwrap().unwrap();
 
+        let recursion_tracker = items
+            .iter()
+            .filter_map(|i| match i {
```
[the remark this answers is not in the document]
celinval: Also, can you use find instead?

celinval (commented):
Do you know why we use name instead of mangled_name

feliperodri:
> Do you know why we use name instead of mangled_name

@celinval do we have available mangled name for `StaticDef`? The name and path are sufficient to get the correct variable.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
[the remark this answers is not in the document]
feliperodri: Should I replace the filter map used below to get the `instance_under_contract` with a find as well?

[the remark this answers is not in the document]
feliperodri: I implemented the update using `find` as you suggested.

> What if there's another item with reentry in the name?

There is the chance it would pick up the wrong item. I could complement the comparison using RegEx or by adding something like
```rust
if (*recursion_tracker).name().contains("REENTRY") &&
   (*recursion_tracker).name().contains("recursion_wrapper") =>
```
However, as we discussed offline, a more reliable approach would be to add an attribute to the item that could mark it as "instrumentation code for contracts". There are other clashing names issues due to this instrumentation (e.g., https://github.com/model-checking/kani/issues/2597), so I suggest to merge this fix and address the name clashing issues on contracts in a separate PR.

celinval:
> > Do you know why we use name instead of mangled_name
> 
> @celinval do we have available mangled name for `StaticDef`? The name and path are sufficient to get the correct variable.
> 
> ```
> error[E0599]: no method named `mangled_name` found for struct `StaticDef` in the current scope
>   --> kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs:88:40
>    |
> 88 |             instance_recursion_tracker.mangled_name(),
>    |                                        ^^^^^^^^^^^^ method not found in `StaticDef`
> ```

You should invoke mangled_name in the instance, not in the `StaticDef`.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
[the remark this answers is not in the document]
celinval: These are different problems. The problem here is not name clashing per se. The problem here is a reliable mechanism for the compiler to retrieve the name of this specific static.

This would only clash with a name if the name of a parameter is `REENTRY`, but this is very unlikely since it breaks the Rust naming convention.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -65,11 +79,11 @@ impl<'tcx> GotocCtx<'tcx> {
         let location_of_recursion_wrapper = self.codegen_span(&span_of_recursion_wrapper);
 
         let full_name = format!(
-            "{}:{}::REENTRY",
+            "{}:{}",
             location_of_recursion_wrapper
                 .filename()
                 .expect("recursion location wrapper should have a file name"),
-            tcx.item_name(recursion_wrapper_id),
+            recursion_tracker.name(),
```
JustusAdam, line 86: Given that this is the second time we've had to fix the exact same issue I would add more documentation.
I would refactor this into a variable and add a comment that explains that this needs to use the same name that is emitted into the symbol table. 
At the same time I would add a comment where the name is selected for the symbol table and link it to this one (or maybe the original issue) explaining that these two names must be the same or contracts will silently fail.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -37,6 +37,20 @@ impl<'tcx> GotocCtx<'tcx> {
         let function_under_contract_attrs = KaniAttributes::for_item(tcx, function_under_contract);
         let wrapped_fn = function_under_contract_attrs.inner_check().unwrap().unwrap();
 
+        let recursion_tracker = items.iter().find_map(|i| match i {
+            MonoItem::Static(recursion_tracker)
+                if (*recursion_tracker).name().contains("REENTRY") =>
```
JustusAdam, line 42: Is there a way to make this name match more specific? For instance like only searching the siblings of the harness function? There is code in `attributes.rs` that does similar stuff, or you may consider using `TyCtxt::parent_module_from_def_id` to get the module containing the harness and then `TyCtxt::module_children` to get the siblings that you can search.

Also I would suggest using `filter_map` and fail if the resulting iterator has more than one matching element.

JustusAdam (changes requested):
Thanks for fixing this. 

I would mention a "Closes" for the original issue that this addresses.

celinval:
Hey @feliperodri, can you please make sure this code works if the contract is added to a generic function? I don't know if we already have a test for that.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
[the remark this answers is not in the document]
feliperodri: I've updated the way we search for `REENTRY` and add an assertion, so we can fail in the case we find multiple instance. In the future, we can add a `KaniAttribute` to track this variable.

feliperodri:
@celinval 
> You should invoke mangled_name in the instance, not in the `StaticDef`.

I added a comment to clarify why we can't use now `mangled_name`. I also reported the issue on CBMC (see https://github.com/diffblue/cbmc/issues/8225).

> can you please make sure this code works if the contract is added to a generic function? I don't know if we already have a test for that.

I added a test with generics and recursion. I didn't find any tests like that in our regression.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
feliperodri: https://github.com/model-checking/kani/pull/3045#discussion_r1508420857

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
feliperodri: I added more documentation to highlight some of these issues.

feliperodri:
> I would mention a "Closes" for the original issue that this addresses.

@JustusAdam I added to the PR description.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -35,8 +35,45 @@ impl<'tcx> GotocCtx<'tcx> {
...
         let tcx = self.tcx;
         let function_under_contract_attrs = KaniAttributes::for_item(tcx, function_under_contract);
-        let wrapped_fn = function_under_contract_attrs.inner_check().unwrap().unwrap();
 
+        let recursion_wrapper_id =
+            function_under_contract_attrs.checked_with_id().unwrap().unwrap();
+        let mut recursion_tracker = items.iter().filter_map(|i| match i {
+            MonoItem::Static(recursion_tracker)
+                if (*recursion_tracker).name().contains(
+                    format!("{}::REENTRY", tcx.item_name(recursion_wrapper_id)).as_str(),
```
JustusAdam, line 44: This is much better, though I wonder if instead of checking for `contains` we could check directly for equality? And also, as a small optimization, I would move the `format` outside the loop. E.g.

```rs
let expecterd_name = format!("{}::REENTRY", tcx.item_name(recursion_wrapper_id);
let mut recursion_tracker = items.iter().filter_map(|i| match i {
    MonoItem::Static(recursion_tracker)
        if (*recursion_tracker).name() == expected_name.as_str() =>
    {
        Some(*recursion_tracker)
    }
    _ => None,
});

File: kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs [hunk 5]
```diff
@@ -35,8 +35,45 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                ) =>
+            {
+                Some(*recursion_tracker)
+            }
+            _ => None,
+        });
+        let recursion_tracker_def = recursion_tracker.next().unwrap();
```
JustusAdam, line 51: I would add a short message here in case this `unwrap` fails.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -35,8 +35,45 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        );
+
+        let span_of_recursion_wrapper = tcx.def_span(recursion_wrapper_id);
+        let location_of_recursion_wrapper = self.codegen_span(&span_of_recursion_wrapper);
+        // The name and location for the recursion tracker should match the exact information added
+        // to the symbol table, otherwise our contract instrumentation will silently failed.
+        // This happens because Kani relies on `--nondet-static-exclude` from CBMC to properly
+        // handle this tracker. CBMC silently fails if there is no match in the symbol table
+        // that correspond to the argument of this flag.
+        // More details at https://github.com/model-checking/kani/pull/3045.
```
JustusAdam, line 64: Nice. I like that a lot

JustusAdam: I would however recommend adding a sister comment to the code that emits the names into the symbol table to make sure that when someone edits that they know that they might have to review this piece of code also.

JustusAdam (commented):
Good stuff.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
celinval: Can you also add a link to the CBMC issue instead? So we know when this can be adjusted.

tests/expected/function-contract/generic_infinity_recursion.rs
```diff
@@ -0,0 +1,11 @@
+#[kani::requires(x != 0)]
```
celinval, line 1: Please add the copyright header and a quick module documentation explaining what you are testing

celinval (commented):
Looks good

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
feliperodri: We can't check for directly equality because `tcx.item_name(recursion_wrapper_id)` doesn't give us the module of the function, so there is still an information missing and that's why I used contains. But I can definitely move the `format` outside of the loop.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
feliperodri: Implemented as suggested.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
feliperodri: @celinval the comment right bellow this one contains the CBMC issue. Should I remove this link and just keep the CBMC issue or it's ok to leave them both?

tests/expected/function-contract/generic_infinity_recursion.rs
feliperodri: Implemented as suggested.

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
celinval: Oh. I missed that. That's OK. Thanks

celinval (approved):

JustusAdam (approved):

PR #3047, opened 2024-02-28
Fix `codegen_atomic_binop` for `atomic_ptr`

### LINKED ISSUE #3042: Kani crashes when handling code related to AtomicPtr
Caught by the M1 [regression](https://github.com/model-checking/kani/actions/runs/7995213474/job/21871709415), Kani seems to crash when handling harness that use atomic pointers.

Here is a minimal example that causes the crash, in both linux and mac versions of Kani

```
#![feature(strict_provenance_atomic_ptr, strict_provenance)]
use std::sync::atomic::{AtomicPtr, Ordering};

#[kani::proof]
fn check_fetch_byte() {
    let atom = AtomicPtr::<i64>::new(core::ptr::invalid_mut(1));
    assert_eq!(atom.fetch_byte_sub(1, Ordering::Relaxed).addr(), 1);
    assert_eq!(atom.load(Ordering::Relaxed).addr(), 0);
}
``` 

This is the crash

```
Kani Rust Verifier 0.42.0 (standalone)
thread 'rustc' panicked at cprover_bindings/src/goto_program/stmt.rs:170:9:
assertion `left == right` failed: Error: assign statement with unequal types lhs Pointer { typ: Signedbv { width: 64 } } rhs CInteger(SSizeT)
  left: Pointer { typ: Signedbv { width: 64 } }
 right: CInteger(SSizeT)

[Kani] current codegen item: codegen_function: std::sync::atomic::atomic_sub::<*mut i64>
_RINvNtNtCslE7LfYgGyNS_4core4sync6atomic10atomic_subOxECsgohT1pLHQU2_4main
[Kani] current codegen location: Loc { file: "./.rustup/toolchains/nightly-2023-11-28-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/sync/atomic.rs", function: None, start_line: 3334, start_col: Some(1), end_line: 3334, end_col: Some(73) }
error: ./.kani/kani-0.42.0/bin/kani-compiler exited with status exit status: 101
```

```
kani main.rs
```

with Kani version: 0.46.0, 0.45.0, 0.42.0 (and possible older versions).

qinheping:
Fetch functions of `atomic_ptr` calls atomic intrinsic functions with pointer-type arguments (`invalid_mut`), which will cause typecheck failures. The change in this commit add support of pointer-type arguments into `codegen_atomic_binop` to fix the issue. The new `codegen_atomic_binop` will cast pointer arguments to `size_t`, apply op on them, and then cast the op result back to the pointer type.

The new test include all fetch functions of `atomic_ptr` except for `fetch_ptr_add` and `fetch_ptr_sub`, which do not call intrinsic functions.

Resolves #3042.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
@@ -240,9 +240,9 @@ impl<'tcx> GotocCtx<'tcx> {
         // var = tmp;
         // -------------------------
         //
-        // In fetch functions of atomic_ptr, the type of var2 can be pointer (invalid_mut).
+        // In fetch functions of atomic_ptr such as https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html#method.fetch_byte_add,
+        // the type of var2 can be pointer (invalid_mut).
         // In such case, atomic binops are transformed as follows to avoid typecheck failure.
-        //
         // -------------------------
         // var = atomic_op(var1, var2)
         // -------------------------
@@ -262,20 +262,18 @@ impl<'tcx> GotocCtx<'tcx> {
                 let (tmp, decl_stmt) =
                     self.decl_temp_variable(var1.typ().clone(), Some(var1.to_owned()), loc);
                 let var2 = fargs.remove(0);
+                let op_expr: Expr;
                 if var2.typ().is_pointer() {
-                    let op_expr = (var1.clone().cast_to(Type::c_size_t()))
+                    op_expr = (var1.clone().cast_to(Type::c_size_t()))
                         .$op(var2.cast_to(Type::c_size_t()))
                         .with_location(loc)
                         .cast_to(var1.typ().clone());
-                    let assign_stmt = (var1.clone()).assign(op_expr, loc);
-                    let res_stmt = self.codegen_expr_to_place_stable(place, tmp.clone());
-                    Stmt::atomic_block(vec![decl_stmt, assign_stmt, res_stmt], loc)
                 } else {
-                    let op_expr = (var1.clone()).$op(var2).with_location(loc);
-                    let assign_stmt = (var1.clone()).assign(op_expr, loc);
-                    let res_stmt = self.codegen_expr_to_place_stable(place, tmp.clone());
-                    Stmt::atomic_block(vec![decl_stmt, assign_stmt, res_stmt], loc)
+                    op_expr = (var1.clone()).$op(var2).with_location(loc);
                 }
+                let assign_stmt = (var1.clone()).assign(op_expr, loc);
+                let res_stmt = self.codegen_expr_to_place_stable(place, tmp.clone());
+                Stmt::atomic_block(vec![decl_stmt, assign_stmt, res_stmt], loc)
             }};
         }
 
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
@@ -262,15 +262,14 @@ impl<'tcx> GotocCtx<'tcx> {
                 let (tmp, decl_stmt) =
                     self.decl_temp_variable(var1.typ().clone(), Some(var1.to_owned()), loc);
                 let var2 = fargs.remove(0);
-                let op_expr: Expr;
-                if var2.typ().is_pointer() {
-                    op_expr = (var1.clone().cast_to(Type::c_size_t()))
+                let op_expr = if var2.typ().is_pointer() {
+                    (var1.clone().cast_to(Type::c_size_t()))
                         .$op(var2.cast_to(Type::c_size_t()))
                         .with_location(loc)
-                        .cast_to(var1.typ().clone());
+                        .cast_to(var1.typ().clone())
                 } else {
-                    op_expr = (var1.clone()).$op(var2).with_location(loc);
-                }
+                    (var1.clone()).$op(var2).with_location(loc)
+                };
                 let assign_stmt = (var1.clone()).assign(op_expr, loc);
                 let res_stmt = self.codegen_expr_to_place_stable(place, tmp.clone());
                 Stmt::atomic_block(vec![decl_stmt, assign_stmt, res_stmt], loc)
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
@@ -239,6 +239,19 @@ impl<'tcx> GotocCtx<'tcx> {
         // *var1 = op(*var1, var2);
         // var = tmp;
         // -------------------------
+        //
+        // In fetch functions of atomic_ptr, the type of var2 can be pointer (invalid_mut).
+        // In such case, atomic binops are transformed as follows to avoid typecheck failure.
+        //
+        // -------------------------
+        // var = atomic_op(var1, var2)
+        // -------------------------
+        // unsigned char tmp;
+        // tmp = *var1;
+        // *var1 = (typeof var1)op((size_t)*var1, (size_t)var2);
+        // var = tmp;
+        // -------------------------
+        //
         // Note: Atomic arithmetic operations wrap around on overflow.
         macro_rules! codegen_atomic_binop {
             ($op: ident) => {{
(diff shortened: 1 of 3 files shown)


qinheping:
Partially resolve #25. The benchmarks included in this PR test atomic intrinsic with type of `*mut i64`.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -249,10 +262,20 @@ impl<'tcx> GotocCtx<'tcx> {
                 let (tmp, decl_stmt) =
                     self.decl_temp_variable(var1.typ().clone(), Some(var1.to_owned()), loc);
                 let var2 = fargs.remove(0);
-                let op_expr = (var1.clone()).$op(var2).with_location(loc);
-                let assign_stmt = (var1.clone()).assign(op_expr, loc);
-                let res_stmt = self.codegen_expr_to_place_stable(place, tmp.clone());
-                Stmt::atomic_block(vec![decl_stmt, assign_stmt, res_stmt], loc)
+                if var2.typ().is_pointer() {
+                    let op_expr = (var1.clone().cast_to(Type::c_size_t()))
```
adpaco-aws, line 266: Only `op_expr` depends on `var2` being a pointer or not. The rest of code can be out of the if-then-else blocks.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -239,6 +239,19 @@ impl<'tcx> GotocCtx<'tcx> {
         // *var1 = op(*var1, var2);
         // var = tmp;
         // -------------------------
+        //
+        // In fetch functions of atomic_ptr, the type of var2 can be pointer (invalid_mut).
+        // In such case, atomic binops are transformed as follows to avoid typecheck failure.
```
adpaco-aws, line 244: Can you add a reference to at least one of such functions?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -239,6 +239,19 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        //
+        // -------------------------
+        // var = atomic_op(var1, var2)
+        // -------------------------
+        // unsigned char tmp;
+        // tmp = *var1;
+        // *var1 = (typeof var1)op((size_t)*var1, (size_t)var2);
```
adpaco-aws, line 251: This is the only line that changes w.r.t. the previous example. Why don't we edit the original example so that the possibility of `var2` being a pointer is considered? Something like:
```
        // unsigned char tmp;
        // tmp = *var1;
        // if var2.is_pointer() {
        //   *var1 = (typeof var1)op((size_t)*var1, (size_t)var2);
        // } else {
        //   *var1 = op(*var1, var2);
        // }
        // var = tmp;
```

tests/kani/AtomicPtr/atomic_ptr.rs
```diff
@@ -0,0 +1,65 @@
+#![feature(strict_provenance_atomic_ptr, strict_provenance)]
```
adpaco-aws, line 1: Needs the copyright header.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -249,10 +262,20 @@ impl<'tcx> GotocCtx<'tcx> {
...
-                let op_expr = (var1.clone()).$op(var2).with_location(loc);
-                let assign_stmt = (var1.clone()).assign(op_expr, loc);
-                let res_stmt = self.codegen_expr_to_place_stable(place, tmp.clone());
-                Stmt::atomic_block(vec![decl_stmt, assign_stmt, res_stmt], loc)
+                if var2.typ().is_pointer() {
+                    let op_expr = (var1.clone().cast_to(Type::c_size_t()))
+                        .$op(var2.cast_to(Type::c_size_t()))
+                        .with_location(loc)
+                        .cast_to(var1.typ().clone());
+                    let assign_stmt = (var1.clone()).assign(op_expr, loc);
```
zhassan-aws, line 270: This line and the next 3 are repeated in the else branch. Can you refactor to avoid code duplication?

tests/kani/AtomicPtr/atomic_ptr.rs
```diff
@@ -0,0 +1,65 @@
...
+use std::sync::atomic::{AtomicPtr, Ordering};
+
```
adpaco-aws, line 3: Needs a description (i.e., what are you testing?).

zhassan-aws, line 3: Please add a comment that describes what those tests are checking.

tests/kani/AtomicPtr/atomic_ptr.rs
adpaco-aws, line 1: I understand why you may want all these tests to be in one file, but I think they should live in `tests/kani/Intrinsics/Atomic/Stable` since they're testing intrinsics through a stable interface.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
zhassan-aws, line 266: Are the casts here necessary, or is the last one (casting the whole expression to `var1.typ()`) sufficient?

zhassan-aws (changes requested):

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
qinheping: Yes, they are necessary. For example, in https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html#method.fetch_byte_add, we do pointer arithmetic `*var1-var2` in the intrinsic where `var2` is a pointer `invalid_mut`. It will lead to typecheck failure since we don't allow `pointer - pointer`. Actually, the hack in `fetch_byte_add` is to use `invalid_mut` as integer to do the pointer arithmetic. So we also cast the variables to integer type in the codegen macro to avoid type check failures.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
qinheping: The two examples are the two cases of how we transform atomic binop into goto. There is no if-then-else in the transformed snippet. Combining the two examples and introducing an if-then-else statement in it will be misleading.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -262,20 +262,18 @@ impl<'tcx> GotocCtx<'tcx> {
                 let (tmp, decl_stmt) =
                     self.decl_temp_variable(var1.typ().clone(), Some(var1.to_owned()), loc);
                 let var2 = fargs.remove(0);
+                let op_expr: Expr;
                 if var2.typ().is_pointer() {
```
zhassan-aws, line 266: The more "Rusty" way to do this is:
```rust
                let op_expr = if var2.typ().is_pointer() {
                    (var1.clone().cast_to(Type::c_size_t()))
                        .$op(var2.cast_to(Type::c_size_t()))
                        .with_location(loc)
                        .cast_to(var1.typ().clone())
                } else {
                    (var1.clone()).$op(var2).with_location(loc)
                };
```

tests/kani/Intrinsics/Atomic/Stable/AtomicPtr/main.rs
```diff
@@ -1,3 +1,8 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+//! Test atomic intrinsics through the stable interface of atomic_ptr.
+
```
zhassan-aws, line 5: Can you elaborate on the "special case" these tests are checking? Example:
```
Specifically, it checks that Kani correctly handles atomic_ptr's fetch methods, in which the second argument is a pointer type.
```
or something along those lines. You can also mention that this wasn't handled correctly before and point to https://github.com/model-checking/kani/issues/3042.

tests/kani/Intrinsics/Atomic/Stable/AtomicPtr/main.rs
```diff
@@ -0,0 +1,72 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// Test atomic intrinsics through the stable interface of atomic_ptr.
+// Specifically, it checks that Kani correctly handles atomic_ptr's fetch methods, in which the second argument is a pointer type.
+// These methods were not correctly handles as explained in https://github.com/model-checking/kani/issues/3042.
```
zhassan-aws, line 6:
```suggestion
// These methods were not correctly handled as explained in https://github.com/model-checking/kani/issues/3042.
```

zhassan-aws (approved):
Thanks @qinheping!

adpaco-aws (approved):

PR #3053, opened 2024-03-01
Update s2n-quic submodule weekly through dependabot

zhassan-aws:
Use dependabot to update the `s2n-quic` submodule weekly (every Monday). This is to avoid having to do this effort manually, and to guarantee that none of the `s2n-quic` proofs break due to changes in Kani.

I've tested this in my own fork, and it seemed to work correctly:

https://github.com/zhassan-aws/kani/network/updates/19481419/jobs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,10 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "gitsubmodule"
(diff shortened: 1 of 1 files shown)


adpaco-aws (approved):

adpaco-aws:
Note: We don't have access to the "network" link you posted, so I'll omit it from the commit message. It's not a big deal, let's merge it now and see what happens on Monday!

PR #3056, opened 2024-03-04
Add `--use-local-toolchain` to Kani setup

jaisnan:
> Please ensure your PR description includes the following:
> 4. If you had to perform any manual test, please describe them.
>
> **Make sure you remove this list from the final PR description.**

Adds `--use-local-toolchain` to Kani's setup flow, which accepts a local toolchain and then uses that to finish the Kani setup. 

Some notes:

1. Why? This is mainly for installing GPG verified toolchains. 
2. This doesn't switch the cargo being invoked by kani-driver (yet). 

Resolves #ISSUE-NUMBER

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,7 @@ mod cmd;
 mod os_hacks;
 mod setup;
 
+use clap::{App, Arg, SubCommand};
 use std::ffi::OsString;
 use std::os::unix::prelude::CommandExt;
 use std::path::{Path, PathBuf};
@@ -28,16 +29,16 @@ use anyhow::{bail, Context, Result};
 /// `bin` should be either `kani` or `cargo-kani`
 pub fn proxy(bin: &str) -> Result<()> {
     match parse_args(env::args_os().collect()) {
-        ArgsResult::ExplicitSetup { use_local_bundle } => setup::setup(use_local_bundle),
+        ArgsResult::ExplicitSetup { use_local_bundle, use_local_toolchain } => setup::setup(use_local_bundle, use_local_toolchain),
         ArgsResult::Default => {
             fail_if_in_dev_environment()?;
             if !setup::appears_setup() {
-                setup::setup(None)?;
+                setup::setup(None, None)?;
             } else {
                 // This handles cases where the setup was left incomplete due to an interrupt
                 // For example - https://github.com/model-checking/kani/issues/1545
                 if let Some(path_to_bundle) = setup::appears_incomplete() {
-                    setup::setup(Some(path_to_bundle.clone().into_os_string()))?;
+                    setup::setup(Some(path_to_bundle.clone().into_os_string()), None)?;
                     // Suppress warning with unused assignment
                     // and remove the bundle if it still exists
                     let _ = fs::remove_file(path_to_bundle);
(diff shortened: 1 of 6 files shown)


kani-driver/src/concrete_playback/playback.rs
```diff
@@ -96,6 +96,21 @@ fn build_test(install: &InstallType, args: &KaniPlaybackArgs) -> Result<PathBuf>
 /// TODO: This should likely be inside KaniSession, but KaniSession requires `VerificationArgs` today.
 /// For now, we just use InstallType directly.
 fn cargo_test(install: &InstallType, args: CargoPlaybackArgs) -> Result<()> {
+    // Recreating the match from `setup_cargo_command` here because this function takes InstallType
+    // This whole function needs refactoring to use KaniSession instead
```
celinval, line 100: Not really. The correct fix is to get rid of InstallType

kani-driver/src/session.rs
```diff
@@ -64,6 +64,31 @@ impl KaniSession {
         })
     }
 
+    /*
+    Setup the default version of cargo being run, based on the type/mode of installation for kani
+    If kani is being run in developer mode, then we use the one provided by rustup as we can assume that the developer will have rustup installed
+    For release versions of Kani, we use a version of cargo that's in the toolchain that's been symlinked during `cargo-kani` setup. This will allow
+    Kani to remove the runtime dependency on rustup later on.
+    */
+    pub fn setup_cargo_command(&self) -> Result<Command> {
```
celinval, line 73: This function doesn't really need to be part of KaniSession. If you move it out, you don't need to duplicate the code.

src/lib.rs
```diff
@@ -51,28 +54,56 @@ pub fn proxy(bin: &str) -> Result<()> {
...
-    match &args_ez[..] {
-        &[_, Some("setup"), Some("--use-local-bundle"), _] => {
-            ArgsResult::ExplicitSetup { use_local_bundle: Some(args[3].clone()) }
-        }
-        &[_, Some("kani"), Some("setup"), Some("--use-local-bundle"), _] => {
-            ArgsResult::ExplicitSetup { use_local_bundle: Some(args[4].clone()) }
-        }
-        &[_, Some("setup")] | &[_, Some("kani"), Some("setup")] => {
-            ArgsResult::ExplicitSetup { use_local_bundle: None }
+    let matches = App::new("kani")
```
celinval, line 63: Can you use the derive mode instead?

src/setup.rs
```diff
@@ -139,14 +142,22 @@ pub(crate) fn get_rust_toolchain_version(kani_dir: &Path) -> Result<String> {
...
     // Currently this means we require the bundle to have been unpacked first!
     let toolchain_version = get_rust_toolchain_version(kani_dir)?;
     println!("[3/5] Installing rust toolchain version: {}", &toolchain_version);
-    Command::new("rustup").args(["toolchain", "install", &toolchain_version]).run()?;
 
-    let toolchain = home::rustup_home()?.join("toolchains").join(&toolchain_version);
+    // Symlink to a local toolchain if the user explicitly requests
+    if let Some(local_toolchain_path) = use_local_toolchain {
+        let toolchain_path = Path::new(&local_toolchain_path);
+        // TODO: match the version against which kani was built
```
celinval, line 153: Add issue link please

celinval (commented):
Looks good. Just have a few smaller improvements. 

Question, how did you test this so far?

jaisnan:
> Looks good. Just have a few smaller improvements.
> 
> Question, how did you test this so far?

I have tested it locally so far, but can include a CI test for --use-local-toolchain as well that replicates the setup process.
What I've done locally is the following:

1. Build Kani using cargo build-dev. This installs the toolchain mentioned in rust-toolchain.toml
2. Then, delete that nightly toolchain in ~/.rustup/toolchains
3. Fetch the same toolchain's tarball from the rust lang server
4. Unzip tarball, install using install script into a specific directory inside kani work directory (though this can be anywhere else as well) using --prefix=/toolchain_path/
5. setup using `cargo kani setup --use-local-toolchian ./toolchain_path/
6. Run Kani on a few example crates

So far, this has been a manual process, but I don't see why we can't make it CI test. Would this process work?

celinval:
Sure, but please skip step 1 and 2 and maybe download the toolchain into a "random" directory. Also, we had issues with simlink before when the target is in a different partition. Would that be an issue here too?

src/lib.rs
jaisnan: Will add this in my cleanup issue and PR.

celinval (approved):
Please add a CI job to test the new setup path. It would also be nice to switch to clap derive later.

PR #3062, opened 2024-03-06
Add method to assert a pointer is valid

celinval:
This new method will assert that a pointer isn't null, and that its is valid for access of a given type.

I'm creating this as a draft for now as I'm still adding more tests.


library/kani/src/mem.rs
```diff
@@ -0,0 +1,282 @@
...
+//!    be dereferenceable: the memory range of the given size starting at the pointer must all be
+//!    within the bounds of a single allocated object. Note that in Rust, every (stack-allocated)
+//!    variable is considered a separate allocated object.
+//!    Even for operations of size zero, the pointer must not be pointing to deallocated memory,
+//!    i.e., deallocation makes pointers invalid even for zero-sized operations.
+//! 3. However, casting any non-zero integer literal to a pointer is valid for zero-sized
+//!    accesses, even if some memory happens to exist at that address and gets deallocated.
+//!    This corresponds to writing your own allocator: allocating zero-sized objects is not very
+//!    hard. The canonical way to obtain a pointer that is valid for zero-sized accesses is
+//!    NonNull::dangling.
```
zhassan-aws, line 19:
```suggestion
//!    `NonNull::dangling`.
```

library/kani/src/mem.rs
```diff
@@ -0,0 +1,282 @@
...
+//! 4. All accesses performed by functions in this module are non-atomic in the sense of atomic
+//!    operations used to synchronize between threads.
+//!    This means it is undefined behavior to perform two concurrent accesses to the same location
+//!    from different threads unless both accesses only read from memory.
+//!    Notice that this explicitly includes read_volatile and write_volatile:
```
zhassan-aws, line 24:
```suggestion
//!    Notice that this explicitly includes `read_volatile` and `write_volatile`:
```

library/kani/src/mem.rs
```diff
@@ -0,0 +1,282 @@
...
+//! from a non-zero integer literal.
+
+use crate::kani_intrinsic;
+use crate::mem::private::Internal;
+use std::mem::{align_of, size_of};
+use std::ptr::{DynMetadata, NonNull, Pointee};
+
+/// Assert that the pointer is valid for access according to [crate::mem] conditions 1, 2 and 3.
+///
+/// Note that a unaligned pointer is still considered valid.
```
zhassan-aws, line 44:
```suggestion
/// Note that an unaligned pointer is still considered valid.
```

library/kani/src/mem.rs
```diff
@@ -0,0 +1,282 @@
...
+///
+/// TODO: Kani will automatically add those checks when a de-reference happens.
```
zhassan-aws, line 46:
```suggestion
/// TODO: Kani should automatically add those checks when a de-reference happens.
```

library/kani/src/mem.rs
```diff
@@ -0,0 +1,282 @@
...
+/// https://github.com/model-checking/kani/issues/2975
+///
+/// This function will either panic or return `true`. This is to make it easier to use it in
+/// contracts.
+#[crate::unstable(
+    feature = "mem-predicates",
+    issue = 2690,
+    reason = "experimental memory predicate API"
+)]
+pub fn assert_valid_ptr<T>(ptr: *const T) -> bool
```
zhassan-aws, line 56: Is the `bool` return necessary?

zhassan-aws (commented):
Awesome, thanks @celinval!

Can we add at least one `expected` test so that we can match against the descriptions of the new pointer checks (Dangling pointer, non-null pointer, valid pointer, etc.).

library/kani/src/mem.rs
celinval: It's not necessary, but it is handy for contracts, where it expects a boolean.

Not ideal, but I think it's better than forcing users to write `{ assert_valid_ptr(ptr); true }` in their contract.

zhassan-aws: I see. I'm not clear how this can be used with contracts though given the limitation in using them in assumptions. Perhaps we would need new special annotations, e.g. `#[kani::requires_valid_ptr(...)]`?

celinval: It is still sound to use this function inside an assumption. It is redundant but still sound since it cannot evaluate to false.

celinval: I added an expected test using contracts. Let me know if that's what you are looking for.

tests/expected/function-contract/valid_ptr.rs
```diff
@@ -0,0 +1,61 @@
...
+// kani-flags: -Zfunction-contracts -Zmem-predicates
+
+//! Test that it is sound to use `assert_valid_ptr` inside a contract pre-condition.
+//! We cannot validate post-condition yet. This can be done once we fix:
+//! <https://github.com/model-checking/kani/issues/2997>
+#![feature(pointer_is_aligned)]
+
+mod pre_condition {
+    /// This contract should succeed only if the input is a valid pointer.
+    #[kani::requires(kani::mem::assert_valid_ptr(ptr) && ptr.is_aligned())]
```
zhassan-aws, line 12: I find it ugly/non-intuitive to use an assertion inside a requires/ensures. I would prefer if we instead provide a separate attribute for pointer validity whose semantics are unambiguous. For example, we could add `#[kani::assert_valid_ptr(...)]` to make it clear that this attribute only asserts, both in the context of a `proof_for_contract` and `stub_verified`.

celinval: I totally agree, but I'm also optimistic that this is a temporary solution until we have a proper method to query pointer validity. I'm not sure it grants a new attribute.

In fact, we would need two different attributes, one for pre and another for post-condition. We would also need to define order of evaluation, which might not be straight forward. This would also not work well if the pointer validity check is only relevant if another predicate is true.

zhassan-aws (approved):
Approving as this is a temporary solution per the discussions.

celinval:
> Approving as this is a temporary solution per the discussions.

Would you like me to document that this signature is temporary? The method is marked as unstable too.

PR #3063, opened 2024-03-07
Emit `dead` goto-instructions on MIR StatementDead

### LINKED ISSUE #3061: Kani should fail due to invalid access of a dead local variable via raw pointer
I tried this code:
```rust
#[kani::proof]
pub fn check_invalid_ptr() {
    let raw_ptr = {
        let var = 10;
        &var as *const _
    };
    // This should fail since it is de-referencing a dead object.
    assert_eq !(unsafe {*raw_ptr}, 10);
}

```

```
kani invalid_ptr.rs
```

with Kani version: 0.47

I expected to see this happen: Verification should fail since the de-referenced pointer is pointing to a dead object.

karkhaz:
This commit adds a new `Dead` goto-instruction that gets codegened
whenever Kani sees a MIR `StatementDead` statement. This new
goto instruction corresponds to the CBMC [code_deadt](
https://diffblue.github.io/cbmc/classcode__deadt.html) statement
that marks the point where a local variable goes out of scope.

This new instruction is needed to detect invalid accesses of dead local
variables.

The commit also codegens a CBMC `Decl` instruction upon seeing a MIR
StatementLive. This ensures that variables that go out of scope at the
end of a loop are not falsely marked as having a dead dereference when
they are accessed on the next loop iteration.

Resolves #3061

### DIFF
--- a/kani-driver/src/call_single_file.rs
+++ b/kani-driver/src/call_single_file.rs
@@ -122,6 +122,8 @@ impl KaniSession {
                 "symbol-mangling-version=v0",
                 "-Z",
                 "panic_abort_tests=yes",
+                "-Z",
+                "sanitizer=address",
             ]
             .map(OsString::from),
         );
--- a/tests/expected/dead-invalid-access-via-raw/main.rs
+++ b/tests/expected/dead-invalid-access-via-raw/main.rs
@@ -1,0 +1,12 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#[kani::proof]
+pub fn check_invalid_ptr() {
+    let raw_ptr = {
+        let var = 10;
+        &var as *const _
+    };
+    // This should fail since it is de-referencing a dead object.
+    assert_eq!(unsafe { *raw_ptr }, 10);
+}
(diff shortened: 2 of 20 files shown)


kani-driver/src/call_single_file.rs
```diff
@@ -122,6 +122,8 @@ impl KaniSession {
                 "symbol-mangling-version=v0",
                 "-Z",
                 "panic_abort_tests=yes",
+                "-Z",
+                "sanitizer=address",
```
celinval, line 126: Can you please make sure the code from the standard library has these statements?

If not, we may have to add the same argument here: https://github.com/model-checking/kani/blob/ea710b36/tools/build-kani/src/sysroot.rs#L89

celinval (commented):
It looks good. Can you also please add a cargo test to ensure dependencies also contains these statements?

For that, you will have to create 2 crates. You add the defective code in one crate, and the harness in the other crate. Thanks!

tests/expected/dead-invalid-access-via-raw/main.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: Please add a comment that explains what this test is checking. It would be useful to add a reference to #3061 as well.

kani-driver/src/call_single_file.rs
zhassan-aws: Per our offline discussion, we will follow up on this separately.

PR #3064, opened 2024-03-08
Replace internal reverse_postorder by a stable one

celinval:
The `reverse_postorder` function used before is internal to the compiler and reflects the internal body representation.

Besides that, I would like to introduce transformation on the top of SMIR body, which will break the 1:1 relationship between basic blocks from internal body and monomorphic body from StableMIR.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
@@ -74,6 +74,13 @@ impl<'tcx> GotocCtx<'tcx> {
     }
 }
 
+/// Iterate over the basic blocks in reverse post-order.
+///
+/// The `reverse_postorder` function used before was internal to the compiler and reflected the
+/// internal body representation.
+///
+/// As we introduce transformations on the top of SMIR body, there will be not guarantee of a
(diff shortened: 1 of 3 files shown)


zhassan-aws:
Can you explain the motivation for this change? Is the compiler's `reverse_postorder` function unstable in some way?

celinval:
> Can you explain the motivation for this change? Is the compiler's `reverse_postorder` function unstable in some way?

Totally. I updated the description. Let me know if that answers your question.

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
```diff
@@ -72,3 +73,25 @@ impl<'tcx> GotocCtx<'tcx> {
         }
     }
 }
+
```
zhassan-aws, line 76: Can you replicate the explanation in the PR's description here? Just want to make sure we don't wonder in the future why we're not using the version available from the compiler :)

zhassan-aws (approved):

kani-compiler/src/codegen_cprover_gotoc/codegen/block.rs
celinval: sure

PR #3074, opened 2024-03-13
Add CI test for --use-local-toolchain

jaisnan:
Adds a CI test for --use-local-toolchain.
This test will run after every push to main.

Additional cleanup to the setup as mentioned in #3060 in next PR's. (In interest of keeping changes small & focused).


.github/workflows/check-setup.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Ensure that the rustup toolchain is not present
+        run: ls -la ~/.rustup/toolchains/
+
+      - name: Run tests
+        run: |
+          ls ./tests/cargo-ui/
+          for dir in function multiple-harnesses verbose; do
+            >&2 echo "Running test $dir"
+            pushd ./tests/cargo-ui/$dir
+            cargo kani
```
celinval, line 76: Won't this step install the toolchain?

.github/workflows/check-setup.yml
```diff
@@ -0,0 +1,78 @@
...
+          curl --proto '=https' --tlsv1.2 -O https://static.rust-lang.org/dist/$DATE/rust-nightly-${{ matrix.rust_target }}.tar.gz
+          tar -xzf rust-nightly-${{ matrix.rust_target }}.tar.gz
+          ./rust-nightly-${{ matrix.rust_target }}/install.sh --prefix=${{ github.workspace }}/custom_toolchain
+
+      - name: Ensure installation is correct
+        run: |
+          cargo kani setup --use-local-toolchain ${{ github.workspace }}/custom_toolchain/
+
+      - name: Ensure that the rustup toolchain is not present
+        run: ls -la ~/.rustup/toolchains/
```
celinval, line 68: This is logging but not checking.

celinval (commented):
I recommend not running any command from Kani's folder, since Rustup will see the toolchain file and install the toolchain.

.github/workflows/check-setup.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Run tests
```
celinval, line 70: Can you please also test standalone `kani` as well as `--version` and `--help`? :)

jaisnan: Couldn't agree more lol

.github/workflows/check-setup.yml
```diff
@@ -65,14 +65,33 @@ jobs:
...
             popd
           done
+
+      - name: Check --help and --version
+        run: |
+          cargo kani --help && cargo kani --version
+
+      - name: Run standalone kani test
+        run: |
+          for dir in Assert; do
```
celinval, line 92: you don't need this for loop, do you? Also, maybe get a test from `tests/kani` instead.

.github/workflows/check-setup.yml
```diff
@@ -0,0 +1,97 @@
...
+          ./rust-nightly-${{ matrix.rust_target }}/install.sh --prefix=${{ github.workspace }}/custom_toolchain
+
+      - name: Ensure installation is correct
+        run: |
+          cargo kani setup --use-local-toolchain ${{ github.workspace }}/custom_toolchain/
+
+      - name: Ensure that the rustup toolchain is not present
+        run: |
+          ls -la ~/.rustup/toolchains/
+          if [ ! -f "~/.rustup/toolchains/" ]; then
```
karkhaz, line 70: I suggest `-d` rather than `-f`, which checks that the path is a regular file, but `toolchains` is a directory so I think the else statement will never be reached here. `-d` checks that the path is a directory and `-e` checks that it exists (any file type).

jaisnan: Thank you!

.github/workflows/check-setup.yml
```diff
@@ -0,0 +1,191 @@
...
+          os: ubuntu-18.04
+
+      - name: Build bundle
+        id: bundle
+        uses: ./.github/actions/build-bundle
+        with:
+          os: linux
+          arch: x86_64-unknown-linux-gnu
+
+  test-use-local-toolchain:
```
celinval, line 98: Can you move this job to be part of the Release bundle workflow and remove the rest (which seems to be duplicated)

.github/workflows/check-setup.yml
```diff
@@ -0,0 +1,191 @@
...
+          path: kani
+
+      - name: Setup Kani Dependencies
+        uses: ./kani/.github/actions/setup
+        with:
+          os: ${{ matrix.os }}
+          kani_dir: 'kani'
+
+      - name: Build Kani
+        run: cargo install --locked --path ./kani
```
celinval, line 128: Why do you build Kani here? You should be able to download package from the previous job and install it from there. Please see how we do that in the release workflow:

https://github.com/model-checking/kani/blob/f8c30d9d5849a8abcef8df4aae11c2bc61f7213f/.github/workflows/release.yml#L131-L135

.github/workflows/check-setup.yml
```diff
@@ -0,0 +1,191 @@
...
+          else
+            echo "::error::Default toolchain exists despite not installing."
+            exit 1
+          fi
+
+      - name: Run cargo-kani tests
+        run: |
+          for dir in function multiple-harnesses verbose; do
+            >&2 echo "Running test $dir"
+            pushd ./tests/cargo-ui/$dir
```
celinval, line 177: also, I don't think this directory exists, since you checked out Kani inside `kani` folder.

.github/workflows/check-setup.yml
celinval: BTW, this will happen if the folder you are running cargo from, or one of its parents folder, has a file `rust-toolchain.toml`. To avoid that to happen, you should either remove the `rust-toolchain.toml` file from the `kani` folder, or move the tests folder to be outside of `kani` folder.

celinval (approved):

PR #3076, opened 2024-03-13
Add a multi-crate test for #3061

zhassan-aws:
This is a follow-up on #3063 that adds a test with multiple crates to make sure this scenario is correctly handled and that Kani reports the bug.


celinval (approved):
Thank you!

PR #3077, opened 2024-03-14
Add optional scatterplot to benchcomp output

tautschnig:
Scatterplots should make it easier to immediately spot performance trends (and indeed any differences) rather than having to process a (large) number of table rows.

Uses mermaid-js to produce markdown-embedded plots that will display on the job summary page. Scatterplots are not directly supported by mermaid-js at this point (xycharts only do line or bar charts), so quadrant plots are employed with various diagram items drawn in white to make them disappear.

### DIFF
--- a/tools/benchcomp/benchcomp/visualizers/__init__.py
+++ b/tools/benchcomp/benchcomp/visualizers/__init__.py
@@ -247,7 +247,7 @@ class dump_markdown_results_table:
 
             {% if scatterplot and metric in d["scaled_metrics"] and d["scaled_variants"][metric]|length == 2 -%}
             ```mermaid
-            %%{init: { "quadrantChart": { "titlePadding": 10, "xAxisLabelPadding": 20, "yAxisLabelPadding": 20, "quadrantLabelFontSize": 0, "pointRadius": 2, "pointLabelFontSize": 2 }, "themeVariables": { "quadrant1Fill": "#FFFFFF", "quadrant2Fill": "#FFFFFF", "quadrant3Fill": "#FFFFFF", "quadrant4Fill": "#FFFFFF", "quadrant1TextFill": "#FFFFFF", "quadrant2TextFill": "#FFFFFF", "quadrant3TextFill": "#FFFFFF", "quadrant4TextFill": "#FFFFFF", "quadrantInternalBorderStrokeFill": "#FFFFFF" } } }%%
+            %%{init: { "quadrantChart": { "titlePadding": 15, "xAxisLabelPadding": 20, "yAxisLabelPadding": 20, "quadrantLabelFontSize": 0, "pointRadius": 2, "pointLabelFontSize": 2 }, "themeVariables": { "quadrant1Fill": "#FFFFFF", "quadrant2Fill": "#FFFFFF", "quadrant3Fill": "#FFFFFF", "quadrant4Fill": "#FFFFFF", "quadrant1TextFill": "#FFFFFF", "quadrant2TextFill": "#FFFFFF", "quadrant3TextFill": "#FFFFFF", "quadrant4TextFill": "#FFFFFF", "quadrantInternalBorderStrokeFill": "#FFFFFF" } } }%%
             quadrantChart
                 title {{ metric }}
                 x-axis {{ d["scaled_variants"][metric][0] }}
--- a/tools/benchcomp/benchcomp/visualizers/__init__.py
+++ b/tools/benchcomp/benchcomp/visualizers/__init__.py
@@ -257,7 +257,9 @@ class dump_markdown_results_table:
                 quadrant-3 3
                 quadrant-4 4
                 {%- for bench_name, bench_variants in d["scaled_metrics"][metric].items () %}
-                "{{ bench_name }}": [{{ bench_variants[d["scaled_variants"][metric][0]] }}, {{ bench_variants[d["scaled_variants"][metric][1]] }}]
+                {% set v0 = bench_variants[d["scaled_variants"][metric][0]] -%}
+                {% set v1 = bench_variants[d["scaled_variants"][metric][1]] -%}
+                "{{ bench_name }}": [{{ v0|round(3) }}, {{ v1|round(3) }}]
                 {%- endfor %}
             ```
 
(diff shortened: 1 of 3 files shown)


tautschnig:
See https://github.com/model-checking/kani/actions/runs/8296849261?pr=3077 for what those diagrams now look like.

tools/benchcomp/benchcomp/visualizers/__init__.py
```diff
@@ -206,12 +223,47 @@ def _eval_column_text(column_spec):
...
+            sys.exit(1)
+
+
     @staticmethod
     def _get_template():
         return textwrap.dedent("""\
             {% for metric, benchmarks in d["metrics"].items() %}
             ## {{ metric }}
 
+            {% if scatterplot and metric in d["scaled_metrics"] and d["scaled_variants"][metric]|length == 2 -%}
```
jaisnan, line 248: Non blocking but any chance these values can be stored separately in a config structure and re-used wherever? In case we want to change values later etc.

tools/benchcomp/test/test_regression.py
```diff
@@ -461,6 +462,20 @@ def test_markdown_results_table(self):
                 run_bc.stdout, textwrap.dedent("""
                     ## runtime
 
+                    ```mermaid
```
jaisnan, line 465: Same here.

jaisnan (commented):
From the images shown, can you leave a comment here or in documentation somewhere on how to interpret those values? What do the values on the top-right of the plot mean, what do the values on the bottom-left mean etc?

tools/benchcomp/benchcomp/visualizers/__init__.py
tautschnig: Which "values" are you specifically referring to?

tools/benchcomp/test/test_regression.py
tautschnig: Same as ... what?

tools/benchcomp/benchcomp/visualizers/__init__.py
jaisnan: Ah apologies, should have pointed to line 250. I meant the values initialized for the quadrantChart. For example pointRadius: 2. If we could store these values separately, imo it would make them more re-usable. Hope that clarifies.

tautschnig:
> From the images shown, can you leave a comment here or in documentation somewhere on how to interpret those values? What do the values on the top-right of the plot mean, what do the values on the bottom-left mean etc?

I am now adding text output to spell out the axis ranges, thank you for calling this out!

tools/benchcomp/benchcomp/visualizers/__init__.py
tautschnig: I could put those in the Python code in some other place and make Jinja read that variable if that's what you are asking for? I don't really think that makes much of a difference, though? If you are asking for a configuration file: GitHub does not support this, all plot configuration must be done inline.

jaisnan: I was referring to doing jinja read the variable from a python enum, but you're right, not much of a difference. You can resolve this comment.

karkhaz (commented):
Thank you, just a few comments.

To be honest the latest version is still a bit hard to read. It's less the axis labels and more that I'm not confident that I would be able to spot a 20% increase or decrease from the *y*=*x* line, for example. Also if a particular point has moved far away from the *y*=*x* line, it's not that easy to determine which proof or benchmark that actually corresponds to.

As a totally different approach to quickly identifying outlying benchmarks---would it be easier if I actually implemented the `benchcomp filter` command, which allows you to filter benchmarks based on certain criteria before they get emitted to the Markdown table?

That way, we could print a "filtered" set of Markdown tables (containing only regressions) followed by the complete results. That might be superior to the current approach where regressions are highlighted in bold in the complete table.

So the action would look something like this

```yaml
      - name: Perf Regression Table
        run: |
          echo "# Regressions" >> "$GITHUB_STEP_SUMMARY"
          new/tools/benchcomp/bin/benchcomp \
            --config new/tools/benchcomp/configs/perf-regression.yaml \
            visualize --only dump_markdown_results_table >> "$GITHUB_STEP_SUMMARY"

      - name: Perf All Results Table
        run: |
          echo "# All Results" >> "$GITHUB_STEP_SUMMARY"
          new/tools/benchcomp/bin/benchcomp \
            --config new/tools/benchcomp/configs/perf-all-results.yaml \
            visualize --only dump_markdown_results_table >> "$GITHUB_STEP_SUMMARY"

      - name: Run other visualizations
        run: |
          new/tools/benchcomp/bin/benchcomp \
            --config new/tools/benchcomp/configs/perf-regression.yaml \
            visualize --except dump_markdown_results_table
```

tools/benchcomp/benchcomp/visualizers/__init__.py
```diff
@@ -228,7 +281,48 @@ def _get_variant_names(results):
...
 
     @staticmethod
-    def _organize_results_into_metrics(results):
+    def _compute_scaled_metric(data_for_metric, log_scaling):
+        min_value = None
+        max_value = None
+        for bench, bench_result in data_for_metric.items():
+            for variant, variant_result in bench_result.items():
+                if isinstance(variant_result, (bool, str)):
+                    return None
```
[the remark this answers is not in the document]
tautschnig: No, for all that I can tell it does not: `isinstance(True, int)` returns `True`.

tautschnig:
> To be honest the latest version is still a bit hard to read. It's less the axis labels and more that I'm not confident that I would be able to spot a 20% increase or decrease from the _y_=_x_ line, for example. Also if a particular point has moved far away from the _y_=_x_ line, it's not that easy to determine which proof or benchmark that actually corresponds to.
> 

Yes, I'd indeed love to be able to draw that diagonal line into the diagram, but I haven't been able to figure out a way to do this in Mermaid. Perhaps it's just about time to create bunch of feature requests against Mermaid.

> As a totally different approach to quickly identifying outlying benchmarks---would it be easier if I actually implemented the `benchcomp filter` command, which allows you to filter benchmarks based on certain criteria before they get emitted to the Markdown table?

Perhaps we should have all of these tools available and then see what helps us the most?

karkhaz (approved):
Thank you, looking forward to seeing this in CBMC also!

PR #3078, opened 2024-03-14
Revert clap from arg parsing during setup

jaisnan:
`--help` and `--version` commands were being overriden by clap, so we are going back to the simple parsing with the added logic for `use-local-toolchain`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,15 +34,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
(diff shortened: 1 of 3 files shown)


celinval:
I think we should unblock the release by merging this, but if you want to try re-adding clap later, you can disable the parser [`--help`](https://docs.rs/clap/latest/clap/struct.Command.html#method.disable_help_flag) and [`--version`](https://docs.rs/clap/latest/clap/struct.Command.html#method.disable_version_flag)

jaisnan:
> I think we should unblock the release by merging this, but if you want to try re-adding clap later, you can disable the parser [`--help`](https://docs.rs/clap/latest/clap/struct.Command.html#method.disable_help_flag) and [`--version`](https://docs.rs/clap/latest/clap/struct.Command.html#method.disable_version_flag)

Yeah, I just wanted to go to a quick solution we know works for sure. In the meanwhile, I'll try to refactor this change to use clap with those options.

zhassan-aws (approved):

Issue #3068, opened 2024-03-11
Remove `--enable-unstable`


celinval:
**Proposed change:** Deprecate and remove `--enable-unstable`. For legacy arguments like `--cbmc-args`, `--visualize`, `--function` and other unstable arguments, the proposal was to create a unstable feature `-Z unstable-options`.

For things that are only added for debug / development purpose, I propose we add something like `Z dev-features`.

**Motivation:** This is the missing part of the [Unstable APIs RFC](https://model-checking.github.io/kani/rfc/rfcs/0006-unstable-api.html).

PR #3080, opened 2024-03-14
Ensure storage markers are kept in std code

### LINKED ISSUE #3079: Kani may miss invalid accesses in std
The fix to https://github.com/model-checking/kani/issues/3061 only applies to code outside of `std`. We need to extend the fix to `std` as well, to make sure any invalid accesses in `std` are captured by Kani.

zhassan-aws:
This is a follow-up to #3063 that turns off that MIR pass while compiling `std` as well to ensure any bugs of the same nature in `std` are captured by Kani.

Resolves #3079 


zhassan-aws:
I moved the change that directly turns off the MIR pass that removes storage markers instead of using the address sanitizer into a separate PR (#3083). So this PR is now only concerned with keeping storage markers in `std`.

kani-driver/src/args/mod.rs
```diff
@@ -252,6 +252,14 @@ pub struct VerificationArgs {
     #[arg(long, hide_short_help = true, requires("enable_unstable"))]
     pub ignore_global_asm: bool,
 
+    /// Ignore lifetimes of local variables. This effectively extends their
+    /// lifetimes to the function scope, and hence may cause Kani to miss
+    /// undefined behavior resulting from using the variable after it dies.
+    /// This option may impact the soundness of the analysis and may cause false
+    /// proofs and/or counterexamples
+    #[arg(long, hide_short_help = true, requires("enable_unstable"))]
```
celinval, line 260: we shouldn't use `--enable-unstable` anymore

celinval: Instead, can you convert this flag into `-Z ignore_locals_lifetime`?

celinval (commented):
Just a small comment

kani-driver/src/args/mod.rs
zhassan-aws: I went back and forth on this one. The -Z flags seemed to me more like unstable features. Since this flag isn't really a feature, but a workaround for a bug, -Z didn't seem appropriate to me. I saw that https://github.com/model-checking/kani/issues/3068 mentioned adding a -Z unstable-options that can be used for such cases, so I thought I'd use enable-unstable for the time being. What do you think?

tautschnig (approved):

kani-driver/src/args/mod.rs
tautschnig: Resolving for the time being so that this PR can make it into the release. @celinval @zhassan-aws please follow-up if still needed.

PR #3083, opened 2024-03-15
Disable removal of storage markers

zhassan-aws:
Instead of turning on the address sanitizer to ensure storage markers (`StorageLive` and `StorageDead`) are kept in MIR, directly disable the MIR pass that removes them from MIR (https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/.E2.9C.94.20Keeping.20MIR's.20storage.20markers). This is to ensure we don't get additional instrumentation that is not relevant for Kani, and may unnecessarily increase the code size and compilation time.


karkhaz (approved):
Thank you!

PR #3084, opened 2024-03-16
Upgrade toolchain to nightly-2024-03-15

celinval:
- Kani can no longer automatically detect loop bounds when using range.

### DIFF
--- a/docs/src/getting-started/verification-results/src/main.rs
+++ b/docs/src/getting-started/verification-results/src/main.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 #[kani::proof]
+#[kani::unwind(4)]
 // ANCHOR: success_example
 fn success_example() {
     let mut sum = 0;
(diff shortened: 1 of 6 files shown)


docs/src/getting-started/verification-results/src/main.rs
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 #[kani::proof]
+#[kani::unwind(4)]
```
feliperodri, line 5: Why we didn't need to unwind these cases before?

feliperodri (approved):

adpaco-aws:
@karkhaz the test cases in this PR have regressed and Kani can no longer automatically unwind them.

docs/src/getting-started/verification-results/src/main.rs
celinval: I still haven't had time to investigate why

celinval:
I created https://github.com/model-checking/kani/issues/3088 to track the unwind issue.

docs/src/getting-started/verification-results/src/main.rs
celinval: I created #3088 for us to investigate what can be done to remove these attributes.

Issue #2998, opened 2024-02-07
Automatically ensure validity invariants of unsafe operations


celinval:
Requested feature: Kani should automatically check if unsafe operations can break validity invariants. This follows the concept of validity from https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html.
Use case: Checking that unsafe code doesn't trigger UB.
Link to relevant documentation (Rust reference, Nomicon, RFC):

Test case:
```rust
#[kani::requires(!ptr.is_null())]
#[kani::ensures(!result.as_ptr().is_null())] // Compiler warns that this should always be nonnull().
pub unsafe fn new_unchecked<T>(ptr: *mut T) -> NonNull<T> {
    NonNull::<T>::new_unchecked(ptr)
}
```

The ensure clause here might be redundant, since the compiler might decide to prune it, i.e., you cannot write Rust code to check for UB occurrence.

celinval:
Note for self: We should use implement https://github.com/rust-lang/project-stable-mir/issues/58 and use the valid range from Scalar type. We would need to traverse structures to add proper checks. Maybe experiment with transforming StableMIR instead of regular MIR. This would play well since we want the monomorphic version to add validity checks. Here are a few good places to see how rustc / MIRI handle these things:
- ValidityVisitor: https://github.com/rust-lang/rust/blob/master/compiler/rustc_const_eval/src/interpret/validity.rs#L655
- List of errors: https://github.com/rust-lang/rust/blob/d8b00690ec3aa2d2fb0d3eb00e862195501186f9/compiler/rustc_const_eval/messages.ftl#L416-L474 (Some are not redundant for this issue)

PR #3085, opened 2024-03-16
Implement validity checks

### LINKED ISSUE #301: Missing check for object correctness invariants
Rust requires that all values have to be valid when assigned to or read from a place, passed to a function/primitive operation or returned from a function/primitive operation.  It is UB to produce an invalid bit-pattern of a given type.  We do not currently have any checks for this in Kani.

## Likelihood:

If code contains this bug, Kani will not detect it. We do not have any data as to how often this occurs in practice.

## Mitigation: 

Warn users that this class of UB is outside the current scope of Kani.

## Path to soundness:

Add a check for this.

celinval:
This is still incomplete, but hopefully it can be merged as an unstable feature. I'll publish an RFC shortly.

This instruments the function body with assertion checks to see if users are generating invalid values. This covers:
  - Union access
  - Raw pointer dereference
  - Transmute value
  - Field assignment of struct with invalid values
  - Aggregate assignment

Things not covered today should trigger ICE or a delayed verification failure due to unsupported feature.

## Design

This change has two main design changes which are inside the new `kani_compiler::kani_middle::transform` module:
  1- Instance body should now be retrieved from the `BodyTransformation` structure. This structure will run transformation passes on instance bodies (i.e.: monomorphic instances) and cache the result.
  2- Create a new transformation pass that instruments the body of a function for every potential invalid value generation.
  3- Create a body builder which contains all elements of a function body and mutable functions to modify them accordingly.

## Call-outs
  1. I am planning to add expected tests in a follow up PR.
  2. This PR is currently blocked since it requires toolchain nightly-2024-03-15.
  3. One potential improvement for later is to replace most of the build check logic by an internal Kani library function that handle the comparison.

Related to #2998
Fixes #301 

### DIFF
--- a/kani-compiler/src/args.rs
+++ b/kani-compiler/src/args.rs
@@ -72,7 +72,7 @@ pub struct Arguments {
     /// A legacy flag that is now ignored.
     goto_c: bool,
     /// Enable specific checks.
-    #[clap(long, conflicts_with = "check_all")]
+    #[clap(long)]
     pub ub_check: Vec<ExtraChecks>,
 }
 
--- a/kani-compiler/src/args.rs
+++ b/kani-compiler/src/args.rs
@@ -74,9 +74,6 @@ pub struct Arguments {
     /// Enable specific checks.
     #[clap(long, conflicts_with = "check_all")]
     pub ub_check: Vec<ExtraChecks>,
-    /// Enable all extra checks.
-    #[clap(long)]
-    pub check_all: bool,
 }
 
 #[derive(Debug, Clone, Copy, AsRefStr, EnumString, VariantNames, PartialEq, Eq)]
--- a/kani-compiler/src/args.rs
+++ b/kani-compiler/src/args.rs
@@ -71,4 +71,18 @@ pub struct Arguments {
     #[clap(long)]
     /// A legacy flag that is now ignored.
     goto_c: bool,
+    /// Enable specific checks.
+    #[clap(long, conflicts_with = "check_all")]
+    pub ub_check: Vec<ExtraChecks>,
+    /// Enable all extra checks.
+    #[clap(long)]
+    pub check_all: bool,
+}
+
+#[derive(Debug, Clone, Copy, AsRefStr, EnumString, VariantNames, PartialEq, Eq)]
+#[strum(serialize_all = "snake_case")]
+pub enum ExtraChecks {
+    /// Check that produced values are valid except for uninitialized values.
+    /// See https://github.com/model-checking/kani/issues/920.
+    Validity,
 }
--- a/kani-compiler/src/kani_middle/transform/mod.rs
+++ b/kani-compiler/src/kani_middle/transform/mod.rs
@@ -1,0 +1,131 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+//! This module is responsible for optimizing and instrumenting function bodies.
+//!
+//! We make transformations on bodies already monomorphized, which allow us to make stronger
+//! decisions based on the instance types and constants.
+//!
+//! The main downside is that some transformation that don't depend on the specialized type may be
+//! applied multiple times, one per specialization.
+//!
+//! Another downside is that these modifications cannot be applied to concrete playback, since they
+//! are applied on the top of StableMIR body, which cannot be propagated back to rustc's backend.
+//!
+//! # Warn
+//!
+//! For all instrumentation passes, always use exhaustive matches to ensure soundness in case a new
+//! case is added.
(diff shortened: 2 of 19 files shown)


tests/kani/ValidValues/write_invalid.rs
```diff
@@ -0,0 +1,37 @@
...
+
+//! Check that Kani can identify UB after writing an invalid value.
+//! Writing invalid bytes is not UB as long as the incorrect value is not read.
+//! However, we over-approximate for sake of simplicity and performance.
+
+use std::num::NonZeroU8;
+
+#[kani::proof]
+#[kani::should_panic]
+pub fn write_invalid_bytes_no_ub() {
```
zhassan-aws, line 13: This test seems to be subsumed by the next one.

zhassan-aws (commented):
The first round of comments. I haven't finished reviewing the entire PR yet.

tests/kani/ValidValues/custom_niche.rs
```diff
@@ -0,0 +1,151 @@
...
+        Rating { stars }
+    }
+}
+
+#[kani::proof]
+#[kani::should_panic]
+pub fn check_new_with_ub() {
+    assert_eq!(Rating::new(10), None);
+}
+#[test]
```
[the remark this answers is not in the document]
celinval: No. Thanks

tests/kani/ValidValues/custom_niche.rs
```diff
@@ -0,0 +1,151 @@
...
+    ];
+    kani::concrete_playback_run(concrete_vals, check_unchecked_new_ub);
+}
+
+#[kani::proof]
+#[kani::should_panic]
+pub fn check_new_with_ub_limits() {
+    let stars = kani::any_where(|s: &u8| *s == 0 || *s > 5);
+    let _ = Rating::new(stars);
+    unreachable!("Call to new should always fail");
```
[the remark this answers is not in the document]
celinval: I will remove that. I was planning to make this an expected test, but I'm waiting for some feedback on the RFC

zhassan-aws (commented):
Some nits

kani-compiler/src/args.rs
```diff
@@ -71,4 +71,18 @@ pub struct Arguments {
     #[clap(long)]
     /// A legacy flag that is now ignored.
     goto_c: bool,
+    /// Enable specific checks.
+    #[clap(long, conflicts_with = "check_all")]
+    pub ub_check: Vec<ExtraChecks>,
+    /// Enable all extra checks.
+    #[clap(long)]
+    pub check_all: bool,
```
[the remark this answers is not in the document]
celinval: Actually, I think I'll just remove this for now. I'll leave this for after we decide on #3092

kani_metadata/src/unstable.rs
```diff
@@ -84,6 +84,8 @@ pub enum UnstableFeature {
     FunctionContracts,
     /// Memory predicate APIs.
     MemPredicates,
+    /// Automatically check for that all values checks.
```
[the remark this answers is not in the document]
celinval: it's incomplete... let me fix that

tests/kani/ValidValues/write_invalid.rs
celinval: I actually would prefer keeping it. In an ideal world, this test should not panic. This test shows that the UB detection for this case today is an over-approximation. I could turn it into a fixme test, but it's not really a bug at this point, it is a decision made by design.

kani-compiler/src/kani_middle/transform/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub fn dummy() -> Self {
+        BodyTransformation { opt_passes: vec![], inst_passes: vec![], cache: Default::default() }
+    }
+
+    pub fn body(&mut self, tcx: TyCtxt, instance: Instance) -> Option<Body> {
+        match self.cache.get(&instance) {
+            Some(TransformationResult::Modified(body)) => Some(body.clone()),
+            Some(TransformationResult::NotModified) => instance.body(),
+            None => {
+                let mut body = instance.body()?;
```
zhassan-aws, line 69: Is there a way this can return `None`?

zhassan-aws (approved):
Awesome work, @celinval!

kani-compiler/src/kani_middle/transform/mod.rs
celinval: There are cases where instance.body() does return `None`, but I believe all callers to this function unwrap the result. So I could remove the `Option`.

zhassan-aws: That would make sense.

PR #3091, opened 2024-03-18
Stabilize cover statement and update contracts RFC

celinval:
I would like to propose that we stabilize the cover statement as is. Any further improvements or changes can be done separately, with or without an RFC.

I am also updating the contracts RFC status since parts of it have been integrated to Kani, but it is still unstable.

### Call-out

This PR requires at least 2 approvals.


feliperodri (approved):

adpaco-aws (approved):
Thanks @celinval

Issue #3089, opened 2024-03-18
Kani should detect Rust specific undefined behavior


celinval:
Requested feature: Kani should implement UB checks for all behavior listed [in the Rust reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html), such as value validity tests, pointer aliasing rules.
Use case: Rust code that uses unsafe
Link to relevant documentation (Rust reference, Nomicon, RFC): https://doc.rust-lang.org/reference/behavior-considered-undefined.html

Related issues:

- [ ] #3085 
- [ ] #920
- [ ] #314

PR #3092, opened 2024-03-18
RFC for Rust UB checks

celinval:
Create a new RFC for standardizing Kani's UB checks.

I left the "Software Design" section empty for now so we can first focus on the user experience.

Note that this RFC might need some adjustment depending on the outcome of https://github.com/diffblue/cbmc/pull/8242 and https://github.com/diffblue/cbmc/pull/8226.

Related to #3089 

Rendered version: https://celinval.github.io/kani-dev/rfc/rfcs/0010-rust-ub-checks.html


rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Rationale and alternatives
+
+Kani already has undefined behavior checks, some are implemented on the CBMC side, such as access out bounds, and
+arithmetic overflow.
+While others are implemented on Kani's side, such as intrinsics validation checks, and some are a
+result of instrumentation to the Rust standard library.
+
+The behavior of these checks varies according to how they are added.
+The ones on CBMC side do not interrupt the analysis of a path, while the ones on Kani / Rust std side will.
```
zhassan-aws, line 109:
```suggestion
The ones on the CBMC side do not interrupt the analysis of a path, while the ones on the Kani / Rust std side will.
```

rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
...
+
+We would like to keep taking advantage of existing instrumentation to maximize the UB check coverage.
+The mechanism used to build the check, however, shouldn't influence in the user experience.
+
+### Opt-out vs opt-in
+
+The default behavior matches the one that we believe is the recommended for most users due to soundness.
+In some cases, users may need to disable some checks for reasons such as spurious CEX or proof scalability.
+Thus, we suggest that every category of checks should have an opt-out mechanism that can either be global or local;
+i.e.: via command argument or harness attribute.
```
zhassan-aws, line 156: Since some of our checks are over-approximating, we may also need to introduce code-level control at some point, e.g. to enable/disable the checks for a few lines of code.

rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
...
+The detection of an undefined behavior should impact the status of any assertion that may be reachable from the
+detected undefined behavior.
+
+For simplicity, we propose that all undefined behavior checks are modelled as assert-assume pairs.
+We propose that the status of all passing assertions that may be affected by this check to be displayed as
+UNDETERMINED if one or more UB checks fail. [^all-passing]
+
+All failing assertions should have their status preserved, unless the UB check is implemented using "demonic"
+non-determinism, such as CBMC's deallocated check.
+See [0010-rust-ub-checks.md#open-questions] for alternatives to be considered.
```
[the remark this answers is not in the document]
celinval: Yeah, I need to fix that.

rfc/src/rfcs/0010-rust-ub-checks.md
celinval: Indeed. It might be hard for dependencies though.

rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
...
+
+Command line arguments should follow the existing name schema:
+`--no-[CHECK-NAME]-checks` where the CHECK-NAME is kebab case. There is no need to add `--[CHECK-NAME]-checks` since
+they are enabled by default.
+
+We propose to add the following attribute:
+
+```rust
+#[kani::proof]
+#[kani::ignore(CHECK_NAMES)]
```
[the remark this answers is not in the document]
celinval:
```suggestion
#[kani::disable(CHECK_NAMES)]
```

rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
...
+
+```rust
+#[kani::proof]
+#[kani::disable(CHECK_NAMES)]
+fn harness() {}
+```
+
+Where `CHECK_NAMES` is a list of check category names in snake case, such as `memory_safety`, `valid_values`.
+
+For checks that are unstable, Kani should support both opt-out mechanisms.
```
adpaco-aws, line 94: Why this requirement? I think it'll be trivial to support opt-outs once the groundwork is in place, but otherwise it's putting obstacles to unstable features.

rfc/src/rfcs/0010-rust-ub-checks.md
adpaco-aws: +1

rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
...
+    - Counter examples detected together with a non-deterministic UB check.
+    - Over-approximating UB checks.
+- How should UB checks affect concrete playback? Is a comment enough?
+    - Another possibility would be to add a `unreachable!()` statement at the end of the test case with a message
+      explaining why the statement should be unreachable.
+- Should we deprecate the `--no-default-checks` and all `--[CHECK-NAME]-checks` checks.
+
+## Out of scope / Future Improvements
+
+- **List enabled checks:** Kani could include a command that list all the categories of undefined behaviors supported.
```
adpaco-aws, line 173: I'd like to have a table (generated from the source code?) which is available in our online documentation.

rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
...
+further instrumentation of undefined behavior in Kani.
+
+## User Experience
+
+A UB check should behave consistently independent of how and where it is implemented.
+The detection of an undefined behavior should impact the status of any assertion that may be reachable from the
+detected undefined behavior.
+
+For simplicity, we propose that all undefined behavior checks are modelled as assert-assume pairs.
+We propose that the status of all passing assertions that may be affected by this check to be displayed as
```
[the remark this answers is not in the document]
adpaco-aws: Yep, I don't think this became clearer to me after reviewing the rest of the RFC. Why only passing assertions? Don't you want to refer to any checks in general, also regardless of their status?

rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
...
+For example, the code below shows how reading a boolean value in safe code may trigger UB due to an
+invalid write:
+
+```rust
+#[kani::proof]
+fn buggy_code() {
+    let mut value: bool = false;
+    unsafe {
+        let ptr: *mut u8 = &mut value as *mut _ as *mut u8;
+        *ptr = 8; // This assignment does not cause UB...
```
adpaco-aws, line 52: According to [the reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html), producing the value is undefined behavior in Rust.

adpaco-aws (commented):
Thanks, @celinval ! I'm all down for a more consistent and intuitive experience like the one proposed in this RFC.

However, I'd have liked to see some content regarding categorization of UB checks. At the same time, I acknowledge that this subject deserves its own RFC, so I'd be satisfied if this RFC included a forward declaration about a future categorization effort that tries to be as exhaustive as possible (think of it as a more detailed version of [this reference section](https://doc.rust-lang.org/reference/behavior-considered-undefined.html)).

rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
+- **Feature Name:** Rust UB Checks (`rust-ub-checks`)
+- **Feature Request Issue:** [#3089](https://github.com/model-checking/kani/issues/3089)
+- **RFC PR:** [#XXXX](https://github.com/model-checking/kani/pull/3091)
+- **Status:** Under Review
```
tautschnig, line 4: Nit pick: I suppose the above need to be updated?!

rfc/src/rfcs/0010-rust-ub-checks.md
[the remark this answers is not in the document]
celinval: I believe we have two sound options here:
  1. Change the status of ALL passing assertions as undetermined.
  2. Change the status of passing assertions only if they are downstream from the UB check that has failed. I.e., do something similar to what is being proposed in the Fatal Assertions PR (https://github.com/diffblue/cbmc/pull/8226). In this case, you are traversing the control flow graph to rule out assertions that can only occur before the UB check.

I did leave this slightly ambiguous on purpose. Both options above are an over-approximation of the set of assertions that could be reached after the UB. The option 2 is an optimization of option 1, and I'm curious on how function pointers and other things will play out here.

In my opinion, they are both acceptable (as long as they are sound) and better than what we have today. I am not convinced the optimization adds a lot of value, since users shouldn't really be using partial results from a program with UB.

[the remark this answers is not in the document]
celinval: I tried to make it more explicit. Let me know if that makes sense now.

rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
...
+
+For example, the code below shows how reading a boolean value in safe code may trigger UB due to an
+invalid write:
+
+```rust
+#[kani::proof]
+fn buggy_code() {
+    let mut value: bool = false;
+    unsafe {
+        let ptr: *mut u8 = &mut value as *mut _ as *mut u8;
```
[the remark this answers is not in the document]
celinval: Because we first need to cast the reference to a raw pointer, then we need to cast to a `u8` pointer.

rfc/src/rfcs/0010-rust-ub-checks.md
celinval: Humm... I think you are correct. Miri only detects this failure in the `assert` statement, but possibly because it doesn't track the original value type. Let me clarify with the unsafe code guidelines.

celinval: BTW, I'm changing the example here.

rfc/src/rfcs/0010-rust-ub-checks.md
```diff
@@ -0,0 +1,175 @@
...
+It may not be scalable for Kani to verify that every value produced is valid.
+In cases like that, we may opt for implementing one of the following UB checks:
+
+1. Perform static analysis to first identify which code may be affected by unsafe code, and only instrument those that
+   are reachable from the unsafe code.
+2. Add eager UB detection in unsafe code that will fail verification if there are side effects that we may not be
+   able to detect efficiently. For example, we could fail if casting between types that have different value
+   requirements, i.e.: fail in the cast `*mut bool as *mut u8`.
+
+For option 2, Kani may fail verification even if the code does not have UB.
```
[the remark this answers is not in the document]
celinval: I agree that this is not ideal, but the ideal solution may not scale to basic examples. 

In cases where option 1 is not feasible, or not yet implemented, option 2 may enable us to verify most existing code out there and fail for very specific use cases.

For example, we won't likely be able to implement a full stacked borrows in Kani since the runtime overhead can be overwhelming. In that case, modeling an over-abstraction may allow us to reason about aliasing violation in most use cases, but we may need to reject some code that is accepted by Miri.

I would be OK with that, as long as the failure explicitly mentions that this may be due to an over-approximation. Users would still be able to run the counter-example in Miri to check if it is spurious or not. Maybe we can even eventually integrate with Miri to check that.

rfc/src/rfcs/0010-rust-ub-checks.md
celinval: Removed this

rfc/src/rfcs/0010-rust-ub-checks.md
celinval: Is it OK if I leave this as future work?

rfc/src/rfcs/0010-rust-ub-checks.md
celinval: I would love that too. :smiley:

rfc/src/rfcs/0014-rust-ub-checks.md
```diff
@@ -0,0 +1,187 @@
...
+
+Command line arguments should follow the existing name schema:
+`--no-[CHECK-NAME]-checks` where the CHECK-NAME is kebab case. There is no need to add `--[CHECK-NAME]-checks` since
+they are enabled by default.
+
+We propose to add the following attribute:
+
+```rust
+#[kani::proof]
+#[kani::ignore(CHECK_NAMES)]
```
zhassan-aws, line 99: "ignore" is a bit too general. Maybe use `ignore_ub`?

rfc/src/rfcs/0014-rust-ub-checks.md
```diff
@@ -0,0 +1,187 @@
...
+For checks that are unstable, they should only be included if their respective unstable flag are passed, i.e.,
+`-Z [CHECK-NAME]`.
+
+## Software Design
+
+*To be filled in the next iteration*
+
+## Rationale and alternatives
+
+Kani already has undefined behavior checks, some are implemented on the CBMC side, such as access out bounds, and
```
zhassan-aws, line 114:
```suggestion
Kani already has undefined behavior checks, some are implemented on the CBMC side, such as out-of-bounds access, and
```

rfc/src/rfcs/0014-rust-ub-checks.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+Failing due to an assumption check inside `get_unchecked` function.
+
+```plaintext
+Failed Checks: assumption failed
+```
+
+We would like to keep taking advantage of existing instrumentation to maximize the UB check coverage.
+The mechanism used to build the check, however, shouldn't influence in the user experience.
```
zhassan-aws, line 160:
```suggestion
The mechanism used to build the check, however, shouldn't influence the user experience.
```

rfc/src/rfcs/0014-rust-ub-checks.md
```diff
@@ -0,0 +1,187 @@
...
+## Open questions
+
+- Should we add a new status for potentially spurious counter examples? Such as:
+    - Counter examples detected together with a non-deterministic UB check.
+    - Over-approximating UB checks.
+- How should UB checks affect concrete playback? Is a comment enough?
+    - Another possibility would be to add a `unreachable!()` statement at the end of the test case with a message
+      explaining why the statement should be unreachable.
+- Should we deprecate the `--no-default-checks` and all `--[CHECK-NAME]-checks` checks.
+- Should we add an attribute that allows us to opt-out some UB checks for a function or some code block?
```
zhassan-aws, line 181:
```suggestion
- Should we add an attribute that allows us to opt-out of some UB checks for a function or some code block?
```

zhassan-aws (approved):

zhassan-aws:
Please take a look at the modifications suggested in the previous round of review as well.

rfc/src/rfcs/0014-rust-ub-checks.md
carolynzech: +1

rfc/src/rfcs/0014-rust-ub-checks.md
```diff
@@ -0,0 +1,187 @@
...
+    let idx: usize = 4;
+    assert_eq!(unsafe { *var.get_unchecked(idx) }, 0);
+}
+```
+
+Failing due to an assumption check inside `get_unchecked` function.
+
+```plaintext
+Failed Checks: assumption failed
+```
```
carolynzech, line 157:
```suggestion
Failing due to an [assumption check](https://github.com/rust-lang/rust/blob/a1d7676d6a8c6ff13f9165e98cc25eeec66cb592/library/core/src/slice/index.rs#L255) inside the `get_unchecked` function.

```plaintext
Failed Checks: assumption failed
File: "library/core/src/slice/index.rs", line 255, in <usize as std::slice::SliceIndex<[i32]>>::get_unchecked
```

carolynzech: (Just to make it a bit clearer that this is an assumption in the standard library itself, not some Kani model of `get_unchecked`).

rfc/src/rfcs/0014-rust-ub-checks.md
carolynzech: This section does a great job of explaining the current division of UB checks and how those manifest in different experiences for users, but I'm missing a call to action--what are you proposing we do about it? Should we add a model of `get_unchecked` to Kani so that we can catch this error ourselves and provide a better error message? Or not, because then we're not relying on existing instrumentation? We don't need to land on a precise answer now (I know the implementation section is empty), but can we give some options at least?
Also, I'm a bit confused about the scope of this RFC. The User Experience section made me think that we were just talking about whether a passing check should pass or be undetermined based on a failing UB check before it. Now it seems like we're also considering the UX of the UB checks themselves. I would recommend adding a sentence or two about consistent messaging for failing UB checks to the User Experience section if you intend for that to be part of the "consistent framework" you're proposing.

rfc/src/rfcs/0014-rust-ub-checks.md
```diff
@@ -0,0 +1,187 @@
...
+Provide a consistent user experience for all UB checks in Kani, while establishing the best mechanisms for
+further instrumentation of undefined behavior in Kani.
+
+## User Experience
+
+A UB check should behave consistently independent on how and where it is implemented.
+The detection of an undefined behavior should impact the status of any assertion that may be reachable from the
+detected undefined behavior.
+
+For simplicity, we propose that all undefined behavior checks are modelled as assert-assume pairs.
```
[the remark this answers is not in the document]
carolynzech: Ok, after finishing the RFC, I believe the demonic nondeterminism mention later addresses the soundness concerns from the linked comment.

carolynzech (commented):
Thanks! High-level comment: I think the RFC would benefit from a longer summary that outlines the key elements of what defines a "consistent framework." Bullet points are fine--something like:
- Consistent standards for the status of assertions downstream of a failing UB check
- Consistent error message style for failing UB checks, regardless of whether the UB was detected in Kani, CBMC, or the Rust standard library
- Consistent mechanisms by which a user can opt in or out of a given UB check

Laying this out in the beginning would give the reader a better idea of what they can expect from reading the rest of the RFC.

rfc/src/rfcs/0014-rust-ub-checks.md
```diff
@@ -0,0 +1,187 @@
...
+We propose that the status of all passing assertions that may be affected by this check to be displayed as
+`UNDETERMINED` if one or more UB checks fail.
+For simplicity, Kani currently over-approximate this set by marking all passing assertions as `UNDETERMINED`.
```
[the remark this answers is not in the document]
celinval: I actually don't want to be super specific at this point. Maybe let me change the proposal to say that we shall mark all passing assertions as `UNDETERMINED`, and mentioned that refining the set of assertions marked as `UNDETERMINED` is out of scope of this RFC.

celinval:
@zhassan-aws @carolynzech thank you for the latest round of review.

I gave this PR a second thought, and I decided to close it. Unfortunately, I don't think I have enough time to address the open questions, and maybe the UX we want can be driven by the std verification effort. It will give us opportunity to incorporate more user feedback and ensure we have a robust solution.

PR #3105, opened 2024-03-25
Add `benchcomp filter` command

karkhaz:
This allows benchcomp to pass the list of results to an external program for modification before the results are visualized. This can be used, for example, to visualize only a relevant subset of results.

### DIFF
--- a/docs/src/benchcomp-conf.md
+++ b/docs/src/benchcomp-conf.md
@@ -40,6 +40,7 @@ filters:
 The value of `filters` is a list of dicts.
 Currently the only legal key for each of the dicts is `command_line`.
 Benchcomp invokes each `command_line` in order, passing the results as a JSON file on stdin, and interprets the stdout as a YAML-formatted modified set of results.
+Filter scripts can emit either YAML (which might be more readable while developing the script), or JSON (which benchcomp will parse as a subset of YAML).
 
 
 ## Built-in visualizations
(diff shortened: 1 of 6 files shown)


docs/src/benchcomp-conf.md
```diff
@@ -24,6 +24,24 @@ variants:
...
+
+```yaml
+filters:
+    - command_line: ./scripts/remove-redundant-results.py
+    - command_line: cat
+```
+
+The value of `filters` is a list of dicts.
+Currently the only legal key for each of the dicts is `command_line`.
+Benchcomp invokes each `command_line` in order, passing the results as a JSON file on stdin, and interprets the stdout as a YAML-formatted modified set of results.
```
zhassan-aws, line 42: Why are the input and output formats different (JSON and YAML)? Does this mean that any specified filter command will need to translate from JSON and YAML (in addition to performing some filtering)?

karkhaz: > Why are the input and output formats different (JSON and YAML)?

I don't want to require anybody to parse YAML, especially since in Python this would require a pip installation where one previously might not have been required. So the input format should be JSON IMO. On the other hand, when writing one of these filter scripts, it can be nice to have an easier-to-read format when debugging. So I think it's nice for `benchcomp filter` to continue to work even if the filter script emits YAML.

>  Does this mean that any specified filter command will need to translate from JSON and YAML

I just pushed a commit adding the following to the documentation:

```
Filter scripts can emit either YAML (which might be more readable while developing the script), or JSON (which benchcomp will parse as a subset of YAML).
```

zhassan-aws (approved):

PR #3114, opened 2024-03-27
Use `intrinsic_name` to get the intrinsic name

adpaco-aws:
Let's close the loop and use `intrinsic_name`, the StableMIR API we added in https://github.com/rust-lang/rust/pull/122203, to retrieve the intrinsic name and avoid post-processing it.


celinval (approved):
Awesome! Thanks @adpaco-aws

PR #3116, opened 2024-03-29
Upgrade Rust toolchain to nightly-2024-03-29

feliperodri:
Resolves https://github.com/model-checking/kani/issues/3115.


feliperodri:
There are two test cases failing for lack of support for `typed_swap`. Here are the test cases:
```
failures:
    [kani] kani/FunctionAbstractions/mem_swap.rs
    [kani] kani/Vectors/any/sorting.rs
```
Here is the error message:
```
Failed Checks: typed_swap is not currently supported by Kani. Please post your example at https://github.com/model-checking/kani/issues/new/choose
 File: "/home/ubuntu/.rustup/toolchains/nightly-2024-03-29-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/mod.rs", line 731, in std::mem::swap::<u32>
```

@celinval should we open a GitHub issue to track the support for `typed_swap` and address it on a separate PR, or should we implement it on this one?

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
```diff
@@ -39,6 +39,7 @@ use stable_mir::mir::mono::Instance;
 use stable_mir::mir::Body;
 use stable_mir::ty::Allocation;
 
+#[allow(dead_code)]
```
zhassan-aws, line 42: Why was this added? It might be better to address the cause of the warning (e.g. delete unused code) rather than suppress the warning.

kani-compiler/src/kani_middle/mod.rs
```diff
@@ -224,6 +223,7 @@ pub fn dump_mir_items(tcx: TyCtxt, items: &[MonoItem], output: &Path) {
 /// Structure that represents the source location of a definition.
 /// TODO: Use `InternedString` once we move it out of the cprover_bindings.
 /// <https://github.com/model-checking/kani/issues/2435>
+#[allow(dead_code)]
```
zhassan-aws, line 226: Same here

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
feliperodri: Removed all dead code as suggested.

kani-compiler/src/kani_middle/mod.rs
feliperodri: Same as above.

kani-compiler/src/kani_middle/transform/body.rs
```diff
@@ -231,7 +231,7 @@ pub enum CheckType {
     /// This is used by default when the `kani` crate is available.
     Assert(Instance),
     /// When the `kani` crate is not available, we have to model the check as an `if { panic!() }`.
-    Panic(Instance),
+    Panic(),
```
zhassan-aws, line 234:
```suggestion
    Panic,
```

zhassan-aws (commented):
Thanks. Regarding `typed_swap`, I suspect that it'll occur in real code, so I would prefer if we handle it in this PR.

zhassan-aws:
For example, the `s2n-quic-core` regression is failing (likely due to `typed_swap`): https://github.com/model-checking/kani/actions/runs/8484198509/job/23246699235?pr=3116#step:5:81

celinval:
I agree with @zhassan-aws  that this will need to be addressed in this PR.

tools/bookrunner/src/litani.rs
```diff
@@ -16,6 +16,7 @@ use std::process::{Child, Command};
 /// attributes in such files, but it may require you to
 /// extend it if advanced features are used (e.g., pools)
 #[derive(Debug, Deserialize)]
+#[allow(dead_code)]
```
feliperodri, line 19: Following the previous Zyad's advice, why do we need to keep this code around with `dead_code` attribute? Maybe @adpaco-aws @jaisnan could tell us whether or not we should just delete it.

tautschnig: I will create a separate PR to completely delete bookrunner.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1943,6 +1944,47 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    /// Swaps the memory contents pointed to by arguments `x` and `y`, respectively, which is
+    /// required for the `typed_swap` intrinsic.
+    ///
+    /// The standard library API requires that `x` and `y` are readable and writable as their
+    /// (common) type (which auto-generated checks for dereferencing will take care of), and the
+    /// memory regions pointed to must be non-overlapping.
+    pub fn codegen_swap(&mut self, fargs: Vec<Expr>, loc: Location) -> Stmt {
+        // if(same_object(x, y)) {
+        //   assert(x + 1 <= y || y + 1 <= x);
+        //   assume(x + 1 <= y || y + 1 <= x);
```
adpaco-aws, line 1957: Is this how the check is implemented in CBMC? I don't know if we want to re-implement the check because there are a bunch of corner cases we could be missing. For example, will this properly handle ZSTs? What about aliasing?

Because of that, I'm wondering if it's better to replace the assignments by calls to `memcpy` which already include those checks IIRC.

kani-compiler/src/kani_middle/mod.rs
```diff
@@ -127,20 +127,16 @@ pub fn dump_mir_items(tcx: TyCtxt, items: &[MonoItem], output: &Path) {
 pub struct SourceLocation {
     pub filename: String,
     pub start_line: usize,
-    pub start_col: usize,
     pub end_line: usize,
-    pub end_col: usize,
```
adpaco-aws, line 132: Assuming you're getting warnings from the linter here, do they still happen if you remove the `pub` keyword?

tools/bookrunner/src/litani.rs
adpaco-aws: Agreed, the `#[allow(dead_code)]` annotations are OK here for now.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1943,6 +1944,47 @@ impl<'tcx> GotocCtx<'tcx> {
...
         cast_ptr.rem(align).eq(zero)
     }
+
+    /// Swaps the memory contents pointed to by arguments `x` and `y`, respectively, which is
+    /// required for the `typed_swap` intrinsic.
+    ///
+    /// The standard library API requires that `x` and `y` are readable and writable as their
+    /// (common) type (which auto-generated checks for dereferencing will take care of), and the
+    /// memory regions pointed to must be non-overlapping.
+    pub fn codegen_swap(&mut self, fargs: Vec<Expr>, loc: Location) -> Stmt {
```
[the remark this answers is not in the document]
tautschnig: kani/Vectors/any/sorting.rs (among others) will exercise this. Are you implying some extra testing is needed?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1943,6 +1944,47 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        // if(same_object(x, y)) {
+        //   assert(x + 1 <= y || y + 1 <= x);
+        //   assume(x + 1 <= y || y + 1 <= x);
+        // }
+        let one = Expr::int_constant(1, Type::c_int());
+        let non_overlapping = fargs[0]
```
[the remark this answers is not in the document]
tautschnig: Why?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
tautschnig: 1) Yes, CBMC's `memcpy` uses the same checks.
2) Aliasing is handled via the `same_object` condition.
3) Using `memcpy` would introduce overhead as we have type information here and don't need to go for byte-by-byte copying. No need to compute type sizes (in bytes), which in turn may raise questions about possible arithmetic overflow.

kani-compiler/src/kani_middle/mod.rs
tautschnig: Why would removing `pub` be the right fix when these are unused members? (Removal was done in response to https://github.com/model-checking/kani/pull/3116#discussion_r1544682573.)

adpaco-aws: I'm trying to understand what's the problem first. Does `clippy` throw an error because you don't read the value ever? Or because you don't read the value AND it's public?

The removal of these fields in `struct`s that derive `Serialize` or `Deserialize` would cause (de)serialization to stop working altogether. That's not the case here, but it's what prevents you from removing the attributes in `LitaniRun`, for example. Moreover, both you and I know that a source location contains columns, and removing it here might be confusing for other developers who need to use them later.

I think this warning could be silenced by either removing `pub` or adding an underscore `_` at the beginning of the field names. Again, I don't even know what the warning is, I just want us to be reasonable when it comes to `clippy` warnings and such.

adpaco-aws (approved):

zhassan-aws (approved):
LGTM. Thanks!

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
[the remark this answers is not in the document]
tautschnig: Thanks, test added.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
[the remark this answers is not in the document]
tautschnig: Thank you for clarifying. Now done.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1943,6 +1944,47 @@ impl<'tcx> GotocCtx<'tcx> {
...
+    ///
+    /// The standard library API requires that `x` and `y` are readable and writable as their
+    /// (common) type (which auto-generated checks for dereferencing will take care of), and the
+    /// memory regions pointed to must be non-overlapping.
+    pub fn codegen_swap(&mut self, fargs: Vec<Expr>, loc: Location) -> Stmt {
+        // if(same_object(x, y)) {
+        //   assert(x + 1 <= y || y + 1 <= x);
+        //   assume(x + 1 <= y || y + 1 <= x);
+        // }
+        let one = Expr::int_constant(1, Type::c_int());
```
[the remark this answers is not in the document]
tautschnig: Done.

kani-compiler/src/kani_middle/mod.rs
tautschnig: `rustc` yields a warning, and `clippy` makes that (in our configuration) an error. I have now added `#[allow(dead_code)]` for just those two fields and also added a comment.

PR #3119, opened 2024-04-02
Adopt Rust's source-based code coverage instrumentation

adpaco-aws:
This PR replaces the line-based coverage instrumentation we introduced in #2609 with the standard source-based code coverage instrumentation performed by the Rust compiler.

As a result, we now insert code coverage checks in the `StatementKind::Coverage(..)` statements produced by the Rust compiler during compilation. These checks include coverage-relevant information[^note-internal] such as the coverage counter/expression they represent [^note-instrument]. Both the coverage metadata (`kanimap`) and coverage results (`kaniraw`) are saved into files after the verification stage.

Unfortunately, we currently have a chicken-egg problem with this PR and #3121, where we introduce a tool named `kani-cov` to postprocess coverage results. As explained in #3143, `kani-cov` is expected to be an alias for the `cov` subcommand and provide most of the postprocessing features for coverage-related purposes. But, the tool will likely be introduced after this change. Therefore, we propose to temporarily print a list of the regions in each function with their associated coverage status (i.e., `COVERED` or `UNCOVERED`).

### Source-based code coverage: An example

The main advantage of source-based coverage results is their precision with respect to the source code. The [Source-based Code Coverage](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html) documentation explains more details about the LLVM coverage workflow and its different options.

For example, let's take this Rust code:
```rust
1 fn _other_function() {
2    println!("Hello, world!");
3 }
4
5 fn test_cov(val: u32) -> bool {
6     if val < 3 || val == 42 {
7         true
8     } else {
9         false
10    }
11 }
12
13 #[cfg_attr(kani, kani::proof)]
14 fn main() {
15    let test1 = test_cov(1);
16    let test2 = test_cov(2);
17    assert!(test1);
18    assert!(test2);
19 }
```

Compiling and running the program with `rustc` and the `-C instrument-coverage` flag, and using the LLVM tools can get us the following coverage result:

Image

In contrast, the `cargo kani --coverage -Zsource-coverage` command currently generates:

```
src/main.rs (main)
 * 14:1 - 19:2 COVERED

src/main.rs (test_cov)
 * 5:1 - 6:15 COVERED
 * 6:19 - 6:28 UNCOVERED
 * 7:9 - 7:13 COVERED
 * 9:9 - 9:14 UNCOVERED
 * 11:1 - 11:2 COVERED
```

which is a verification-based coverage result almost equivalent to the runtime coverage results.

### Benchmarking

We have evaluated the performance impact of the instrumentation using the `kani-perf.sh` suite (14 benchmarks). For each test, we compare the average time to run standard verification against the average time to run verification with the source-based code coverage feature enabled[^note-line-evaluation].

The evaluation has been performed on an EC2 `m5a.4xlarge` instance running Ubuntu 22.04. The experimental data has been obtained by running the `kani-perf.sh` script 10 times for each version (`only verification` and `verification + coverage`), computing the average and standard deviation. We've split this data into `small` (tests taking 60s or less) and `large` (tests taking more than 60s) and drawn the two graphs below.

#### Performance comparison - `small` benchmarks

performance_comparison_small

#### Performance comparison - `large` benchmarks

performance_comparison_large

#### Comments on performance

Looking at the small tests, the performance impact seems negligible in such cases. The difference is more noticeable in the large tests, where the time to run verification and coverage can take 2x or even more. It wouldn't be surprising that, as programs become larger, the complexity of the coverage checking grows exponentially as well. However, since most verification jobs don't take longer than 30min (1800s), it's OK to say that coverage checking represents a 100-200% slowdown in the worst case w.r.t. standard verification.

It's also worth noting a few other things:
 * The standard deviation remains similar in most cases, meaning that the coverage feature doesn't have an impact on their stability.
 * We haven't tried any SAT solvers other than the ones used by default for each benchmark. It's possible that other solvers perform better/worse with the coverage feature enabled.

### Call-outs
 * The soundness issue documented in #3441.
 * The issue with saving coverage mappings for non-reachable functions documented in #3445.
 * I've modified the test cases in `tests/coverage/` to test this feature. Since this technique is simpler, we don't need that many test cases. However, it's possible I've left some test cases which don't contribute much. Please let me know if you want to add/remove a test case.

[^note-internal]: The coverage mappings can't be accessed through the StableMIR interface so we retrieve them through the internal API.

[^note-instrument]: The instrumentation replaces certain counters with expressions based on other counters when possible to avoid a part of the runtime overhead. More details can be found [here](https://github.com/rust-lang/rustc-dev-guide/blob/master/src/llvm-coverage-instrumentation.md#mir-pass-instrumentcoverage). Unfortunately, we can't avoid instrumenting expressions at the moment.

[^note-line-evaluation]: We have not compared performance against the line-based code coverage feature because it doesn't seem worth it. The line-based coverage feature is guaranteed to include more coverage checks than the source-based one for any function. In addition, source-based results are more precise than line-based ones. So this change represents both a quantitative and qualitative improvement.


tests/coverage/reachable/bounds/reachable_fail/test.rs
```diff
@@ -2,10 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
```
zhassan-aws, line 3: Is the absence of anything uncovered due to https://github.com/model-checking/kani/issues/3441? If so, please add a comment that references it.

tests/coverage/reachable/div-zero/reachable_fail/test.rs
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
```
zhassan-aws, line 3: Same here.

tests/coverage/unreachable/assume_assert/main.rs
```diff
@@ -1,5 +1,11 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This test should check that the region after `kani::assume(false)` is
+//! `UNCOVERED`. However, due to a technical limitation in `rustc`'s coverage
+//! instrumentation, only one `COVERED` region is reported for the whole
```
zhassan-aws, line 6: Bummer :(

tests/coverage/unreachable/debug-assert/expected
```diff
@@ -1,4 +1,10 @@
...
-coverage/unreachable/debug-assert/main.rs, 8, NONE
-coverage/unreachable/debug-assert/main.rs, 10, NONE
+Source-based code coverage results:
+
+tests/coverage/unreachable/debug-assert/main.rs (main)
+ * 10:1 - 10:11 COVERED\
+ * 11:9 - 11:10 COVERED\
+ * 11:14 - 11:18 COVERED\
+ * 12:30 - 12:71 UNCOVERED\
+ * 13:9 - 13:23 UNCOVERED\
```
zhassan-aws, line 8: I'm surprised this one works as expected despite #3441.

tests/coverage/unreachable/variant/main.rs
```diff
@@ -12,6 +12,7 @@ enum Direction {
 }
 
 fn print_direction(dir: Direction) {
+    // For some reason, `dir`'s span is reported as `UNCOVERED` too
```
zhassan-aws, line 15: This is weird. Any insights why the span is reported as uncovered?

tools/compiletest/src/runtest.rs
```diff
@@ -272,7 +272,9 @@ impl<'test> TestCx<'test> {
             .arg("kani")
             .arg("--target-dir")
             .arg(self.output_base_dir().join("target"))
-            .current_dir(parent_dir);
+            .current_dir(parent_dir)
+            .arg("--coverage")
```
zhassan-aws, line 276: Isn't this turning on coverage for all `cargo kani` tests? Is this intentional?

kani-driver/Cargo.toml
```diff
@@ -15,6 +15,7 @@ publish = false
 kani_metadata = { path = "../kani_metadata" }
 cargo_metadata = "0.18.0"
 anyhow = "1"
+chrono = "0.4.38"
```
zhassan-aws, line 18: This crate seems to bring in tons of other dependencies as indicated by the `Cargo.lock` diff. Since all that is needed is to get the current time and print it, it the [`time` crate](https://crates.io/crates/time/0.3.36/dependencies) seems to be a lighter-weight version.

kani-driver/src/call_single_file.rs
```diff
@@ -155,6 +155,13 @@ impl KaniSession {
     pub fn kani_rustc_flags(&self, lib_config: LibConfig) -> Vec<OsString> {
         let mut flags: Vec<_> = base_rustc_flags(lib_config);
         // We only use panic abort strategy for verification since we cannot handle unwind logic.
+        if self.args.coverage {
+            flags.extend_from_slice(
+                // &["-C", "instrument-coverage", "-Z", "no-profiler-runtime", "--emit=mir"].map(OsString::from),
```
zhassan-aws, line 160: Please remove comment.

kani-driver/src/call_single_file.rs
```diff
@@ -155,6 +155,13 @@ impl KaniSession {
...
+                &["-C", "instrument-coverage", "-Z", "no-profiler-runtime", "-C", "opt-level=0"]
```
zhassan-aws, line 161: Is `-C opt-level=0` necessary?

zhassan-aws (commented):
Nice!

First round of comments.

adpaco-aws:
Thanks for you review, @zhassan-aws ! I'll address your comments soon. Also, please note that I have updated the benchmarking section in the PR description. I'm hoping this makes clear the performance impact of running coverage checking with this technique.

kani-driver/src/call_single_file.rs
adpaco-aws: No, it's not - forgot to remove it. Thanks for pointing this out!

kani-driver/src/call_single_file.rs
adpaco-aws: Done.

tools/compiletest/src/runtest.rs
adpaco-aws: This was intended as a temporary temporary

kani-driver/Cargo.toml
adpaco-aws: Great suggestion! We're now using the `time` crate with the `formatting` feature.

tests/coverage/if-statement-multi/test.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: --coverage -Zsource-coverage
+
+//! Checks that we print a line which points the user to the path where coverage
```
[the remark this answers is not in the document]
adpaco-aws: Updated the comment.

tests/coverage/reachable/assert/reachable_pass/expected
```diff
@@ -1,4 +1,9 @@
-coverage/reachable/assert/reachable_pass/test.rs, 6, FULL
-coverage/reachable/assert/reachable_pass/test.rs, 7, PARTIAL
-coverage/reachable/assert/reachable_pass/test.rs, 8, FULL
-coverage/reachable/assert/reachable_pass/test.rs, 10, FULL
+Source-based code coverage results:
+
+test.rs (main)\
+ * 5:1 - 6:34 COVERED\
+ * 7:14 - 9:6 COVERED\
+ * 9:6 - 9:7 UNCOVERED
```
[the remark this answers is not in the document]
adpaco-aws: Good catch. I didn't realize this, so I've documented the issue here: #3455

tests/coverage/reachable/bounds/reachable_fail/test.rs
adpaco-aws: Yes, the "return" should be `UNCOVERED` but there's only one region. Added a comment referencing the issue.

tests/coverage/reachable/div-zero/reachable_fail/test.rs
adpaco-aws: This one I actually removed because didn't seem to add anything new.

tests/coverage/unreachable/assume_assert/main.rs
adpaco-aws: Yeah, this was an unpleasant surprise for me too :(

tests/coverage/unreachable/debug-assert/expected
adpaco-aws: Indeed, it's quite surprising. I suppose it's due to the macros expanding to more elaborate code behind the scenes, and their spans being correctly propagated.

tests/coverage/unreachable/variant/main.rs
adpaco-aws: No insights, unfortunately. I'm going to ask around because it's really weird. For the time being, I've created #3456 and pasted it into the test.

adpaco-aws:
Ready for review again - I've taken the chance to reorganize the `tests/coverage` folder so that it's not so nested. There's even a `known_issue` folder that contains the tests with known issues. I've removed some more tests that seemed redundant to me too. Hope that makes it easier to review.

zhassan-aws (commented):
Looks great! I haven't looked at the CBMC result parsing part yet.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -216,3 +216,69 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    return Some(mapping.code_region.clone());
+                }
+            }
+        }
+        None
+    }
+
+    fn parse_coverage_opaque(coverage_opaque: &Opaque) -> Option<CovTerm> {
+        let coverage_str = coverage_opaque.to_string();
+        if coverage_str == "Zero" {
```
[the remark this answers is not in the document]
adpaco-aws: Yes, I've added a reference to the [`CovTerm` enum documentation](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/coverage/enum.CovTerm.html) which determines the strings. This appears in the RFC but it's a good idea to include here too.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -216,3 +216,69 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        } else if let Some(rest) = coverage_str.strip_prefix("CounterIncrement(") {
+            let (num_str, _rest) = rest.split_once(')').unwrap();
+            let num = num_str.parse::<u32>().unwrap();
+            Some(CovTerm::Counter(num.into()))
+        } else if let Some(rest) = coverage_str.strip_prefix("ExpressionUsed(") {
+            let (num_str, _rest) = rest.split_once(')').unwrap();
+            let num = num_str.parse::<u32>().unwrap();
+            Some(CovTerm::Expression(num.into()))
+        } else {
+            None
```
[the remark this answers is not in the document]
adpaco-aws: This branch is not reachable at the moment, but I was trying to defend against future changes in the `CovTerm` enum. I've changed it now to return just `CovTerm` since it hasn't changed at all for quite some time, but if you notice this fails often in the future I'd suggest changing it to return a `Result<CovTerm>` then.

kani-driver/src/args/coverage_args.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+use crate::args::CommonArgs;
```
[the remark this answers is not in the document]
adpaco-aws: Thanks for this comment because... this module should've been removed 😳 

In a previous version, the coverage feature was enabled through a subcommand, which was defined/implemented in this module. However, thanks to the feedback in #3143, we decided to keep the subcommand as an alias for `kani-cov` (#3121) and enable coverage through the `--coverage` option.

kani-driver/src/coverage/cov_session.rs
```diff
@@ -0,0 +1,176 @@
...
+        let metadata = self.cargo_metadata(build_target)?;
+        let target_dir = self
+            .args
+            .target_dir
+            .as_ref()
+            .unwrap_or(&metadata.target_directory.clone().into())
+            .clone()
+            .join("kani");
+
+        let outdir = target_dir.join(build_target).join(format!("kanicov_{stamp}"));
```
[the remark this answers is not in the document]
adpaco-aws: Sure! The purpose is to make coverage results easily identifiable.

We could also use UUIDs for this purpose, but UUIDs are better suited for scenarios where you generate many runs and you just want them to be different. Here, in addition to being (slightly) different, we're often interested in the most recent results, especially when we're proof-debugging. Using a timestamp makes it easy to find them, and also helps with cleaning up older results which are not interesting to us anymore. This is a problem I always have with `benchcomp`/`litani`, which uses UUIDs for run IDs, forcing me to look at logs or `ls -l` to know what's the ID associated with the most recent run.

BTW I assumed you wanted the clarification here, but let me know if you want me to add this as a comment to the timestamp generation.

kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -216,3 +216,69 @@ impl<'tcx> GotocCtx<'tcx> {
         self.reset_current_fn();
     }
 }
+
+pub mod rustc_smir {
```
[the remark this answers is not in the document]
adpaco-aws: Yeah, it would be ideal to have these types and APIs in StableMIR. Until now, I've refrained from opening an issue in the StableMIR repo because the coverage APIs cannot be considered stable by any means, and I suspect no stabilization will happen until branch coverage has been implemented/tested. But I'm open to create the issue now and refer to it here if you think it's appropriate.

kani-driver/src/coverage/cov_session.rs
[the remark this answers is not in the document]
zhassan-aws: A comment would help.

kani-driver/src/call_cbmc.rs
```diff
@@ -404,6 +425,60 @@ fn determine_failed_properties(properties: &[Property]) -> FailedProperties {
...
+
+    if cov_properties.is_empty() {
+        return None;
+    }
+
+    let counter_re = {
+        static COUNTER_RE: OnceLock<Regex> = OnceLock::new();
+        COUNTER_RE.get_or_init(|| {
+            Regex::new(
+                r#"^(?<kind>CounterIncrement|ExpressionUsed)\((?<counter_num>[0-9]+)\) \$(?<func_name>[^\$]+)\$ - (?<span>.+)"#,
```
zhassan-aws, line 440: Can we add an example pattern this regex is matching against?

kani-driver/src/cbmc_property_renderer.rs
```diff
@@ -431,72 +422,31 @@ pub fn format_result(
     result_str
 }
 
-/// Separate checks into coverage and non-coverage based on property class and format them separately for --coverage. We report both verification and processed coverage
-/// results
+/// Separate checks into coverage and non-coverage based on property class and
+/// format them separately for `--coverage``. Then we report both verification
```
zhassan-aws, line 426:
```suggestion
/// format them separately for `--coverage`. Then we report both verification
```

zhassan-aws (approved):
A few nits. Thanks @adpaco-aws!

kani-driver/src/cbmc_property_renderer.rs
adpaco-aws: Added manually.

kani-driver/src/call_cbmc.rs
```diff
@@ -312,23 +325,31 @@ impl VerificationResult {
...
+            coverage_results: None,
         }
     }
 
     pub fn render(
         &self,
         output_format: &OutputFormat,
         should_panic: bool,
-        coverage_mode: bool,
+        _coverage_mode: bool,
```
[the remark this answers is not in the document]
adpaco-aws: Yes, removed.

kani-driver/src/call_cbmc.rs
adpaco-aws: Done. I've also linked to #3445 in the comment.

kani-driver/src/coverage/cov_session.rs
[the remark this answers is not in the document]
adpaco-aws: Added a comment for clarification.

adpaco-aws:
Thank you, @zhassan-aws ! I will hold the merge of this PR until #3143 is merged.

Also, the other we talked about this feature breaking the coverage feature in the VS Code extension. I recall agreeing to keeping the line-based coverage option but, on a second thought, I think it'd just be easier to remove it here and integrate the VS Code extension with the source-based feature. I'm meeting @jaisnan soon to make sure that's the case and coordinate on the integration.

feliperodri (approved):

PR #3121, opened 2024-04-03
`kani-cov`: A coverage tool for Kani

adpaco-aws:
This PR introduces `kani-cov`, a coverage-oriented tool for Kani. This new tool will help users aggregate raw coverage results produced by Kani and generate human-readable coverage summaries and reports.

`kani-cov` allows using three different subcommands: `merge`, `summary` and `report`. In the following, we will show how each one of these commands can be used.

### The `merge` subcommand

The `merge` subcommand takes raw coverage results from Kani (AKA "kaniraw" files) and generates aggregated coverage results which are stored into another file (AKA a "kanicov" file). This new file can then be used for generating summaries or reports.

For example, if we assume this Rust code is in `main.rs`:

```rust
1 fn _other_function() {
2    println!("Hello, world!");
3 }
4
5 fn test_cov(val: u32) -> bool {
6     if val < 3 || val == 42 {
7         true
8     } else {
9         false
10    }
11 }
12
13 #[cfg_attr(kani, kani::proof)]
14 fn main() {
15    let test1 = test_cov(1);
16    let test2 = test_cov(2);
17    assert!(test1);
18    assert!(test2);
19 }
```
and we execute Kani with coverage enabled
```
kani main.rs --coverage -Zsource-coverage
```
the raw coverage results will be saved to a folder
```
[info] Coverage results saved to /home/ubuntu/coverage-experiments/src/kanicov_2024-09-23_23-49
```
we now can aggregate those results with the `merge` subcommand
```
kani-cov merge kanicov_2024-09-23_23-49/*kaniraw.json
```
which by default produces a `default_kanicov.json` file. We can also use the `--output <path>` option to change the name of the file.

The type of file we just produced is known as the "profile" or the "kanicov" file, and contains aggregated coverage results which may have been originated in multiple verification sessions. This file is a required input for the other two subcommands.

### The `summary` subcommand

Now that we have both the "kanicov" file and the "kanimap" file, we are ready to produce coverage metrics with the `summary` subcommand:

```
kani-cov summary kanicov_2024-09-23_23-49/kanicov_2024-09-23_23-49_kanimap.json --profile default_kanicov.json
```

This outputs a table including coverage metrics for functions, lines and regions for each file referenced in the coverage mappings:

```
| Filename                                      | Function (%) | Line (%)     | Region (%)  |
| --------------------------------------------- | ------------ | ------------ | ----------- |
| /home/ubuntu/coverage-experiments/src/main.rs | 2/3 (66.67)  | 8/12 (66.67) | 4/6 (66.67) |
```

Since the default format is markdown, the table can also be rendered as in this PR description:

| Filename                                      | Function (%) | Line (%)     | Region (%)  |
| --------------------------------------------- | ------------ | ------------ | ----------- |
| /home/ubuntu/coverage-experiments/src/main.rs | 2/3 (66.67)  | 8/12 (66.67) | 4/6 (66.67) |

At present, no other formats are available. But the JSON and CSV formats would be really nice and easy to have.

### The `report` subcommand

Coverage reports go a little further and allow users to visualize coverage information on the source code.

The `report` requires the same input files as the `summary` subcommand:

```
kani-cov report kanicov_2024-09-23_23-49/kanicov_2024-09-23_23-49_kanimap.json --profile default_kanicov.json
```

When calling this command from a terminal, it will render like this:

Screenshot 2024-10-04 at 5 49 51 PM

However, if you're redirecting the output to a file, it will back off to another text-based mode which simply uses ` ``` ` and ` ''' ` to represent opening and closing escapes for the coverage regions. This is just another format called `escapes` which can be explicitly requested as

```
kani-cov report kanicov_2024-09-23_23-49/kanicov_2024-09-23_23-49_kanimap.json --profile default_kanicov.json --format=escapes
```

Similar to the `summary` command, this will iterate over the files in the coverage mappings, compute coverage information and highlight any uncovered regions (except #3543).

### Testing

This PR also changes `compiletest` to automatically call the `merge` and `report` subcommands on the `coverage` test suite. The tests have been blessed using the `--fix-expected` option and this Python3 script that removes the first line (path to the file) and adds `\` after each line so an exact match is expected.

```
import os

def process_files(directory):
    for filename in os.listdir(directory):
        if filename == "expected":
            file_path = os.path.join(directory, filename)
            with open(file_path, "r") as file:
                lines = file.readlines()
            
            # Remove the first line
            lines = lines[1:]
            
            # Add the character `\` after every line
            processed_lines = []
            for i, line in enumerate(lines):
                processed_lines.append(line.replace("\n", "\\\n"))
            
            # Write the processed lines back to the file
            with open(file_path, "w") as file:
                file.writelines(processed_lines)
```

### Callouts / Known issues

This PR only introduces `kani-cov`. It doesn't plug `kani-cov` into `kani-driver` so it's automatically run when using the `--coverage` option, nor it prepares the `kani-cov` binary for distribution. This is better to address later.

This PR also spawns a few issues:
 - #3543 forces us to filter out certain coverage results to avoid issues with the report command. It is important that we investigate where these coverage results come from and their semantics as this represents a soundness issue.
 - #3541 is to avoid duplicating certain data structures that currently live in `kani-driver`. We should probably have a different crate for these structures and import them from `kani-driver` and `kani-cov`, but this is more of a distribution problem that doesn't need to be tackled here.
 - #3542 requires us to match filenames when extracting function coverage results. It's not a big deal, but it'll likely improve the tool's robustness in the long term.


celinval (commented):
Are we planning to release this tool or is it only for developers?

tools/kanicov/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "visualize_coverage"
```
celinval, line 2: why not `kani-cov`?

adpaco-aws:
> Are we planning to release this tool or is it only for developers?

The plan is to release the tool. More details in the upcoming RFC.

Cargo.toml
```diff
@@ -41,6 +41,7 @@ members = [
   "tools/bookrunner",
   "tools/compiletest",
   "tools/build-kani",
+  "tools/kanicov",
```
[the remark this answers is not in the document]
adpaco-aws: Thanks for the suggestion! Yes, let's use `kani-cov` instead.

adpaco-aws:
Note to self: Consider highlighting uncovered spans in red as the default behavior. At the very least, discuss in the RFC.

adpaco-aws:
Just marked this as ready for review!

tools/kani-cov/src/args.rs
```diff
@@ -0,0 +1,111 @@
...
+pub enum ReportFormat {
+    Terminal,
+    Escapes,
+}
+
+/// Validate general arguments and delegate validation of command-specific
+/// arguments.
+pub fn validate_args(args: &Args) -> Result<()> {
+    if args.command.is_none() {
+        bail!("subcommand needs to be specified")
```
feliperodri, line 101: Nitpick: I'd include here the command options available.

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Basically, this aggregates the information of one or more `CoverageCheck`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CovResult {
+    pub function: String,
+    pub region: CoverageRegion,
+    pub times_covered: u32,
+    pub total_times: u32,
+}
+
+/// A coverage region.
```
feliperodri, line 88: I'd include
```suggestion
/// A coverage region.
/// `start` and `end` are tuples containing the line and column numbers.
```

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct CoverageResults {
+    pub data: BTreeMap<String, Vec<CoverageCheck>>,
+}
+
+/// Aggregated coverage results.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct CombinedCoverageResults {
+    pub data: BTreeMap<String, Vec<(String, Vec<CovResult>)>>,
+}
```
feliperodri, line 76: Wouldn't be better to use a `HashMap` here? Does the order matter? I also imagine this data could be quite large.

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+#[derive(Debug, Clone, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct CoverageRegion {
+    pub file: String,
+    pub start: (u32, u32),
+    pub end: (u32, u32),
+}
+
+impl Display for CoverageRegion {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}:{} - {}:{}", self.start.0, self.start.1, self.end.0, self.end.1)
```
feliperodri, line 101: Maybe let's include the file name as well.
```suggestion
        write!(f, "{}:{}:{} - {}:{}", self.file, self.start.0, self.start.1, self.end.0, self.end.1)
```

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    let mut cursor = tree.walk();
+    let first_child_exists = cursor.goto_first_child();
+
+    if !first_child_exists {
+        return vec![];
+    }
+
+    let mut function_info: Vec<FunctionInfo> = Vec::new();
+
```
feliperodri, line 159: We could also declare here:
```rust
let source_bytes = source_code.as_bytes();
```
and avoid calling it multiple times.

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+    if cursor.node().kind() == "function_item" {
+        function_info.push(function_info_from_node(cursor.node(), source_code.as_bytes()))
+    };
+
+    while cursor.goto_next_sibling() {
+        if cursor.node().kind() == "function_item" {
+            function_info.push(function_info_from_node(cursor.node(), source_code.as_bytes()))
+        }
+    }
```
feliperodri, line 168: Why not use an iterator?
```suggestion
for node in tree.root_node().children(&mut cursor) {
    if node.kind() == "function_item" {
        function_info.push(function_info_from_node(node, source_bytes));
    }
}
```

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+    parser.set_language(&tree_sitter_rust::language()).expect("Error loading Rust grammar");
+
+    let tree = parser.parse(&source_code, None).expect("Failed to parse file");
+
+    let mut cursor = tree.walk();
+    let first_child_exists = cursor.goto_first_child();
+
+    if !first_child_exists {
+        return vec![];
+    }
```
feliperodri, line 156:
```suggestion
    if !cursor.goto_first_child() {
        return vec![];
    }
```

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Extract the coverage results associated to a function
+pub fn function_coverage_results(
+    info: &FunctionInfo,
+    file: &Path,
+    results: &CombinedCoverageResults,
+) -> Option<(String, Vec<CovResult>)> {
+    // The filenames in "kaniraw" files are not absolute, so we need to match
+    // them with the ones we have in the aggregated results (i.e., the filenames
+    // in the "kanimap" files).
+    let filename = file.to_path_buf().into_os_string().into_string().unwrap();
```
feliperodri, line 194: Wouldn't this also works? It's also safer since it returns valid UTF-8. 
```suggestion
    let filename = file.to_str()?;
```

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+    let right_filename = results.data.keys().find(|p| filename.ends_with(*p)).unwrap();
+    // TODO: The filenames in kaniraw files should be absolute, just like in metadata.
+    // Otherwise the key for `results` just fails... <https://github.com/model-checking/kani/issues/3542>
+    let file_results = results.data.get(right_filename).unwrap();
+    let function = info.name.clone();
```
feliperodri, line 199: Why do you need a clone here? The reference might be fine.

tools/kani-cov/src/merge.rs
```diff
@@ -0,0 +1,150 @@
...
+    Ok(raw_results)
+}
+
+/// Combine raw coverage results into an aggregated form
+fn combine_raw_results(results: &Vec<CoverageResults>) -> CombinedCoverageResults {
+    let all_file_function_names = function_names_from_results(results);
+
+    let mut new_data: BTreeMap<String, Vec<(String, Vec<CovResult>)>> = BTreeMap::new();
+
+    for (file_name, fun_name) in all_file_function_names {
```
feliperodri, line 61: Wouldn't references be better here?
```suggestion
    for (file_name, fun_name) in &all_file_function_names {
```

tools/kani-cov/src/merge.rs
```diff
@@ -0,0 +1,150 @@
...
+            for check in checks {
+                let function = check.function.clone();
+                let file_function = (file.clone(), function);
+                if !file_function_pairs.contains(&file_function) {
+                    file_function_pairs.push(file_function);
+                }
+            }
+        }
+    }
+    file_function_pairs
```
feliperodri, line 149:
```suggestion
    let mut file_function_pairs = HashSet::new();
    for result in results {
        for (file, checks) in &result.data {
            for check in checks {
                file_function_pairs.insert((file.clone(), check.function.clone()));
            }
        }
    }
    file_function_pairs.into_iter().collect()
```

tools/kani-cov/src/report.rs
```diff
@@ -0,0 +1,262 @@
...
+/// ```fn _other_function() {
+///     println!("Hello, world!");
+/// }'''
+///
+pub fn output_coverage_results(
+    format: &ReportFormat,
+    filepath: PathBuf,
+    results: Vec<LineResults>,
+) -> Result<()> {
+    let flattened_results: Vec<(usize, Option<(u32, MarkerInfo)>)> =
```
feliperodri, line 116: Couldn't this be empty?

feliperodri (changes requested):
Thank you so much for pushing this to the finish line. I can't wait to use it, only a few comments.

tools/compiletest/src/runtest.rs
```diff
@@ -441,10 +441,68 @@ impl<'test> TestCx<'test> {
...
+        let kanicov_proc = self.run_kanicov_report(&kanimap, &kaniraws, &kanicov);
         let expected_path = self.testpaths.file.parent().unwrap().join("expected");
-        self.verify_output(&proc_res, &expected_path);
+        self.verify_output(&kanicov_proc, &expected_path);
+    }
+
+    fn run_kanicov_report(
+        &self,
+        kanimap: &PathBuf,
+        kaniraws: &Vec<PathBuf>,
```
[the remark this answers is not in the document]
adpaco-aws: Done.

tools/compiletest/src/runtest.rs
```diff
@@ -441,10 +441,68 @@ impl<'test> TestCx<'test> {
...
+        (kanimap, kaniraws, kanicov)
     }
 
+    fn extract_cov_results_path(&self, proc_res: &ProcRes) -> PathBuf {
+        let output_lines = proc_res.stdout.split('\n').collect::<Vec<&str>>();
+        let coverage_info = output_lines.iter().find(|l| l.contains("Coverage results saved to"));
+        if coverage_info.is_none() {
+            self.fatal_proc_rec("failed to find the path to the coverage results", proc_res);
+        }
+        let coverage_path = coverage_info.unwrap().split(' ').last().unwrap();
```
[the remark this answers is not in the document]
adpaco-aws: Done.

tools/compiletest/src/runtest.rs
```diff
@@ -441,10 +441,68 @@ impl<'test> TestCx<'test> {
...
+        // self.error(&format!("kaniraw_glob: {kaniraw_glob}"));
+        let kaniraws: Vec<PathBuf> = glob::glob(&kaniraw_glob)
+            .expect("Failed to read glob pattern")
+            .filter_map(|entry| entry.ok())
+            .collect();
+
+        (kanimap, kaniraws, kanicov)
     }
 
+    fn extract_cov_results_path(&self, proc_res: &ProcRes) -> PathBuf {
```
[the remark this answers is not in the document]
adpaco-aws: Because we need access to `fatal_proc_rec`, which is used for graceful failures in `compiletest`.

tools/kani-cov/src/args.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use anyhow::{Result, bail};
+use clap::{arg, command};
+
+use crate::{merge, report, summary};
+
+/// We define three subcommands:
+///  * `merge` for merging raw Kani coverage results (AKA "kaniraw" files)
+///  * `summary` for producing a summary containing coverage metrics
+///  * `report` for generating human-readable coverage reports
```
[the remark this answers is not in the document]
adpaco-aws: Since the PR description already includes an example, I've done some modifications and added it here.

tools/kani-cov/src/args.rs
```diff
@@ -0,0 +1,111 @@
...
+    pub command: Option<Subcommand>,
+}
+
+/// Arguments for the `merge` subcommand
+#[derive(Debug, clap::Args)]
+pub struct MergeArgs {
+    #[arg(long)]
+    pub output: Option<PathBuf>,
+    #[arg(required = true)]
+    pub files: Vec<PathBuf>,
```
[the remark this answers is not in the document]
adpaco-aws: I don't know, this seems to be working as intended. Why do you think we should add it to all required arguments?

tools/kani-cov/src/args.rs
```diff
@@ -0,0 +1,111 @@
...
+    // The format of the summary
+    #[arg(long, short, value_parser = clap::value_parser!(SummaryFormat), default_value = "markdown")]
+    pub format: SummaryFormat,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, clap::ValueEnum)]
+pub enum SummaryFormat {
+    Markdown,
+    // Json,
+}
```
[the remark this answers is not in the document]
adpaco-aws: Done!

tools/kani-cov/src/args.rs
adpaco-aws: Done.

tools/kani-cov/src/coverage.rs
adpaco-aws: Done.

tools/kani-cov/src/coverage.rs
adpaco-aws: The order doesn't really matter. Changed to `HashMap`.

tools/kani-cov/src/coverage.rs
adpaco-aws: Ok, done.

tools/kani-cov/src/coverage.rs
adpaco-aws: Yep, done.

tools/kani-cov/src/coverage.rs
adpaco-aws: Thanks! Yes, this looks better.

tools/kani-cov/src/coverage.rs
adpaco-aws: Yes, it works. I've updated the code to use your suggestion, thanks!

tools/kani-cov/src/coverage.rs
adpaco-aws: This code is now gone because we're using the iterator.

tools/kani-cov/src/coverage.rs
adpaco-aws: True, done.

tools/kani-cov/src/merge.rs
adpaco-aws: Updated, thanks!

tools/kani-cov/src/merge.rs
adpaco-aws: We can use references but that forces us to use a bunch of `*` and `clone()` in the code below. I didn't think it looked better than the current solution, but happy to commit it if you think otherwise.

tools/kani-cov/src/report.rs
adpaco-aws: Yes, this could be empty. For that to happen, there must be no functions in the file. But if there's at least one, and even if it's unreachable, `flattened_results` will have been populated with full-line markers.

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub data: HashMap<String, Vec<(String, Vec<CovResult>)>>,
+}
+
+/// The coverage result associated to a particular coverage region.
+/// Basically, this aggregates the information of one or more `CoverageCheck`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CovResult {
+    pub function: String,
+    pub region: CoverageRegion,
+    pub times_covered: u32,
```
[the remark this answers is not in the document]
adpaco-aws: True, and it's probably better to use `usize` anyway. I've replaced `u32` with `usize` in most types/structures, and added more documentation to `CovResult` regarding `total_times` and `times_covered`.

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,201 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This module defines coverage-oriented data structures shared among
+//! subcommands and other utilities like the Rust tree-sitter.
```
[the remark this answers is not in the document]
adpaco-aws: This is a great suggestion, and it's now implemented.

feliperodri (changes requested):
Thanks for implementing all my comments. I have one last before approving it. Why the issue https://github.com/model-checking/kani/issues/3543 is classified as a soundness issue now? As far as I understood, the coverage report feature will crash in the cases you mentioned in the issue. Will it report false coverage reports in any case? This is important to fix before stabilization, but it seems to me unnecessary to classify this as a soundness issue.

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+use serde_derive::{Deserialize, Serialize};
+use std::path::{Path, PathBuf};
+use std::{collections::HashMap, fmt::Display};
+use std::{fmt, fs};
+use tree_sitter::{Node, Parser};
+
+type Function = String;
+type Filename = String;
+type LineNumber = usize;
+
```
carolynzech, line 17:
```suggestion
type ColumnNumber = usize;

```

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+/// The coverage result associated to a particular coverage region.
+///
+/// Basically, this aggregates the information of one or more `CoverageCheck`
+/// for a particular region. Thus, `total_times` represents the total number of
+/// such checks, while `times_covered` keeps track of how many of those checks
+/// had the `COVERED` status.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CovResult {
+    pub function: String,
```
carolynzech, line 90:
```suggestion
    pub function: Function,
```

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+/// A coverage region.
+/// `start` and `end` are tuples containing the line and column numbers.
+///
+/// Note: This data structure should not be duplicated in Kani -
+/// <https://github.com/model-checking/kani/issues/3541>
+#[derive(Debug, Clone, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct CoverageRegion {
+    pub file: String,
```
carolynzech, line 103:
```suggestion
    pub file: Filename,
```

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub start: (usize, usize),
+    pub end: (usize, usize),
```
carolynzech, line 105:
```suggestion
    pub start: (LineNumber, ColumnNumber),
    pub end: (LineNumber, ColumnNumber),
```

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+        CoverageMetric { covered, total }
+    }
+}
+
+/// Function information obtained through a tree-sitter
+#[derive(Debug)]
+pub struct FunctionInfo {
+    pub name: Function,
+    pub start: (usize, usize),
+    pub end: (usize, usize),
```
carolynzech, line 149:
```suggestion
    pub start: (LineNumber, ColumnNumber),
    pub end: (LineNumber, ColumnNumber),
```

tools/kani-cov/src/coverage.rs
```diff
@@ -0,0 +1,209 @@
...
+    let end = (node.end_position().row + 1, node.end_position().column + 1);
+    FunctionInfo { name, start, end }
+}
+
+/// Extract the coverage results associated to a function
+pub fn function_coverage_results(
+    info: &FunctionInfo,
+    file: &Path,
+    results: &CombinedCoverageResults,
+) -> Option<(String, Vec<CovResult>)> {
```
carolynzech, line 190:
```suggestion
) -> Option<(Function, Vec<CovResult>)> {
```

tools/kani-cov/src/summary.rs
```diff
@@ -0,0 +1,301 @@
...
+/// Validate arguments to the `summary` subcommand in addition to clap's
+/// validation.
+pub fn validate_summary_args(_args: &SummaryArgs) -> Result<()> {
+    // No validation is done at the moment
+    Ok(())
+}
+
+/// Computes coverage results from a line-based perspective.
+///
+/// Basically, for each line we produce an `<Option<(u32, MarkerInfo)>>` result
```
carolynzech, line 112:
```suggestion
/// Basically, for each line we produce an `<Option<(usize, MarkerInfo)>>` result
```

carolynzech (approved):
I'd recommend making the type aliases public and incorporating them throughout `merge.rs` and `summary.rs` as well, but not a blocker for this PR.
Excited to use this feature!

adpaco-aws:
> Why the issue #3543 is classified as a soundness issue now? As far as I understood, the coverage report feature will crash in the cases you mentioned in the issue. Will it report false coverage reports in any case? This is important to fix before stabilization, but it seems to me unnecessary to classify this as a soundness issue.

The reason #3543 was classified as a soundness issue is because we were filtering out the results which could lead to a crash in `kani-cov`. Filtering out these results can be considered a soundness issue because, even if they're spurious failures coming from the Rust coverage instrumentation, we could be hiding other issues. It all depends on how we define soundness for coverage and the origin of these coverage artifacts.

Regardless, I've changed how we handle the checks with these out-of-bounds regions to replicate the spurious failure instead. Since we only need to extend the line by one space to represent the check in such cases, we just do that now. In other words, we are now reporting the same spurious failures, which may be more confusing for users but avoids soundness concerns. For example, for the test in `tests/coverage/assert` we get this report:

````
/home/ubuntu/kani/tests/coverage/assert/test.rs
   1|     | // Copyright Kani Contributors
   2|     | // SPDX-License-Identifier: Apache-2.0 OR MIT
   3|     | 
   4|     | #[kani::proof]
   5|    1| fn foo() {
   6|    1|     let x: i32 = kani::any();
   7|    1|     if x > 5 {
   8|     |         // fails
   9|    1|         assert!(x < 4);
  10|    1|         if x < 3 ```{'''
  11|    0| ```            // unreachable'''
  12|    0| ```            assert!(x == 2);'''
  13|    0| ```        }'''``` '''
  14|    1|     } else {
  15|    1|         // passes
  16|    1|         assert!(x <= 5);
  17|    1|     }
  18|     | }
````

Note the single region represented at the end of line 13? In the original source code, that space doesn't exist. But we can create that space in this representation and report the same result that the instrumentation gives us. It's less noticeable in the terminal output, but we also "draw it" there:

Screenshot 2024-10-04 at 5 49 51 PM

In addition to this change, I've added some more documentation to the code handling this special case and updated #3543 (labeled as "spurious failure" now).  Please don't hesitate to ask for more clarifications if you need them. @carolynzech I've also implemented your suggestions to use the type aliases in other modules of `kani-cov`. Thanks!

feliperodri (approved):

Issue #3005, opened 2024-02-07
Restore mdbook-graphviz installation


tautschnig:
Revert the change to `scripts/setup/ubuntu/install_doc_deps.sh` done in #3004 once https://github.com/dylanowen/mdbook-graphviz has released a new version with consistent dependencies (their CI is also failing in https://github.com/dylanowen/mdbook-graphviz/pull/106 and https://github.com/dylanowen/mdbook-graphviz/pull/104).

PR #3123, opened 2024-04-04
Remove bookrunner

tautschnig:
We never updated the books and do not actually require the evaluation data anymore.


.github/workflows/kani.yml
```diff
@@ -96,39 +96,14 @@ jobs:
...
-        run: cat build/output/latest/html/bookrunner.txt
-
-      - name: Print book runner failures grouped by stage
-        run: python3 scripts/ci/bookrunner_failures_by_stage.py build/output/latest/html/index.html
-
-      - name: Detect unexpected book runner failures
-        run: ./scripts/ci/detect_bookrunner_failures.sh build/output/latest/html/bookrunner.txt
-
       - name: Install book dependencies
         run: ./scripts/setup/ubuntu/install_doc_deps.sh
```
adpaco-aws, line 108: You can remove the commented out `cargo install ...` in `install_doc_deps.sh`

adpaco-aws (approved):
Thanks, @tautschnig !

zhassan-aws (approved):
Thanks!

.github/workflows/kani.yml
tautschnig: No, that's required to render the graph in `docs/src/benchcomp-cli.md`, as @karkhaz reminded me. But then he told me that we should be able to remove any litani-related left-overs, so I'll be doing this in an extra commit in this PR.

zhassan-aws: For some reason, the graph referred to in https://model-checking.github.io/kani/benchcomp-cli.html doesn't appear rendered for me. Is anyone seeing the same?

karkhaz: Yes, that's because of a separate PR that removed the mdbook-graphviz crate. See [this](https://github.com/model-checking/kani/issues/3005)

PR #3124, opened 2024-04-04
Remove unnecessary build step for some workflows

zhassan-aws:
Workflows that run the perf regressions (`scripts/kani-perf.sh`) do not need to build Kani, as the perf script itself builds Kani in release mode:

https://github.com/model-checking/kani/blob/dc0978043c52492112e4ad37a617fd3c8100ef1f/scripts/kani-perf.sh#L12

The extra build step was causing Kani to be built two extra times for the `benchcomp` flow because the workflow was building it in debug mode, so the script will end up rebuilding it in release mode. This should save about 3 minutes for the `benchcomp` flow.


zhassan-aws:
Turns out this occurs in non-perf flows as well because `scripts/kani-regression.sh` builds Kani using `RUSTFLAGS="-D warnings" cargo build-dev` while the workflow builds it using `cargo build-dev`, so we end up always building Kani twice. See https://github.com/model-checking/kani/actions/runs/8562243400/job/23465096745 for an example run where this occurred.

adpaco-aws (approved):
Thanks, @zhassan-aws !

Issue #2640, opened 2023-07-28
Report region-based coverage results


adpaco-aws:
Requested feature: The approach introduced in #2612 is a good start for coverage, but the end goal is to have region-based coverage results which are clearer than line-based for users. Maybe this way we can avoid #2639 while at the same time improving how report coverage.
Use case: Coverage results
Link to relevant documentation (Rust reference, Nomicon, RFC): #2612 

Test case:

```rust
<code>
```

PR #3143, opened 2024-04-15
RFC: Source-based code coverage

adpaco-aws:
Upgrades the Kani coverage feature with the source-based code coverage implementation used in the Rust compiler.

"Software Design" section intentionally left blank for now so we can focus on the user experience. Other sections are also missing some content due to other dependencies (e.g., benchmarking results).

Related to #2640 


rfc/src/rfcs/0008-source-coverage.md
```diff
@@ -0,0 +1,277 @@
...
+but other coverage frameworks don't support the concept of code regions.
+
+### The Kani coverage workflow
+
+The proposed Kani coverage workflow imitates the LLVM coverage workflow as much as possible.
+
+The two main components of the Kani coverage workflow will be the following:
+ 1. A new subcommand `cov` that drives the coverage workflow in Kani and
+    produces machine-readable coverage results.
+ 2. A new tool `kani-cov` that consumes the machine-readable coverage results
```
karkhaz, line 142: It's not clear to me why we need our own tool to produce human-readable outputs.

I think I understand the need for step 1. And I agree that there needs to be a post-processing tool for some of the steps you describe below, like merging the `.kaniraw` files and excluding stdlib results.

However, why can't we make the final output (after merging) of `kani-cov` be a `.profraw` file that can be given to `llvm-cov`? This would presumably allow for the same coverage summary and terminal visualization as you have described above, for free.

If the answer is that `llvm-cov` also requires the instrumented binary (which you're replacing here with `*.kanimap`)---could we make it compatible with LLVM by writing out a degenerate LLVM file that contains only a coverage map section, and pass that to `llvm-cov`?

(I'm sure you have good answers to these but it's just not clear to me. Below, it says

> At the moment, there's only another unstable, GCC-compatible code coverage implementation based on the Gcov format. The Gcov format is line-based so it's not able to report region coverage results.

but that seems contradicted by the fact that `llvm-cov` does support region coverage and seems to give us what we need here)

rfc/src/rfcs/0008-source-coverage.md
```diff
@@ -0,0 +1,277 @@
...
+```
+
+[^note-source]: The LLVM project refers to their own coverage feature as "source-based code coverage".
+It's not rare to see the term "region coverage" being used instead to refer to the same thing.
+That's because LLVM's source-based code coverage feature can report coverage for code regions,
+but other coverage frameworks don't support the concept of code regions.
+
+### The Kani coverage workflow
+
+The proposed Kani coverage workflow imitates the LLVM coverage workflow as much as possible.
```
karkhaz, line 137: This seems to imply that to verify the code you run `kani`, and to produce the coverage information you run `kani cov`. My guess is that both of these will need to run symex (though only the first one will run SAT, or at least the property-checking part of SAT).

This seems unnecessary. Why not make coverage an option to kani instead of a separate subcommand? That is, `kani --dump-coverage` means "verify the program, and while you're at it, dump out the `.kaniraw` and `.kanimap` that will later be postprocessed".

I agree that coverage checking will become a vital part of many users' workflow (as you've noted with the bullets beginning with "we expect users to employ this coverage-related option on several stages of a verification effort"). So I think it's important not to waste time (both compute time and running multiple commands). One thing I haven't seen mentioned is IDE integration; I would imagine users would want to press a single button and get both the verification results, and also coverage info highlighted in the IDE... it seems to me that with the proposed flow, the IDE would have to run kani twice.

rfc/src/rfcs/0008-source-coverage.md
```diff
@@ -0,0 +1,277 @@
...
+They will have to use the `show` command for that:
+
+```sh
+kani-cov show my-coverage/default.kanimap -format=html -instr-profile=my-coverage.kanicov -o coverage-report
+```
+
+This time, the command will generate a `coverage-report` folder including a
+browsable HTML webpage that highlights the regions covered in the source
+according to the coverage results in `my-coverage.kanicov`.
+
```
karkhaz, line 238: What about IDE integration? (However, a quick search seems to indicate that the [VSCode code coverage plugin](https://github.com/markis/vscode-code-coverage) only highlights whole lines. That would be a good reason to only support terminal and HTML. But I still think it's worth calling out explicitly)

rfc/src/rfcs/0008-source-coverage.md
adpaco-aws: 1. You're right, there's no need to have a subcommand. I haven't found a reason to have it be a subcommand, and I think it'd actually complicate things more than anything. So I've changed it to be the `-cov`/`--coverage` option again.
2. At the moment, a single run with the coverage option will get you both the verification results and the coverage results. I don't know why you think users will have to run Kani twice for an IDE, although that's what I expect to have in most CI jobs (i.e., run standard verification to decide whether to block or not a PR, then run coverage to maybe block if coverage percentage decreased too much or just to have reports available to reviewers).
3. Sorry, I've neglected IDE integration. Let me give you a more elaborate reply on your other comment about IDE integration.

rfc/src/rfcs/0008-source-coverage.md
adpaco-aws: Sorry, I've pushed a new subsection to discuss the integration with the Kani VS Code Extension. To integrate with that particular plugin we'd just need to be able to export to the LCOV format. Let me call it out there.

rfc/src/rfcs/0008-source-coverage.md
```diff
@@ -0,0 +1,300 @@
...
+The [draft implementation](https://github.com/model-checking/kani/pull/3119)
+uses the JSON format but we might consider using a binary format if it doesn't
+scale.
+
+[^note-kanimap]: Note that the `.kanimap` generation isn't implemented in [#3119](https://github.com/model-checking/kani/pull/3119).
+The [draft implementation of `kani-cov`](https://github.com/model-checking/kani/pull/3121)
+simply reads the source files referred to by the code coverage checks, but it
+doesn't get information about code trimmed out by the MIR linker.
+
+#### The `kani-cov` tool
```
celinval, line 190: I am curious on why this is a separate command instead of a sub-command.

rfc/src/rfcs/0008-source-coverage.md
```diff
@@ -0,0 +1,300 @@
...
+Hence, the `kani-cov` tool corresponds to the set of LLVM tools
+(`llvm-profdata`, `llvm-cov`, etc.) that are used to produce coverage outputs
+through the LLVM coverage workflow.
+
+In contrast to LLVM, we'll have a single tool for all Kani coverage-related needs.
+We suggest that the tool initially offers three subcommands[^note-export]:
+ - `merge`: to combine the coverage results of one or more `.kaniraw` files into
+ a single `.kanicov` file, which will be required for the other subcommands.
+ - `report`: to display a summary of the coverage results.
+ - `show`: to produce source-based code coverage reports in human-readable formats (e.g., HTML).
```
celinval, line 203: Do we need this? I'm wondering if we really need to split Kani's coverage flow into multiple commands.

It makes sense to me that LLVM does that, since users want to execute the binaries independently and merge the results after the fact. But with Kani, `kani-driver` is already responsible for compiling and running the verification, so could `kani-driver` also aggregate the results at the end?

BTW, I'm OK if we want to put that logic in its own separate binary (`kani-cov`), but I'm wondering if `kani-driver` should be the one invoking `kani-cov`.

celinval (commented):
Thanks for diving deep into this one. It's not a trivial feature to implement.

rfc/src/rfcs/0008-source-coverage.md
```diff
@@ -0,0 +1,300 @@
+- **Feature Name:** Source-based code coverage (`source-coverage`)
```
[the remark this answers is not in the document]
adpaco-aws: That's what I did when I started writing it, but then went back to this one I could reuse a lot of the content in "User Impact". Let me create a new one and I'll copy some of that content over. Thanks for the suggestion!

rfc/src/rfcs/0008-source-coverage.md
```diff
@@ -0,0 +1,300 @@
...
+
+Therefore, the first part of the coverage workflow will be managed by Kani.
+Then, in the other part, we will use the `kani-cov` tool to produce the coverage
+output(s) we're interested in.
+
+In the following, we describe each one of these components in more detail.
+
+#### The `-cov` option
+
+The coverage workflow will be kicked off through a new `-cov` option:
```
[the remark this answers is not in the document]
adpaco-aws: Yes, Clap didn't let me do this.

rfc/src/rfcs/0008-source-coverage.md
adpaco-aws: We don't really need this, but I think there two main reasons to propose it:
 1. **To resemble the LLVM workflow as much as possible**, potentially easing an integration with the LLVM framework in the future. Unfortunately, the integration with LLVM tooling isn't completely explained in the RFC at the moment, but I think you get the idea.
 2. **To allow users to aggregate and extract coverage data however they want** (i.e., to have flexible coverage options).

But you're right in that this may not represent the best experience for Kani users. The three sub-commands provide much flexibility, yes. But based on my own experience, they aren't that ergonomic and I'm always copy-pasting the same ones most of the time.

celinval: Maybe we can have a unified flow out of the box, and later allow users to aggregate them in different ways

rfc/src/rfcs/0008-source-coverage.md
adpaco-aws: Yes, it'd make a lot of sense to have a sub-command instead. Calling this from `kani-driver` as you mention in the other thread is likely the best choice, but it's not clear to me if should have a binary or just a library for `kani-cov`. What do you think?

rfc/src/rfcs/0008-source-coverage.md
adpaco-aws: > Maybe we can have a unified flow out of the box, and later allow users to aggregate them in different ways

Yes, I'm going to make changes to the RFC and propose something like that. Thanks!

rfc/src/rfcs/0008-source-coverage.md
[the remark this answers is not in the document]
adpaco-aws: Fixed.

rfc/src/rfcs/0008-source-coverage.md
adpaco-aws: Made a few updates w.r.t. this comment:
 1. `--coverage` is an option in the standard Kani verification workflow. `cov` is a subcommand that aliases `kani-cov`. Also made clearer how `.kaniraw` and `.kanimap` files are produced and used in the coverage workflow.
 2. Added a subsection to discuss Kani VS Code integration. It's not very concrete since it'll likely just use the tooling we design here.

rfc/src/rfcs/0008-source-coverage.md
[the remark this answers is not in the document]
adpaco-aws: The previous RFC is now marked as "Cancelled" in this PR. The present RFC also mentions replacing the feature explicitly.

zhassan-aws (approved):
Very-well written, and the proposal looks great!

rfc/src/rfcs/0008-source-coverage.md
adpaco-aws: This should be much clearer now thanks to the design and future work sections. Resolving.

rfc/src/rfcs/0011-source-coverage.md
```diff
@@ -0,0 +1,589 @@
...
+writes the coverage files into the `target/kani/<target_triple>/cov/` directory.
+
+Users aren't expected to read the information in any of these files.
+Therefore, there's no need to restrict their format.
+The [draft implementation](https://github.com/model-checking/kani/pull/3119)
+uses the JSON format but we might consider using a binary format if it doesn't
+scale.
+
+In addition, Kani will produce two types of coverage results:
+ 1. A coverage summary
```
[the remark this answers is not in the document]
adpaco-aws: My bad, I forgot about this. I've completed this as follows:

> In addition, Kani will produce two types of coverage results:
> 1. A coverage summary with the default options.
> 2. A terminal-based coverage report with the default options. However, we will
 only do this if the program is composed of a single source
 file*.

Suggestions are welcome on this "default" behavior. It makes sense to me that users get coverage results printed to terminal if their programs are small enough, as this should be a good experience for tutorials and simple tests. But maybe they will be confused when they run it on a larger program and nothing comes out...

Aside from that, I've also added a subsection on "Optimization with coverage-counter expressions" to "Future Possibilites".

feliperodri (approved):

PR #3151, opened 2024-04-19
Loop Contracts Annotation for While-Loop

qinheping:
This PR introduce the loop contracts annotation for while-loops using proc-macro.
A while loop of the form 
 ``` rust
#[kani::loop_invariant(inv)]
while guard {
    body
}
```
is annotated as 
 ``` rust
#[inline(never)]
#[kanitool::fn_marker = "kani_register_loop_contract"]
const fn kani_register_loop_contract_id<T, F: FnOnce() -> T>(f: F) -> T {
    unreachable!()
}
while 
kani_register_loop_contract_id(|| -> bool {inv}) && guard {
    loop_body;
}
```
We then replace the function body of the register function `kani_register_loop_contract_id` by a call to its function argument `f`. 

In the loop-contract transformation, we move the calls to register functions to a new basic block as new loop latches and redirect all loop latches (for loops containing `continue` there can be multiple latches pointing to the same loop head) to the new loop latches to make sure that
1. each loop contain only one loop latch;
2. the terminator of the loop latch is a call to the register function.

In detail, we transform
```ignore
loop_head_block: {
     loop_head_stmts
     _v = kani_register_loop_contract(move args) -> [return: next_idx];
}

 ...
loop_body_blocks
 ...

ori_loop_latch_block: {
     loop_latch_stmts
     goto -> loop_head_block;
}
```
to blocks
```ignore
// loop head block
loop_head_block: {
     _v = true
     goto -> next_idx
}

...
loop_body_blocks
...

ori_loop_latch_block: {
     loop_latch_stmts
     goto -> new_loop_latch_block;
}

new_loop_latch_block: {
     _v = kani_register_loop_contract(move args) -> [return: next_idx];
}
```
The register functions will be transformed to
```ignore
    bb0: {
        _0 = closure@fn as std::ops::FnOnce::call_once(move _1, ()) -> [return: bb1, unwind unreachable];
    }
    bb1: {
        return;
    }
```

At the end, the call to the register function will be codegened as backward goto with loop contracts annotated.


library/kani/src/lib.rs
```diff
@@ -140,6 +140,25 @@ pub const fn assert(cond: bool, msg: &'static str) {
 #[rustc_diagnostic_item = "KaniCover"]
 pub const fn cover(_cond: bool, _msg: &'static str) {}
 
+/// This function is only used for loop contract annotation.
+/// It behaves as a placeholder to telling us where is the loop head and loop end
+/// that we want to annotate to.
+#[inline(never)]
+#[rustc_diagnostic_item = "KaniLoopInvariant"]
+pub const fn kani_loop_invariant(_cond: bool) {}
```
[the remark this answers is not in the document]
celinval: These functions seemed to be used as markers, so maybe rename that so it is explicit. This one for example seems to be marking the end of the loop, so it could be something like.
```rust
pub const fn kani_loop_end_marker() {}
```

celinval (commented):
I think I'm missing the big picture here, since the PR itself is currently a no-op and the attribute itself has no documentation.

library/kani_macros/src/lib.rs
```diff
@@ -346,6 +358,77 @@ mod sysroot {
...
+            let closre_name = Ident::new(
+                &format!("__kani_loop_invariant_{LOOP_INVARIANT_COUNT}"),
+                Span::call_site(),
+            );
+            quote!(
+                let #closre_name = ||(#inv_expr);
+                kani::kani_loop_invariant_begin();
+                #closre_name();
+                kani::kani_loop_invariant_end();
+                #loop_stmt
```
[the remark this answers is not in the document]
qinheping: Updated the PR description to make them consistent.

tests/ui/loop_contracts/small_slice_eq/small_slcie_eq.rs
```diff
@@ -0,0 +1,41 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// kani-flags: -Z loop-contracts --keep-temps --enable-unstable --cbmc-args --arrays-uf-always --no-standard-checks
```
adpaco-aws, line 4: Do you need `--keep-temps`? Do you need `--no-standard-checks`?

tests/ui/loop_contracts/small_slice_eq/small_slcie_eq.rs
```diff
@@ -0,0 +1,41 @@
...
+
+// Check if loop contracts are correctly applied.
```
adpaco-aws, line 6: Please add module-level documentation with `//!` and a more concrete description of what the test is checking.

tests/ui/loop_contracts/small_slice_eq/small_slcie_eq.rs
adpaco-aws, line 1: Note the typo in `slcie`.

adpaco-aws (commented):
It's looking good but I'm also missing the big picture when it comes to
 * the general approach to loop invariants - are we planning on doing the same for `for` loops?
 * the code generation for invariant/latch blocks - do you describe this transformation in the RFC? 
 * postprocessing of reachability checks - not sure what's being done or why it's needed

kani-compiler/src/codegen_cprover_gotoc/context/loop_contracts_ctx.rs
```diff
@@ -0,0 +1,140 @@
...
+///     In this state, we push all codegen stmts into the invariant block.
+///     We enter this state when codegen for `KaniLoopInvariantBegin`.
+///     We exit this state when codegen for `KaniLoopInvariantEnd`.
+/// 2. loop latch block
+///     In this state, we codegen a statement expression from the
+///     invariant_block annotate the statement expression to the named sub
+///     of the next backward jumping we codegen.
+///     We enter this state when codegen for `KaniLoopInvariantEnd`.
+///     We exit this state when codegen for the first backward jumping.
+#[allow(dead_code)]
```
adpaco-aws, line 35: Why is this needed?

tests/ui/loop_contracts/small_slice_eq/small_slcie_eq.rs
```diff
@@ -0,0 +1,41 @@
...
+#![feature(proc_macro_hygiene)]
+#![feature(ptr_sub_ptr)]
+unsafe fn small_slice_eq(x: &[u8], y: &[u8]) -> bool {
+    debug_assert_eq!(x.len(), y.len());
+    unsafe {
+        let (mut px, mut py) = (x.as_ptr(), y.as_ptr());
+        let (pxend, pyend) = (px.add(x.len() - 4), py.add(y.len() - 4));
+        #[kani::loop_invariant( px as isize >= x.as_ptr() as isize
+        && py as isize >= y.as_ptr() as isize
+        && px as isize - x.as_ptr() as isize == (py as isize - y.as_ptr() as isize))]
```
adpaco-aws, line 18: Can you please add a simpler test with a single variable that doesn't use pointer arithmetic? Or is that not possible for some reason?

cprover_bindings/src/goto_program/stmt.rs
```diff
@@ -283,7 +286,18 @@ impl Stmt {
     pub fn goto<T: Into<InternedString>>(dest: T, loc: Location) -> Self {
         let dest = dest.into();
         assert!(!dest.is_empty());
-        stmt!(Goto(dest), loc)
+        stmt!(Goto { dest, loop_invariants: None }, loc)
+    }
+
+    /// `goto dest;` with loop invariant
+    pub fn goto_with_loop_inv<T: Into<InternedString>>(
```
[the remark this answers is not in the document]
qinheping: Good idea. Updated the code.

qinheping:
@celinval  @feliperodri , I refactored the implementation with closure. It is ready for review now.

celinval:
> @celinval @feliperodri , I refactored the implementation with closure. It is ready for review now.

Can you please update the description of the PR with an overview of the approach you implemented? Thanks

kani-compiler/src/codegen_cprover_gotoc/overrides/loop_contracts_hooks.rs
```diff
@@ -0,0 +1,37 @@
...
+use cbmc::goto_program::{CIntType, Expr, Stmt, Type};
+use rustc_middle::ty::TyCtxt;
+use rustc_span::Symbol;
+use stable_mir::mir::mono::Instance;
+use stable_mir::mir::{BasicBlockIdx, Place};
+use stable_mir::ty::Span;
+
+pub struct LoopInvariantRegister;
+
+impl GotocHook for LoopInvariantRegister {
```
celinval, line 17: tbh, I would just keep it in the same file as the other hooks. But not a big deal, can you document what this is expanding to? Thanks!

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,266 @@
...
+///    to invoke the closure.
+///
+/// 2. Replace the dummy call to the register function with the actual call, i.e., transform
+///
+/// ```ignore
+/// let kani_loop_invariant = || -> bool {inv};
+/// kani_register_loop_contract(kani_loop_invariant)
+/// while guard {
+///     loop_body;
+///     kani_register_loop_contract(||->bool{true});
```
celinval, line 36: wait, why do you register a dummy closure down here?

I think we want to redefine the closure so it captures the arguments again and don't violate the borrow checker. Something like:

```rust
let kani_loop_invariant_entry = || -> bool {inv};
kani_register_loop_contract(kani_loop_invariant_entry)
while guard {
    loop_body;
    let kani_loop_invariant_latch = || -> bool {inv};
    kani_register_loop_contract(kani_loop_invariant_latch);
}
```

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,266 @@
...
+/// }
+/// ```
+///
+///    to
+///
+/// ```ignore
+/// let kani_loop_invariant = || -> bool {inv};
```
celinval, line 43: don't we need to check the invariant before the loop entry?

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+//! Implementation of the loop contracts code generation.
+//!
+
+use proc_macro::TokenStream;
+use proc_macro_error2::abort_call_site;
+use quote::{format_ident, quote};
+use syn::spanned::Spanned;
+use syn::{Expr, Stmt};
+
+fn generate_unique_id_from_span(stmt: &Stmt) -> String {
```
celinval, line 13: nit: I recommend to move the private functions to the bottom of the file for readability sake.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,266 @@
...
+/// while guard {
+///     loop_body;
+///     kani_register_loop_contract(kani_loop_invariant);
```
celinval, line 46: This could break the aliasing and ownership model.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,266 @@
...
+/// }
+///
+/// ```
+///
+/// 3. Move the call to the register function to the loop latch terminator. This is required
```
celinval, line 51: Oh... maybe clarify that we need the call to be latch's terminator.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,266 @@
...
+        Self: Sized,
+    {
+        TransformationType::Stubbing
+    }
+
+    fn is_enabled(&self, _query_db: &QueryDb) -> bool
+    where
+        Self: Sized,
+    {
+        true
```
celinval, line 76: What happens if the loop invariant check hasn't been enabled?

kani-driver/src/call_goto_instrument.rs
```diff
@@ -183,6 +192,18 @@ impl KaniSession {
         self.call_goto_instrument(&args)
     }
 
+    /// Apply annotated loop contracts.
+    pub fn instrument_loop_contracts(&self, file: &Path) -> Result<()> {
+        let args: Vec<OsString> = vec![
+            "--apply-loop-contracts".into(),
+            "--loop-contracts-no-unwind".into(),
+            "--disable-loop-contracts-side-effect-check".into(),
```
celinval, line 200: Why is that?

kani_metadata/src/unstable.rs
```diff
@@ -83,6 +83,8 @@ pub enum UnstableFeature {
     SourceCoverage,
     /// Enable function contracts [RFC 9](https://model-checking.github.io/kani/rfc/rfcs/0009-function-contracts.html)
     FunctionContracts,
+    /// Enable loop contracts [RFC 12]
```
celinval, line 86: Please add the link.

BTW, will this enable the entire flow where a loop is checked and later replaced?

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+                //      kani_register_loop_contract(#inv_ident);
+                let inv_end_stmt: Stmt = syn::parse(
+                    quote!(
+                        #register_ident(||->bool{true});)
+                    .into(),
+                )
+                .unwrap();
+                ew.body.stmts.push(inv_end_stmt);
+            }
+            _ => (),
```
celinval, line 81: Shouldn't this also print an error?

tests/ui/loop_contracts/small_slice_eq/small_slice_eq.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// kani-flags: -Z loop-contracts --enable-unstable --cbmc-args --arrays-uf-always --no-standard-checks
+
+//! Check if loop contracts are correctly applied. The flag --no-standard-checks should be
+//! removed once same_object predicate is supported in loop contracts.
+//! This function is originally the std function
+//! https://github.com/rust-lang/rust/blob/master/library/core/src/str/pattern.rs#L1885
```
celinval, line 9: I think we might want to use the modifications copyright here.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
celinval, line 1: This transformation seems very CBMC specific. I wonder if we should make this more explicit here. @zhassan-aws any thoughts on how would loop invariant be handled with the Aeneas backend?

celinval (commented):
This is awesome! Looking forward to merging this!

kani_metadata/src/unstable.rs
qinheping: Yes, this will enable the entire flow. The two test cases run with loop contracts enable, where the running time is only ~1 seconds comparing to timeout when running without loop contracts

kani-driver/src/call_goto_instrument.rs
qinheping: Because loop contracts now are wrapped in a closure which will be a side-effect expression in CBMC even they may not contain side-effect. So we disable the side-effect check for now and will implement a better check instead of simply rejecting function calls and statement expressions.
https://github.com/diffblue/cbmc/issues/8393

kani-compiler/src/kani_middle/transform/loop_contracts.rs
qinheping: It should be true only when loop contracts is enable. Updated.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
qinheping: The idea of this implementation (and also the loop contracts in CBMC) is that we want to use one single expression as the loop invariants, and let CBMC do the transformation:  rewrite loop to if, duplicate the loop invariant (check and induction assumption) to the entry of the loop as base case, instrument havocing instructions after the induction assumption, and check the loop invariant at the end of the loop.

The one single loop invariant expression in my mind should be `kani_loop_invariant` which is `|| -> bool {inv}`. As it will be called before the entry of the loop after the transformation, it should be assigned before the loop. However, without dummy call, expanding the macro in the following way will violate the borrow check.
```
let kani_loop_invariant = || -> bool {inv};
while guard {
    loop_body;
    kani_register_loop_contract(kani_loop_invariant);
}
```

So I choose to use a dummy call as a place holder and later replace it with the actual call during MIR transform.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
qinheping: We annotate the loop invariants in the loop latch goto and let CBMC instrument the checks.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
qinheping: I understand that it will violate the borrow check in Rust. But is it also a problem if we do it in MIR level?

kani-compiler/src/kani_middle/transform/loop_contracts.rs
qinheping: updated

qinheping:
@celinval 
I modified the description and addressed most of the comments. The one I am not sure about is will the MIR transform break  aliasing and ownership model?

Thank you!

kani-compiler/src/kani_middle/transform/loop_contracts.rs
qinheping: Yes, the transformer is specified for CBMC loop contracts. But we can build new transformers for other backend if they are incompatible with this one.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
celinval: But this MIR transformation is creating an invalid MIR that does not respect borrow checking rules.

kani-driver/src/call_goto_instrument.rs
celinval: Can you please link this issue here with your explanation?

kani-compiler/src/kani_middle/transform/loop_contracts.rs
celinval: Yes. The borrow checker is trying to prevent users from making mistakes since mutation may invalidate memory.

Here is an example where I think this won't work:

```rust
type Data = u8;

struct Container {
    data: Box<Data>,
}

impl Container {
    fn new_data(&mut self, data: Data) {
        self.data = Box::new(data)
    }
}

/// Dummy function that reallocates data inside the loop.
pub fn target() {
    let mut container = Container { data: Box::new(1) };
    let mut i = 0;
    // This captures the pointer to data that will be deallocated
    let invariant = || {
        *container.data > i
    };
    kani_register_loop_contract(invariant);
    while i < 10 {
        i += 1;
        container.new_data(*container.data + i);
        let place_holder = || { true };
        kani_register_loop_contract(place_holder);
    }
}

const fn kani_register_loop_contract<T, F: FnOnce() -> T>(f: F) -> T {
    unreachable!()
}
```

I believe the transformation from this PR would trigger an invalid read by looking at the MIR generated using [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=458da18d5b304ebcf38ddec1224a5884).

It looks like the `invariant` closure captures a pointer to the allocated data (not a reference to the box):
```asm
_1 = Container { data: move _2 };
_3 = const 0_u8;
_18 = deref_copy (_1.0: std::boxed::Box<u8>);
_20 = (((_18.0: std::ptr::Unique<u8>).0: std::ptr::NonNull<u8>).0: *const u8);
_5 = &(*_20);
_6 = &_3;
_4 = {closure@[src/lib.rs:18:21: 18:23](..)} { container: move _5, i: move _6 };
```

Thus, invoking this closure in the latch (after `new_data()` call that reallocates the box) will trigger an invalid memory read.

Can you please check if that's the case?

qinheping: Good point. I checked the case and it does cause problem in current transformation. I am thinking about not using closure to capture the invariants, and rolling back to the marker functions idea. I will try to refactor my code and update here.

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    quote!(
+        {
+        // Dummy function used to force the compiler to capture the environment.
+        // We cannot call closures inside constant functions.
+        // This function gets replaced by `kani::internal::call_closure`.
+        #[inline(never)]
+        #[kanitool::fn_marker = "kani_register_loop_contract"]
+        const fn #register_ident<T, F: FnOnce() -> T>(f: F) -> T {
+            unreachable!()
+        }
```
celinval, line 86:
```suggestion
        const fn #register_ident<F: FnOnce() -> bool>(_f: F) -> bool {
            unreachable!()
        }
```
BTW, I believe that if you take `f` by reference, you are able to return `true` here and fix #3586.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,467 @@
...
+    /// The bb_idx of the new loop latch block.
+    /// Keys are the original loop head.
+    new_loop_latches: HashMap<usize, usize>,
+    /// Statements of loop head with loop contracts.
+    registered_stmts: HashMap<DefId, Vec<Statement>>,
+    /// If loop contracts is enabled.
+    loop_contracts_enabled: bool,
+}
+
+impl GlobalPass for LoopContractPass {
```
celinval, line 85: why is this a global pass? I believe this should be a `TransformPass` instead since the transformation is local to each function.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,467 @@
...
+            for instance in &instances {
+                let body = instance.body().unwrap();
+                let (modified, new_body) =
+                    self.remove_register_function_calls(tcx, body, *instance);
+                if modified {
+                    transformer.cache.entry(*instance).and_modify(|transformation_result| {
+                        *transformation_result = TransformationResult::Modified(new_body);
+                    });
+                }
+            }
```
celinval, line 134: If you modify the body of `kani_register_loop_contract` to just return true, you should be able to remove this since the function is a no-op.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,467 @@
...
+            // First transform functions with loop contracts.
+            for instance in &instances {
+                let body = instance.body().unwrap();
+                let (modified, new_body) = self.transform_main_body(tcx, body, *instance);
+                if modified {
+                    transformer.cache.entry(*instance).and_modify(|transformation_result| {
+                        *transformation_result = TransformationResult::Modified(new_body);
+                    });
+                }
+            }
```
celinval, line 112: Please use the `TransformPass` and remove this code.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,467 @@
...
+            // Move loop head statements with loop contracts to the corresponding register functions.
+            for instance in &instances {
+                let body = instance.body().unwrap();
+                let (modified, new_body) = self.transform_register_body(tcx, body, *instance);
+                if modified {
+                    transformer.cache.entry(*instance).and_modify(|transformation_result| {
+                        *transformation_result = TransformationResult::Modified(new_body);
+                    });
+                }
+            }
```
celinval, line 123: ditto

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,467 @@
...
+///    functions for later codegen them as statement_expression in CBMC loop contracts.
+#[derive(Debug, Default)]
+pub struct LoopContractPass {
+    /// Cache KaniRunContract function used to implement contracts.
+    run_contract_fn: Option<FnDef>,
+    /// The bb_idx of the new loop latch block.
+    /// Keys are the original loop head.
+    new_loop_latches: HashMap<usize, usize>,
+    /// Statements of loop head with loop contracts.
+    registered_stmts: HashMap<DefId, Vec<Statement>>,
```
celinval, line 80: What is this for? Why are we caching statements?

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,467 @@
...
+                let (modified, new_body) = self.transform_register_body(tcx, body, *instance);
+                if modified {
+                    transformer.cache.entry(*instance).and_modify(|transformation_result| {
+                        *transformation_result = TransformationResult::Modified(new_body);
+                    });
+                }
+            }
+        } else {
+            for instance in &instances {
+                let body = instance.body().unwrap();
```
celinval, line 126: We shouldn't be using `instance.body()` since they don't contain changes from other transformation passes. This should also go away by using `TransformPass`.

kani-compiler/src/kani_middle/transform/body.rs
```diff
@@ -428,6 +428,15 @@ impl MutableBody {
         self.blocks.push(BasicBlock { statements: Vec::default(), terminator })
     }
 
+    /// Replace statements from the given basic block
+    pub fn replace_statements(
+        &mut self,
+        source_instruction: &SourceInstruction,
```
celinval, line 434: Please pass the basic block instead. The source instruction is that the function adjusts the instruction so it still points to the same instruction as before.

This code is replacing the basic block code which invalidates the source instruction.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,467 @@
...
+///    }
+///
+///    bb_new_loop_latch: {
+///         loop_head_body
+///         _v = kani_register_loop_contract(move args) -> [return: bb_idx];
+///    }
+///    ```
+///
+/// 3. Move the statements `loop_head_stmts` of loop head  with contracts to the body of register
+///    functions for later codegen them as statement_expression in CBMC loop contracts.
```
celinval, line 71: Sorry, how does this work? What about the local declarations? How do we know this is sound?

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,467 @@
...
+use std::collections::{HashMap, HashSet};
+use std::fmt::Debug;
+
+use super::GlobalPass;
+
+/// This pass will perform the following operations:
+/// 1. Replace the body of `kani_register_loop_contract` by `kani::internal::run_contract_fn`
+///    to invoke the closure.
+///
+/// 2. Transform loops with contracts from
```
celinval, line 32: why do we need this?

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,467 @@
...
+};
+use stable_mir::ty::{FnDef, MirConst, RigidTy};
+use std::collections::VecDeque;
+use std::collections::{HashMap, HashSet};
+use std::fmt::Debug;
+
+use super::GlobalPass;
+
+/// This pass will perform the following operations:
+/// 1. Replace the body of `kani_register_loop_contract` by `kani::internal::run_contract_fn`
```
celinval, line 29: Do we even need to invoke the loop contract at this point? I thought CBMC would handle this.

celinval (commented):
Sorry, might need to sync about this one offline. I'm not sure I understand your transformation

library/kani_macros/src/sysroot/loop_contracts/mod.rs
qinheping: I got the error `error[E0493]: destructor of `F` cannot be evaluated at compile-time`.
I think we can use the same idea of function contracts here that copy the loops when expanding the macro
```
let kani_loop_contract_mode = kani::internal::ORIGINAL
match kani_loop_contract_mode {
   kani::internal::loop_contract_check => #loop_with_loop_contracts_macro
   _ => #original_loop
}
```
And later modify the value of `kani_loop_contract_mode` during transformation.

But I plan to implement this after this PR.

celinval: What I'm saying is that the error will go away if you change it to a reference though. Something like:
```rust
        const fn #register_ident<F: FnOnce() -> bool>(_f: &F) -> bool {
            true
        }
```

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,313 @@
...
+///         _v = kani_register_loop_contract(move args) -> [return: terminator_target];
+///    }
+///    ```
+#[derive(Debug, Default)]
+pub struct LoopContractPass {
+    /// Cache KaniRunContract function used to implement contracts.
+    run_contract_fn: Option<FnDef>,
+    /// The map from original loop head to the new loop latch.
+    /// We use this map to redirect all original loop latches to a new single loop latch.
+    new_loop_latches: HashMap<usize, usize>,
```
celinval, line 75: You can't cache this here. This object will get re-used for all body transformations. If you need to cache something that is relevant to a specific transformation, you should create a new object that only lives during the transformation of a function.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    fn is_enabled(&self, query_db: &QueryDb) -> bool
+    where
+        Self: Sized,
+    {
+        query_db.args().unstable_features.contains(&"loop-contracts".to_string())
+    }
+
+    /// Run a transformation pass on the whole codegen unit.
```
celinval, line 94: please document what this is actually doing.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,313 @@
...
+    fn transform(&mut self, tcx: TyCtxt, body: Body, instance: Instance) -> (bool, Body) {
+        match instance.ty().kind().rigid().unwrap() {
+            RigidTy::FnDef(_func, args) => {
```
celinval, line 97: What about closures (and co-routines)... won't they be `ClosureDef`?

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -0,0 +1,313 @@
...
+                if KaniAttributes::for_instance(tcx, instance).fn_marker()
+                    == Some(Symbol::intern("kani_register_loop_contract"))
+                {
+                    // Replace the body of the register function with `run_contract_fn`'s.
+                    let run = Instance::resolve(self.run_contract_fn.unwrap(), args).unwrap();
+                    (true, run.body().unwrap())
```
celinval, line 103: I believe this is redundant today.

library/kani_macros/src/sysroot/loop_contracts/mod.rs
qinheping: Yes, I also added two tests cases showing harness can be correctly proved when loop contracts disabled.

tests/expected/loop-contract/box.rs
celinval, line 1: This should be a fixme test. Please also create an issue and add a link to it in the comments.

celinval (approved):
As we discussed offline, I am OK merging this just to unblock experimentation. This transformation is not sound since you are drastically moving where variables are used, and we cannot guarantee that the code does not violate the borrow checker (hence the fixme test).

Can you also please add a few more checks, including:
1. Function with multiple loops with annotations
2. Function with nested loops
3. Harness that invokes different functions that have loop contract in them
4. Test that has a `loop_invariant` but that does not pass `-Z loop-contracts`

Thanks

kani-compiler/src/kani_middle/transform/loop_contracts.rs
qinheping: Good points. I think the current implementation also support loop contracts in closures and co-routines. I will enable them and add tests for them. Tracked in #3599.

tests/expected/loop-contract/box.rs
qinheping: Added.

Issue #3099, opened 2024-03-21
Spurious failures caused by handling of storage markers


zhassan-aws:
The handling of MIR's storage markers (`StorageLive` and `StorageDead`) introduced in #3063 causes spurious failures.

For example, on `tests/kani/Spurious/storage_fixme.rs`, the following failures are reported:
```
** 33 of 305 failed (272 undetermined)
Failed Checks: rust_dealloc must be called on an object whose allocated size matches its layout
 File: "/home/ubuntu/git/kani/library/kani/kani_lib.c", line 86, in __rust_dealloc
Failed Checks: misaligned pointer dereference: address must be a multiple of its type's alignment
 File: "src/main.rs", line 149, in NodeRef::<marker::Dying, marker::Leaf>::len
Failed Checks: explicit panic
 File: "src/main.rs", line 300, in NodeRef::<marker::Dying, marker::Leaf>::force
Failed Checks: called `Option::unwrap()` on a `None` value
 File: "/home/ubuntu/.rustup/toolchains/nightly-2024-03-15-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs", line 1985, in std::option::unwrap_failed
Failed Checks: attempt to add with overflow
 File: "src/main.rs", line 336, in Handle::<NodeRef<marker::Dying, marker::Leaf>, marker::KV>::right_edge
Failed Checks: assertion failed: idx <= node.len()
 File: "src/main.rs", line 344, in Handle::<NodeRef<marker::Dying, marker::Leaf>, marker::Edge>::new_edge
Failed Checks: free argument must be NULL or valid pointer
 File: "/home/ubuntu/git/kani/library/kani/kani_lib.c", line 88, in __rust_dealloc
Failed Checks: free argument must be dynamic object
 File: "/home/ubuntu/git/kani/library/kani/kani_lib.c", line 88, in __rust_dealloc
Failed Checks: free argument has offset zero
 File: "/home/ubuntu/git/kani/library/kani/kani_lib.c", line 88, in __rust_dealloc
Failed Checks: double free
 File: "/home/ubuntu/git/kani/library/kani/kani_lib.c", line 88, in __rust_dealloc
Failed Checks: dereference failure: pointer NULL
 File: "src/main.rs", line 177, in NodeRef::<marker::Dying, marker::Leaf>::ascend
Failed Checks: dereference failure: pointer invalid
[...]

PR #3164, opened 2024-04-26
Codegen storage markers as assignments to __CPROVER_dead_object

tautschnig:
This changes our handling of storage markers to be marking is-alive only rather than treating StorageLive as creating a new object. That is, object instances are now tied to their Mir-provided declarations (which, at present, only appear once per function). To still account for when Rust scopes deem an object to be alive, we use StorageLive and StorageDead to update `__CPROVER_dead_object`. This (global) variable is used by CBMC's pointer checks to track when a pointer may not be safe to dereference for it could be pointing to an object that no longer is in scope.

Resolves: #3099

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
@@ -81,9 +81,24 @@ impl<'tcx> GotocCtx<'tcx> {
                 {
                     Stmt::skip(location)
                 } else {
-                    let sym_e = self.codegen_local(*var_id);
-                    let init = self.codegen_default_initializer(&sym_e);
-                    Stmt::decl(sym_e, init, location)
+                    let global_dead_object = self.ensure_global_var(
+                        "__CPROVER_dead_object",
+                        false,
+                        Type::void_pointer(),
+                        Location::none(),
+                        |_, _| None,
+                    );
+                    Stmt::assign(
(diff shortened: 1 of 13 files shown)


tautschnig:
This still requires a fix on the CBMC side for contracts instrumentation not to fail.

kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -76,17 +76,50 @@ impl<'tcx> GotocCtx<'tcx> {
                 self.codegen_set_discriminant(dest_ty, dest_expr, *variant_index, location)
             }
             StatementKind::StorageLive(var_id) => {
-                if self.queries.args().ignore_storage_markers {
+                if !self.current_fn().is_address_taken_local(*var_id) {
                     Stmt::skip(location)
                 } else {
-                    Stmt::decl(self.codegen_local(*var_id), None, location)
+                    let global_dead_object = self.ensure_global_var(
```
celinval, line 82: Can you please encapsulate this declaration inside the goto bindings create?

tautschnig: Addressed in https://github.com/model-checking/kani/pull/2995/commits/7f996e70d98cd9a7e9f5ff985688599e45673898.

tautschnig:
Closing as the changes are now included in #2995.

PR #3167, opened 2024-04-30
RFC: Loop Contracts

qinheping:
RFC for loop contracts in Kani.

Rendered version available [here](https://github.com/qinheping/kani/blob/rfc-loop-contracts/rfc/src/rfcs/0012-loop-contracts.md).

### DIFF
--- a/rfc/src/rfcs/0012-loop-contracts.md
+++ b/rfc/src/rfcs/0012-loop-contracts.md
@@ -82,11 +82,17 @@ In this case, Kani verifies that the loop invariant `x >= 1` is inductive, i.e.,
 
 Also, once Kani proved that the loop invariant is inductive, it can safely use the loop
 invariants to abstract the loop out of the verification process.
-The idea is, instead of exploring all possible branches of the loop, Kani can safely
-substitute the loops with any of program states that satisfy the loop invariants and
-the negation of the loop guard. The requirement of satisfying the negation of the loop
-guard comes from the fact that a path exits the loop must fail the loop guard.
-After the loop is abstracted, the program will be equivalent to the following:
+The idea is, instead of exploring all possible branches of the loop, Kani only needs to
+prove those branches reached from an arbitrary program state that satisfies the loop contracts,
+after the execution of one iteration of the loop.
+
+So, for loops without break statements, proving post-loops properties with loop contracts is
+equivalent to proving the properties with the loop abstracted out as assuming the post-states
+of the loops should satisfying the disjunction of the invariant and the negation of the loop guard.
+The requirement of satisfying the negation of the loop guard comes from the fact that a path
+exits loops without break statements must fail the loop guard.
+
+For example, applying loop contracts in `simple_loop` function is equivalent to the following:
 ```rs
 fn simple_loop_transformed() {
     let mut x: u64 = kani::any_where(|i| *i >= 1);
--- a/rfc/src/rfcs/0012-loop-contracts.md
+++ b/rfc/src/rfcs/0012-loop-contracts.md
@@ -131,14 +131,10 @@ Then when use the loop contracts to abstract the loop, Kani will only havoc the 
 location `x` and keep `y` unchanged. Note that if the `modifies` clause contains also
 `y`, Kani will havoc both `x` and `y`, and hence violate the assertion `y == 1`.
 
-
-Kani will also verify if all writing targets in the loop are included in the `modifies`
-clause.
-
-
-Note that the `modifies` clause is optional, and Kani will infer the write set if not
-provided.
-
+Kani can employs CBMC's write set inference to infer the write set of the loop.
+So users have to specify the `modifies` clauses by their self only when the inferred write
+sets are not complete---there exists some target that could be written to in the loop but 
+is not in the inferred write set.
 
 ### Proof of termination
     
@@ -244,8 +240,16 @@ easier.
 - How do we translate back modify targets that inferred by CBMC to Rust level? 
 - It is not clear how the CBMC loop modifies inference works for Rust code. We need to
   experiment more to decide what would be the best UX of using loop modifies.
+- How do we handle havocing in unsafe code where it is fine to break the safety invariant
+  of Rust? In that case, we may need havocing function that preserves validity invariant
+  but not safety invariant.
 
 ## Future possibilities
+
+- We can improve the UX by allowing users to choose if they want to apply loop contracts
+  to certain annotated loops or not.
+- We can employ CBMC's decreases inference to infer the decreases clauses to reduce the
+  user burden of specifying the decreases clauses.
 
 <!-- For Developers -->
 
--- a/rfc/src/rfcs/0012-loop-contracts.md
+++ b/rfc/src/rfcs/0012-loop-contracts.md
@@ -194,8 +194,8 @@ Kani implements the functionality of loop contracts in three places.
 
 ### Procedural macros `loop_invariant`, `loop_modifies`, and `loop_decreases`.
 
-We extend the three macros `loop_invariant`, `loop_modifies`, and `loop_decreases` to
-corresponding Rust code. Kani will then compile them into MIR-level code.
+We will implement the three proc-macros `loop_invariant`, `loop_modifies`, and `loop_decreases` to
+embed the annotation logic as Rust code. Kani will then compile them into MIR-level code.
 
 
 ### Code Generation for Builtin Functions
--- a/rfc/src/rfcs/0012-loop-contracts.md
+++ b/rfc/src/rfcs/0012-loop-contracts.md
@@ -194,46 +194,17 @@ Kani implements the functionality of loop contracts in three places.
 
 ### Procedural macros `loop_invariant`, `loop_modifies`, and `loop_decreases`.
 
-The `loop_invariant` macro perform code generation for the loop invariant clause.
-The generated code consists of two parts: 
-1. a closure definition to wrap the loop invariant, which is an Boolean expression.
-2. a call to a builtin function `kani_loop_invariant` at end of the loop.
-
-As an example, in the above program, the following code will be generated for the
-loop invariants clauses:
-
-```rs
-fn simple_loop_macro_expanded() {
-    let mut x: u64 = kani::any_where(|i| *i >= 1);
-
-    while x > 1{
-        x = x - 1;
-        kani::kani_loop_invariant_begin_marker();
-        let __kani_loop_invariant: bool = x >= 1;
-        kani::kani_loop_invariant_end_marker();
-    };
-
-    assert!(x == 1);
-}
-```
-
-Similarly, we generate calls to the corresponding builtin functions for `modifies` and `decreases`
-clauses.
+We extend the three macros `loop_invariant`, `loop_modifies`, and `loop_decreases` to
+corresponding Rust code. Kani will then compile them into MIR-level code.
 
 
 ### Code Generation for Builtin Functions
 
-When generating GOTO program from MIR, Kani will first scan for the placeholder function
-calls `kani_loop_invariant_begin_marker` and `kani_loop_invariant_end_marker` in the MIR. 
-Then Kani will generate the corresponding GOTO-level statement expression for all instructions
-between the two placeholder function calls. At last, Kani will add the statement expression
-to the loop latches---the jumps back to the loop head.
+Then in the MIR, we codegen the loop contracts as GOTO-level expressions and annotate them
+into the corresponding loop latches---the jumps back to the loop head.
 
 The artifact `goto-instrument` in CBMC will extract the loop contracts from the named-subs
 of the loop latch, and then apply and prove the extracted loop contracts.
-
-Similarly, Kani will add the `modifies` targets into the named-subs of the loop latch for
-CBMC to extract and prove the loop contracts.
 
 
 ### GOTO-Level Havocing
@@ -271,6 +242,8 @@ easier.
   loop contracts are not enough prove the harness, we expect the loop-contract synthesizer
   can fix the loop contracts.
 - How do we translate back modify targets that inferred by CBMC to Rust level? 
+- It is not clear how the CBMC loop modifies inference works for Rust code. We need to
+  experiment more to decide what would be the best UX of using loop modifies.
 
 ## Future possibilities
 
--- a/rfc/src/rfcs/0012-loop-contracts.md
+++ b/rfc/src/rfcs/0012-loop-contracts.md
@@ -1,0 +1,235 @@
+- **Feature Name:** Loop Contracts
+- **Feature Request Issue:** 
+- **RFC PR:** 
+- **Status:** Under Review
+- **Version:** 1
+- **Proof-of-concept:** 
+
+-------------------
+
+## Summary
+
+Loop contracts provide way to safely abstract loops of a program, typically
+in order to accelerate the verification process, and remove the loop unwinding
+bounds. The key idea is to overapproximate the possible set of program states,
+while still being precise enough to be able to prove the desired property.
+
+## User Impact
+
+Loop contracts provide an interface for a verified, sound abstraction.
+The goal for specifying loop contracts in the source code is three fold:
+
+* Unbounded verification: Currently Kani does not support proving correctness
+  (i.e. assertions never fail) on programs with unbounded control flow (e.g. 
+  loops with dynamic bounds). Kani unrolls all unbounded loops until a few
+  iterations and then verifies this unrolled program — it thus provides a much
+  weaker guarantee on correctness.
+* Faster CI runs: These contracts, when provided, would also significantly
+  improve Kani's verification time since all loops would be unrolled only to
+  a single iteration, as opposed to a small number of iterations which is
+  Kani's current behavior.
+
+
+
+Loop contracts are completely optional with no user impact if unused. This
+RFC proposes the addition of new attributes, and functions, that shouldn't
+interfere with existing functionalities.
+
+
+## User Experience
+
+A loop contract specifies the behavior of a loop as a predicate that
+can be checked against the loop implementation, and used to abstract out
+the loop in the verification process.
+
+We illustrate the usage of loop contracts with an example.
+Consider the following program:
+```rs
+fn main() {
+    let mut x: u64 = kani::any_where(|i| *i >= 1);
+
+    while x > 1{
+        x = x - 1;
+    };
+
+    assert!(x == 1);
+}
+```
+The loop in the `main` function keep subtracting 1 from `x` until `x` is 1.
+However, Kani currently needs to unroll the loop for `u64::MAX` number of times
+to verify the assertion at the end of the program. 
+
+With loop contracts, the user can specify the behavior of the loop as follows:
+```rs
+fn main() {
+    let mut x: u64 = kani::any_where(|i| *i >= 1);
+
+    #[kani::loop_invariant(x >= 1)]
+    while x > 1{
+        x = x - 1;
+    };
+
+    assert!(x == 1);
+}
+```
+The loop invariant clause `#[kani::loop_invariant(x >= 1)]` specifies the loop
+invariants that must hold at the beginning of each iteration of the loop right before
+checking the loop guard.
+
+In this case, Kani verifies that the loop invariant `x >= 1` is inductive, i.e.,
+`x` is always greater than or equal to 1 at each iteration before checking `x > 1`.
+
+
+Also, once Kani proved that the loop invariant is inductive, it can safely use the loop
+invariants to abstract the loop out of the verification process.
+The idea is, instead of exploring all possible branches of the loop, Kani can safely
+substitute the loops with any of program states that satisfy the loop invariants and
+the negation of the loop guard. The requirement of satisfying the negation of the loop
+guard comes from the fact that a path exits the loop must fail the loop guard.
+After the loop is abstracted, the program will be equivalent to the following:
+```rs
+fn main() {
+    let mut x: u64 = kani::any_where(|i| *i >= 1);
+
+    x = kani::any(); // Arbitrary program state that
+    kani::assume( !(x > 1) && x >= 1); // satisfies !`guard` && `inv` 
+
+    assert!(x == 1);
+}
+```
+The assumption above is actually equivalent to `x == 1`, hence the assertion at the end
+of the program is proved.
+
+### Write Sets and Havocking
+
+For those memory locations that are not modified in the loop, loop invariants state
+that they stay unchanged throughout the loop are inductive. In other words, Kani should
+only havoc the memory locations that are modified in the loop. This is achieved by
+specifying the `modifies` clause for the loop. For example, the following program:
+```rs
+fn main() {
+    let mut x: u64 = kani::any_where(|i| *i >= 1);
+    let mut y: u64 = kani::any_where(|i| *i >= 1);
+
+    #[kani::loop_invariant(x >= 1)]
+    #[kani::loop_modifies(x)]
+    while x > 1{
+        x = x - 1;
+    };
+
+    assert!(x == 1);
+}
+```
+write to only `x` in the loop, hence the `modifies` clause contains only `x`.
+Then when use the loop contracts to abstract the loop, Kani will only havoc the memory
+location `x` and keep `y` unchanged. 
+
+
+Kani will also verify if all writing targets in the loop are included in the `modifies`
+clause.
+
+
+Note that the `modifies` clause is optional, and Kani will infer the write set if not
+provided.
+
+## Detailed Design
+
+
+Kani implements the functionality of loop contracts in three places.
+
+1. Procedural macros `loop_invariant` and `loop_modifies`.
+2. Code generation for builtin functions expanded from the above two macros.
+3. GOTO-level loop contracts using CBMC's contract language generated in
+   `kani-compiler` for `loop-modifies` clauses.
+
+### Procedural macros `loop_invariant` and `loop_modifies`.
+
+The `loop_invariant` macro perform code generation for the loop invariant clause.
+The generated code consists of two parts: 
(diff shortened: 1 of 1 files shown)


rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+
+```rs
+fn main() {
+    let mut x: u64 = kani::any_where(|i| *i >= 1);
+
+    let _loop_invariant_closure = || x >= 1;
+    while x > 1{
+        x = x - 1;
+        kani_loop_invariant_begin();
+        _loop_invariant_closure()
```
celinval, line 163: Invoking this closure will limit the usage of loop contracts in functions declared `const`. It will also not work if the closure captures a mutable reference that is used inside the loop.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+}
+```
+
+Similarly, we generate a call to the builtin function `kani_loop_modifies` for modifies
+clauses.
+
+
+### Code Generation for Builtin Functions
+
+When generating GOTO program from MIR, Kani will first scan for the placeholder function
```
celinval, line 177: Our codegen generate codes for each instruction at a time. Are you proposing to somehow change that? Is that really necessary?

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+```rs
+fn main() {
+    let mut x: u64 = kani::any_where(|i| *i >= 1);
+
+    let _loop_invariant_closure = || x >= 1;
+    while x > 1{
+        x = x - 1;
+        kani_loop_invariant_begin();
+        _loop_invariant_closure()
+        kani_loop_invariant_end();
```
celinval, line 164: would it make sense to replace this with something like we did for Kani contracts:

```suggestion
        // Allow us to programmatically retrieve the closure captures 
        kani_register_invariant(_loop_invariant_closure);
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+clauses.
+
+
+### Code Generation for Builtin Functions
+
+When generating GOTO program from MIR, Kani will first scan for the placeholder function
+calls `kani_loop_invariant_begin` and `kani_loop_invariant_end` in the MIR. Then Kani
+will generate the corresponding GOTO-level statement expression for all instructions
+between the two placeholder function calls. At last, Kani will add the statement expression
+to the loop latch---the jump back to the loop head.
```
celinval, line 181: latches, I'm guessing

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+CBMC to extract and prove the loop contracts.
+
+
+### GOTO-Level Havocing
+
+The ordinary havocing in CBMC is not aware of the type constraints of Rust type.
+Hence, we will use customized havocing functions for modifies targets. In detail,
+Kani will generate code for the definition of corresponding `kani::any()` functions
+for each modifies target. Then Kani will create a map from the modifies target to the
+the name of its `kani::any()` function, and add the map to the loop latch too.
```
celinval, line 196: Are you adding a new property to goto-program that will allow us to provide that information, or would this be a separate artifact that we will need to pass to `goto-instrument`? Can we expand that to handle havoc in contracts modifies?

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Rust-Level Transformation vs CBMC 
+
+Besides transforming the loops in GOTO level using `goto-instrument`,
+we could also do the transformation in Rust level using procedural macros, or
+in MIR level.
+
+There are two reasons we prefer the GOTO-level transformation.
+First, `goto-instrument` is a mature tool that can correctly instrument the frame
+condition checking for the transformed loop, which will safe us from reinventing
```
celinval, line 213:
```suggestion
condition checking for the transformed loop, which will save us from reinventing
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+The loop in the `main` function keep subtracting 1 from `x` until `x` is 1.
+However, Kani currently needs to unroll the loop for `u64::MAX` number of times
+to verify the assertion at the end of the program. 
+
+With loop contracts, the user can specify the behavior of the loop as follows:
+```rs
+fn main() {
+    let mut x: u64 = kani::any_where(|i| *i >= 1);
+
+    #[kani::loop_invariant(x >= 1)]
```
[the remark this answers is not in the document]
feliperodri: No, the assertion after the loop would fail if the invariant is just `true`. We need an inductive invariant to properly replace this loop that is specific enough to prove the remaining properties of the code.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+    while x > 1{
+        x = x - 1;
+    };
+
+    assert!(x == 1);
+}
+```
+write to only `x` in the loop, hence the `modifies` clause contains only `x`.
+Then when use the loop contracts to abstract the loop, Kani will only havoc the memory
+location `x` and keep `y` unchanged.
```
feliperodri, line 125: This example is a bit weird, cause we never even use `y`. It'd be nice to see a case where, if `y` is havoced, the property would fail due to an over-approximation.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+Then when use the loop contracts to abstract the loop, Kani will only havoc the memory
+location `x` and keep `y` unchanged. 
+
+
+Kani will also verify if all writing targets in the loop are included in the `modifies`
+clause.
+
+
+Note that the `modifies` clause is optional, and Kani will infer the write set if not
+provided.
```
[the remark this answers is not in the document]
feliperodri: We cannot infer in all case. If Kani cannot infer all writing targets, it will fail when checking if all writing targets are included in the write set, as previously described.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Detailed Design
+
+
+Kani implements the functionality of loop contracts in three places.
+
+1. Procedural macros `loop_invariant` and `loop_modifies`.
```
feliperodri, line 140: What about decreases clauses?

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+2. Code generation for builtin functions expanded from the above two macros.
+3. GOTO-level loop contracts using CBMC's contract language generated in
+   `kani-compiler` for `loop-modifies` clauses.
+
+### Procedural macros `loop_invariant` and `loop_modifies`.
+
+The `loop_invariant` macro perform code generation for the loop invariant clause.
+The generated code consists of two parts: 
+1. a closure definition to wrap the loop invariant, which is an boolean expression.
```
feliperodri, line 149:
```suggestion
1. a closure definition to wrap the loop invariant, which is an Boolean expression.
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+### Code Generation for Builtin Functions
+
+When generating GOTO program from MIR, Kani will first scan for the placeholder function
+calls `kani_loop_invariant_begin` and `kani_loop_invariant_end` in the MIR. Then Kani
+will generate the corresponding GOTO-level statement expression for all instructions
+between the two placeholder function calls. At last, Kani will add the statement expression
+to the loop latch---the jump back to the loop head.
+
+The artifact `goto-instrument` in CBMC will extract the loop contracts from the named-subs
+of the loop latch, and the apply and prove the extracted loop contracts.
```
feliperodri, line 184:
```suggestion
of the loop latch, and then apply and prove the extracted loop contracts.
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+### GOTO-Level Havocing
+
+The ordinary havocing in CBMC is not aware of the type constraints of Rust type.
+Hence, we will use customized havocing functions for modifies targets. In detail,
+Kani will generate code for the definition of corresponding `kani::any()` functions
+for each modifies target. Then Kani will create a map from the modifies target to the
+the name of its `kani::any()` function, and add the map to the loop latch too.
+
+On the CBMC site, `goto-instrument` will extract the map and instrument the customized
+havocing functions for the modifies targets.
```
feliperodri, line 199: So basically you created a mechanism in CBMC that allow users to specify *how* they want to havoc the write set? How is that? A pair of memory to be havoced + invariant? Shouldn't this be expose to contracts?

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+
+<!-- For Developers -->
+
+- The idea of using closure to wrap the loop invariant is a bit hacky. It is not
+  clear what behavior of the loop will move the variables in the closure, and hence
+  invalidate the closure. Is there a better way to do this?
+<!-- 
+- Is there any part of the design that you expect to resolve through the RFC process?
+- What kind of user feedback do you expect to gather before stabilization? How will this impact your design? 
+-->
```
feliperodri, line 229:
```suggestion
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Future possibilities
+
+<!-- For Developers -->
```
feliperodri, line 233: I'd at least mention decreases clauses here if you don't want to include it in this stage.

feliperodri (changes requested):

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,235 @@
...
+
+In this case, Kani verifies that the loop invariant `x >= 1` is inductive, i.e.,
+`x` is always greater than or equal to 1 at each iteration before checking `x > 1`.
+
+
+Also, once Kani proved that the loop invariant is inductive, it can safely use the loop
+invariants to abstract the loop out of the verification process.
+The idea is, instead of exploring all possible branches of the loop, Kani can safely
+substitute the loops with any of program states that satisfy the loop invariants and
+the negation of the loop guard. The requirement of satisfying the negation of the loop
```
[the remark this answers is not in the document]
qinheping: For loops with break, there is no general form of the post conditions that the existing state should satisfy. I will clarify in the text.

rfc/src/rfcs/0012-loop-contracts.md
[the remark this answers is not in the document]
celinval: So the clause is not always optional. I'm trying to understand what is the UX here. How will the user know when the `modifies` is needed? At least for function contracts, our modifies errors are currently very cryptic, and it can include temporary variables.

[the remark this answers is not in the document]
qinheping: CBMC will check all assignments to see if the target is in the modifies set for loop contracts, and report target-not-assignable failure if it finds any assigning targets not in `modifies`. So we should notify users to provide `modifies` if we see such failures.

rfc/src/rfcs/0012-loop-contracts.md
celinval: We can discuss implementation details a bit later, but I'm not convinced that this is the best approach since it adds more complexity to the codegen code.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,274 @@
...
+
+Also, once Kani proved that the loop invariant is inductive, it can safely use the loop
+invariants to abstract the loop out of the verification process.
+The idea is, instead of exploring all possible branches of the loop, Kani only needs to
+prove those branches reached from an arbitrary program state that satisfies the loop contracts,
+after the execution of one iteration of the loop.
+
+So, for loops without break statements, proving post-loops properties with loop contracts is
+equivalent to proving the properties with the loop abstracted out as assuming the post-states
+of the loops should satisfying the disjunction of the invariant and the negation of the loop guard.
```
celinval, line 91: I'm having a hard time parsing this sentence.

rfc/src/rfcs/0012-loop-contracts.md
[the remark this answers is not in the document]
celinval: 1. How will the user know what Kani has inferred as the write set? What are the rules here?
2. CBMC suggestions at this point for modifies is fairly useless. How are you planning to address this? See #3336

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,274 @@
...
+
+Note that the `modifies` clause is optional, and Kani will infer the write set if not
+provided.
+
+
+### Proof of termination
+    
+    Loop contracts also provide a way to prove the termination of the loop.
+    Without the proof of termination, the loop contracts could lead to a false
+    positive result. For example, consider the following program:
```
celinval, line 148:
```suggestion
Loop contracts also provide a way to prove the termination of the loop.
Without the proof of termination, the loop contracts could lead to a false
positive result. For example, consider the following program:
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,274 @@
...
+
+
+Note that the `modifies` clause is optional, and Kani will infer the write set if not
+provided.
+
+
+### Proof of termination
+    
+    Loop contracts also provide a way to prove the termination of the loop.
+    Without the proof of termination, the loop contracts could lead to a false
```
celinval, line 147: What do you call "false positive" here? That would be OK since this is an over-abstraction, right?

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,274 @@
...
+
+```rs
+fn simple_loop_macro_expanded() {
+    let mut x: u64 = kani::any_where(|i| *i >= 1);
+
+    while x > 1{
+        x = x - 1;
+        kani::kani_loop_invariant_begin_marker();
+        let __kani_loop_invariant: bool = x >= 1;
+        kani::kani_loop_invariant_end_marker();
```
celinval, line 213: FYI, loops are allowed in constant code. If we decide to embed the code as proposed here, this will restrict what can be used in the invariant annotation. I.e., the loop invariant inside constant functions will only be able to invoke other constant functions, and it may limit the usage of Kani intrinsics.

This pattern does occur in functions such as `<[u8]>.is_ascii()`.

rfc/src/rfcs/0012-loop-contracts.md
celinval: Not sure we can assume that to always be the case. Have you tried enabling optimizations? Anyway, I'm OK if the implementation for now assumes that, as long as the implementation can be easily extended later

rfc/src/rfcs/0012-loop-contracts.md
celinval: One thing that may be worth calling out is that it is perfectly fine for a loop to break the safety invariant of a value. So calling `kani::any()` could potentially be unsound if the loop is inside an unsafe block / function. If that is the case, you may need something in between. I.e., something that preserves validity invariant, but not safety invariant.

The UX question becomes how will the user specify the different scenarios. Do we create different modifies clauses? Or do we force the user to state the invariant that the modified type is safe?

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,274 @@
...
+
+There are two reasons we prefer the GOTO-level transformation.
+First, `goto-instrument` is a mature tool that can correctly instrument the frame
+condition checking for the transformed loop, which will save us from reinventing
+the error-prone wheel. Second, the loop contracts synthesis tool we developed and
+are developing are all based on GOTO level. Hence, doing the transformation in
+the GOTO level will make the integration of loop contracts with the synthesis tool
+easier.
+
+## Open questions
```
celinval, line 268: I think you should at least mention here that we need to figure out how this work interacts with the synthesis tool. Do we expect the synthesis tool to be able to generate contracts later on? Can the synthesis tool somehow be used to complement the loop invariants, allowing users to write incomplete invariants? Or will they be incompatible?

rfc/src/rfcs/0012-loop-contracts.md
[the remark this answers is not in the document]
qinheping: > 1. How will the user know what Kani has inferred as the write set? What are the rules here?

CBMC reports the inferred write targets. We need to translate them back to rust targets.

>     2. CBMC suggestions at this point for modifies is fairly useless. How are you planning to address this? See [Modifies property points to a temporary variable #3336](https://github.com/model-checking/kani/issues/3336)

We could either map mir variable to their corresponding rust variable with static analysis, or do the assign targets 
inference in MIR level instead with CBMC.

rfc/src/rfcs/0012-loop-contracts.md
qinheping: Sorry I was wrong about it. Loops can simply have multiple latches by containing `continue`. But we can identify all the latches of the same loop as they should point to the same loop head according to the [loop terminology](https://llvm.org/docs/LoopTerminology.html).

rfc/src/rfcs/0012-loop-contracts.md
qinheping: I think it should be consistent as how `kani::any()` works in an unsafe block. I propose to allow users provide custom `any()` functions for each modify target.

rfc/src/rfcs/0012-loop-contracts.md
qinheping: It talks about how loop contracts affect the post-states of the loop. I will update the text and try to make it more understandable.

rfc/src/rfcs/0012-loop-contracts.md
qinheping: Kani can report success of some assertions that are actually unreachable due to the non-termination of the loop. I will clarify in the text.

rfc/src/rfcs/0012-loop-contracts.md
qinheping: Let's discuss offline to see what can be a workaround for it.

rfc/src/rfcs/0012-loop-contracts.md
qinheping: Good point!

qinheping:
@celinval 
I remove implementation detail as it was too much for RFC; and add more open questions.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,252 @@
...
+    };
+
+    assert!(x >= 1);
+}
+```
+After abstracting the loop, the loop will be transformed to no-op, and the assertion
+`x >= 1` will be proved. However, the loop is actually an infinite loop, and the
+assertion will never be reached.
+
+For this reason, Kani will also require the user to provide a `decreases` clause that
```
celinval, line 166: Can we over-approximate if the decreases is not provided or if the loop may not terminate?

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,252 @@
...
+
+1. Procedural macros `loop_invariant`, `loop_modifies`, and `loop_decreases`.
+2. Code generation for builtin functions expanded from the above macros.
+3. GOTO-level loop contracts using CBMC's contract language generated in
+   `kani-compiler`.
+
+### Procedural macros `loop_invariant`, `loop_modifies`, and `loop_decreases`.
+
+We extend the three macros `loop_invariant`, `loop_modifies`, and `loop_decreases` to
+corresponding Rust code. Kani will then compile them into MIR-level code.
```
celinval, line 198:
```suggestion
We will implement the three proc-macros `loop_invariant`, `loop_modifies`, and `loop_decreases` to
embed the annotation logic as Rust code. Kani will then compile them into MIR-level code.
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,252 @@
...
+The artifact `goto-instrument` in CBMC will extract the loop contracts from the named-subs
+of the loop latch, and then apply and prove the extracted loop contracts.
+
+
+### GOTO-Level Havocing
+
+The ordinary havocing in CBMC is not aware of the type constraints of Rust type.
+Hence, we will use customized havocing functions for modifies targets. In detail,
+Kani will generate code for the definition of corresponding `kani::any()` functions
+for each modifies target. Then Kani will create a map from the modifies target to the
```
celinval, line 215: But isn't modifies target optional? In this case, I'm assuming you need to create a map of `Type -> kani::any()` implementation.

rfc/src/rfcs/0012-loop-contracts.md
celinval: Users don't really use `kani::any()` in code outside the harness / stubs. I suggest that you leave this as an open question for now.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,252 @@
...
+    assert!(y == 1);
+}
+```
+write to only `x` in the loop, hence the `modifies` clause contains only `x`.
+Then when use the loop contracts to abstract the loop, Kani will only havoc the memory
+location `x` and keep `y` unchanged. Note that if the `modifies` clause contains also
+`y`, Kani will havoc both `x` and `y`, and hence violate the assertion `y == 1`.
+
+
+Kani will also verify if all writing targets in the loop are included in the `modifies`
```
celinval, line 135: This sentence contradicts the one from line 139.

celinval (commented):
Thanks @qinheping! It's looking pretty good.

feliperodri (approved):
Thank you for addressing all the comments. LGTM!

rfc/src/rfcs/0012-loop-contracts.md
celinval: But I'm talking about cases where there is no decreases to be inferred.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,256 @@
...
+  out loops, and hence remove the need for unwinding loops.
+* Faster CI runs: In most cases, the provided contracts would also significantly
+  improve Kani's verification time since all loops would be unrolled only to
+  a single iteration.
+
+
+
+Loop contracts are completely optional with no user impact if unused. This
+RFC proposes the addition of new attributes, and functions, that shouldn't
+interfere with existing functionalities.
```
celinval, line 35: It is still not clear to me how users opt-in to use loop contracts. Will there be a global option?

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -0,0 +1,252 @@
...
+```
+The loop invariant clause `#[kani::loop_invariant(x >= 1)]` specifies the loop
+invariants that must hold at the beginning of each iteration of the loop right before
+checking the loop guard.
+
+In this case, Kani verifies that the loop invariant `x >= 1` is inductive, i.e.,
+`x` is always greater than or equal to 1 at each iteration before checking `x > 1`.
+
+
+Also, once Kani proved that the loop invariant is inductive, it can safely use the loop
```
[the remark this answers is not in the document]
celinval: Future possibilities are things that are not meant to be resolved as part of this work. However, I do think this is more like an open question. Once the feature is stabilized, do we expect all loop contracts to be applied to all harnesses? Will there be a mechanism to opt-out from using the loop contract? If so, what will be the granularity (per-harness / per-crate). Thanks!

celinval (commented):
I only have one last question about feature opt-in / opt-out. Like contracts, I am not convinced that a user always wants the loop to be abstracted.

rfc/src/rfcs/0012-loop-contracts.md
qinheping: Not sure if I understand this comment correctly. If no decreases is provided/inferred, we can't prove the termination of the loops and hence can only prove the partial correctness.

qinheping (commented):
@celinval Thank you!

rfc/src/rfcs/0012-loop-contracts.md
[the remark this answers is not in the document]
celinval: Can you please add that to the open-questions? I don't think we should stabilize this without this mechanism, but we don't need to pin down the details yet.

rfc/src/rfcs/0012-loop-contracts.md
celinval: Would users be able to use loop contracts in a function like [this `read_char` function](https://github.com/console-rs/console/blob/de2f15a31a8fef0b0e65ef4bdf92cd03c3061dac/src/term.rs#L267-L285)?

celinval (approved):
Thanks @qinheping

rfc/src/rfcs/0012-loop-contracts.md
qinheping: Users can specify loop contracts for `loop` loops, like in the `read_char` function. However, for this function, I don't think we can write any non-trivial contracts for the loop, And it is not guarantee that the loop will terminate as it is possible that `read_key` always return non-char and non-enter.

PR #3169, opened 2024-05-01
Remove kani::Arbitrary from the modifies contract instrumentation

feliperodri:
This is an additional fix for https://github.com/model-checking/kani/pull/3098. With this fix, Kani should be able to check for contracts using modifies clauses that contain references to types that doesn't implement `kani::Arbitrary`. The verification will still fail if the same contract is used as a verified stub.


library/kani_macros/src/sysroot/contracts/check.rs
```diff
@@ -60,7 +60,8 @@ impl<'a> ContractConditionsHandler<'a> {
                 let wrapper_args = if let Some(wrapper_call_args) =
                     inner.iter_mut().find_map(|stmt| try_as_wrapper_call_args(stmt, &wrapper_name))
                 {
-                    let wrapper_args = make_wrapper_args(wrapper_call_args.len(), attr.len());
+                    let wrapper_args =
+                        make_wrapper_idents(wrapper_call_args.len(), attr.len(), "_wrapper_arg_");
```
zhassan-aws, line 64: Since `_wrapper_arg_` is used multiple times, please add a `const` string for it.

library/kani_macros/src/sysroot/contracts/check.rs
```diff
@@ -129,15 +130,34 @@ impl<'a> ContractConditionsHandler<'a> {
...
+            let type_params = make_wrapper_idents(
+                self.annotated_fn.sig.inputs.len(),
+                attr.len(),
+                "WrapperArgType",
+            );
             let sig = &mut self.annotated_fn.sig;
-            for arg in wrapper_args.clone() {
+            for (arg, arg_type) in wrapper_args.clone().zip(type_params) {
+                // Add the type parameter to the function signature's generic parameters list
+                sig.generics.params.push(syn::GenericParam::Type(syn::TypeParam {
```
zhassan-aws, line 147: Can you clarify what the purpose of adding it to the generic parameters list is?

library/kani_macros/src/sysroot/contracts/check.rs
```diff
@@ -193,8 +214,12 @@ fn try_as_wrapper_call_args<'a>(
 
 /// Make `num` [`Ident`]s with the names `_wrapper_arg_{i}` with `i` starting at `low` and
```
zhassan-aws, line 215: Should the doc comment be updated?

zhassan-aws (changes requested):
How is this tested?

library/kani_macros/src/sysroot/contracts/check.rs
feliperodri: Implemented as suggested.

library/kani_macros/src/sysroot/contracts/check.rs
feliperodri: Updated the header documentation to clarify this point. Is it more clear now?

library/kani_macros/src/sysroot/contracts/check.rs
feliperodri: Yes, I included a small update.

feliperodri:
> How is this tested?

@zhassan-aws I included an extra test to cover this case. I was testing this on the function contracts we have fore stdlib.

library/kani_macros/src/sysroot/contracts/check.rs
```diff
@@ -191,10 +221,14 @@ fn try_as_wrapper_call_args<'a>(
...
+/// Make `num` [`Ident`]s with the names `prefix{i}` with `i` starting at `low` and
 /// increasing by one each time.
-fn make_wrapper_args(low: usize, num: usize) -> impl Iterator<Item = syn::Ident> + Clone {
-    (low..).map(|i| Ident::new(&format!("_wrapper_arg_{i}"), Span::mixed_site())).take(num)
+fn make_wrapper_idents(
+    low: usize,
+    num: usize,
+    prefix: &'static str,
+) -> impl Iterator<Item = syn::Ident> + Clone + 'static {
+    (low..).map(move |i| Ident::new(&format!("{}{}", prefix, i), Span::mixed_site())).take(num)
```
zhassan-aws, line 231: Nit:
```suggestion
    (low..).map(move |i| Ident::new(&format!("{prefix}{i}"), Span::mixed_site())).take(num)
```

PR #3179, opened 2024-05-13
Do not run perf tests in parallel

zhassan-aws:
Currently, perf tests (whether run through benchcomp or separately) use the available parallelism of the machine. This sometimes causes memouts when multiple memory-hungry tests run in parallel, causing failures in CI. This PR forces perf regressions to always use a single thread regardless of the available parallelism. While this may cause them to take longer to complete, they should:
1. Tame the spurious memouts occurring in CI
2. Reduce the runtime variability in benchcomp results

### DIFF
--- a/scripts/kani-perf.sh
+++ b/scripts/kani-perf.sh
@@ -27,7 +27,7 @@ done
 suite="perf"
 mode="cargo-kani-test"
 echo "Check compiletest suite=$suite mode=$mode"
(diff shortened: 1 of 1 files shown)


zhassan-aws:
@karkhaz pointed out that benchcomp already runs in CI with a single thread:

https://github.com/model-checking/kani/blob/main/tools/benchcomp/configs/perf-regression.yaml#L15

so this PR is not necessary. Closing.

PR #3188, opened 2024-05-16
[DRAFT] RFC for Ghost State

celinval:
This is still a draft, since it's not clear if this approach can actually be implemented today using CBMC's shadow memory support.

We could consider implementing our own / using @remi-delmas-3000 library.

### DIFF
--- a/rfc/src/rfcs/0010-ghost-state.md
+++ b/rfc/src/rfcs/0010-ghost-state.md
@@ -1,0 +1,207 @@
+- **Feature Name:** Ghost State (`ghost-state`)
+- **Feature Request Issue:** [#3184](https://github.com/model-checking/kani/issues/3184)
+- **RFC PR:**
+- **Status:** Unstable
+- **Version:** 0
+- **Proof-of-concept:** N/A
+
+-------------------
+
+## Summary
+
+Add support to ghost state to allow users to track metadata useful for verification purpose without extra
+runtime overhead.
+
+## User Impact
+
+TODO
+
+## User Experience
+
+We propose that ghost state to be implemented using a trait which is extended by Kani.
+
+```rust
+/// Trait that indicates a type has an associated ghost code of type `S`.
+///
+/// A type can have many multiple implementations, each will be mapped to a different ghost state.
+pub trait GhostState<S: Default + GhostValue>: Sized + GhostStateExt<S> {}
+
+/// Trait that specifies types that are safe to be used as a ghost value.
+///
+/// # Safety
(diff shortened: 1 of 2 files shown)


rfc/src/rfcs/0010-ghost-state.md
```diff
@@ -0,0 +1,207 @@
...
+which variant.
+
+```rust
+union Size {
+    unsigned: usize,
+    signed: isize,
+}
+
+/// Declare ghost state to track initialization.
+impl kani::GhostState<IsInit> for Size {}
```
remi-delmas-3000, line 63: This will not scale if we want to generalize initialisation analysis to all program state, or do more general taint analysis. CBMC shadow maps is may restricted in size but the nice point is that they are global and the mapping is implicit.
Here it seems like we don't really use that to the max of its possiblities.

rfc/src/rfcs/0010-ghost-state.md
```diff
@@ -0,0 +1,207 @@
...
+    #[doc(hidden)]
+    fn private(&self, _: Internal);
+}
+```
+
+For example, let's say a user wants to check their union have been initialized, together with
+which variant.
+
+```rust
+union Size {
```
remi-delmas-3000, line 57: The name is confusing (wrt to the trait "Sized")

rfc/src/rfcs/0010-ghost-state.md
celinval: Wouldn't that be the same as creating a `impl kani::GhostState<IsInit> for *const u8 {}`? So whenever you want to get the initialization state of a byte, you cast its address to a `*const u8`, and get its ghost state?

celinval: That said, I don't think tracking initialization of every byte is scalable or needed. In Rust, you only need to track initialization of raw pointers and unions. Everything else is guaranteed to be initialized.

No matter the approach, I think my main concern is how to make this map compact enough.

rfc/src/rfcs/0010-ghost-state.md
celinval: Sure. I can change it, this is just a dummy example.

rfc/src/rfcs/0010-ghost-state.md
celinval: I added this concern and potential mitigations to open questions.

celinval:
This is no longer a priority. Closing it for now.

Issue #3095, opened 2024-03-19
Add support to type invariants


celinval:
Requested feature: Users should be able to specify and verify type safety invariants.
Use case: Users often encode safety of a type as a relationship between its internal variables, and Kani should provide a mechanism for users to specify the invariants and verify them.
Link to relevant documentation (Rust reference, Nomicon, RFC): https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html

PR #3190, opened 2024-05-16
(Re)introduce `Invariant` trait

adpaco-aws:
This PR reintroduces the `Invariant` trait as a mechanism for the specification of type safety invariants. The trait is defined in the Kani library where we also provide `Invariant` implementations for primitive types.

In contrast to the previous `Invariant` trait, this version doesn't require the `Arbitrary` bound (i.e., it has the same requirements as `Arbitrary`). This way, the user isn't required to provide an `Arbitrary` implementation in addition to the `Invariant` one.

Related #3095 

### DIFF
--- a/library/kani/src/invariant.rs
+++ b/library/kani/src/invariant.rs
@@ -15,22 +15,6 @@ where
     fn is_safe(&self) -> bool;
 }
 
-impl Invariant for bool {
-    #[inline(always)]
-    fn is_safe(&self) -> bool {
-        let value = *self as u8;
-        value < 2
-    }
-}
-
-impl Invariant for char {
-    #[inline(always)]
-    fn is_safe(&self) -> bool {
-        let value = *self as u32;
-        value <= 0xD7FF || (0xE000..=0x10FFFF).contains(&value)
-    }
-}
-
 /// Any value is considered safe for the type
 macro_rules! trivial_invariant {
     ( $type: ty ) => {
@@ -64,3 +48,5 @@ trivial_invariant!(f32);
 trivial_invariant!(f64);
 
 trivial_invariant!(());
+trivial_invariant!(bool);
+trivial_invariant!(char);
--- a/library/kani/src/invariant.rs
+++ b/library/kani/src/invariant.rs
@@ -1,0 +1,13 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This module introduces the Invariant trait.
+
+/// This trait should be used to specify and check type safety invariants for a
+/// type.
+pub trait Invariant
+where
+    Self: Sized,
+{
+    fn is_valid(&self) -> bool;
+}
--- a/library/kani/src/invariant.rs
+++ b/library/kani/src/invariant.rs
@@ -1,13 +1,66 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! This module introduces the Invariant trait.
+//! This module introduces the `Invariant` trait as well as its implementation
+//! for primitive types.
 
 /// This trait should be used to specify and check type safety invariants for a
 /// type.
+///
+/// Note: Type safety might be checked automatically by Kani in the future.
 pub trait Invariant
 where
     Self: Sized,
 {
     fn is_safe(&self) -> bool;
 }
+
+impl Invariant for bool {
+    #[inline(always)]
+    fn is_safe(&self) -> bool {
+        let value = *self as u8;
+        value < 2
+    }
+}
+
+impl Invariant for char {
+    #[inline(always)]
+    fn is_safe(&self) -> bool {
+        let value = *self as u32;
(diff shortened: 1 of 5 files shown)


library/kani/src/invariant.rs
```diff
@@ -0,0 +1,13 @@
...
+
+//! This module introduces the Invariant trait.
+
+/// This trait should be used to specify and check type safety invariants for a
+/// type.
+pub trait Invariant
+where
+    Self: Sized,
+{
+    fn is_valid(&self) -> bool;
```
celinval, line 12: I would like to suggest `is_safe(&self)`... what do you think?

adpaco-aws: It's a great suggestion since a lot of related contents refers to this as "type safety". Changed to `is_safe`.

adpaco-aws:
The PR is ready for review.

Question: Since we're considering checking automatically these invariants at some level in the future, would it make sense to warn Kani users in the release notes?

library/kani/src/invariant.rs
```diff
@@ -0,0 +1,66 @@
...
+macro_rules! trivial_invariant {
+    ( $type: ty ) => {
+        impl Invariant for $type {
+            #[inline(always)]
+            fn is_safe(&self) -> bool {
+                true
+            }
+        }
+    };
+}
```
celinval, line 44: You don't need this. These are validity invariants and they should be enforced by the compiler.

adpaco-aws: But won't this be an issue when we have automatically-derived implementations of `Invariant` (e.g., for a `struct`) which rely on member types to implement `Invariant`?

celinval: Validity invariant has to be guaranteed at construction time, it's not something you can check after the fact.

Let's say we were to implement a safe `any_raw()` using the trait `Invariant`, it would look like this:
```rust
fn any_raw<T: Invariant>() -> T {
  // First ensure memory is initialized with a valid value.
  let maybe : MaybeUninit<T> = kani::any();
  kani::assume(kani::has_valid_value(maybe.as_ptr()));

  // Now ensure the object safety invariants are respected.
  let val = unsafe { maybe.assume_init() };
  kani::assume(val.is_safe());
  val
}
```

library/kani/src/invariant.rs
```diff
@@ -0,0 +1,66 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This module introduces the `Invariant` trait as well as its implementation
+//! for primitive types.
+
+/// This trait should be used to specify and check type safety invariants for a
+/// type.
+///
+/// Note: Type safety might be checked automatically by Kani in the future.
+pub trait Invariant
```
celinval, line 11: I think it would be great if we can improve this documentation a bit, and maybe include an example. I would also add a link to this definition:
https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#validity-and-safety-invariant

I was also wondering if we should just name this trait `SafetyInvariant`.

tests/kani/Invariant/percentage.rs
```diff
@@ -0,0 +1,33 @@
...
+//! Check that the `Invariant` implementation behaves as expected when used on a
+//! custom type.
+
+extern crate kani;
+use kani::Invariant;
+
+// We use the default `Arbitrary` implementation, which allows values that
+// shouldn't be considered safe for the `Percentage` type.
+#[derive(kani::Arbitrary)]
+struct Percentage(u8);
```
celinval, line 13: Can I suggest we enhance this test a bit? Maybe add a few functions to this type, like `try_new`, and a `value` method to show that users can only violate the safety invariant by using unsafe code.

celinval (approved):
Btw, great documentation. Thanks

tests/kani/Invariant/percentage.rs
adpaco-aws: Done. Also added an `increase` method which uses `try_new` which is demonstrated to be safe w.r.t. the invariant through another harness.

library/kani/src/invariant.rs
```diff
@@ -0,0 +1,97 @@
...
+///    the compiler to perform optimizations.
+///  - *Safety invariant*: An invariant that safe code may assume all data to
+///    uphold. This invariant can be temporarily violated by unsafe code, but
+///    must always be upheld when interfacing with unknown safe code.
+///
+/// Therefore, validity invariants must be upheld at all times, while safety
+/// invariants only need to be upheld at the boundaries to safe code.
+///
+/// Safety invariants are particularly interesting for user-defined types, and
+/// the `Invariant` trait allows you to check them with Kani. For example,
```
feliperodri, line 24: This can also be used in tests. It's a programmatic way to specify (in Rust) properties over your data types. Since it's written in Rust, it can be used for static and dynamic checking.

feliperodri (approved):
This is such a great addition. In the future, we could have something like `safe_any()`, which would return non-deterministic values that automatically assume the safety invariants of a type.

PR #3200, opened 2024-05-25
Add simple API for shadow memory

### LINKED ISSUE #3184: Add support to shadow memory
Requested feature: Add support to shadow memory
Use case: Allow users to extend their code for verification without impacting runtime

Test case:
```rust
union Data {
    word: usize,
    byte: u8,
}

impl Data {
    unsafe fn assume_word(&self) -> usize {
        self.word
    }
    unsafe fn assume_byte(&self) -> u8 {
        self.byte
    }
}

struct Memory {
    data: [Data; 1024],
    word: bool,
}

impl Memory {
    unsafe fn read_word(&self, index: usize) -> usize {
        assert!(self.word);
        unsafe { self.data[index].assume_word() }
    }

    unsafe fn read_byte(&self, index: usize) -> u8 {
        assert!(!self.word);
        unsafe { self.data[index].assume_byte() }
    }

    fn new(word: bool) { /* */}
}

#[kani::proof]
fn check_memory() {
    // info
}
```

Dummy example where users may want to use shadow memory to track how this code is being used. E.g.: Shadow memory could be used to track if the union has been initialized and which type has been used.

zhassan-aws:
Towards #3184 

### DIFF
--- a/library/kani/src/shadow.rs
+++ b/library/kani/src/shadow.rs
@@ -15,22 +15,34 @@ impl ShadowMem {
         Self { is_init: [[false; 64]; 1024] }
     }
 
-    pub fn is_init(&self, ptr: *const u8) -> bool {
-        read(&self.is_init, ptr)
+    /// # Safety
+    /// 
+    /// `ptr` must be valid
+    pub unsafe fn is_init(&self, ptr: *const u8) -> bool {
+        unsafe { read(&self.is_init, ptr) }
     }
 
-    pub fn set_init(&mut self, ptr: *const u8, init: bool) {
-        write(&mut self.is_init, ptr, init);
+    /// # Safety
+    /// 
+    /// `ptr` must be valid
+    pub unsafe fn set_init(&mut self, ptr: *const u8, init: bool) {
+        unsafe { write(&mut self.is_init, ptr, init) };
     }
 }
 
-pub fn read(sm: &[[bool; 64]; 1024], ptr: *const u8) -> bool {
+/// # Safety
+/// 
+/// `ptr` must be valid
+pub unsafe fn read(sm: &[[bool; 64]; 1024], ptr: *const u8) -> bool {
     let obj = unsafe { __KANI_pointer_object(ptr) };
     let offset = unsafe { __KANI_pointer_offset(ptr) };
     sm[obj][offset]
 }
 
-pub fn write(sm: &mut [[bool; 64]; 1024], ptr: *const u8, val: bool) {
+/// # Safety
+/// 
+/// `ptr` must be valid
+pub unsafe fn write(sm: &mut [[bool; 64]; 1024], ptr: *const u8, val: bool) {
     let obj = unsafe { __KANI_pointer_object(ptr) };
     let offset = unsafe { __KANI_pointer_offset(ptr) };
     sm[obj][offset] = val;
--- a/library/kani/src/shadow.rs
+++ b/library/kani/src/shadow.rs
@@ -18,7 +18,7 @@ impl<T: Copy> ShadowMem<T> {
     }
 
     /// Get the shadow memory value of the given pointer
-    pub fn get(&self, ptr: *const u8) -> T {
+    pub fn get<U>(&self, ptr: *const U) -> T {
         let obj = crate::mem::pointer_object(ptr);
         let offset = crate::mem::pointer_offset(ptr);
         crate::assert(obj < MAX_NUM_OBJECTS, MAX_NUM_OBJECTS_ASSERT_MSG);
@@ -27,7 +27,7 @@ impl<T: Copy> ShadowMem<T> {
     }
 
     /// Set the shadow memory value of the given pointer
-    pub fn set(&mut self, ptr: *const u8, val: T) {
+    pub fn set<U>(&mut self, ptr: *const U, val: T) {
         let obj = crate::mem::pointer_object(ptr);
         let offset = crate::mem::pointer_offset(ptr);
         crate::assert(obj < MAX_NUM_OBJECTS, MAX_NUM_OBJECTS_ASSERT_MSG);
--- a/library/kani/src/shadow.rs
+++ b/library/kani/src/shadow.rs
@@ -33,24 +33,32 @@ impl ShadowMem {
     }
 }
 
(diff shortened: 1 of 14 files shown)


library/kani/src/shadow.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+impl ShadowMem {
+    pub const fn new() -> Self {
+        Self { is_init: [[false; MAX_OBJECT_SIZE]; MAX_NUM_OBJECTS] }
+    }
+
+    /// # Safety
+    ///
+    /// `ptr` must be valid
```
celinval, line 23: Why raw pointer? Why does it need to be valid?

library/kani/src/shadow.rs
```diff
@@ -0,0 +1,64 @@
...
+
+const MAX_NUM_OBJECTS: usize = 1024;
+const MAX_OBJECT_SIZE: usize = 64;
+
+pub struct ShadowMem {
+    is_init: [[bool; MAX_OBJECT_SIZE]; MAX_NUM_OBJECTS],
+}
+
+impl ShadowMem {
+    pub const fn new() -> Self {
```
celinval, line 17: can you please add `#[unstable]` attribute to the public functions?

library/kani/src/shadow.rs
```diff
@@ -0,0 +1,64 @@
...
+        Self { is_init: [[false; MAX_OBJECT_SIZE]; MAX_NUM_OBJECTS] }
+    }
+
+    /// # Safety
+    ///
+    /// `ptr` must be valid
+    pub unsafe fn is_init(&self, ptr: *const u8) -> bool {
```
celinval, line 24: Can we use generic `T` instead of u8.

library/kani/src/shadow.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+extern "C" {
+    pub fn __KANI_pointer_object(ptr: *const u8) -> usize;
```
celinval, line 5: I would prefer not adding a new way of implementing Kani intrinsic for now. Can we just use a hook to keep it consistent with other Kani operations?

library/kani/src/shadow.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub fn __KANI_pointer_offset(ptr: *const u8) -> usize;
+}
+
+const MAX_NUM_OBJECTS: usize = 1024;
+const MAX_OBJECT_SIZE: usize = 64;
+
+pub struct ShadowMem {
+    is_init: [[bool; MAX_OBJECT_SIZE]; MAX_NUM_OBJECTS],
```
celinval, line 13: Is this shadow memory specific for tracking initialization? Is there a reason why we force this to be a boolean?

tests/expected/shadow/uninit_array/test.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use std::alloc::{alloc, dealloc, Layout};
+
+static mut SM: kani::shadow::ShadowMem = kani::shadow::ShadowMem::new();
+
+fn write(ptr: *mut i32, offset: usize, x: i32) {
+    unsafe {
+        let p = ptr.offset(offset as isize);
+        p.write(x);
+        SM.set_init(p as *mut u8, true);
```
celinval, line 12: For initialization, this should actually write 4 times to the shadow memory.

Since this is just a sanity check, I would just change the types you are using to `i8` so you don't have to bother.

celinval (commented):
Thank you for doing this! I added a few comments.

library/kani/src/shadow.rs
zhassan-aws: Sure.

library/kani/src/shadow.rs
zhassan-aws: No, I had intended to make this generic. I'll update.

library/kani/src/shadow.rs
zhassan-aws: Makes sense.

library/kani/src/shadow.rs
zhassan-aws: > Why raw pointer?

Do you suggest using a reference instead?

> Why does it need to be valid?

I thought `__CPROVER_POINTER_OBJECT` required a valid pointer, but the [CBMC docs](https://diffblue.github.io/cbmc/cprover-manual/md_memory_primitives.html) indicates it doesn't have any preconditions.

library/kani/src/shadow.rs
zhassan-aws: Sure

tests/expected/shadow/uninit_array/test.rs
zhassan-aws: Indeed. I'll change it.

library/kani/src/shadow.rs
celinval: I think the raw pointer interface makes sense if you want to be able to reason about potentially invalid pointer. The initialization case is definitely one where you do want a raw pointer exactly for that reason.

celinval: That said, I don't think this function is unsafe or require valid pointer. So making this clear is probably the best outcome

zhassan-aws: Removed `unsafe` and validity requirement.

library/kani/src/shadow.rs
zhassan-aws: Done.

library/kani/src/shadow.rs
zhassan-aws: Done.

tests/expected/shadow/uninit_array/test.rs
zhassan-aws: Done.

library/kani/src/shadow.rs
zhassan-aws: Done.

library/kani/src/shadow.rs
zhassan-aws: Done.

celinval (approved):
Awesome, thanks

PR #3205, opened 2024-05-28
Fix `{,e}println!()`

### LINKED ISSUE #3204: `println!()` with no trailing semicolon causes an error
I tried this code:
```rust
fn main() {
    println!()
}
```

```shell
kani test.rs
```

with Kani version: 0.51.0

I expected to see this happen: all is ok

Instead, this happened:
```
error: macro expansion ends with an incomplete expression: expected expression
 --> test.rs:2:5
  |
2 |     println!()
  |     ^^^^^^^^^^ expected expression
  |
  = note: the macro call doesn't expand to an expression, but it can expand to a statement
help: add `;` to interpret the expansion as a statement
  |
2 |     println!();
  |               +

```

GrigorenkoPV:
Change the implementation of `println!` & `eprintln!` with no arguments to call `print!("\n")`
 & `eprint!("\n")` respectively instead of producing no tokens. This is what std does. [^println][^eprintln]
 
[^println]: https://github.com/rust-lang/rust/blob/8c127df75fde3d5ad8ef9af664962a7676288b52/library/std/src/macros.rs#L140
[^eprintln]: https://github.com/rust-lang/rust/blob/8c127df75fde3d5ad8ef9af664962a7676288b52/library/std/src/macros.rs#L218

Probably resolves #320. Haven't tested, sorry. It might also make sense to add a test for this.

### DIFF
--- a/library/std/src/lib.rs
+++ b/library/std/src/lib.rs
@@ -135,14 +135,14 @@ macro_rules! eprint {
 #[cfg(not(feature = "concrete_playback"))]
 #[macro_export]
 macro_rules! println {
-    () => { };
+    () => { $crate::print!("\n") };
     ($($x:tt)*) => {{ let _ = format_args!($($x)*); }};
 }
 
 #[cfg(not(feature = "concrete_playback"))]
 #[macro_export]
 macro_rules! eprintln {
-    () => { };
+    () => { $crate::eprint!("\n") };
     ($($x:tt)*) => {{ let _ = format_args!($($x)*); }};
 }
 


zhassan-aws:
Thanks for the contribution @GrigorenkoPV!

> It might also make sense to add a test for this.

Can you please add one? It can be placed under `tests/kani/Print`.

Thank you!

zhassan-aws:
> Probably resolves https://github.com/model-checking/kani/pull/320.

I think you meant #3204

PR #3221, opened 2024-06-03
Inject pointer validity check when casting raw pointers to references

### LINKED ISSUE #2975: Check for pointer validity when casting raw pointer to reference.
Requested feature: Kani should automatically check if a raw pointer is valid before any conversion to a reference, strengthening the verification of unsafe code.
Use case: This will help enable modular verification of safe abstractions. Kani detects invalid memory access only when the reference or dangling pointer is used.

Test case:
```rust
struct Store<'a, const LEN: usize> {
    data: [&'a i128; LEN],
}

impl<'a, const LEN: usize> Store<'a, LEN> {
    pub fn from(var: &i64) -> Self {
        let ref1: *const i64 = var;
        let ref2: *const i128 = ref1 as *const i128;
        unsafe {
            Store { data: [&*ref2; LEN] }  // ---- THIS LINE SHOULD FAIL
        }
    }
}

#[cfg_attr(not(kani), test)]
pub fn check_store() {
    let val = 1;
    let broken = Store::<3>::from(&val);
    assert_eq!(broken.data.len(), 3)
}

```

Note: The harness above succeeds under playground run (version 1.75.0), and with Kani, but it correctly fails under MIRI.

artemagvanian:
Resolves #2975

This PR makes Kani able to check if a raw pointer is valid before casting it to a reference.

To check for the pointer validity when casting it to a reference, inject asserting `__CPROVER_r_ok(ptr, sz)` into places where a raw pointer is dereferenced and a reference is immediately taken.

Since this check seems to cause some of the CIs to run out of memory, it is only enabled under `-Z ptr-to-ref-cast-checks` flag. 


artemagvanian:
That makes sense given the fact that the harness was already close to running OOM. I disabled the test by changing the `expected` file to take into account its failure. Not sure I know of a better way to do it, but happy to learn.

artemagvanian:
Instead of changing the `expected` file, I added the copies of the files without the problematic harness into the `overlay` directory, so the harness will not run on the CI.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -323,4 +324,63 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    let raw_ptr_read_ok_expr = Expr::read_ok(
+                        ptr.cast_to(Type::void_pointer()),
+                        Expr::int_constant(pointee_size, Type::size_t()),
+                    )
+                    .cast_to(Type::Bool);
+                    // Finally, assert that the pointer points to a valid memory location.
+                    let raw_ptr_read_ok = self.codegen_assert(
+                        raw_ptr_read_ok_expr,
+                        PropertyClass::SafetyCheck,
+                        "dereferencing a pointer to invalid memory location",
```
zhassan-aws, line 377: To be consistent with the message from CBMC checks, we should use "dereference failure: pointer invalid"

tests/expected/ptr_to_ref_cast/expected
```diff
@@ -0,0 +1,75 @@
+Checking harness check_zst_deref...
+
+Status: FAILURE
```
zhassan-aws, line 3: Adding a backslash at the end, i.e.
```suggestion
Status: FAILURE\
```
ensures that the script checks that the next line in the expected file appears after the next line in the output. This is necessary to make sure that the failure is for that specific check.

tests/expected/ptr_to_ref_cast/expected
```diff
@@ -0,0 +1,75 @@
...
+Description: "dereferencing a pointer to invalid memory location"
+Location: tests/expected/ptr_to_ref_cast/ptr_to_ref_cast.rs:131:35 in function zst_deref
```
zhassan-aws, line 5: We tend to avoid having line numbers in expected files as they are fragile, e.g. due to changing the file preamble. You can keep the function name though, e.g.
```suggestion
in function zst_deref
```
and add a backslash at the end of the previous line.

tests/perf/overlays/s2n-quic/quic/s2n-quic-core/src/inet/checksum.rs
```diff
@@ -0,0 +1,482 @@
+// Copyright Kani Contributors
```
zhassan-aws, line 1: It's problematic to turn off those tests in our regressions because as soon as a new version of Kani is released, their CI is going to break.

@tautschnig AFAIK, the stabilization plan for this feature includes reducing only making this check if necessary. Thus, I think we should put it under a flag for the time being because the flag is eventually going to be removed once the feature is stabilized.

zhassan-aws (changes requested):
Great work, @artemagvanian!

We need to avoid overriding the files in s2n-quic, as the regression is meant to ensure we don't break our customers' CI.

tests/perf/overlays/s2n-quic/quic/s2n-quic-core/src/inet/checksum.rs
artemagvanian: @zhassan-aws I have put the check behind the unstable flag for now

zhassan-aws (approved):

Issue #2909, opened 2023-12-01
Diverging `modifies` clauses cause unsound vacuity


JustusAdam:
Diverging (panicking) expressions, such as indexing, in a `modifies` clause lead to vacuous verification.

celinval:
Can you please clarify what you mean here? Maybe an example with the current output would be helpful. Will the overall verification succeed?

jsalzbergedu:
An example of code that verifies vacuously, found with the help of @pi314mm :
```
// Copyright Kani Contributors
// SPDX-License-Identifier: Apache-2.0 OR MIT
// kani-flags: -Zfunction-contracts

// Test that a modifies clause works when a (function call)
// expression is provided
fn infinite_looper() -> usize {
    loop {};
    1
}

#[kani::requires(arr[idx] == 1)]
#[kani::modifies(&arr[infinite_looper()])]
#[kani::ensures(arr[idx] == 2)]
fn modify(idx: usize, arr: &mut [i64]) {
    arr[idx] += 1;
}

#[kani::proof_for_contract(modify)]
fn main() {
    let mut arr: [i64; 4] = [1, 2, 3, 4];
    modify(1, &mut arr);
}
```

zhassan-aws:
This can be reproduced without `modifies` clauses as well, e.g.:
```rust
#[kani::ensures(result == 1)]
fn foo() -> i32 {
    loop {};
    2
}

#[kani::proof_for_contract(foo)]
fn check_foo() {
    let _ = foo();
}
```

```
$ kani contract_loop.rs -Zfunction-contracts
Kani Rust Verifier 0.51.0 (standalone)
warning: unreachable expression
 --> contract_loop.rs:4:5
  |
3 |     loop {};
  |     ------- any code following this expression is unreachable
4 |     2
  |     ^ unreachable expression
  |
  = note: `#[warn(unreachable_code)]` on by default

warning: 1 warning emitted

warning: unreachable expression
 --> contract_loop.rs:4:5
  |
3 |     loop {};
  |     ------- any code following this expression is unreachable
4 |     2
  |     ^ unreachable expression
  |
  = note: `#[warn(unreachable_code)]` on by default

warning: Found the following unsupported constructs:
             - caller_location (1)
             - foreign function (2)
         
         Verification will fail if one or more of these constructs is reachable.
         See https://model-checking.github.io/kani/rust-feature-support.html for more details.

warning: 2 warnings emitted

Checking harness check_foo...
[...]

PR #3223, opened 2024-06-04
Do not turn trivially diverging loops into assume(false)

tautschnig:
CBMC's symbolic execution by default turns `while(true) {}` loops into `assume(false)` to counter trivial non-termination of symbolic execution. When unwinding assertions are enabled, however, we should report the non-termination of such loops.

Resolves: #2909

### DIFF
--- a/kani-driver/src/call_cbmc.rs
+++ b/kani-driver/src/call_cbmc.rs
@@ -183,6 +183,7 @@ impl KaniSession {
 
         if self.args.checks.unwinding_on() {
             args.push("--unwinding-assertions".into());
+            args.push("--no-self-loops-to-assumptions".into());
         }
 
         if self.args.extra_pointer_checks {
--- a/kani-driver/src/call_cbmc.rs
+++ b/kani-driver/src/call_cbmc.rs
@@ -182,6 +182,7 @@ impl KaniSession {
         }
 
         if self.args.checks.unwinding_on() {
+            // TODO: With CBMC v6 the below can be removed as those are defaults.
             args.push("--unwinding-assertions".into());
             args.push("--no-self-loops-to-assumptions".into());
         }
--- a/tests/expected/function-contract/diverging_loop.expected
+++ b/tests/expected/function-contract/diverging_loop.expected
@@ -1,0 +1,3 @@
+Failed Checks: unwinding assertion loop 0
+
+VERIFICATION:- FAILED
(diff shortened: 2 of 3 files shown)


zhassan-aws:
Thanks @tautschnig for the quick fix! Should we consider this a soundness issue in CBMC and fix it on the CBMC side instead? i.e., when unwinding assertions are enabled, self loops should not be replaced with `assume(false)`?

tautschnig:
> Thanks @tautschnig for the quick fix! Should we consider this a soundness issue in CBMC and fix it on the CBMC side instead? i.e., when unwinding assertions are enabled, self loops should not be replaced with `assume(false)`?

I think we should indeed seek to make this consistent on the CBMC side for version 6. I'll come up with a PR.

qinheping:
Yes, when applying loop contracts, lack of decreases may lead to vacuous verification, which is the case of Zyad's comment. Now the synthesizer for Kani doesn't synthesize decrease clauses, and should does before it became stable.

tautschnig:
> > > Thanks @tautschnig for the quick fix! Should we consider this a soundness issue in CBMC and fix it on the CBMC side instead? i.e., when unwinding assertions are enabled, self loops should not be replaced with `assume(false)`?
> > 
> > 
> > I think we should indeed seek to make this consistent on the CBMC side for version 6. I'll come up with a PR.
> 
> @tautschnig if we fix this on version 6, we do not need to merge this PR, correct? Shall we close it in favor of the new update?

Merging this PR would help disentangle a Kani issue from the CBMC release. Once CBMC v6 is released and Kani starts using it we can then remove this, but then we will anyway end up removing several options from the command line being used to invoke CBMC.

kani-driver/src/call_cbmc.rs
```diff
@@ -183,6 +183,7 @@ impl KaniSession {
 
         if self.args.checks.unwinding_on() {
             args.push("--unwinding-assertions".into());
+            args.push("--no-self-loops-to-assumptions".into());
```
feliperodri, line 186: Should we add a comment here saying this must be updated once we move to CBMC 6.0?

feliperodri (approved):

kani-driver/src/call_cbmc.rs
zhassan-aws: +1

zhassan-aws (approved):

PR #3225, opened 2024-06-04
Update the rust toolchain to nightly-2024-06-11

tautschnig:
Changes required due to:
- rust-lang/rust@a34c26e7ec Make body_owned_by return the body directly.
- rust-lang/rust@333458c2cb Uplift TypeRelation and Relate
- rust-lang/rust@459ce3f6bb Add an intrinsic for `ptr::metadata`
- rust-lang/rust@7e08f80b34 Split smir `Const` into `TyConst` and `MirConst`
- rust-lang/rust@eb584a23bf offset_of: allow (unstably) taking the offset of slice tail fields
- rust-lang/rust@16e8803579 Update cargo

Resolves: #3218


feliperodri (approved):

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -814,6 +815,51 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                    } else if let Some(len_typ) =
+                        src_goto_expr.typ().lookup_field_type("len", &self.symbol_table)
+                    {
+                        assert_eq!(len_typ, dst_goto_typ);
+                        src_goto_expr.member("len", &self.symbol_table)
+                    } else {
+                        unreachable!(
+                            "fat pointer with neither vtable nor len: {:?}",
+                            src_goto_expr
+                        );
```
celinval, line 860: Can this intrinsic be called with a thin pointer though? I think in that case it should just return a ZST of the given type.

celinval (commented):
Looks good! Can you please also add tests for `ptr_metadata` for the other types of pointers? Like `&dyn T` and `&char`? Thanks!

celinval (changes requested):
I think the thin pointer case needs to be handled. Please double check

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Tests for `&dyn T` and `&char` added. What would a test with a thin pointer look like?

celinval: `&char` is a thin pointer.

kani-compiler/src/kani_middle/transform/contracts.rs
```diff
@@ -11,7 +11,7 @@ use rustc_middle::ty::TyCtxt;
 use rustc_smir::rustc_internal;
 use stable_mir::mir::mono::Instance;
 use stable_mir::mir::{Body, Constant, Operand, TerminatorKind};
-use stable_mir::ty::{Const as MirConst, FnDef, RigidTy, TyKind};
+use stable_mir::ty::{FnDef, MirConst, RigidTy, TyKind};
```
celinval, line 14: You're welcome... 😝

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
tautschnig: Looks like we are already doing the right thing then - the extended test passes just fine.

celinval:
Ah... That's because we skip statements that generate zst values. It does make me wonder if that's appropriate. But that's a different problem. Thanks @tautschnig

celinval (approved):

Issue #3226, opened 2024-06-04
Add an option to Kani to verify a custom version of the standard library.


celinval:
Create an unstable sub-command that users can invoke to verify a local version of the standard library.

Something like:
```
kani verify-std <path-std>
```

Issue #3153, opened 2024-04-19
Create a `#[no_core]` Kani library


celinval:
In order to support annotations in a `#[no_core]` environment, and subsequently, `#[no_std]`, we should consider creating a version of Kani library that has no dependency.

PR #3227, opened 2024-06-04
Add `kani_core` library placeholder and build logic

celinval:
While at it, I also added a `--skip-libs` to skip rebuilding the Kani libraries and standard library at every `cargo build-dev` execution.

We usually only need to rebuild the libraries when we make changes to them or when we update the Rust toolchain. Rebuilding them can be quite time consuming when you are making changes to Kani.

Towards #3226 #3153


jaisnan (approved):
LGTM, thank you for this.

PR #3229, opened 2024-06-05
Contracts: History Expressions via a new function contract "remembers"

### LINKED ISSUE #2803: History Expression Support for Function Contracts

pi314mm:
This adds a new contract known as "remember" to handle history variables. The remember expression occurs before the computation of the function, so it can be used to store arguments from the past to be used later. For example, the following code works:

```rust
#[kani::remember(ptr.clone())]
#[kani::ensures(|result, old_ptr : &NoCopy<u32>| (*old_ptr).0 + 1 == ptr.0)]
#[kani::remember(true)]
#[kani::ensures(|result, old_ptr : &NoCopy<u32>, t : &bool| (*old_ptr).0 + 1 == ptr.0 && *t)]
#[kani::requires(ptr.0 < 100)]
#[kani::modifies(&mut ptr.0)]
fn modify(ptr: &mut NoCopy<u32>) {
    ptr.0 += 1;
}
```

Note that the first ensures statement has one argument to the second closure while the second ensures statement has two arguments. This is because the second remember occurred in between the first and second ensures statements. This functionality is intended as we can think of the remembers as adding something new to our memory as we go down line by line in the contracts listed.

Resolves #2803


pi314mm:
Closing in favor for the #3232 implementation instead of using remembers contract

PR #3230, opened 2024-06-05
Add intrinsics and Arbitrary support for no_core

jaisnan:
### Approach So far

The approach so far has been to shift the `kani` library's functionality to `kani_core` instead, which has no dependencies of its own. By moving these API functions to macros, we delay the compilation of these to when Kani is invoked.

Tested by using the regression tests added @celinval  in https://github.com/model-checking/kani/pull/3236 using the new `verify-std` subcommand.

### Running Kani

To test Kani itself, we injected a proof inside the core library by making these changes.

```  rust
#[cfg(kani)]
kani_core::kani_lib!(core);

#[cfg(kani)]
#[unstable(feature = "kani", issue = "none")]
pub mod verify {
    use crate::kani;

    #[kani::proof]
    pub fn harness() {
        kani::assert(true, "yay");
    }
}
```


scripts/std-lib-regression.sh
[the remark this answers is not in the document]
jaisnan: Not really, reverting changes to this file.

library/kani/Cargo.toml
```diff
@@ -9,7 +9,7 @@ license = "MIT OR Apache-2.0"
 publish = false
 
 [dependencies]
-kani_macros = { path = "../kani_macros" }
+kani_macros = { path = "../kani_macros", default-features = false }
```
celinval, line 12: Why do we need this?

celinval (approved):
Awesome! Thanks

tests/script-based-pre/verify_std_cmd/verify_std.sh
celinval, line 1: Actually, can you please change this code to invoke the `kani_lib` macro you just created? Thanks!

PR #3231, opened 2024-06-05
Add a new verify-std subcommand to Kani

### LINKED ISSUE #3226: Add an option to Kani to verify a custom version of the standard library.

celinval:
This subcommand will take the path to the standard library.
   
It will then use `cargo build -Z build-std` to build the custom standard
library and verify any harness found during the build.

## Call out

- So far I only performed manual tests. I'm going to add a few unit tests and a script in the next revision.

Resolves #3226 

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
@@ -458,6 +458,45 @@ impl GotocHook for UntrackedDeref {
     }
 }
 
+struct InitContracts;
+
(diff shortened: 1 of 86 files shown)


jaisnan (approved):
Approving for now, thanks for the change. I'm hoping the tests and scripts are going to come next?

celinval:
@jaisnan do you prefer a follow up PR?

jaisnan:
> @jaisnan do you prefer a follow up PR?

I do in this case, but only for the sake of speed here.

Issue #2747, opened 2023-09-07
Comprehensive Documentation of Contract Usage


JustusAdam:
There should be an example driven documentation of how contracts are to be used that we can link to from e.g. the `requires`, `ensures`, `stub_verified`, `proof_for_contract` attribute documentation.

pi314mm:
Also documentation for `old` for history expressions

PR #3232, opened 2024-06-06
Contracts: History Expressions via "old" monad

### LINKED ISSUE #2803: History Expression Support for Function Contracts

pi314mm:
Add support for History Expressions. This allows an ensures clause to evaluate expressions before the function runs, bing them to anonymous variables and use them after calculating the result to be able to compare to the state before the function evaluation.

Resolves #2803


library/kani_macros/src/sysroot/contracts/shared.rs
```diff
@@ -144,3 +147,151 @@ pub fn try_as_result_assign(stmt: &syn::Stmt) -> Option<&syn::LocalInit> {
...
+                    subpat: _,
+                }) => ident.to_string().starts_with("remember_kani_internal_"),
+                _ => false,
+            },
+            _ => false,
+        })
+        .count()
+}
+
+pub fn build_ensures(data: &ExprClosure, remember_count: usize) -> (TokenStream2, Expr) {
```
celinval, line 171: please doc what this is doing... I find that giving an example of the expansion is very helpful.

tests/expected/function-contract/history/noclone_fail.expected
```diff
@@ -0,0 +1,4 @@
+error[E0382]: borrow of moved value: `ptr_renamed`
```
celinval, line 1: Humm... maybe name this something like `kani_old_<var_name>`, so it's a bit more intuitive to the user what's going on.

Another possibility is to create a dummy function:
```rust
fn remember<T: Copy>(val: T) -> T {
  val
}
```

and change the generated code to be:
```rust
let ptr_renamed = kani::internal::remember(ptr);
```

This way, the error will be that `ptr` does not implement Copy trait.

tests/expected/function-contract/history/old_old.expected
```diff
@@ -0,0 +1,2 @@
+error: Nested calls to `old` are prohibited
```
celinval, line 1: that's a nice error

celinval (commented):
I didn't do a thorough review, but so far so good. Thanks!

tests/expected/function-contract/history/noclone_fail.expected
pi314mm: The naming is actually because of the renaming of variables within the ensures clause
For this particular example,
```rust
#[kani::ensures(|result| old(ptr).0 + 1 == ptr.0)]
```
becomes
```rust
let ptr_renamed = kani::internal::untracked_deref(&ptr);
let result_kani_internal = function_call(ptr);
kani::assert((|result| old(ptr_renamed).0 + 1 == ptr_renamed.0)(result_kani_internal))
std::mem::forget(ptr_renamed)
```
So the "_renamed" comes from the renaming the ensures statement which is something tangential to this PR, but a plausible solution is provided in #3234. 

And then what this PR is implementing is lifting "old" up to the top
```rust
let ptr_renamed = kani::internal::untracked_deref(&ptr);
let remember_kani_internal_0 = ptr_renamed;
let result_kani_internal = function_call(ptr);
kani::assert((|result| remember_kani_internal_0.0 + 1 == ptr_renamed.0)(result_kani_internal))
std::mem::forget(ptr_renamed)
```
This specific error statement comes from the fact that ownership of `ptr_renamed` gets given to the `remember_kani_internal`, but then we try to access it again via the assert statement which causes the borrow of moved value error at the assert.

This error could be solved by making yet another copy of the `let ptr_for_remember = kani::internal::untracked_deref(&ptr);` which essentially functions as contraction within rust affine's logic. This would allow us to remember non-copyable/non-cloneable data, because the move into the `remember_kani_internal_0` wouldn't clash with borrows within the ensures statement later.

tests/expected/function-contract/history/old_old.expected
pi314mm: credit for that goes to #2993

library/kani_macros/src/sysroot/contracts/shared.rs
pi314mm: I now added various documentation in shared.rs and library/kani/src/contracts.rs, and added an example of the expansion within mod.rs

pi314mm:
@JustusAdam could you please review this change as it is closely related (same `old` visitor) to your previous implementation #2993?

library/kani_macros/src/sysroot/contracts/shared.rs
```diff
@@ -144,3 +147,171 @@ pub fn try_as_result_assign(stmt: &syn::Stmt) -> Option<&syn::LocalInit> {
 pub fn try_as_result_assign_mut(stmt: &mut syn::Stmt) -> Option<&mut syn::LocalInit> {
     try_as_result_assign_pat!(stmt, as_mut)
 }
+
+/// This function goes through the vector of statements and counts the number of variables
+/// with idents prefixed by `remember_kani_internal_`. This avoid variable name collisions by
```
JustusAdam, line 152: Are these remember variables ever exposed to the user? If not I would prefer to use a short hash of the entire function body as a prefix or suffix rather than a count.

tests/expected/function-contract/history/block.expected
```diff
@@ -0,0 +1 @@
+VERIFICATION:- SUCCESSFUL
```
JustusAdam, line 1: These should generally also list the important (i.e. specifically relevant to the particular test case) as expected. The reason is that then we also ensure that the test case is not vacuous.

tests/expected/function-contract/history/noclone_pass.rs
JustusAdam, line 1: I would prefer `copy_pass` as a name here

library/kani/src/contracts.rs
```diff
@@ -225,4 +225,27 @@
 //! Rust pointer type (`&T`, `&mut T`, `*const T` or `*mut T`). In addition `T`
 //! must implement [`Arbitrary`](super::Arbitrary). This is used to assign
 //! `kani::any()` to the location when the function is used in a `stub_verified`.
+//!
+//! ## History Variables
+//!
+//! Additionally, an ensures clause is allowed to refer to history computations
+//! via an `old` monad. Any instance of `old(computation)` will evaluate the
+//! computation before the function is called. It is required that this computation
+//! is effect free and closed with respect to the function arguments.
```
JustusAdam, line 234: A very important point here is to mention whether the absence of side effects here is _required_ or whether it is _checked_. I believe the latter is the case, because CBMC will deny any side-effect use in contracts. 

Might even warrant a test case.

library/kani/src/contracts.rs
```diff
@@ -225,4 +225,27 @@
 //! Rust pointer type (`&T`, `&mut T`, `*const T` or `*mut T`). In addition `T`
 //! must implement [`Arbitrary`](super::Arbitrary). This is used to assign
 //! `kani::any()` to the location when the function is used in a `stub_verified`.
+//!
+//! ## History Variables
```
JustusAdam, line 229:
```suggestion
//! ## History Expressions
```

library/kani/src/contracts.rs
```diff
@@ -225,4 +225,27 @@
...
+//!
+//! Additionally, an ensures clause is allowed to refer to history computations
```
JustusAdam, line 231: "history computations" is not defined or a well known term. I would suggest 

 ```suggestion
//! Additionally, an ensures clause is allowed to refer to the state of the function arguments before function execution and perform simple computations on them
```

**Review by JustusAdam — CHANGES_REQUESTED** [review]
Thanks for your contribution. I recognize some of the code here 😄 

I'm a bit confused as to the terminology of an "old _monad_". To me this terminology suggests that there is some way to produce values in a category representing the state before the function execution (lets call this category O) as opposed to some "current" category (lets call it T). I would then expect some way to apply transformations to values in O and some way to extract O -> T in a specific context. For example I would expect I could write code similar to this 

```rs
fn check_args(x: u32, y: u32) -> O<bool> {
     let x0 : O<u32> = old(x);
     x0.and_then(|x1| O::pure(x1 + y < 9))
}

// Implicitly in the `O` monad
#[ensures(check_args(x, y))]
fn foo(x: u32, y: u32) { ... }
```

I guess what I'm trying to say is I don't see what the natural transformations of this monad would be. 

Aside: One more reason this is likely not a monad, nesting in a monad is always well defined, whereas we deny expressions of the form `old(old(...))`.

pi314mm:
Thank you for the questions about monads! The old monad links the "language of the past" to the "language of the present" so it relates to your explanation via categories.

It is important to understand there is an underlying theory and the theory is related to but not fully implemented by this particular implementation. While we are trying to express the "old monad," implementing the full generality of the monad is rather difficult, so we focus on a particular usage of the monad.

In particular, if we have an ensures statement like
```rust
#[kani::ensures(old(*ptr)+1==*ptr)]
```
Then we comprehend this as syntax for the statement (not within rust)
```
bind (*ptr : O(u32)) (|remember : u32| remember + 1 == *ptr)
```
Which then gets compiled to (within rust)
```rust
let remember = *ptr;
let result = ...;
kani::assert(remember + 1 == *ptr)
```
This means that the underlying principle of the monad is there, but external syntax appears to be less like a monad because otherwise it would be too difficult to implement, and the user most likely only cares about this particular construction of prefixing all the `bind` operations.

As for nested old, such as `old(old(*ptr)+*ptr)`, it is reasonable to interpret this as compiling to
```rust
let remember_1 = *ptr;
let remember_0 = remember_1 + *ptr;
let result = ...;
...
```
so the restriction is just a matter of there not being implementation support for this kind of statement rather than the theory itself. It is not particularly useful to implement this because we claim that there should be no effectful computation within the contracts, so you can substitute the `remember_1` into the second line without worrying about the effects. Hence, we opt for simply restricting this behavior instead of implementing it.

library/kani/src/contracts.rs
pi314mm: Required and not enforced by kani. Added a new test case in that fails due to side effects within the old computation. You may still be right for other kani contracts, but it isn't checked for old.

tests/expected/function-contract/history/noclone_fail.expected
pi314mm: decided to do a shuffling around of code so that it uses the name `ptr` which is tracked by rust and gives a nicer error message

tests/expected/function-contract/history/block.expected
```diff
@@ -0,0 +1,5 @@
+assertion\
+	 - Status: SUCCESS\
+	 - Description: "|result| old({ let x = &ptr; let y = **x; y + 1 }) == *ptr"\
+
+VERIFICATION:- SUCCESSFUL
```
feliperodri, line 5: Add new line.

tests/expected/function-contract/history/function_call.expected
```diff
@@ -0,0 +1,5 @@
+assertion\
+	 - Status: SUCCESS\
+	 - Description: "|result| old(add1(dereference(ptr))) == *ptr"\
+
+VERIFICATION:- SUCCESSFUL
```
feliperodri, line 5: Add new line.

tests/expected/function-contract/history/copy_pass.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
```
feliperodri, line 3: It would be nice if test cases had a little description about what is intended behavior to be tested here. Sometimes the test description contains an error message that is very obvious, which might ultimately making this extra description unnecessary. So, you can use your judgment where to add it.

tests/expected/function-contract/history/old_old.expected
```diff
@@ -0,0 +1,2 @@
...
+#[kani::ensures(|result| old(*ptr + old(1)) == *ptr)]
```
feliperodri, line 2: Add new line.

tests/expected/function-contract/history/side_effect.expected
```diff
@@ -0,0 +1,2 @@
+Failed Checks: |result| old({ *ptr+=1; *ptr }) == _val
+VERIFICATION:- FAILED
```
feliperodri, line 2: Add new line.

tests/expected/function-contract/history/simple_fail.expected
```diff
@@ -0,0 +1,7 @@
+assertion\
+- Status: FAILURE\
+- Description: "|result| old(*ptr) == *ptr"
+
+Failed Checks: |result| old(*ptr) == *ptr
+
+VERIFICATION:- FAILED
```
feliperodri, line 7: Add new line.

tests/expected/function-contract/history/simple_pass.expected
```diff
@@ -0,0 +1,5 @@
+assertion\
+- Status: SUCCESS\
+- Description: "|result| old(*ptr + 1) == *ptr"\
+
+VERIFICATION:- SUCCESSFUL
```
feliperodri, line 5: Add new line.

tests/expected/function-contract/history/simple_fail.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
+
+#[kani::ensures(|result| old(*ptr) == *ptr)]
```
feliperodri, line 5: We can replace `|result|` with simply `||` if we don't even have a return value. Maybe it makes the case cleaner.

tests/expected/function-contract/history/simple_pass_twice.expected
```diff
@@ -0,0 +1,9 @@
+assertion\
+- Status: SUCCESS\
+- Description: "|result| old(*ptr + 1) == *ptr"\
+
+assertion\
+- Status: SUCCESS\
+- Description: "|result| old(*ptr + 1) == *ptr"\
+
+VERIFICATION:- SUCCESSFUL
```
feliperodri, line 9: Add new line.

tests/expected/function-contract/history/simple_pass_twice.rs
```diff
@@ -0,0 +1,17 @@
...
+
+#[kani::ensures(|result| old(*ptr + 1) == *ptr)]
+#[kani::ensures(|result| old(*ptr + 1) == *ptr)]
+#[kani::requires(*ptr < 100)]
+#[kani::modifies(ptr)]
+fn modify(ptr: &mut u32) {
+    *ptr += 1;
+}
+
+#[kani::proof_for_contract(modify)]
```
feliperodri, line 13: What about test cases that use `stub_verified`? These are important because `old` can help us to write more precise ensure clauses, which would restrict more the value havoced during replacement. Could you add some of those as well? Also, what happen if I use old without an modifies clause? Would this lead to an error?

library/kani/src/contracts.rs
```diff
@@ -225,4 +225,27 @@
...
+//! #[kani::ensures(|result| old(*a).wrapping_add(1) == *a)]
+//! #[kani::ensures(|result : &u32| old(*a).wrapping_add(1) == *result)]
+//! fn add1(a : &mut u32) -> u32 {
+//!     *a=a.wrapping_add(1);
+//!     *a
+//! }
+//! ```
+//!
+//! Here, the value stored in `a` is precomputed and remembered after the function
+//! is called, even though the contents of `a` changed during the function execution.
```
feliperodri, line 249: Could you also update our function contracts [RFC](https://github.com/model-checking/kani/blob/main/rfc/src/rfcs/0009-function-contracts.md#history-expressions) to reflect this new implementation? Let's also add a comment in the issue #2747 to remember to include a section in our documentation about history variables with examples.

library/kani_macros/src/sysroot/contracts/helpers.rs
```diff
@@ -268,3 +268,21 @@ pub fn short_hash_of_token_stream(stream: &proc_macro::TokenStream) -> u64 {
...
+    ($condition:expr, $span_source:expr, $msg:expr $(,)?) => {
+        if !$condition {
+            $span_source.span().unwrap().error($msg).emit();
+            assert!(false);
+        }
+    };
+    ($condition:expr, $span_source:expr) => {
+        assert_spanned_err!($condition, $span_source, concat!("Failed assertion ", stringify!($condition)))
+    };
+}
```
feliperodri, line 288: Add a short comment/documentation to this macro rule. Take a look at this [example](https://github.com/model-checking/kani/blob/9ee1d296de22f363b399df7a38d56c5796bfed1b/library/kani/src/tuple.rs#L13).

library/kani_macros/src/sysroot/contracts/mod.rs
```diff
@@ -233,15 +233,92 @@
...
 //! }
 //! ```
+//!
+//! Additionally, there is functionality that allows the referencing of
+//! history values within the ensures statement. This means we can
+//! precompute a value before the function is called and have access to
+//! this value in the later ensures statement. This is done via the
+//! `old` monad which lets you access the old state within the present
+//! state. Each occurance of old is lifted, so is is necessary that
+//! each lifted occurance is closed with respect to the function arguments.
```
feliperodri, line 243:
```suggestion
//! state. Each occurrence of `old` is lifted, so is is necessary that
//! each lifted occurrence is closed with respect to the function arguments.
```

library/kani_macros/src/sysroot/contracts/shared.rs
feliperodri: @JustusAdam Why would that be preferable in this case?

library/kani_macros/src/sysroot/contracts/shared.rs
```diff
@@ -176,3 +179,246 @@ pub fn try_as_result_assign(stmt: &syn::Stmt) -> Option<&syn::LocalInit> {
...
+
+/// When a #[kani::ensures(|result|expr)] is expanded, this function is called on
+/// with build_ensures(|result|expr, remember_count) where remember_count is the total number of
+/// remember_kani_internal_ variables that exist before building this ensures statement.
+/// This function goes through the expr and extracts out all the `old` expressions and creates a sequence
+/// of statements that instantiate these expressions as `let remember_kani_internal_x = old_expr;` with
+/// `x` starting t the `remember_count` and incrementing from there. This is returned as the first return
+/// parameter along with changing all the variables to _renamed. The second parameter is the closing of
+/// all the unsafe argument copies. The third return parameter is the expression formed by passing in the
+/// result variable into the input closure and changing all the variables to _renamed.
```
feliperodri, line 214:
```suggestion
/// When a `#[kani::ensures(|result|expr)]` is expanded, this function is called on
/// with `build_ensures(|result|expr, remember_count)` where `remember_count` is the total number of
/// `remember_kani_internal_` variables that exist before building this ensures statement.
/// This function goes through the expr and extracts out all the `old` expressions and creates a sequence
/// of statements that instantiate these expressions as `let remember_kani_internal_x = old_expr;` with
/// `x` starting the `remember_count` and incrementing from there. This is returned as the first return
/// parameter along with changing all the variables to `_renamed`. The second parameter is the closing of
/// all the unsafe argument copies. The third return parameter is the expression formed by passing in the
/// result variable into the input closure and changing all the variables to `_renamed`.
```

library/kani_macros/src/sysroot/contracts/shared.rs
```diff
@@ -176,3 +179,246 @@ pub fn try_as_result_assign(stmt: &syn::Stmt) -> Option<&syn::LocalInit> {
...
+    /// indicates whether the entire `old()` call should be replaced by the
+    /// (potentially altered) first argument.
+    ///
+    /// The second argument is the span of the original `old` expr.
+    ///
+    /// The third argument is the number of remember variables that have already been
+    /// instantiated in the surrounding environment.
+    ///
+    /// The fourth argument is a collection of all the expressions that need to be lifted
+    /// into the past environment as new remember variables.
```
feliperodri, line 249:
```suggestion
    /// You are provided with the expression that is the first argument of the
    /// `old()` call. You may modify it as you see fit. The return value
    /// indicates whether the entire `old()` call should be replaced by the
    /// (potentially altered) first argument.
    ///
    /// The second argument is the span of the original `old` expression.
    ///
    /// The third argument is the number of remember variables that have already been
    /// instantiated in the surrounding environment.
    ///
    /// The fourth argument is a collection of all the expressions that need to be lifted
    /// into the past environment as new remember variables.
```

library/kani_macros/src/sysroot/contracts/shared.rs
```diff
@@ -176,3 +179,246 @@ pub fn try_as_result_assign(stmt: &syn::Stmt) -> Option<&syn::LocalInit> {
...
+                    qself: None,
+                    path: Path { leading_colon: None, segments },
+                }) if segments.len() == 1
+                    && segments.first().map_or(false, |sgm| sgm.ident == "old") =>
+                {
+                    let first_segment = segments.first().unwrap();
+                    assert_spanned_err!(first_segment.arguments.is_empty(), first_segment);
+                    assert_spanned_err!(attrs.is_empty(), call);
+                    assert_spanned_err!(func_attrs.is_empty(), func);
+                    assert_spanned_err!(args.len() == 1, call);
```
feliperodri, line 297: Let's add a string description for each assertion to provide a better error message.

library/kani_macros/src/sysroot/contracts/shared.rs
```diff
@@ -176,3 +179,246 @@ pub fn try_as_result_assign(stmt: &syn::Stmt) -> Option<&syn::LocalInit> {
...
+        e: &mut Expr,
+        _: Span,
+        remember_count: usize,
+        remembers_exprs: &mut HashMap<Ident, Expr>,
+    ) -> bool {
+        let mut denier = OldVisitor { t: OldDenier, remember_count, remembers_exprs };
+        // This ensures there are no nested calls to `old`
+        denier.visit_expr_mut(e);
+
+        // The index of the remembers_exprs is offset by the remember_count of the surrounding environment
```
feliperodri, line 339:
```suggestion
        // The index of the `remembers_exprs` is offset by the `remember_count` of the surrounding environment
```

feliperodri (changes requested):
Only a few minor changes (mostly documentation) before we merge. Thanks for this!!

feliperodri:
@JustusAdam could you give another review?

library/kani_macros/src/sysroot/contracts/shared.rs
JustusAdam: Counting is brittle, plus code that matches over syntax (e.g. this function) are effort to maintain if the AST format changes. We should already have a routine that calculated a unique hash (it may even be stored in the generators already). Using a hash would completely eliminate this syntax matching function.

JustusAdam (approved):
Looks good to me otherwise

celinval (changes requested):
Can you please add a few `ui/` tests where the user misuses the API, for example:
  1. User tries to pass an object of type `T: !Copy` to `old`.
  2. `old` has no argument.
  3. `old` argument is not an expression.
  4. User tries to pass `result` to `old`.

tests/expected/function-contract/history/ui/statement.expected
```diff
@@ -0,0 +1 @@
+error
```
celinval, line 1: What is the error message that the user would see in this case?

celinval (commented):
Thanks for adding the `ui` tests. I think we need to work on these error messages though. I'm OK with making it a follow up PR, but I'll leave it up to you and @feliperodri.

tests/expected/function-contract/history/ui/statement.expected
pi314mm:
```
error[E0433]: failed to resolve: maybe a missing crate `core`?
 --> tests/expected/function-contract/history/ui/statement.rs:5:41
  |
5 | #[kani::ensures(|result| old(let x = 1+1;) old(*ptr + 1) == *ptr)]
  |                                         ^ maybe a missing crate `core`?
  ```
  Not sure if it's the error message we actually want, but it's also very malformated so it's really hard to get a good error message.

File: library/kani_macros/src/sysroot/contracts/shared.rs (hunk 4 above)
Comment (pi314mm) [reply library/kani_macros/src/sysroot/contracts/shared.rs]: Now using syn::Expr's Hash trait to hash the contents of `old(expr)`. If there is a hash collision, it would be because the exprs are identical, which means we can use the same variable for both expressions.

File: tests/expected/function-contract/history/ui/noncopy_ignore.expected [hunk 27]
```diff
@@ -0,0 +1 @@
+use of moved value: `ptr`
```
feliperodri, line 1: New line.

tests/expected/function-contract/history/ui/no_args.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
+
+#[kani::ensures(|result| {let x = old(); true})]
```
feliperodri, line 5: Add a description of what you are trying to test here.

tests/expected/function-contract/history/ui/statement.expected
feliperodri: @celinval should we create a GitHub issue to improve these error messages on a separate PR?

feliperodri (approved):

tests/expected/function-contract/history/ui/statement.expected
celinval: Sure. I'm OK with this being a follow up PR, but we definitely need to improve these messages.

pi314mm: Issue mentioned in #3273

library/kani_macros/src/sysroot/contracts/helpers.rs
pi314mm: Issue mentioned in #3273

library/kani_macros/src/sysroot/contracts/shared.rs
pi314mm: Issue mentioned in #3273

PR #3233, opened 2024-06-06
Remove empty box creation from contracts impl

celinval:
There seems to be an issue in CBMC contracts implementation that it assumes that `free` must have a body. However, slicing can remove `free` body if the harness does not allocate anything.

We used to create an empty Box before to force `free` to be in scope. Instead, just invoke `free(NULL)` which is a no-op.


jaisnan (approved):
Thanks for the quick fix, will be handy when replacing std invocations.

jaisnan:
Relevant issue filed on the CBMC side:
https://github.com/diffblue/cbmc/issues/8317

PR #3234, opened 2024-06-06
Contracts: Preventing "_renamed" from being visible to the user via reshadowing

### LINKED ISSUE #3026: Contracts: Error refers to nonvisible `*e_renamed` expression
Requested feature: When trying out the contracts feature in one of my projects, I got the following errors:

```
error[E0507]: cannot move out of `*e_renamed` which is behind a shared reference
  --> src/lib.rs:79:5
   |
79 |     #[kani::ensures(match result { Ok(t) => Self::typecheck_2(*e, t), Err(_) => true})]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ move occurs because `*e_renamed` has type `Expr`, which does not implement the `Copy` trait
   |

error[E0507]: cannot move out of `*e_renamed` which is behind a shared reference
  --> src/lib.rs:79:5
   |
79 |     #[kani::ensures(match result { Ok(t) => Self::typecheck_2(*e, t), Err(_) => true})]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ move occurs because `*e_renamed` has type `Expr`, which does not implement the `Copy` trait
   |

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0507`.
error: could not compile `interpreter` (lib) due to 3 previous errors
error: Failed to execute cargo (exit status: 101). Found 3 compilation errors.
```

The error refers to an unknown  `*e_renamed` expression which may be confusing for users. Moreover, the error message is printed twice despite the diagnostic being the same for both errors.

This error can be reproduced using the [kani-contracts](https://github.com/adpaco-aws/interpreter/tree/kani-contracts) branch of my project. Is it possible to improve the error message in this case?

pi314mm:
Currently, the ensures macro is expanded by taking the user provided expression and renaming all the function arguments to "arg_renamed". We attempt to hide the renaming by having the produced  kani::assert(renamed,"original") have the original version of the expression as a string for the error message.

However, this abstraction is occasionally broken when the "renamed" expression doesn't type check properly, and the error message propagated to the user features the renamed expression instead of the original one, causing the kind of confusion in #3026.

Instead of changing the input expression, it is possible to rebind the arguments back into their original form in the same way that they were renamed. This pull request introduce a new let binding to do so and remove the VisitMut that replaces the variables in the input expression.

Resolves #3026

### DIFF
--- a/library/kani_macros/src/sysroot/contracts/check.rs
+++ b/library/kani_macros/src/sysroot/contracts/check.rs
@@ -34,11 +34,13 @@ impl<'a> ContractConditionsHandler<'a> {
                 )
             }
             ContractConditionsData::Ensures { argument_names, attr } => {
-                let (arg_copies, copy_clean) = make_unsafe_argument_copies(&argument_names);
+                let (arg_copies, copy_rename, copy_clean) =
+                    make_unsafe_argument_copies(&argument_names);
 
                 // The code that enforces the postconditions and cleans up the shallow
                 // argument copies (with `mem::forget`).
                 let exec_postconditions = quote!(
+                    #copy_rename
                     kani::assert(#attr, stringify!(#attr_copy));
                     #copy_clean
                 );
--- a/library/kani_macros/src/sysroot/contracts/initialize.rs
+++ b/library/kani_macros/src/sysroot/contracts/initialize.rs
@@ -154,7 +154,4 @@ impl<'ast> Visit<'ast> for ArgumentIdentCollector {
         self.0.insert(i.ident.clone());
         syn::visit::visit_pat_ident(self, i)
     }
-    fn visit_receiver(&mut self, _: &'ast syn::Receiver) {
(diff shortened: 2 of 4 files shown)


pi314mm:
There may be an issue with this fix applied to "self"
It seems like there used to be a mechanism to do "let self_renamed = untracked_deref(&self)"
But to rebind this would mean doing
"let self = untracked_deref(&self_renamed)"
But this is not allowed as self is a keyword

All tests pass by ignoring the self parameter, so I am unsure of what test case would cause this to be an issue. Maybe a function that consumes self instead of call by reference?

celinval:
Can you please expand the description a bit? What was the issue and how this change fixes it. Thanks!

PR #3236, opened 2024-06-06
Add a regression test for running `verify-std` command.

celinval:
In #3226, we added a new command that allow Kani to verify properties in a custom standard library. In this PR, we create a script test that create a modified version of the standard library and runs Kani against it.

I also moved the script based tests to run after the more generic regression jobs. I think the script jobs are a bit harder to debug, they may take longer, and they are usually very specific to a few features. So probably best if we run the more generic tests first.

### DIFF
--- a/tests/script-based-pre/verify_std_cmd/verify_std.sh
+++ b/tests/script-based-pre/verify_std_cmd/verify_std.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+set +e
+
+TMP_DIR="tmp_dir"
+
+rm -rf ${TMP_DIR}
+mkdir ${TMP_DIR}
+
+# Create a custom standard library.
+SYSROOT=$(rustc --print sysroot)
+STD_PATH="${SYSROOT}/lib/rustlib/src/rust/library"
+cp -r "${STD_PATH}" "${TMP_DIR}"
+
+# Insert a small harness in one of the standard library modules.
+CORE_CODE='
+#[cfg(kani)]
+#[unstable(feature = "kani", issue = "none")]
+pub mod kani {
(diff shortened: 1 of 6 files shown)


tests/script-based-pre/verify_std_cmd/verify_std.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 4: Please add a description for the test.

tests/script-based-pre/verify_std_cmd/verify_std.sh
```diff
@@ -0,0 +1,76 @@
...
+'
+
+STD_CODE='
+#[cfg(kani)]
+mod verify {
+    use core::kani;
+    #[kani::proof]
+    fn check_non_zero() {
+        let orig: u32 = kani::any_raw_inner();
+        let Some(val) = core::num::NonZeroU32::new(orig) else { assert!(orig == 0); return };
```
zhassan-aws, line 61: Nit: an `if let Some` here will be slightly neater as it wouldn't require the `return`.

celinval: that is really nit-picking... I just thought the one liner would be cleaner.

zhassan-aws: Well, it's not really a one-liner given that the `else` has two statements 😄 

Feel free to leave it as is of course.

celinval: it's OK.. I need to figure out the MacOS regression anyway. If this was the only change then I would push back.

BTW, thanks for the quick review.

jaisnan (approved):
lgtm now

PR #3237, opened 2024-06-06
Add a few examples of using shadow memory to check initialization of slices

zhassan-aws:
A follow-up on #3200: use API to check that slices produced by some slice operations that internally use `unsafe` are properly initialized.


artemagvanian:
Nice! Those are interesting examples, although I am a bit confused about the semantics of the checks. 

Right now, the only thing that is actually checked is that the underlying memory of the slices points to the same memory objects that were previously tagged as initialized. In other words, if the methods under verification somehow deinitialized the memory or allocated and initialized new memory, the checks wouldn't catch it. 

Maybe we should add some comments explaining the limitations of the checks?

zhassan-aws:
Thanks for taking a look @artemagvanian.

> Right now, the only thing that is actually checked is that the underlying memory of the slices points to the same memory objects that were previously tagged as initialized.

Correct. The purpose of the tests is to check that those _unsafe_ methods do not mess up the start/end of the slices produced in a way that makes them go out of bounds of the original arrays/slices.

> In other words, if the methods under verification somehow deinitialized the memory or allocated and initialized new memory, the checks wouldn't catch it.

Correct. The proper way to check for such cases is what you did in #3264. The main purpose of the tests in this PR is to demonstrate a possible usage of the shadow memory APIs. They're not really verifying those methods. I can add a comment that clarifies that.

zhassan-aws:
@artemagvanian I added clarifying comments in those tests.

artemagvanian (approved):
Looks good to me!

PR #3242, opened 2024-06-07
RFC: Quantifiers

feliperodri:
RFC describing the support for existential and universal quantifiers in Kani.

### DIFF
--- a/rfc/src/rfcs/0010-quantifiers.md
+++ b/rfc/src/rfcs/0010-quantifiers.md
@@ -2,7 +2,7 @@
 - **Feature Request Issue:** [#2546](https://github.com/model-checking/kani/issues/2546) and [#836](https://github.com/model-checking/kani/issues/836)
 - **RFC PR:** [#](https://github.com/model-checking/kani/pull/)
 - **Status:** Unstable
-- **Version:** 1
+- **Version:** 1.0
 
 -------------------
 
@@ -27,11 +27,11 @@ This new feature doesn't introduce any breaking changes to users. It will only a
 We propose a syntax inspired by ["Pattern Types"](https://github.com/rust-lang/rust/pull/120131). The syntax of existential (i.e., `kani::exists`) and universal (i.e., `kani::forall`) quantifiers are:
 
 ```rust
-kani::exists(|<var>: <type> is <range-expr> | <boolean-expression>)
-kani::forall(|<var>: <type> is <range-expr> | <boolean-expression>)
+kani::exists(|<var>: <type> [is <range-expr>] | <boolean-expression>)
+kani::forall(|<var>: <type> [is <range-expr>] | <boolean-expression>)
 ```
 
-CBMC's SAT backend only supports bounded quantification under **constant** lower and upper bounds (for more details, see the [documentation for quantifiers in CBMC](https://diffblue.github.io/cbmc/contracts-quantifiers.html)). In contracts, the SMT backend supports arbitrary Boolean expressions.
+If `<range-expr>` is not provided, we assume `<var>` can range over all possible values of the given `<type>` (i.e., syntactic sugar for full range `|<var>: <type> as .. |`). CBMC's SAT backend only supports bounded quantification under **constant** lower and upper bounds (for more details, see the [documentation for quantifiers in CBMC](https://diffblue.github.io/cbmc/contracts-quantifiers.html)). The SMT backend, on the other hand, supports arbitrary Boolean expressions. In any case, `<boolean-expression>` should not have side effects, as the purpose of quantifiers is to assert a condition over a domain of objects without altering the state.
 
 Consider the following example adapted from the documentation for the [from_raw_parts](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts) function:
 
@@ -179,7 +179,7 @@ The usage of quantifiers should be valid in any part of the Rust code analysed b
 
 <!-- For the implementors or the hackers -->
 
-Kani should have the same support that CBMC has for quantifiers with its SAT backend. For more details, see [Quantifiers](https://github.com/diffblue/cbmc/blob/0a69a64e4481473d62496f9975730d24f194884a/doc/cprover-manual/contracts-quantifiers.md).
+Kani should have the same support that CBMC has for quantifiers. For more details, see [Quantifiers](https://github.com/diffblue/cbmc/blob/0a69a64e4481473d62496f9975730d24f194884a/doc/cprover-manual/contracts-quantifiers.md).
 
 
 ## Open questions
@@ -192,6 +192,7 @@ Kani should have the same support that CBMC has for quantifiers with its SAT bac
   interface is familiar to developers, but the code generation is tricky, as
   CBMC level quantifiers only allow certain kinds of expressions. This
   necessitates a rewrite of the `Fn` closure to a compliant expression.
+    - Which kind of expressions should be accepted as a "compliant expression"? 
 
 
 ## Future possibilities
--- a/rfc/src/rfcs/0010-quantifiers.md
+++ b/rfc/src/rfcs/0010-quantifiers.md
@@ -8,35 +8,30 @@
 
 ## Summary
 
-Quantifiers are like logic-level loops and a powerful reasoning helper, which allow us to express statements about objects in a domain that satisfy a given condition. There are two primary quantifiers: the existential quantifier (∃) and the universal quantifier (∀).
-
-1. The existential quantifier (∃): represents the statement "there exists." We use to express that there is at least one object in the domain that satisfies a given condition. For example, "∃x P(x)" means "there exists an x such that P(x) is true."
-
-2. The universal quantifier (∀): represents the statement "for all" or "for every." We use it to express that a given condition is true for every object in the domain. For example, "∀x P(x)" means "for every x, P(x) is true."
+Quantifiers are logical operators that allow users to express that a property or condition applies to some or all objects within a given domain.
 
 ## User Impact
 
-Quantifiers enhance users' expressive capabilities, enabling the verification of more intricate properties in a concise manner.
+There are two primary quantifiers: the existential quantifier (∃) and the universal quantifier (∀).
+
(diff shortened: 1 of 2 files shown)


rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -0,0 +1,207 @@
+- **Feature Name:** Quantifiers
+- **Feature Request Issue:** [#2546](https://github.com/model-checking/kani/issues/2546) and [#836](https://github.com/model-checking/kani/issues/836)
+- **RFC PR:** [#](https://github.com/model-checking/kani/pull/)
+- **Status:** Unstable
+- **Version:** 1
+
+-------------------
+
+## Summary
```
celinval, line 9: I would move this definitions to the section below, and just write 1-2 sentences on what this is proposing.

Add support to quantifiers in Kani to allow users to express properties about objects over an entire domain that satisfy a given condition.

celinval (commented):
Thanks Felipe. This is a very well-written RFC!

rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -0,0 +1,207 @@
...
+
+Quantifiers are like logic-level loops and a powerful reasoning helper, which allow us to express statements about objects in a domain that satisfy a given condition. There are two primary quantifiers: the existential quantifier (∃) and the universal quantifier (∀).
```
tautschnig, line 11: That first part of the sentence trips me up: this is perhaps too simplistic a language?

rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -0,0 +1,207 @@
...
+
+1. The existential quantifier (∃): represents the statement "there exists." We use to express that there is at least one object in the domain that satisfies a given condition. For example, "∃x P(x)" means "there exists an x such that P(x) is true."
```
tautschnig, line 13: Maybe say "there exists *a value* x"

rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -0,0 +1,207 @@
...
+
+2. The universal quantifier (∀): represents the statement "for all" or "for every." We use it to express that a given condition is true for every object in the domain. For example, "∀x P(x)" means "for every x, P(x) is true."
```
tautschnig, line 15: Say "for every *value* x"

rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -0,0 +1,207 @@
...
+
+## User Impact
+
+Quantifiers enhance users' expressive capabilities, enabling the verification of more intricate properties in a concise manner.
```
tautschnig, line 19: Is it actually try that expressive capabilities are enhanced (I'm not even sure that statement is well defined). In a Turing-complete language expressiveness is not going to change through quantifiers.

rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -0,0 +1,207 @@
...
+  bool`, where `T` is some arbitrary type that can be quantified over. This
+  interface is familiar to developers, but the code generation is tricky, as
+  CBMC level quantifiers only allow certain kinds of expressions. This
+  necessitates a rewrite of the `Fn` closure to a compliant expression.
+
+
+## Future possibilities
+
+<!-- For Developers -->
+- CBMC has an SMT backend which allow the use of quantifiers with arbitrary Boolean expressions. Kani must include an option for users to experiment with this backend.
```
tautschnig, line 205: Nit pick: which allow -> which allows

tautschnig (approved):
Ok from my point of view, except I'd prefer a somewhat more precise language.

rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -0,0 +1,207 @@
...
+
+This new feature doesn't introduce any breaking changes to users. It will only allow them to write properites using the existential (∃) and universal (∀) quantifiers.
+
+## User Experience
+
+We propose a syntax inspired by [Prusti](https://viperproject.github.io/prusti-dev/user-guide/syntax.html?highlight=quanti#quantifiers). Quantifiers must have only one bound variable (restriction imposed today by CBMC). The syntax of existential (i.e., `kani::exists`) and universal (i.e., `kani::forall`) quantifiers are:
+
+```rust
+kani::exists(|<bound variable>: <bound variable type>| <range> && <Boolean expression>)
```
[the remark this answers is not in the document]
pnkfelix: I agree that I wouldn't use `<range> && <Boolean expression>` ; overloading `&&` that way might be problematic.

There is an experimental proposal on the table for a ["Pattern Types" syntax](https://github.com/rust-lang/rust/pull/120131), but I would put the emphasis on "experiment" there... still, if you were to take inspiration from that, then I think the closure in your example would be written `|<var>: <type> is <range>| <boolean-expression>`; but I don't know whether that's **better** for your purposes than what @celinval has proposed.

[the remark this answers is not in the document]
qinheping: Why is the range expression special here? Couldn't the range expression be just part of the boolean expression?

rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -0,0 +1,207 @@
...
+kani::forall(|<bound variable>: <bound variable type>| <range> ==> <Boolean expression>)
+```
+
+where `<range>` is an expression of the form
+
+```rust
+<lower bound> <= <bound variable> && <bound variable> < <upper bound>
+```
+
+where `lower bound` and `upper bound` are constants. The bound predicates could be strict (e.g., `<lower bound> < <bound variable>`), or non-strict (e.g., `<upper bound> <= <bound variable>`), but both the bounds must be **constants**. CBMC's SAT backend only supports bounded quantification under **constant** lower and upper bounds.
```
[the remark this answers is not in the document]
qinheping: With SMT-backends, CBMC supports not-constants bound. Is the restriction of being constant bounded necessary here?

rfc/src/rfcs/0010-quantifiers.md
feliperodri: I wrote that phrase with "enabling developers to write more general properties" or something with that sort in mind. Removing it for clarity.

rfc/src/rfcs/0010-quantifiers.md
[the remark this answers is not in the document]
feliperodri: I included this restriction only applies for the SAT backend.

rfc/src/rfcs/0010-quantifiers.md
feliperodri: I re-wrote the summary and moved the text for the second section.

rfc/src/rfcs/0010-quantifiers.md
feliperodri: I removed the first sentence.

rfc/src/rfcs/0010-quantifiers.md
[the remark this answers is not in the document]
feliperodri: We need a way to reference to the variable used to specify the range in the Boolean expression. I really liked the suggestion by @pnkfelix about Pattern Types. @celinval what do you think of the new syntax proposal?

rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -0,0 +1,202 @@
...
+Rather than exhaustively listing all elements in a domain, quantifiers enable users to make statements about the entire domain at once. This compact representation is crucial when dealing with large or unbounded inputs. Quantifiers also facilitate abstraction and generalization of properties. Instead of specifying properties for specific instances, quantified properties can capture general patterns and behaviors that hold across different objects in a domain. Additionally, by replacing loops in the specification with quantifiers, Kani can encode the properties more efficiently within the specified bounds, making the verification process more manageable and computationally feasible.
+
+This new feature doesn't introduce any breaking changes to users. It will only allow them to write properites using the existential (∃) and universal (∀) quantifiers.
+
+## User Experience
+
+We propose a syntax inspired by ["Pattern Types"](https://github.com/rust-lang/rust/pull/120131). The syntax of existential (i.e., `kani::exists`) and universal (i.e., `kani::forall`) quantifiers are:
+
+```rust
+kani::exists(|<var>: <type> is <range-expr> | <boolean-expression>)
```
celinval, line 30: Can the `<boolean-expression>` have side effects?

rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Open questions
+
+<!-- For Developers -->
+- **Function Contracts RFC** - CBMC has support for both `exists` and `forall`, but the
+  code generation is difficult. The most ergonomic and easy way to implement
+  quantifiers on the Rust side is as higher-order functions taking `Fn(T) ->
+  bool`, where `T` is some arbitrary type that can be quantified over. This
+  interface is familiar to developers, but the code generation is tricky, as
+  CBMC level quantifiers only allow certain kinds of expressions. This
```
celinval, line 193: So which kinds of expressions are allowed? I think we either call it out as an open question or document as part of the UX?

rfc/src/rfcs/0010-quantifiers.md
celinval: We could also accept:
```rust
kani::exists(|<var>: <type>| <bool_expr>);
```
as syntactic sugar to full range:
```rust
kani::exists(|<var>: <type> as .. | <bool_expr>);
```

celinval (approved):
Looks good! Thanks

qinheping (approved):
LGTM

rfc/src/rfcs/0010-quantifiers.md
feliperodri: `<boolean-expression>` should not have side effects. The purpose of quantifiers is to assert a condition over a domain of objects without altering the state of any memory. I'll update the text to highlight this point.

rfc/src/rfcs/0010-quantifiers.md
feliperodri: I'll add it as an open question.

Issue #3238, opened 2024-06-07
Issue injecting Kani module to dependency (zerocopy)


zi0Black:
I tried to just create a basic proof on the Aptos-Core project, but I faced an issue with building with Kani, likely a bug with module injection:
```
error[E0433]: failed to resolve: use of undeclared crate or module `kani`
    --> /home/xcd/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zerocopy-0.7.34/src/lib.rs:8238:9
     |
8238 |         kani::assume(matches!(base.size_info, SizeInfo::SliceDst(..)));
     |         ^^^^ use of undeclared crate or module `kani`

error[E0433]: failed to resolve: use of undeclared crate or module `kani`
    --> /home/xcd/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zerocopy-0.7.34/src/lib.rs:8247:33
     |
8247 |         let layout: DstLayout = kani::any();
     |                                 ^^^^ use of undeclared crate or module `kani`

error: aborting due to 37 previous errors

For more information about this error, try `rustc --explain E0433`.
error: could not compile `zerocopy` (lib) due to 38 previous errors
warning: build failed, waiting for other jobs to finish...
error: Failed to execute cargo (exit status: 101). Found 38 compilation errors.
```
I believe this bug triggers only on certain combination of features, cargo build environment, and previous issue: https://github.com/model-checking/kani/issues/3109, where there was some issue with replicating the environment.

I tried this code:
1. Clone Aptos Core: `git clone https://github.com/aptos-labs/aptos-core`
2. Run dev setup: `./scripts/dev_setup.sh`
3. Install and setup Kani
4. Open `consensus/src/transaction_shuffler/sender_aware.rs`
5. Add the following code at the bottom:
```rust
#[cfg(kani)]
mod verification {
    use super::*;

    #[kani::proof]
    pub fn test_3_sender_shuffling() {

    }
}
```
6. Run: `cargo kani --package aptos-consensus --harness transaction_shuffler::sender_aware::verification::test_3_sender_shuffling`

with Kani version: 0.52

I expected to see this happen: Kani building the project.

Instead, this happened: A dependency which uses Kani, zerocopy, fails to build because there is a missing Kani module.

```
xcd@ttt:~/aptos-core$ RUST_BACKTRACE=1 cargo kani --package aptos-consensus --harness transaction_shuffler::sender_aware::verification::test_3_sender_shuffling
[...]

PR #3244, opened 2024-06-08
Use cfg=kani_host for host crates

### LINKED ISSUE #3101: Kani symbols not available in path dependency when build script depends on same local crate
Hey, I am currently running into a compilation error when I have two local crates where one depends on the other in the build script.

TL;DR: `kani` symbols are not available in path dependency crate if dependent crate has a build script that also imports the same path dependency.

This is the directory structure:

```
binary
binary/build.rs
...
constants
```

And binary has a both a build and normal dependency on `constants` (`binary/Cargo.toml`):

```
[dependencies]
constants = { path = "../constants" }

[build-dependencies]
constants = { path = "../constants" }
```

Both crates have some `#[cfg(kani)]` directives in their code. When I run `cargo kani` in `constants`, it compiles and verifies. However, when running `cargo kani` in the `binary` crate, I get this error:

```
$ cd binary && cargo kani -v
Kani Rust Verifier 0.48.0 (cargo plugin)
Skipped the following unsupported targets: 'build-script-build'.
[Kani] Running:

[...]

error[E0433]: failed to resolve: use of undeclared crate or module `kani`
 --> constants/src/lib.rs:9:6
  |
9 | impl kani::Arbitrary for SomeStruct {
  |      ^^^^ use of undeclared crate or module `kani`

error[E0433]: failed to resolve: use of undeclared crate or module `kani`
  --> constants/src/lib.rs:12:25
   |
12 |             some_field: kani::any()
   |                         ^^^^ use of undeclared crate or module `kani`
```

This only happens if the `build.rs` script that also depends on the `constants` crate is present; if you delete it (no need to adjust `Cargo.toml`, just remove the script), everything works again. However, I need the build script for my normal build.

I tried this code: https://github.com/xarantolus/kani-issue-build-script, it contains the mentioned code and structure.

```
cd binary && cargo kani -v
```

with Kani version: `Kani Rust Verifier 0.48.0 (cargo plugin)`

I expected to see this happen: compilation should work for the dependency crate (`constants`) even when a build script that depends on `constants` is present

Instead, this happened: It seems like kani is not injected into the path dependency `constants` if the package we want to verify has a build script that also depends on the `constants` package.

tautschnig:
We want to run the proofs in the target crate and don't need to build (or run) the proofs in any of the host crates. This avoids a need to make available the `kani` crate to any such host crates.

Resolves #3101, #3238


tests/cargo-kani/build-rs-plus-host-with-kani-proofs/README.md
```diff
@@ -0,0 +1,10 @@
+This repo contains contains a minimal example that breaks compilation when using [kani](https://github.com/model-checking/kani), where I would expect compilation to work.
```
feliperodri, line 1:
```suggestion
This repository contains a minimal example that breaks compilation when using [kani](https://github.com/model-checking/kani), where we would expect compilation to work.
```

tests/cargo-kani/build-rs-plus-host-with-kani-proofs/constants/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+[package]
+name = "constants"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
feliperodri, line 8:
```suggestion
```

feliperodri (approved):

docs/src/usage.md
```diff
@@ -84,6 +84,7 @@ When Kani builds your code, it does two important things:
 
 1. It sets `cfg(kani)`.
 2. It injects the `kani` crate.
+3. For host crates (crates that your target crate depends on), Kani sets `cfg(kani_host)`.
```
celinval, line 87: 1. Line 83 says "two important things". We should increment that. :)
2. I would change the language of 1 and 3.
> 1. It sets `kani` configuration for target crates compilation (including dependencies).
> 3. It sets `kani_host` configuration for host build targets such as build script and procedural macro crates.

tests/cargo-kani/build-rs-plus-host-with-kani-proofs/constants/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// From https://github.com/model-checking/kani/issues/3101
+
+pub const SOME_CONSTANT: u32 = 42;
```
celinval, line 6: I would suggest to enhance this test a bit. You can define a different value for this constant based on whether this is `kani_host`, `kani` or none. something like:

```suggestion
#[cfg(not(any(kani, kani_host)))]
pub const SOME_CONSTANT: u32 = 0;
#[cfg(kani)]
pub const SOME_CONSTANT: u32 = 1;
#[cfg(kani_host)]
pub const SOME_CONSTANT: u32 = 2;
```

So in the build script, you can add:
```rust

#[cfg(not(kani_host))]
assert_eq!(constants::SOME_CONSTANT, 0);
#[cfg(kani_host)]
assert_eq!(constants::SOME_CONSTANT, 2);
```

and inside the harness, you can add:
```rust
assert_eq!(constants::SOME_CONSTANT, 1);
```

celinval (commented):
Thanks for doing this

tests/cargo-kani/build-rs-plus-host-with-kani-proofs/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+Deleting the `binary/build.rs` script makes the compilation work suddenly, despite it being skipped anyways:
+
+```
+binary$ cargo kani -v
+Kani Rust Verifier 0.48.0 (cargo plugin)
+Skipped the following unsupported targets: 'build-script-build'.
```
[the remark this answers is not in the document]
tautschnig: Admitted I also don't know what "build-script-build" means - could you enlighten me so that I can add specific text for this case (rather than throwing it into the catch-all `ignored-unsupported`?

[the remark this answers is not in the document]
celinval: That target represents the `build.rs` file. https://doc.rust-lang.org/cargo/reference/build-scripts.html

Issue #3152, opened 2024-04-19
Enable Kani to verify the standard library crates


celinval:
As we experiment with verifying the standard library, it might be helpful to enable users to verify the standard library crates directly, instead of having to create wrappers. Those wrappers can also only be used to verify public components.

There are a few things that need to be done in order to do that:

- [ ] Create a `#[no_core]` Kani library.
- [ ] Instrument contracts in `core` without proc_macros.
- [ ] Build std fork with `cargo build -Z build-std`.
- [ ] Run verification together with `cargo build -Z build-std`.

celinval:
See https://github.com/rust-lang/wg-cargo-std-aware/issues/7 for workarounds on how to build a modified `std` library.

celinval:
We have added a sub-command to Kani (`verify-std`) that is able to verify the std library implementation.

Issue #2435, opened 2023-05-09
Use InternedString in other parts of our code (such as Metadata)


celinval:
**Proposed change:** This would require the following changes:

1. Assess the memory consumption due to using String in many places such as Metadata crate.
2. Move the InternedString code out of the CBMC crate. Maybe create a `kani_common` crate.
3. Modify the crates that still heavily rely on String, such as the Metadata crate.

**Motivation:** Keeping strings can increase the memory consumption of our software.

PR #3245, opened 2024-06-09
Refactor stubbing so Kani compiler only invoke rustc once per crate

### LINKED ISSUE #3072: Migrate stubbing transformation to be performed in the instance body
### Proposed change:

Perform stubbing in the instance body instead of the generic one.

### Motivation

This will simplify stubbing code, remove the redundant compilation and provide more precise error handling.

### Background:

The stubbing transformation and validation are spread into multiple stages. The stubbing itself is done in the generic body by hijacking rustc's transformation query. Because of that, we have the following limitations:
1. If the function being stubbed is generic, we delay the instantiation requirements validation to be done during the reachability analysis.
2. We need to re-run the compilation for every set of stubs to invalidate the query cache and apply the right set of stubs.
3. We have to override `collect_and_partition_mono_items` query to avoid an ICE related to the delayed validation of the instantiation requirements.

There is a downside though. Today, we are able to apply the stubs to concrete playback, which wouldn't be possible until changes to the StableMIR are visible to the rust compiler.

celinval:
Using stubs or function contracts as part of the `verify-std` sub-command does not work with multiple rustc executions as previous implementation. This happens because we now enable verifying dependencies, and cargo crashes due to a race condition. As soon as the first rustc invocation succeeds, cargo starts the compilation of the dependents crate. However, new executions can override files.

Instead, we moved the stub logic to the new transformation framework, which is done on the top of the StableMIR body, and doesn't affect the Rust compiler session. We are now able to apply stub without restarting the compiler. This is a much better user experience as well, since multiple calls to the compiler can print the same warnings multiple times.

Resolves #3072
Towards #3152 


jaisnan (approved):
Thank you! It's a big refactor but the regressions passing gives me enough confidence to approve the change. I would like to see some examples related to the refactor added to the description if possible, to make why this was necessary a bit more clear.

kani-compiler/src/kani_compiler.rs
```diff
@@ -69,537 +54,64 @@ fn backend(queries: Arc<Mutex<QueryDb>>) -> Box<CodegenBackend> {
...
-            debug!(?queries, "config end");
-        }
+        // Configure queries.
+        let queries = &mut (*self.queries.lock().unwrap());
+        queries.set_args(args);
+        debug!(?queries, "config end");
     }
 
-    /// During the initialization state, we collect the crate harnesses and prepare for codegen.
+    /// After analysis, we check the crate items for Kani API misusage or configuration issues.
```
feliperodri, line 103: typo.
```suggestion
    /// After analysis, we check the crate items for Kani API misuse or configuration issues.
```

kani-compiler/src/kani_middle/codegen_units.rs
```diff
@@ -0,0 +1,214 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This module is responsible for extracting grouping harnesses that can be processed together
+//! by codegen.
+//!
+//! Today, only stub / contracts can affect the harness codegen. Thus, we group the harnesses
+//! according to their stub configuration.
+
+// TODO: Move this out of CBMC crate.
```
feliperodri, line 10: Is this `#TODO` new? Are their any GitHub issues explaining this problem?

kani-compiler/src/kani_middle/transform/body.rs
```diff
@@ -288,3 +284,136 @@ impl SourceInstruction {
 fn find_instance(tcx: TyCtxt, diagnostic: &str) -> Option<Instance> {
     Instance::resolve(find_fn_def(tcx, diagnostic)?, &GenericArgs(vec![])).ok()
 }
+
+/// Basic mutable body visitor.
+///
+/// We removed a many methods for simplicity.
```
feliperodri, line 290: typo.
```suggestion
/// We removed many methods for simplicity.
```

kani-compiler/src/kani_middle/transform/body.rs
```diff
@@ -288,3 +284,136 @@ impl SourceInstruction {
...
+///
+/// TODO: Contribute this to stable_mir.
```
feliperodri, line 292: Should we track this in a GitHub issue?

kani-compiler/src/kani_queries/mod.rs
```diff
@@ -2,65 +2,28 @@
...
-    path::PathBuf,
-    sync::{Arc, Mutex},
-};
+use std::sync::{Arc, Mutex};
 
 use crate::args::Arguments;
 
 /// This structure should only be used behind a synchronized reference or a snapshot.
+///
+/// TODO: Merge this with arguments
```
feliperodri, line 11: What does that mean? Do we have a GitHub issue to track this?

tests/cargo-kani/stubbing-double-extern-path/harness/expected
```diff
@@ -1 +1,2 @@
-error[E0391]: cycle detected when optimizing MIR for `crate_b::assert_false`
+error: Cannot stub `crate_b::assert_false`. Stub configuration for harness `check_inverted` has a cycle
```
feliperodri, line 1: Way better error message.

tests/script-based-pre/stubbing_compiler_sessions/check_compiler_sessions.expected
```diff
@@ -1,3 +0,0 @@
-Complete - 6 successfully verified harnesses, 0 failures, 6 total.
```
feliperodri, line 1: Why can we safely delete all these test-related files?

tests/script-based-pre/verify_std_cmd/verify_std.sh
```diff
@@ -85,7 +100,7 @@ cat ${TMP_DIR}/std_lib.rs >> ${TMP_DIR}/library/std/src/lib.rs
 
 echo "[TEST] Run kani verify-std"
 export RUST_BACKTRACE=1
-kani verify-std -Z unstable-options "${TMP_DIR}/library" --target-dir "${TMP_DIR}/target"
+kani verify-std -Z unstable-options "${TMP_DIR}/library" --target-dir "${TMP_DIR}/target" -Z stubbing
```
feliperodri, line 103: Why `-Z unstable-options` doesn't include stubbing? Should we also add `-Z function-contracts`

feliperodri (approved):

kani-compiler/src/kani_middle/codegen_units.rs
celinval: https://github.com/model-checking/kani/issues/2435

celinval: Basically interned string is not something specific to CBMC. We added it there a long time ago when dealing with high memory consumption in Kani due to String usage in the CBMC crate.

kani-compiler/src/kani_middle/transform/body.rs
celinval: Sure: https://github.com/rust-lang/project-stable-mir/issues/81

kani-compiler/src/kani_queries/mod.rs
celinval: This struct is now redundant, since it is only wrapping Args.

I prefer avoiding GitHub issues for cleanup work that has no effect to the end user.

tests/script-based-pre/stubbing_compiler_sessions/check_compiler_sessions.expected
celinval: I can get the test back if you want.

tests/script-based-pre/verify_std_cmd/verify_std.sh
celinval: We can add `-Z function-contracts` with an example once Jai's PR has been merged (#3153).

kani-compiler/src/kani_middle/codegen_units.rs
celinval: I deleted the TODO

tests/script-based-pre/stubbing_compiler_sessions/check_compiler_sessions.expected
celinval: I'll move the test itself to the `expected` suite, since it's now just a regular test to ensure we can verify harnesses with different sets of stubs.

PR #3247, opened 2024-06-10
Fix "unused mut" warnings created by generated code.

### LINKED ISSUE #3010: kani contracts cause false-positives for unused mut lint on the gcd example
I tried this code, which was reduced from the [recent blog post](https://model-checking.github.io/kani-verifier-blog/2024/01/29/function-contracts.html)

```rust
#[kani::ensures(a % result == 0 && b % result == 0 && result != 0)]
fn gcd(mut a: u64, mut b: u64) -> u64 {
    (&mut a, &mut b);
    0
}
```

```
cargo kani  -Zfunction-contracts
```

with Kani version: Kani Rust Verifier 0.45.0 (cargo plugin)

I expected to see this happen: Kani runs with no incorrect lint warnings from Rust.

Instead, this happened: Kani prints this:

```
warning: variable does not need to be mutable
 --> src/main.rs:2:8
  |
2 | ...cd(mut a: u...
  |       ----^
  |       |
  |       help: remove this `mut`
  |
  = note: `#[warn(unused_mut)]` on by default

warning: variable does not need to be mutable
 --> src/main.rs:2:20
  |
2 | ...4, mut b: u...
  |       ----^
  |       |
  |       help: remove this `mut`

warning: 2 warnings emitted
```

The advice from these warnings is not correct; the `mut` modifiers on the bindings need to be there, due to the way the function bodies are written. (And if you follow the given advice, you do get hard errors about needing to put the `mut`'s back.)

My advice: Just turn off the `unused_mut` linting in tool invocation done by the proc-macro, if you can.

jsalzbergedu:
This change adds "unused_mut" to the list of suppressed lints for wrappers generated by the contracts macros. This will get rid of spurious errors caused by mutable parameters to functions.

This fixes the example from https://github.com/model-checking/kani/issues/3010 .

It can be tested by adding the example from the issues to tests/expected/test_macros/gcd.rs, creating a file tests/expected/test_macros/gcd.expected, then running 
```bash
cargo build-dev
RUST_BACKTRACE=1 cargo run -p compiletest -- --logfile logfile.txt --suite expected --mode expected --ignored --no-fail-fast --src-base tests/expected/test_macros
 ```
 
 RESOLVES #3010 

### DIFF
--- a/library/kani_macros/src/sysroot/contracts/shared.rs
+++ b/library/kani_macros/src/sysroot/contracts/shared.rs
@@ -50,7 +50,7 @@ impl<'a> ContractConditionsHandler<'a> {
     pub fn emit_common_header(&mut self) {
         if self.function_state.emit_tag_attr() {
             self.output.extend(quote!(
-                #[allow(dead_code, unused_variables)]
+                #[allow(dead_code, unused_variables, unused_mut)]
             ));
         }
         self.output.extend(self.annotated_fn.attrs.iter().flat_map(Attribute::to_token_stream));
--- a/library/kani_macros/src/sysroot/contracts/bootstrap.rs
+++ b/library/kani_macros/src/sysroot/contracts/bootstrap.rs
@@ -85,7 +85,7 @@ impl<'a> ContractConditionsHandler<'a> {
 
         let result = Ident::new(INTERNAL_RESULT_IDENT, Span::call_site());
         self.output.extend(quote!(
-            #[allow(dead_code, unused_variables)]
+            #[allow(dead_code, unused_variables, unused_mut)]
             #[kanitool::is_contract_generated(recursion_wrapper)]
             #wrapper_sig {
                 static mut REENTRY: bool = false;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: library/kani_macros/src/sysroot/contracts/shared.rs [hunk 1]
```diff
@@ -50,7 +50,7 @@ impl<'a> ContractConditionsHandler<'a> {
     pub fn emit_common_header(&mut self) {
         if self.function_state.emit_tag_attr() {
             self.output.extend(quote!(
-                #[allow(dead_code, unused_variables)]
+                #[allow(dead_code, unused_variables, unused_mut)]
```
feliperodri, line 53: Could you add or modify an existing test case to confirm this fix? What you can do is to update some existing `.expected` file to make sure there are no more warnings of this kind.

feliperodri (changes requested):

library/kani_macros/src/sysroot/contracts/shared.rs
jsalzbergedu: Changes to the .expected file didn't seem to do it; I tried instead binding x and y as mut in the original gcd test and adding #![deny(warnings)]. I've confirmed locally that #![deny(warnings)] will cause the regression test to fail unless you add unused_mut.

feliperodri (approved):

PR #3250, opened 2024-06-10
Add a `#[derive(Invariant)]` macro

adpaco-aws:
This PR adds a `#[derive(Invariant)]` macro for structs which allows users to automatically derive the `Invariant` implementations for any struct. The derived implementation determines the invariant for the struct as the conjunction of invariants of its fields. In other words, the invariant is derived as `true && self.field1.is_safe() && self.field2.is_safe() && ..`.

For example, for the struct

```rs
#[derive(kani::Invariant)]
struct Point<X, Y> {
    x: X,
    y: Y,
}
```

we derive the `Invariant` implementation as

```rs
impl<X: kani::Invariant, Y: kani::Invariant> kani::Invariant for Point<X, Y> {
    fn is_safe(&self) -> bool {
        true && self.x.is_safe() && self.y.is_safe()
    }
}
```

Related #3095 


celinval (commented):
Can you add a little test where an invariant created fails? Something like you have a type `A` with a custom safety invariant and a struct `B` with a derived `Invariant` that has a member of type `A`. Now ensure that breaking the `A` invariant in the `B` field also breaks `B` invariant. Thanks!

adpaco-aws:
Added a test where the invariant check fails but let me know if you have suggestions for the setup step (right now it's just an overly generic `Arbitrary` implementation).

tests/expected/derive-invariant/invariant_fail/invariant_fail.rs
celinval, line 1: That's an interesting one... This exposes a problem where we shouldn't let users use `#[derive(Arbitrary)]` if they actually have any value restriction. I.e.: The `kani::any()` in this case could generate unsafe values. Not sure there's a way to error out here, except by adding some sort of validation in the compiler.

That said, I think we should extend both derive clauses to provide an attribute helper to make it easier for users to add value restrictions that would apply to both:

```rust
#[derive(kani::Arbitrary, kani::Invariant)]
struct NotNegative {
  #[invariant(inner > 0)]
  inner: i32
};
```

tests/expected/derive-invariant/invariant_fail/invariant_fail.rs
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Check that a verification failure is triggered when the derived `Invariant`
+//! method is checked but not satisfied.
+
+extern crate kani;
+use kani::Invariant;
+
+#[derive(kani::Arbitrary)]
+struct NotNegative(i32);
```
celinval, line 11: Maybe add a comment here:

```suggestion
/// Incorrect usage of `Arbitrary` and `Invariant` types.
///
/// `Arbitrary` implementation should respect the type invariant.
#[derive(kani::Arbitrary)]
struct NotNegative(i32);
```

celinval (approved):
We need to figure out how to ensure `Arbitrary` respects `Invariant`.

tests/expected/derive-invariant/invariant_fail/invariant_fail.rs
adpaco-aws: I've opened #3265 to have a record of the problem and discuss potential solutions for it.

On the attribute helper, it sounds great to me. I'm currently developing the `#[invariant(<predicate>)]` at the struct level, which would look like
```rs
#[kani::invariant(self.inner >= 0)]
struct NotNegative {
  inner: i32
};
```
but I'm not sure these two could (or should) be made compatible.

tests/expected/derive-invariant/invariant_fail/invariant_fail.rs
adpaco-aws: Added manually in addition to the link to the issue.

PR #3255, opened 2024-06-11
Fix a few issues with std verification

### LINKED ISSUE #3251: Kani contract cannot be applied to functions with `rustc_diagnostic_item`
I tried this code (on the `core` crate):
```rust
#[inline]
#[kani_core::requires(ub_checks::can_dereference(src))]
#[stable(feature = "volatile", since = "1.9.0")]
#[cfg_attr(miri, track_caller)] // even without panics, this helps for Miri backtraces
#[rustc_diagnostic_item = "ptr_read_volatile"]
pub unsafe fn read_volatile<T>(src: *const T) -> T {
  // ..
}
```

```
kani verify-std ./library/ -Z unstable-options --target-dir target/verify-std
```

with Kani version:

I expected to see this happen: 0.52.0

Instead, this happened: Kani failed to compile the code:

```
error: duplicate diagnostic item in crate `core`: `ptr_read_volatile`.
    --> library/core/src/ptr/mod.rs:1693:1
     |
1693 | pub unsafe fn read_volatile<T>(src: *const T) -> T {
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
note: the diagnostic item is first defined here
    --> library/core/src/ptr/mod.rs:1689:1
     |
1689 | #[kani_core::requires(ub_checks::can_dereference(src))]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     = note: this error originates in the attribute macro `kani_core::requires` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### LINKED ISSUE #3254: Cannot verify a constant function with contracts
I tried this code:

```rust
#[kani::requires(true)]
#[inline]
const unsafe fn dummy<T>(arg: *const T) -> T {
    std::ptr::read(arg)
}

#[kani::proof_for_contract(dummy)]
fn check() {
    unsafe { dummy(&kani::any::<u8>()) };
}

```

```
kani -Z function-contracts dummy.rs
```

with Kani version:

I expected to see this happen: verification succeed

Instead, this happened: Kani failed to compile

celinval:
- Contracts cannot duplicate some attributes. I think we need a better solution than just duplicate all attributes, but for now just filter the ones we know are problematic.
  - Do not make contract generated functions `const` when annotating constant functions.
  - I also moved the compilation of no_core up since it is much faster.

## Call-out

Need to add a test.

Resolves #3251 
Resolves #3254 


library/kani_core/src/lib.rs
```diff
@@ -54,7 +61,7 @@ macro_rules! kani_lib {
 /// TODO: Use this inside kani library so that we dont have to maintain two copies of the same intrinsics.
 #[macro_export]
 macro_rules! kani_intrinsics {
-    () => {
+    ($core:tt) => {
```
jaisnan, line 64: I'm guess this still doesnt change how kani needs to be invoked from the std library or the user library?

library/kani_macros/src/sysroot/contracts/shared.rs
```diff
@@ -53,7 +53,34 @@ impl<'a> ContractConditionsHandler<'a> {
...
+                #[unstable(feature="kani", issue="none")]
+            ));
+            self.output.extend(
+                self.annotated_fn
+                    .attrs
+                    .iter()
+                    .filter(|attr| {
+                        if let Some(ident) = attr.path().get_ident() {
+                            let name = ident.to_string();
+                            !name.starts_with("rustc") && !(name == "stable")
```
jaisnan, line 76: Is there any chance of an attribute that is not this, but shares these properties?  I suppose we can always deal with that case later.

celinval: Yes. For sure there is! That's why I mentioned that we need to find a more robust way to do this. I was thinking about creating an associated function instead. @JustusAdam do you remember why we decided against it?

library/kani_core/src/lib.rs
celinval: No... I don't think so. Users are supposed to invoke `kani!()` macro, and that hasn't changed.

jaisnan (approved):
I'm approving this. I assume there might some other attributes that might cause issues, but we can always add more to the filter or improve the filter in any case.

celinval:
I think unstable might be one... not sure yet.

Issue #3182, opened 2024-05-14
Intrinsic `type_swapped` fails for unit types


celinval:
I tried this code:
```rust
//! swap.rs

#[kani::proof]
pub fn check_swap_unit() {
    let mut x: () = kani::any();
    let mut y: () = kani::any();
    std::mem::swap(&mut x, &mut y)
}

```

using the following command line invocation:

```
kani swap.rs
```

with Kani version: 0.51.0

I expected to see this happen: Verification succeeded

Instead, this happened: Verification failed

```
SUMMARY:
 ** 18 of 21 failed
Failed Checks: dereference failure: pointer NULL
Failed Checks: dereference failure: pointer invalid
Failed Checks: dereference failure: deallocated dynamic object
Failed Checks: dereference failure: dead object
Failed Checks: dereference failure: pointer outside object bounds
Failed Checks: dereference failure: invalid integer address
Failed Checks: dereference failure: pointer NULL
 File: "/home/ANT.AMAZON.COM/celinval/.rustup/toolchains/nightly-2024-04-21-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/mod.rs", line 731, in std::mem::swap::<()>
Failed Checks: dereference failure: pointer invalid
 File: "/home/ANT.AMAZON.COM/celinval/.rustup/toolchains/nightly-2024-04-21-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/mod.rs", line 731, in std::mem::swap::<()>
Failed Checks: dereference failure: deallocated dynamic object
 File: "/home/ANT.AMAZON.COM/celinval/.rustup/toolchains/nightly-2024-04-21-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/mod.rs", line 731, in std::mem::swap::<()>
Failed Checks: dereference failure: dead object
 File: "/home/ANT.AMAZON.COM/celinval/.rustup/toolchains/nightly-2024-04-21-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/mod.rs", line 731, in std::mem::swap::<()>
Failed Checks: dereference failure: pointer outside object bounds
[...]

PR #3256, opened 2024-06-11
Fix typed_swap for ZSTs

tautschnig:
typed_swap needs to be a no-op on ZSTs as pointers to those have an arbitrary value in Kani.

Resolves: #3182


tautschnig:
perf test failure is `inet::checksum::tests::differential` running out of memory (on this occasion).

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1991,30 +1991,42 @@ impl<'tcx> GotocCtx<'tcx> {
...
-        let non_overlapping_stmt =
-            Stmt::if_then_else(x.clone().same_object(y.clone()), non_overlapping_check, None, loc);
-
-        // T t = *y; *y = *x; *x = t;
         let deref_y = y.clone().dereference();
-        let (temp_var, assign_to_t) =
-            self.decl_temp_variable(deref_y.typ().clone(), Some(deref_y), loc);
-        let assign_to_y = y.dereference().assign(x.clone().dereference(), loc);
-        let assign_to_x = x.dereference().assign(temp_var, loc);
+        if deref_y.typ().sizeof(&self.symbol_table) == 0 {
```
adpaco-aws, line 1995: Should we use `is_zst(..)` here?

tautschnig: For all that I know `is_zst(..)` exists for MIR types only, and what we have is a pointer type where the ZST would be the pointed-to type. If anyone can teach me how to get the pointed-to type at MIR level then I'd happily use `is_zst(..)` (and we should then also consider doing so in other places).

adpaco-aws: There are actually a few `is_zst*` methods defined in `kani-compiler`. I think you can directly use [this one](https://github.com/model-checking/kani/blob/ca110f7b2c1d7cf219595863ef355713b306dc28/kani-compiler/src/codegen_cprover_gotoc/codegen/ty_stable.rs#L42-L44) in your code.

celinval: https://github.com/model-checking/kani/blob/ca110f7b2c1d7cf219595863ef355713b306dc28/kani-compiler/src/codegen_cprover_gotoc/codegen/ty_stable.rs#L125

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1991,30 +1991,42 @@ impl<'tcx> GotocCtx<'tcx> {
...
+            // do not attempt to dereference (and assign) a ZST
+            Stmt::skip(loc)
```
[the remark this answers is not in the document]
celinval: That's a great question. I was actually wondering that in the toolchain update PR (see [this comment](https://github.com/model-checking/kani/pull/3225#issuecomment-2159549709)).

I think we are incorrectly ignoring projections into a ZST. I think they can still trigger UB.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
tautschnig: Thank you!

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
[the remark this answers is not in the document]
celinval:
```rust

struct Dummy<T> {
    val: u8,
    last: T,
}

struct Wrapper<T>(Dummy<T>);

pub unsafe fn dumb<T: Copy>() -> T {
    let ptr: *const Wrapper<T> = std::ptr::NonNull::dangling().as_ptr();
    // Note that this will dereference `ptr` first. Gerating the following MIR:
    // _0 = (((*_1).0: Dummy<T>).1: T);
    unsafe { (*ptr).0.last }
}

/// This will trigger UB in Miri.
fn main() {
    unsafe { dumb::<()>() };
}
```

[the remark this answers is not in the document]
tautschnig: I have created #3262 to track this as this isn't a `typed_swap`-specific regression.

PR #3259, opened 2024-06-11
Fix contract of constant fn with effect feature

### LINKED ISSUE #3258: Cannot verify a constant function if crate has `effect` feature enabled
I tried this code:

```rust
#![feature(effects)]

#[kani::requires(kani::mem::can_dereference(arg))]
const unsafe fn dummy<T>(arg: *const T) -> T {
    std::ptr::read(arg)
}

#[kani::proof_for_contract(dummy)]
fn check() {
    unsafe { dummy(&kani::any::<u8>()) };
}
```

```
kani -Z function-contracts -Z mem-predicates contract.rs
```

with Kani version:

I expected to see this happen: Verification should pass;

Instead, this happened: Compilation fails

```
error: mismatch in the number of generic parameters: original function/method `dummy` takes 2 generic parameters(s), stub `dummy_recursion_wrapper_8b0f77` takes 1
  --> constant.rs:15:1
   |
15 | #[kani::proof_for_contract(dummy)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```

celinval:
We now check if the host effect index is present. If so, remove it before performing stub operation.

Resolves #3258


kani-compiler/src/kani_middle/stubbing/mod.rs
```diff
@@ -61,15 +79,26 @@ pub fn check_compatibility(tcx: TyCtxt, old_def: FnDef, new_def: FnDef) -> Resul
...
-    let stub_num_generics = tcx.generics_of(new_def_id).count();
-    if old_num_generics != stub_num_generics {
+    let old_ty = rustc_internal::stable(tcx.type_of(old_def_id)).value;
+    let new_ty = rustc_internal::stable(tcx.type_of(new_def_id)).value;
+    let TyKind::RigidTy(RigidTy::FnDef(_, mut old_args)) = old_ty.kind() else {
+        unreachable!("Expected Fn")
+    };
+    let TyKind::RigidTy(RigidTy::FnDef(_, new_args)) = new_ty.kind() else {
+        unreachable!("Expected Fn")
+    };
```
feliperodri, line 89: I'd include at least the `DefId` in the unreachable message.

kani-compiler/src/kani_middle/stubbing/mod.rs
```diff
@@ -61,15 +79,26 @@ pub fn check_compatibility(tcx: TyCtxt, old_def: FnDef, new_def: FnDef) -> Resul
...
+    if let Some(idx) = contract_host_param(tcx, old_def, new_def) {
+        old_args.0.remove(idx);
+    }
+
+    // TODO: We should check for the parameter type too or replacement will fail.
+    if old_args.0.len() != new_args.0.len() {
```
feliperodri, line 95: So right now this will only enable stubbing and checking function contracts, but not the use of stub verified? Should we track this on an issue?

feliperodri (approved):

kani-compiler/src/kani_middle/stubbing/mod.rs
celinval: No... we are enabling them all. I actually need to push a new PR where we re-export everything.

celinval: the problem here is that I don't know what will happen if we try the following stub pair:
```rust

fn orig<const C: usize>() -> usize {
    C
}

fn new<T: Into<usize> + Default>() -> usize {
    T::default().into()
}
```

kani-compiler/src/kani_middle/stubbing/mod.rs
```diff
@@ -61,15 +79,26 @@ pub fn check_compatibility(tcx: TyCtxt, old_def: FnDef, new_def: FnDef) -> Resul
     // Check whether the numbers of generic parameters match.
     let old_def_id = rustc_internal::internal(tcx, old_def.def_id());
     let new_def_id = rustc_internal::internal(tcx, new_def.def_id());
-    let old_num_generics = tcx.generics_of(old_def_id).count();
-    let stub_num_generics = tcx.generics_of(new_def_id).count();
-    if old_num_generics != stub_num_generics {
+    let old_ty = rustc_internal::stable(tcx.type_of(old_def_id)).value;
+    let new_ty = rustc_internal::stable(tcx.type_of(new_def_id)).value;
+    let TyKind::RigidTy(RigidTy::FnDef(_, mut old_args)) = old_ty.kind() else {
+        unreachable!("Expected Fn")
```
celinval, line 85:
```suggestion
        unreachable!("Expected function, but found {}", old_ty)
```

kani-compiler/src/kani_middle/stubbing/mod.rs
```diff
@@ -61,15 +79,26 @@ pub fn check_compatibility(tcx: TyCtxt, old_def: FnDef, new_def: FnDef) -> Resul
...
+    };
+    let TyKind::RigidTy(RigidTy::FnDef(_, new_args)) = new_ty.kind() else {
+        unreachable!("Expected Fn")
```
celinval, line 88:
```suggestion
        unreachable!("Expected function, but found {old_ty}")
```

PR #3264, opened 2024-06-12
Towards Proving Memory Initialization

artemagvanian:
This PR enables automatic memory initialization proofs for raw pointers in Kani. This is done without any extra instrumentation from the user. 

Currently, due to high memory consumption and only partial support of pointee types for which memory initialization proofs work, this feature is gated behind `-Z uninit-checks` flag. Note that because it uses shadow memory under the hood, programs using this feature need to pass `-Z ghost-state` flag as well.

This PR is a part of working towards #3300.


kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -290,14 +300,24 @@ impl GotocHook for PointerOffset {
...
+            let ptr = fargs.pop().unwrap();
+            let place_ty = instance.args().0[0].expect_ty().clone();
+            // Handle both fat and thin pointers.
+            if gcx.use_thin_pointer_stable(place_ty) {
+                ptr
+            } else {
+                ptr.member("data", &gcx.symbol_table)
+            }
+        }
+        .cast_to(Type::void_pointer());
```
celinval, line 320: can we do that as part of the kani library implementation using `to_raw_parts()`?

celinval (commented):
Can you please double check that invoking methods through the vtable works well? You can for example test drop of `Box<dyn Debug>` for cases where it is a valid value or it is an uninitialized value.

kani-compiler/src/kani_middle/transform/body.rs
```diff
@@ -182,42 +206,157 @@ impl MutableBody {
...
+    /// Add a new call to the basic block indicated by the given index.
+    ///
+    /// The new call will have the same span as the source instruction, and the basic block
+    /// will be split. The source instruction will be adjusted to point to the first instruction in
+    /// the new basic block.
+    pub fn add_call(
+        &mut self,
+        callee: &Instance,
+        source: &mut SourceInstruction,
+        position: InsertPosition,
```
celinval, line 218: What if we just merge these two? I.e.: Maybe make `InsertPosition` include the source instruction.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
celinval, line 1: I still need to finish reviewing this one.

kani-compiler/src/kani_middle/transform/kani_intrinsics.rs
```diff
@@ -120,11 +127,139 @@ impl IntrinsicGeneratorPass {
...
+                    InsertPosition::Before,
+                    result,
+                    &reason,
+                );
+            }
+        }
+        new_body.into()
+    }
+
+    fn is_initialized_body(&self, tcx: TyCtxt, body: Body) -> Body {
```
celinval, line 148: Can you please document this function? Maybe add a pseudo code of what this will look like.

kani-compiler/src/kani_middle/transform/mod.rs
```diff
@@ -64,6 +66,7 @@ impl BodyTransformation {
         // This has to come after stubs since we want this to replace the stubbed body.
         transformer.add_pass(queries, AnyModifiesPass::new(tcx, &unit));
         transformer.add_pass(queries, ValidValuePass { check_type: check_type.clone() });
+        transformer.add_pass(queries, UninitPass { check_type: check_type.clone() });
```
celinval, line 69: does the order matter?

celinval (commented):
That's amazing! Thank you

kani-compiler/src/kani_middle/transform/body.rs
```diff
@@ -35,6 +36,16 @@ pub struct MutableBody {
 
     /// The span that covers the entire function body.
     span: Span,
+
+    /// Set of basic block indices for which analyzing first statement should be skipped.
```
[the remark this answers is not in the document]
artemagvanian: Previously, all checks were inserted before the source instruction, which was in turn moved to the next basic block. Hence, we would not need to look at it again as a part of new basic block. Now, if the check is inserted after the source instruction, we still need to look at the first statement of the new basic block, so we need to keep track of which basic blocks were created as a part of injecting checks after the source instruction.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+    fn build_check(
+        &self,
+        tcx: TyCtxt,
+        body: &mut MutableBody,
+        instruction: InitRelevantInstruction,
+    ) {
+        debug!(?instruction, "build_check");
+        // Need to partition operations to make sure we add prefix operations before postfix operations
+        // to ensure instruction pointer shifts correctly.
+        let (operations_before, operations_after): (Vec<_>, Vec<_>) = instruction
```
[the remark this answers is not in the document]
artemagvanian: Not always. We have a couple of `bless` operations, which allow us change the memory initialization state before the source instruction. This is used, for example, for blessing statics

kani-compiler/src/kani_middle/transform/mod.rs
artemagvanian: I have not tested it directly, but I would be inclined to place `UninitPass` after `ValidValuePass`. Right now, `UninitPass` only instruments read and writes via raw pointers, and I don't think that `ValidValuePass` emits any code that uses raw pointers. Vice-versa, I could see `ValidValuePass` emitting additional checks in the code generated by `UninitPass`.

kani-compiler/src/kani_middle/transform/body.rs
artemagvanian: I added `InsertPosition` as a field to `SourceOp`

kani-compiler/src/kani_middle/mod.rs
```diff
@@ -116,7 +116,8 @@ pub fn dump_mir_items(
         let mut writer = BufWriter::new(out_file);
 
         // For each def_id, dump their MIR
-        for instance in items.iter().filter_map(get_instance) {
+        for instance in items.iter().filter_map(get_instance).filter(|instance| instance.has_body())
```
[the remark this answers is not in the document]
artemagvanian: Undid it

kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
```diff
@@ -118,6 +118,27 @@ impl<'tcx> GotocCtx<'tcx> {
             .typ
             .clone();
 
+        let shadow_memory_assign = self
```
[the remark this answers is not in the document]
artemagvanian: This is a noop if the static is not in scope because `.get(&rustc_span::symbol::Symbol::intern("KaniShadowMemory"))` returns an option

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+            .operations
+            .into_iter()
+            .partition(|operation| operation.should_be_inserted_before());
+        let operations: Vec<_> =
+            vec![operations_before, operations_after].into_iter().flatten().collect();
+
+        let mut source = instruction.source;
+        for operation in operations {
```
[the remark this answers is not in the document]
artemagvanian: Addressed that

library/kani/src/shadow.rs
```diff
@@ -81,3 +81,98 @@ impl<T: Copy> ShadowMem<T> {
         self.mem[obj][offset] = val;
     }
 }
+
+/// Global shadow memory object.
```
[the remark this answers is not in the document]
artemagvanian: Changed it

kani-compiler/src/kani_middle/transform/body.rs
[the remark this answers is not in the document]
celinval: I am not convinced this should be part of the `MutableBody` though. Isn't this a transformation pass responsibility?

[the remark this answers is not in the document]
artemagvanian: I was under the impression that depending on the basic block splitting logic, it would be hard to know ahead whether we need to skip the new block. But it seems to me that it's actually not the case, I will move that logic into the transformation pass.

library/kani/src/shadow.rs
```diff
@@ -81,3 +81,98 @@ impl<T: Copy> ShadowMem<T> {
...
+                    .set((ptr as *const u8).add(count * N + offset), value && layout[offset]);
+            }
+            offset += 1;
+        }
+        count += 1;
+    }
+}
+
+#[rustc_diagnostic_item = "KaniShadowMemoryGet"]
+pub fn __kani_global_sm_get<const N: usize, T: Sized>(
```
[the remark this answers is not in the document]
artemagvanian: Correct, will change that

library/kani/src/shadow.rs
```diff
@@ -81,3 +81,98 @@ impl<T: Copy> ShadowMem<T> {
...
+    value: bool,
+) {
+    let (ptr, _) = ptr.to_raw_parts();
+    __kani_global_sm_set_inner(ptr, layout, n, value);
+}
+
+// This method should only be called if T is known to be a slice.
+#[rustc_diagnostic_item = "KaniShadowMemoryGetSlice"]
+pub fn __kani_global_sm_get_slice<const N: usize, T: ?Sized>(
+    ptr: *const T,
```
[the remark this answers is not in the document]
artemagvanian: Changed it to `*const [T]` and added a separate function handling `*const str`

library/kani/src/shadow.rs
```diff
@@ -81,3 +81,98 @@ impl<T: Copy> ShadowMem<T> {
...
+    layout: [bool; N],
+    n: usize,
+) -> bool {
+    let (ptr, meta) = ptr.to_raw_parts();
+    let meta: usize = unsafe { std::mem::transmute_copy(&meta) };
```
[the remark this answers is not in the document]
artemagvanian: Got rid of that

library/kani/src/shadow.rs
```diff
@@ -81,3 +81,98 @@ impl<T: Copy> ShadowMem<T> {
         self.mem[obj][offset] = val;
     }
 }
+
+/// Global shadow memory object.
+#[rustc_diagnostic_item = "KaniShadowMemory"]
+static mut __KANI_GLOBAL_SM: ShadowMem<bool> = ShadowMem::new(false);
+
+// Get initialization setate of `n` items laid out according to the `layout` starting at address `ptr`.
+fn __kani_global_sm_get_inner<const N: usize>(ptr: *const (), layout: [bool; N], n: usize) -> bool {
```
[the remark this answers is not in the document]
artemagvanian: `n` is there for functions like `std::intrinsics::copy`, where there is a `count` attached, so more than one object is copied

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,275 @@
...
+            }
+        }
+        layout_mask
+    }
+}
+
+// Depending on whether the type is statically or dynamically sized,
+// the layout of the element or the layout of the actual type is returned.
+pub enum PointeeLayout {
+    Static { layout: TypeLayout },
```
celinval, line 41:
```suggestion
    /// Layout of sized objects
    Sized { layout: TypeLayout },
```
I think it makes sense to just use the Rust term for these things.

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,275 @@
...
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+struct DataBytes {
+    /// Offset in bytes.
+    offset: usize,
+    /// Size of this data chunk.
+    size: MachineSize,
+}
+
+pub struct TypeLayout {
```
celinval, line 19: Please document this structure. I'm assuming you are representing the non-padding bytes of a type.

celinval: BTW, do you even need this structure to be public? It seems that only `to_byte_mask()` is relevant, so why not expose the byte mask instead?

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,275 @@
...
+        }
+        layout_mask
+    }
+}
+
+// Depending on whether the type is statically or dynamically sized,
+// the layout of the element or the layout of the actual type is returned.
+pub enum PointeeLayout {
+    Static { layout: TypeLayout },
+    Slice { element_layout: TypeLayout },
```
celinval, line 42: Is `&str` included in this case?

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,275 @@
...
+                        Ok(PointeeInfo { pointee_ty: ty, layout })
+                    } else {
+                        Err(
+                            "Can only determine unsized type layout information for slices and trait objects.".to_string(),
+                        )
+                    }
+                }
+            },
+            TyKind::Alias(..) | TyKind::Param(..) | TyKind::Bound(..) => {
+                Err("Can only determine type layout information for RigidTy.".to_string())
```
celinval, line 91: Feel free to panic here. We shouldn't encounter these cases in monomorphic code.

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,275 @@
...
+                    if ty.layout().unwrap().shape().is_sized() {
+                        let size_in_bytes = ty.layout().unwrap().shape().size.bytes();
+                        let data_chunks = data_bytes_for_ty(&MachineInfo::target(), ty, 0)?;
+                        let layout = PointeeLayout::Static {
+                            layout: TypeLayout { size_in_bytes, data_chunks },
+                        };
+                        Ok(PointeeInfo { pointee_ty: ty, layout })
+                    } else {
+                        Err(
+                            "Can only determine unsized type layout information for slices and trait objects.".to_string(),
```
celinval, line 85:
```suggestion
                            format!("Cannot determine type layout for type `{ty}`"),
```

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,275 @@
...
+        &self.layout
+    }
+}
+
+/// Get a size of an initialized scalar.
+fn scalar_ty_size(machine_info: &MachineInfo, ty: Ty) -> Option<DataBytes> {
+    let shape = ty.layout().unwrap().shape();
+    match shape.abi {
+        ValueAbi::Scalar(Scalar::Initialized { value, .. })
+        | ValueAbi::ScalarPair(Scalar::Initialized { value, .. }, _) => {
```
celinval, line 110: Don't you need the second scalar in this case?

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,275 @@
...
+            vec![size]
+        } else {
+            vec![]
+        }
+    };
+    match layout.fields {
+        FieldsShape::Primitive => Ok(ty_size()),
+        FieldsShape::Array { stride, count } if count > 0 => {
+            let TyKind::RigidTy(RigidTy::Array(elem_ty, _)) = ty.kind() else { unreachable!() };
+            let elem_validity = data_bytes_for_ty(machine_info, elem_ty, current_offset)?;
```
celinval, line 140: please rename `elem_validity` and all the other `*validity` variables. :wink:

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,275 @@
...
+                                            let field_offset = offsets[field_idx].bytes();
+                                            let field_ty = fields[field_idx].ty_with_args(&args);
+                                            fields_validity.append(&mut data_bytes_for_ty(
+                                                machine_info,
+                                                field_ty,
+                                                field_offset + current_offset,
+                                            )?);
+                                        }
+                                    }
+                                    if fields_validity.is_empty() {
```
celinval, line 203: I think in your case you want to check if their padding byte locations are the same. I think you also want to add the data where the variant lives to make sure that is initialized.

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,275 @@
...
+                                            "Unsupported Enum `{}` check",
+                                            def.trimmed_name()
+                                        ))
+                                    }
+                                }
+                            }
+                        }
+                        AdtKind::Union => unreachable!(),
+                        AdtKind::Struct => {
+                            // If the struct range has niche add that.
```
celinval, line 216: remove this comment

kani-compiler/src/kani_middle/transform/kani_intrinsics.rs
```diff
@@ -120,11 +127,141 @@ impl IntrinsicGeneratorPass {
...
+        new_body.insert_stmt(stmt, &mut terminator, InsertPosition::Before);
+
+        // The first argument type.
+        let arg_ty = new_body.locals()[1].ty;
+        let TyKind::RigidTy(RigidTy::RawPtr(target_ty, _)) = arg_ty.kind() else { unreachable!() };
+        let pointee_info = PointeeInfo::from_ty(target_ty);
+        match pointee_info {
+            Ok(pointee_info) => {
+                match pointee_info.layout() {
+                    PointeeLayout::Static { layout } => {
```
celinval, line 182: Can we short circuit this if target_ty is a ZST?

kani-compiler/src/kani_middle/transform/kani_intrinsics.rs
```diff
@@ -120,11 +127,141 @@ impl IntrinsicGeneratorPass {
...
+                                GenericArgKind::Const(
+                                    TyConst::try_from_target_usize(
+                                        layout.to_byte_mask().len() as u64
+                                    )
+                                    .unwrap(),
+                                ),
+                                GenericArgKind::Type(*pointee_info.ty()),
+                            ]),
+                        )
+                        .unwrap();
```
celinval, line 195: I think every time you call `get_kani_sm_function` you are trying to resolve an instance and this chunk of code is replicated. Can you create a function that takes the `FnDef`, a `usize` and a `Ty` returns an instance instead?

kani-compiler/src/kani_middle/transform/kani_intrinsics.rs
```diff
@@ -120,11 +127,141 @@ impl IntrinsicGeneratorPass {
...
+                        );
+                        new_body.add_call(
+                            &shadow_memory_get_instance,
+                            &mut terminator,
+                            InsertPosition::Before,
+                            vec![Operand::Copy(Place::from(1)), layout_operand],
+                            Place::from(ret_var),
+                        );
+                    }
+                    PointeeLayout::TraitObject => unimplemented!(),
```
celinval, line 242: can you add an unsupported check like the one bellow?

celinval (commented):
I just need to review the instrumentation pass now. So far so good! Thanks

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+
+pub use ty_layout::{PointeeInfo, PointeeLayout, TypeLayout};
+use uninit_visitor::{CheckUninitVisitor, InitRelevantInstruction, SourceOp};
+
+const SKIPPED_DIAGNOSTIC_ITEMS: &[&str] = &["KaniMemInitSMGetInner", "KaniMemInitSMSetInner"];
+
+/// Retrieve a function definition by diagnostic string, caching the result.
+pub fn get_kani_sm_function(tcx: TyCtxt, diagnostic: &'static str) -> FnDef {
+    lazy_static! {
+        static ref KANI_SM_FUNCTIONS: Mutex<HashMap<&'static str, FnDef>> =
```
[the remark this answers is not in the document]
artemagvanian: Changed it. As a part of that, I had to make the `transform` function of each pass accept `self` as mutable, since the cache is updated as the pass progresses

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
artemagvanian: > BTW, do you even need this structure to be public? It seems that only `to_byte_mask()` is relevant, so why not expose the byte mask instead?

Good idea, did that ^

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
artemagvanian: Yep, added a comment to explain that. `&str` is treated as `&[u8]`

kani-compiler/src/kani_middle/transform/kani_intrinsics.rs
artemagvanian: I think I meant to just leave that case empty, sorry. We generate a failing assertion for any unsize cast right now, so this case presumably won't even happen.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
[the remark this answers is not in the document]
celinval: I was thinking that you could cache all functions you need as part of the pass constructor. But that should be fine too.

kani-compiler/src/kani_middle/transform/kani_intrinsics.rs
celinval: If that's the case, please replace this by `unreachable!()`.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+        instruction: InitRelevantInstruction,
+    ) {
+        debug!(?instruction, "build_check");
+        // Need to partition operations to make sure we add prefix operations before postfix operations
+        // to ensure instruction pointer shifts correctly.
+        let (operations_before, operations_after): (Vec<_>, Vec<_>) = instruction
+            .operations
+            .into_iter()
+            .partition(|operation| operation.should_be_inserted_before());
+        let operations: Vec<_> =
```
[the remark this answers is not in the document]
artemagvanian: That part got refactored, so this no longer applies

library/kani/src/shadow.rs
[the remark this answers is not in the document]
artemagvanian: Good idea. I experimented with representing type layout as a `u64` or `u128` and it seemed to perform better. In any case -- I will work on that in subsequent PRs.

library/kani/src/shadow.rs
```diff
@@ -81,3 +81,117 @@ impl<T: Copy> ShadowMem<T> {
         self.mem[obj][offset] = val;
     }
 }
+
```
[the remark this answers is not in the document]
artemagvanian: Moved memory initialization checking code into a separate module. 

Talked about the unit tests offline. They are hard to implement because all the calls from the added functions lead to `kani_intrinsic`. In theory, we could shim the intrinsics, but decided not to do it right now

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
artemagvanian: Added a comment addressing that

kani-compiler/src/kani_middle/transform/kani_intrinsics.rs
artemagvanian: Talked offline. Added an `assert(false)` for cases where a pointer to trait objects is used

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
artemagvanian: Talked about it offline. Now it uses both of the scalars for fat pointers.

library/kani/src/shadow.rs
```diff
@@ -81,3 +81,98 @@ impl<T: Copy> ShadowMem<T> {
...
+) -> bool {
+    let (ptr, _) = ptr.to_raw_parts();
+    __kani_global_sm_get_inner(ptr, layout, n)
+}
+
+#[rustc_diagnostic_item = "KaniShadowMemorySet"]
+pub fn __kani_global_sm_set<const N: usize, T: Sized>(
+    ptr: *const T,
+    layout: [bool; N],
+    n: usize,
```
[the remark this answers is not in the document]
artemagvanian: I am planning to work on it in a follow-up PR, as this might change depending on how we choose to represent the layout

library/kani/src/mem_init.rs
```diff
@@ -0,0 +1,122 @@
...
+/// Get initialization state of `len` items laid out according to the `layout` starting at address `ptr`.
+#[rustc_diagnostic_item = "KaniIsUnitPtrInitialized"]
+fn is_unit_ptr_initialized<const N: usize>(ptr: *const (), layout: Layout<N>, len: usize) -> bool {
+    let mut count: usize = 0;
+    while count < len {
+        let mut offset: usize = 0;
+        while offset < N {
+            unsafe {
+                if layout[offset]
+                    && !MEM_INIT_SHADOW_MEM.get((ptr as *const u8).add(count * N + offset))
```
remi-delmas-3000, line 43: to avoid multiple lookups into the shadow map per check, could you instead rebase `ptr` to the shadow object and then iterate on the shadow object bytes directly by bumping the pointer ?

library/kani/src/mem_init.rs
```diff
@@ -0,0 +1,122 @@
...
+    len: usize,
+    value: bool,
+) {
+    let mut count: usize = 0;
+    while count < len {
+        let mut offset: usize = 0;
+        while offset < N {
+            unsafe {
+                MEM_INIT_SHADOW_MEM
+                    .set((ptr as *const u8).add(count * N + offset), value && layout[offset]);
```
remi-delmas-3000, line 69: here we should try to get rid of the loop using a bulk assignment `__CPROVER_array_update` or somthing like that

library/kani/src/mem_init.rs
artemagvanian: Thanks for the suggestion! I have a follow-up PR that represents the layout as a `u128` and, I believe, solves this problem

library/kani/src/mem_init.rs
artemagvanian: As per the previous reply, this array will be removed in the subsequent PR

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -0,0 +1,423 @@
...
+                self.build_set(tcx, body, source, operation, pointee_ty_info, skip_first)
+            }
+            SourceOp::Unsupported { .. } => {
+                unreachable!()
+            }
+        }
+    }
+
+    // Inject a load from shadow memory tracking memory initialization and an assertion that all
+    // non-padding bytes are initialized.
```
celinval, line 175: please use `///` for function documentation.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -0,0 +1,423 @@
...
+            PointeeLayout::TraitObject => {
+                try_mark_new_bb_as_skipped(&operation, body, skip_first);
+                let reason = "Kani does not support reasoning about memory initialization of pointers to trait objects.";
+                self.unsupported_check(tcx, body, source, operation.position(), reason);
+                return;
+            }
+        };
+
+        // Make sure all non-padding bytes are initialized.
+        try_mark_new_bb_as_skipped(&operation, body, skip_first);
```
celinval, line 243: nit: A function with `try` in its name means that it is a fallible operation. Something like `collect_skipped` might make more sense.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -0,0 +1,423 @@
...
+                            const_: MirConst::from_bool(value),
+                        }),
+                    ],
+                    ret_place,
+                );
+            }
+            PointeeLayout::TraitObject => {
+                try_mark_new_bb_as_skipped(&operation, body, skip_first);
+                let reason = "Kani does not support reasoning about memory initialization of pointers to trait objects.";
+                self.unsupported_check(tcx, body, source, operation.position(), reason);
```
celinval, line 337: Do we actually need this one? How can one directly change the initialization state of a TraitObject?

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -0,0 +1,423 @@
...
+            const_: MirConst::from_bool(false),
+            span,
+            user_ty: None,
+        }));
+        let result = body.new_assignment(rvalue, source, position);
+        body.add_check(tcx, &self.check_type, source, position, result, reason);
+    }
+}
+
+/// Generate a bit array denoting padding vs data bytes for a layout.
```
celinval, line 361: Can you be a bit more specific? Maybe an example.

I'm assuming this is generating an array where `false` means padding and `true` means data?

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,328 @@
...
+/// Represents a chunk of data bytes in a data structure.
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+struct DataBytes {
+    /// Offset in bytes.
+    offset: usize,
+    /// Size of this data chunk.
+    size: MachineSize,
+}
+
+/// Bytewise mask, representing which bytes of a type are data and which are padding.
```
celinval, line 20: Where `false` represents padding and `true` represents data.

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,328 @@
...
+        ValueAbi::Scalar(Scalar::Initialized { value, .. }) => {
+            DataBytes { offset: current_offset, size: value.size(machine_info) }
+        }
+        ValueAbi::ScalarPair(
+            Scalar::Initialized { value: value_first, .. },
+            Scalar::Initialized { value: value_second, .. },
+        ) => DataBytes {
+            offset: current_offset,
+            size: MachineSize::from_bits(
+                value_first.size(machine_info).bits() + value_second.size(machine_info).bits(),
```
celinval, line 120: What if these are not contiguous? Don't we need to generate two `DataBytes`?

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
```diff
@@ -0,0 +1,328 @@
...
+                                    let tag_size = match tag {
+                                        Scalar::Initialized { value, .. } => {
+                                            value.size(&machine_info)
+                                        }
+                                        Scalar::Union { .. } => {
+                                            unreachable!("Enum tag should not be a union.")
+                                        }
+                                    };
+                                    let tag_data_bytes =
+                                        vec![DataBytes { offset: current_offset, size: tag_size }];
```
celinval, line 202: Is it safe to assume that the tag is always at offset 0?

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,783 @@
...
+use stable_mir::mir::{
+    BasicBlockIdx, CastKind, ConstOperand, LocalDecl, MirVisitor, Mutability,
+    NonDivergingIntrinsic, Operand, Place, PointerCoercion, ProjectionElem, Rvalue, Statement,
+    StatementKind, Terminator, TerminatorKind,
+};
+use stable_mir::ty::{ConstantKind, MirConst, RigidTy, Span, TyKind, UintTy};
+use strum_macros::AsRefStr;
+
+#[derive(AsRefStr, Clone, Debug)]
+pub enum SourceOp {
```
celinval, line 19: please document this enum and its variants.

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,783 @@
...
+    Get { operand: Operand, count: Operand, position: InsertPosition },
+    Set { operand: Operand, count: Operand, value: bool, position: InsertPosition },
+    SetRef { operand: Operand, count: Operand, value: bool, position: InsertPosition },
+    Unsupported { reason: String, position: InsertPosition },
```
celinval, line 23: Do we need an insert position for this?

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,783 @@
...
+    BasicBlockIdx, CastKind, ConstOperand, LocalDecl, MirVisitor, Mutability,
+    NonDivergingIntrinsic, Operand, Place, PointerCoercion, ProjectionElem, Rvalue, Statement,
+    StatementKind, Terminator, TerminatorKind,
+};
+use stable_mir::ty::{ConstantKind, MirConst, RigidTy, Span, TyKind, UintTy};
+use strum_macros::AsRefStr;
+
+#[derive(AsRefStr, Clone, Debug)]
+pub enum SourceOp {
+    Get { operand: Operand, count: Operand, position: InsertPosition },
```
celinval, line 20: Shouldn't position for `Get` always be Before?

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,783 @@
...
+                                | "maxnumf32"
+                                | "maxnumf64"
+                                | "min_align_of"
+                                | "min_align_of_val"
+                                | "minnumf32"
+                                | "minnumf64"
+                                | "mul_with_overflow"
+                                | "nearbyintf32"
+                                | "nearbyintf64"
+                                | "needs_drop" => {}
```
celinval, line 493: I really appreciate the thoroughness. May I suggest that you move the cases that we don't have to do any work to a validation function? Maybe you can change the default cases to be something like:
```rust
intrinsic_name if unsupported_intrinsic(intrinsic_name) => {
        // and unsupported operation instead
    }
_ => { /* no-op  */ }
```
Where `unsupported_intrinsic()` checks if the name does not match any of the names you have listed here and in other places. BTW, I'm not great at names, so feel free to rename the function

celinval (commented):
Almost done. Just need to finish reviewing the visitor.

tests/expected/uninit/alloc-to-slice/alloc-to-slice.rs
```diff
@@ -0,0 +1,26 @@
...
+        let ptr = alloc(layout);
+        *ptr = 0x41;
+        *ptr.add(1) = 0x42;
+        *ptr.add(2) = 0x43;
+        *ptr.add(3) = 0x44;
+        *ptr.add(16) = 0x00;
+        // Forming a slice from unitialized memory is UB.
+        let slice1 = from_raw_parts(ptr, 16);
+        let slice2 = from_raw_parts(ptr.add(16), 16);
+        drop(slice1.cmp(slice2));
```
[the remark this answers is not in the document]
artemagvanian: This is unnecessary, you are right -- removed it

tests/expected/uninit/alloc-to-slice/alloc-to-slice.rs
```diff
@@ -0,0 +1,26 @@
...
+fn main() {
+    let layout = Layout::from_size_align(32, 8).unwrap();
+    unsafe {
+        let ptr = alloc(layout);
+        *ptr = 0x41;
+        *ptr.add(1) = 0x42;
+        *ptr.add(2) = 0x43;
+        *ptr.add(3) = 0x44;
+        *ptr.add(16) = 0x00;
+        // Forming a slice from unitialized memory is UB.
```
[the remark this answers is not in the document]
artemagvanian: I've added comments. Which multiple harnesses did you have in mind?

tests/expected/uninit/struct-padding-and-arr-uninit/struct-padding-and-arr-uninit.rs
```diff
@@ -0,0 +1,32 @@
...
+#![feature(custom_mir)]
+
+use std::intrinsics::mir::*;
+use std::ptr;
+
+#[repr(C)]
+struct S(u8, u16);
+
+#[kani::proof]
+#[custom_mir(dialect = "runtime", phase = "optimized")]
```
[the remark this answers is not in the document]
artemagvanian: Nope -- the original test from MIRI used `custom_mir`, I removed it now

tests/std-checks/core/src/slice.rs
```diff
@@ -0,0 +1,32 @@
...
+    use super::*;
+
+    const LEN_MIN: usize = 1;
+    const LEN_MAX: usize = 2;
+
+    #[kani::proof_for_contract(contracts::from_raw_parts)]
+    #[kani::unwind(25)]
+    pub fn check_from_raw_parts_primitive() {
+        let len: usize = kani::any();
+        kani::assume(len >= LEN_MIN);
```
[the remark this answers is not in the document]
artemagvanian: If `len == 0`, the following assertion fails:
```
 File: ".../kani/library/kani/src/mem.rs", line 171, in kani::mem::is_inbounds::<(), u8>
```
However, expressing the precondition as:
```
#[requires(can_dereference(std::ptr::slice_from_raw_parts(data, len)))]
```
solves the issue.

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
artemagvanian: `scalar_ty_size` is only used for determining sizes of primitives and references / pointers. I believe `ValueAbi::ScalarPair` could not occur in the former case, and it is contiguous in the latter case.

That makes me think: should we just "inline" this function in two places and remove unnecessary match arms for each place?

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
artemagvanian: Not if the enum is a field in another struct, for example

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
artemagvanian: I don't think it matters, I can remove it

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
artemagvanian: I can't think of an example where it wouldn't be Before. I will assume it is always Before for now.

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
celinval: Inline makes sense. You could also add an assertion that the data is contiguous and add the your explanation as a comment.

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
celinval: I mean, 0 for the current type (i.e. `current_offset`). Can you add an assertion here too?

artemagvanian: Added an assertion

kani-compiler/src/kani_middle/transform/check_uninit/ty_layout.rs
artemagvanian: I inlined it into two places and actually retrieved the offsets in the pointer / ref case without assuming it is contiguous

tests/expected/uninit/alloc-to-slice/alloc-to-slice.rs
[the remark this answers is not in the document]
artemagvanian: I simplified the harness for now so that it finishes on line 21.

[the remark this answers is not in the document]
celinval: Perfect

celinval (commented):
So close!! Can you please add a few tests for catching intrinsics misuse?

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,812 @@
...
+                        if place_without_deref.ty(&self.locals).unwrap().kind().is_raw_ptr() {
+                            self.push_target(MemoryInitOp::Set {
+                                operand: Operand::Copy(place_without_deref),
+                                count: mk_const_operand(1, location.span()),
+                                value: true,
+                                position: InsertPosition::After,
+                            });
+                        }
+                    }
+                    // Check whether Rvalue creates a new initialized pointer previously not captured inside shadow memory.
```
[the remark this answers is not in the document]
artemagvanian: Whenever a raw pointer is created via an `AddressOf`:
```rust
let value: u8 = 42;
let ptr = std::ptr::addr_of!(value);
```

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,812 @@
...
+                                    assert!(matches!(
+                                        args[0].ty(self.locals).unwrap().kind(),
+                                        TyKind::RigidTy(RigidTy::RawPtr(_, Mutability::Mut))
+                                    ));
+                                    self.push_target(MemoryInitOp::Set {
+                                        operand: args[0].clone(),
+                                        count: mk_const_operand(1, location.span()),
+                                        value: true,
+                                        position: InsertPosition::After,
+                                    });
```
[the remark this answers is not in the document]
artemagvanian: Ah, I see -- thanks for catching this.

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,812 @@
...
+                                }
+                                "atomic_xchg_seqcst"
+                                | "atomic_xchg_acquire"
+                                | "atomic_xchg_acqrel"
+                                | "atomic_xchg_release"
+                                | "atomic_xchg_relaxed"
+                                | "atomic_store_seqcst"
+                                | "atomic_store_release"
+                                | "atomic_store_relaxed"
+                                | "atomic_store_unordered" => {
```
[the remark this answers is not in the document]
artemagvanian: Not anymore since it was changed to `Get`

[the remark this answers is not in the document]
artemagvanian: Will merge those ^

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,812 @@
...
+                                        count: args[2].clone(),
+                                    });
+                                    self.push_target(MemoryInitOp::Set {
+                                        operand: args[1].clone(),
+                                        count: args[2].clone(),
+                                        value: true,
+                                        position: InsertPosition::After,
+                                    });
+                                }
+                                "volatile_load" => {
```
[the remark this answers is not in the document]
artemagvanian: I reorganized intrinsics a bit, so I merged all `atomic` intrinsics altogether instead since all of them are now `Get`

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,812 @@
...
+                        }
+                        _ => {}
+                    }
+                }
+                TerminatorKind::Drop { place, .. } => {
+                    self.super_terminator(term, location);
+                    let place_ty = place.ty(&self.locals).unwrap();
+                    // When drop is codegen'ed, a reference is taken to the place which is later implicitly coerced to a pointer.
+                    // Hence, we need to bless this pointer as initialized.
+                    self.push_target(MemoryInitOp::SetRef {
```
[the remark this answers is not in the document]
artemagvanian: `SetRef`is actually a `Set` that applies to the pointer to the place, not the place itself. This is needed because when drop is codegen'd, a reference to the place is coerced under the hood to a pointer, which, without this blessing, will fail a check assertion upon access.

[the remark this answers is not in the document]
artemagvanian: Renamed `Get` to `Check` though

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,812 @@
...
+                        });
+                    };
+                }
+            }
+        }
+        self.super_operand(operand, location);
+    }
+
+    fn visit_rvalue(&mut self, rvalue: &Rvalue, location: Location) {
+        if let Rvalue::Cast(CastKind::PointerCoercion(PointerCoercion::Unsize), _, _) = rvalue {
```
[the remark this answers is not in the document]
artemagvanian: I added an extra check to only add `Unsupported` if the cast is to a pointer to a trait object

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,812 @@
...
+                        count: copy.count.clone(),
+                        value: true,
+                        position: InsertPosition::After,
+                    });
+                }
+                StatementKind::Assign(place, rvalue) => {
+                    // First check rvalue.
+                    self.visit_rvalue(rvalue, location);
+                    // Check whether we are assigning into a dereference (*ptr = _).
+                    if let Some(place_without_deref) = try_remove_topmost_deref(place) {
```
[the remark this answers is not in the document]
artemagvanian: Good point. I assumed they would always be split, but I see your point. I added extra logic to check for that.

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,712 @@
...
+                        // First, check that we are not dereferencing extra pointers along the way
```
celinval, line 176: BTW, you could reuse the `visit_place` code. But not a blocker.

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -0,0 +1,712 @@
...
+                | StatementKind::Coverage(_)
+                | StatementKind::ConstEvalCounter
+                | StatementKind::Intrinsic(NonDivergingIntrinsic::Assume(_))
+                | StatementKind::Nop => self.super_statement(stmt, location),
+            }
+        }
+        let SourceInstruction::Statement { idx, bb } = self.current else { unreachable!() };
+        self.current = SourceInstruction::Statement { idx: idx + 1, bb };
+    }
+    fn visit_terminator(&mut self, term: &Terminator, location: Location) {
```
celinval, line 241:
```suggestion
    }

    fn visit_terminator(&mut self, term: &Terminator, location: Location) {
```

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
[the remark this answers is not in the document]
celinval: I'm still confused about this one. Maybe we can check offline. You just added this comment that makes it sound like a check to me:
> ```rust
>    /// Check memory initialization of data bytes in a memory region starting from the reference to
>    /// `operand` and of length `count * sizeof(operand)` bytes.
> ```

celinval (approved):
Thank you

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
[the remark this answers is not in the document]
artemagvanian: My bad, the comment is incorrect. "Check" here should mean "Set". Fixed it in the last commit.

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
artemagvanian: You're right -- I'll fix it in the follow-up PR

Issue #3027, opened 2024-02-13
Contracts: Can't include non-`Copy` types in contract


adpaco-aws:
When trying out the contracts feature (from Kani 0.46.0) in one of my projects, I got the following errors:

```
error[E0507]: cannot move out of `*e_renamed` which is behind a shared reference
  --> src/lib.rs:79:5
   |
79 |     #[kani::ensures(match result { Ok(t) => Self::typecheck_2(*e, t), Err(_) => true})]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ move occurs because `*e_renamed` has type `Expr`, which does not implement the `Copy` trait
   |
   = note: this error originates in the attribute macro `kani::ensures` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0507]: cannot move out of `*e_renamed` which is behind a shared reference
  --> src/lib.rs:79:5
   |
79 |     #[kani::ensures(match result { Ok(t) => Self::typecheck_2(*e, t), Err(_) => true})]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ move occurs because `*e_renamed` has type `Expr`, which does not implement the `Copy` trait
   |
   = note: this error originates in the attribute macro `kani::ensures` (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0507`.
error: could not compile `interpreter` (lib) due to 3 previous errors
error: Failed to execute cargo (exit status: 101). Found 3 compilation errors.
```

I've opened #3026 for improving the error message, but anyway it's clear to me that it's asking for `Expr` to implement `Copy`. In my case, `Expr` is an `enum` defined as follows:

```rs
pub enum Expr {
    Int(i32),
    Bool(bool),
    If {
        test_expr: Box<Expr>,
        then_expr: Box<Expr>,
        else_expr: Box<Expr>,
    },
    BinaryApp {
        op: BinaryOp,
        arg1: Box<Expr>,
        arg2: Box<Expr>,
    },
}
```

[...]

PR #3274, opened 2024-06-18
Function Contracts: remove instances of _renamed

### LINKED ISSUE #3026: Contracts: Error refers to nonvisible `*e_renamed` expression

### LINKED ISSUE #3239: Contracts: macro allowing for contraction of specific variables
Let's say we are verifying function `f` with conditions `P` and `Q` where `P` can depend on the arguments of the function and `Q` can depend on those arguments, along with past computations and the result of the computation.

```rust
#[kani::requires(P(arg1,arg2))]
#[kani::ensures(Q(result)(old_comp(arg1,arg2))(arg1,arg2))]
fn f(arg1, arg2){
    ...
}
```

This will expand to the format of

```rust
fn f(arg1, arg2){
    ...
}
fn check_f(arg1, arg2){
    assume(P(arg1,arg2));
    let old_value = old_comp(arg1,arg2);
    let result = f(arg1,arg2);
    assert(Q(result)(old_value)(arg1,arg2));
    result
}
```

In this setting, we can clearly see that each line uses `arg1` and `arg2`. This is worrying in a type system such as rust.

Rust's borrow checker is designed with affine logic, which means that once a variable is used, it cannot be used again (at most once). It is clear that we are using the variable 4 separate times here, and 4>1.

In the event that all arguments implement a copy or clone, then this issue is dodged as we clone the variable in each separate usage of it, never consuming the resource. This is too restrictive as we want to reason about functions `f` that potentially take ownership of the resources of the arguments.

Affine logic is a substructural logic where you take structural logic and remove the "contraction" rule. Contraction allows for `let y,z = x;` which performs a clone of the `x` into two separate `y` and `z` variables.

We currently see this being implemented within the contract macro expansion via the
```let x_renamed = kani::internal::untracked_deref(&x);```
As this statement lets us convert the single reference of `x` into two references `x` and `x_renamed`, so when `f` consumes the `x` resource, the ensures is still able to refer to `x_renamed`, bypassing the affine logic of the rust borrow checker.

Since we are only doing this contraction once, we only have 2 copies of the arguments to the function. One copy (the `x_renamed`) is being given to the `Q` and `old_comp`, while another copy (the `x`) is being given to the `P` and `f`. This means that if any of the `old_comp` were to consume the resource, and we later want the resource in `Q`, then we get the error of `error[E0382]: borrow of moved value: x_renamed`. Likewise, if `P` consumes the resource, then we will be unable to give the resource to function `f`.

A good example is the following code:

```rust
struct NoCopy<T>(T);

impl<T: kani::Arbitrary> kani::Arbitrary for NoCopy<T> {
    fn any() -> Self {
        Self(kani::any())
    }
}

#[kani::modifies(a)]
#[kani::ensures(|result| old(*a).0.wrapping_add(1) == a.0)]
fn add1(a : &mut NoCopy<u32>){
    *a=NoCopy(a.0.wrapping_add(1))
}

#[kani::proof_for_contract(add1)]
fn verify_success() {
    let a : &mut NoCopy<u32> = &mut kani::any();
    add1(a);
}
```

Here, we specifically force the `a` to not have a copy function. The old computation `old(*a)` thus needs to acquire the resource of `a_renamed` and this causes conflict when we want to use `a_renamed` within the `|result| old(*a).0 + 1 == a.0` ensures clause.

If we change the ensures clause to `#[kani::ensures(|result| old(a.0).wrapping_add(1) == a.0)]` then the `old` expression computes a value that has a copy and the test passes.

There are two options for how to proceed. One way would be to force users to comply with rust semantics so that no borrows of move semantics occur. This would restrict how people use the function arguments in their requires and ensures and `old` monad because they could each potentially move the arguments and cause a compilation error.

Another way is to give up on the rust affine logic and allow for specific arguments to be lifted into structural logic, thus allowing for contraction.

The original example becomes something like the following

```rust
fn f(arg1, arg2){
    ...
}
fn check_f(arg1, arg2){
    contraction!(arg1);
    contraction!(arg2);
    assume(P(arg1,arg2));
    let old_value = old_comp(arg1,arg2);
    let result = f(arg1,arg2);
    assert(Q(result)(old_value)(arg1,arg2));
    result
}
```

We can then expand these contractions to

```rust
fn f(arg1, arg2){
    ...
}
fn check_f(arg1, arg2){
    let arg1_box = arg1;
    let arg2_box = arg2;

    let arg1 = kani::internal::untracked_deref(&arg1_box);
    let arg2 = kani::internal::untracked_deref(&arg2_box);
    assume(P(arg1,arg2));
    free(arg1);
    free(arg2);

    let arg1 = kani::internal::untracked_deref(&arg1_box);
    let arg2 = kani::internal::untracked_deref(&arg2_box);
    let old_value = old_comp(arg1,arg2);
    free(arg1);
    free(arg2);

    let arg1 = kani::internal::untracked_deref(&arg1_box);
    let arg2 = kani::internal::untracked_deref(&arg2_box);
    let result = f(arg1,arg2);
    free(arg1);
    free(arg2);

    let arg1 = kani::internal::untracked_deref(&arg1_box);
    let arg2 = kani::internal::untracked_deref(&arg2_box);
    assert(Q(result)(old_value)(arg1,arg2));
    free(arg1);
    free(arg2);

    result
}
```
With this, every time we create a new line that can consume `arg`, we create a copy of it.

It is suggested that we merge #3232 before this change as this is a tangential issue. The #3232 PR is failing on the case where `old_comp` moves the argument as seen in the conversation. This new change would resolve that issue and add the functionality for capturing the argument within the contract propositions.

This change would also subsume the changes within #3234. That PR would no longer be required and this new change would also solve issue #3026.

pi314mm:
The current method for creating the modifies wrapper requires changing the `ensures` clause to have `_renamed` variables which are unsafe copies of the original function arguments. This causes issues with regards to some possible tests as in #3239.

This change removes the `_renamed` variables and instead simply changes the modifies clauses within the replace to unsafely dereference the pointer to modify the contents of it unsafely, condensing all instances of unsafe Rust into a single location.

Resolves #3239
Resolves #3026
May affect #3027. In my attempt to run this example with slight modification to fit the current implementation, I got the error `CBMC appears to have run out of memory. You may want to rerun your proof in an environment with additional memory or use stubbing to reduce the size of the code the verifier reasons about.` This suggests that the compilation is working appropriately but the test case is just too large for CBMC.

### DIFF
--- a/library/kani_macros/src/sysroot/contracts/mod.rs
+++ b/library/kani_macros/src/sysroot/contracts/mod.rs
@@ -242,8 +242,7 @@
 //! state. Each occurrence of `old` is lifted, so is is necessary that
 //! each lifted occurrence is closed with respect to the function arguments.
 //! The results of these old computations are placed into
-//! `remember_kani_internal_XXX` variables of incrementing index to avoid
-//! collisions of variable names. Consider the following example:
+//! `remember_kani_internal_XXX` variables which are hashed. Consider the following example:
 //!
 //! ```
 //! #[kani::ensures(|result| old(*ptr + 1) == *ptr)]
@@ -265,49 +264,72 @@
 //! This expands to
 //!
 //! ```
+//! #[kanitool::checked_with = "modify_recursion_wrapper_633496"]
+//! #[kanitool::replaced_with = "modify_replace_633496"]
+//! #[kanitool::inner_check = "modify_wrapper_633496"]
+//! fn modify(ptr: &mut u32) { { *ptr += 1; } }
+//! #[allow(dead_code, unused_variables, unused_mut)]
+//! #[kanitool::is_contract_generated(recursion_wrapper)]
+//! fn modify_recursion_wrapper_633496(arg0: &mut u32) {
+//!     static mut REENTRY: bool = false;
+//!     if unsafe { REENTRY } {
+//!             modify_replace_633496(arg0)
+//!         } else {
+//!            unsafe { REENTRY = true };
+//!            let result_kani_internal = modify_check_633496(arg0);
+//!            unsafe { REENTRY = false };
+//!            result_kani_internal
+//!        }
+//! }
 //! #[allow(dead_code, unused_variables, unused_mut)]
 //! #[kanitool::is_contract_generated(check)]
 //! fn modify_check_633496(ptr: &mut u32) {
 //!     let _wrapper_arg_1 =
 //!         unsafe { kani::internal::Pointer::decouple_lifetime(&ptr) };
 //!     kani::assume(*ptr < 100);
-//!     let remember_kani_internal_1 = *ptr + 1;
-//!     let ptr_renamed = kani::internal::untracked_deref(&ptr);
-//!     let remember_kani_internal_0 = *ptr + 1;
-//!     let ptr_renamed = kani::internal::untracked_deref(&ptr);
+//!     let remember_kani_internal_92cc419d8aca576c = *ptr + 1;
+//!     let remember_kani_internal_92cc419d8aca576c = *ptr + 1;
 //!     let result_kani_internal: () = modify_wrapper_633496(ptr, _wrapper_arg_1);
 //!     kani::assert((|result|
-//!                     (remember_kani_internal_0) ==
-//!                         *ptr_renamed)(&result_kani_internal),
+//!                     (remember_kani_internal_92cc419d8aca576c) ==
+//!                         *ptr)(&result_kani_internal),
 //!         "|result| old(*ptr + 1) == *ptr");
-//!     std::mem::forget(ptr_renamed);
 //!     kani::assert((|result|
-//!                     (remember_kani_internal_1) ==
-//!                         *ptr_renamed)(&result_kani_internal),
+//!                     (remember_kani_internal_92cc419d8aca576c) ==
+//!                         *ptr)(&result_kani_internal),
 //!         "|result| old(*ptr + 1) == *ptr");
-//!     std::mem::forget(ptr_renamed);
 //!     result_kani_internal
 //! }
 //! #[allow(dead_code, unused_variables, unused_mut)]
 //! #[kanitool::is_contract_generated(replace)]
 //! fn modify_replace_633496(ptr: &mut u32) {
 //!     kani::assert(*ptr < 100, "*ptr < 100");
-//!     let remember_kani_internal_1 = *ptr + 1;
-//!     let ptr_renamed = kani::internal::untracked_deref(&ptr);
-//!     let remember_kani_internal_0 = *ptr + 1;
-//!     let ptr_renamed = kani::internal::untracked_deref(&ptr);
+//!     let remember_kani_internal_92cc419d8aca576c = *ptr + 1;
+//!     let remember_kani_internal_92cc419d8aca576c = *ptr + 1;
 //!     let result_kani_internal: () = kani::any_modifies();
-//!     *unsafe { kani::internal::Pointer::assignable(ptr) } =
-//!         kani::any_modifies();
+//!     *unsafe {
+//!                 kani::internal::Pointer::assignable(kani::internal::untracked_deref(&(ptr)))
+//!             } = kani::any_modifies();
 //!     kani::assume((|result|
-//!                     (remember_kani_internal_0) ==
-//!                         *ptr_renamed)(&result_kani_internal));
-//!     std::mem::forget(ptr_renamed);
+//!                     (remember_kani_internal_92cc419d8aca576c) ==
+//!                         *ptr)(&result_kani_internal));
 //!     kani::assume((|result|
-//!                     (remember_kani_internal_1) ==
-//!                         *ptr_renamed)(&result_kani_internal));
-//!     std::mem::forget(ptr_renamed);
+//!                     (remember_kani_internal_92cc419d8aca576c) ==
+//!                         *ptr)(&result_kani_internal));
 //!     result_kani_internal
+//! }
+//! #[kanitool::modifies(_wrapper_arg_1)]
+//! #[allow(dead_code, unused_variables, unused_mut)]
+//! #[kanitool::is_contract_generated(wrapper)]
+//! fn modify_wrapper_633496<'_wrapper_arg_1,
+//!     WrapperArgType1>(ptr: &mut u32, _wrapper_arg_1: &WrapperArgType1) {
+//!     *ptr += 1;
+//! }
+//! #[allow(dead_code)]
+//! #[kanitool::proof_for_contract = "modify"]
+//! fn main() {
+//!     kani::internal::init_contracts();
+//!     { let mut i = kani::any(); modify(&mut i); }
 //! }
 //! ```
 
--- a/tests/expected/function-contract/modifies/field_replace_pass.rs
+++ b/tests/expected/function-contract/modifies/field_replace_pass.rs
@@ -9,7 +9,7 @@ struct S<'a> {
 #[kani::requires(*s.target < 100)]
 #[kani::modifies(s.target)]
 #[kani::ensures(|result| *s.target == prior + 1)]
-fn modify(s: S, prior: u32) {
+fn modify(s: &mut S, prior: u32) {
     *s.target += 1;
 }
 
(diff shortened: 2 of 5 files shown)


feliperodri:
Why would this change affect #3027?

library/kani_macros/src/sysroot/contracts/mod.rs
celinval, line 1: btw, line 248 and 249 add the same ensures...

library/kani_macros/src/sysroot/contracts/mod.rs
```diff
@@ -265,50 +264,73 @@
...
 //! #[allow(dead_code, unused_variables, unused_mut)]
 //! #[kanitool::is_contract_generated(replace)]
 //! fn modify_replace_633496(ptr: &mut u32) {
 //!     kani::assert(*ptr < 100, "*ptr < 100");
-//!     let remember_kani_internal_1 = *ptr + 1;
-//!     let ptr_renamed = kani::internal::untracked_deref(&ptr);
-//!     let remember_kani_internal_0 = *ptr + 1;
-//!     let ptr_renamed = kani::internal::untracked_deref(&ptr);
+//!     let remember_kani_internal_92cc419d8aca576c = *ptr + 1;
+//!     let remember_kani_internal_92cc419d8aca576c = *ptr + 1;
```
celinval, line 308: isn't this overshadowing variable 307?

library/kani_macros/src/sysroot/contracts/mod.rs
```diff
@@ -265,50 +264,73 @@
...
 //!     let result_kani_internal: () = kani::any_modifies();
-//!     *unsafe { kani::internal::Pointer::assignable(ptr) } =
-//!         kani::any_modifies();
+//!     *unsafe {
```
celinval, line 310: why is this unsafe?

tests/expected/function-contract/modifies/field_replace_pass.rs
```diff
@@ -9,7 +9,7 @@ struct S<'a> {
 #[kani::requires(*s.target < 100)]
 #[kani::modifies(s.target)]
 #[kani::ensures(|result| *s.target == prior + 1)]
-fn modify(s: S, prior: u32) {
+fn modify(s: &mut S, prior: u32) {
```
celinval, line 12: Why are you changing this?

celinval (commented):
That's a nice cleanup. Thanks

library/kani_macros/src/sysroot/contracts/mod.rs
pi314mm: This is intentional as you get to see what happens with multiple ensures and the hash collision.

library/kani_macros/src/sysroot/contracts/mod.rs
pi314mm: This is intentional as they are the exact same history expression. Overshadowing is perfectly fine

library/kani_macros/src/sysroot/contracts/mod.rs
feliperodri: +1

feliperodri (approved):

tests/expected/function-contract/modifies/field_replace_pass.rs
pi314mm: The test originally did not make sense as the S struct gets consumed by the modify function and so you cannot refer to the contents of it within the ensures. The test originally passed unsafely as it ignored the fact that it referred to a potentially deallocated memory address. Changing it to a pass by reference is better than deleting the test.

library/kani_macros/src/sysroot/contracts/mod.rs
pi314mm: Tried to remove the unsafe and got the following error:
```call to unsafe function `kani::internal::Pointer::assignable` is unsafe and requires unsafe function or block```
My guess is that assignable is an unsafe dereference to force the havoc into the memory location.

File: library/kani_macros/src/sysroot/contracts/mod.rs (hunk 1 above)
Comment (celinval) [reply library/kani_macros/src/sysroot/contracts/mod.rs:308]: But would that happen if they were not the same expression? That is a confusing example

File: tests/expected/function-contract/modifies/field_replace_pass.rs (hunk 3 above)
Comment (celinval) [reply tests/expected/function-contract/modifies/field_replace_pass.rs:12]: Did the test break with your changes? I still believe we need tests to capture what the behavior would be here. Broken tests should trigger an error, not deleted since a user could make the same mistake

Comment (pi314mm) [reply tests/expected/function-contract/modifies/field_replace_pass.rs:12]: The test broke, but error cannot be represented within the current `expected` test framework because valid rust code is not produced after the macro expansion, so the kani compilation fails. To include this failing test case, the harness for `expected` needs to be changed.

Basically, the test case expands to something like
```rust
modify(s, prior); // s is consumed here
kani::assert(*s.target == prior + 1, "*s.target == prior + 1"); // s is used here after it is consumed
```
which is definitely not valid Rust code, and kani is unable to compile it, and the `expected` test fails even with the error message being included into the `.expected` file.

A better test case that is now fixed by this code change was mentioned in the issue #3239 
```rust
#[derive(Clone,Copy)]
struct Five;

impl Five {
    fn five(self : &Five) -> u32{
        5
    }
}

#[kani::ensures(|result : &Five| x.five() == result.five())]
fn id(x : Five) -> Five {
    x
}

#[kani::proof_for_contract(id)]
fn eat_harness() {
    let x = Five;
    id(x);
}
```
If you leave the copy/clone line, the test works (did not work before this change). If you delete the copy/clone line, you get the following error
```
error[E0382]: borrow of moved value: `x`
  --> ../test.rs:9:17
   |
9  | #[kani::ensures(|result : &Five| x.five() == result.five())]
   |                 ^^^^^^^^^^^^^^^^ - borrow occurs due to use in closure
   |                 |
   |                 value borrowed here after move
10 | fn id(x : Five) -> Five {
   |       -
   |       |
   |       value moved here
   |       move occurs because `x` has type `Five`, which does not implement the `Copy` trait
   |
note: consider changing this parameter type in function `id_wrapper_8110d0` to borrow instead if owning the value isn't necessary
  --> ../test.rs:10:11
   |
9  | #[kani::ensures(|result : &Five| x.five() == result.five())]
   | ------------------------------------------------------------ in this function
10 | fn id(x : Five) -> Five {
   |           ^^^^ this parameter takes ownership of the value
note: if `Five` implemented `Clone`, you could clone the value
  --> ../test.rs:1:1
   |
1  | struct Five;
   | ^^^^^^^^^^^ consider implementing `Clone` for this type
...
10 | fn id(x : Five) -> Five {
   |       - you could clone this value
   ```

celinval: We usually use the `ui` suite to ensure compilation errors are user friendly. In this case, I would recommend to move the broken test there. Thanks

PR #3276, opened 2024-06-19
Fix `--emit mir` handling of constants

celinval:
It shouldn't assume that the body of statics are available otherwise the compiler will ICE.


jaisnan (approved):
Ty! What issue is this in service of?

celinval:
The `--emit mir` is a debug flag, so I didn't bother creating an issue for this.

tautschnig:
> The `--emit mir` is a debug flag, so I didn't bother creating an issue for this.

Can the PR description/commit message perhaps still say what was going wrong/when things were doing wrong?

PR #3278, opened 2024-06-20
Remove support for the unstable argument `--function`

### LINKED ISSUE #2129: Remove `--function` option

celinval:
This is a legacy argument that we have very limited support to. We kept it around for bookrunner tests, which has been removed already.

Resolves #2129 


kani-compiler/src/codegen_cprover_gotoc/utils/names.rs
```diff
@@ -51,12 +51,9 @@ impl<'tcx> GotocCtx<'tcx> {
...
 
     /// Return the mangled name to be used in the symbol table.
-    ///
-    /// We special case main function in order to support `--function main`.
-    // TODO: Get rid of this: https://github.com/model-checking/kani/issues/2129
+    #[inline(always)]
     pub fn symbol_name_stable(&self, instance: Instance) -> String {
-        let pretty = instance.name();
-        if pretty == "main" { pretty } else { instance.mangled_name() }
+        instance.mangled_name()
```
adpaco-aws, line 56: Might as well just remove this function and call `instance.mangled_name()` wherever you need it.

kani-driver/src/args/mod.rs
```diff
@@ -569,7 +559,7 @@ impl ValidateArgs for VerificationArgs {
         if self.cbmc_args.contains(&OsString::from("--function")) {
             return Err(Error::raw(
                 ErrorKind::ArgumentConflict,
-                "Invalid flag: --function should be provided to Kani directly, not via --cbmc-args.",
+                "Invalid flag: --function is not supported in Kani.",
```
adpaco-aws, line 562: Why keep this case? Can't we just treat as any other argument that doesn't exist?

kani-driver/src/harness_runner.rs
```diff
@@ -173,26 +172,22 @@ impl KaniSession {
                     "Complete - {succeeding} successfully verified harnesses, {failing} failures, {total} total."
                 );
             } else {
-                match (self.args.harnesses.as_slice(), &self.args.function) {
-                    ([], None) =>
+                match self.args.harnesses.as_slice() {
+                    [] =>
```
adpaco-aws, line 176: Nit: I wonder if the code would be easier to read if we just do an `if ... then ... else` on the length of `self.args.harnesses`.

kani-driver/src/metadata.rs
```diff
@@ -115,12 +114,7 @@ impl KaniSession {
         &self,
         all_harnesses: &[&'a HarnessMetadata],
     ) -> Result<Vec<&'a HarnessMetadata>> {
-        let harnesses = if self.args.harnesses.is_empty() {
-            BTreeSet::from_iter(self.args.function.iter())
-        } else {
-            BTreeSet::from_iter(self.args.harnesses.iter())
-        };
-
+        let harnesses = BTreeSet::from_iter(self.args.harnesses.iter());
```
adpaco-aws, line 117: Shame on `clippy` 😆

kani-driver/src/project.rs
```diff
@@ -49,15 +35,6 @@ pub struct Project {
...
     artifacts: Vec<Artifact>,
-    /// A flag that indicated whether all artifacts have been merged or not.
-    ///
-    /// This allow us to provide a consistent behavior for `--function`.
-    /// For this option, we still merge all the artifacts together, so the
-    /// `merged_artifacts` flag will be set to `true`.
-    /// When this flag is `true`, there should only be up to one artifact of any given type.
-    /// When this flag is `false`, there may be multiple artifacts for any given type. However,
-    /// only up to one artifact for each
-    pub merged_artifacts: bool,
```
adpaco-aws, line 60: So `merged_artifacts` in a project would only be used through the `--function` flag?

adpaco-aws: I need to stop the review, I'll continue from here.

kani-driver/src/args/mod.rs
celinval: This is the CBMC --function args

kani-driver/src/harness_runner.rs
celinval: I don't think it makes much difference

adpaco-aws (approved):
Thanks, @celinval !

Do we want to mention this as a breaking change in the changelog? I doubt anyone depends on it so maybe listing the PR as usual is just enough.

celinval:
> Thanks, @celinval !
> 
> Do we want to mention this as a breaking change in the changelog? I doubt anyone depends on it so maybe listing the PR as usual is just enough.

Good question. This was an unstable feature that we provided very limited support, so I am not sure where to add it. That said, Maybe adding a callout might still be a good idea.

kani-compiler/src/codegen_cprover_gotoc/utils/names.rs
celinval: How much I love my IDE. That was a one click change. :smile:

PR #3279, opened 2024-06-21
Use older pip version to remain compatible with Python 3.7

tautschnig:
Fixes BuildBundle-Linux as an upstream change dropped support for Python 3.7. See also https://github.com/pypa/get-pip/issues/214 for some discussion.


jaisnan (approved):
Thanks, will probably be removing all this soon when we remove the dependency on `cbmc-viewer`

Issue #3235, opened 2024-06-06
Kani fails to detect use of a dangling pointer dereference in println


jsalzbergedu:
Given the following code:
```rust
static mut MY_POINTER: Option<*mut i32> = None;

fn reference_dies() {
  let mut x: i32 = 2;
  unsafe {
    MY_POINTER = Some(&mut x as *mut i32);
    println!("use to ensure that compiler doesn't optimize to Some(0), {}", &mut x);
  }
}

#[kani::proof]
fn main() {
  reference_dies();
  unsafe {
    match MY_POINTER {
        None => { unreachable!() },
        Some(x) => { println!("My pointer, {}", *x) },
    }
  }
  assert!(true);
}
```
kani v0.45 with the following flags:

```bash
kani --enable-unstable --memory-safety-checks --extra-pointer-checks aliasing_ub_fail.rs
```
gives a verification success even though cbmc gives a verification failure for

```C
#include <memory.h>
#include <stdio.h>
static int *MY_POINTER = NULL;

void reference_dies() {
    int x = 2;
    MY_POINTER = &x;
    return;
}

int main() {
    reference_dies();
    printf("%d", *MY_POINTER);
    return 0;
}
```

I'm hesitant to call this a "bug", because it falls under a reference being used outside of its lifetime, but it seems like something Kani should be able to get...

zhassan-aws:
Miri correctly reports the issue:
```
$ cargo miri run
Preparing a sysroot for Miri (target: x86_64-unknown-linux-gnu)... done
   Compiling ref_dies v0.1.0 (/home/ubuntu/examples/ref_dies)
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `/home/ubuntu/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo-miri runner target/miri/x86_64-unknown-linux-gnu/debug/ref_dies`
use to ensure that compiler doesn't optimize to Some(0), 2
error: Undefined Behavior: out-of-bounds pointer use: alloc838 has been freed, so this pointer is dangling
  --> src/main.rs:16:49
   |
16 |         Some(x) => { println!("My pointer, {}", *x) },
   |                                                 ^^ out-of-bounds pointer use: alloc838 has been freed, so this pointer is dangling
   |
   = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
[...]

PR #3280, opened 2024-06-21
Add tests checking if Kani detects dangling pointer dereference inside println macro

artemagvanian:
Extract tests from the discussion in #3235.

### DIFF
--- a/tests/expected/dangling-ptr-println/main.rs
+++ b/tests/expected/dangling-ptr-println/main.rs
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 // kani-flags: -Z ptr-to-ref-cast-checks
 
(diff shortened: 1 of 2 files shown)


tests/expected/dangling-ptr-println/main.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Z ptr-to-ref-cast-checks
+
+// These tests check that Kani correctly detects dangling pointer dereference inside println macro.
```
celinval, line 5:
```suggestion
//! These tests check that Kani correctly detects dangling pointer dereference inside println macro.
//! Related issue: <https://github.com/model-checking/kani/issues/3235>.
```

celinval (approved):
Awesome! Thanks

Issue #2832, opened 2023-10-20
Simplify installation


stefnotch:
Requested feature: Kani currently has Python as a dependency (on Windows and other not-directly-supported OSes at least). And Kani [also has the following dependencies](https://model-checking.github.io/kani/build-from-source.html). Since that adds dependencies that are not automatically installed by Cargo to a Rust project, it'd be worthwhile to improve this sometime in the future.

Use case: If I develop an open source library, then I want it to be as easy as possible for people to contribute to it. `cargo run` is beautiful.
Saying "install Python if you're on Windows" is obviously less beautiful, especially considering how un-fun Python installations on Windows can be.

Solution ideas:
- Rewriting the Python code. Rewriting all dependencies is probably infeasible.
- Using [RustPython](https://github.com/RustPython/RustPython) instead of the Python code. No idea if good equivalents exist for the other dependencies. There might be [cosmopolitan](https://github.com/jart/cosmopolitan) for the C dependencies, I guess.
- Compiling the dependency code to WASM and running that. That'd certainly be portable, and the documentation could include a section on "how to use the native dependencies to get rid of the performance overhead".
- Shipping Windows binaries. Only a partial solution, I suppose?

And of course I wouldn't request this feature anytime soon. I'd just love it if the team were to consider it at some unspecified point in the far future, when Kani is becoming even more popular!

celinval:
Hi @stefnotch, thank you for the issue. Another possibility would be to make the `--visualize` feature optional. That is the only reason we depend on Python today.

celinval:
We're now deprecating `--visualize` and we will soon remove the python dependency. :smiley:

stefnotch:
That's amazing, I look forward to verification tools being more accessible! It is insanely nice when the entry barrier is as low as possible, because it makes it so much easier to kani to certain projects, without causing a lot of extra effort for other contributors.

PR #3281, opened 2024-06-21
Deprecate `--visualize` in favor of concrete playback

celinval:
We believe the `--visualize` is much harder to use than concrete playback. In the rare cases where a trace might be relevant, users can still use CBMC trace.

For most users, this will simplify installation since Kani will no longer depend on Python3.

Note: As opposed to `--function` which was purely an internal feature, I believe we have mentioned `--visualize` to users before, so I think it's important to have a deprecation period.

Related #2832 


feliperodri (approved):
🎉

PR #3283, opened 2024-06-21
Enable an `#[safety_constraint(...)]` attribute helper for the `Arbitrary` and `Invariant` macros

adpaco-aws:
This PR enables an `#[safety_constraint(...)]` attribute helper for the `#[derive(Arbitrary)]` and `#[derive(Invariant)]` macro.

For the `Invariant` derive macro, this allows users to derive more sophisticated invariants for their data types by annotating individual named fields with the `#[safety_constraint(<cond>)]` attribute, where `<cond>` represents the predicate to be evaluated for the corresponding field. In addition, the implementation  always checks `#field.is_safe()` for each field.

For example, let's say we are working with the `Point` type from #3250
```rs
#[derive(kani::Invariant)]
struct Point<X, Y> {
    x: X,
    y: Y,
}
```

and we need to extend it to only allow positive values for both `x` and `y`.
With the `[safety_constraint(...)]` attribute, we can achieve this without explicitly writing the `impl Invariant for ...` as follows:

```rs
#[derive(kani::Invariant)]
struct PositivePoint {
    #[safety_constraint(*x >= 0)]
    x: i32,
    #[safety_constraint(*y >= 0)]
    y: i32,
}
```

For the `Arbitrary` derive macro, this allows users to derive more sophisticated `kani::any()` generators that respect the specified invariants. In other words, the `kani::any()` will assume any invariants specified through the `#[safety_constraint(...)]` attribute helper. Going back to the `PositivePoint` example, we'd expect this harness to be successful:

```rs
#[kani::proof]
fn check_invariant_helper_ok() {
    let pos_point: PositivePoint = kani::any();
    assert!(pos_point.x >= 0);
    assert!(pos_point.y >= 0);
}
```

The PR includes tests to ensure it's working as expected, in addition to UI tests checking for cases where the arguments provided to the macro are incorrect. Happy to add any other cases that you feel are missing.

Related #3095 


celinval:
Quick question... wouldn't it make sense to also parse this helper in the `#[derive(Arbitrary)]` macro?

adpaco-aws:
I'm not sure about that. How would users create an unsafe value then? This isn't clear to me and I think it's a use-case we should continue to support.

To me, it'd make more sense to have another API `kani::any_safe()` which combines `let x = kani::any()`and `kani::assume(x.is_safe())` (no need to replicate the parsing code). This way, the code would be explicit about the "safe values" assumption, which I don't think is clear otherwise.

celinval:
`kani::any()` is a safe function and it should only return safe values. Anything beyond that is incorrect.

tests/ui/derive-invariant/helper-wrong-expr/helper-wrong-expr.rs
```diff
@@ -0,0 +1,18 @@
...
+//! Check the compilation error for the invariant attribute helper when the
+//! argument cannot be evaluated in the struct's context.
+
+extern crate kani;
+use kani::Invariant;
+
+#[derive(kani::Arbitrary)]
+#[derive(kani::Invariant)]
+struct PositivePoint {
+    // Note: `x` is unknown, we should refer to `*x`
```
celinval, line 13:
```suggestion
    // Note: `x` is a reference in this context, we should refer to `*x`
```

tests/ui/derive-invariant/helper-empty/expected
```diff
@@ -0,0 +1,6 @@
+error: Cannot derive impl for `PositivePoint`
+   |
+11 | #[derive(kani::Invariant)]
+   |          ^^^^^^^^^^^^^^^
+   |
+note: invariant condition in field `x` could not be parsed - `Err("expected attribute arguments in parentheses: #[invariant(...)]")`
```
celinval, line 6: Can we remove the "Err" portion of the error? This should be:
```suggestion
note: invariant condition in field `x` could not be parsed: expected attribute arguments in parentheses: #[invariant(...)]
```

celinval (commented):
Nice! Can you please add a test where the invariant calls a function and one that tries to mutate the object? Thank you

zhassan-aws (commented):
Thanks. I find it a bit unexpected that `#[derive(Arbitrary)]` takes invariants from `#[invariant]` into account but not those from `impl Invariant`.

zhassan-aws:
For example, what happens when the field type already has an `Arbitrary` implementation (e.g. `char`):
```rust
#[derive(Arbitrary)]
struct Foo {
    #[invariant(*c != 'X')]
    c: char,
    #[invariant(*i > 2)]
    i: i32,
}
```
Would the safety invariant of `char` still be maintained?

tests/ui/derive-invariant/helper-empty/expected
adpaco-aws: Done.

adpaco-aws:
> Nice! Can you please add a test where the invariant calls a function and one that tries to mutate the object? Thank you

Added tests named `invariant_helper_funs` and `invariant_fail_mut` to check for these cases specifically.

adpaco-aws:
> For example, what happens when the field type already has an `Arbitrary` implementation (e.g. `char`):
> 
> ```rust
> #[derive(Arbitrary)]
> struct Foo {
>     #[invariant(*c != 'X')]
>     c: char,
>     #[invariant(*i > 2)]
>     i: i32,
> }
> ```
> 
> Would the safety invariant of `char` still be maintained?

The **safety** invariant for `char` is trivially satisfied (see https://github.com/model-checking/kani/pull/3190), so in this particular case it'd be maintained. It's important to note that the `Invariant` trait is not designed to work with **validity** invariants, which are likely the ones you're thinking about.

But you're right: the current implementation "rewrites" the field invariants. One option would be to call `is_safe()` for the field **in addition** to checking the specified invariants when generating the derived implementations. The problem is then is that, if the user wants to avoid the existing `impl Invariant` for the field type, they're forced to explicitly do so by writing the `impl Invariant` for the entire struct. In my opinion, it's more flexible to *not add `is_safe()` calls by default* because users can still define it as `#[invariant(*c != 'X' && c.is_safe())]` through the helper.

celinval:
> For example, what happens when the field type already has an `Arbitrary` implementation (e.g. `char`):
> 
> ```rust
> #[derive(Arbitrary)]
> struct Foo {
>     #[invariant(*c != 'X')]
>     c: char,
>     #[invariant(*i > 2)]
>     i: i32,
> }
> ```
> 
> Would the safety invariant of `char` still be maintained?

It should. The derive will use the `Arbitrary` implementation of the field type which should respect it's safety invariant.

adpaco-aws:
> For example, what happens when the field type already has an `Arbitrary` implementation (e.g. `char`):
> 
> ```rust
> #[derive(Arbitrary)]
> struct Foo {
>     #[invariant(*c != 'X')]
>     c: char,
>     #[invariant(*i > 2)]
>     i: i32,
> }
> ```
> 
> Would the safety invariant of `char` still be maintained?

The `Arbitrary` implementations should maintain the safety invariants, so in this example it'd be maintained.

But I'm not sure if you're alluding to the option of enforcing this by assuming those invariants in its `Arbitrary` implementation. For reference, this is what we're generating for that example:

```
impl kani::Arbitrary for Foo {
    fn any() -> Self {
        let obj = Foo {
            c: kani::any(),
            i: kani::any(),
        };
        let c = &obj.c;
        let i = &obj.i;
        kani::assume(*c != 'X' && *i > 2);
        obj
    }
}
```

We could add `c.is_safe() && i.is_safe()` to the condition assumed in `kani::assume(...)`, but that would create a dependency between the `Arbitrary` and `Invariant` traits, which is not desirable. Does this help?

celinval:
I'm saying that Arbitrary should always return safe values. The `invariant` attributes is to further constraint the field beyond its type constraints.

zhassan-aws:
I mistakenly thought that the derive `Arbitrary` implementation was only using the invariants for the individual fields. But perhaps the safety invariants are included through this line :
```
let obj = #body;
```
?

zhassan-aws:
I would still prefer keeping the `Arbitrary` implementation independent of the invariants to prevent surprises (e.g. vacuous proofs). I understand this would require manually adding `.is_safe()` in many places, but it seems safer to me from the soundness perspective.

adpaco-aws:
Update: On a separate discussion, we agreed on keeping the implementation as is while we ensure that the documentation communicates the danger of over-constraining values generated with `kani::any()` when using the `#[safety_constraint(...)]` (yes, we also agreed on a new name because `#[invariant(...)]` may give the impression that only works for the `Invariant` trait).

For the documentation, I've added it to the derive functions so that it's visible in the crates documentation. We could add another section to the "Reference" section of our documentation with this, but I don't know how we could keep them sync'd. We also ensure now that `is_safe()` gets called for each field regardless of whether they have the attribute or not.

zhassan-aws (commented):
Thanks. Do we check anywhere (or need to check) that the annotated field matches with the variable used in the predicate? 

For example, does Kani error out in the following case?
```rust
struct Foo {
    #[safety_constraint(*y == 5)]
    x: i32,
    y: i32,
}
```
If not, perhaps the safety constraints can be added anywhere inside the struct and not necessarily annotated on the individual fields?

tests/expected/derive-arbitrary/safety_constraint_helper/safety_constraint_helper.rs
```diff
@@ -0,0 +1,31 @@
...
+
+//! Check that the `#[safety_constraint(...)]` attribute helper adds the conditions provided to
+//! the attribute to the derived `Arbitrary` implementation.
+
+extern crate kani;
+use kani::Invariant;
+
+#[derive(kani::Arbitrary)]
+struct PositivePoint {
+    #[safety_constraint(*x >= 0)]
```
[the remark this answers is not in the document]
adpaco-aws: It is possible to use `self.x` if you're only deriving `Invariant`. That's because `self` doesn't mean anything in `any()`'s context, and we want to use the same condition in both. Otherwise, the error message is this:
```
error[E0424]: expected value, found module `self`
  --> src/main.rs:35:25
   |
32 | #[derive(kani::Arbitrary)]
   |          --------------- this function doesn't have a `self` parameter
...
35 |     #[safety_constraint(self.x == 0)]
   |                         ^^^^ `self` value is a keyword only available in methods with a `self` parameter

```

adpaco-aws:
> Thanks. Do we check anywhere (or need to check) that the annotated field matches with the variable used in the predicate?

We don't check the expressions for that. In my opinion, that'd be a nice feature to have but it's not needed at this point.

> For example, does Kani error out in the following case?
> 
> ```rust
> struct Foo {
>     #[safety_constraint(*y == 5)]
>     x: i32,
>     y: i32,
> }
> ```
> 
> If not, perhaps the safety constraints can be added anywhere inside the struct and not necessarily annotated on the individual fields?

No, it doesn't error out. We could change it so the safety constraints can be added somewhere else but I don't know what's the advantage. Also, note that the feature in #3270 is expected to allow constraints relating multiple fields of the struct. But if we're talking about constraints associated to individual fields, I think this feature is more appropriate because the annotations are easier to read.

tests/expected/derive-arbitrary/safety_constraint_helper/safety_constraint_helper.rs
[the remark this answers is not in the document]
adpaco-aws: BTW this has been updated in the PR description.

zhassan-aws (approved):

tests/ui/derive-invariant/helper-wrong-expr/expected
```diff
@@ -0,0 +1,9 @@
+error[E0308]: mismatched types
+   |
+14 |     #[safety_constraint(x >= 0)]
```
celinval, line 3: Can you please remove the line number of expected files. It makes any change to the file cumbersome, even adding comments.

tests/expected/derive-invariant/safety_constraint_helper_funs/safety_constraint_helper_funs.rs
```diff
@@ -0,0 +1,47 @@
...
+fn check_invariant_helper_ok() {
+    let pos_point: PositivePoint = kani::any();
+    assert!(pos_point.is_safe());
+}
+
+#[kani::proof]
+#[kani::should_panic]
+fn check_invariant_helper_fail() {
+    // In this case, `kani::any()` returns unconstrained values that don't
+    // respect the invariants.
```
celinval, line 34: nit: This comment took me by surprise, but looking at the code I see what you mean. The case is that building the struct from arbitrary values is not the same as creating an arbitrary struct.

celinval (commented):
It looks good to me. Can you please add a test where all the attributes are guarded by `kani` configuration. I.e.: `#[cfg_attr(kani, safety_attribute(/*content*/))]`.

Other tests that I would like to see is inadvertently specifying an expression with side effect, and mentioning other fields in the invariant expression. Thanks!

celinval (approved):

tests/ui/derive-invariant/helper-wrong-expr/expected
adpaco-aws: Removed.

tests/expected/derive-invariant/safety_constraint_helper_funs/safety_constraint_helper_funs.rs
adpaco-aws: I've rewritten this comment, hope it's clear now.

celinval (approved):
Thank you!!

Issue #3293, opened 2024-06-25
Contract implementation is unsafe and may trigger UB


celinval:
Looking at the `kani::internal::Pointer` implementation for `*mut T`:

https://github.com/model-checking/kani/blob/1491dd6eaf453efc61e915e0a3d247c2ce03abaf/library/kani/src/internal.rs#L59-L63

This can trigger UB if the location pointed by `*mut T` does not contain a valid value of type `T` since it is converting it to a `&T`.
Converting `*const T` into `&mut T` and `*mut T` to `&mut T` is also unsafe, and may break aliasing rules.

celinval:
Note that https://github.com/model-checking/kani/pull/3363 removes the `decouple_lifetime` function, but `untracked_deref` function is still unsafe and my cause UB for arguments that are dropped inside the function.

I was wondering if we can use something similar to Prusti's snapshots: https://viperproject.github.io/prusti-dev/dev-guide/encoding/types-snap.html to replace `untracked_deref`.

feliperodri:
The specific UB pattern (`decouple_lifetime` converting `*mut T` to `&'a T` via `&**self`) no longer exists. The Pointer trait was rewritten. `decouple_lifetime` was replaced with `assignable(self) -> *mut Self::Inner` which returns a raw pointer, avoiding the UB. The `impl Pointer` for `*mut T` at kani_core/src/lib.rs:476 now just returns self as `*mut T`.

Issue #3206, opened 2024-05-28
Error when annotating an associated function with a contract


zhassan-aws:
I tried this code:

```rust
extern crate kani;

#[derive(PartialEq, Eq)]
enum Foo {
    A,
    B,
}

impl Foo {
    #[kani::ensures(result == Foo::A)]
    pub fn new() -> Foo {
        Foo::A
    }
}

#[kani::proof_for_contract(Foo::new)]
fn check() {
    let _ = Foo::new();
}
```

using the following command line invocation:

```
kani enum_ensures.rs -Zfunction-contracts
```

with Kani version: https://github.com/model-checking/kani/commit/f10e61c232c9f76d920c5aaab822014ecc97685f

I expected to see this happen: Verification succeeds

Instead, this happened:
```
$ kani enum_ensures.rs -Zfunction-contracts
Kani Rust Verifier 0.51.0 (standalone)
error[E0425]: cannot find function `new_wrapper_faac06` in this scope
  --> enum_ensures.rs:10:5
   |
10 |     #[kani::ensures(result == Foo::A)]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
   |
   = note: this error originates in the attribute macro `kani::ensures` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider using the associated function on `Self`
   |
10 |     Self::#[kani::ensures(result == Foo::A)]
   |     ++++++

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0425`.
error: /home/ubuntu/git/kani/target/kani/bin/kani-compiler exited with status exit status: 1
```

zhassan-aws:
Found a workaround: the error goes away if I use `Self` as the return type instead of `Foo`:
```diff
$ diff enum_ensures.rs enum_ensures_self.rs 
11c11
<     pub fn new() -> Foo {
---
>     pub fn new() -> Self {
```
Verification fails though:
```
$ kani enum_ensures_self.rs -Zfunction-contracts
...
SUMMARY:
 ** 1 of 1336 failed
Failed Checks: result == Foo::A
 File: "enum_ensures_self.rs", line 10, in Foo::new_check_478ff5

VERIFICATION:- FAILED
Verification Time: 1.9594536s
```

zhassan-aws:
The issue in the previous comment seems to be independent of associated functions, so I filed a separate issue for it (#3228).

celinval:
[...]

PR #3295, opened 2024-06-26
Function Contracts: Modify Slices

### LINKED ISSUE #2908: Support for slices in `modifies` clauses
The `modifies` clause currently supports only pointers to a single object. To also support sequences (arrays) it should allow and specially handle slices.

pi314mm:
Using the `__CPROVER_object_upto` function to pass modifies clauses to asserts clauses in goto for rust slices.

Resolves #2908

Todo: Testing and Documentatiaon


pi314mm:
We are currently relying on `kani::any_modifies` markers that are compiled away into `kani::any`. The purpose of them is to allow for the function contracts to type check even if the modifies data does not implement `Arbitrary` as long as `stub_contract` is not used.

Instead of the `kani::any_modifies`,  we should be using `kani::any_fill` that indicates that the parameter should be dereferenced and assigned to, which will allow us to handle both slim pointers and fat pointers, as the method to assign to them is different.

library/kani/src/lib.rs
celinval, line 1: Please move all these functions into `internal` module and add the unstable tag to them using the function contract feature gate.

library/kani/src/lib.rs
```diff
@@ -183,6 +183,39 @@ pub fn any_modifies<T>() -> T {
...
+#[rustc_diagnostic_item = "KaniHavocSlim"]
+#[inline(always)]
+pub fn havoc_slim<T: Arbitrary>(pointer: &mut T) {
+    *pointer = T::any()
+}
+
+#[rustc_diagnostic_item = "KaniHavocStr"]
+#[inline(always)]
+pub fn havoc_str(s: &mut str) {
+    unsafe { s.as_bytes_mut() }.fill_with(u8::any)
```
celinval, line 215: This is UB.

library/kani/src/lib.rs
```diff
@@ -183,6 +183,39 @@ pub fn any_modifies<T>() -> T {
...
+#[rustc_diagnostic_item = "KaniHavocSlice"]
+#[inline(always)]
+pub fn havoc_slice<T: Arbitrary>(slice: &mut [T]) {
+    slice.fill_with(T::any)
+}
+
+#[rustc_diagnostic_item = "KaniHavocSlim"]
+#[inline(always)]
+pub fn havoc_slim<T: Arbitrary>(pointer: &mut T) {
+    *pointer = T::any()
```
celinval, line 209: We should probably take `*mut T` so this can be used for havocking `&mut` and `*mut`. We should also use `ptr::write` otherwise it will be UB for `*mut T` that hasn't been initialized yet.

library/kani/src/lib.rs
```diff
@@ -183,6 +183,39 @@ pub fn any_modifies<T>() -> T {
...
 
+/// Recieves a reference to a pointer-like object and assigns kani::any_modifies to that object.
+/// Only for use within function contracts and will not be replaced if the recursive or function stub
+/// replace contracts are not used.
+#[rustc_diagnostic_item = "KaniHavoc"]
+#[inline(never)]
+#[doc(hidden)]
+pub fn havoc<T: ?Sized>(_pointer: &T) {
+    // This function should not be reacheable.
+    // Users must include `#[kani::recursion]` in any function contracts for recursive functions;
```
celinval, line 194: I don't understand the relationship.

library/kani/src/lib.rs
```diff
@@ -183,6 +183,39 @@ pub fn any_modifies<T>() -> T {
     unreachable!()
 }
 
+/// Recieves a reference to a pointer-like object and assigns kani::any_modifies to that object.
+/// Only for use within function contracts and will not be replaced if the recursive or function stub
+/// replace contracts are not used.
+#[rustc_diagnostic_item = "KaniHavoc"]
+#[inline(never)]
+#[doc(hidden)]
+pub fn havoc<T: ?Sized>(_pointer: &T) {
```
celinval, line 192: I would suggest that we rename these `havoc` functions to follow the `any` convention. Maybe `write_any` since we are writing an arbitrary value into a memory location.

celinval (changes requested):
Thanks for doing this. I haven't reviewed the compiler changes yet.

library/kani/src/lib.rs
pi314mm: We are doing an unsafe cast with kani::internal::Pointer to change &T, &mut T, *const T, and *mut T all into a &mut T. Would this fix issue #3293 if we changed everything into a *mut T instead of a &mut T?

library/kani/src/lib.rs
pi314mm: Yes. Supporting strings is not the objective of this PR, but we will ultimately need to use UB to hack strings as there is no other way. Just need to add string validation to make it safe.

library/kani/src/lib.rs
celinval: I don't think it would solve all the issues, but it would definitely help. I'm actually removing the decouple_lifetime function in my next PR for issue #3206 .

library/kani/src/lib.rs
pi314mm: Strings were replaced with a todo. We are not introducing new UB in this PR

library/kani/src/internal.rs
```diff
@@ -97,3 +96,42 @@ pub fn init_contracts() {}
...
+}
+
+#[crate::unstable(feature = "function-contracts", issue = "none", reason = "function-contracts")]
+#[rustc_diagnostic_item = "KaniWriteAnyStr"]
+#[inline(always)]
+pub unsafe fn write_any_str(_s: *mut str) {
+    //TODO: strings introduce new UB
+    //(*s).as_bytes_mut().fill_with(u8::any)
+    //TODO: String validation
+    unreachable!("strings are not yet implemented")
```
celinval, line 136:
```suggestion
    unimplemented!("Kani does not support creating arbitrary `str`")
```

kani-compiler/src/kani_middle/transform/contracts.rs
```diff
@@ -117,6 +142,42 @@ impl AnyModifiesPass {
                 *func = Operand::Constant(new_func);
                 changed = true;
             }
+
+            if let TyKind::RigidTy(RigidTy::FnDef(def, instance_args)) =
```
celinval, line 146: Can you please add a comment here saying what you are doing?

celinval (approved):
Thanks!

library/kani/src/internal.rs
```diff
@@ -97,3 +96,42 @@ pub fn init_contracts() {}
...
+
+#[crate::unstable(feature = "function-contracts", issue = "none", reason = "function-contracts")]
+#[rustc_diagnostic_item = "KaniWriteAnyStr"]
+#[inline(always)]
+pub unsafe fn write_any_str(_s: *mut str) {
+    //TODO: strings introduce new UB
+    //(*s).as_bytes_mut().fill_with(u8::any)
+    //TODO: String validation
+    unimplemented!("Kani does not support creating arbitrary `str`")
+}
```
feliperodri, line 137: We need a regression test that shows that we report this error correctly in such cases.

library/kani/src/internal.rs
```diff
@@ -97,3 +96,42 @@ pub fn init_contracts() {}
...
+pub unsafe fn write_any_slice<T: Arbitrary>(slice: *mut [T]) {
+    (*slice).fill_with(T::any)
+}
+
+#[crate::unstable(feature = "function-contracts", issue = "none", reason = "function-contracts")]
+#[rustc_diagnostic_item = "KaniWriteAnySlim"]
+#[inline(always)]
+pub unsafe fn write_any_slim<T: Arbitrary>(pointer: *mut T) {
+    ptr::write(pointer, T::any())
+}
```
feliperodri, line 127: What about the documentation for these other functions?

kani-compiler/src/kani_middle/transform/contracts.rs
```diff
@@ -25,6 +25,10 @@ use tracing::{debug, trace};
 pub struct AnyModifiesPass {
     kani_any: Option<FnDef>,
     kani_any_modifies: Option<FnDef>,
+    kani_write_any: Option<FnDef>,
+    kani_write_any_slim: Option<FnDef>,
+    kani_write_any_slice: Option<FnDef>,
+    kani_write_any_str: Option<FnDef>,
```
feliperodri, line 31: We need more documentation stating why we need these different markers.

tests/expected/function-contract/modifies_fat_pointer/u8slice.rs
```diff
@@ -0,0 +1,17 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zfunction-contracts
+
+// Test that modifies a slice containing bytesize data
+
+#[kani::modifies(x)]
+#[kani::ensures(|_| x.iter().map(|v| *v == 0).fold(true,|a,b|a&b))]
+fn zero(x: &mut [u8]) {
+    x.fill(0)
+}
```
feliperodri, line 11: Can you include a test case where the parameter is an array, but we want to specify in the assigns clause only a slice of that array?

feliperodri (changes requested):

library/kani/src/internal.rs
pi314mm: Reported in #3349

feliperodri (approved):

PR #3305, opened 2024-06-28
Fix contract handling of promoted constants and constant static

### LINKED ISSUE #3228: Contract fails due to promoted constants being havocked
I tried this code:

```rust
extern crate kani;

#[derive(PartialEq, Eq)]
pub enum Foo {
    A,
    B,
}

#[kani::ensures(result == Foo::A)]
pub fn a() -> Foo {
    Foo::A
}

#[kani::proof_for_contract(a)]
fn check() {
    let _ = a();
}
```

```
kani enum_ensures_bug.rs -Zfunction-contracts
```

with Kani version: 7bf23a25dd3

Instead, this happened:
```
SUMMARY:
 ** 1 of 1336 failed
Failed Checks: result == Foo::A
 File: "enum_ensures_bug.rs", line 9, in a_check_7b31b

VERIFICATION:- FAILED
Verification Time: 1.9666983s
```

celinval:
When verifying contracts, CBMC initializes all static variables to non-deterministic values, except for those with constant types or with types / values annotated with `ID_C_no_nondet_initialization`.

Kani compiler never set these flags, which caused spurious failures when verification depended on promoted constants or constant static variables. This fix changes that.

First, I did a bit of refactoring since we may need to set this `Symbol` property at a later time for static variables.
I also got rid of the initialization function, since the allocation initialization can be done directly from an expression.

Then, I added the new property to the `Symbol` type. In CBMC, this is a property of the type or expression.  However, I decided to add it to `Symbol` to avoid having to add this attribute to all variants of `Type` and `Expr`.

Resolves #3228 


celinval:
@remi-delmas-3000 do you mind taking a look?

cprover_bindings/src/goto_program/symbol.rs
```diff
@@ -46,6 +48,14 @@ pub struct Symbol {
     pub is_thread_local: bool,
     pub is_volatile: bool,
     pub is_weak: bool,
+
+    /// This flag marks a variable as constant (IrepId: `ID_C_constant`).
+    ///
+    /// In CBMC, this is a property of the type or expression. However, we keep it here to avoid
+    /// having to propagate the attribute to all variants of `Type` and `Expr`.
+    ///
+    /// During contract verification, CBMC will not havoc static variables marked as constant.
```
remi-delmas-3000, line 57: Should we distinguish promoted constants from user-declared symbols declared as constants ? (the first type are really literal constants, the second one may be modified through unsafe) ?

cprover_bindings/src/goto_program/symbol.rs
```diff
@@ -157,6 +167,7 @@ impl Symbol {
             is_lvalue: false,
             is_parameter: false,
             is_static_lifetime: false,
+            is_static_const: false,
```
remi-delmas-3000, line 170: are all _const_ necessarily _static_ ? we already have a flag for _static_lifetime_

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -513,50 +512,40 @@ impl<'tcx> GotocCtx<'tcx> {
                 .collect()
         });
 
+        // Create the allocation from an array byte array.
```
remi-delmas-3000, line 515:
```suggestion
        // Create the allocation from a byte array.
```

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -484,13 +484,12 @@ impl<'tcx> GotocCtx<'tcx> {
         mem_place.address_of()
     }
 
-    /// Insert an allocation into the goto symbol table, and generate a goto function that will
-    /// initialize it.
+    /// Insert an allocation into the goto symbol table, and generate an init value.
```
remi-delmas-3000, line 487: What does _allocation_ mean in this context ?

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -513,50 +512,40 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                alloc_data
+                    .iter()
+                    .map(|d| match d {
+                        AllocData::Bytes(bytes) => Expr::array_expr(
+                            Type::unsigned_int(8).array_of(bytes.len()),
+                            bytes
+                                .iter()
+                                // We should consider adding a poison / undet where we have none
+                                // This mimics the behaviour before StableMIR though.
+                                .map(|b| Expr::int_constant(b.unwrap_or(0), Type::unsigned_int(8)))
```
remi-delmas-3000, line 528: is it really the case that every individual byte is potentially undefined ?

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
```diff
@@ -513,50 +512,40 @@ impl<'tcx> GotocCtx<'tcx> {
...
+                        AllocData::Expr(e) => e.clone(),
+                    })
+                    .collect(),
+                &gcx.symbol_table,
+            );
+            if val.typ() == &var.typ { val } else { val.transmute_to(var.typ, &gcx.symbol_table) }
+        };
+
         // The global static variable may not be in the symbol table if we are dealing
         // with a literal that can be statically allocated.
```
remi-delmas-3000, line 540: is this really referring to a promoted constant ? if so use the dedicated term

kani-compiler/src/codegen_cprover_gotoc/codegen/static_var.rs
```diff
@@ -37,9 +36,8 @@ impl<'tcx> GotocCtx<'tcx> {
         // havoc static variables. Kani uses the location and pretty name to identify
         // the correct variables. If the wrong name is used, CBMC may fail silently.
         // More details at https://github.com/diffblue/cbmc/issues/8225.
```
remi-delmas-3000, line 38: Is it the right place for that comment ?  shouldn't it be moved to where `--nondet-static-exclude` is actually used ?

```suggestion
        // CBMC contract instrumentation havocs all statics by default (for soundness). 
        // However, some static variables added by Kani's own contract instrumentation must not be havoced
        // in order for the instrumentation to work as intented (if these variables get havoced,
        // the analysis may fail spuriously, cf.  https://github.com/diffblue/cbmc/issues/8225).
        // Kani uses location and pretty name to identify these variables and uses 
        // `--nondet-static-exclude` to exclude these variables from havocing.
```

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
```diff
@@ -200,34 +203,58 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        if matches!(sym.value, SymbolValues::None) {
+            // Clone sym so we can use `&mut self`.
+            let sym = sym.clone();
+            let init_expr = SymbolValues::Expr(init(self, sym));
+            // Need to lookup again since symbol table might've changed.
+            let sym = self.symbol_table.lookup_mut(name).unwrap();
+            sym.value = init_expr;
+            sym
+        } else {
+            self.symbol_table.lookup_mut(name).unwrap()
```
remi-delmas-3000, line 235: What if the symbol was present but its `init_expr` field was not defined, or defined to something different from the current `init` ?

cprover_bindings/src/goto_program/symbol.rs
celinval: No, I don't think so. We should however handle `UnsafeCell`. We could for now check if the type has one. If so, we keep it as mutable.

Note that mutating a constant is immediate UB.

>Mutating immutable bytes. All bytes inside a [const](https://doc.rust-lang.org/reference/items/constant-items.html) item are immutable. The bytes owned by an immutable binding are immutable, unless those bytes are part of an [UnsafeCell<U>](https://doc.rust-lang.org/std/cell/struct.UnsafeCell.html).

https://doc.rust-lang.org/reference/behavior-considered-undefined.html

cprover_bindings/src/goto_program/symbol.rs
celinval: We are only marking static constants for now. If we were to start marking variables constants, then we could rename this.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
celinval: Because we don't care about the value and 0 is simpler. Reading uninitialized values is always UB, hence we are adding code to check for that.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
celinval: sure... I can update this comment

kani-compiler/src/codegen_cprover_gotoc/codegen/static_var.rs
celinval: I didn't change this code, but we can update the doc. No problem.

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
celinval: The former will enter the `if` block since `sym.value` will be `None`. We could add a sanity check for the value initialization to catch a potential rustc bug. That would require us to implement a proper expression comparison, which we don't have today and I believe it is out of scope of this PR.

celinval:
> seems like filling up undefined bytes with `false` in Allocation initialisers may be underapproximating.

Reading undefined bytes is UB. Any analysis that read uninitialized bytes is unsound, hence the value doesn't matter.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
remi-delmas-3000: Do we have the guarantee that the check for uninitialised values isn't inhibited by this replacement of nondet by 0 ?

cprover_bindings/src/goto_program/symbol.rs
celinval: Note that the following program crashes:
```rust
static STATIC_CONST: u32 = 0;

fn check_static() {
    assert_eq!(STATIC_CONST, 0);
}

fn assign(ptr: *mut u32) {
    unsafe { *ptr = 10 };
}

fn main() {
    let ptr = &STATIC_CONST as *const _ as *mut _;
    assign(ptr); // -- Doing the assignment here triggers a compilation error.
    check_static();
}
```

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
celinval: The uninitialized value check should consider the type of the static. Is that the case today, @artemagvanian?

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
celinval: It means a memory allocation.

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
artemagvanian: I think it does if I understand you correctly: here is the code snippet for blessing statics:
```rust
fn visit_operand(&mut self, operand: &Operand, location: Location) {
    if let Operand::Constant(constant) = operand {
        if let ConstantKind::Allocated(allocation) = constant.const_.kind() {
            for (_, prov) in &allocation.provenance.ptrs {
                if let GlobalAlloc::Static(_) = GlobalAlloc::from(prov.0) {
                    if constant.ty().kind().is_raw_ptr() {
                        // If a static is a raw pointer, need to mark it as initialized.
                        self.push_target(MemoryInitOp::Set {
                            operand: Operand::Constant(constant.clone()),
                            value: true,
                            position: InsertPosition::Before,
                        });
                    }
                };
            }
        }
    }
    self.super_operand(operand, location);
}
```
It does consider the type of the static and only blesses raw pointers

celinval:
Sorry, I accidentally rebased the branch. :(

kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
remi-delmas-3000: What do you mean by "blessing" ? Is it synonymous for "checking" ?

artemagvanian: By "blessing" I mean marking it as initialized in that case -- the opposite for checking. Sorry for the confusion!

celinval: Let me add a check here.

celinval: @remi-delmas-3000 I added a test to ensure that uninitialized checks correctly identifies the poison value in the static padding.

remi-delmas-3000 (approved):

kani-compiler/src/kani_middle/mod.rs
```diff
@@ -19,8 +19,11 @@ use rustc_span::Span;
 use rustc_target::abi::call::FnAbi;
 use rustc_target::abi::{HasDataLayout, TargetDataLayout};
 use stable_mir::mir::mono::MonoItem;
-use stable_mir::ty::{FnDef, RigidTy, Span as SpanStable, TyKind};
+use stable_mir::ty::{FnDef, RigidTy, Span as SpanStable, Ty, TyKind};
+use stable_mir::visitor::{Visitable, Visitor as TyVisitor};
 use stable_mir::CrateDef;
+use std::ops::ControlFlow;
+
```
celinval, line 26:
```suggestion
```

PR #3307, opened 2024-06-28
Function Contracts: Closure Type Inference

### LINKED ISSUE #3304: UX Improvement: Ensures clause requires type annotation
Requested feature: Improve UX for ensure clauses. 
Use case: Since #3207, ensures are now represented as closures that receive a reference of the return value. However, when using this annotation, we often get an error that requires the type to be specified.

Test case:
```rust
#[kani::ensures(|result| *result == Foo::A)]
pub fn foo_a() -> Foo {
    Foo::A
}
```

I get the following compilation error:
```rust
  --> promoted_constants_enum.rs:16:18
   |
16 | #[kani::ensures(|result| *result == Foo::A)]
   |                  ^^^^^^  ------- type must be known at this point
   |
help: consider giving this closure parameter an explicit type
   |
16 | #[kani::ensures(|result: /* Type */| *result == Foo::A)]
```

pi314mm:
The rust type inference for closures doesn't work in the particular use case we are using it for ensures clauses. By creating a helper function, we change the path the rust type inference takes and lets the type of the closure be identified properly. This means type annotations are no longer required within ensures clauses.

Resolves #3304


library/kani/src/lib.rs
```diff
@@ -326,3 +326,7 @@ pub use kani_macros::*;
 pub(crate) use kani_macros::unstable_feature as unstable;
 
 pub mod contracts;
+
+pub fn apply_closure<T, U: FnOnce(&T) -> bool>(f: U, x: &T) -> bool {
```
celinval, line 330: We should use `Fn` since the closure should not mutate the state.

celinval: Also, please move this function to the `internal` module, and add documentation that this should only be used by contracts.

Maybe create an expected test where this call can fail, for example, mutating expression to see what the error will look like. Thanks!

celinval: We usually mark internal functions with `#[doc(hidden)]`. However, I don't know if that is the case if they show up in the error message

celinval (approved):
Awesome! Thanks

Issue #3300, opened 2024-06-27
Tracking Issue: Automatically Verify Memory Initialization


artemagvanian:
Goal: Implement automated and efficient checks for memory initialization in Kani.

## Motivation

Safe Rust guarantees that every variable is initialized, and every reference points to a properly allocated and initialized memory position. The same guarantees do not apply to raw pointers, transmute operations and unions, which are available in unsafe Rust. The safe usage of these constructs and operations must be upheld by the Rust developer.

Failures to uphold these safety properties can have unwanted consequences: it can affect the application behavior and even result in the application exposing critical data. 

Kani should be able to detect uninitialized memory accesses automatically and efficiently.

## Examples

Currently, Kani does not catch the following cases of UB due to accessing uninitialized memory:

```rust
#[kani::proof]
fn check_uninit_padding() {
    let s = S(0, 0);
    let ptr: *const u8 = addr_of!(s) as *const u8;
    let padding = unsafe { *(ptr.add(5)) }; // ~ERROR: padding bytes are uninitialized, so reading them is UB.
}
```

```rust
#[kani::proof]
fn check_vec_read_uninit() {
    let v: Vec<u8> = Vec::with_capacity(10);
    let uninit = unsafe { *v.as_ptr().add(5) }; // ~ERROR: reading from unitialized memory is UB.
}
```

## Tasks

- [ ] Implement brute-force memory initialization checks for raw pointers (#3264)
- [ ] Make sure layout representation does not negatively affect performance
- [ ] Implement pruning unnecessary checks via static analysis
- [ ] Add support for `enum`s with variants that have different padding
- [ ] Add support for `transmute`s
- [ ] Add support for `union`s
- [ ] Add support for `extern static`s

zhassan-aws:
> // ~ERROR: padding bytes are uninitialized, so reading them is UB.

The [Rust reference ](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#behavior-considered-undefined) mentions that reading padding bytes is not UB: 

> In other words, the only cases in which reading uninitialized memory is permitted are inside unions and in "padding" (the gaps between the fields/elements of a type).

artemagvanian:
@zhassan-aws thanks for pointing that out -- I interpret the Rust reference a bit differently in this case. 

[...]

PR #3313, opened 2024-07-01
Improve performance and language support of memory initialization checks

artemagvanian:
This is a follow-up PR for #3264, a part of the larger work towards #3300.

This PR introduces:

- Use of demonic non-determinism (prophecy variables) to improve memory initialization checks performance;
    - Instead of keeping track of memory initialization of each possibly uninitialized byte pointed to by some pointer, one is chosen non-deterministically;
- Tests for proper memory initialization checks injection for compiler intrinsics;
- Separate functions for checking memory initialization of slice chunks.


artemagvanian:
Seems like `Release Bundle / BuildBundle-Linux (pull_request)` is failing because of https://github.com/actions/checkout/issues/1809

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -426,3 +464,20 @@ pub fn resolve_mem_init_fn(fn_def: FnDef, layout_size: usize, associated_type: T
     )
     .unwrap()
 }
+
+/// Checks if the instance is a harness -- an entry point of Kani analysis.
+fn is_harness(instance: Instance, tcx: TyCtxt) -> bool {
```
zhassan-aws, line 469: This function could be useful elsewhere, so it might make sense to move it up (not necessarily in this PR).

library/kani_core/src/mem.rs
```diff
@@ -123,9 +123,17 @@ macro_rules! kani_mem {
             <T as Pointee>::Metadata: PtrProperties<T>,
         {
             let (thin_ptr, metadata) = ptr.to_raw_parts();
+            // Need to assert `is_initialized` because non-determinism is used under the hood, so it
+            // does not make sense to use it inside assumption context.
             metadata.is_ptr_aligned(thin_ptr, Internal)
                 && is_inbounds(&metadata, thin_ptr)
-                && is_initialized(ptr, 1)
+                && {
+                    assert!(
```
zhassan-aws, line 131: Neat!

Wouldn't this still result in using `is_initialized` in an assumption though since we encode `assert!` as assert followed by assume?

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -74,6 +74,44 @@ impl TransformPass for UninitPass {
...
+            // First statement or terminator in the harness.
+            let mut source = if !new_body.blocks()[0].statements.is_empty() {
+                SourceInstruction::Statement { idx: 0, bb: 0 }
+            } else {
+                SourceInstruction::Terminator { bb: 0 }
+            };
+
+            // Dummy return place.
+            let ret_place = Place {
+                local: new_body.new_local(
```
zhassan-aws, line 88: Could be wrapped in a helper function.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -74,6 +74,44 @@ impl TransformPass for UninitPass {
...
+                ),
+                &GenericArgs(vec![]),
+            )
+            .unwrap();
+            new_body.add_call(
+                &memory_initialization_init,
+                &mut source,
+                InsertPosition::Before,
+                vec![],
+                ret_place.clone(),
```
zhassan-aws, line 111: Is the clone needed?

zhassan-aws (commented):
Nice!

What happens if there are two accesses to distinct uninitialized memory in a row? Will the first violation block the reporting of the second?

library/kani/src/mem_init.rs
```diff
@@ -25,98 +28,174 @@ use crate::shadow::ShadowMem;
...
     }
-    true
+
+    /// Set currently tracked memory initialization state if `ptr` points to the currently tracked
+    /// object and the tracked offset lies within `LAYOUT_SIZE * num_elts` bytes of `ptr`. Do
+    /// nothing otherwise.
+    ///
+    /// Such definition is necessary since both tracked object and tracked offset are chosen
+    /// non-deterministically.
+    pub fn set<const LAYOUT_SIZE: usize>(
```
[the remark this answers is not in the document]
artemagvanian: Will investigate that

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
artemagvanian: I took a note, will do

library/kani_core/src/mem.rs
artemagvanian: As far as I understand, using it in an assumption context it would be a no-op, wouldn't it?

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
artemagvanian: Nope, good catch -- removed it

tests/std-checks/core/mem.expected
```diff
@@ -1,7 +1,30 @@
...
+
+VERIFICATION:- FAILED
+
 Summary:
+Verification failed for - mem::verify::check_swap_large_adt_no_drop
+Verification failed for - mem::verify::check_swap_adt_no_drop
 Verification failed for - mem::verify::check_swap_unit
-Complete - 6 successfully verified harnesses, 1 failures, 7 total.
+Verification failed for - mem::verify::check_swap_primitive
+Complete - 3 successfully verified harnesses, 4 failures, 7 total.
```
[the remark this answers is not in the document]
artemagvanian: Should we disable it for all `std-checks` for now?

artemagvanian:
> What happens if there are two accesses to distinct uninitialized memory in a row? Will the first violation block the reporting of the second?

Good question: I tested this simple case and all of the violations were reported.

```rust
use std::ptr::addr_of;

#[repr(C)]
struct S(u32, u8);

/// Checks that Kani catches an attempt to access padding of a struct using raw pointers.
#[kani::proof]
fn check_uninit_padding() {
    let s = S(0, 0);
    let ptr: *const u8 = addr_of!(s) as *const u8;
    let padding = unsafe { *(ptr.add(5)) }; // ~ERROR: padding bytes are uninitialized, so reading them is UB.
    let padding_next = unsafe { *(ptr.add(6)) }; // ~ERROR: padding bytes are uninitialized, so reading them is UB.
    inner_fn(ptr);
}

fn inner_fn(ptr: *const u8) {
    let padding_inner = unsafe { *(ptr.add(7)) }; // ~ERROR: padding bytes are uninitialized, so reading them is UB.
}
```

library/kani/src/mem_init.rs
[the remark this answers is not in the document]
artemagvanian: I pushed a commit with these changes -- the performance impact seems to be minimal for the unit tests we have; the code looks cleaner in some places, though

zhassan-aws:
> Good question: I tested this simple case and all of the violations were reported.

Cool, thanks for checking.

library/kani/src/mem.rs
```diff
@@ -294,10 +300,16 @@ unsafe fn has_valid_value<T: ?Sized>(_ptr: *const T) -> bool {
 /// Check whether `len * size_of::<T>()` bytes are initialized starting from `ptr`.
 #[rustc_diagnostic_item = "KaniIsInitialized"]
 #[inline(never)]
-pub fn is_initialized<T: ?Sized>(_ptr: *const T, _len: usize) -> bool {
+pub fn is_initialized<T: ?Sized>(_ptr: *const T) -> bool {
     kani_intrinsic()
 }
 
+/// A helper to to assert `is_initialized` to use it as a part of other predicates.
```
zhassan-aws, line 307:
```suggestion
/// A helper to assert `is_initialized` to use it as a part of other predicates.
```

library/kani_core/src/mem.rs
```diff
@@ -302,10 +306,19 @@ macro_rules! kani_mem {
         /// Check whether `len * size_of::<T>()` bytes are initialized starting from `ptr`.
         #[rustc_diagnostic_item = "KaniIsInitialized"]
         #[inline(never)]
-        pub fn is_initialized<T: ?Sized>(_ptr: *const T, _len: usize) -> bool {
+        pub fn is_initialized<T: ?Sized>(_ptr: *const T) -> bool {
             kani_intrinsic()
         }
 
+        /// A helper to to assert `is_initialized` to use it as a part of other predicates.
```
zhassan-aws, line 313:
```suggestion
        /// A helper to assert `is_initialized` to use it as a part of other predicates.
```

library/kani_core/src/mem.rs
artemagvanian: Do we have an API that generates assert only? This would be easy to implement if needed.

zhassan-aws: I don't think we have one currently. We can add one (e.g. `check`) that is codegened as an assert only (`__CPROVER_assert`).

artemagvanian: I fixed it in the latest commit, also made sure that the instrumentation adds assertions only

tests/std-checks/core/mem.expected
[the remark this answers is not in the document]
artemagvanian: Disabled memory initialization checking in `std-checks` for now

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -143,6 +143,46 @@ impl GotocHook for Assert {
...
+        let msg = gcx.extract_const_message(&msg).unwrap();
+        let target = target.unwrap();
+        let caller_loc = gcx.codegen_caller_span_stable(span);
+
+        let (msg, reach_stmt) = gcx.codegen_reachability_check(msg, span);
+
+        // Since `cond` might have side effects, assign it to a temporary
+        // variable so that it's evaluated once, then assert it
+        // TODO: I don't think `cond` can have side effects, this is MIR, it's going to be temps
+        let (tmp, decl) = gcx.decl_temp_variable(cond.typ().clone(), Some(cond), caller_loc);
```
zhassan-aws, line 173: There is no need for a temporary variable in this case since we're only using the condition once. The temporary is needed in the case where it's going to be referenced multiple times (e.g. in an assert followed by an assume).

zhassan-aws (approved):
Thanks!

PR #3266, opened 2024-06-14
Use GitHub's M1 macOS runner

tautschnig:
An M1 (aarch64, Apple Silicon) runner is available on the free plan since January 2024 (see
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #3320, opened 2024-07-03
Auto-upload macOS ARM binaries to draft release

tautschnig:
In #3266 a job to build macOS ARM binaries was added, but the artifacts weren't yet propagated to the release. This PR adds this missing step.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -322,7 +322,7 @@ jobs:
     if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/kani-') }}
     name: Release
     runs-on: ubuntu-20.04
-    needs: [build_bundle_macos, build_bundle_linux, test_bundle, test_alt_platform]
+    needs: [build_bundle_macos, build_bundle_macos_aarch64, build_bundle_linux, test_bundle, test_alt_platform]
     permissions:
       contents: write
     outputs:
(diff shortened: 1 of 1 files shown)


jaisnan (commented):
Great catch, thank you. Is there any way of testing that this works?

tautschnig:
Looks we are also caught by https://github.com/actions/checkout/issues/1809

tautschnig:
> Great catch, thank you. Is there any way of testing that this works?

Not an easy one that I know of, I'm afraid.

adpaco-aws:
It's not easy to test but also not the end the of the world if it fails for any reason. In that case, one only needs to go the Release Bundle summary (e.g., [this one](https://github.com/model-checking/kani/actions/runs/9780933430) for version 0.53.0), then download and attach the missing bundles to the release.

celinval:
Wait, is there a reason why we are not running any test in the image like we do with the other images?

tautschnig:
> Wait, is there a reason why we are not running any test in the image like we do with the other images?

We could certainly add such tests, but that wouldn't in itself test that the release auto-upload works.

celinval:
No, but it would test the image which I think is more important.

adpaco-aws (approved):

adpaco-aws:
Thanks @tautschnig for adding this! 😃 

> Wait, is there a reason why we are not running any test in the image like we do with the other images?

@celinval I've opened https://github.com/model-checking/kani/issues/3331 to keep track of that work. Please feel free to comment if I missed something.

PR #3322, opened 2024-07-03
CI: Allow use of `node16` for release jobs

### LINKED ISSUE #3321: `BuildBundle-Linux` CI jobs are failing with `version GLIBC_2.28 not found`
`BuildBundle-Linux` CI jobs are failing with ``version `GLIBC_2.28` not found`` as in [this CI run](https://github.com/model-checking/kani/actions/runs/9783697318/job/27013906335?pr=3320). This seems to be caused by https://github.com/actions/checkout/issues/1809 so it might be easier to just upgrade to v4.

adpaco-aws:
actions/checkout@v3 is causing trouble in the `BuildBundle-Linux` workflow as described in #3321 .

This PR sets `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true` to continue using `node16` so we don't find issues with `node20` dependencies in `ubuntu-18.04` jobs. This is the option recommended in this GitHub post: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

Resolves #3321 

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,7 @@ on:
 
 env:
   RUST_BACKTRACE: 1
+  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
 
 jobs:
   build_bundle_macos:


adpaco-aws:
Looks like the CI job also fails with v4 😢

adpaco-aws:
Tried adding the following to the `build_bundle_linux` job:

```yaml
  env:
    ACTIONS_RUNNER_FORCED_INTERNAL_NODE_VERSION: node16
    ACTIONS_RUNNER_FORCE_ACTIONS_NODE_VERSION: node16
```
But it didn't work. Not sure how to export these variables before starting the runners in our case.

adpaco-aws:
Replacing the action used in the `build_bundle_linux` job with `taiki-e/checkout-action@v1` (a 3rd party action that doesn't use node) avoids the error until the `actions/upload-artifact` action is used because it uses the same dependencies (see [this run](https://github.com/model-checking/kani/actions/runs/9841237127/job/27167372493?pr=3322) for example).

tautschnig (approved):
Approving except the PR title now needs adjusting.

tautschnig:
Maybe worth adding a link to https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/ to the PR description (and, thus, the git commit message).

adpaco-aws:
> I've also tried to set ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true at the workflow level (as suggested by https://github.com/actions/checkout/issues/1809#issuecomment-2208202462) but it didn't work (see run [here](https://github.com/model-checking/kani/actions/runs/9842615863/job/27171909854?pr=3322)).

Note: @tautschnig told us this didn't work because v4 will always try to use `node20`. The flag works if we don't upgrade the action's version.

> Maybe worth adding a link to https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/ to the PR description (and, thus, the git commit message).

Yes, added the link. Thanks!

PR #3323, opened 2024-07-03
Fix visibility of some Kani intrinsics

artemagvanian:
This PR fixes inadvertently exposing some of the unstable Kani intrinsics.


jaisnan (commented):
Can you change the equivalent methods in `kani_core` as well? Is that necessary?

artemagvanian:
> Can you change the equivalent methods in `kani_core` as well? Is that necessary?

Yes, it is -- thanks for catching this

artemagvanian:
It seems like the shadow memory tests depend on the shadow memory functions to be public -- I wonder if this is desired

celinval:
For now, I would just add the unstable tag and `#[doc(hidden)]` to them so we don't have to change the tests.

celinval (approved):
Thanks!

PR #3332, opened 2024-07-08
Make Kani reject mutable pointer casts if padding is incompatible and memory initialization is checked

### LINKED ISSUE #3324: Uninitialized memory detection mitigates delayed UB, but does not support it fully
I tried this code:

```rust
#[kani::proof]
fn invalid_value() {
    unsafe {
        let mut value: u128 = 0;
        let ptr = &mut value as *mut _ as *mut (u8, u32, u64);
        *ptr = (4, 4, 4);   // This assignment itself does not cause UB...
        assert!(value > 0); // ...but this reads a padding value! ⚠️
    }
}

```

```
kani -Z ghost-state -Z uninit-checks delayed-ub.rs
```

with Kani version: 0.53

I expected to see this happen: Verification should fail since we are reading padding values.

artemagvanian:
This PR introduces layout checks for types to instrument mutable pointer casts. If two types have incompatible padding (e.g. a padding byte in one is a data byte in the other or vice-versa), an "unsupported check" assertion is inserted. This overapproximates for soundness, since the casts do not cause UB themselves, but an alternative solution involves tracking every MIR place, which is costly.

Resolves #3324


celinval (approved):
Thanks

Issue #3257, opened 2024-06-11
Cleanup work due to enabling standard library verification


celinval:
Keeping a list of things that we need to improve to make the code more maintainable and less fragile:

- [x] Unify `kani` and `kani_core` logic. #3230 introduced duplicated logic.
- [x] Improve contract attribute handling.
  - We cannot just blindly duplicate the attributes, and for now we specialize on a few attributes (see #3255).
- [x] Safe handling rustc ast lowering in cases where they add extra generic argument.
  - One issue we bumped into is with constant functions and feature `effect` enabled.
- [x] Rename `unstable` function and re-add them to the `kani_core::mem` functions.
- [x] Enable `any_array` for the standard library.

PR #3333, opened 2024-07-09
Unify kani library and kani core logic

jaisnan:
1. Unifies `Kani library` and `kani_core` for `Arbitrary`, `mem` and `intrinsics` along with the `internal` module.
2. Adjusts some regression expected files to reflect the new underlying source of Arbitrary

Related to #https://github.com/model-checking/kani/issues/3257


library/kani/src/lib.rs
```diff
@@ -316,6 +107,24 @@ macro_rules! cover {
     };
 }
 
+impl<T> Arbitrary for std::boxed::Box<T>
+where
+    T: Arbitrary,
+{
```
adpaco-aws, line 113: Don't we want these implementations to live in `library/kani_core/src/arbitrary.rs`?

library/kani/src/vec.rs
adpaco-aws, line 1: What's the reason to keep this here? Can't we move it to `kani_core` just like the module `mem`?

library/kani_core/src/lib.rs
```diff
@@ -360,6 +365,13 @@ macro_rules! kani_intrinsics {
             #[doc(hidden)]
             #[rustc_diagnostic_item = "KaniInitContracts"]
             pub fn init_contracts() {}
+
+            /// This should only be used within contracts. The intent is to
+            /// perform type inference on a closure's argument
+            #[doc(hidden)]
+            pub fn apply_closure<T, U: Fn(&T) -> bool>(f: U, x: &T) -> bool {
```
adpaco-aws, line 372: It's not clear to me why we're not removing `apply_closure` in `library/kani/src/internal.rs`.

library/kani_macros/src/sysroot/contracts/shared.rs
```diff
@@ -66,9 +66,6 @@ impl<'a> ContractConditionsHandler<'a> {
         // We should consider a better strategy than just duplicating all attributes.
         #[cfg(feature = "no_core")]
         {
-            self.output.extend(quote!(
-                #[unstable(feature="kani", issue="none")]
-            ));
```
adpaco-aws, line 71: The comment above needs to be updated.

tests/expected/function-contract/modifies/check_invalid_modifies.expected
```diff
@@ -1,7 +1,7 @@
-error: `&str` doesn't implement `kani::Arbitrary`\
-   -->\
+error: `&str` doesn't implement `kani::Arbitrary`.
+  -->
```
adpaco-aws, line 2: No need to keep this arrow.

adpaco-aws (commented):
> In order for this cleanup to work, we need to remove the unstable feature being inserted into every `kani_sysroot` related compilation. This does not change anything for the end user, as long as they add 
> 
> ```
> #[unstable(feature="kani", issue="none")]
> ```
>
> when they invoke `kani` in core.

Do you mean that if a user wants to analyze `core` or `std` with Kani, they need to manually add this attribute to the Kani functions and compile from source?

library/kani/src/lib.rs
jaisnan: I would say no for now. A key requirement for anything in Kani core is that it needs to be something that has an implementation in the core library as well as std, this allows us to switch between the two source locations at compile time once the macros get expanded. Based on what i found, it doesnt seem like there are no_core implementations of boxed or box or time. 

What this means is that, we can have some methods as extensions of our core API, but are not a part of our core API since `kani_core` is a minimal dependency crate.

library/kani_core/src/lib.rs
jaisnan: Thanks for reminding me. We should be removing the entire file `internal.rs` itself, as it is simply duplicated code.

library/kani/src/vec.rs
jaisnan: No particular reason, i wanted to keep this PR minimal and focused on removing duplicated code, not adding new code to `kani_core`, which we most certainly can do in the future. Lmk if you think that should be a part of this PR as well.

jaisnan:
> > In order for this cleanup to work, we need to remove the unstable feature being inserted into every `kani_sysroot` related compilation. This does not change anything for the end user, as long as they add
> > ```
> > #[unstable(feature="kani", issue="none")]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > when they invoke `kani` in core.
> 
> Do you mean that if a user wants to analyze `core` or `std` with Kani, they need to manually add this attribute to the Kani functions and compile from source?

No, what I meant was that they need to add this to their invocation below `cfg(kani)` like we do [here](https://github.com/model-checking/verify-rust-std/blob/df109daf13a1cdb173a0d38f18d5e9a5b7f5b9ed/library/core/src/mem/mod.rs#L1354C1-L1356C13).

library/kani_core/src/lib.rs
jaisnan: Redundant file has been removed.

library/kani_macros/src/sysroot/contracts/shared.rs
celinval: What's the failure you are getting?

jaisnan: I see failures with nearly all function-contract related regression tests (71 failed). The exact failure depends on the type of function contract attribute used, but the underlying error seems to be the same.

For example, this is one of the failures for an `ensures` example. 

```
---- [expected] expected/function-contract/checking_from_external_mod.rs stdout ----

error: test failed: expected output to contain the line(s):
- Status: SUCCESS
- Description: "|result : &u32| (*result == x) | (*result == y)"
in function max
status: exit status: 1
command: "kani" "/home/ubuntu/kani/tests/expected/function-contract/checking_from_external_mod.rs" "-Zfunction-contracts"
stdout:
Kani Rust Verifier 0.53.0 (standalone)
error: /home/ubuntu/kani/target/kani/bin/kani-compiler exited with status exit status: 101

stderr:
error[E0734]: stability attributes may not be used outside of the standard library
 --> /home/ubuntu/kani/tests/expected/function-contract/checking_from_external_mod.rs:5:1
  |
5 | #[kani::ensures(|result : &u32| (*result == x) | (*result == y))]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |

For more information about this error, try `rustc --explain E0734`.

```

and another example, 

```


---- [expected] expected/function-contract/arbitrary_requires_pass.rs stdout ----

error: test failed: expected output to contain the line(s):
arithmetic_overflow
- Status: SUCCESS
- Description: "attempt to divide by zero"
in function div
status: exit status: 1
command: "kani" "/home/ubuntu/kani/tests/expected/function-contract/arbitrary_requires_pass.rs" "-Zfunction-contracts"
stdout:
Kani Rust Verifier 0.53.0 (standalone)
error: /home/ubuntu/kani/target/kani/bin/kani-compiler exited with status exit status: 101

stderr:
error[E0734]: stability attributes may not be used outside of the standard library
 --> /home/ubuntu/kani/tests/expected/function-contract/arbitrary_requires_pass.rs:5:1
  |
5 | #[kani::requires(divisor != 0)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the attribute macro `kani::requires` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0734`.

```

Removing that blob that injects the unstable attribute, fixes all 71 errors at once.

library/kani/src/lib.rs
celinval: But shouldn't this be inside `kani/src/arbitrary.rs`

library/kani_core/src/lib.rs
```diff
@@ -384,6 +389,13 @@ macro_rules! kani_intrinsics {
             #[doc(hidden)]
             #[rustc_diagnostic_item = "KaniInitContracts"]
             pub fn init_contracts() {}
+
+            /// This should only be used within contracts. The intent is to
+            /// perform type inference on a closure's argument
+            #[doc(hidden)]
```
celinval, line 395: What about the `kani::unstable` attribute?

library/kani_macros/src/sysroot/contracts/shared.rs
celinval: Yes, I think this is worth a bit more investigation. As you mentioned, this should not be triggered here. It's possible that the `unstable` proc_macro is broken too.

jaisnan: This part is now fixed. There is no need to remove this unstable injection, and the final result is working as expected. This error was caused by this addition to cargo.toml which made no_core the default feature for kani.
Resolving the comment now.

PR #3338, opened 2024-07-12
Mitigate invalid `transmute` when checking memory initialization

artemagvanian:
This PR addresses another aspect of #3324, where delayed UB could be caused by transmuting a mutable pointer into the one of incompatible padding. It also adds a check to error whenever transmuting between two types of incompatible padding:

```rust
// 5 bytes of data + 3 bytes of padding.
#[repr(C)]
#[derive(kani::Arbitrary)]
struct S(u32, u8);

/// Checks that Kani catches an attempt to access padding of a struct using transmute.
#[kani::proof]
fn check_uninit_padding() {
    let s = kani::any();
    access_padding(s);
}

fn access_padding(s: S) {
    let _padding: u64 = unsafe { std::mem::transmute(s) }; // ~ERROR: padding bytes are uninitialized, so reading them is UB.
}
```

### DIFF
--- a/kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
+++ b/kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
@@ -37,6 +37,8 @@ pub enum MemoryInitOp {
     SetRef { operand: Operand, value: bool, position: InsertPosition },
     /// Unsupported memory initialization operation.
     Unsupported { reason: String },
+    /// Operation that trivially accesses uninitialized memory, results in injecting `assert!(false)`.
+    TriviallyUnsafe { reason: String },
 }
 
 impl MemoryInitOp {
@@ -63,7 +65,9 @@ impl MemoryInitOp {
                 },
                 projection: vec![],
             }),
-            MemoryInitOp::Unsupported { .. } => unreachable!(),
+            MemoryInitOp::Unsupported { .. } | MemoryInitOp::TriviallyUnsafe { .. } => {
+                unreachable!()
+            }
         }
     }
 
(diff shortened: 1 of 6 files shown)


kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
```diff
@@ -581,6 +581,29 @@ impl<'a> MirVisitor for CheckUninitVisitor<'a> {
...
+                            // If casting from a mutable pointer to a mutable pointer with different
+                            // layouts, delayed UB could occur.
+                            self.push_target(MemoryInitOp::Unsupported {
+                                reason: "Kani does not support reasoning about memory initialization in presence of mutable raw pointer casts that could cause delayed UB.".to_string(),
+                            });
+                        }
+                    } else if !tys_layout_compatible(&operand_ty, &ty) {
+                        // If transmuting between two types of incompatible layouts, padding
+                        // bytes are exposed, which is UB.
+                        self.push_target(MemoryInitOp::Unsupported {
```
celinval, line 601: Why is this unsupported? Shouldn't it be a check?

celinval (commented):
Just one minor comment. Thanks!

kani-compiler/src/kani_middle/transform/check_uninit/uninit_visitor.rs
artemagvanian: `Unsupported` compiles into `assert!(false)`, which serves the same purpose. I could rename it to something like `Fail` or `AssertFail`

celinval: I would recommend that you create a new variant since they represent different things. They can be implemented the same way though. Thanks!

celinval (approved):
It looks good. Just please add the new invariant to distinguish the unsupported vs the trivially false cases. Thanks!

PR #3344, opened 2024-07-15
Support for disabling automatically generated pointer checks to avoid reinstrumentation

artemagvanian:
Recently added memory initialization checks (see #3300 for an overview) code suffers from re-instrumentation due to automatically added pointer checks to the instrumentation code. 

This PR adds an internal `kanitool::disable_checks` attribute to disable automatically generated CBMC checks inside internal instrumentation. This, in turn, relies on CBMC pragmas (https://www.cprover.org/cprover-manual/properties/#usinggotoinstrument).


kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -74,6 +74,9 @@ enum KaniAttributeKind {
     /// We use this attribute to properly instantiate `kani::any_modifies` in
     /// cases when recursion is present given our contracts instrumentation.
     Recursion,
+    /// Used to mark functions where generating automatic pointer checks should be disabled. This is
+    /// used later to automatically attach pragma statements to locations.
+    SkipPtrChecks,
```
qinheping, line 79: We might also want to disable other checks in this way in the future. Could you make this associated with a list of strings representing the checks we want to disable?
 
Also, I think the name `DisableChecks` will be more aligned with the name of disable-checks-pragmas in CBMC.

qinheping (commented):
Could you add some tests to illustrate that the checks are actually disabled in the GOTO programs?

artemagvanian:
> Could you add some tests to illustrate that the checks are actually disabled in the GOTO programs?

I wonder what's a good way to test this: if there is a way to somehow retrieve the internal output of goto-instrument, we could check that the appropriate pragmas are added? Since this attribute is supposed to be internal, we would need to test it indirectly

qinheping:
> internal output of goto-instrument, we could check that the appropriate pragmas are added? Since this attribute is supposed to be internal, we would need to test it indirectly

Could you elaborate a bit more about what will happen if we don't disable the pointer checks in the  instrumentation code. Is there any test failed without this PR? If so, we can use them to test this PR indirectly.

Directly testing checks being disabled requires extending `.expected` with exclude-tag --- specifying lines that we don't want to see in the Kani output.

artemagvanian:
> Could you elaborate a bit more about what will happen if we don't disable the pointer checks in the instrumentation code. Is there any test failed without this PR? If so, we can use them to test this PR indirectly.

No test will fail, but extra checks will be added to the instrumentation code itself, which negatively affects verification performance.

> Directly testing checks being disabled requires extending `.expected` with exclude-tag --- specifying lines that we don't want to see in the Kani output.

That could work, let me look into that

adpaco-aws:
> if there is a way to somehow retrieve the internal output of goto-instrument, we could check that the appropriate pragmas are added?

IIRC the human-readable output of `goto-instrument` (via `--show-functions` or something like that) doesn't pretty-print pragmas, so I'd suggest exploring other options.

> Directly testing checks being disabled requires extending .expected with exclude-tag --- specifying lines that we don't want to see in the Kani output.

For more complex behaviors we often recommend [the `exec` mode](https://model-checking.github.io/kani/regression-testing.html#the-exec-mode), which allows you to run a script to determine the outcome and optionally check the output and exit code.

kani-compiler/src/kani_middle/attributes.rs
celinval: I like the idea, but  I would avoid requiring the strings to match the ones from CBMC. In that case, make sure you document which strings are accepted

celinval: We may want to use this later to disable Kani instrumentations

artemagvanian:
@qinheping @adpaco-aws thanks for the pointers. I added a test that checks that automatic checks are not added in the module where we disable it. This feels clunky, but I can't think of a better way since the attribute is private

kani-compiler/src/kani_middle/attributes.rs
artemagvanian: Addressed that in the latest commit

tests/script-based-pre/mem-init-reinstrumentation/mem-init-reinstrumentation.expected
```diff
@@ -0,0 +1 @@
+success: no pointer checks are detected in initialized memory instrumentaiton
```
qinheping, line 1: Missing EOL.

qinheping (approved):
LGTM. Thank you!

PR #3348, opened 2024-07-16
Add support for global transformations

artemagvanian:
This PR adds basic support for global transformations, which will be useful to implement full support of #3324

Changes:
- Add edge annotations to the call graph
- Implement global transformation pass infrastructure
- Implement `dump_mir` as a global transformation pass


kani-compiler/src/kani_middle/transform/dump_mir_pass.rs
```diff
@@ -0,0 +1,73 @@
...
+impl GlobalPass for DumpMirPass {
+    fn is_enabled(&self, _query_db: &QueryDb) -> bool {
+        self.enabled
+    }
+
+    fn transform(
+        &mut self,
+        tcx: TyCtxt,
+        starting_items: &[MonoItem],
+        bodies: Vec<(Body, Instance)>,
```
celinval, line 45: I don't think we should be passing the vector of bodies. We could pass the `BodyTransformer` instead or some abstraction. It would be nice to be able to change this if the memory consumption gets too high.

kani-compiler/src/kani_middle/transform/dump_mir_pass.rs
```diff
@@ -0,0 +1,73 @@
...
+        _call_graph: &CallGraph,
+    ) -> Vec<(bool, Body, Instance)> {
+        // Create output buffer.
+        let file_path = {
+            let base_path = tcx.output_filenames(()).path(OutputType::Object);
+            let base_name = base_path.as_path();
+            let entry_point = (starting_items.len() == 1).then_some(starting_items[0].clone());
+            if let Some(MonoItem::Fn(starting_instance)) = entry_point {
```
celinval, line 53: I think we should stick to the harness name.

kani-compiler/src/kani_middle/transform/dump_mir_pass.rs
```diff
@@ -0,0 +1,73 @@
...
+        let out_file = File::create(file_path.with_extension("kani.mir")).unwrap();
+        let mut writer = BufWriter::new(out_file);
+
+        // For each def_id, dump their MIR
+        for (body, instance) in bodies.iter() {
+            writeln!(writer, "// Item: {} ({})", instance.name(), instance.mangled_name()).unwrap();
+            let _ = body.dump(&mut writer, &instance.name());
+        }
+
+        bodies.into_iter().map(|(body, instance)| (false, body, instance)).collect()
```
celinval, line 71: If we pass the transformer, we can just store the modified bodies.

kani-compiler/src/kani_middle/transform/mod.rs
```diff
@@ -118,6 +128,27 @@ impl BodyTransformation {
         }
     }
 
+    /// Apply all per-body pass followed by the global passes and store the results in cache that
+    /// can later be queried by `body`.
+    pub fn apply_passes(
```
celinval, line 133: maybe `run_global_passes`

kani-compiler/src/kani_middle/transform/mod.rs
```diff
@@ -118,6 +128,27 @@ impl BodyTransformation {
         }
     }
 
+    /// Apply all per-body pass followed by the global passes and store the results in cache that
```
celinval, line 131: wait... why are we applying the per-body passes again?

kani-compiler/src/kani_middle/reachability.rs
```diff
@@ -118,7 +125,24 @@ where
...
+    VTableMethod,
+    Static,
+    StaticDrop,
+}
+
+/// A destination of the edge in the call graph.
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+struct CollectedItem {
+    item: MonoItem,
+    reason: CollectionReason,
```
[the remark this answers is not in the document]
artemagvanian: It ultimately depends on how we are going to use those reasons. I think we could change that once we have a good example of using those in a real global transformation pass.

kani-compiler/src/kani_middle/transform/dump_mir_pass.rs
artemagvanian: That's fine by me. I just copied whatever was there before, but happy to make it simpler

kani-compiler/src/kani_middle/transform/dump_mir_pass.rs
artemagvanian: The problem with passing `&mut BodyTransformer` is that now there are two mutable references to it, the other one from the global pass we are mutably iterating over. I added a hacky solution to that by moving the global passes out before iteration and then moving them in, but I wonder whether it's worth it

kani-compiler/src/kani_middle/transform/dump_mir_pass.rs
artemagvanian: See the previous comment

kani-compiler/src/kani_middle/transform/dump_mir_pass.rs
celinval: oh... if that's what was there, we should be fine. I guess this is a vector, so the first item is a harness. :)

kani-compiler/src/kani_middle/transform/mod.rs
```diff
@@ -118,6 +128,21 @@ impl BodyTransformation {
...
 
+    /// Run all global passes and store the results in a cache that can later be queried by `body`.
+    pub fn run_global_passes(
+        &mut self,
+        tcx: TyCtxt,
+        starting_items: &[MonoItem],
+        instances: Vec<Instance>,
+        call_graph: CallGraph,
+    ) {
+        let mut global_passes = std::mem::take(&mut self.global_passes);
```
celinval, line 139: The other option would be to extract the cache and pass it here instead.

artemagvanian: I think this is less desirable because we would need to make sure the cache is already populated versus populating it on-demand

celinval: actually, does global passes need to be part of the Transformer at all?

artemagvanian: I don't think so, actually -- I believe we run them only in one place so I could change that

celinval (approved):
Awesome! Thanks

PR #3360, opened 2024-07-19
Harness output individual files

### LINKED ISSUE #3356: Harness Output in Individual Files
Requested feature: Harness Output in Individual Files
Use case: Easily parse output by having each harness output be an a unique file.
Printing is done to the std output but having each harness output be in a file helps with parsing. Possibly files named based on harness name.
This is currently done in each harness here in kani-driver line 124:
```rust
if !self.args.common_args.quiet && self.args.output_format != OutputFormat::Old {
    println!(
        "{}",
        result.render(
            &self.args.output_format,
            harness.attributes.should_panic,
            self.args.coverage
        )
    );
}
```

Alexander-Aghili:
Resolves #3356 

1. Changes allow for directory with individual output of files named by the full harness name. --output-into-files command line argument will allow for placing all output of individual harnesses into files named by the full harness pretty_name. The output directory is either --target-dir or a hard coded default: "kani_output" directory. (These can be changed if there is a better interface). Still prints output to std::out exactly as previous. 
2. Previously, all output was placed into std::out. This will allow for some control over output. It will also enable easier parsing of harness output.
3. Only solved #3356 but could be expanded to include more features.
4. Ran manually to check the flags and output behaved as expected. Indeed: 
--output-into-files enabled will place output into individual files, disabled will not
--output-terse will create terse output to command line, regular output to individual files if --output-into-files is enabled. 
--target-dir will place output into target-dir directory when --output-into-files is enabled, and will not place file output when disabled.

Let me know if you need specific explanations of code segments, a clean list of all the testing configurations, or any feature enhancement for greater configuration options.


kani-driver/src/harness_runner.rs
```diff
@@ -101,6 +104,46 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
...
+        std::fs::create_dir_all(prefix).unwrap();
+        let mut file = File::create(&file_name).unwrap();
+        let file_output = result.render(&OutputFormat::Regular, harness.attributes.should_panic, self.args.coverage);
+
+        if let Err(e) = writeln!(file, "{}", file_output) {
+            eprintln!("Failed to write to file {}: {}", file_name, e);
+        }
+    }
+
+    fn get_target_dir(&self) -> String {
```
celinval, line 138: Can you please rename this to `get_result_output_dir` to separate these files from the other artifacts that Kani may generate?

kani-driver/src/harness_runner.rs
```diff
@@ -101,6 +104,46 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
...
+        self.args.target_dir
+            .clone()
+            .map_or_else(|| "./kani_output".to_string(), |dir| {
+                format!("{}/", dir.into_os_string().into_string().unwrap())
+            })
```
celinval, line 143: Maybe something like:
```suggestion
    let target_dir = self.args.target_dir
            .clone()
            .map_or_else(|| current_dir(), |dir| Ok(dir))?;
    let outdir = target_dir.join("kani_results");
    fs::create_dir_all(&outdir)?; // This is a no-op if directory exists.
```

Alexander-Aghili: Okay, I will get to fixing this on late Sunday or Monday as I am away this weekend. Thanks for the review!

kani-driver/src/args/mod.rs
```diff
@@ -275,6 +275,10 @@ pub struct VerificationArgs {
     )]
     pub synthesize_loop_contracts: bool,
 
+    //Harness Output into individual files
+    #[arg(long, hide_short_help = true, requires("enable_unstable"))]
```
celinval, line 279: I forgot to mention. We are deprecating `enable_unstable`, so can you please make sure the new feature can be enabled via `-Z unstable-options` instead?

jaisnan:
This output format is untested as of this PR. I'd like to see one regression test added at least. This isn't a blocking request, but likely something we want to add immediately, maybe in a follow up PR.

kani-driver/src/args/mod.rs
Alexander-Aghili: Hi there, is there any example of how to do that?

Alexander-Aghili:
> This output format is untested as of this PR. I'd like to see one regression test added at least. This isn't a blocking request, but likely something we want to add immediately, maybe in a follow up PR.

I agree that tests should be added. I'd be fine adding it as a component of this PR or as a separate one, whichever is easier for you guys.

kani-driver/src/args/mod.rs
celinval: Yes, unfortunately this cannot be done with clap directly. Here is an example:

https://github.com/model-checking/kani/blob/7ad4d1cc3109defea1131876f427f13599243ed8/kani-driver/src/args/mod.rs#L632-L640

In this case, you can use [`UnstableFeature::UnstableOptions`](https://github.com/model-checking/kani/blob/7ad4d1cc3109defea1131876f427f13599243ed8/kani_metadata/src/unstable.rs#L97).

Alexander-Aghili:
Hey guys, sorry I haven't gotten to this in a minute. I was testing this on a server used by my research group, but my laptop's SSD failed with my ssh key so I can't really access the server at the moment. Should be fixed in a couple days. Thanks for the patience.

Alexander-Aghili:
Okay I tested it and added the -Z --unstable-options instead of --enable-unstable. Do we still want to add tests as a part of this PR?

jaisnan:
> Okay I tested it and added the -Z --unstable-options instead of --enable-unstable. Do we still want to add tests as a part of this PR?

I'm okay without a regression for this specific output format since its a really special case and testing will involve some work. I can take that up in a follow up PR. You might want to do the following though,

1. Add comments on the functions related to processing the output.
2. Format the code with `cargo fmt` and run `cargo clippy` on it once.
3. (Optionally) Create an issue to test the new output format.

artemagvanian:
> Hey there, I've done most of it but I ran into this issue when running regression tests: ERROR: CBMC version is 5.89.0, expected at least 5.95.0 How can I upgrade the CBMC version?

You can re-run the script responsible for installing the dependencies to install the appropriate CBMC version. Here is a link to the wiki page that explains what exactly to run: https://model-checking.github.io/kani/build-from-source.html#dependencies.

zhassan-aws:
If you're using a Mac, run:
```bash
./scripts/setup/macos/install_cbmc.sh
```
If you're using Ubuntu, run:
```bash
./scripts/setup/ubuntu/install_cbmc.sh
```

Alexander-Aghili:
I don't want to be the bottleneck on this but in short, I am having a difficult time upgrading because of permissions issues on my remote server, and thus, if others want to try and fix the regression tests, then that is fine as well. Otherwise I am going to have to spend some time fixing the issue on the server I do my development work on and I'm out of town.

Alexander-Aghili:
Also added  #3412 for output testing

kani-driver/src/harness_runner.rs
```diff
@@ -101,6 +107,54 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
...
+                "Failed to write to file {}: {}",
+                file_name.into_os_string().into_string().unwrap(),
+                e
+            );
+        }
+    }
+
+    fn result_output_dir(&self) -> Result<PathBuf> {
+        let target_dir =
+            self.args.target_dir.clone().map_or_else(current_dir, Ok)?;
```
celinval, line 154:
```suggestion
        let target_dir = self.args.target_dir.clone().map_or_else(current_dir, Ok)?;
```

This should hopefully fix the format issue from the CI.

jaisnan:
> I don't want to be the bottleneck on this but in short, I am having a difficult time upgrading because of permissions issues on my remote server, and thus, if others want to try and fix the regression tests, then that is fine as well. Otherwise I am going to have to spend some time fixing the issue on the server I do my development work on and I'm out of town at the moment.

Hi @Alexander-Aghili, the regression tests should pass now after the fix from @celinval. You should be able to resume development if you have the time to add tests. If not, please let us know and I can take on the task of adding a test on this PR 😄

Alexander-Aghili:
Hi guys,
Sorry for taking so long, I've been out on holiday. I made an issue for adding regression tests. Do we want to complete that in this PR? If not, what needs to be done with this PR to merge it? If we do, how should the regression test be done?

jaisnan:
> Hi guys, Sorry for taking so long, I've been out on holiday. I made an issue for adding regression tests. Do we want to complete that in this PR? If not, what needs to be done with this PR to merge it? If we do, how should the regression test be done?

For this pr to be merged, we are looking for at least something that uses the new flag and ensure Kani still succeeds. It doesn't need to check that the files exist, just needs to test that Kani doesn't break. If you can add that, I am OK with merging the PR. Lmk if thats putting too much on your plate, and I can take over this PR and add those changes myself as well. In any case, thank you very much for the changes :D.

Alexander-Aghili:
I swear I'm not as flaky as I seem but my computer's screen just broke and I'm going to have to get it replaced which might take some time. If you want to take it over and finish the PR feel free to do so otherwise I'll get on it immediately when I can.

jaisnan:
> I swear I'm not as flaky as I seem but my computer's screen just broke and I'm going to have to get it replaced which might take some time. If you want to take it over and finish the PR feel free to do so otherwise I'll get on it immediately when I can.

No worries :D

Alexander-Aghili:
Finally getting back around to this. I will be creating that test and merging in the new base, then asking for a review to be merged.

jaisnan:
> Finally getting back around to this. I will be creating that test and merging in the new base, then asking for a review to be merged.

Awesome, thank you!

Alexander-Aghili:
Hey, I was checking the other UI tests and they verify the validity of the response by running a proof in a main.rs file and checking it with an expected output which is a single txt. With the proper output being a directory, how should I handle that? Also, do you use git merge or git rebase when updating the git structure?

Alexander-Aghili:
So I think when I merged it included all of the commits in the commit history here. Is there a way to prevent that in the future? Also, I looked at the script-based-pre testing and it seemed like the .expected file checks the output of the command to the command line but I wasn't sure how to verify the generation of the output to directories. I decided that I would do it through a script which is almost done. Would it be better to do it another way or do you need to see the code first to decide?

celinval:
> So I think when I merged, it included all of the commits in the commit history here. Is there a way to prevent that in the future? Also, I looked at the script-based-pre testing and it seemed like the .expected file checks the output of the command to stdout but I wasn't sure how to verify the generation of the output to directories. I decided that I would do it through a script which is almost done. Would it be better to do it another way or do you need to see the code first to decide?

Hi @Alexander-Aghili,  the only problem I see is that you seem to have accidentally a `target/tools` folder. If you remove this, it should be good to go.

Alexander-Aghili:
I updated with an test to ensure the feature operates and behaves as expected... how could I remove the target/tools file from the commit if that matters because git rm --cached is not working.

celinval:
Just create a new commit. It's fine.

Alexander-Aghili:
Given it passes the regression tests, it should be good to go.

jaisnan:
> Hey everyone, any update here?

Almost there, can you resolve the merge conflicts and push again?

Alexander-Aghili:
@jaisnan Fixed

kani-driver/src/harness_runner.rs
```diff
@@ -69,6 +79,8 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
                 .collect::<Result<Vec<_>>>()
         })?;
 
+        println!("Elapsed Time: {:.2?}\n", before.elapsed());
```
celinval, line 82: This also seems unrelated to this PR.

kani-driver/src/harness_runner.rs
```diff
@@ -39,15 +45,19 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
         self.check_stubbing(harnesses)?;
 
         let sorted_harnesses = crate::metadata::sort_harnesses_by_loc(harnesses);
-
+        let max_threads = 8;
```
celinval, line 48: Why did you have to change this?

Alexander-Aghili: Ya this was a test of something else, accidentally slipped through, idk how, sorry. Will fix it up.

Alexander-Aghili:
Okay I updated it. However, it seems like sometimes the tests fail only on the CI?

tests/script-based-pre/individual_file_output/individual_file_output.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Ensure output folder is clean
+rm -rf ${OUT_DIR}
+
+# Move the original source to the output folder since it will be modified
+cp -r sample_crate ${OUT_DIR}
+pushd $OUT_DIR
+
+echo "Run verification..."
+../../../../scripts/cargo-kani -Z unstable-options --output-into-files
```
celinval, line 17: You should just call `cargo kani` here. I think our test infrastructure should ensure path is correct

tests/script-based-pre/individual_file_output/individual_file_output.sh
```diff
@@ -0,0 +1,46 @@
...
+
+OUTPUT_DIR="kani_results"
```
celinval, line 19: Where are you creating this folder? If Kani is generating this folder, shouldn't it be inside target?

celinval (commented):
What is the error?

kani-driver/src/harness_runner.rs
```diff
@@ -39,12 +44,8 @@ impl<'sess, 'pr> HarnessRunner<'sess, 'pr> {
         self.check_stubbing(harnesses)?;
 
         let sorted_harnesses = crate::metadata::sort_harnesses_by_loc(harnesses);
-
         let pool = {
-            let mut builder = rayon::ThreadPoolBuilder::new();
-            if let Some(x) = self.sess.args.jobs() {
-                builder = builder.num_threads(x);
-            }
+            let builder = rayon::ThreadPoolBuilder::new();
```
celinval, line 48:
```suggestion
            let mut builder = rayon::ThreadPoolBuilder::new();
            if let Some(x) = self.sess.args.jobs() {
                builder = builder.num_threads(x);
            }
```

Is there a reason why you reverted this? If not, can you please revert it?

celinval (approved):
Awesome!! Thanks

jaisnan:
Before we merge, can you resolve the merge conflicts locally? It'll allow us to see CI as well

Alexander-Aghili:
Anything more needed here?

zhassan-aws:
We should be good to go after resolving the conflicts.

celinval:
I solved the conflict. So if the CI passes, this should merge. Thanks again for the contribution @Alexander-Aghili!

Alexander-Aghili:
Hooray! Thanks for all the help and guidance everyone.

PR #3363, opened 2024-07-20
Enable contracts in associated functions

### LINKED ISSUE #3206: Error when annotating an associated function with a contract

celinval:
Contracts could not be used with associated function unless they used Self. This is because at a proc-macro level, we cannot determine if we are inside a associated function or not, except in cases where `Self` is used. In cases where the contract generation could not identify an associated function, it would generate incorrect call, triggering a compilation error.

Another problem with our encoding is that users could not annotate trait implementations with contract attributes. This would try to create new functions inside the trait implementation context, which is not allowed.

In order to solve these issues, we decided to wrap contract logic using closures instead of functions. See the discussion in the original issue (#3206) for more details.

The new solution is still split in two:
1. The proc-macro will now expand the code inside the original function to encode all possible scenarios (check, replace, recursive check, and original body).
2. Instead of using stub, we introduced a new transformation pass that chooses which scenario to pick according to the target harness configuration, and cleanup unused logic.

The expanded function will have the following structure:
```rust
#[kanitool::recursion_check = "__kani_recursion_check_modify"]
#[kanitool::checked_with = "__kani_check_modify"]
#[kanitool::replaced_with = "__kani_replace_modify"]
#[kanitool::inner_check = "__kani_modifies_modify"]
fn name_fn(ptr: &mut u32) {
    #[kanitool::fn_marker = "kani_register_contract"]
    pub const fn kani_register_contract<T, F: FnOnce() -> T>(f: F) -> T {
        kani::panic("internal error: entered unreachable code: ")
    }
    let kani_contract_mode = kani::internal::mode();
    match kani_contract_mode {
        kani::internal::RECURSION_CHECK => {
            #[kanitool::is_contract_generated(recursion_check)]
            let mut __kani_recursion_check_name_fn = || { /* recursion check body */ };
            kani_register_contract(__kani_recursion_check_modify)
        }
        kani::internal::REPLACE => {
            #[kanitool::is_contract_generated(replace)]
            let mut __kani_replace_name_fn = || { /* replace body */ };
            kani_register_contract(__kani_replace_name_fn)
        }
        kani::internal::SIMPLE_CHECK => {
            #[kanitool::is_contract_generated(check)]
            let mut __kani_check_name_fn = || { /* check body */ };
            kani_register_contract(__kani_check_name_fn)
        }
        _ => { /* original body */ }
    }
}
```
In runtime, `kani::internal::mode()` will return kani::internal::ORIGINAL, which runs the original body. The transformation will replace this call by a different assignment in case the function needs to be replaced. The body of the unused closures will be replaced by a `UNREACHABLE` statement to avoid unnecessary code to be analyzed.

This is still fairly hacky, but hopefully we can cleanup this logic once Rust adds contract support. :crossed_fingers: 

Resolves #3206


tests/expected/function-contract/simple_replace_pass.expected
```diff
@@ -1,4 +1,4 @@
-div.assertion\
+.assertion\
```
feliperodri, line 1: why are we missing the name of the functions here?

tests/expected/function-contract/modifies/havoc_pass.expected
```diff
@@ -1,38 +1,13 @@
...
-copy.assigns\
+.assigns\
 - Status: SUCCESS\
 - Description: "Check that var_4 is assignable"\
 in function copy
 
-copy.assigns\
-- Status: SUCCESS\
-- Description: "Check that var_3 is assignable"\
-in function copy\
```
feliperodri, line 16: Yeah, these are quite brittle. Any chance we can just add `var_\d+` to cover any number?

library/kani_macros/src/sysroot/contracts/mod.rs
```diff
@@ -400,36 +488,35 @@ pub fn proof_for_contract(attr: TokenStream, item: TokenStream) -> TokenStream {
...
     output: TokenStream2,
-    hash: Option<u64>,
+    /// The name of the check closure.
+    check_name: String,
+    /// The name of the replace closure.
+    replace_name: String,
+    /// The name of the recursion closure.
+    recursion_name: String,
+    /// The name of the modifies closure.
+    modify_name: String,
```
feliperodri, line 519: sometimes we refer as inner check sometimes as modify name, should we be more consistent here?

feliperodri (commented):
First batch of reviews. So far the test looks good and we have confidence in the pre-existing tests that the refactoring did not break any current behavior. I'm gonna give another pass in the instrumentation.

kani-compiler/src/kani_middle/transform/contracts.rs
```diff
@@ -232,8 +228,297 @@ impl AnyModifiesPass {
...
+            RigidTy::FnDef(def, args) => {
+                if let Some(mode) = self.contract_mode(tcx, *def) {
+                    self.mark_unused(tcx, *def, &body, mode);
+                    let new_body = self.set_mode(tcx, body, mode);
+                    (true, new_body)
+                } else if KaniAttributes::for_instance(tcx, instance).fn_marker()
+                    == Some(Symbol::intern("kani_register_contract"))
+                {
+                    let run = Instance::resolve(find_fn_def(tcx, "KaniRunContract").unwrap(), args)
+                        .unwrap();
```
feliperodri, line 322: should we use `.expect()` here?

kani-compiler/src/kani_middle/transform/contracts.rs
```diff
@@ -232,8 +228,297 @@ impl AnyModifiesPass {
...
+            let check_fn = attrs.interpret_for_contract_attribute().map(|(_, def_id, _)| def_id);
+            let replace_fns: HashSet<_> = attrs
+                .interpret_stub_verified_attribute()
+                .iter()
+                .map(|(_, def_id, _)| *def_id)
+                .collect();
+            FunctionWithContractPass { check_fn, replace_fns, unused_closures: Default::default() }
+        } else {
+            // Building the model for tests or public functions.
+            FunctionWithContractPass::default()
```
feliperodri, line 363: I don't understand what you mean here.

kani-compiler/src/kani_middle/transform/contracts.rs
```diff
@@ -232,8 +228,297 @@ impl AnyModifiesPass {
...
+    /// 1. Initialize `kani_contract_mode` variable to the value corresponding to the mode.
+    ///
+    /// Thus replace this call:
+    /// ```ignore
+    ///    let kani_contract_mode = kani::internal::mode(); // ** Replace this call
+    /// ```
+    /// by:
+    /// ```ignore
+    ///    let kani_contract_mode = mode_const;
+    ///    goto bbX;
```
feliperodri, line 398: Why this `goto bbX`?

kani-compiler/src/kani_middle/transform/contracts.rs
```diff
@@ -232,8 +228,297 @@ impl AnyModifiesPass {
...
+                self.unused_closures.insert(check_closure);
+            }
+            ContractMode::SimpleCheck => {
+                self.unused_closures.insert(replace_closure);
+                self.unused_closures.insert(recursion_closure);
+            }
+            ContractMode::Replace => {
+                self.unused_closures.insert(recursion_closure);
+                self.unused_closures.insert(check_closure);
+            }
```
feliperodri, line 485: If the idea is to completely replace the macros with closures, the way we need to handle modifies clauses doesn't impact at all here? We can still generate the inner function with macros?

library/kani_macros/src/sysroot/contracts/mod.rs
```diff
@@ -180,56 +156,106 @@
...
+//! #[kanitool::recursion_check = "__kani_recursion_check_div"]
+//! #[kanitool::checked_with = "__kani_check_div"]
+//! #[kanitool::replaced_with = "__kani_replace_div"]
+//! #[kanitool::inner_check = "__kani_modifies_div"]
+//! fn div(dividend: u32, divisor: u32) -> u32 {
+//!     #[inline(never)]
+//!     #[kanitool::fn_marker = "kani_register_contract"]
+//!     pub const fn kani_register_contract<T, F: FnOnce() -> T>(f: F) -> T {
+//!         kani::panic("internal error: entered unreachable code: ")
+//!     }
```
feliperodri, line 169: Why is this register necessary? Couldn't we just get the mode, set the `kani_contract_mode` and call the original function?

library/kani_macros/src/sysroot/contracts/mod.rs
```diff
@@ -67,61 +64,39 @@
...
-//!
-//!                        State is detected via
+//!                           │  Annotate original + generate closures
+//!                           │
+//!                           ▼
+//!                    ┌──────────┐
+//!                    │ Original │◄─┐
+//!                    └──┬───────┘  │
+//!                       │          │ Expand
+//!                       └──────────┘ closures
```
JustusAdam, line 74: What happened here? I think this new diagram version is not very helpful. We still generate different types of functions (closures in this case) and use marker annotations to distinguish how a particular clause modifies code so I think that should be reflected here. Especially now since this code generation is getting increasingly complex.

tests/expected/function-contract/modifies/havoc_pass.expected
JustusAdam: AFAIK the test harnesses do not support regexes. They are just substring matches :/

kani-compiler/src/kani_middle/transform/contracts.rs
JustusAdam: And should this be done in the constructor so we only do it once?

kani-compiler/src/kani_middle/transform/contracts.rs
JustusAdam: Because we're replacing a terminator. Unless we want to add logic to merge the basic blocks, we need to insert a new terminator (e.g. a goto) that gets us to the next basic block that used to follow the function call.

JustusAdam (commented):
I wonder if we couldn't vastly simplify the generated code by just inlining the closures into the match and pick a branch based on mode? Since we're already replacing the `mode` call with a constant, can't we just run like rustc's folding + dead code removal to select the branch for us?

kani-compiler/src/kani_middle/transform/body.rs
```diff
@@ -402,12 +419,21 @@ impl MutableBody {
...
     }
+
+    /// Replace a terminator from the given basic block
+    pub fn replace_terminator(
+        &mut self,
+        source_instruction: &SourceInstruction,
+        new_term: Terminator,
+    ) {
+        self.blocks.get_mut(source_instruction.bb()).unwrap().terminator = new_term;
+    }
```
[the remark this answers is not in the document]
celinval: In the contract transformation, we remove the call to `mode` function and replace it with an assignment + `goto` terminator.

library/kani/src/internal.rs
```diff
@@ -8,67 +8,48 @@ use std::ptr;
...
-    }
-
     unsafe fn assignable(self) -> *mut Self::Inner {
         self
     }
 }
 
 /// A way to break the ownerhip rules. Only used by contracts where we can
 /// guarantee it is done safely.
+/// TODO: Remove this! This is not safe. Users should be able to use `ptr::read` and `old` if
```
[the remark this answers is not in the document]
celinval: Yes, we have an issue already. I'll add it to this comment

tests/expected/function-contract/simple_replace_pass.expected
celinval: because the name now includes a closure. Something like `div.{closure#1}.assertion`

tests/expected/function-contract/modifies/havoc_pass.expected
celinval: The truth is that we shouldn't have temporaries in user messages.

kani-compiler/src/kani_middle/transform/contracts.rs
celinval: sure... we can do that

kani-compiler/src/kani_middle/transform/contracts.rs
celinval: I'm assuming you mean replace functions with closures. The modifies clauses is not important for the transformation because its logic is not directly reachable from the original function.

We don't need to delete the body, since it will be unreachable once we delete the body of the `check_closure` and `recursion_closure`.

kani-compiler/src/kani_middle/transform/contracts.rs
celinval: Ah, I can update this comment.  This is for cases where the reachability mode is not harnesses.

library/kani_macros/src/sysroot/contracts/mod.rs
celinval: I was actually tempted to remove this, since there is no state tracking. We have 2 states now, untouched and touched. In that sense, the state machine got simpler.

I agree that modifies and old complicates the code generation a bit though, but maybe we can keep the documentation more local so it's easier to maintain.

celinval:
> I wonder if we couldn't vastly simplify the generated code by just inlining the closures into the match and pick a branch based on mode? Since we're already replacing the `mode` call with a constant, can't we just run like rustc's folding + dead code removal to select the branch for us?

The thought did occur to me, but not sure the effort is worth it. Let me see what I can do in a few hours. I would prefer not spending too much time, since a lot of this will be throw away code once the compiler incorporate contracts.

celinval:
> I wonder if we couldn't vastly simplify the generated code by just inlining the closures into the match and pick a branch based on mode? Since we're already replacing the `mode` call with a constant, can't we just run like rustc's folding + dead code removal to select the branch for us?

Actually, we cannot do that today unless if we want to restrict contract of constant functions to only contain constant code. I remember bumping into this issue while trying to verify the `copy` intrinsics.

@feliperodri that is also the reason why we need the `kani_register_contract` function, i.e., we cannot invoke closures from constant functions.

library/kani_macros/src/sysroot/contracts/mod.rs
celinval: Talked offline, these are needed to allow contract of constant functions to contain non-constant logic.

tests/expected/function-contract/simple_replace_pass.expected
celinval: btw, I just didn't want to include the closure name here because it's auto generated and I'm not sure how stable that id is.

feliperodri (approved):
🚢 it!

PR #3375, opened 2024-07-24
Import apply_closure into kani_core

tautschnig:
This enables use of `|result|` when verifying the standard library.


jaisnan (approved):
This was a change that was to be merged in https://github.com/model-checking/kani/pull/3333 but that PR is currently blocked due to a leak in how the unstable proc macro works. However that does not affect this PR, so thanks for this change.

PR #3380, opened 2024-07-24
Use `box` in shadow memory test for reliable object count

tautschnig:
When using stack-allocated objects, MIR is not actually introducing fresh objects in this loop. Us treating them as fresh objects is just an artifact of interpreting StorageLive as creating a fresh object. Using `box` will make sure we genuinely have independent objects.

### DIFF
--- a/tests/expected/shadow/unsupported_num_objects/test.rs
+++ b/tests/expected/shadow/unsupported_num_objects/test.rs
@@ -15,14 +15,14 @@ fn check_max_objects<const N: usize>() {
     // - the NULL pointer whose object ID is 0, and
     // - the object ID for `i`
     while i < N {
-        let x = i;
-        assert_eq!(kani::mem::pointer_object(&x as *const usize), i + 2);
+        let x: Box<usize> = Box::new(i);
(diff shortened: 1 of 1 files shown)


tests/expected/shadow/unsupported_num_objects/test.rs
```diff
@@ -15,14 +15,14 @@ fn check_max_objects<const N: usize>() {
     // - the NULL pointer whose object ID is 0, and
     // - the object ID for `i`
     while i < N {
-        let x = i;
-        assert_eq!(kani::mem::pointer_object(&x as *const usize), i + 2);
+        let x: Box<usize> = Box::new(i);
+        assert_eq!(kani::mem::pointer_object(&*x as *const usize), 2 * i + 2);
```
zhassan-aws, line 19: What is the reason for the multiplication by 2? Is there an object for the box and another for the variable `x`? If so, would inlining the `Box:new` operation in the assert remove the doubling?

tautschnig:
Closing as this has been made part of #2995.

PR #3381, opened 2024-07-24
Maintain test execution performance with newer CaDiCaL versions

tautschnig:
CBMC using CaDiCaL 1.9.2 and later have substantially worse performance on these tests than CaDiCaL 1.7.2, which CBMC 5.95.1 uses. Using MiniSat will make sure performance remains consistent.


feliperodri (approved):

zhassan-aws:
Interesting. How much is the performance difference?

tautschnig:
> Interesting. How much is the performance difference?

CaDiCaL regressed from a few seconds to several minutes on those examples.

PR #3382, opened 2024-07-24
Better explanations and extra functionality in `MutableBody`

artemagvanian:
`MutableBody` is the core data structure for MIR manipulation in instrumentation passes. However, its public methods have limited documentation and unclear semantics. This slowed down the development of instrumentation passes.

This PR aims to fix that by:
- Clarifying how source instruction shifts when the methods are called;
- Adding functionality for inserting basic blocks;
- Expanding the support for different terminator types.

### DIFF
--- a/kani-compiler/src/kani_middle/transform/body.rs
+++ b/kani-compiler/src/kani_middle/transform/body.rs
@@ -88,16 +88,20 @@ impl MutableBody {
 
     pub fn new_str_operand(&mut self, msg: &str, span: Span) -> Operand {
         let literal = MirConst::from_str(msg);
-        Operand::Constant(ConstOperand { span, user_ty: None, const_: literal })
+        self.new_const_operand(literal, span)
     }
 
-    pub fn new_const_operand(&mut self, val: u128, uint_ty: UintTy, span: Span) -> Operand {
+    pub fn new_uint_operand(&mut self, val: u128, uint_ty: UintTy, span: Span) -> Operand {
         let literal = MirConst::try_from_uint(val, uint_ty).unwrap();
+        self.new_const_operand(literal, span)
+    }
+
+    fn new_const_operand(&mut self, literal: MirConst, span: Span) -> Operand {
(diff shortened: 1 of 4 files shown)


kani-compiler/src/kani_middle/transform/body.rs
```diff
@@ -146,9 +150,10 @@ impl MutableBody {
     /// Add a new assert to the basic block indicated by the given index.
     ///
     /// The new assertion will have the same span as the source instruction, and the basic block
-    /// will be split. The source instruction will be adjusted to point to the first instruction in
-    /// the new basic block.
-    pub fn add_check(
+    /// will be split. If `InsertPosition` is `InsertPosition::Before`, `source` will point to the
+    /// same instruction as before. If `InsertPosition` is `InsertPosition::After`, `source` will
+    /// point to the new terminator.
+    pub fn new_check(
```
celinval, line 156: Can you please explain why `new` makes more sense here? For me, `new` implies that this method is returning a new object, not necessarily it's adding to the body. Another option would be to use `insert_` instead.

I acknowledge that not all methods are consistent with this terminology, but I would prefer fixing that. E.g.: `new_assignment` could become `add_assignment` / `insert_assignment`.

celinval (commented):
It looks good. Thanks for improving documentation.

kani-compiler/src/kani_middle/transform/body.rs
artemagvanian: I am happy to change the names of all methods that "add" constructs to the body to "insert_XXX" to be consistent

kani-compiler/src/kani_middle/transform/body.rs
```diff
@@ -128,7 +132,7 @@ impl MutableBody {
 
     /// Add a new assignment.
     ///
-    /// Return  local where the result is saved.
+    /// Return the local where the result is saved.
```
celinval, line 135: Can you also rename this method please? Thanks

celinval (approved):
Looks good. Thanks

PR #3384, opened 2024-07-25
Add missing functionalities to kani_core

celinval:
- Add Arbitrary for array
  - Add Arbitrary for tuples
  - Add missing changes from modifies slices (#3295)

Note that for adding `any_array` I had to cleanup the unnecessary usage of constant parameters from `kani::any_raw`.

### DIFF
--- a/library/kani/src/lib.rs
+++ b/library/kani/src/lib.rs
@@ -7,11 +7,10 @@
 // Used for rustc_diagnostic_item.
 // Note: We could use a kanitool attribute instead.
 #![feature(rustc_attrs)]
-// This is required for the optimized version of `any_array()`
-#![feature(generic_const_exprs)]
-#![allow(incomplete_features)]
 // Used to model simd.
 #![feature(repr_simd)]
+#![feature(generic_const_exprs)]
+#![allow(incomplete_features)]
 // Features used for tests only.
 #![cfg_attr(test, feature(core_intrinsics, portable_simd))]
 // Required for `rustc_diagnostic_item` and `core_intrinsics`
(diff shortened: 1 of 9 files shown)


library/kani/src/lib.rs
```diff
@@ -7,11 +7,10 @@
 // Used for rustc_diagnostic_item.
 // Note: We could use a kanitool attribute instead.
 #![feature(rustc_attrs)]
-// This is required for the optimized version of `any_array()`
-#![feature(generic_const_exprs)]
-#![allow(incomplete_features)]
 // Used to model simd.
 #![feature(repr_simd)]
+#![feature(generic_const_exprs)]
```
jaisnan, line 12: Is this a clippy cleanup?

jaisnan (approved):
Thanks for the cleanup, this might make unification easier as well. Approving as the added function-contracts in the `verify_core` seem to check for regressions in newly added capabilities as well.

library/kani/src/lib.rs
celinval: No. I updated it because before it was needed for `simd` and any_array. However, now it is only needed for `simd`.

celinval:
> Thanks for the cleanup, this might make unification easier as well. Approving as the added function-contracts in the `verify_core` seem to check for regressions in newly added capabilities as well.

BTW, I noticed that `core` crate contains a re-export of itself as `core`, so they can use `core` inside it. So we might be able to just use straight `core::` for both versions (with and without `core`), so we may not need the `$core` argument to our macros.

PR #3386, opened 2024-07-25
Add two new CI workflows to check Kani against `verify-rust-std`

celinval:
This PR adds the following workflows:
1. `verify-std-check`: This workflow will run either in a PR or via workflow call. It pulls the `main` branch of `verify-rust-std` repository, and see if it can verify the repo with the new Kani version from HEAD.
   - If verification succeeds, the workflow will succeed.
   - If the verification fails and this is a PR, the workflow will compare the verification result against the base of the PR. The workflow will succeed if the results match*.
2. `verify-std-update`: This workflow will run the `verify-std-check` workflow. If it succeeds, it will update the `verify-rust-std` branch from HEAD. This workflow will fail if `features/verify-rust-std` branch diverges, and a merge is required.

The motivation for this PR is to help us identify any changes to Kani that may break the Rust verification repository, and to keep the `verify-rust-std` up to date as much as possible.

* The fallback logic is needed since toolchain upgrades can potentially break the std verification. This will happen in cases where the new toolchain version is incompatible with the library version from the `verify-rust-std`. Those cases should not interfere with Kani development, and they should be fixed when we update the `verify-rust-std` repo.

### Call-out

Here is an example of a `verify-std-check` run: https://github.com/celinval/kani-dev/actions/runs/10101499265.
Here is an example of a `verify-std-update` run: https://github.com/celinval/kani-dev/actions/runs/10101909177.


jaisnan (approved):
I am approving under the assumption that should the workflow to update features/verify-rust-std on its own fail, we will modify it again to not push directly.

Is that a correct assumption?

Issue #2681, opened 2023-08-14
Invalid path reported for Rust toolchain on MacOS


celinval:
I got this error:
```
Errors
File [/Users/ec2-user/.rustup/toolchains/nightly-2023-08-04-aarch64-apple-darwin/lib/rustlib/src/rust/library/std/src/sys/unix/thread_local_dtor.rs]
Function [std::sys::unix::thread_local_dtor::register_dtor]
Line [69]
[[trace] call to foreign "C" function `_tlv_atexit` is not currently supported by Kani. Please post your example at https://github.com/model-checking/kani/issues/2423
```

But I don't understand why it says "/Users/ec2-user/". I don't have this user on my machine! :)

_Originally posted by @ithinkicancode in https://github.com/model-checking/kani/issues/2423#issuecomment-1676484187_

jaisnan:
Replicating this issue in a ubuntu environment, gave this response - 

`../../../github/home/.rustup/toolchains/nightly-2023-09-19-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/builders.rs:33:16 in function <core::fmt::builders::PadAdapter<'_, '_> as std::fmt::Write>::write_str`

What this tells me is that, the path to the toolchain is being hardcoded from where the binary is being built. In this case, in the `github action` runner.

celinval:
I believe this is related to this PR: https://github.com/model-checking/kani/pull/976. More specifically, this change: https://github.com/model-checking/kani/pull/976/files#diff-ad7f3edc0c2f338653896927f634fc2b24a45720db4249bda577be21bb9fabb5

See this comment in the PR:

> 2. Fix the file path in the compiler to use local file path.
>    - Note that using the local file path is not recommended to generate artifacts that can be shared, but I don't this we have that use case today.

One side effect of reverting this change, is that Kani will always display invalid path for toolchain files. This could impact the VSCode extension (if it uses the paths printed by Kani) and it will impact the visualize trace which probably won't display the contents of standard library functions.

If we do want to make sure the path points to the local version of the rust toolchain, we will likely need to build all our libraries locally as part of the setup process.

celinval:
@jaisnan, I wonder if we could use `-Z remap-cwd-prefix` tracked [here](https://github.com/rust-lang/rust/issues/89434) while building our libraries.

fedejinich:
[...]

PR #3387, opened 2024-07-26
Tutorial Sections 2.1 and 2.2 Updates

carolynzech:
Update the tutorial, namely:

- Update the [bounds checking and pointers example](https://model-checking.github.io/kani/tutorial-kinds-of-failure.html#bounds-checking-and-pointers). `cargo test` catches the UB in the current example, so this PR modifies the code slightly so that `cargo test` still misses the UB, as desired.
- Rather than including larger sections on experimental features throughout the tutorial, create a separate experimental features section and include (briefer) references to them in the tutorial.
- The old text recommended debugging by generating a trace with `--visualize`, with a briefer mention of `--concrete-playback`. Since `--visualize` is deprecated, revise the debugging exercises to recommend `--concrete-playback` instead.


docs/src/tutorial-kinds-of-failure.md
```diff
@@ -86,33 +86,23 @@ How does Kani's output change, compared to the unsafe operation?
 <details>
 <summary>Click to see explanation for exercise 1</summary>
 
-Having switched back to the safe indexing operation, Kani reports two failures:
+Having switched back to the safe indexing operation, Kani reports a bounds check failure:
```
carolynzech, line 89: It may be worth replacing this example. When I run `cargo test` with the unsafe code line
`return unsafe { *a.get_unchecked(i % a.len() + 1) };`
it now fails with this helpful error message:
```
running 2 tests
thread 'bounds_check::tests::doesnt_crash' panicked at library/core/src/[panicking.rs:221:5](http://panicking.rs:221:5):
unsafe precondition(s) violated: slice::get_unchecked requires that the index is within the slice
thread caused non-unwinding panic. aborting.
error: test failed, to rerun pass --lib
```
whereas Kani's error message, instead of saying
```
```
as desired, now says:
```
SUMMARY:
 ** 1 of 345 failed (8 unreachable)
Failed Checks: assumption failed
 File: "/Users/jaisnan/.rustup/toolchains/nightly-2024-07-01-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/slice/index.rs", line 273, in <usize as std::slice::SliceIndex<[u32]>>::get_unchecked

VERIFICATION:- FAILED
```
I tried a few other unsafe slice functions and found that `split_at_unchecked` produces a more informative error message from Kani:
```
SUMMARY:
 ** 2 of 343 failed (8 unreachable)
Failed Checks: attempt to compute `unchecked_sub` which would overflow
 File: "/Users/jaisnan/.rustup/toolchains/nightly-2024-07-01-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/slice/mod.rs", line 1985, in core::slice::<impl [u32]>::split_at_unchecked
 File: "src/bounds_check.rs", line 15, in bounds_check::get_wrapped

VERIFICATION:- FAILED
```
but `cargo test` catches the error as well. I've held off on replacing it for now in case we want to try to find an example where `cargo test` fails _and_ Kani produces a good error message. I'm happy to do that, but would benefit from some guidance on where to look.

docs/src/tutorial-kinds-of-failure.md
```diff
@@ -126,13 +116,15 @@ In this case, where we just have a couple of `kani::any` values in our proof har
...
 ```
-Step 36: Function bound_check, File src/bounds_check.rs, Line 37
-let size: usize = kani::any();
-size = 2464ul
-
-Step 39: Function bound_check, File src/bounds_check.rs, Line 39
-let index: usize = kani::any();
-index = 2463ul
+Step 523: Function bound_check, File src/bounds_check.rs, Line 37
+<- kani::any::<usize>
```
carolynzech, line 120: I kept the function/file/line information from the previous example (see #3388)

docs/src/tutorial-kinds-of-failure.md
jaisnan: I see https://github.com/model-checking/kani/issues/2681 show up here as well :(. Might need to prioritize it sooner rather than later.

As per your example, we'd want an example where `cargo test` passes, while Kani fails with a good error message right?

celinval: Given the nice error messages, it might be time to revisit: https://github.com/model-checking/kani/issues/1740

celinval: We might also want to enable `ub_checks` if they are not enabled yet.

docs/src/tutorial-first-steps.md
```diff
@@ -200,17 +200,22 @@ Fortunately, Kani is able to report a coverage metric for each proof harness.
 Try running:
 
 ```
-cargo kani --visualize --harness verify_success
+cargo kani --coverage -Z line-coverage --harness verify_success
```
celinval, line 203: I actually think we should remove this from our tutorial. We shouldn't be teaching unstable features in the first step tutorial.

We should have a separate section for experimental features and just refer to them here.

docs/src/tutorial-kinds-of-failure.md
carolynzech: > If you are replacing what the output message in the tutorial is, can you cut the file path in the summary to only show the relative path or the file name?
>
Yes, will do.
> About your example, we'd want an example where `cargo test` passes, while Kani fails with a good error message right? Not an example where `cargo test` fails?
>
Yes, you're right--I meant that we want to find an example where `cargo test` fails to find a problem, but poor wording on my part.

carolynzech: Resolving -- replaced with new example per @zhassan-aws suggestion.

docs/src/reference/experimental/experimental-features.md
```diff
@@ -0,0 +1,5 @@
+# Experimental Features
+
+We elaborate on some of the more commonly used experimental features in Kani.
```
jaisnan, line 3: Is this file going to appear at the top of the section?

docs/src/reference/attributes.md
```diff
@@ -60,7 +60,7 @@ For example, the class in `Check 1: my_harness.assertion.1` is `assertion`, so t
 
 > **NOTE**: The `#[kani::should_panic]` is only recommended for writing
 > harnesses which complement existing harnesses that don't use the same
-> attribute. In order words, it's only recommended to write *negative harnesses*
+> attribute. In other words, it's only recommended to write *negative harnesses*
```
jaisnan, line 63: Nice catch!

jaisnan (commented):
Almost there! Just a few questions. Have you rendered your changes using mdbook to see how it appears to the user? If you have, does the new file for unstable features appear at the top of section?

jaisnan (approved):
Looks good to me! Congrats :D 

As discussed offline, you should go through the tutorial and give it a glance to check if everything rendered properly. You might also catch some more errors in the documentation, and in that case, create another fix for them.

Thank you @carolynzech !

Issue #3326, opened 2024-07-04
Add an option to turn `requires` clauses into assertions


celinval:
Requested feature: Enable contract checking without stubbing.
Use case: Be able to detect if a function violate the contract of any of its dependencies. This can be very helpful for debbuging verification failures due to API misuse. This is especially important for functions that are only be annotated with safety contract, and for those that are not eligible for stubbing.
Link to relevant documentation (Rust reference, Nomicon, RFC):

celinval:
Actually, should we just insert contracts clauses as assertions so they can be used with playback? When `#[cfg(not(kani))]` we already ignore the contract.

jaisnan:
Branch with progress so far: https://github.com/jaisnan/kani/tree/turn-clause-to-contract

PR #3389, opened 2024-07-26
Enable concrete playback for contract and stubs

### LINKED ISSUE #3383: Enable concrete playback for contracts
Use case: Debugging a failure in a `harness_for_contract` is nearly impossible today. Having concrete playback or something else that prints the initial values would really help!

celinval:
This PR enables concrete playback for contract and stubs. Since these two cases may not actually behave as users expect (it can even have an internal failure), I am adding documentation to the generated test calling that out.

As I was testing this issue, I realized that concrete playback didn't quite work for arrays. So I introduced a new private function `any_raw_array`, which doesn't change the behavior during verification, but allow us to special case it in the concrete playback flow.

Resolves #3383 

## Call-out

I just realized I forgot to update `kani-core`. I'll leave it as ready for review, so I can already get some feedback.


kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -43,9 +44,9 @@ impl KaniSession {
             } else {
                 let mut unit_tests: Vec<UnitTest> = harness_values
                     .iter()
-                    .map(|concrete_vals| {
+                    .map(|(prop, concrete_vals)| {
                         let pretty_name = harness.get_harness_name_unqualified();
-                        format_unit_test(&pretty_name, &concrete_vals)
+                        format_unit_test(&pretty_name, &concrete_vals, gen_test_doc(harness, *prop))
```
celinval, line 49:
```suggestion
                        format_unit_test(&pretty_name, &concrete_vals, gen_test_doc(harness, prop))
```

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -168,6 +169,9 @@ impl KaniSession {
                 writeln!(temp_file, "{line}")?;
                 if curr_line_num == proof_harness_end_line {
                     for unit_test in unit_tests.iter() {
+                        // Write an empty line before the unit test.
+                        writeln!(temp_file, "")?;
```
celinval, line 173:
```suggestion
                        writeln!(temp_file)?;
```

kani_metadata/src/harness.rs
```diff
@@ -52,6 +52,34 @@ pub struct HarnessAttributes {
...
+    /// Function was annotated with `#[kani::proof]`.
+    Proof,
+    /// Function was annotated with `#[kani::proof_for_contract(target_fn)]`.
+    ProofForContract { target_fn: String },
+    /// This is a test harness annotated with `#[test]`.
+    Test,
+}
+
+impl HarnessAttributes {
+    /// Create a new harness with of the provided kind.
```
zhassan-aws, line 66:
```suggestion
    /// Create a new harness of the provided kind.
```

library/kani_core/src/arbitrary.rs
```diff
@@ -34,7 +34,7 @@ macro_rules! generate_arbitrary {
                         unsafe { crate::kani::any_raw_internal::<Self>() }
                     }
                     fn any_array<const MAX_ARRAY_LENGTH: usize>() -> [Self; MAX_ARRAY_LENGTH] {
-                        unsafe { crate::kani::any_raw_internal::<[Self; MAX_ARRAY_LENGTH]>() }
+                        unsafe { crate::kani::any_raw_aray::<[Self; MAX_ARRAY_LENGTH]>() }
```
zhassan-aws, line 37: Should this be:
```suggestion
                        unsafe { crate::kani::any_raw_array::<Self; MAX_ARRAY_LENGTH>() }
```
?

zhassan-aws (commented):
Nice!

Can you clarify how concrete playback for a contract can fail?

Can we also add a test for stubs?

celinval:
> Nice!
> 
> Can you clarify how concrete playback for a contract can fail?

It can fail if your pre-conditions are too loose or if the post-conditions are too strict. I found it very hard to debug contract specification otherwise.

> 
> Can we also add a test for stubs?

Sure! I might just add a generic script to run multiple concrete playback tests. It's a big annoying to write a new script for each case.

zhassan-aws:
> > Nice!
> > Can you clarify how concrete playback for a contract can fail?
> 
> It can fail if your pre-conditions are too loose or if the post-conditions are too strict. I found it very hard to debug contract specification otherwise.

I meant when would concrete playback itself fail? I'm specifically referring to this part of the PR description:
> This PR enables concrete playback for contract and stubs. Since these two cases may not actually behave as users expect (it can even have an internal failure), I am adding documentation to the generated test calling that out.

I can see why it would fail for stubs (hence the documentation that you added makes sense for stubs), but I'm not sure if the same applies to contracts? Or perhaps this is referring to the stubbing of a contract?

celinval:
The contract themselves may have calls to `kani::any()`, which could make the test misbehave. Additionally, on normal executions, contract clauses are not added to the execution flow, so a contract failure won't be triggered. If we implement #3326, then we wouldn't have this problem any more.

zhassan-aws (approved):

tests/script-based-pre/playback_expected/playback.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# This will run Kani verification in every `src/*.rs` file followed by playback command.
+# Expected output is generated from individual expected files.
+set -o nounset
+
+run() {
+  input_rs=${1:?"Missing input file"}
```
feliperodri, line 9:
```suggestion
  local input_rs=${1:?"Missing input file"}
```

tests/script-based-pre/playback_expected/playback.sh
```diff
@@ -0,0 +1,39 @@
...
+
+  echo "[TEST] Generate test for $input_rs..."
+  kani ${input_rs} \
```
feliperodri, line 12:
```suggestion
  echo "[TEST] Generate test for ${input_rs}..."
  kani "${input_rs}" \
```
Quoting variables prevents word splitting and globbing issues.

tests/script-based-pre/playback_expected/playback.sh
```diff
@@ -0,0 +1,39 @@
...
+    -Z concrete-playback --concrete-playback=inplace \
+    -Z function-contracts -Z stubbing --output-format terse
+
+  # Note that today one of the tests will succeed since the contract pre-conditions are not inserted by Kani.
+  # Hopefully this will change with https://github.com/model-checking/kani/issues/3326
+  echo "[TEST] Run test for $input_rs..."
+  summary=$(kani playback -Z concrete-playback ${input_rs} -- kani_concrete_playback | grep "test result")
+  echo "Result for $input_rs: $summary"
```
feliperodri, line 20:
```suggestion
  echo "[TEST] Run test for ${input_rs}..."
  summary=$(kani playback -Z concrete-playback "${input_rs}" -- kani_concrete_playback | grep "test result")
  echo "Result for ${input_rs}: ${summary}"
```

tests/script-based-pre/playback_expected/playback.sh
```diff
@@ -0,0 +1,39 @@
...
+}
+
+ROOT_DIR=$(git rev-parse --show-toplevel)
+MODIFIED_DIR=modified
+rm -rf $MODIFIED_DIR
+mkdir $MODIFIED_DIR
```
feliperodri, line 26:
```suggestion
MODIFIED_DIR="${ROOT_DIR}/modified"
rm -rf "${MODIFIED_DIR}"
mkdir -p "${MODIFIED_DIR}"
```

tests/script-based-pre/playback_expected/playback.sh
```diff
@@ -0,0 +1,39 @@
...
+
+for rs in src/*.rs
+do
+  [[ -e "${rs}" ]] || exit 1
+  echo "Running ${rs}"
+  cp "$rs" $MODIFIED_DIR
+  pushd $MODIFIED_DIR
+  run $(basename $rs)
+  popd
+done
```
feliperodri, line 36:
```suggestion
for rs in src/*.rs; do
  if [[ -e "${rs}" ]]; then
    echo "Running ${rs}"
    cp "${rs}" "${MODIFIED_DIR}"
    pushd "${MODIFIED_DIR}" > /dev/null
    run "$(basename "${rs}")"
    popd > /dev/null
  else
    echo "No .rs files found in src directory"
    exit 1
  fi
done
```
Error handling, just in case.

tests/script-based-pre/playback_expected/playback.sh
```diff
@@ -0,0 +1,39 @@
...
+
+  # Cleanup
+rm -rf $MODIFIED_DIR
```
feliperodri, line 39:
```suggestion
  # Cleanup
rm -rf "${MODIFIED_DIR}"
```

kani_metadata/src/harness.rs
```diff
@@ -52,6 +52,34 @@ pub struct HarnessAttributes {
...
+            kind,
+            should_panic: false,
+            solver: None,
+            unwind_value: None,
+            stubs: vec![],
+        }
+    }
+
+    /// Return whether this is a proof harness.
+    pub fn is_proof(&self) -> bool {
```
feliperodri, line 78: I'd keep consistent with `attributes.rs` and just call it `is_proof_harness`.

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -231,8 +235,51 @@ impl KaniSession {
...
+    let mut doc_str = match &harness.attributes.kind {
+        HarnessKind::Proof => {
+            format!("/// Test generated for harness `{}` \n", harness.pretty_name)
+        }
+        HarnessKind::ProofForContract { target_fn } => {
+            format!(
+                "/// Test generated for harness `{}` that checks contract for `{target_fn}`\n",
+                harness.pretty_name
+            )
+        }
```
feliperodri, line 248:
```suggestion
        HarnessKind::ProofForContract { target_fn } => {
            format!(
                "/// Test generated for harness `{}` that checks contract for `{}`\n",
                harness.pretty_name, target_fn
            )
        }
```

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -231,8 +235,51 @@ impl KaniSession {
...
+               ///\n\
+               /// Concrete playback tests combined with stubs or contracts is highly \n\
+               /// experimental, and subject to change.\n\
+               ///\n\
+               /// The original harness has stubs which are not applied to this test.\n\
+               /// This may cause a mismatch of non-deterministic values if the stub\n\
+               /// creates any non-deterministic value.\n\
+               /// The execution path may also differ, which can be used to refine the stub\n\
+               /// logic.",
+        );
```
feliperodri, line 272:
```suggestion
doc_str.push_str(r#"/// 
/// # Warning
/// 
/// Concrete playback tests combined with stubs or contracts is highly 
/// experimental, and subject to change.
/// 
/// The original harness has stubs which are not applied to this test.
/// This may cause a mismatch of non-deterministic values if the stub
/// creates any non-deterministic value.
/// The execution path may also differ, which can be used to refine the stub
/// logic.
"#);
```

kani-driver/src/concrete_playback/test_generator.rs
```diff
@@ -231,8 +235,51 @@ impl KaniSession {
...
+            format!(
+                "/// Test generated for harness `{}` that checks contract for `{target_fn}`\n",
+                harness.pretty_name
+            )
+        }
+        HarnessKind::Test => {
+            unreachable!("Concrete playback for tests is not supported")
+        }
+    };
+    doc_str.push_str("///\n");
```
feliperodri, line 253: Question: wouldn't be more efficient to use the `write!` macro? Do we even care? Can we generate multiple concrete tests or this will always just run once?

tests/script-based-pre/playback_expected/playback.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+# This will run Kani verification in every `src/*.rs` file followed by playback command.
+# Expected output is generated from individual expected files.
+set -o nounset
```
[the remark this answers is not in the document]
celinval: `-u` and `nounset` is the same thing. Since this script is running multiple tests and it's output is compared against the expected file, I prefer not failing if any command fails so it always runs all tests

kani-driver/src/concrete_playback/test_generator.rs
celinval: I don't know if it's really faster, and I don't think we care. We run this once per test case that we generate.

tests/script-based-pre/playback_expected/playback.sh
celinval: We don't want to add the modified folder to our repository root folder

kani-driver/src/concrete_playback/test_generator.rs
celinval: ?

tests/script-based-pre/playback_expected/playback.sh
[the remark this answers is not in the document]
celinval: Note that Kani will return non-zero if the verification fails.

PR #3392, opened 2024-07-29
Update the rust toolchain to nightly-2024-07-29

tautschnig:
Changes required due to:
- rust-lang/rust@9db265048e Bump to 1.82

Resolves: #3369

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #3394, opened 2024-07-29
Fix verify-std-check outcome check

celinval:
It looks like I used `output` instead of `outcome` variable.

I noticed that [this PR](https://github.com/model-checking/kani/pull/3392) should've failed the workflow but it didn't because of this mistake.

### DIFF
--- a/.github/workflows/verify-std-check.yml
+++ b/.github/workflows/verify-std-check.yml
@@ -39,6 +39,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           path: kani
+          fetch-depth: 0
 
       - name: Setup Kani Dependencies
         uses: ./kani/.github/actions/setup
@@ -79,7 +80,7 @@ jobs:
             -Z mem-predicates -Z ptr-to-ref-cast-checks
 
       - name: Compare PR results
-        if: steps.check-head.outcome != 'success' && steps.check-head.output != github.check-base.output
+        if: steps.check-head.outcome != 'success' && steps.check-head.outcome != github.check-base.outcome
         run: |
           echo "New failure introduced by this change"
           exit 1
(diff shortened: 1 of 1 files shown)


jaisnan (approved):
Oops, should have caught this in my review. in any case, thanks!  

Can we add a comment somewhere that all toolchain upgrade failures that are resolved manually should trigger a failure for this CI check?

celinval:
We currently have the following blob at the top of the workflow file:
```yml
# We expect toolchain updates to potentially break this workflow in cases where
# the version tracked in the `verify-rust-std` is not compatible with newer
# versions of the Rust toolchain.
#
# Changes unrelated to the toolchain should match the current status of main.
```
Do you have anything else in mind?

jaisnan:
> We currently have the following blob at the top of the workflow file:
> 
> ```yaml
> # We expect toolchain updates to potentially break this workflow in cases where
> # the version tracked in the `verify-rust-std` is not compatible with newer
> # versions of the Rust toolchain.
> #
> # Changes unrelated to the toolchain should match the current status of main.
> ```
> 
> Do you have anything else in mind?

Just maybe a small clarification like ->

```
# We expect toolchain updates to potentially break this workflow in cases where
# the version tracked in the `verify-rust-std` is not compatible with newer 
# versions of the Rust toolchain. 
# For example, toolchain upgrade breaks such as 
# https://github.com/model-checking/kani/issues/3369.
#
# Changes unrelated to the toolchain should match the current status of main.
```

It's okay if this change is not added, this is purely for me because i like examples in comments.

celinval:
I agree that an example can be very helpful, but I'm not sure the PR you mentioned is a good reference. Maybe we can add a better link later.

The PR didn't actually fail the workflow, and the workflow logs themselves won't be available for more than a few days. So it might be confusing to use this as a reference.

PR #3398, opened 2024-07-31
Reduce CBMC verbosity to CBMC's default

tautschnig:
CBMC v6 included changes to hide timing information and loop unwinding status output at the default verbosity. To not impact Kani users, #2995 included changes to make Kani run CBMC with `--verbosity 9`. This change now makes Kani run CBMC with just the default verbosity, hiding timing information and loop unwinding status by default. If users still wish to see this information, they can invoke Kani with
`--cbmc args --verbosity 9`.

### DIFF
--- a/kani-driver/src/call_cbmc.rs
+++ b/kani-driver/src/call_cbmc.rs
@@ -153,11 +153,6 @@ impl KaniSession {
 
         args.push(file.to_owned().into_os_string());
 
-        // Make CBMC verbose by default to tell users about unwinding progress. This should be
-        // reviewed as CBMC's verbosity defaults evolve.
-        args.push("--verbosity".into());
-        args.push("9".into());
-
         Ok(args)
     }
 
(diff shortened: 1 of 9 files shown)


zhassan-aws (changes requested):
I would suggest packaging this in a Kani-level option (e.g. `--verbose_solver`). The user should not be required to use `--cbmc-args` except in very unusual cases, and I think it will be common for users to need this option to debug a run that is not terminating.

We also need to highlight this change in the CHANGELOG.md.

tautschnig:
> I would suggest packaging this in a Kani-level option (e.g. `--verbose_solver`). The user should not be required to use `--cbmc-args` except in very unusual cases, and I think it will be common for users to need this option to debug a run that is not terminating (e.g. to understand whether it's stuck in unwinding, symex, or solver stage).
> 
> We also need to highlight this change in the CHANGELOG.md.

Marking this PR as "Draft," which I actually meant to do from the start: we should really review the verbosity levels on the CBMC side as at least some of the information that's now hidden at the default verbosity (such as unwinding information) is practically very important. (Timing information, however, is probably not very important.)

I suggest we revisit this PR once we have a better situation on the CBMC side, at which point it may well be possible to merge this one as-is.

PR #3405, opened 2024-08-01
Update crates documentation with info. on `#[safety_constraint(...)]` attribute for structs

adpaco-aws:
This PR updates the crates documentation we already had for the `#[safety_constraint(...)]` attribute to account for the changes in #3270.

The proposal includes notes/guidelines on where to use the attribute (i.e., the struct or its fields) depending on the type safety condition to be specified. Also, it removes the `rs` annotations on the code blocks that appear in the documentation because they don't seem to have the intended effect (the blocks are not being highlighted at all).

The current version of this documentation can be found [here](https://model-checking.github.io/kani/crates/doc/kani/derive.Arbitrary.html) and [here](https://model-checking.github.io/kani/crates/doc/kani/derive.Invariant.html).

Related #3095 


library/kani_macros/src/lib.rs
```diff
@@ -158,6 +161,45 @@ pub fn unstable_feature(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// #[derive(Arbitrary)]
+/// #[safety_constraint(vector.len() <= *capacity)]
+/// struct MyVector<T> {
+///     vector: Vec<T>,
+///     capacity: usize,
+/// }
+/// ```
+///
+/// However, if we were defining a struct whose fields are not related in any
+/// way, we would prefer using the `#[safety_constraint(...)]` attribute in its
```
zhassan-aws, line 191:
```suggestion
/// way, we would prefer using the `#[safety_constraint(...)]` attribute on its
```

library/kani_macros/src/lib.rs
```diff
@@ -212,6 +258,45 @@ pub fn derive_arbitrary(item: TokenStream) -> TokenStream {
 ///
 /// Note: the assignments to `obj` and `inner` are made so that we can treat the
 /// fields as if they were references.
+///
+/// ### Adding `#[safety_constraint(...)]` to the struct as opposed to its fields
+///
+/// As mentioned earlier, the `#[safety_constraint(...)]` attribute can be added
+/// to to either the struct or its fields, but not to both. Adding the
```
zhassan-aws, line 265:
```suggestion
/// to either the struct or its fields, but not to both. Adding the
```

library/kani_macros/src/lib.rs
```diff
@@ -212,6 +258,45 @@ pub fn derive_arbitrary(item: TokenStream) -> TokenStream {
...
+/// #[derive(Invariant)]
+/// #[safety_constraint(vector.len() <= *capacity)]
+/// struct MyVector<T> {
+///     vector: Vec<T>,
+///     capacity: usize,
+/// }
+/// ```
+///
+/// However, if we were defining a struct whose fields are not related in any
+/// way, we would prefer using the `#[safety_constraint(...)]` attribute in its
```
zhassan-aws, line 288:
```suggestion
/// way, we would prefer using the `#[safety_constraint(...)]` attribute on its
```

zhassan-aws (commented):
Thanks!

Only a few nits.

library/kani_macros/src/lib.rs
```diff
@@ -103,16 +103,19 @@ pub fn unstable_feature(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// or its fields (but not both) to indicate a type safety invariant condition
+/// `<cond>`. Since `kani::any()` is always expected to produce type-safe
+/// values, **adding `#[safety_constraint(...)]` to the struct or any of its
+/// fields will further constrain the objects generated with `kani::any()`**.
 ///
 /// For example, the `check_positive` harness in this code is expected to
 /// pass:
 ///
-/// ```rs
+/// ```
```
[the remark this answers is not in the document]
celinval: I think by default mdbook considers code blocks to be Rust.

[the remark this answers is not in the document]
adpaco-aws: `rs` is a valid alias in most highlighters (e.g., [this one](https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md) or [this one](https://pygments.org/docs/lexers/)) and what I normally use for my comments here. But it's possible that's not the case here. Let me use `rust` and if it doesn't work I'll just remove it.

zhassan-aws (approved):

PR #3406, opened 2024-08-01
Stacked Borrows in Kani

jsalzbergedu:
This PR adds stacked borrows to Kani, strengthening its model of pointers to detect violations of Rust's aliasing model.

It does this by automatically instrumenting code with hooks at every stack and heap allocation, ensuring that a live mutable pointer is never aliased.

This instrumentation is subject to several limitations:

1. Currently, function calls are not permitted. To implement them, the "protector" mechanism of the stacked borrows calculus has to be implemented first.
2. Only one byte is monitored; it is picked nondeterministic-ally at the beginning of the instrumented code. Function contracts may need to keep track of relations between multiple references and raw pointers in their pre & post conditions, and are therefore not yet tracked.
3. The address of locals is taken before they are initialized; because of this, optimizations to the "local storage" part of the stack frame may break this analysis. In our cases so far, no such optimizations are performed. It also means that initialization must be checked separately using the Uninit pass
4. Arrays are not yet modelled
5. Malloc is not yet modelled
7. Pointer manipulation is not yet modelled
8. Two phase borrows are not yet modelled.

Two phase borrows:
The rust compiler will desugar the following:
```rust
fn p(mut x: Vec<usize>) { x.push(x.len()) } 
```
to
```rust
fn p(mut x: Vec<usize>) {
  let y: &mut Vec<i32> = &mut x;
  let z: usize = Vec::len(x);
  Vec::push(y, z)
}
```
Marking the borrow into y as "two phase." 

This may be enabled in the future via a hack (ref: https://www.ralfj.de/blog/2023/06/02/tree-borrows.html) and so the aliasing model may in the future change to accommodate these differently.

10. There are numerous unhandled syntactic constructs.
- Assignments that Copy/Move
- Terminators
- Fields
- Arrays/Slices/Indices
These constructs _must_ be handled in the future to say that we have ported stacked borrows to Kani.
Currently, print statements are made; these constructs are expected to do pointer manipulation, but to no effect.
In the future, we will want to instrument all of these.


kani-compiler/src/kani_middle/transform/check_aliasing/cached_body_mutator.rs
```diff
@@ -0,0 +1,94 @@
+use std::collections::HashMap;
+use super::{BodyMutator, Local, MirInstance, Mutability, LocalDecl, Body, Ty, Statement, MutatorIndex, MutatorIndexStatus, Instruction, Span};
+
+/// Body mutator which wraps the BodyMutator
+/// interface with a cache of the locals that
+/// store function calls.
+pub struct CachedBodyMutator {
```
artemagvanian, line 7: It seems to me that the only functionality this struct introduces is caching function locals to use in instrumented calls and providing a unit local to use as a return target. I think those could be just incorporated into the transformation pass.

kani-compiler/src/kani_middle/transform/check_aliasing/body_mutator.rs
```diff
@@ -0,0 +1,258 @@
...
+    var_debug_info: Vec<VarDebugInfo>,
+    spread_arg: Option<Local>,
+    span: Span,
+
+    ghost_locals: Vec<LocalDecl>,
+    ghost_blocks: Vec<BasicBlock>,
+    ghost_statements: Vec<Statement>,
+}
+
+impl BodyMutator {
```
artemagvanian, line 24: This has the same purpose and functionality as `MutableBody`. I think the instrumentation needs to integrate with the existing framework instead.

kani-compiler/src/kani_middle/transform/check_aliasing/body_mutation.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub fn instrument_locals(&mut self) -> Result<(), MirError> {
+        self.instrumentation_data.instrument_locals(&self.values)
+    }
+
+    pub fn instrument_instructions(&mut self) -> Result<(), MirError> {
+        self.instrumentation_data.instrument_instructions()?;
+        Ok(())
+    }
+
+    pub fn finalize(mut self) -> Body {
```
artemagvanian, line 22: This is the only method that adds extra functionality. I would rather have it as a single helper function instead of creating a separate struct for this.

kani-compiler/src/kani_middle/transform/check_aliasing/function_cache.rs
```diff
@@ -0,0 +1,66 @@
+pub use stable_mir::ty::GenericArgKind as GenericArg;
+pub use stable_mir::ty::GenericArgs;
+use super::{MirInstance, MirError, TyCtxt, super::super::find_fn_def};
+/// FunctionSignature encapsulates the data
+/// for rust functions with generic arguments
+/// to ensure that it can be cached.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Signature {
```
artemagvanian, line 8: Is there any reason why you wouldn't use `FnDef`?

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
```diff
@@ -0,0 +1,300 @@
+use std::collections::HashMap;
+use super::{TyCtxt, CachedBodyMutator, Cache, Local, Place, Rvalue, Mutability, StatementKind, Statement, MirError, Signature, Span, Ty, GenericArg, MutatorIndex, Instruction, BorrowKind, ProjectionElem, TyKind, RigidTy, MutatorIndexStatus};
```
artemagvanian, line 2: Perhaps split those imports to indicate better which one of those are from `stable_mir` and which one are new.

kani-compiler/src/kani_middle/transform/check_aliasing/function_cache.rs
```diff
@@ -0,0 +1,66 @@
+pub use stable_mir::ty::GenericArgKind as GenericArg;
```
artemagvanian, line 1: Renaming of commonly used structures makes it harder to track what's going on. I'd rather avoid that.

kani-compiler/src/kani_middle/transform/check_aliasing/local_collection.rs
```diff
@@ -0,0 +1,64 @@
...
+    Body, BodyMutationPassState, CachedBodyMutator, InstrumentationData, Local, LocalDecl,
+    MirVisitor, TyCtxt,
+};
+use std::collections::HashMap;
+
+/// Collect local visitor visits the body
+/// and collects all of the (non argument)
+/// locals. In the future it will collect
+/// argument locals and initialize arguments.
+struct CollectLocalVisitor {
```
artemagvanian, line 12: Similarly to the previous comments, I don't think you need a new struct for that.

kani-compiler/src/kani_middle/transform/check_aliasing/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+    "std::panic",
+    "core::panic",
+    "std::fmt",
+    "std::iter",
+    "core::ub_checks",
+    "std::cmp",
+    "core::slice",
+    "std::mem",
+    // This blacklist needs expansion.
+];
```
artemagvanian, line 103: I am not a huge fan of that. Right now, you are skipping instrumenting _all_ instances of these functions. What if some are legitimately called from the code you are verifying?

kani-compiler/src/kani_middle/transform/check_aliasing/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+    {
+        let args = query_db.args();
+        args.ub_check.contains(&ExtraChecks::Aliasing)
+    }
+
+    fn transform(&mut self, tcx: TyCtxt, body: Body, instance: MirInstance) -> (bool, Body) {
+        trace!(function=?instance.name(), "transform: aliasing pass");
+        if ALIASING_BLACKLIST
+            .iter()
+            .fold(false, |blacklisted, member| blacklisted || instance.name().contains(member))
```
artemagvanian, line 134: Regardless of the previous comment, this is imprecise and could cause troubles later on. I think using diagnostic items would be a better way to go.

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Associate every pointer object with a tag
+    static mut TAGS: ShadowMem<PointerTag> = ShadowMem::new(0);
+    /// Next pointer id: the next pointer id in sequence
+    static mut NEXT_TAG: PointerTag = 0;
+
+    #[non_exhaustive]
+    struct Access;
+    #[allow(unused)]
+    impl Access {
+        pub(self) const READ: bool = false;
```
artemagvanian, line 21: `pub(self)` is a no-op AFAIK, why would you need them?

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,254 @@
+const STACK_DEPTH: usize = 15;
+type PointerTag = u8;
+
+use crate::mem::pointer_object;
+use crate::mem::pointer_offset;
+use crate::shadow::ShadowMem;
+use crate::{any, assume};
+
+/// The stacked borrows state.
+pub mod sstate {
```
artemagvanian, line 10: As a person who is not very familiar with stacked borrows, I would appreciate better module-level documentation, explaining what are those things for.

kani-compiler/src/kani_middle/transform/check_aliasing/body_mutator.rs
artemagvanian: Also, I am not too confident this handles control flow or multiple basic blocks correctly. We need more tests for that.

artemagvanian (commented):
Thanks! Having Stacked Borrows instrumentation in Kani is very cool! 

It seems like the code uses a version of body mutation that is quite similar to `MutableBody` that we already have. I think integrating with current instrumentation (and improving it) is better than creating a different version of it. Moreover, I am not sure the instrumentation introduced in this PR actually works for any non-trivial CFG and there are no tests to convince me of the opposite. 

I also added several small comments here and there.

kani-compiler/src/kani_middle/transform/check_aliasing/function_cache.rs
jsalzbergedu: I could use FnDef instead in the signature, however, I'm trying to cache different generic versions of the same function.

jsalzbergedu: I believe that FnDef doesn't encapsulate the generic arguments.

artemagvanian: Perhaps a combination of FnDef + GenericArgs would work?

artemagvanian: Alternatively, you could rename `name` to `diagnostic` to make it clear that this is the name of the diagnostic item

celinval (commented):
Thanks @jsalzbergedu! I haven't looked at the instrumentation code yet. I was mostly trying to understand the library changes.

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,316 @@
...
+                    STACK_TOP = new_top;
+                    STACK_VALID = STACK_VALID && found;
+                }
+            }
+        }
+    }
+
+    #[rustc_diagnostic_item = "KaniInitializeSState"]
+    pub fn initialize() {
+        self::monitors::prepare_monitors();
```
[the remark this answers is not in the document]
jsalzbergedu: This was meant to prepare the way for future rewrites of the passes in which the prepare monitors might be more involved (non-deterministically pick the # of monitors for example)

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,316 @@
...
+    ///
+    /// Every function call in the source program stack-allocates
+    /// the local variables that it uses; references to these
+    /// variables are only valid after these variables are initialized (first written).
+    /// Therefore this function can be used by supplying an initialized flag
+    /// set to true after the first write, a track flag set to the value
+    /// of a query to a demonic nondeterminism oracle (when this feature is used)
+    /// and a reference to the stack location.
+    #[rustc_diagnostic_item = "KaniInitializeLocal"]
+    pub fn initialize_local<U>(pointer: *const U) {
```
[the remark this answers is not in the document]
jsalzbergedu: I might be able to -- are you saying that we should do this to avoid instrumenting too many variables in the case that there are many variables but not many references or allocations? 

In the regression tests that I've written, this hasn't been a "rate-limiting factor" -- there are more references then there are "value" locals, and the cost of initializing a local is low compared to the cost of a stack check.

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        /// Initialize local when track local is true, picking a monitor,
+        /// and setting its object and offset to within pointer.
+        pub(super) unsafe fn track_local<U>(tag: u8, pointer: *const U) {
+            // Decide whether to initialize the stacks
+            // for location:location+size_of(U).
+            // Offset has already been picked earlier.
+            unsafe {
+                if demonic_nondet() && STATE == MonitorState::UNINIT {
```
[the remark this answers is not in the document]
jsalzbergedu: I do not, I can remove this. It was only there for separation of concerns; I wanted it to be possible to initialize up to n monitors eventually at the beginning of the program so that relational info for up to n pointers could be tracked.

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub(super) mod monitors {
+        static mut STATE: bool = false;
+        static mut OBJECT: usize = 0;
+        static mut OFFSET: usize = 0;
+        static mut STACK_TAGS: [u8; STACK_DEPTH] = [0; STACK_DEPTH];
+        static mut STACK_PERMS: [u8; STACK_DEPTH] = [0; STACK_DEPTH];
+        static mut STACK_TOP: usize = 0;
+
+        #[non_exhaustive]
+        struct MonitorState;
```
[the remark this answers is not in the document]
jsalzbergedu: This was to make sure I don't lose efficiency, and to allow extending to edge cases of the stacked borrows calculus. If i switch from associated constants to enums, and add #[repr(u8)], will it be "as cheap" in symex?

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,316 @@
+const STACK_DEPTH: usize = 15;
+type PointerTag = u8;
+
+use crate::mem::{pointer_object, pointer_offset};
+use crate::shadow::ShadowMem;
+
+/// The stacked borrows state.
+///
+/// The stacked borrows state associates every pointer value
```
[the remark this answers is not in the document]
jsalzbergedu: Stacked borrows should be a stricter calculus than pointer provenance, disallowing not only out-of-bounds accesses but also aliased live mutable pointers. My understanding is that it's closer to an implementation of lexical lifetimes for unsafe code than it is to pointer provenance. Accesses through a pointer object are however restricted to be within the size of the allocation, and a "root object" is kept for each allocation, from which all other pointers to the memory are "derived" -- which I think captures provenance

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,316 @@
...
+        }
+    }
+
+    #[rustc_diagnostic_item = "KaniStackCheckPtr"]
+    pub fn stack_check_ptr<U>(pointer_value: *const *mut U) {
+        unsafe {
+            let tag = TAGS.get(pointer_value);
+            let perm = PERMS.get(pointer_value);
+            let pointer = *pointer_value;
+            for i in 0..std::mem::size_of::<U>() {
```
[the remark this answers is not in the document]
jsalzbergedu: This only exists for separation of concerns

library/kani/src/aliasing.rs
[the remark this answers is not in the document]
celinval: Yes, it should.

library/kani/src/aliasing.rs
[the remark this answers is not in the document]
celinval: You could add this up later when we implement the `n` pointers logic

library/kani/src/aliasing.rs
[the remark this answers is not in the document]
celinval: I would keep it simple for now and add it later when we actually need this. Thanks for the explanation

jsalzbergedu:
@celinval @artemagvanian ,
I believe I've added all of the changes that are required for this pr -- Celina mentioned that some of the changes could be implemented in a follow up PR.

Can you see if this PR is now in a "approvable" state to your standards?

Thanks,

Jacob.

kani-compiler/src/kani_middle/transform/check_aliasing/actions.rs
```diff
@@ -0,0 +1,237 @@
...
+                self.actions.push(Action::NewStackReference { lvalue, rvalue });
+            }
+            [ProjectionElem::Deref] => {
+                // Reborrow
+                // x : &mut T = &*(y : *mut T OR &mut T)
+                let lvalue = to; // Copy to avoid borrow
+                let rvalue = from.local; // Copy to avoid borrow
+                self.visit_assign_reference_dereference(lvalue, rvalue);
+            }
+            _ => { /* not yet handled */ }
```
artemagvanian, line 74: Perhaps we should somehow notify the user that this has not been handled yet?

kani-compiler/src/kani_middle/transform/check_aliasing/actions.rs
```diff
@@ -0,0 +1,237 @@
...
+            }
+            [ProjectionElem::Deref] => {
+                // Dereference -- instrument stack check
+            }
+            _ => {
+                // Field access -- not yet handled.
+                return;
+            }
+        };
+        match self.locals[place.local].ty.kind() {
```
artemagvanian, line 123: You could use `place.ty(...)` instead.

kani-compiler/src/kani_middle/transform/check_aliasing/actions.rs
```diff
@@ -0,0 +1,237 @@
...
+                        // Do nothing for the constants case
+                    }
+                    Rvalue::BinaryOp(_, _, _) => {
+                        eprintln!("Binary op not yet handled");
+                    }
+                    Rvalue::CheckedBinaryOp(_, _, _) => {
+                        eprintln!("Checked binary op not yet handled");
+                    }
+                    _ => {
+                        panic!("Rvalue kind: {:?} not yet handled", rvalue);
```
artemagvanian, line 217: Is there any reason why this is a panic, unlike other variants?

kani-compiler/src/kani_middle/transform/check_aliasing/function_cache.rs
```diff
@@ -0,0 +1,90 @@
...
+        for i in 0..cache.len() {
+            if sig == cache[i].signature {
+                return Ok(&cache[i].instance);
+            }
+        }
+        let fndef = find_fn_def(*ctx, &sig.diagnostic)
+            .ok_or(MirError::new(format!("Not found: {}", &sig.diagnostic)))?;
+        let instance = super::MirInstance::resolve(fndef, &GenericArgs(sig.args.clone()))?;
+        cache.push(Instance::new(sig, instance));
+        Ok(&cache[cache.len() - 1].instance)
```
artemagvanian, line 88: Could you call `self.register(ctx, Signature { diagnostic: "KaniAssert", args: vec![] })` instead?

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
```diff
@@ -0,0 +1,446 @@
...
+        let valid = body.new_local(
+            Ty::from_rigid_kind(stable_mir::ty::RigidTy::Bool),
+            span,
+            Mutability::Mut,
+        );
+        let bb = body.blocks().len() - 1;
+        let min_processed = match body.blocks()[bb].statements.len() {
+            0 => SourceInstruction::Terminator { bb },
+            n => SourceInstruction::Statement { idx: n - 1, bb },
+        };
```
artemagvanian, line 98: Perhaps add a comment explaining what this is responsible for?

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
```diff
@@ -0,0 +1,446 @@
...
+            min_processed,
+            ghost_index,
+            body,
+        }
+    }
+
+    /// Register the function described by the diagnostic
+    /// and generic arguments in "Signature".
+    fn register_fn(&mut self, callee: Signature) -> Result<Local> {
+        let cache = &mut self.cache;
```
artemagvanian, line 118: You don't have to introduce extra variables: e.g.: calling `self.cache.register(...)` is good, too

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
```diff
@@ -0,0 +1,446 @@
...
+                body.replace_terminator(&enter_ghost_block, terminator);
+                // Instrument execute ghost:
+                let target = execute_terminator_block.bb();
+                let kind = TerminatorKind::Goto { target };
+                let terminator = Terminator { kind, span };
+                body.replace_terminator(&execute_ghost_block, terminator);
+                // Instrument execute terminator
+                body.replace_terminator(&execute_terminator_block, original);
+
+                self.ghost_index = execute_ghost_block;
```
artemagvanian, line 235: It seems like what you are trying to do here is to insert a bunch of statements *before* the terminator in the basic block, so perhaps you should do it directly.

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
```diff
@@ -0,0 +1,446 @@
...
+                self.process_instruction();
+            }
+            for action in actions {
+                eprintln!("Action is: {:?}", action);
+                self.instrument_action(action)?;
+            }
+            self.min_processed = match self.min_processed {
+                SourceInstruction::Statement { idx: 0, bb: 0 } => {
+                    break;
+                }
```
artemagvanian, line 416: Perhaps a better way to express this would be writing it out as something like `SourceInstruction::Statement { idx: 0, bb: 0 } | SourceInstruction::Terminator { bb: 0 } => { break; }`

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
```diff
@@ -0,0 +1,446 @@
...
+    valid: Local,
+    /// A map associating resolved generic functions with
+    /// locals in the body that can be used to call them
+    fn_pointers: HashMap<MirInstance, Local>,
+    /// The span of the body
+    span: Span,
+    /// The minimum processed instruction.
+    /// All instructions before this one belong to the original
+    /// source code, and have not yet been analyzed.
+    min_processed: SourceInstruction,
```
artemagvanian, line 49: Totally up to you, but I would call it `current` or something of this sort

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
```diff
@@ -0,0 +1,446 @@
...
+            vec![local_ptr],
+            self.unit,
+        )?;
+        Ok(())
+    }
+
+    /// Split at the minimum processed instruction,
+    /// allowing instrumentation of ghost code following
+    /// that source instruction.
+    pub fn process_instruction(&mut self) {
```
artemagvanian, line 195: I believe this function could use some of the functionality of `insert_bb` from `MutableBody`.

kani-compiler/src/kani_middle/transform/check_aliasing/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+        transformer: &mut super::BodyTransformation,
+    ) {
+        let mut found = HashSet::new();
+        // Collect
+        for instance in &instances {
+            if instance
+                .def
+                .all_attrs()
+                .into_iter()
+                .fold(false, |is_proof, attr| is_proof || attr.as_str().contains("kanitool::proof"))
```
artemagvanian, line 128: Should it also look for `proof_for_contract`?

kani-compiler/src/kani_middle/transform/check_aliasing/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+            {
+                let (items, _) =
+                    collect_reachable_items(tcx, transformer, &[MonoItem::Fn(*instance)]);
```
artemagvanian, line 131: I don't think re-running reachability is a good idea. I think you could extract all reachable items from the call graph instead.

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
```diff
@@ -0,0 +1,446 @@
...
+    pub fn instrument_stack_check(&mut self) -> Result<()> {
+        let span = match self.min_processed {
+            SourceInstruction::Statement { idx, bb } => self.body.blocks()[bb].statements[idx].span,
+            SourceInstruction::Terminator { bb } => self.body.blocks()[bb].terminator.span,
+        };
+        self.call(Signature::new("KaniStackValid", &[]), vec![], self.valid)?;
+        let msg = format!(
+            "Stacked borrows aliasing model violated at {:?}:{:?}",
+            span.get_filename(),
+            span.get_lines()
```
artemagvanian, line 278: I might be wrong, but doesn't the failed assertion contain the location by default?

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
```diff
@@ -0,0 +1,446 @@
...
+            self.instrument_local(local)?
+        }
+        Ok(())
+    }
+
+    /// Fetch the actions to be instrumented at the current instruction.
+    pub fn instruction_actions(&self) -> Vec<Action> {
+        let mut visitor = CollectActions::new(self.body.locals());
+        match self.min_processed {
+            SourceInstruction::Terminator { .. } => { /* not yet handled */ }
```
artemagvanian, line 370: Notify the user about it?

artemagvanian (commented):
Thanks! I added some comments and suggestions for the instrumentation pass itself.

kani-compiler/src/kani_middle/transform/check_aliasing/actions.rs
jsalzbergedu: This is the fall-through case; the other variants occur in the "control_flow.rs" file, but don't alter the stacked borrows state.

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
jsalzbergedu: Artem thank you, what about this needs extra clarification? It is the min_processed field of the struct...

kani-compiler/src/kani_middle/transform/check_aliasing/mod.rs
jsalzbergedu: function contracts are not yet supported

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
jsalzbergedu: This is more general than insert_bb, even in just the statement case -- insert_bb grabs mut target ref, which  may only be one with a target. This is instead just splitting the basic block -- not inserting a basic block after another basic block

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
artemagvanian: For example, I don't quite understand why the terminator is getting skipped if there is one or more statements present.

kani-compiler/src/kani_middle/transform/check_aliasing/instrumentation.rs
jsalzbergedu: uh if you have bb0 you can still have instructions to process

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,316 @@
...
+        const WRITE: bool = true;
+    }
+
+    #[non_exhaustive]
+    struct Permission;
+    impl Permission {
+        const UNIQUE: u8 = 0;
+        const SHAREDRW: u8 = 1;
+        const SHAREDRO: u8 = 2;
+        const DISABLED: u8 = 3;
```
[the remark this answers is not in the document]
jsalzbergedu: Enum for any of the data (Permission, Monitor, Access) causes a perf regression when I tested, doubling the runtime of my test cases (100s for control_flow.rs to 200s).

This is even with`#[repr(u8)]`.

library/kani/src/aliasing.rs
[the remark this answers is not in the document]
jsalzbergedu: Copying the comment of above:


Enum for any of the data (Permission, Monitor, Access) causes a perf regression when I tested, doubling the runtime of my test cases (100s for control_flow.rs to 200s).

This is even with#[repr(u8)].

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,316 @@
...
+            // for location:location+size_of(U).
+            // Offset has already been picked earlier.
+            unsafe {
+                if demonic_nondet() && STATE == MonitorState::UNINIT {
+                    STATE = MonitorState::INIT;
+                    OBJECT = pointer_object(pointer);
+                    OFFSET = 0;
+                    crate::assume(OFFSET < std::mem::size_of::<U>());
+                    STACK_TAGS[STACK_TOP] = tag;
+                    STACK_PERMS[STACK_TOP] = Permission::UNIQUE;
```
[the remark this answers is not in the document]
jsalzbergedu: That is not actually checked there. Rather, it is only possible to obtain a mutable pointer by first going through an immutable reference. The mutable pointer will then not have write perms.

This can be see in the original SB calculus in this example here:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b82326f44255c2e10eac00a4c82cb5e4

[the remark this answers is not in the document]
jsalzbergedu: (If you run with miri)

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Push a tag with a permission perm at pointer
+    pub(super) fn push<U>(tag: PointerTag, perm: PermissionByte, pointer: *const U) {
+        // Decide whether to initialize the stacks
+        // for location:location+size_of(U).
+        // Offset has already been picked earlier.
+        unsafe {
+            use self::*;
+            if STATE == MonitorState::ON
+                && pointer_object(MONITORED) == pointer_object(pointer)
+                && pointer_offset(MONITORED) <= std::mem::size_of::<U>()
```
celinval, line 182: I don't think this will work for a struct field or an array element. This will also not work for unsized types (I believe you need to use `size_of_val_raw` for that and add `U: ?Sized`). If you do, please make this function `unsafe`.

I think you need this to be something like:
```suggestion
                && MONITORED.addr() >= pointer.addr()
                && MONITORED.addr() < pointer.addr().wrapping_add(std::mem::size_of_val_raw(pointer))
```

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,313 @@
...
+            {
+                STACK_TAGS[STACK_TOP] = tag;
```
celinval, line 184: Can you please add an assert here that STACK_TOP <= STACK_DEPTH with an error friendly message that the current limit has been reached like line 20 (maybe use an aux function for that).

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,313 @@
...
+            }
+        }
+    }
+
+    /// Run a stack check on the monitored byte for the given
+    /// tag and the given access permission.
+    pub(super) fn stack_check(tag: PointerTag, access: AccessBit) {
+        unsafe {
+            use self::*;
+            if STATE == MonitorState::ON {
```
celinval, line 196: you don't need this

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,313 @@
...
+        NEXT_TAG += 1;
+        monitors::track_local(tag, pointer);
+    }
+}
+
+#[rustc_diagnostic_item = "KaniStackCheckPtr"]
+fn stack_check_ptr<U>(pointer_value: *const *mut U) {
+    unsafe {
+        let tag = TAGS.get(pointer_value);
+        let perm = PERMS.get(pointer_value);
```
celinval, line 249: please move this to be inside the `if` block. This logic adds unnecessary overhead if this is not the pointer currently being monitored.

library/kani/src/aliasing.rs
```diff
@@ -0,0 +1,313 @@
...
+                STACK_TOP += 1;
+            }
+        }
+    }
+
+    /// Run a stack check on the monitored byte for the given
+    /// tag and the given access permission.
+    pub(super) fn stack_check(tag: PointerTag, access: AccessBit) {
+        unsafe {
+            use self::*;
```
celinval, line 195: you don't need this

tests/expected/aliasing/box.rs.fixme
```diff
@@ -0,0 +1,14 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zaliasing
+
+#[kani::proof]
+fn main() {
```
celinval, line 6: please use a more descriptive name for your harnesses.

tests/expected/aliasing/box.rs.fixme
```diff
@@ -0,0 +1,14 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Zaliasing
```
celinval, line 3: please add documentation on what it is the expected behavior

celinval (commented):
Ok. I think I'm almost done reviewing the library changes. Thanks for your patience!

I think it will be very helpful if you add more tests with things like structure fields, arrays, more elaborated tests as well. Even if they don't pass today, it will be helpful to know how these cases are mitigated. It will also be helpful if you can add tests where the alias analysis finds no error and one that it finds multiple errors.

PR #3409, opened 2024-08-02
Enable "Auto label" for merge-queue actions

tautschnig:
This will Fix merge-queue CI actions as this is a required CI action, but wasn't being run for jobs entering the queue.

### DIFF
--- a/.github/workflows/extra_jobs.yml
+++ b/.github/workflows/extra_jobs.yml
@@ -18,7 +18,9 @@
 # See <https://github.com/actions/labeler/issues/121> for more details.
 
 name: Kani Extra
-on: pull_request_target
+on:
+  pull_request_target:
+  merge_group:
 
 jobs:
   # Keep this job minimal since it requires extra permission


jaisnan (approved):
Thanks!

celinval (changes requested):
This workflow will potentially invoke the benchcomp workflow, which I think may require changes as well, since it has code that depends on the event trigger. Can you please double check! Thanks

tautschnig:
> This workflow will potentially invoke the benchcomp workflow, which I think may require changes as well, since it has code that depends on the event trigger. Can you please double check! Thanks

The job for invoking `benchcomp` workflows is defined as `verification-bench` in the very same file that this PR touches, so I believe we are actually fine here.

celinval:
In that workflow, we have this code to extract the commits that will be compared:

https://github.com/model-checking/kani/blob/4a9255706dfcb336db59d30fbe327f9ed2680396/.github/workflows/bench.yml#L20-L30

I think in the new case, it won't enter any of these steps, and the commits to be compared will not be defined.

tautschnig:
> In that workflow, we have this code to extract the commits that will be compared:
> 
> [...]

Oh, thank you for highlighting this, you are absolutely right. I will fix this, but the way I intend to fix this is by not running this performance check workflow for merge queues at all. My reasons for doing this are:
1. The job is not marked as "Required," so failures will not block the queue from progressing.
2. No one will look at the individual job outcomes when the merging attempt does not fail.
3. We run performance verification upon pushing to `main` anyway.

celinval:
> > In that workflow, we have this code to extract the commits that will be compared:
> > [...]
> 
> Oh, thank you for highlighting this, you are absolutely right. I will fix this, but the way I intend to fix this is by not running this performance check workflow for merge queues at all. My reasons for doing this are:
> 
>     1. The job is not marked as "Required," so failures will not block the queue from progressing.
> 
>     2. No one will look at the individual job outcomes when the merging attempt does not fail.
> 
>     3. We run performance verification upon pushing to `main` anyway.

Sounds good to me!

celinval (approved):

PR #3422, opened 2024-08-06
Handle intrinsics systematically

artemagvanian:
Since compiler intrinsics are handled both in codegen and instrumentation, it would make sense to have a single source of truth with regard to which intrinsics we support and update it systematically. 

This PR introduces an `Intrinsic` enum, which allows to parse intrinsic name into one of the predefined variants we support or a catch-all `Unsupported` variant.


kani-compiler/src/kani_middle/points_to/points_to_analysis.rs
```diff
@@ -202,161 +204,119 @@ impl<'a, 'tcx> Analysis<'tcx> for PointsToAnalysis<'a, 'tcx> {
...
-                                        self.successors_for_deref(state, args[0].node.clone());
-                                    let destination_set =
-                                        state.resolve_place(*destination, self.instance);
-                                    state.extend(&destination_set, &state.successors(&dst_set));
-                                    state.extend(&dst_set, &src_set);
-                                }
-                                // All `atomic_load` intrinsics take `src` as an argument.
-                                // This is equivalent to `destination = *src`.
-                                name if name.starts_with("atomic_load") => {
-                                    assert_eq!(
```
JustusAdam, line 232: Maybe I'm reading it wrong, but where did these assertions about the length of arguments go?

And perhaps, now that we have a separate type for this, maybe we should have some method like `Intrinsic::arg_len() -> usize` or `Intrinsic::arg_len_matches(usize) -> bool` that checks if the argument number is correct in one place?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -289,342 +290,274 @@ impl<'tcx> GotocCtx<'tcx> {
...
-            "simd_or" => codegen_intrinsic_binop!(bitor),
-            "simd_shl" | "simd_shr" => {
-                self.codegen_simd_shift_with_distance_check(fargs, intrinsic, place, loc)
+            Intrinsic::SimdOr => codegen_intrinsic_binop!(bitor),
+            Intrinsic::SimdShl | Intrinsic::SimdShr => {
+                self.codegen_simd_shift_with_distance_check(fargs, intrinsic_str, place, loc)
+            }
+            Intrinsic::SimdShuffle(stripped) => {
+                assert!(fargs.len() == 3, "`simd_shuffle` had unexpected arguments {fargs:?}");
+                let n: u64 = self.simd_shuffle_length(stripped, farg_types, span);
```
adpaco-aws, line 502: Why do this here? Wouldn't it be better to have `n` be the field for `Intrinsic::SimdShuffle`?

kani-compiler/src/intrinsics.rs
```diff
@@ -0,0 +1,359 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Single source of truth about which intrinsics we support.
+
+// Enumeration of all intrinsics we support right now, with the last option being a catch-all. This
+// way, adding an intrinsic would highlight all places where they are used.
+#[allow(unused)]
+#[derive(Clone, Copy, Debug)]
+pub enum Intrinsic<'a> {
```
adpaco-aws, line 10: Have you considered using the [symbols](https://github.com/rust-lang/rust/blob/d3a393932eeafa4638ae22f5ecbc38bf38760d0e/compiler/rustc_span/src/symbol.rs#L128) from `rustc_span` instead?

kani-compiler/src/kani_middle/points_to/points_to_analysis.rs
```diff
@@ -202,161 +204,119 @@ impl<'a, 'tcx> Analysis<'tcx> for PointsToAnalysis<'a, 'tcx> {
...
-                        name if name.starts_with("atomic") => {
-                            match name {
-                                // All `atomic_cxchg` intrinsics take `dst, old, src` as arguments.
-                                // This is equivalent to `destination = *dst; *dst = src`.
-                                name if name.starts_with("atomic_cxchg") => {
-                                    assert_eq!(
-                                        args.len(),
-                                        3,
-                                        "Unexpected number of arguments for `{name}`"
-                                    );
```
adpaco-aws, line 215: I'm not familiar with this code, but where are these checks being done now? Shouldn't they be done in `from_str`?

kani-compiler/src/kani_middle/points_to/points_to_analysis.rs
artemagvanian: These assertions were duplicating similar checks that happen in the codegen implicitly, so they seemed unnecessary and fragile to me. 

I agree that we need a better way to ensure argument types/length for the intrinsics. Perhaps we can make the `Intrinsic` enum constructor to accept a function signature (e.g. a list of StableMIR types) that is validated against what we expect it to be?

kani-compiler/src/kani_middle/points_to/points_to_analysis.rs
```diff
@@ -652,3 +612,136 @@ impl<'a, 'tcx> PointsToAnalysis<'a, 'tcx> {
...
+        | Intrinsic::UncheckedDiv
+        | Intrinsic::UncheckedRem
+        | Intrinsic::Unlikely
+        | Intrinsic::VtableSize
+        | Intrinsic::VtableAlign
+        | Intrinsic::WrappingAdd
+        | Intrinsic::WrappingMul
+        | Intrinsic::WrappingSub
+        | Intrinsic::WriteBytes => {
+            /* Intrinsics that do not interact with aliasing beyond propagating it. */
```
adpaco-aws, line 713: Sir, is this a C++ comment?

kani-compiler/src/intrinsics.rs
```diff
@@ -0,0 +1,359 @@
...
+            name if name.starts_with("atomic_nand") => {
+                Self::AtomicNand(name.strip_prefix("atomic_nand_").unwrap())
+            }
+            name if name.starts_with("atomic_or") => {
+                Self::AtomicOr(name.strip_prefix("atomic_or_").unwrap())
+            }
+            name if name.starts_with("atomic_singlethreadfence") => Self::AtomicSingleThreadFence(
+                name.strip_prefix("atomic_singlethreadfence_").unwrap(),
+            ),
+            name if name.starts_with("atomic_store") => {
```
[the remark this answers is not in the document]
artemagvanian: I like the macro approach! I will add it to the code.

kani-compiler/src/kani_middle/transform/check_uninit/ptr_uninit/uninit_visitor.rs
```diff
@@ -182,46 +185,30 @@ impl MirVisitor for CheckUninitVisitor {
...
                         InstanceKind::Intrinsic => {
-                            match instance.intrinsic_name().unwrap().as_str() {
+                            match Intrinsic::from_str(instance.intrinsic_name().unwrap().as_str()) {
                                 intrinsic_name if can_skip_intrinsic(intrinsic_name) => {
                                     /* Intrinsics that can be safely skipped */
                                 }
-                                name if name.starts_with("atomic") => {
-                                    let num_args = match name {
-                                        // All `atomic_cxchg` intrinsics take `dst, old, src` as arguments.
-                                        name if name.starts_with("atomic_cxchg") => 3,
```
adpaco-aws, line 192: Again, I don't where these checks have gone.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
artemagvanian: Agreed. I simply copied this code from before the match and kept it as-is.

kani-compiler/src/kani_middle/points_to/points_to_analysis.rs
artemagvanian: @JustusAdam asked a similar question. These checks are implicitly done in the codegen, but I agree that having a centralized way to check those properties is better.

kani-compiler/src/intrinsics.rs
artemagvanian: I don't think this would make it easier because Kani does not support every single intrinsic. Using symbols also would not allow us to split intrinsics by their prefix

kani-compiler/src/kani_middle/points_to/points_to_analysis.rs
adpaco-aws: Ok, but there are many other checks that aren't done implicitly (or at least they weren't when I was doing the work) like for example the type-checking done in `codegen_intrinsic_simd_extract`. What do you think we should do with them?

artemagvanian: I am leaning towards the approach of making the the `Intrinsic` enum constructor to accept a function signature (e.g. a list of StableMIR types) that is validated against what we expect it to be

artemagvanian:
@JustusAdam @adpaco-aws I added argument type/count checks, which run every time an intrinsic enum is created. I think this is a better and more systematic approach

kani-compiler/src/intrinsics.rs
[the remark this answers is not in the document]
artemagvanian: Actually, given that the code has changed quite a bit, I wonder if adding another macro is a good idea. I would be leaning to leaving this in if it is ok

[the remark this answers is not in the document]
JustusAdam: I think what I care most about is to avoid the unnecessary `unwrap`. I personally prefer macros to do it, but you can just as well use the if-let chain I suggested first.

JustusAdam:
I also saw a lot of matches on groups of these operations. Perhaps it would make sense to move those into helper functions like `is_simd()`?

kani-compiler/src/intrinsics.rs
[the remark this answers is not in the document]
adpaco-aws: In case you are still wondering about this: I recommend using the `if let Some(...) = ...` chains.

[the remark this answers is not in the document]
artemagvanian: Added that ^

adpaco-aws (approved):

artemagvanian:
> I also saw a lot of matches on groups of these operations. Perhaps it would make sense to move those into helper functions like `is_simd()`?

Good idea, I added that too

PR #2532, opened 2023-06-15
[Draft] Introduce global conditions with support for `fail_uncoverable` (new condition) and `should_panic` (rework)

adpaco-aws:
### Description of changes: 

In this draft PR we:
 * Introduce **Global Conditions** (currently being discussed in #2516)
 * Re-implement the `should_panic` feature as a global condition
 * Add support for a new `fail_uncoverable` condition

The PR has been opened to serve as a Proof-of-Concept for #2516 , therefore it's not recommended to review it unless the PR has been merged (I'll update the description when that happens).

### Resolved issues:

Resolves #2299 

### Related RFC:

Optional #2516 , #2272 , #1905 

### Call-outs:

There are a few things to be done for this implementation. One thing I'm considering is moving all `struct`s and logic related to global conditions to a new module. Another one is to convert the `compute_<cond>_condition` into associated functions of `GlobalCondition`.

### Testing:

* How is this change tested? New tests.

* Is this a refactor change? No.

### Checklist
- [x] Each commit message has a non-empty body, explaining why the change was made
- [x] Methods or procedures are documented
- [x] Regression or unit tests are included, or existing tests cover the modified code
- [x] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #3424, opened 2024-08-07
`cover_or_fail!` macro

### LINKED ISSUE #2792: fail harness if `cover` is not satisfied
Requested feature: fail harness 
Use case:

I just realized that the current [bolero integration](https://github.com/camshaft/bolero/blob/8d8464791313b755465e846062d0921ecf1d030d/lib/bolero-kani/src/lib.rs#L55) allows a proof to go through, even if you've specified a generator that is impossible to satisfy:

```#[test]
#[kani::proof]
fn my_test() {
    bolero::check!().with_type::<u8>().filter(|_| false).for_each(|_| {
        panic!();
    });
}```

```$ cargo kani --tests --harness my_test --output-format=terse
Checking harness my_test...

VERIFICATION RESULT:
 ** 0 of 255 failed

 ** 0 of 1 cover properties satisfied (1 unreachable) // <= this should ideally fail the proof

VERIFICATION:- SUCCESSFUL
Verification Time: 0.14628594s

Complete - 1 successfully verified harnesses, 0 failures, 1 total.```

Ideally, we'd replace the call to `kani::assume` with a cover call that would force harness failure. This could be called something like `kani::cover_or_fail!` or `kani::assert_cover!`.

Doing a test with multiple harnesses, it seems to behave as one would expect and only applies to the individual harness:

```rust
#[test]
#[kani::proof]
fn my_test() {
    bolero::check!().with_type::<u8>().for_each(|_| {
        // noop
    });
}

#[test]
#[kani::proof]
fn my_test() {
    bolero::check!().with_type::<u8>().filter(|_| false).for_each(|_| {
        panic!();
    });
}

#[kani::proof]
fn my_kani_proof() {
    assert_eq!(1 + 2, 3);
}
```

```
VERIFICATION:- SUCCESSFUL
Verification Time: 0.89504564s

Checking harness add_test...

VERIFICATION RESULT:
 ** 0 of 1502 failed (69 unreachable)

 ** 1 of 1 cover properties satisfied // <= the generator produced at least 1 valid value


VERIFICATION:- SUCCESSFUL
Verification Time: 0.8548021s

Checking harness my_test...

VERIFICATION RESULT:
 ** 0 of 255 failed

 ** 0 of 1 cover properties satisfied (1 unreachable) // <= the generator was UNSAT; should fail the harness!


VERIFICATION:- SUCCESSFUL
Verification Time: 0.14286228s

Checking harness my_kani_proof...

VERIFICATION RESULT:
 ** 0 of 2 failed  // <= this harness doesn't use bolero at all; no references to the cover property

VERIFICATION:- SUCCESSFUL
Verification Time: 0.024725173s
```

carolynzech:
This PR adds a `cover_or_fail!` macro to Kani. This macro has the same functionality as the existing `cover!` macro, except it causes verification to fail when the outcome is unsatisfiable or unreachable.

Currently, a caller can invoke this macro from anywhere in their code (not just harnesses), which is also true for `cover`. However, we only analyze reachability starting from a given proof harness (the entry point), so if the `cover_or_fail` call is contained inside a function that the harness does not call, the verification will succeed even though the call actually is not covered. I'm concerned that this behavior will produce confusing results for users. Specifically, see the test cases inside `cover-unreachable-outside-harness-pass` (which pass verification) and `cover-unreachable-outside-harness-fail` (which fail verification). See also the discussion of this issue [here](https://github.com/model-checking/kani/pull/1906#issuecomment-1318036336). Is this the behavior that we want?

Our options are:

1. Leave the implementation like this, with clear documentation of the reachability behavior. The benefit of this design is that users can more easily use `cover_or_fail` throughout their code for debugging.
2. Add a pass which scans the entire crate for calls to `cover_or_fail` and makes sure that CBMC marked them all as reachable. This allows using `cover_or_fail` throughout the codebase and avoids any confusing reachability behavior, but may be slow.
3. Restrict calls to `cover_or_fail` to proof harnesses only. This approach solves the reachability issue, since CBMC will see every call to `cover_or_fail`, but does restrict users--the author of the issue that prompted this PR, for example, wanted to use it outside of a proof harness.

I'm leaning toward #1 or #3, but curious to hear others' thoughts.

Resolves #2792

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
@@ -45,6 +45,11 @@ pub enum PropertyClass {
     ///
     /// SPECIAL BEHAVIOR: "Errors" for this type of assertion just mean "reachable" not failure.
     Cover,
+    /// See [GotocCtx::codegen_cover] below. Generally just an `assert(false)` that's not an error.
+    ///
+    /// SPECIAL BEHAVIOR: Same as PropertyClass::Cover, except unreachable or unsatisfiable  
+    /// conditions cause verification failure.
+    CoverOrFail,
     /// The class of checks used for code coverage instrumentation. Only needed
     /// when working on coverage-related features.
     ///
@@ -145,6 +150,16 @@ impl<'tcx> GotocCtx<'tcx> {
         // https://github.com/diffblue/cbmc/issues/6613). So for now use
         // assert(!cond).
         self.codegen_assert(cond.not(), PropertyClass::Cover, msg, loc)
+    }
+
+    /// Generate code to cover the given condition at the current location
+    pub fn codegen_cover_or_fail(&self, cond: Expr, msg: &str, span: SpanStable) -> Stmt {
+        let loc = self.codegen_caller_span_stable(span);
+        // Should use Stmt::cover, but currently this doesn't work with CBMC
+        // unless it is run with '--cover cover' (see
+        // https://github.com/diffblue/cbmc/issues/6613). So for now use
+        // assert(!cond).
+        self.codegen_assert(cond.not(), PropertyClass::CoverOrFail, msg, loc)
     }
 
     /// Generate a cover statement for code coverage reports.
--- a/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
@@ -79,6 +79,47 @@ impl GotocHook for Cover {
     }
 }
 
+/// A hook for Kani's `cover_or_fail` function (declared in `library/kani/src/lib.rs`).
+/// The function takes two arguments: a condition expression (bool) and a
+/// message (&'static str).
+/// The hook codegens the function as a cover property that checks whether the
+/// condition is satisfiable. If the condition is undetermined, unreachable,
+/// or unsatisfiable, verification fails.
+struct CoverOrFail;
+impl GotocHook for CoverOrFail {
+    fn hook_applies(&self, tcx: TyCtxt, instance: Instance) -> bool {
+        matches_function(tcx, instance.def, "KaniCoverOrFail")
+    }
+
+    fn handle(
+        &self,
+        gcx: &mut GotocCtx,
+        _instance: Instance,
+        mut fargs: Vec<Expr>,
+        _assign_to: &Place,
+        target: Option<BasicBlockIdx>,
+        span: Span,
+    ) -> Stmt {
+        assert_eq!(fargs.len(), 2);
+        let cond = fargs.remove(0).cast_to(Type::bool());
+        let msg = fargs.remove(0);
+        let msg = gcx.extract_const_message(&msg).unwrap();
+        let target = target.unwrap();
+        let caller_loc = gcx.codegen_caller_span_stable(span);
+
+        let (msg, reach_stmt) = gcx.codegen_reachability_check(msg, span);
+
+        Stmt::block(
+            vec![
+                reach_stmt,
+                gcx.codegen_cover_or_fail(cond, &msg, span),
+                Stmt::goto(bb_label(target), caller_loc),
+            ],
+            caller_loc,
+        )
+    }
+}
+
 struct Assume;
 impl GotocHook for Assume {
     fn hook_applies(&self, tcx: TyCtxt, instance: Instance) -> bool {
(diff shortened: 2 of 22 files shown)


adpaco-aws:
Thanks @carolynzech for this! It's nice to finally see a concrete implementation for this feature.

I do have some thoughts about it too. I suppose you know about the very related option `--fail-uncoverable` which essentially converts all `cover!` macros into something like this. The proposal was added to the [RFC: Cover statement](https://model-checking.github.io/kani/rfc/rfcs/0003-cover-statement.html#impact-on-overall-verification-status) in #2967 but we didn't get to finish the implementation in #2532 (#2967 in particular had a [nice discussion](https://github.com/model-checking/kani/pull/2967#issuecomment-1881486704) based on theoretical implementations of the `cover_or_fail!` macro 😄 ). This is mostly to say that the `cover_or_fail!` macro can be implemented in a number of ways, and we won't know which option is the best without user feedback. We can try to guess which one is the best though.

In my opinion, option 3 would restrict too much the usage for the macro to be useful. Option 1 seems fine to me and it's the same we were doing for `--fail-uncoverable`. But I'm wondering now: Can we do better? Maybe we could something like a mix of options 1 and 2 where (1) we scan the codebase and emit a warning for each `cover_or_fail!` macro that is going to get sliced away and (2) perform the analysis as it's being proposed in option 1. However, I'm not sure this option would be acceptable if we need to add another compiler pass just for it - it would be a noticeable overhead even when users don't have any `cover_or_fail!` macros in their code. What do you think?

carolynzech:
> I suppose you know about the very related option `--fail-uncoverable` which essentially converts all `cover!` macros into something like this. 
>
Yes, I did see that. This implementation is subtly different in that it allows users to mix and match `cover!` with `cover_or_fail!`, rather than use a flag to force all covers to become failures. That might be nice for debugging if a user just wants to focus on a few failing cover statements at a time, rather than get a wall of output for all of the cover statements. In practice, it's probably not a huge difference, and the command line flag would have the benefit of reducing some of the code duplication between `cover/cover_or_fail` that we see here.
> In my opinion, option 3 would restrict too much the usage for the macro to be useful. 
>
I would tend to agree--it's a less messy solution for sure, but that's moot if it doesn't help users in the first place.

> Maybe we could something like a mix of options 1 and 2 where (1) we scan the codebase and emit a warning for each `cover_or_fail!` macro that is going to get sliced away and (2) perform the analysis as it's being proposed in option 1. However, I'm not sure this option would be acceptable if we need to add another compiler pass just for it - it would be a noticeable overhead even when users don't have any `cover_or_fail!` macros in their code. What do you think?
>
If I imagine myself as a user of Kani unfamiliar with the implementation, I would expect that if I put `cover_or_fail` throughout my code and verification succeeded, Kani found _all_ those statements to be reachable from the harness. My concern is that users won't read the fine print re: slicing and take the successful verification to mean something it doesn't.
However, your performance point is well-taken--we don't expect people to use this macro _that_ often, so doing a whole extra pass just to check for it seems like a waste. Perhaps we flip the order of your proposed steps--by default, we do (2) and do the analysis as proposed in option 1, but offer a `--comprehensive` flag that will do (1). Additionally, for each `cover_or_fail` that we do see, if we detect that `--comprehensive` was _not_ passed, we can emit a warning suggesting it. This of course does not help in the case that no cover statements were detected in the first place, but (hopefully) the user is more likely to notice that anyway. I think this warning would cover the more common human error case where a few cover statements of many were not analyzed and the user doesn't notice.

adpaco-aws:
> Perhaps we flip the order of your proposed steps--by default, we do (2) and do the analysis as proposed in option 1, but offer a `--comprehensive` flag that will do (1). Additionally, for each `cover_or_fail` that we do see, if we detect that `--comprehensive` was _not_ passed, we can emit a warning suggesting it. This of course does not help in the case that no cover statements were detected in the first place, but (hopefully) the user is more likely to notice that anyway. I think this warning would cover the more common human error case where a few cover statements of many were not analyzed and the user doesn't notice.

If we are going to have an option that warns about `cover_or_fail!`s being left out, we might as well error out instead. That will make sure the user notices 😆 In fact, I believe that's the main thing users want out of this feature: if some line of code stops being verified for whatever reason, they want to know. Warnings will likely go unnoticed, so they would not know.

I'd be happy with a feature like that. We need to make sure the reachability caveat is well documented, and I'd probably use a more detailed name for the flag, but the rest sounds good to me.

carolynzech:
Closing for now; this feature requires further design discussion.

PR #3426, opened 2024-08-07
Stabilize pointer-to-reference cast validity checks

### LINKED ISSUE #3425: Stabilize pointer-to-reference cast validity checks
#3221 introduced additional pointer validity checks when performing pointer-to-reference casts to aid with modular verification. 

This led to one harness run on the CI inside `s2n-quic` timing out, so the checks were introduced behind an unstable flag. The reason for the timeout was that a check was inserted inside a nested loop, which blew up verification time. 

One of the pathways for stabilizing the checks is excluding the harness from the CI.

artemagvanian:
This PR stabilizes pointer-to-reference cast validity checks, so that they are run by default. 

Resolves #3425 

### DIFF
--- a/.github/workflows/verify-std-check.yml
+++ b/.github/workflows/verify-std-check.yml
@@ -59,7 +59,7 @@ jobs:
         continue-on-error: true
         run: |
           kani verify-std -Z unstable-options ./library --target-dir ${{ runner.temp }} -Z function-contracts \
-            -Z mem-predicates -Z ptr-to-ref-cast-checks
+            -Z mem-predicates
 
       # If the head failed, check if it's a new failure.
       - name: Checkout base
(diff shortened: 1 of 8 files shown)


feliperodri (approved):
How can we know for sure that any performance issues related to `--ptr-to-ref-cast-checks` have been resolved?

artemagvanian:
The only performance issue that we had was that the `perf` check failed because one harness consistently ran out of memory on the CI. In the current run all required checks passed.

zhassan-aws:
I think it would be useful to have an option to turn those checks off. Is there such an option?

artemagvanian:
Not as of now, but it won't be too hard to implement. I wonder if it is worth it, though.

Just to give some context: the performance impact of the check itself is negligible. We encountered the issue above because the check was inserted into a place nested inside two loops, which magnified its impact. 

On top of that, as far as I remember, the problematic harness ran OOM on the CI sometimes already, so the checks seemed to just be the last drop. It all makes me believe this is an isolated instance rather than a systemic problem.

PR #3431, opened 2024-08-09
Update CBMC build instructions for Amazon Linux 2

tautschnig:
The default compiler on Amazon Linux 2 is GCC 7, which is not sufficiently recent for building CBMC v6+. Install and use GCC 10, adjust warnings to cope with the flex version provided by Amazon Linux 2, and handle the non-default std::filesystem support. Furthermore, apply a workaround until https://github.com/diffblue/cbmc/issues/8357 has been fixed on the CBMC side.


feliperodri (approved):
Thank you!

PR #3436, opened 2024-08-12
Actually apply CBMC patch

tautschnig:
The patch introduced in #3431 not only needs to be created, but also needs to be applied.


feliperodri (approved):
😂

PR #3438, opened 2024-08-14
Avoid corner-cases by grouping instrumentation into basic blocks and using backward iteration

artemagvanian:
This PR is a follow-up to https://github.com/model-checking/kani/pull/3374.

Its main purpose is to properly handle a corner-case when multiple instrumentation instructions are added to the same instruction and not all of them are skipped during subsequent instrumentation.

For example, if instrumentation is added after a terminator, a new basic block will be created, containing the added instrumentation. However, we currently only mark the first statement (or the terminator, if there are none) of the new basic block as skipped for subsequent instrumentation. That means that if instrumentation in this case contains some instrumentation targets itself, it will never terminate. 

Coincidentally, this is not currently the case, but could lead to subtle bugs if we decide to change instrumentation. In fact, this bug was only surfaced when I experimented with checking all memory accesses, which introduced significantly more checks. Ultimately, this shows that the current way to avoiding instrumentation is limited and needs to be changed.

The PR introduces the following changes:

- Group instrumentation into separate basic blocks instead of adding instructions one-by-one.
- Use backward iteration to avoid having to reason about which instructions need to be skipped.


celinval:
Can you please describe what the corner-cases are?

artemagvanian:
> Can you please describe what the corner-cases are?

I expanded a little bit on that in the PR description. I encountered this corner case when testing enabling instrumentation with standard library. I am not sure of the exact configuration that triggered it, but the general issue was that skipping only the first statement in the new basic block when splitting was not enough and instrumentation was caught in the infinite loop.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -465,36 +522,43 @@ impl<'a> UninitInstrumenter<'a> {
...
 
 /// If injecting a new call to the function before the current statement, need to skip the original
 /// statement when analyzing it as a part of the new basic block.
-fn collect_skipped(operation: &MemoryInitOp, body: &MutableBody, skip_first: &mut HashSet<usize>) {
-    if operation.position() == InsertPosition::Before {
+fn collect_skipped(
+    position: InsertPosition,
+    source: &SourceInstruction,
+    body: &MutableBody,
+    skip_first: &mut HashSet<usize>,
```
celinval, line 557: shouldn't this be `HashSet<BasicBlockIdx>`?

celinval (commented):
I think this logic to track which instructions we still need to visit is adding unnecessary complexity to this code. You mentioned that @jsalzbergedu suggested visiting the basic blocks in reverse order. I think that would make this code much cleaner, don't you think?

Also, since this is meant to fix an issue, I expect at least one test that ensures this issue doesn't regress later. Thanks

artemagvanian:
> I think this logic to track which instructions we still need to visit is adding unnecessary complexity to this code. You mentioned that @jsalzbergedu suggested visiting the basic blocks in reverse order. I think that would make this code much cleaner, don't you think?

That's a good idea! I implemented a version of it, and it seemed to make the code significantly cleaner. 

The only issue, which has nothing to do with the approach itself, is that I can't call `visit_statement` and `visit_terminator` directly when implementing backward iteration. The reason for that is a private field in the `Location` newtype struct. I managed to find a workaround to unblock myself by transmuting between those objects and it seems to work, but it's obviously not a good idea. Perhaps we should propose an upstream change in StableMIR to address that properly.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
artemagvanian: We were able to get rid of it altogether! 😁

kani-compiler/src/kani_middle/transform/check_uninit/delayed_ub/instrumentation_visitor.rs
```diff
@@ -44,13 +37,23 @@ impl<'a, 'tcx> TargetFinder for InstrumentationVisitor<'a, 'tcx> {
...
     ) -> Option<InitRelevantInstruction> {
-        self.skip_next = skip_first;
-        self.current = SourceInstruction::Statement { idx: 0, bb };
         self.target = None;
-        self.visit_basic_block(&body.blocks()[bb]);
+        match *source {
+            SourceInstruction::Statement { idx, bb } => {
+                let BasicBlock { statements, .. } = &body.blocks()[bb];
+                let stmt = &statements[idx];
+                self.visit_statement(stmt, self.__location_hack_remove_before_merging(stmt.span))
```
celinval, line 47: Do we need to use the `Visitor` trait at this point?

artemagvanian: Added a commit removing `MirVisitor` dependency

kani-compiler/src/kani_middle/transform/check_uninit/delayed_ub/instrumentation_visitor.rs
```diff
@@ -63,43 +57,52 @@ impl<'a, 'tcx> InstrumentationVisitor<'a, 'tcx> {
...
-        } else if self.target.is_none() {
-            // Check all inner places.
-            self.super_statement(stmt, location);
-        }
+        self.super_statement(stmt, location);
         // Switch to the next statement.
-        let SourceInstruction::Statement { idx, bb } = self.current else { unreachable!() };
-        self.current = SourceInstruction::Statement { idx: idx + 1, bb };
+        if let SourceInstruction::Statement { idx, bb } = self.current_instruction {
+            self.current_instruction = SourceInstruction::Statement { idx: idx + 1, bb }
```
celinval, line 93: don't you need to set the current instruction before visiting the instruction itself? I.e.: before `super_statement` call?

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -31,13 +31,8 @@ mod relevant_instruction;
...
 
 /// Trait that the instrumentation target providers must implement to work with the instrumenter.
-trait TargetFinder {
-    fn find_next(
-        &mut self,
-        body: &MutableBody,
-        bb: BasicBlockIdx,
-        skip_first: bool,
-    ) -> Option<InitRelevantInstruction>;
+pub trait TargetFinder {
```
celinval, line 34: why do you need a trait?

kani-compiler/src/kani_middle/transform/check_uninit/delayed_ub/instrumentation_visitor.rs
```diff
@@ -41,17 +38,14 @@ pub struct InstrumentationVisitor<'a, 'tcx> {
...
-        self.visit_basic_block(&body.blocks()[bb]);
-        self.target.clone()
+    fn find_all(&mut self, body: &MutableBody) -> Vec<InitRelevantInstruction> {
+        for (bb_idx, bb) in body.blocks().iter().enumerate() {
+            self.current_instruction = SourceInstruction::Statement { idx: 0, bb: bb_idx };
+            self.visit_basic_block(bb);
+        }
+        // Push the last current target into the list.
+        self.targets.push(self.current_target.clone());
+        self.targets.clone()
```
celinval, line 48: Do we need to clone this? You could make this function take `self` by value and return `self.targets`.

I'm assuming we don't want this function to be called multiple times, right?

kani-compiler/src/kani_middle/transform/check_uninit/delayed_ub/instrumentation_visitor.rs
```diff
@@ -18,20 +18,17 @@ use rustc_middle::ty::TyCtxt;
...
-    /// later.
-    skip_next: bool,
-    /// The instruction being visited at a given point.
-    current: SourceInstruction,
-    /// The target instruction that should be verified.
-    pub target: Option<InitRelevantInstruction>,
+    /// All target instructions in the body.
+    targets: Vec<InitRelevantInstruction>,
+    /// Currently analyzed instruction.
+    current_instruction: SourceInstruction,
```
celinval, line 29: This seems to be a duplicate of `current_target.source`... what am I missing?

celinval (commented):
This is cleaner. Thanks!

celinval:
> Also, since this is meant to fix an issue, I expect at least one test that ensures this issue doesn't regress later. Thanks

I'm wondering if you missed this comment. Can you please add a test?

kani-compiler/src/kani_middle/transform/check_uninit/delayed_ub/instrumentation_visitor.rs
artemagvanian: It is initially set before `visit_basic_block` is created and then incremented on every iteration:
https://github.com/model-checking/kani/pull/3438/files/58b689a7cd817e9eeb9402f22abb9016911bdc8c#diff-471a7a8b457c94203de190d77f830f88b61e027680d347e9768c24f926a19270R42-R44

I could potentially change the `current_instruction` to be of the `Option<SourceInstruction>` type and initialize it if it is set to None, but it becomes a bit more complicated when we need to account for multiple basic blocks visited.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
artemagvanian: There are two so-called target finders: one for finding regular uninitialized memory places to instrument and one for delayed UB. Both of them cause the exact same instrumentation to be added (though in different places). Instrumentation is handled by `UninitInstrumenter` struct, so to avoid duplicating the code I decided to parametrize it over `TargetFinder`, hence the trait.

kani-compiler/src/kani_middle/transform/check_uninit/delayed_ub/instrumentation_visitor.rs
artemagvanian: Correct, I'll change that. Thanks!

kani-compiler/src/kani_middle/transform/check_uninit/delayed_ub/instrumentation_visitor.rs
artemagvanian: My intention was to use `current_instruction` as a counter, which is changed while iterating over basic blocks and `current_target.source` to be a part of the instrumentation target. The difference between two was supposed to indicate that we have already switched to the next statement/terminator. 

I agree though that `current_instruction` seems redundant. I will refactor it slightly to use `current_target.source` for both purposes.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
celinval: ok... thanks

celinval (approved):
Modulo missing tests.

celinval:
I wonder if there was a bug in the experiment itself, not in the instrumentation. It is a bit unsettling that this potential bug is not clearly understood, but the new approach seems safe. Can you still please ensure we have tests that add multiple instrumentation for a single MIR instruction? Thanks

artemagvanian:
> I wonder if there was a bug in the experiment itself, not in the instrumentation. It is a bit unsettling that this potential bug is not clearly understood, but the new approach seems safe. Can you still please ensure we have tests that add multiple instrumentation for a single MIR instruction? Thanks

I added tests where an instruction gets instrumented more than once. Regarding the experiment, I ran a naive version of delayed UB checks, where instead of filtering places to instrument by aliasing, all of the places get instrumented. I ran it under the assumption that places added by instrumentation will be correctly skipped, which was not the case.

PR #3120, opened 2024-04-02
Add code scanner tool

celinval:
This is just a utility tool that allow us to scan a crate and extract some metrics out of it. I also added a script to scan the standard library.

Currently, the tool will produce a few CSV files with raw data.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #3443, opened 2024-08-16
Add loop scanner to tool-scanner

qinheping:
This extend #3120 with loop scanner that counts the number of loops in each function and the number of functions that contain loops.

### DIFF
--- a/tools/scanner/src/analysis.rs
+++ b/tools/scanner/src/analysis.rs
@@ -573,9 +573,9 @@ impl<'a> MirVisitor for IteratorVisitor<'a> {
                     self.props.loops += 1;
                 }
             }
-            TerminatorKind::InlineAsm { destination, .. } => {
-                if let Some(target) = destination {
-                    self.props.loops += self.visited_blocks.contains(target) as usize;
+            TerminatorKind::InlineAsm { destination: Some(target), .. } => {
+                if self.visited_blocks.contains(target) {
+                    self.props.loops += 1;
                 }
             }
             // No targets in other terminators.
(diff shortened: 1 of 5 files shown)


tools/scanner/src/analysis.rs
```diff
@@ -461,10 +494,23 @@ impl FnLoops {
...
+    }
+
     fn visit_terminator(&mut self, term: &Terminator, location: Location) {
+        // A goto is identified as a loop latch if its target number is less than
+        // the number of visited basic block.
+        if let TerminatorKind::Goto { target } = &term.kind {
+            if self.num_visited_blocks > *target {
+                self.props.loops += 1;
+            }
+        }
```
celinval, line 513: This is not true for MIR though, you need to traverse the graph.

qinheping: I changed it to find all goto with target that has been visited.

qinheping: I also add the visitor for other terminators that also contains targets beyond GOTO.

tools/scanner/src/analysis.rs
```diff
@@ -461,50 +494,94 @@ impl FnLoops {
...
-                            "rposition",
-                            "nth",
-                            "count",
-                            "last",
-                            "find",
-                        ]
-                        .contains(last)
-                    {
-                        self.props.iterators += 1;
+        match &term.kind {
```
celinval, line 507: I think you can simplify this by calling `successors()` in the terminator before the match statement: https://github.com/rust-lang/rust/blob/4fe1e2bd5bf5a6f1cb245f161a5e9d315766f103/compiler/stable_mir/src/mir/body.rs#L130

celinval (approved):
Thanks

PR #3452, opened 2024-08-19
Make points-to analysis handle all intrinsics explicitly

### LINKED ISSUE #3447: Point-to-analysis fail if an instance does not have a body
I tried this code:

```rust
#[kani::proof]
fn check_sz() {
    let refstr = "hello";
    let ptr = refstr as *const _;
    assert_eq!(size_of_val(&ptr), 4);
}
```

```
kani check_sz.rs -Z uninit-checks
```

with Kani version:

Instead, this happened: Kani crashes

```
Kani Rust Verifier 0.54.0 (standalone)
thread 'rustc' panicked at kani-compiler/src/kani_middle/points_to/points_to_analysis.rs:449:54:
called `Option::unwrap()` on a `None` value

[Kani] no current codegen item.
[Kani] no current codegen location.

```

artemagvanian:
Initially, points-to analysis tried to determine the body of an intrinsic (if it was available) to avoid enumerating them all. However, it turned out this logic was faulty, and the analysis attempted to query the body for intrinsics that didn't have it and ICEd.

I added a couple of missing intrinsics, which had a side benefit of removing some duplicate assertion failures.

Resolves #3447

### DIFF
--- a/kani-compiler/src/kani_middle/points_to/points_to_analysis.rs
+++ b/kani-compiler/src/kani_middle/points_to/points_to_analysis.rs
@@ -203,119 +203,108 @@ impl<'a, 'tcx> Analysis<'tcx> for PointsToAnalysis<'a, 'tcx> {
             };
             match instance.def {
                 // Intrinsics could introduce aliasing edges we care about, so need to handle them.
-                InstanceKind::Intrinsic(def_id) => {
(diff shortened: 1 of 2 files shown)


celinval (approved):
Thanks

Issue #2646, opened 2023-07-31
[Feature Request] Ability to stub out function on primitive types (such as `[T]`)


roypat:
Requested feature: I recently ran across a situation where I wanted to stub the `binary_search_by_key` function on slices because I knew my slice would always have length 1, which would allow me to save on some unrolls. I ran into some problems expressing this stub however, since `[T]::binary_search_by_key` is not a valid path (even the `impl [T]` in the standard library seems to be special-cased into the compiler).  

Test case:

I'm really not sure how to express this, since `[T]` cannot really be part of paths, but I guess I'd like 
```rust
// kani_example.rs
fn stub_len<T>(this: &[T]) -> usize {
    0
}

#[kani::proof]
#[kani::stub(std::slice::[T]::len, stub_len)]
fn verify_len() {
    assert_eq!(vec![1].as_slice().len(), 0);
}
```
to work. Currently, the output I get from running `kani kani_example.rs` is 

```
$ kani kani_example.rs 
Kani Rust Verifier 0.33.0 (standalone)
warning: unused variable: `this`
 --> kani_example.rs:1:16
  |
1 | fn stub_len<T>(this: &[T]) -> usize {
  |                ^^^^ help: if this is intentional, prefix it with an underscore: `_this`
  |
  = note: `#[warn(unused_variables)]` on by default

warning: function `stub_len` is never used
 --> kani_example.rs:1:4
  |
1 | fn stub_len<T>(this: &[T]) -> usize {
  |    ^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

error: attribute `kani::stub` takes two path arguments; found 0
 --> kani_example.rs:6:1
  |
6 | #[kani::stub(std::slice::[T]::len, stub_len)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the attribute macro `kani::stub` (in Nightly builds, run with -Z macro-backtrace for more info)
```

celinval:
Just to clarify, the problem is that slice methods are represented using [qualified paths](https://doc.rust-lang.org/reference/paths.html#qualified-paths), and today we only support simple paths.

For example, here is a dummy test that calls `len()` via a function pointer:

```rust
#[test]
fn test_len() {
    assert_eq!(len_ptr(&[1u8]), 1);
}

fn len_ptr<T>(slice: &[T]) -> usize {
    let ptr = <[T]>::len;
[...]

PR #3457, opened 2024-08-22
Extend name resolution to support qualified paths (Partial Fix)

celinval:
This is the first part needed for us to add support to stubbing trait implementations (https://github.com/model-checking/kani/issues/1997) and primitive types (https://github.com/model-checking/kani/issues/2646).

After this change, we need to change stubs to accept a `FnResolution` instead of `FnDef`. We also need to find out how to retrieve methods of primitive types.


kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -787,20 +790,47 @@ fn parse_unwind(tcx: TyCtxt, attr: &Attribute) -> Option<u32> {
...
-        if let Err(err) = result {
-            tcx.dcx().span_err(attr.span, format!("failed to resolve `{name}`: {err}"));
+    let check_resolve = |attr: &Attribute, path: &TypePath| {
+        let result = resolve_fn_path(tcx, current_module.to_local_def_id(), path);
+        match result {
+            Ok(FnResolution::Fn(_)) => { /* no-op */ }
+            Ok(FnResolution::FnImpl { .. }) => {
+                tcx.dcx().span_err(
+                    attr.span,
+                    "Kani currently does not support stubbing trait implementation.",
```
zhassan-aws, line 800:
```suggestion
                    "Kani currently does not support stubbing trait implementations.",
```

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -787,20 +790,47 @@ fn parse_unwind(tcx: TyCtxt, attr: &Attribute) -> Option<u32> {
...
     };
     attributes
         .iter()
-        .filter_map(|attr| match parse_paths(attr) {
-            Ok(paths) => match paths.as_slice() {
+        .filter_map(|attr| {
+            let paths = parse_paths(attr).unwrap_or_else(|_| {
+                tcx.dcx().span_err(
+                    attr.span,
+                    "attribute `kani::stub` takes two path arguments; found argument that is not \
```
zhassan-aws, line 817: Use `KaniAttributeKind::Stub` instead of `kani::stub`?

zhassan-aws (approved):
LGTM

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -989,3 +991,54 @@ pub fn matches_diagnostic<T: CrateDef>(tcx: TyCtxt, def: T, attr_name: &str) ->
...
+    fn segments_str<'a, I>(segments: I) -> String
+    where
+        I: IntoIterator<Item = &'a PathSegment>,
+    {
+        // We don't bother with path arguments for now since users shouldn't provide them.
+        segments
+            .into_iter()
+            .map(|segment| segment.to_token_stream().to_string())
+            .intersperse("::".to_string())
+            .collect()
```
feliperodri, line 1028: Wouldn't be better to use a String buffer here? We can replace `intersperse` with a `peekable` iterator to avoid creating an intermediate collection

feliperodri: Maybe the way you choose is more readable since we won't need loops. Is that why?

kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -78,25 +143,93 @@ pub fn expect_resolve_fn<T: CrateDef>(
...
+            unreachable!()
+        }
+    }
+}
+
+/// Checks if a Path segment represents a primitive
+fn is_primitive(ident: &Ident) -> bool {
+    let token = ident.to_string();
+    let Ok(typ) = syn::parse_str(&token) else { return false };
+    #[warn(non_exhaustive_omitted_patterns)]
```
feliperodri, line 191: is the list of patterns too long?

kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -445,3 +583,25 @@ fn resolve_in_type<'tcx>(
...
+    debug!(?name, ?trait_id, "resolve_in_trait");
+    let missing_item_err =
+        || ResolveError::MissingItem { tcx, base: trait_id, unresolved: name.to_string() };
+    let trait_def = tcx.trait_def(trait_id);
+    // Try the inherent `impl` blocks (i.e., non-trait `impl`s).
+    tcx.associated_item_def_ids(trait_def.def_id)
+        .iter()
+        .copied()
+        .find(|item| {
+            let item_path = tcx.def_path_str(*item);
```
feliperodri, line 602: Why not `tcx.item_name(item)` to avoid string operations?

tests/ui/function-stubbing-error/unsupported_resolutions.expected
```diff
@@ -0,0 +1,8 @@
+error: failed to resolve `<[char ; 10]>::foo`: Kani currently cannot resolve path including primitive types
```
tautschnig, line 1: Somewhat random place for comment: do I understand correctly that no test is being added that demonstrates that the feature works? As far as I can tell, we only have tests demonstrating that name resolution fails?!

kani-compiler/src/kani_middle/mod.rs
```diff
@@ -225,10 +225,16 @@ fn find_fn_def(tcx: TyCtxt, diagnostic: &str) -> Option<FnDef> {
...
+
+/// Try to convert a internal DefId to a FnDef.
+pub fn stable_fn_def(tcx: TyCtxt, def_id: InternalDefId) -> Option<FnDef> {
+    if let TyKind::RigidTy(RigidTy::FnDef(def, _)) =
+        rustc_internal::stable(tcx.type_of(def_id)).value.kind()
+    {
+        Some(def)
+    } else {
+        None
+    }
```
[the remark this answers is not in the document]
celinval: I don't think so. I think the first is more common in Rust code.

kani-compiler/src/kani_middle/attributes.rs
celinval: I thought it was more readable, and this code is only invoked for error messages, which shouldn't occur a lot, thus performance is not a concern. Let me know if you want me to make an explicit loop.

kani-compiler/src/kani_middle/resolve.rs
celinval: The problem here is that this enum is marked as `non_exhaustive`, which means you always need to use a default arm, and there is no compilation error if a new variant is added. This attribute makes sure that all variants are covered despite the default arm. See [this](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) for more details.

tests/ui/function-stubbing-error/unsupported_resolutions.expected
celinval: That's exactly what this PR is doing. This is not yet adding a new feature or a new fix, hence only the failure case. This PR is the first step to implement the feature. In this PR, we are now able to correctly parse the user input for fully qualified path and primitive types, and distinguish between them. We still need to create a follow up PR to handle each case identified here.

kani-compiler/src/kani_middle/resolve.rs
celinval: Because `def_path_str` returns a string suitable for user output.

tests/ui/function-stubbing-error/unsupported_resolutions.expected
```diff
@@ -0,0 +1,8 @@
...
+error: failed to resolve `<& [u32]>::foo`: Kani currently cannot resolve path including primitive types
+error: failed to resolve `<& [u32] as Foo>::foo`: Kani currently cannot resolve path including primitive types
+error: failed to resolve `<(i32 , i32) as Foo>::foo`: Kani currently cannot resolve path including primitive types
+error: failed to resolve `u8::foo`: Kani currently cannot resolve path including primitive types
+error: failed to resolve `<Bar as Foo>::bar`: unable to find `bar` inside trait `Foo`
+error: Kani currently does not support stubbing trait implementations
+error: failed to resolve `<Bar>::foo`: Kani currently cannot resolve qualified bare function paths
```
celinval, line 8:
```suggestion
error: failed to resolve `<Bar>::foo`: Kani currently cannot resolve qualified bare function paths

```

tests/ui/function-stubbing-error/unsupported_resolutions.expected
tautschnig: The PR name says "Extend name resolution to support qualified paths," so I was assuming that we can now successfully resolve some (qualified) paths that we weren't able to resolve before?

celinval: Yes. We can resolve the names now. But we cannot use the resolution for stubbing or contracts yet.

kani-compiler/src/kani_middle/attributes.rs
feliperodri: Not a problem! If performance is not a concern, we should focus on readability and the way it is is fine.

feliperodri (approved):

tests/ui/function-stubbing-error/unsupported_resolutions.expected
celinval: I edited the name of this PR so it is explicit that this is a partial fix. Sorry for the confusion.

PR #3461, opened 2024-08-26
Update deny action

zhassan-aws:
The current `cargo deny` configuration in `deny.toml` uses several keys that have been deprecated. This PR removes the deprecated keys, and updates the deny action to use v2 (as well as renames it from `audit.yml` to `deny.yml`).

The only semantic difference is that `cargo deny` will now reject crates that are maintained or have a notice on them, whereas previously, our configuration set both to "warn". As mentioned in the docs though, one can add an "ignore" if needed to bypass those advisories:

https://embarkstudios.github.io/cargo-deny/checks/advisories/cfg.html#the-version-field-optional

### DIFF
--- a/.github/workflows/deny.yml
+++ b/.github/workflows/deny.yml
@@ -4,7 +4,7 @@
 # 1. Checks licenses for allowed license.
 # 2. Checks Rust-Sec registry for security advisories.
 
-name: Cargo Audit
+name: Cargo Deny
 on:
   pull_request:
   merge_group:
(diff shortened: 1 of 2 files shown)


jaisnan (approved):
Thanks! Any way to test this change?

zhassan-aws:
Yes, it passed in CI: https://github.com/model-checking/kani/actions/runs/10568647124/job/29280011523?pr=3461#step:4:1

zhassan-aws:
I also tested locally that the cargo deny check fails if a dependency uses a disallowed license.

PR #3462, opened 2024-08-27
Clarify comment in RFC Template

carolynzech:
Looking at the RFC template file, I thought that the recommendation to leave the software design section empty referred to everything below the comment, rather than just that section. This PR updates the comment.


jaisnan (approved):
Thanks!

Issue #2573, opened 2023-06-27
RFC: Output - Harness listing/enumeration


adpaco-aws:
Requested feature: A section of our output dedicated to enumerating or listing harnesses found in a project, how many will run for verification, etc. This is likely a requirement for #2563 , which will be an attempt to represent multi-harness progress.
Use case: Multi-harness verification - Nowadays users don't get any feedback about which harnesses are run or in which order.
Link to relevant documentation (Rust reference, Nomicon, RFC): #2563

celinval:
I think this is related to #1612. I think we should add a list option / subcommand. I would also suggest that we list all the contracts in Kani

PR #3463, opened 2024-08-27
RFC for List Subcommand

carolynzech:
RFC for the `kani list` subcommand.

I saw the comment [in the template](https://github.com/model-checking/kani/blob/main/rfc/src/template.md#software-design) to leave the Software Design section empty, but I was looking at the raw MD file and made the mistake of thinking it referred to all sections below the comment. In other words, I thought when it said "We recommend you to leave this empty for the first version of your RFC" it meant that everything below the comment should be empty (so Rationale, Open Questions, and Future Work) and not Software Design. I realized my mistake when I was making this PR, but I figured I may as well leave it since I already wrote it. I [opened a PR](https://github.com/model-checking/kani/pull/3462) to update the comment.

Resolves [#2573](https://github.com/model-checking/kani/issues/2573), [#1612](#https://github.com/model-checking/kani/issues/1612).


rfc/src/rfcs/0011-list.md
```diff
@@ -0,0 +1,157 @@
...
+    target: mock_random
+}
+```
+and so on for other attributes.
+
+## Software Design
+
+We will add a new subcommand to `kani-driver` which follows the normal verification workflow of invoking `kani-compiler` and constructing a `Project` from the result. Then, instead of verifying the project, Kani will iterate over the `KaniMetadata` in the `Project` to construct the output.
+
+### Contracts
```
celinval, line 87: Is this really necessary? I would prefer if we keep it simple, and leave this as an open question for now. I'm not sure we want the `list` sub-command to print all the attributes as they are.

rfc/src/rfcs/0011-list.md
```diff
@@ -0,0 +1,157 @@
...
+├── mod implementation
+│   └── mod verify
+```
+
+### Calculating Totals
+Since we already have the information about each harness, contract, etc., calculating totals is as simple as taking the length of those data structures.
+
+## Rationale and alternatives
+
+We organize the output by module so that users can compare Kani integration across their package. The output can be quite long, since Kani prints the whole module tree. As an alternative, we could support a more compact output format, e.g. one that just maps the Rust path (`example::verify::foo`) to the metadata. This format would have the advantage of being more concise, but the disadvantage of making it harder for the user to visualize how Kani fits into their package structure.
```
celinval, line 136: I actually prefer the more compact view. It's simpler and concise. Since we are sorting by the function name, it will naturally sort them by their modules.

For the json output, we could add a field `module` and I would also suggest adding the file location.

rfc/src/rfcs/0011-list.md
```diff
@@ -0,0 +1,157 @@
...
+
+Then, in `kani-compiler`, we add a new branch of the code generation process for the `list` subcommand. When `codegen_crate` encounters this reachability type, it will:
+- Construct a `CodegenUnits` object.
+- Call a new `CodegenUnits` method that filters local crate instances to those with the `contract_text` attribute, then stores the following in a new `contracted_functions` field:
+    - the type of contract (`Requires`, `Ensures`, or `Modifies`), 
+    - the text of the contract, and 
+    - the fully qualified path to it.
+- Call `write_metadata` on the `CodegenUnits` object, which will construct a `KaniMetadata` object with a new field containing the contracted function information, then write it to a file.
+
+### Constructing the Tree
```
celinval, line 117: I think using a BTreeMap with the fully qualified name of the function would suffice, no?

celinval (commented):
I would recommend to simplify the human output and the implementation. For the `json` format, we should include a more detailed output, including the function location (file, line, column) to help users / VSCode extension to retrieve any extra information.

BTW, I haven't seen any information on the json format / versioning. Can you please include this here? Thanks!

rfc/src/rfcs/0011-list.md
```diff
@@ -0,0 +1,157 @@
...
+- **Feature Request Issue:** [#2573](https://github.com/model-checking/kani/issues/2573), [#1612](https://github.com/model-checking/kani/issues/1612)
+- **RFC PR:** *Link to original PR*
+- **Status:** Under Review
+- **Version:** 0
+
+-------------------
+
+## Summary
+
+Add a subcommand `kani list` that lists the locations of Kani attributes in a project.
```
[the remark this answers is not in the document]
carolynzech: Went for slightly different wording:

> Add a subcommand `kani list` that, for each crate under verification, lists information relevant to its verification.

rfc/src/rfcs/0011-list.md
```diff
@@ -0,0 +1,157 @@
...
+
+Internally, this feature will be useful for tracking our customers' use of Kani and our progress with standard library verification. Externally, users can leverage this feature to get a high-level view of which areas of their projects have Kani attributes (and, by extension, which areas are still in need of verification). This feature can act as a complement to our existing line-by-line coverage feature--one can imagine, for example, using this feature to *informally* identify which areas seem to have few to no harnesses, then using the coverage feature to *formally* identify which areas remain unverified.
+
+This feature will not cause any regressions for exisiting users.
+
+## User Experience
+
+Users run a `kani list` subcommand, which prints metadata about the harnesses, contracts, and stubs in each module.
+For `cargo kani`, this will be every module in the package; for `kani`, it is every module in the crate.
```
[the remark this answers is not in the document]
carolynzech: No, I just thought that it would be a useful way of looking at it. Not married to it.

carolynzech:
Thanks for the feedback @celinval! Addressed your feedback.

rfc/src/rfcs/0012-list.md
```diff
@@ -0,0 +1,174 @@
...
+As the name implies, the goal of the `human` output is to be friendly for human readers. If the user wants an output format that's more easily parsed by a script, they can use the `json` option. 
+
+The JSON format will contain the same information as the human format, with the addition of file paths and JSON version. The JSON version number will start at zero and increment whenever we make an update to the format. This way, any users relying on this format for their scripts can realize that changes have occurred and update their logic accordingly.
+
+For example:
+
+```json
+{
+    kani-version: 0.5.4,
+    json-version: 0,
```
celinval, line 80: maybe just `version` or `file-version`. `json-version` sounds like the `json` version.

rfc/src/rfcs/0012-list.md
```diff
@@ -0,0 +1,174 @@
...
+3. Which contracts are verified, and by which harnesses?
+4. How many harnesses and contracts are there?
+
+We believe these questions are the most important for our use cases of tracking verification progress for customers and the standard library. The UX is designed to answer these questions clearly and concisely.
+
+We could have a more verbose or granular output, e.g., printing the metadata on a per-crate or per-module level, or including stubs or other attributes. Such a design would have the benefit of providing more information, with the disadvantage of being more complex to implement and more information for the user to process.
+
+If we do not implement this feature, users will have to obtain this metadata through manual searching, or by writing a script to do it themselves. This feature will improve our internal productivity by automating the process.
+
+## Open questions
```
celinval, line 156: I wonder if we also want to add an option to filter things by name. Example:

```
cargo kani list --harnesses "my_crate::my_module::*"
```

celinval (approved):
Looking good! Just minor comments, and I'm happy to adjust as we implement this. Thanks!

jaisnan (commented):
I would say, the RFC is missing one small section on the schema/fields that will be presented to the user via "human"/"pretty" or "json". I can see through the examples what fields they show but its not clear if that's an exhaustive list or if some of these fields are optional etc. 

If possible, can you add a section that is essentially formalizing the fields that a user can expect? Maybe something like this with JSON types for the JSON and the json version used as well

```
### Output Schema
1. Kani Version: Int
2. Standard Harnesses: List[] | None.
   2.1 function: string
   2.2 file: string
4. Contract Harnesses: List[] | None
 ...
```

PR #3464, opened 2024-08-27
Adjust test patterns so as not to check for trivial properties

tautschnig:
With diffblue/cbmc#8413, CBMC will no longer create property checks for assigns clauses that are trivially true.

We had CBMC-Nightly failing since August 17th given the CBMC change. This will bring CBMC-Nightly back to a passing state.


carolynzech (approved):
Thanks!

PR #3465, opened 2024-08-27
Cross-function union instrumentation

artemagvanian:
This PR introduces support for memory initialization checks for unions passed across the function boundary.

Whenever a union is passed as an argument, we need to make sure that its initialization state is preserved. Unlike pointers, unions do not have a stable memory address which could identify them in shadow memory. Hence, we need to pass extra information across function boundary since unions are passed “by value”. 

We introduce a global variable to store the previous address of unions passed as function arguments, which allows us to effectively tie the initialization state of unions passed between functions. This struct is written to by the caller and read from by the callee. 

For more information about planned functionality, see https://github.com/model-checking/kani/issues/3300


library/kani/src/mem_init.rs
```diff
@@ -115,6 +115,23 @@ impl MemoryInitializationState {
                 self.tracked_offset += to_offset - from_offset;
                 // Note that this preserves the value.
             }
+        } else if self.tracked_object_id == to_obj
+            && self.tracked_offset >= to_offset
+            && self.tracked_offset < to_offset + num_elts * LAYOUT_SIZE
+        {
+            self.value = true;
+        }
```
artemagvanian, line 123: If we are copying to an object that is currently tracked, need to unconditionally initialize the value, since the actual `copy` execution check will happen in a different non-deterministic branch.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -532,6 +548,59 @@ impl<'a, 'tcx> UninitInstrumenter<'a, 'tcx> {
         );
     }
 
+    /// Construct a piece of instrumentation storing information about unions passed as arguments.
```
celinval, line 551:
```suggestion
    /// Instrument the code to pass information about arguments containing unions.
```

library/kani/src/mem_init.rs
```diff
@@ -331,3 +347,55 @@ fn copy_init_state<const LAYOUT_SIZE: usize, T>(from: *const T, to: *const T, nu
...
+#[rustc_diagnostic_item = "KaniResetArgumentBuffer"]
+#[kanitool::disable_checks(pointer)]
+fn reset_argument_buffer() {
+    unsafe { ARGUMENT_BUFFER = None }
+}
+
+#[rustc_diagnostic_item = "KaniStoreArgument"]
+#[kanitool::disable_checks(pointer)]
+fn store_argument<const LAYOUT_SIZE: usize, T>(from: *const T, selected_argument: usize) {
+    if LAYOUT_SIZE == 0 {
```
celinval, line 368: Can we remove this? I'm assuming your transformation should've taken care of this case already, and in the worst case scenario, this shouldn't impact the code correctness.

tests/expected/uninit/fixme_unions.rs
```diff
@@ -48,14 +47,59 @@ unsafe fn union_as_subfields_should_pass() {
...
 
-/// Tests unions composing with other unions.
+/// Tests unions composing with other unions and reading non-padding data.
 #[kani::proof]
 unsafe fn uber_union_should_pass() {
     let u = Outer { u: U { b: 0 } }; // `u` is initialized for 4 bytes.
     let non_padding = u.a; // Read 4 bytes from `u`.
 }
+
+/// Tests unions composing with other unions and reading padding data.
```
celinval, line 72: This is an `expected` test, right? Don't we need to add an expectation that will fail if we fix this issue?

I would also prefer having multiple fixme tests, one per limitation.

celinval (commented):
Thanks Artem! It's looking good.

kani-compiler/src/kani_middle/transform/check_uninit/ptr_uninit/uninit_visitor.rs
```diff
@@ -38,11 +38,38 @@ impl TargetFinder for CheckUninitVisitor {
...
+                }
             };
+            if bb_idx == 0 {
+                let union_args: Vec<_> = body
+                    .locals()
+                    .iter()
+                    .enumerate()
+                    .skip(1)
+                    .take(body.arg_count())
+                    .filter(|(_, local)| local.ty.kind().is_union())
```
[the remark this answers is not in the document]
artemagvanian: Yes, ideally. However, we do not yet support types containing unions. This is currently mitigated at the union creation site.

kani-compiler/src/kani_middle/transform/check_uninit/ptr_uninit/uninit_visitor.rs
```diff
@@ -127,9 +170,10 @@ impl MirVisitor for CheckUninitVisitor {
                 // if a union as a subfield is detected, `assert!(false)` will be injected from
                 // the type layout code.
                 let is_inside_union = {
-                    let mut contains_union = false;
                     let mut place_to_add_projections =
                         Place { local: place.local, projection: vec![] };
+                    let mut contains_union =
+                        place_to_add_projections.ty(&self.locals).unwrap().kind().is_union();
```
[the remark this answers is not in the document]
artemagvanian: No, it wasn't. The change happened because I moved the union checks below into the if statement block as a part of adding mitigation.

kani-compiler/src/kani_middle/transform/check_uninit/mod.rs
```diff
@@ -592,6 +661,25 @@ impl<'a, 'tcx> UninitInstrumenter<'a, 'tcx> {
         let result = body.insert_assignment(rvalue, source, position);
         body.insert_check(tcx, &self.check_type, source, position, result, reason);
     }
+
+    fn inject_argument_buffer_reset(
```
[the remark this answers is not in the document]
artemagvanian: Yes, we actually can! That's a good idea, thanks!

tests/expected/uninit/fixme_unions.rs
artemagvanian: I wasn't sure what an appropriate failing behavior would be, so I left it as a TODO comment for now.

library/kani_core/src/mem_init.rs
```diff
@@ -119,6 +127,21 @@ macro_rules! kani_mem_init {
                         self.tracked_offset += to_offset - from_offset;
                         // Note that this preserves the value.
                     }
+                } else {
+                    self.bless::<LAYOUT_SIZE>(to_ptr, 1);
+                }
+            }
+
+            #[kanitool::disable_checks(pointer)]
+            pub fn bless<const LAYOUT_SIZE: usize>(&mut self, ptr: *const u8, num_elts: usize) {
```
celinval, line 136: Please add comment. This seems to be assuming that all bytes are initialized, is that the case?

library/kani_core/src/mem_init.rs
```diff
@@ -343,5 +362,54 @@ macro_rules! kani_mem_init {
...
+
+        /// Load information from the argument buffer (if the argument position matches) via copying
+        /// the memory initialization information from an address in the caller to an address in the
+        /// callee. Otherwise, mark that the argument as initialized, as it will be checked by
+        /// another non-deterministic branch. Reset the argument buffer after loading from it.
+        #[kanitool::disable_checks(pointer)]
+        #[rustc_diagnostic_item = "KaniLoadArgument"]
+        fn load_argument<const LAYOUT_SIZE: usize, T>(to: *const T, selected_argument: usize) {
+            let (to_ptr, _) = to.to_raw_parts();
+            unsafe {
```
celinval, line 402: since this function is disabling pointer checks, it would be very nice if we could minimize the usage of `unsafe`.

celinval (approved):
Thanks @artemagvanian!

library/kani_core/src/mem_init.rs
artemagvanian: Yes, correct. Will do!

Issue #3029, opened 2024-02-14
crates.io stub and docs.rs documentation


GrigorenkoPV:
I really wish it was possible to view documentation for `kani`-the-lib, the one that provides all the `any`, `assume`, and `proof` stuff. However, since the kani crate on crates.io is a stub, docs.rs doesn't have any useful documentation.

So now if I want to read the docs, I have to
```shell
cd $KANI_HOME/kani-0.46.0/library/kani
cargo +nightly doc
```
which is not that convinient.

Could you please consider publishing an actual version of the lib to crates.io? If the reason for having it as a stub is to prevent users from downloading the wrong thing, maybe adding something like
```rust
#[cfg(not(docs)]
compile_error!("This is an internal crate. You were probably looking for `kani-verifier` instead.");
```
to the published version could be used?

zhassan-aws:
Hi @GrigorenkoPV. Thanks for opening the issue. The documentation for the Kani library is currently available here:

https://model-checking.github.io/kani/crates/doc/kani/

This can be accessed from the Kani documentation page:

https://model-checking.github.io/kani/crates/index.html

We plan to publish the documentation on crates.io once we publish the `kani` crate.

GrigorenkoPV:
Thank you for the response!

> The documentation for the Kani library is currently available here:
> https://model-checking.github.io/kani/crates/doc/kani/

> We plan to publish the documentation on crates.io once we publish the kani crate.

Would it be feasible to update the stub in the meantime to feature a link to where the actual documentation can be found?

celinval:
That's a good idea.

PR #3474, opened 2024-08-30
Update Kani Book

carolynzech:
Add the following to the book:

- **Reference section on contracts**. The point of this section is not to be an exhaustive explanation--we have good documentation on contracts already, and we don't want to maintain it in two places. The goal of the section is to give readers an intuition for the main idea and then direct them to more detailed resources.
- **Add a reference to the blog**--it's a good resource to learn about Kani and it would be nice to have an easier way to find it. Inspired by [this post](https://users.rust-lang.org/t/can-somebody-explain-how-kani-verifier-works/113918).
- **Move the crates documentation earlier.** This partly addresses [#3029](https://github.com/model-checking/kani/issues/3029) by at least making our crates documentation more visible in the book. (I don't have this PR as resolving that issue, since we would need to actually publish the docs on `crates.io` to do that.)


docs/src/reference/experimental/contracts.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+Since `gcd` performs `max % min` (and perhaps swaps those values), passing zero as an argument could cause a division by zero. The `requires` contract tells Kani to restrict the range of nondeterministic inputs to nonzero ones so that we don't run into this error. The `ensures` contract is what actually checks that the result is a correct divisor for the inputs. (The `recursion` attribute is required when using contracts on recursive functions).
+
+Then, we would write a harness to *verify* those contracts, like so:[^1]
+
+```rust
+#[kani::proof_for_contract(gcd)]
+fn check_gcd() {
+    let mut max: u64 = kani::any();
```
zhassan-aws, line 33: The `mut` on this line and the next is not needed.

docs/src/reference/experimental/contracts.md
```diff
@@ -0,0 +1,58 @@
...
+fn check_foo() {
+    let x: usize = kani::any();
+    foo(x);
+}
+```
+By leveraging the stubbing feature, we can replace the (expensive) `gcd` call with a *verified abstraction* of its behavior, greatly reducing verification time for `foo`.
+
+There is far more information to learn about contracts. We highly recommend reading our [blog post about contracts](https://model-checking.github.io/kani-verifier-blog/2024/01/29/function-contracts.html) (from which this `gcd` example is taken). We also recommend looking at the `contracts` module in our [documentation](../../crates/index.md).
+
+[^1]: (If you are attempting to run this example locally, note you may run into performance issues with unconstrained `u64` inputs. If that is the case, try using `kani::assume` to restrict the inputs to a smaller range.)
```
zhassan-aws, line 58: Does it work with `u32`? If so, I suggest we switch to `u32` so that at least the example in the tutorial works out of the box.

carolynzech: I tried `u32` and `u16` and cut them off after 2 minutes; I figured a reader won't want to wait for much longer than that. `u8` finished in 15 seconds, so went with that.

zhassan-aws (approved):
Thank you!

Issue #3022, opened 2024-02-12
ice: unequal types in stable mir


matthiaskrgr:
I tried this code:

```rust
type BuiltIn = for<'a> fn(&str);

struct Function {
    inner: BuiltIn,
}

impl Function {
    fn new(subr: BuiltIn) -> Self {
        Self { inner: subr }
    }
}

fn dummy(_: &str) {}

#[kani::proof]
fn main() {
    let func1 = Function::new(dummy);
    let func2 = Function::new(dummy);
    let inner: fn(&'static _) -> _ = func1.inner;
    assert!(inner == func2.inner);
}

```

using the following command line invocation:

```
kani file.rs 
```

with Kani version: 0.46.0

I expected to see this happen: *explanation*

Instead, this happened: *explanation*
```
Kani Rust Verifier 0.46.0 (standalone)
thread 'rustc' panicked at compiler/stable_mir/src/mir/body.rs:384:17:
assertion `left == right` failed
  left: Ty { id: 5, kind: RigidTy(FnPtr(Binder { value: FnSig { inputs_and_output: [Ty { id: 4, kind: RigidTy(Ref(Region { kind: ReErased }, Ty { id: 10, kind: RigidTy(Str) }, Not)) }, Ty { id: 6, kind: RigidTy(Tuple([])) }], c_variadic: false, unsafety: Normal, abi: Rust }, bound_vars: [] })) }
 right: Ty { id: 2, kind: RigidTy(FnPtr(Binder { value: FnSig { inputs_and_output: [Ty { id: 9, kind: RigidTy(Ref(Region { kind: ReBound(0, BoundRegion { var: 0, kind: BrAnon }) }, Ty { id: 10, kind: RigidTy(Str) }, Not)) }, Ty { id: 6, kind: RigidTy(Tuple([])) }], c_variadic: false, unsafety: Normal, abi: Rust }, bound_vars: [Region(BrAnon)] })) }
stack backtrace:
   0:     0x7f3644d8be86 - std::backtrace_rs::backtrace::libunwind::trace::haa62de98ce20d13c
                               at /rustc/7ffc697ce10f19447c0ce338428ae4b9bc0c041c/library/std/src/../../backtrace/src/backtrace/libunwind.rs:104:5
   1:     0x7f3644d8be86 - std::backtrace_rs::backtrace::trace_unsynchronized::h4bc7f582e9f49dbd
[...]

PR #3483, opened 2024-09-03
Add test for issue 3022

carolynzech:
#3022 no longer happens in Kani v0.5.4. Add a test with the code from this issue.

I figured this was worth doing since it's a relatively new issue and this way we can make sure we don't regress.


carolynzech:
Closing in favor of #3484

PR #3490, opened 2024-09-04
List RFC revisions

carolynzech:
Addresses @jaisnan's feedback on list RFC--change `human` output to `pretty` and fix Kani version typo.


rfc/src/rfcs/0012-list.md
```diff
@@ -132,6 +152,30 @@ For example:
 }
 ```
 
+#### Output Schema
+The formal output schema is:
+```
+Kani Version: float
+File Version: int
```
jaisnan, line 159: Nit: I'm assuming you mean JSON version?

rfc/src/rfcs/0012-list.md
```diff
@@ -132,6 +152,30 @@ For example:
...
+    - file: string
+    - harnesses: List[string]
+Totals: Object
+    - standard-harnesses: int
+    - contract-harnesses: int
+    - functions-with-contracts: int
+    - contracts: int
+```
+
+All sections will be present in the output, regardless of the result.
```
jaisnan, line 177: What happens if there are no contracts, do we just present an empty list?

jaisnan (commented):
Almost there! Just a few small points left to address

rfc/src/rfcs/0012-list.md
carolynzech: No, see https://github.com/model-checking/kani/pull/3463#discussion_r1739455268.

rfc/src/rfcs/0012-list.md
carolynzech: Yes, the output schema just has a `List` (without None) since the list can just be empty. The example has a case of this. Can add text to make that explicit if it's unclear.

jaisnan: Yes, would be better to add to the RFC to make it explicit

rfc/src/rfcs/0012-list.md
jaisnan: In that case, can you clarify in the RFC what version/format of JSON are you using in your output?

celinval: Are you suggesting that we also include the JSON format?

jaisnan: Yes, but since we anticipate this to be a mostly constant value, I just thought we should document this in the RFC itself, and then we dont have to include the jSON version in the JSON response object itself.

rfc/src/rfcs/0012-list.md
```diff
@@ -66,17 +66,37 @@ Totals:
 
 A "Standard Harness" is a `#[proof]` harness, while a "Contract Harness" is a `#[proof_for_contract]` harness.
 
+#### Output Schema
+The formal output schema is:
+```
+Kani Version: float
```
celinval, line 72: We already print something like:
```
Kani Rust Verifier 0.54.0 ([cargo plugin|standalone]) 
```

rfc/src/rfcs/0012-list.md
```diff
@@ -66,17 +66,37 @@ Totals:
 
 A "Standard Harness" is a `#[proof]` harness, while a "Contract Harness" is a `#[proof_for_contract]` harness.
 
+#### Output Schema
```
celinval, line 69: I don't think we should / need to define a formal output schema for the `pretty` output in here. This will likely change as we implement this feature and get feedback from users.

rfc/src/rfcs/0012-list.md
```diff
@@ -66,17 +66,37 @@ Totals:
...
+
+All sections will be present in the output, regardless of the result. If a list is empty, Kani will output a `NONE` string.
+
 ### JSON Format
 
-As the name implies, the goal of the `human` output is to be friendly for human readers. If the user wants an output format that's more easily parsed by a script, they can use the `json` option. 
+If the user wants an output format that's more easily parsed by a script, they can use the `json` option. 
 
-The JSON format will contain the same information as the human format, with the addition of file paths and file version. The file version will start at zero and increment whenever we make an update to the format. This way, any users relying on this format for their scripts can realize that changes have occurred and update their logic accordingly.
+The JSON format will contain the same information as the pretty format, with the addition of file paths and file version. The file version will start at zero and increment whenever we make an update to the format. This way, any users relying on this format for their scripts can realize that changes have occurred and update their logic accordingly.
```
celinval, line 93: Should we follow semantic versioning here too? In this case, we probably want to start with `0.1`.

rfc/src/rfcs/0012-list.md
```diff
@@ -132,6 +152,30 @@ For example:
 }
 ```
 
+#### Output Schema
+The formal output schema is:
+```
+Kani Version: float
```
celinval, line 158: This is not a float though. It's a String that represents semantic versioning (https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions)

rfc/src/rfcs/0012-list.md
celinval, line 159: I think this should also be a String representing semver format.

rfc/src/rfcs/0012-list.md
```diff
@@ -132,6 +152,30 @@ For example:
 }
 ```
 
+#### Output Schema
```
celinval, line 155: I also would prefer adding this in the documentation rather than the RFC since it will evolve. Otherwise, just mention that this will be the initial version.

celinval (commented):
Just a few comments

rfc/src/rfcs/0012-list.md
carolynzech: Yes, that's what I meant, but I should have made it match precisely--will fix.

rfc/src/rfcs/0012-list.md
carolynzech: I feel like if the JSON version subject to change in future versions of Kani there's not much point putting it in the RFC? I'd prefer to only put information in the RFC that we expect to be (relatively) stable--we'll almost certainly update the JSON version over time.

rfc/src/rfcs/0012-list.md
carolynzech: I agree documentation is better--going to delete both output schemas.

jaisnan (approved):
Approving under the assumption that the schemas will be added to documentation sooner rather than later 👍

carolynzech:
> Approving under the assumption that the schemas will be added to documentation sooner rather than later 👍

Yes, I'll make sure that the first release of this has schema documentation.

celinval (approved):

rfc/src/rfcs/0013-list.md
celinval, line 1: Can you please bump the version to `1`

PR #3491, opened 2024-09-04
Fix contract expansion for `old`

### LINKED ISSUE #3359: The `old` construct does not respect the function requirements
I tried this code:
```rust
#[kani::requires(val < i32::MAX)]
#[kani::ensures(|result| *result == old(val + 1))]
pub fn next(mut val: i32) -> i32 {
    val + 1
}

#[kani::proof_for_contract(next)]
pub fn check_next() {
    // Placing the restriction before calling makes the problem go away.
    // let _ = next(kani::any_where(|val: &i32| *val < i32::MAX));
    let _ = next(kani::any());
}
```

```
kani -Z function-contracts next.rs
```

with Kani version:

I expected to see this happen: Verification succeeded

Instead, this happened: Verification failed due to overflow:
```
Failed Checks: attempt to add with overflow
 File: "next.rs", line 2, in next_check_53aede
```

carolynzech:
Fixes the macro expansion for contracts to properly place history expressions.

## Problem
Before this PR, instantiations of "remembers variables" (i.e., the variables that save the value before the function executes) were always put *above* any statements from previous macro expansions. For example, for this code (from #3359):

```rust
#[kani::requires(val < i32::MAX)]
#[kani::ensures(|result| *result == old(val + 1))]
pub fn next(mut val: i32) -> i32 {
    val + 1
}
```
Kani would first expand the `requires` attribute and insert `kani::assume(val < i32::MAX)`. The expansion of `ensures` would then put the remembers variables first, generating this:

```
let remember_kani_internal_1e725538cd5566b8 = val + 1;
kani::assume(val < i32::MAX);
```
which causes an integer overflow because we don't restrict the value of `val` before adding 1. Instead, we want:

```
kani::assume(val < i32::MAX);
let remember_kani_internal_1e725538cd5566b8 = val + 1;
```

## Solution
The solution is to insert the remembers variables immediately after preconditions--that way, they respect the preconditions but are still declared before the function under contract executes. When we're expanding an `ensures` clause, we iterate through each of the already-generated statements, find the position where the preconditions end, then insert the remembers variables there. For instance:

```
kani::assume(x < 100);
kani::assume(y < 10);
kani::assume(x + y < 105);
<-- remembers variables go here -->
let _wrapper_arg = ...
```
---

Resolves #3359

### DIFF
--- a/library/kani_macros/src/sysroot/contracts/check.rs
+++ b/library/kani_macros/src/sysroot/contracts/check.rs
@@ -9,8 +9,8 @@ use std::mem;
 use syn::{parse_quote, Block, Expr, FnArg, Local, LocalInit, Pat, PatIdent, ReturnType, Stmt};
 
 use super::{
-    helpers::*, shared::build_ensures, ContractConditionsData, ContractConditionsHandler,
-    INTERNAL_RESULT_IDENT,
+    helpers::*, shared::build_ensures, ClosureType, ContractConditionsData,
+    ContractConditionsHandler, INTERNAL_RESULT_IDENT,
 };
 
 const WRAPPER_ARG: &str = "_wrapper_arg";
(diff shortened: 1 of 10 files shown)


celinval (approved):
Thank you for fixing this!

PR #3496, opened 2024-09-05
Enable stubbing and function contracts for primitive types

### LINKED ISSUE #2646: [Feature Request] Ability to stub out function on primitive types (such as `[T]`)

celinval:
Previously, Kani was not able to resolve methods that belong to primitive types, which would not allow users to specify primitive types as stub target, neither for contracts.

Thus, extend the name resolution to be able to convert a type expression into a `Ty` for primitive types, and add a new method to search for function implementation for primitive types.

I also moved the type resolution logic to its own module to make it a bit cleaner.

Resolves #2646 


carolynzech (approved):
Nice! You may want to add tests for the other `PrimitiveIdent` variants (bool and floats) for the sake of being exhaustive, but if you're confident enough that the logic is the same then I'm good either way.

kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -594,14 +562,53 @@ fn resolve_in_trait<'tcx>(
...
-            let last = item_path.split("::").last().unwrap();
-            last == name
-        })
+        .find(|item| is_item_name(tcx, *item, name))
         .ok_or_else(missing_item_err)
 }
+
+/// Resolves a primitive type function.
+///
+/// This function assumes that the Simplified Type is a primitive.
```
celinval, line 575:
```suggestion
/// This function assumes that `ty` is a primitive.
```

PR #3511, opened 2024-09-09
Reduce object-bits for test to avoid OOM

zhassan-aws:
Kani often runs out of memory on `tests/expected/function-contract/history/stub.rs` when running the regressions. In particular, `quadruple_harness` consumes over 9 GB of memory. This PR reduces the object bits for this test to 8 to avoid OOM issues. This brings down memory usage to ~125 MB.

Before:
```bash
$ /usr/bin/time -v kani -Zfunction-contracts stub.rs --harness quadruple_harness
        Maximum resident set size (kbytes): 9136036
```

After:
```bash
$ /usr/bin/time -v kani -Zfunction-contracts stub.rs --harness quadruple_harness --enable-unstable --cbmc-args --object-bits 8
...
        Maximum resident set size (kbytes): 125172
```

### DIFF
--- a/tests/expected/function-contract/history/stub.rs
+++ b/tests/expected/function-contract/history/stub.rs
@@ -1,6 +1,8 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
-// kani-flags: -Zfunction-contracts
+// This test consumes > 9 GB of memory with 16 object bits. Reducing the number
+// of object bits to 8 to avoid running out of memory.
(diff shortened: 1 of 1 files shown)


feliperodri (approved):

celinval:
I wonder if we should tweak with the current default

zhassan-aws:
I tried tweaking the test, but eventually gave up. For example, switching from 32-bit integers to 16 or 8-bits does not decrease the memory usage much.

I'm open to suggestions.

PR #3513, opened 2024-09-11
Instrument validity checks for pointer to reference casts for slices and str's

### LINKED ISSUE #3498: Kani does not detect UB when generating an invalid slice reference from an invalid slice pointer
I tried this code:

```rust
//! Converting a raw pointer to a reference should trigger UB if the metadata is not valid.
#![feature(set_ptr_value)]

// Generate invalid fat pointer by incrementing the pointer address without adjusting the metadata.
#[kani::proof]
fn check_invalid_read() {
    let data = "hello";
    let ptr = data as *const str;
    // This should trigger UB since the metadata does not get adjusted.
    let val = unsafe { &*ptr.byte_add(1) };
    assert_eq!(val.len(), data.len());
}

// Generate invalid fat pointer by combining the metadata.
#[kani::proof]
fn check_with_metadata() {
    let short = [0u32; 2];
    let long = [0u32; 10];
    let ptr = &short as *const [u32];
    // This should trigger UB since the slice is not valid for the new length.
    let fake_long = unsafe { &*ptr.with_metadata_of(&long) };
    assert_eq!(fake_long.len(), long.len());
}
```

```
kani invalid_fat_reference.rs
```

with Kani version: 0.54.0-dev

I expected to see this happen: Both harnesses should fail due to UB. Note that both cases fail in MIRI.

zhassan-aws:
As pointed out in #3498, validity checks for pointer to reference casts (added in #3221) were not instrumented in the case of fat pointers (e.g. array and string slices). This PR extends the instrumentation of validity checks to handle those cases.

Resolves #3498 

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
@@ -23,9 +23,11 @@ use cbmc::goto_program::{Expr, Location, Stmt, Type};
 use cbmc::InternedString;
 use rustc_middle::mir::coverage::SourceRegion;
 use stable_mir::mir::{Place, ProjectionElem};
-use stable_mir::ty::{Span as SpanStable, TypeAndMut};
+use stable_mir::ty::{Span as SpanStable, Ty};
 use strum_macros::{AsRefStr, EnumString};
 use tracing::debug;
+
+use super::intrinsic::SizeAlign;
 
 /// Classifies the type of CBMC `assert`, as different assertions can have different semantics (e.g. cover)
 ///
@@ -333,6 +335,8 @@ impl<'tcx> GotocCtx<'tcx> {
     pub fn codegen_raw_ptr_deref_validity_check(
         &mut self,
         place: &Place,
+        place_ref: Expr,
+        place_ref_ty: Ty,
         loc: &Location,
     ) -> Option<Stmt> {
         if let Some(ProjectionElem::Deref) = place.projection.last() {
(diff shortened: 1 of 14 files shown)


tests/kani/Projection/slice_slice_projection.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-verify-fail
```
zhassan-aws, line 3: This test now fails with the new checks. I'm not clear on the purpose of the unsafe operation, so I'm not sure how to fix it.

CC @celinval

celinval: This test is basically trying to create an ADT that is a DST. I reviewed the test, and it is doing a few incorrect and very unsafe things, and your fix probably uncovered some of them.

I don't think we should add `// kani-verify-fail` here though. I actually think we should deprecate this option since it basically accepts any failure.

I think it's worth fixing this test to ensure we are still correctly testing projection code, and we should derive a new test that trigger the issues that we were missing before. Those issues are creating a reference to a misaligned value, and a reference with invalid size.

To fix the test at hand, I believe you need to change the MyStr to have alignment of `1` or use `ptr.offset()` to ensure the alignment is correct.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -346,41 +350,35 @@ impl<'tcx> GotocCtx<'tcx> {
             let ptr_place_ty = self.place_ty_stable(&ptr_place);
             if ptr_place_ty.kind().is_raw_ptr() {
                 // Extract the size of the pointee.
-                let pointee_size = {
-                    let TypeAndMut { ty: pointee_ty, .. } =
-                        ptr_place_ty.kind().builtin_deref(true).unwrap();
-                    let pointee_ty_layout = pointee_ty.layout().unwrap();
-                    pointee_ty_layout.shape().size.bytes()
-                };
+                let SizeAlign { size: sz, .. } =
```
celinval, line 353: we should also check alignment.

celinval (commented):
I pushed a fix for the test, but as I mentioned before. I think we could create a new test from the existing one that shows the new code catch those issues.

BTW, I think the code we added still does not check alignment.

tests/kani/Projection/slice_slice_projection.rs
celinval:
```suggestion
```

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
zhassan-aws: I checked, and it seems the alignment check is instrumented by the compiler when performing the dereference. For example, the following test:
```rust
#[repr(align(4))]
#[derive(Clone, Copy)]
struct AlignedI32(i32);

#[kani::proof]
fn f() {
    let data = AlignedI32(42);
    let ptr = &data as *const AlignedI32;
    
    let _ = unsafe { *ptr.byte_add(1) };
}
```
fails today with:
```

Check 3: f.assertion.1
         - Status: FAILURE
         - Description: "misaligned pointer dereference: address must be a multiple of its type's alignment"
         - Location: test.rs:10:22 in function f
```
Are there more alignment checks needed?

zhassan-aws:
>  I think we could create a new test from the existing one that shows the new code catch those issues.

Added as an expected test.

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
celinval: But not the casting though, does it?

zhassan-aws: Oh, I see. I'll look into checking alignment.

zhassan-aws: Added in https://github.com/model-checking/kani/pull/3513/commits/47811273c88689757db0f5a03b80518430b0a6c8.

PR #3514, opened 2024-09-14
Add an LLBC backend

zhassan-aws:
This PR adds a new codegen backend that generates low-level borrow calculus (LLBC), which is the format defined by [Charon](https://github.com/AeneasVerif/charon) and [Aeneas](https://github.com/AeneasVerif/aeneas).

The backend can be invoked using `-Zaeneas`, and will generate a `.llbc` file. This file can then be passed to Aeneas to generate Lean for example. Currently, Aeneas needs to be manually run on the generated LLBC.

The PR translates stable MIR to unstructured LLBC (ULLBC) and then uses the Charon library to apply transformations that reconstruct the control flow to regain some of the high-level structure (loops, if statements, etc.).

In this PR, very few MIR constructs are handled, but the PR is already pretty big. More support to come in subsequent PRs.

Call-outs:

- Charon is currently not released on crates.io, so the library is added as a dependency through GitHub with a particular commit.
- The Kani driver is currenty heavily tailored towards CBMC, so currently it's expecting a goto file, and thus errors out when the Aeneas backend is invoked.


tautschnig (approved):
Approving with the following understanding:
1. Some cleanup is needed as indicated by the comments.
2. The codegen for Aeneas does need further work, but this doesn't impact any existing user who continues to use the cprover back-end.

kani-compiler/src/args.rs
```diff
@@ -4,6 +4,20 @@
...
+pub enum BackendOption {
+    /// Aeneas (LLBC) backend
+    Aeneas,
+
+    /// CProver (Goto) backend
+    CProver,
+
+    /// Backend option was not explicitly set
+    #[default]
+    None,
```
celinval, line 18: Why none? Shouldn't we make the argument optional instead?

kani-compiler/src/args.rs
```diff
@@ -4,6 +4,20 @@
 use strum_macros::{AsRefStr, EnumString, VariantNames};
 use tracing_subscriber::filter::Directive;
 
+#[derive(Debug, Default, Clone, Copy, AsRefStr, EnumString, VariantNames, PartialEq, Eq)]
+#[strum(serialize_all = "snake_case")]
+pub enum BackendOption {
+    /// Aeneas (LLBC) backend
+    Aeneas,
```
celinval, line 11: This option should only be available if the compiler feature was enabled, right?

celinval (commented):
I haven't reviewed the backend code yet, but I'm curious about how this affects our released software. Are we shipping the new backend?

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
```diff
@@ -0,0 +1,770 @@
...
+            let mut parents = vec![def_id];
+            let mut cur_id = def_id;
+            while let Some(parent) = tcx.opt_parent(cur_id) {
+                parents.push(parent);
+                cur_id = parent;
+            }
+            parents.into_iter().rev().collect()
+        };
+
+        // Rk.: below we try to be as tight as possible with regards to sanity
```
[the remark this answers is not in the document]
Nadrieril: That's a comment copied from Charon, "Rk" is short for @R1kM I believe

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
```diff
@@ -0,0 +1,770 @@
...
+                    error_assert!(self, span, data.disambiguator == 0);
+
+                    // This should be the beginning of the path
+                    error_assert!(self, span, name.is_empty());
+                    found_crate_name = true;
+                    name.push(PathElem::Ident(crate_name.clone(), disambiguator));
+                }
+                DefPathData::Impl => todo!(),
+                DefPathData::OpaqueTy => {
+                    // TODO: do nothing for now
```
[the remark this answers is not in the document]
Nadrieril: This is copied over from Charon, I think the answer was "whenever we understand why we'd need the name of an opaque type". We could probably make this an error? I haven't tested if this can ever occur with the current code.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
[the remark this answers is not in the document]
R1kM: I believe it stands for "Remark" here

kani-compiler/src/args.rs
zhassan-aws: Good point.

kani-compiler/src/args.rs
zhassan-aws: I think I was trying to simplify the backend checking logic under different features. I'll clean this up.

zhassan-aws (commented):
> I haven't reviewed the backend code yet, but I'm curious about how this affects our released software. Are we shipping the new backend?

That was the plan (it's under an unstable feature). If we prefer to not include it in the release at this point, we can disable the feature by default and have a separate set of regressions for this feature.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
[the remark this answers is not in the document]
Nadrieril: Hah that makes so much more sense x)

kani-compiler/src/codegen_aeneas_llbc/compiler_interface.rs
```diff
@@ -0,0 +1,418 @@
...
+        for pass in charon_lib::transform::LLBC_PASSES.iter() {
+            pass.transform_ctx(&mut ccx)
+        }
+
+        println!("# Final LLBC before serialization:\n\n{}\n", ccx);
+
+        // Display an error report about the external dependencies, if necessary
+        ccx.errors.report_external_deps_errors();
+
+        trace!("Done");
```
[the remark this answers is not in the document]
zhassan-aws: Removed this one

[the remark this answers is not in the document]
zhassan-aws: I made the printing of the final LLBC dependent on a new option (`--print-llbc`) that is used in the current expected tests. Now, by default, nothing will get printed except the name of the LLBC file.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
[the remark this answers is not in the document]
zhassan-aws: Fun conversation :)

I was hesitant to remove the comments because we may need to import any future upstream changes to this function, so I wanted to minimize the diff. @tautschnig Let me know if you think the comments create confusion, and I'll remove them.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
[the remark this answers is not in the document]
zhassan-aws: Same comment here @tautschnig

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
```diff
@@ -0,0 +1,770 @@
...
+        let lineinfo = span.get_lines();
+        let rspan = RawSpan {
+            file_id,
+            beg: Loc { line: lineinfo.start_line, col: lineinfo.start_col },
+            end: Loc { line: lineinfo.end_line, col: lineinfo.end_col },
+            rust_span_data: rustc_internal::internal(self.tcx(), span).data(),
+        };
+
+        // TODO
+        CharonSpan { span: rspan, generated_from_span: None }
```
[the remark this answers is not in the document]
zhassan-aws: Expanded comment.

kani-compiler/Cargo.toml
```diff
@@ -10,6 +10,7 @@ publish = false
 
 [dependencies]
 cbmc = { path = "../cprover_bindings", package = "cprover_bindings", optional = true }
+charon = { git = "https://github.com/AeneasVerif/charon", optional = true, default-features = false, rev = "cdc1dcde447a50cbc20336c79b21b42ac977b7fd" }
```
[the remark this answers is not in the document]
celinval: I'm curious. Why not use a git submodule?

kani-compiler/src/args.rs
```diff
@@ -69,11 +82,12 @@ pub struct Arguments {
...
-    /// A legacy flag that is now ignored.
-    goto_c: bool,
-    /// Enable specific checks.
-    #[clap(long)]
     pub ub_check: Vec<ExtraChecks>,
+    #[clap(long = "backend")]
+    pub backend: Option<BackendOption>,
+    /// Print the final LLBC file to stdout.
+    #[clap(long)]
+    pub print_llbc: bool,
```
celinval, line 90: Why would we print to the stdout?

kani-compiler/src/codegen_aeneas_llbc/compiler_interface.rs
```diff
@@ -0,0 +1,420 @@
...
+        rustc_metadata: EncodedMetadata,
+        _need_metadata_module: bool,
+    ) -> Box<dyn Any> {
+        let ret_val = rustc_internal::run(tcx, || {
+            // Queries shouldn't change today once codegen starts.
+            let queries = self.queries.lock().unwrap().clone();
+
+            // Codegen all items that need to be processed according to the selected reachability mode:
+            //
+            // - Harnesses: Generate one model per local harnesses (marked with `kani::proof` attribute).
```
celinval, line 217: I'm curious. Do we want to keep this model for Aeneas backend too? Not something that we need to change for this PR, but it's something that it may be worth thinking about.

kani-compiler/src/codegen_aeneas_llbc/compiler_interface.rs
celinval, line 1: I think there's a lot of boiler plate code here that is duplicated from the cprover version. It may be worth factoring somethings out to `kani_middle` to reduce the burden. Also not a blocker for this PR.

kani-compiler/src/codegen_aeneas_llbc/compiler_interface.rs
```diff
@@ -0,0 +1,420 @@
...
+    fn metadata_loader(&self) -> Box<MetadataLoaderDyn> {
+        Box::new(rustc_codegen_ssa::back::metadata::DefaultMetadataLoader)
+    }
+
+    fn provide(&self, providers: &mut Providers) {
+        provide::provide(providers, &self.queries.lock().unwrap());
+    }
+
+    fn print_version(&self) {
+        println!("Kani-llbc version: {}", env!("CARGO_PKG_VERSION"));
```
celinval, line 197: Are we using `llbc` or `aeneas`? We should try to be coherent and only use one of them, don't you think?

kani-driver/src/args/mod.rs
celinval, line 1: Do we want to throw an error if users pass flags that are CBMC specific?

kani-driver/src/call_single_file.rs
```diff
@@ -95,6 +100,14 @@ impl KaniSession {
         to_rustc_arg(vec![format!("--reachability={}", self.reachability_mode())])
     }
 
+    pub fn backend_arg(&self) -> Option<String> {
+        if self.args.common_args.unstable_features.contains(UnstableFeature::Aeneas) {
+            Some(to_rustc_arg(vec!["--backend=aeneas".into()]))
+        } else {
+            None
```
celinval, line 107: Why do we return `None` here but `cprover` on line 59?

kani-compiler/src/args.rs
```diff
@@ -69,11 +82,12 @@ pub struct Arguments {
     /// Pass the kani version to the compiler to ensure cache coherence.
     check_version: Option<String>,
     #[clap(long)]
-    /// A legacy flag that is now ignored.
-    goto_c: bool,
-    /// Enable specific checks.
-    #[clap(long)]
     pub ub_check: Vec<ExtraChecks>,
+    #[clap(long = "backend")]
+    pub backend: Option<BackendOption>,
```
celinval, line 87: Please document. Also, what do you think about:

```suggestion
    #[clap(long = "backend", default_value_t = BackendOption::CProver)]
    pub backend: BackendOption,
```

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
```diff
@@ -0,0 +1,797 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
```
celinval, line 2: Should this be a modification copyright?

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
```diff
@@ -0,0 +1,797 @@
...
+                tid
+            }
+        };
+        *tid.as_fun()
+    }
+
+    /// Compute the meta information for a Rust item identified by its id.
+    fn translate_item_meta_from_rid(&mut self, instance: Instance) -> Result<ItemMeta, Error> {
+        let span = self.translate_instance_span(instance);
+        let name = self.def_id_to_name(instance.def.def_id())?;
```
celinval, line 148: I'm confused. Why can't we use `instance.mangled_name()`?

celinval (commented):
A few more comments

kani-compiler/Cargo.toml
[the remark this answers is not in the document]
zhassan-aws: Good question. I very much prefer using a submodule, and that's how I had it initially, but I couldn't get it to pass CI because of clippy issues in Charon (clippy checks the Charon crate as it is considered part of the Kani workspace). I could not find a way to disable running clippy on a specific crate (the exclude option seems to be broken: https://github.com/rust-lang/rust-clippy/issues/9555), and I eventually gave up.

I'll try to open a PR to address the clippy issues in the library, and once this is done, we can switch to using a git submodule.

kani-compiler/src/args.rs
zhassan-aws: This is for testing purposes as part of the `expected` test suite. Since Aeneas is not yet called by the driver, the only option I could think of to test the code is to check that the LLBC is as expected.

kani-compiler/src/codegen_aeneas_llbc/compiler_interface.rs
zhassan-aws: Good point. For Aeneas, I was thinking that the public function mode would be the default.

kani-driver/src/args/mod.rs
zhassan-aws: Yes. I'll add a TODO.

kani-compiler/Cargo.toml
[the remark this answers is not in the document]
Nadrieril: Happy to fix the clippy issues and run clippy inour CI (as long as you don't have too pedantic settings on). The main issue is that you're using a somewhat old charon commit so you'd have to upgrade to get these fixes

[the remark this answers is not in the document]
zhassan-aws: Thanks @Nadrieril. I'll try to reproduce the clippy issues I ran into and create a PR that fixes them or at least post them somewhere.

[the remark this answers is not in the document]
zhassan-aws: I (re-)tried to use a git submodule, and found out that the blocker was not the clippy issues in Charon, but rather that there were compile errors because clippy tries to compile `hax` which is not compatible with our rust toolchain version:
```bash
$ cargo clippy --all  -- -D warnings 
    Checking hax-frontend-exporter v0.1.0-pre.1 (https://github.com/hacspec/hax?branch=main#29d28fcb)
    Checking cprover_bindings v0.55.0 (/home/ubuntu/git/kani/cprover_bindings)
   Compiling kani-compiler v0.55.0 (/home/ubuntu/git/kani/kani-compiler)
    Checking compiletest v0.0.0 (/home/ubuntu/git/kani/tools/compiletest)
    Checking kani_metadata v0.55.0 (/home/ubuntu/git/kani/kani_metadata)
    Checking kani-driver v0.55.0 (/home/ubuntu/git/kani/kani-driver)
error[E0599]: no method named `predicates_defined_on` found for struct `rustc_middle::ty::TyCtxt<'tcx>` in the current scope
   --> /home/ubuntu/.cargo/git/checkouts/hax-310c6deb65d920f9/29d28fc/frontend/exporter/src/state.rs:300:26
    |
300 |             clauses: tcx.predicates_defined_on(did).predicates.sinto(s),
    |                          ^^^^^^^^^^^^^^^^^^^^^
    |
help: there is a method `predicates_of` with a similar name
    |
300 |             clauses: tcx.predicates_of(did).predicates.sinto(s),
    |                          ~~~~~~~~~~~~~

error[E0023]: this pattern has 1 field, but the corresponding tuple variant has 2 fields
    --> /home/ubuntu/.cargo/git/checkouts/hax-310c6deb65d920f9/29d28fc/frontend/exporter/src/types/copied.rs:1290:14
     |
1290 |     Lifetime(Symbol),
     |              ^^^^^^ expected 2 fields, found 1
    --> /home/ubuntu/.rustup/toolchains/nightly-2024-09-12-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/compiler/rustc_ast/src/token.rs:334:22
     |
     = note: tuple variant has 2 fields
     |
help: use `_` to explicitly ignore each field
     |
1290 |     Lifetime(Symbol, _),
     |                    +++
</snip>
error: could not compile `hax-frontend-exporter` (lib) due to 16 previous errors
```
Apparently, clippy attempts to build the entire Charon package, and not just the library?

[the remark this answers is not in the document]
zhassan-aws: OK, I got it to work using:
```
cargo clippy --all --exclude charon --no-deps -- -D warnings
```
I'll switch to using a git submodule.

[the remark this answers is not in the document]
zhassan-aws: Opened https://github.com/AeneasVerif/charon/pull/381 to fix the clippy issues I ran into.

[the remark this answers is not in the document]
zhassan-aws: Oh, and also ran into errors due to `cargo deny` 😬 

I'll need to update `deny.toml` to work around those.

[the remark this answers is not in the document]
zhassan-aws:
```
$ cargo deny --workspace --all-features check advisories
error[unmaintained]: ansi_term is Unmaintained
  ┌─ /home/ubuntu/git/kani/Cargo.lock:5:1
  │
5 │ ansi_term 0.12.1 registry+https://github.com/rust-lang/crates.io-index
  │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
  │
  ├ ID: RUSTSEC-2021-0139
  ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2021-0139
  ├ The maintainer has advised that this crate is deprecated and will not receive any maintenance.
    
    The crate does not seem to have much dependencies and may or may not be ok to use as-is.
    
    Last release seems to have been three years ago.
    
    ## Possible Alternative(s)
    
     The below list has not been vetted in any way and may or may not contain alternatives;
    
     - [ansiterm](https://crates.io/crates/ansiterm)
     - [anstyle](https://github.com/epage/anstyle)
     - [console](https://crates.io/crates/console)
     - [nu-ansi-term](https://crates.io/crates/nu-ansi-term)
     - [owo-colors](https://crates.io/crates/owo-colors)
     - [stylish](https://crates.io/crates/stylish)
     - [yansi](https://crates.io/crates/yansi)
    
    ## Dependency Specific Migration(s)
    
     - [structopt, clap2](https://github.com/clap-rs/clap/discussions/4172)
  ├ Announcement: https://github.com/ogham/rust-ansi-term/issues/72
  ├ Solution: No safe upgrade is available!
  ├ ansi_term v0.12.1
    └── assert_tokens_eq v0.1.0
        └── (dev) macros v0.1.0
            └── charon v0.1.36
                └── kani-compiler v0.55.0

error[unmaintained]: difference is unmaintained
   ┌─ /home/ubuntu/git/kani/Cargo.lock:56:1
   │
56 │ difference 2.0.0 registry+https://github.com/rust-lang/crates.io-index
   │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
   │
   ├ ID: RUSTSEC-2020-0095
   ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2020-0095
   ├ The author of the `difference` crate is unresponsive.
     
     Maintained alternatives:
     
     - [`dissimilar`](https://crates.io/crates/dissimilar)
     
     - [`similar`](https://crates.io/crates/similar)
     
     - [`treediff`](https://crates.io/crates/treediff)
     
     - [`diffus`](https://crates.io/crates/diffus)
   ├ Announcement: https://github.com/johannhof/difference.rs/issues/45
   ├ Solution: No safe upgrade is available!
   ├ difference v2.0.0
     └── assert_tokens_eq v0.1.0
         └── (dev) macros v0.1.0
             └── charon v0.1.36
                 └── kani-compiler v0.55.0

advisories FAILED
```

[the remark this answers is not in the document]
zhassan-aws: One more: `kani-regression.sh` fails because of unformatted Charon code. I'll need to add an entry in `rustfmt.toml` to ignore `charon`.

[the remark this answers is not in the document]
zhassan-aws: Sooooooo, to summarize, adding Charon as a git submodule has 3 implications:
- It gets checked by clippy
- It gets checked by cargo fmt
- It gets checked by cargo deny
I worked around all 3 issues, but we should fix the issues upstream and remove the workarounds.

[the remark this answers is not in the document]
Nadrieril: The cargo deny issues look easy to fix. Clippy I can add to our CI. Formatting I thought we already had this in our CI, what was the diff?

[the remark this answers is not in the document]
zhassan-aws: Here's the diff:

https://gist.github.com/zhassan-aws/d72bacfcb7bc93a94fd978d9feb4b71e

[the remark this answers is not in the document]
celinval: Having `cargo deny` failing is not a bad thing, is it? I'm actually surprised it wasn't failing before, doesn't it check the entire dependency tree? You should be able to go around `cargo fmt` and `cargo deny`, since this should not be a crate in the workspace.

kani-compiler/src/args.rs
celinval: I think this has to at least be marked as a hidden option, since it is not a behavior we want to keep. This should also be documented as a temporary option, and that it requires `-Z aenas`.

For these things, we usually use the `script-based-pre` tests, but depending on the number of tests you want to add, it can become very painful to maintain. So I'll leave it up to you.

kani-compiler/Cargo.toml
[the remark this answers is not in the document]
zhassan-aws: @Nadrieril it looks like the formatting is due to the configuration we're using in https://github.com/model-checking/kani/blob/main/rustfmt.toml

@celinval do you know why we're not using the default formatting configuration?

kani-compiler/src/codegen_aeneas_llbc/compiler_interface.rs
celinval: I wouldn't use CBMC as a model of consistency. :laughing:

I think this is a usability question, and being consistent and using a meaningful name might be the way to go here. I don't think any of our users really know what `goto` is, and using it in a user visible manner is probably more confusing than anything else. Try googling "goto" and you'll see what I'm saying. LOL

My main question to you is, what feature are the users enabling, `aeneas`, `llbc` or `lean`? Which ones are they going to be interacting with?

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
zhassan-aws: My understanding is that this tries to preserve the original names as much as possible so that it's easy for the user to relate the generated code to the original Rust.

kani-driver/src/args/mod.rs
zhassan-aws: Done.

kani-driver/src/call_single_file.rs
zhassan-aws: Good catch. I updated the one on line 59.

kani-compiler/src/args.rs
zhassan-aws: Makes sense. Updated.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
zhassan-aws: Updated.

kani-compiler/Cargo.toml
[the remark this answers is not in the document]
zhassan-aws: > Having `cargo deny` failing is not a bad thing, is it? I'm actually surprised it wasn't failing before, doesn't it check the entire dependency tree? You should be able to go around `cargo fmt` and `cargo deny`, since this should not be a crate in the workspace.

Good question. It was previously added in the `allow-git` section of the `deny.toml`:
```
allow-git = [
    "https://github.com/AeneasVerif/charon",
]
```
Perhaps this makes `cargo deny` not look at its dependencies?

kani-compiler/src/args.rs
zhassan-aws: > I think this has to at least be marked as a hidden option, since it is not a behavior we want to keep.

I recall trying to hide the option, but I couldn't find a way to do that. I'll check if that's possible.

> and that it requires -Z aenas.

I added that to the option documentation. The option validation would already error out if the option is specified without `-Zaeneas`.

> For these things, we usually use the script-based-pre tests, but depending on the number of tests you want to add, it can become very painful to maintain. So I'll leave it up to you.

Yeah, I considered this option, but it was indeed going to be painful.

kani-compiler/src/codegen_aeneas_llbc/compiler_interface.rs
Nadrieril: I'd say it depends on whether you want to use other tools than Aeneas on the output llbc. We have Eurydice that compiles llbc to C and we're developing simple static analyses at the moment.

zhassan-aws: @celinval My thinking was that the kani-compiler is an internal binary that is not meant to be used directly. What the kani-compiler is concerned with is to take MIR and translate it to GOTO or LLBC. The kani-driver is the  binary users interact with, and thus the options at that level should mention Lean or Aeneas.

What do you think?

kani-compiler/Cargo.toml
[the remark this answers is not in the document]
zhassan-aws: > @celinval do you know why we're not using the default formatting configuration?

Answering my own question: it seems we inherited the non-default configuration from rustc.

kani-compiler/src/args.rs
zhassan-aws: The option is now hidden.

kani-compiler/src/codegen_aeneas_llbc/compiler_interface.rs
celinval: For internal names in the compiler, I think `llbc` or `aeneas` would make sense. For user visible things, I think it will depend on what will be visible for the user. If users are not expected to interact with aeneas directly, I would prefer using Lean.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
celinval: I think it would be great to get more clarity on why we need this extra complexity. I'm also concerned that this could impact the analysis soundness (in case of 2+ names being resolved to the same name). It's also not clear to me how this is handling generic parameters.

celinval: Another potential issue is with different crate versions.. I wonder if they will have the same name for multiple items.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+                // TODO: find out if any of the information below is needed
+                let generic_args = GenericArgs {
+                    regions: Vector::new(),
+                    types,
+                    const_generics: Vector::new(),
+                    trait_refs: Vector::new(),
+                };
+                CharonTy::Adt(TypeId::Tuple, generic_args)
+            }
+            RigidTy::FnDef(def_id, _args) => {
```
celinval, line 449: I don't think you should be ignoring `args` here. You need to instantiate the type even when handling generic code.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+            generics: GenericParams::default(),
+            parent_params_info: None,
+            inputs: args,
+            output: ret_type,
+        }
+    }
+
+    fn translate_function_body(&mut self) -> Result<BodyId, Opaque> {
+        let instance = self.instance;
+        let mir_body = instance.body().unwrap();
```
celinval, line 381: From our offline chat. I think what you want is the following, you only want to generate code for generic code, but you want to use the reachability code to collect everything that is reachable from the user input. In that case, I recommend you do the following.

1. After the collector runs, build a set of `CrateDef` from all the collected instances. You can convert an instance to a `CrateItem` using the `TryFrom` implementation [here](https://github.com/rust-lang/rust/blob/baaf3e65abab5a281715230bcb1785610629ae7d/compiler/stable_mir/src/mir/mono.rs#L214-L226).
2. Most places where you use instance, you can replace by the crate item. I think the only place that you will need to change is where you the the `fn_abi()`. In that case, you need to use the function signature instead.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+            body: Ok(body),
+        };
+
+        self.translated.fun_decls.set_slot(fid, fun_decl);
+
+        Ok(())
+    }
+
+    /// Get or create a `FunDeclId` for the given function
+    fn register_fun_decl_id(&mut self, def_id: InternalDefId) -> FunDeclId {
```
celinval, line 133: Why do you need internal id here?

celinval (commented):
I think it would be great if we could eventually simplify the naming schema, and remove the name generation code. This seems fairly fragile, hard to maintain and I worry that this will not work (or even be unsound) if 2+ crates have the same name. That said, I think it is out of scope of this PR since you are only trying to match Aeneas name schema.

The handling of monomorphic vs generic code also need some improvement. To unblock this PR, can we add an error if we encounter generic code? I think none of your tests depend on generic code.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
zhassan-aws: Charon's data structures are in terms of the internal `DefId`: https://github.com/AeneasVerif/charon/blob/4caeeb35331b7e6ee59565a3eb698a200c6e90e9/charon/src/ast/krate.rs#L110

This was removed recently (in https://github.com/AeneasVerif/charon/commit/8a80793e440b256611baf747483a1b24613a043e), so once we upgrade to that commit, we should no longer need it.

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
zhassan-aws: I added an `unimplemented` for non-empty `args`. Would that work?

celinval (approved):
Sweet! I this is a great first step. Thanks @zhassan-aws

kani-compiler/src/codegen_aeneas_llbc/mir_to_ullbc/mod.rs
Nadrieril: These names are used in two ways: as @zhassan-aws said we use them when displaying info to the user so they can recognize which item we're talking about, and we also sometimes pattern-match on them to recognize built-ins. As of latest charon we use lang/diagnostic items wherever possible, but we still rely on names to recognize e.g. `core::panicking::panic_explicit`. That's not crucial.

Generic parameters are only relevant for inherent impl blocks. We include them in the name:
```rust
impl<T: Clone> Foo<T> {
    fn method(&self) {}
}
```
This method has name `crate::module::{impl<T: Clone> Foo<T>}::method`.

Well spotted on crate versions, I opened an [issue](https://github.com/AeneasVerif/charon/issues/412) for that.

PR #3522, opened 2024-09-17
Call `check_proof_attribute` for contract harnesses

carolynzech:
Kani enforces that `[kani::proof]` attribute is not applied to generic functions. We do not currently enforce this restriction on contract harnesses. When the compiler [searches for harnesses to verify](https://github.com/model-checking/kani/blob/dba8f3926a61025f5078de787ebd8d21278333ca/kani-compiler/src/kani_middle/codegen_units.rs#L63), it only looks at monomorphized functions. Thus, currently a user can write this code:

```rust
#[kani::requires(true)]
fn foo() {}

#[kani::proof_for_contract(foo)]
fn check_foo<T>() {
    foo()
}
```

and get 

"No proof harnesses (functions with #[kani::proof]) were found to verify."

In the case where a user is running many harnesses, they may not notice that Kani skipped the harness. For example, we currently have [this harness](https://github.com/model-checking/verify-rust-std/blob/149f6dd5409fac01a983d7b98c51d51666c74e45/library/core/src/ptr/unique.rs#L288) in the standard library, which doesn't actually run. (PR to fix is [here](https://github.com/model-checking/verify-rust-std/pull/86)). After this PR merges, the code snippet above would instead error with:

```rust
error: the '#[kani::proof_for_contract]' attribute cannot be applied to generic functions
 --> src/lib.rs:4:1
  |
4 | #[kani::proof_for_contract(foo)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |

```

### DIFF
--- a/kani-compiler/src/kani_middle/attributes.rs
+++ b/kani-compiler/src/kani_middle/attributes.rs
@@ -358,7 +358,7 @@ impl<'tcx> KaniAttributes<'tcx> {
                         );
                     }
                     expect_single(self.tcx, kind, &attrs);
-                    attrs.iter().for_each(|attr| self.check_proof_attribute(attr))
+                    attrs.iter().for_each(|attr| self.check_proof_attribute(kind, attr))
                 }
                 KaniAttributeKind::Unstable => attrs.iter().for_each(|attr| {
                     let _ = UnstableAttribute::try_from(*attr).map_err(|err| err.report(self.tcx));
@@ -370,6 +370,7 @@ impl<'tcx> KaniAttributes<'tcx> {
                         );
                     }
                     expect_single(self.tcx, kind, &attrs);
+                    attrs.iter().for_each(|attr| self.check_proof_attribute(kind, attr))
                 }
                 KaniAttributeKind::StubVerified => {
                     expect_single(self.tcx, kind, &attrs);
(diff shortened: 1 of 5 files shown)


carolynzech:
Note that we should merge https://github.com/model-checking/verify-rust-std/pull/86 first to avoid breaking the standard library.

celinval:
> Note that we should merge [model-checking/verify-rust-std#86](https://github.com/model-checking/verify-rust-std/pull/86) first to avoid breaking the standard library. (I would appreciate if someone would "Request changes" on this PR to avoid accidental auto-merge; I can't do it myself).

Merging this shouldn't break the std repository, right? BTW, you could leave it as draft for now too since you don't think this is ready to be merged.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -583,15 +584,29 @@ impl<'tcx> KaniAttributes<'tcx> {
...
+        if let KaniAttributeKind::Proof = kind {
+            expect_no_args(tcx, kind, proof_attribute);
+        }
+
         if tcx.def_kind(self.item) != DefKind::Fn {
-            tcx.dcx().span_err(span, "the `proof` attribute can only be applied to functions");
+            tcx.dcx().span_err(
+                span,
+                format!(
+                    "the '#[kani::{}]' attribute can only be applied to functions",
```
celinval, line 598: Interesting. I don't think we have been very consistent with how we name attributes in our messages. But looking at `rustc` messages, it doesn't seem like there is any patterns to follow.

celinval (approved):
Thanks

carolynzech:
> > Note that we should merge [model-checking/verify-rust-std#86](https://github.com/model-checking/verify-rust-std/pull/86) first to avoid breaking the standard library. (I would appreciate if someone would "Request changes" on this PR to avoid accidental auto-merge; I can't do it myself).
> 
> Merging this shouldn't break the std repository, right? BTW, you could leave it as draft for now too since you don't think this is ready to be merged.

It will break our harnesses in the std repository--see the new error messages in the logs:

```
2024-09-17T20:23:16.3424994Z    Compiling core v0.0.0 (/home/runner/work/kani/kani/verify-rust-std/library/core)
2024-09-17T20:23:26.2257417Z error: the '#[kani::proof_for_contract]' attribute cannot be applied to generic functions
2024-09-17T20:23:26.2258407Z    --> /home/runner/work/kani/kani/verify-rust-std/library/core/src/ptr/unique.rs:287:5
2024-09-17T20:23:26.2258952Z     |
2024-09-17T20:23:26.2259233Z 287 |     #[kani::proof_for_contract(Unique::cast<U>)]
2024-09-17T20:23:26.2259645Z     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-09-17T20:23:26.2259971Z     |
2024-09-17T20:23:26.2260750Z     = note: this error originates in the attribute macro `kani::proof_for_contract` (in Nightly builds, run with -Z macro-backtrace for more info)
2024-09-17T20:23:26.2261394Z 
2024-09-17T20:23:26.4192473Z error: aborting due to 1 previous error
```

Marking as draft for now.

PR #3537, opened 2024-09-23
Update to CBMC 6.3.1 and fix auto-update script

tautschnig:
Update to CBMC release 6.3.1. This release includes a full fix for the build problem that we carried a patch for (in #3431 and #3436), thus dropping the patching step.

The automatic update of CBMC requires actually installing that newer version of CBMC, else regression tests fail as seen in https://github.com/model-checking/kani/actions/runs/10987766383/job/30503118786.

Resolves: #3507, #3536


.github/workflows/cbmc-update.yml
```diff
@@ -51,6 +51,8 @@ jobs:
             sed -i "s/^CBMC_MINOR=.*/CBMC_MINOR=\"$CBMC_LATEST_MINOR\"/" kani-dependencies
             sed -i "s/^CBMC_VERSION=.*/CBMC_VERSION=\"$CBMC_LATEST\"/" kani-dependencies
             git diff
+            # install the newer CBMC version
+            scripts/setup/ubuntu/install_cbmc.sh
```
zhassan-aws, line 55: Maybe add a `./` prefix to be consistent with other script invocations in this file.

scripts/setup/al2/install_cbmc.sh
```diff
@@ -21,80 +21,6 @@ git clone \
 
 pushd "${WORK_DIR}"
 
-# apply workaround for https://github.com/diffblue/cbmc/issues/8357 until it is
```
zhassan-aws, line 24: Nice!

zhassan-aws (approved):
Thanks!

PR #3538, opened 2024-09-23
Add experimental API to generate arbitrary pointers

celinval:
This change adds a pointer generator that can non-deterministically generate a pointer with different properties.

This generator allows users to build pointers with different allocation status, initialization and alignment.
It contains an internal buffer that it uses to generate `InBounds` and `OutOfBounds` pointers.
In those cases, the pointers will have the same provenance as the generator, and the same lifetime.

This approach is different than generating a pointer from an arbitrary `usize`. Kani uses demonic non-determinism to track
allocation lifetimes, which makes hard to reason about during verification. I.e., one cannot assume a pointer is valid, and initialized, and this can only be accomplished by manually tracking the pointer status.

I added the new API under `-Z mem-predicates` since it allows reasoning about memory, and I was hoping this wouldn't need another unstable flag. :smile: 


tautschnig:
Appears to break std verification:
```
error[E0433]: failed to resolve: use of undeclared crate or module `kani`
   --> /home/runner/work/kani/kani/verify-rust-std/library/core/src/lib.rs:426:1
    |
426 | kani_core::kani_lib!(core);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `kani`
    |
    = note: this error originates in the macro `kani_core::generate_arbitrary` which comes from the expansion of the macro `kani_core::kani_lib` (in Nightly builds, run with -Z macro-backtrace for more info)
```

library/kani_core/src/arbitrary.rs
```diff
@@ -169,6 +180,171 @@ macro_rules! generate_arbitrary {
...
+            /// Dangling pointers
+            Dangling,
+            /// Pointer to dead object
+            DeadObj,
+            /// Null pointers
+            Null,
+            /// In bounds pointer (it may be unaligned)
+            InBounds,
+            /// Out of bounds
+            OutBounds,
```
tautschnig, line 196: Can we actually call it `OutOfBounds`, please?

library/kani_core/src/arbitrary.rs
```diff
@@ -169,6 +180,171 @@ macro_rules! generate_arbitrary {
...
         arbitrary_tuple!(A, B, C, D, E, F, G, H, I, J, K);
         arbitrary_tuple!(A, B, C, D, E, F, G, H, I, J, K, L);
+
+        /// Enumeration with the cases currently covered by the pointer generator.
+        #[derive(Copy, Clone, Debug, PartialEq, Eq, kani::Arbitrary)]
+        pub enum AllocationStatus {
+            /// Dangling pointers
+            Dangling,
+            /// Pointer to dead object
+            DeadObj,
```
tautschnig, line 190: Can we spell this out as `DeadObject`?

library/kani_core/src/arbitrary.rs
```diff
@@ -169,6 +180,171 @@ macro_rules! generate_arbitrary {
...
+        ///
+        /// This generator allows users to build pointers with different safety properties.
+        /// It contains an internal buffer that it uses to generate `InBounds` and `OutBounds` pointers.
+        /// In those cases, the pointers will have the same provenance as the generator, and the same lifetime.
+        ///
+        /// For example:
+        /// ```ignore
+        ///     let generator = PointerGenerator<char, 5>::new();
+        ///     let arbitrary = generator.generate_ptr();
+        ///     kani::assume(arbitrary.status == AllocationStatus::InBounds);
```
tautschnig, line 209: Can you please spell out what this actually does? What do `char` and `5` imply here?

library/kani_core/src/arbitrary.rs
```diff
@@ -169,6 +180,171 @@ macro_rules! generate_arbitrary {
...
+        ///     kani::cover!(arbitrary.ptr != generator.generate_ptr());
+        /// ```
+        ///
+        /// Note: This code is different than generating a pointer with any address. I.e.:
+        /// ```ignore
+        ///     // This pointer represents any address.
+        ///     let ptr = kani::any::<usize>() as *const u8;
+        ///     // Which is different from:
+        ///     let generator = PointerGenerator<u8, 5>::new();
+        ///     let ptr = generator.generate_ptr().ptr;
```
tautschnig, line 230: You say it is "different," but could you please elaborate in what way this is different?

tests/expected/arbitrary/ptrs/pointer_inbounds.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// kani-flags: -Z mem-predicates
+//! Check different cases for `PointerGenerator` for in-bounds pointers.
+//! TODO: Enable initialization checks once we enable reasoning about unions.
```
tautschnig, line 6: I'm curious: where is "reasoning about unions" disabled?

tests/expected/arbitrary/ptrs/pointer_generator_error.expected
```diff
@@ -0,0 +1,10 @@
+error[E0080]: evaluation of `kani::PointerGenerator::<char, 0>::_VALID` failed
+
+the evaluated program panicked at 'PointerGenerator requires non-zero length.'
+
+note: the above error was encountered while instantiating `fn kani::PointerGenerator::<char, 0>::new`\
+pointer_generator_error.rs\
+|\
+|     let _generator = PointerGenerator::<char, 0>::new();\
+|                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
tautschnig, line 9: I wasn't able to find any documentation that said what the valid values were?

tautschnig (commented):
Thank you for working on this, but I think we really need more documentation to enable safe and sound use of this capability in any proofs that users may be constructing.

tests/expected/arbitrary/ptrs/pointer_inbounds.rs
celinval: I'm talking about enabling `-Z uninit-checks`. Kani memory initializatin instrumentation does not work in the presence of `MaybeUninit` which we use to implement `PointerGenerator`. Kani will detect the usage of `MaybeUninit` and it fail the verification.

tautschnig: Can you perhaps include that clarification in the comment?

celinval:
@tautschnig, hopefully I addressed all your comments. I also moved the pointer type to be an argument of the methods instead of the generator type.

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,327 @@
...
+        /// Compilation will fail if you try to create a generator of size `0`.
+        ///
+        /// Additionally, the verification will fail if you try to generate a pointer for a type
+        /// with size greater than the buffer size.
+        ///
+        /// Use larger buffer size if you want to cover scenarios where the distance
+        /// between the generated pointers matters.
+        ///
+        /// The only caveats of using very large numbers are:
+        ///  1. The value cannot exceed the solver maximum object size, neither Rust's
```
zhassan-aws, line 96:
```suggestion
        ///  1. The value cannot exceed the solver maximum object size (currently 2^48 by default), neither Rust's
```

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,327 @@
...
+        ///     maximum object size (`isize::MAX`).
+        ///  2. Larger sizes could impact performance.
```
zhassan-aws, line 98:
```suggestion
        ///  2. Larger sizes could impact performance as they can lead to an exponential increase in the placement of pointers within the buffer.
```

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,327 @@
...
+        ///
+        /// I.e.:
+        /// ```ignore
+        ///     // This pointer represents any address, and it may point to anything in memory,
+        ///     // allocated or not.
+        ///     let ptr1 = kani::any::<usize>() as *const u8;
+        ///
+        ///     // This pointer address will either point to unallocated memory, to a dead object
+        ///     // or to allocated memory within the generator address space.
+        ///     let generator = PointerGenerator<u8, 5>::new();
```
zhassan-aws, line 156: Should it be `PointerGenerator::<5>::new()`?

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,327 @@
...
+        ///     let ptr2 = generator.any_alloc_status().ptr;
+        /// ```
+        ///
+        /// Kani cannot reason about a pointer allocation status (except for asserting its validity).
+        /// Thus, the generator was introduced to help writing harnesses that need to impose
```
zhassan-aws, line 161: This sentence (and perhaps this entire section) needs to be moved to the beginning since it explains the whole purpose of the generator.

zhassan-aws (commented):
Nice!

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,327 @@
...
+        ///
+        /// ## Buffer size
+        ///
+        /// The internal buffer is used to generate pointers, and its size determines the maximum
+        /// number of pointers it can generate without overlapping.
+        /// Larger values will also impact on the maximum distance between each generated pointer.
+        ///
+        /// We recommend that you pick a size that is at least big enough to
+        /// cover the cases where all pointers produced are non-overlapping.
+        /// The buffer size in bytes must be big enough to fit distinct objects for each call
```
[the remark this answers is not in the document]
celinval: Must. There's an assertion that will fail otherwise.

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,327 @@
...
+        /// # use kani::*;
+        /// # #[kani::proof]
+        /// # fn harness() {
+        ///     let generator = PointerGenerator<6>::new();
+        ///     let ptr1: *mut u8 = generator.any_in_bounds().ptr;
+        ///     let ptr2: *mut u8 = generator.any_in_bounds().ptr;
+        ///     let ptr3: *mut u32 = generator.any_in_bounds().ptr;
+        ///     // This cover is satisfied.
+        ///     cover!((ptr1 as usize) > (ptr2 as usize) + size_of::<char>()
+        ///            && (ptr2 as usize) > (ptr3 as usize) + size_of::<char>());
```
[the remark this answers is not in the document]
celinval: tbh, these examples were the reason I decided to enable doctests. It's too painful to run them in my head. :smile:

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,327 @@
...
+        /// Use the same generator if you want to handle cases where 2 or more pointers may overlap. E.g.:
+        /// ```ignore
+        /// # use kani::*;
+        /// # #[kani::proof]
+        /// # fn harness() {
+        ///     let generator = pointer_generator::<char, 5>();
+        ///     let ptr1 = generator.any_in_bounds::<char>().ptr;
+        ///     let ptr2 = generator.any_in_bounds::<char>().ptr;
+        ///     // This cover is satisfied.
+        ///     cover!(ptr1 == ptr2)
```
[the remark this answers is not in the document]
celinval: Not really. Raw pointer comparison is a stable and safe operation in Rust.

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,327 @@
...
+        /// ```
+        ///
+        /// # Pointer Generator vs Pointer with any address
+        ///
+        /// Creating a pointer using the generator is different than generating a pointer
+        /// with any address.
+        ///
+        /// I.e.:
+        /// ```ignore
+        ///     // This pointer represents any address, and it may point to anything in memory,
```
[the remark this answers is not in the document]
celinval: The `#` is to omit some parts of the example from the documentation. But you are correct that the indentation here is wrong. so I'll fix it. Thanks!

library/kani_core/src/arbitrary/pointer.rs
celinval: I was debating about this. Writing the why before the what makes sense in proposals, but I'm not sure it reads better in the API documentation.

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,327 @@
...
+            )]
+            pub fn any_alloc_status<'a, T>(&'a mut self) -> ArbitraryPointer<'a, T>
+             where T: kani::Arbitrary
+            {
+                assert!(core::mem::size_of::<T>() <= Self::BUF_LEN,
+                    "Cannot generate in-bounds object of the requested type. Buffer is not big enough."
+                );
+
+                // Create an arbitrary pointer, but leave `ptr` as null.
+                let mut arbitrary = ArbitraryPointer {
```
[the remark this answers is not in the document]
celinval: sure

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,327 @@
...
+                feature = "mem-predicates",
+                issue = 2690,
+                reason = "experimental memory predicates and manipulation feature"
+            )]
+            pub fn any_in_bounds<'a, T>(&'a mut self) -> ArbitraryPointer<'a, T>
+            where T: kani::Arbitrary {
+                assert!(core::mem::size_of::<T>() <= Self::BUF_LEN,
+                    "Cannot generate in-bounds object of the requested type. Buffer is not big enough."
+                );
+                let buf_ptr = addr_of_mut!(self.buf) as *mut u8;
```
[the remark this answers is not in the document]
celinval: Sure! The code here converged nicely as I re-implemented this a few times, so I don't see why not.

tests/expected/arbitrary/ptrs/pointer_generator.expected
```diff
@@ -0,0 +1,21 @@
...
+Status: SUCCESS\
+Description: ""NullPtr""
+
+Status: FAILURE\
+Description: ""DeadObject""
+
+Status: SATISFIED\
+Description: "Dangling"
+
+Status: UNREACHABLE\
```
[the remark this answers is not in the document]
celinval: Because `can_write_unaligned` will always panic

library/kani_core/src/arbitrary/pointer.rs
celinval: What if I add one sentence to the beginning of the doc and refer to this section for more details?

tests/expected/arbitrary/ptrs/pointer_inbounds.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[kani::proof]
+fn check_overlap() {
+    let mut generator = kani::pointer_generator::<u16, 5>();
+    let ptr_1 = generator.any_in_bounds::<u16>().ptr as usize;
+    let ptr_2 = generator.any_in_bounds::<u16>().ptr as usize;
+    kani::cover!(ptr_1 == ptr_2, "Same");
+    kani::cover!(ptr_1 == ptr_2 + 1, "Overlap");
+    kani::cover!(ptr_1 > ptr_2 + 4, "Greater");
+    kani::cover!(ptr_2 > ptr_1 + 4, "Smaller");
```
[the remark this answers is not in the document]
celinval: The distance between the two is at most 8 bytes, right?

library/kani_core/src/arbitrary/pointer.rs
zhassan-aws: I guess it's OK. From the review perspective, it would've helped if I saw this sentence first, but from the documentation perspective, having the what before the why makes more sense.

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,371 @@
...
+            )]
+            pub fn any_alloc_status<'a, T>(&'a mut self) -> ArbitraryPointer<'a, T>
+             where T: kani::Arbitrary
+            {
+                assert!(core::mem::size_of::<T>() <= Self::BUF_LEN,
+                    "Cannot generate in-bounds object of the requested type. Buffer is not big enough."
+                );
+
+                let status = kani::any();
+                let mut is_initialized = false;
```
zhassan-aws, line 265: This doesn't need to be `mut`.

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,371 @@
...
+                        return self.create_in_bounds_ptr();
+                    }
+                    AllocationStatus::OutOfBounds => {
+                        let buf_ptr = addr_of_mut!(self.buf) as *mut u8;
+                        let offset = kani::any_where(|b: &usize| *b < size_of::<T>());
+                        unsafe { buf_ptr.add(Self::BUF_LEN - offset) as *mut T }
+                    }
+                };
+
+                // Create an arbitrary pointer, but leave `ptr` as null.
```
zhassan-aws, line 288: This comment is no longer valid.

library/kani_core/src/arbitrary/pointer.rs
```diff
@@ -0,0 +1,371 @@
...
+        /// # use kani::*;
+        /// # #[kani::proof]
+        /// # fn harness() {
+        ///     let mut generator = PointerGenerator::<6>::new();
+        ///     let ptr1: *mut u8 = generator.any_in_bounds().ptr;
+        ///     let ptr2: *mut u8 = generator.any_in_bounds().ptr;
+        ///     let ptr3: *mut u32 = generator.any_in_bounds().ptr;
+        ///     // This cover is satisfied.
+        ///     cover!((ptr1 as usize) >= (ptr2 as usize) + size_of::<u8>()
+        ///            && (ptr2 as usize) >= (ptr3 as usize) + size_of::<u8>());
```
zhassan-aws, line 88: Should this be `u32`:
```suggestion
        ///            && (ptr2 as usize) >= (ptr3 as usize) + size_of::<u32>());
```

zhassan-aws (approved):

library/kani_core/src/arbitrary/pointer.rs
celinval:
```suggestion
                let is_initialized = false;
```

PR #3539, opened 2024-09-23
Update toolchain to 2024-09-20

tautschnig:
Changes required due to:
- rust-lang/rust@bdacdfe95f Minimize visibilities.

Resolves: #3534


qinheping (approved):
I think it is OK to copy those functions in our code while waiting to see if they should be public in the upstream, and unblock the following toolchain updates.

Which way is better to remind us to review if these functions become pub in the future: mark them as FIXME, or create an issue to track them?

celinval (changes requested):
Can you try the following solution instead?

In the `if` block from the `link` function, I would replace the body by a call to `link_binary` as other backend implementations do, [e.g. here](https://github.com/rust-lang/rust/blob/648d024a7859e1ab7fdffe5e419b6e35ccb16a4a/compiler/rustc_codegen_gcc/src/lib.rs#L269-L271).

You will also need to implement the ArchiveBuilderBuilder trait. I believe a dummy implementation as [this should](https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_cranelift/src/archive.rs) work.

Let me know if that makes sense.

celinval:
I guess it took me too long to post my comment. I created https://github.com/model-checking/kani/pull/3545 with the suggestion above.

PR #3546, opened 2024-09-24
Analyze unsafe code reachability

celinval:
Add call graph analysis to scanner in order to find the distance between functions in a crate and unsafe functions.

For that, we build the crate call graph and collect the unsafe functions. After that, do reverse BFS traversal from the unsafe functions and store the distance to other functions. The result is stored in a new csv file.


docs/src/rustc-hacks.md
```diff
@@ -52,19 +52,27 @@ Note that pretty printing for the Rust nightly toolchain (which Kani uses) is no
...
-# The best way is to use the rustup path. Replace <toolchain> with the
-# proper name to your toolchain.
-rustc_driver = { path = "~/.rustup/toolchains/<toolchain>/lib/rustlib/rustc-src/rust/compiler/rustc_driver" }
-rustc_interface = { path = "~/.rustup/toolchains/<toolchain>/lib/rustlib/rustc-src/rust/compiler/rustc_interface" }
+# E.g.: ~/.rustup/toolchains/<toolchain>/lib/rustlib/rustc-src/rust/compiler
+rustc_smir = { path = "<path_to_rustc>/rustc_smir", optional = true }
+stable_mir = { path = "<path_to_rustc>/stable_mir", optional = true }
+
+[features]
+clion = ['rustc_smir', 'stable_mir']
```
tautschnig, line 75: How are the changes in this file related to what the PR description says?

scripts/std-analysis.sh
```diff
@@ -104,7 +105,7 @@ for f in $results/*overall.csv; do
     fname=$(basename $f)
     crate=${fname%_scan_overall.csv}
     echo -n "$crate," >> $summary
-    tr -d [:alpha:]_,; < $f | tr -s '\n' ',' \
+    tr -d "[:alpha:]_,;" < $f | tr -s '\n' ',' \
```
tautschnig, line 108: Thank you, but how did this ever work?

tests/script-based-pre/tool-scanner/test.rs
```diff
@@ -102,3 +107,11 @@ pub fn start_recursion() {
...
     let _ = ok();
 }
+
+extern "C" {
+    fn external_function();
+}
+
+pub fn call_external() {
+    unsafe { external_function() };
+}
```
tautschnig, line 117: Perhaps those various functions could have comments explaining why they are there?

tools/scanner/src/call_graph.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Provide different static analysis to be performed in the call graph
```
tautschnig, line 4: What does "different" mean here?

tools/scanner/src/analysis.rs
```diff
@@ -230,24 +238,24 @@ impl OverallStats {
...
         }
 
         impl $name {
-            const fn num_props() -> usize {
+            pub const fn num_props() -> usize {
                 [$(stringify!($prop),)+].len()
             }
 
-            fn new(fn_name: String) -> Self {
+            pub fn new(fn_name: String) -> Self {
```
tautschnig, line 258: Why are those newly public? Probably just a lack of understanding on my part, but I couldn't spot a new use of them.

docs/src/rustc-hacks.md
celinval: I already removed them... They are not. :)

scripts/std-analysis.sh
celinval: I think it was a typo when I created the script from my local script. This script doesn't run as part of the CI, which is probably something that we need to address.

tools/scanner/src/call_graph.rs
celinval: let me improve this...

tools/scanner/src/analysis.rs
celinval: I thought I was using them in the `call_graph` module, but I can double check if that is still needed.

scripts/std-analysis.sh
tautschnig: Yes, would love for that to be in CI in some way or other!

celinval: I agree, but I believe this is out of scope of this PR, and it should be done if a follow up PR.

tools/scanner/src/analysis.rs
celinval: Yes, we do need the `new` in the `call_graph` module now

PR #3549, opened 2024-09-26
Update toolchain to 2024-09-26

tautschnig:
Changes required due to:
- https://github.com/rust-lang/rust/pull/130234 improve compile errors for invalid ptr-to-ptr casts with trait objects
- rust-lang/rust@cfb8419900 Separate collection of crate-local inherent impls from error reporting
- rust-lang/rust@40fca8f7a8 Bump Clippy version -> 0.1.83

Resolves: #3548


cprover_bindings/src/goto_program/symbol.rs
```diff
@@ -20,8 +20,6 @@ pub struct Symbol {
     /// Contracts to be enforced (only supported for functions)
     pub contract: Option<Box<FunctionContract>>,
 
-    /// Optional debugging information
```
carolynzech, line 23: Why are we deleting this?

carolynzech (approved):

cprover_bindings/src/goto_program/symbol.rs
tautschnig: I don't know why we had it in the first place: none of the fields below this are "debug" information, they are rather essential to describe that symbol.

PR #3557, opened 2024-10-01
Enable doctests to avoid coding mistakes in our API documentation

celinval:
In order to enable the tests, I had to fix a few tests, enhance tests
with hidden code, and mark a few as `no_run` instead of `rust`.

The tests marked as `no_run` will at least compile, but execution would
be stuck in an infinite loop, since we provide dummy implementation
to a lot of our intrinsics with `loop {}`.

Finally, as I was cleaning up the documentation, I noticed that we
accidently published an API that was meant to be internal to Kani.
I marked this API as deprecated for now.

### DIFF
--- a/library/kani/src/invariant.rs
+++ b/library/kani/src/invariant.rs
@@ -39,6 +39,14 @@
 /// ```
 /// You can specify its safety invariant as:
 /// ```rust
+/// # #[derive(kani::Arbitrary)]
+/// # pub struct MyDate {
+/// #  day: u8,
+/// #  month: u8,
+/// #  year: i64,
+/// # }
+/// # fn days_in_month(_: i64, _: u8) -> u8 { 31 }
+///
 /// impl kani::Invariant for MyDate {
 ///   fn is_safe(&self) -> bool {
 ///     self.month > 0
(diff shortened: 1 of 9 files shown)


library/kani_core/src/lib.rs
```diff
@@ -58,6 +59,11 @@ macro_rules! kani_lib {
         kani_core::kani_intrinsics!(std);
         kani_core::generate_arbitrary!(std);
 
+        kani_core::check_intrinsic! {
+        msg="This API was accidently added as a public function and it will be made private in future releases.",
```
tautschnig, line 63: Can you please create a tracking issue with a proposed timeline so that we don't forget to act upon this?

tautschnig (approved):

library/kani/src/invariant.rs
```diff
@@ -49,12 +57,33 @@
...
+/// # pub struct MyDate {
+/// #  day: u8,
+/// #  month: u8,
+/// #  year: i64,
+/// # }
+/// #
+/// # fn days_in_month(_: i64, _: u8) -> u8 { todo!() }
+/// # fn increase_date(_: &mut MyDate, _: u8) { todo!() }
+/// #
+/// # impl kani::Invariant for MyDate {
```
zhassan-aws, line 73:
```suggestion
/// # impl Invariant for MyDate {
```

zhassan-aws: or remove the `use kani::Invariant` above.

library/kani_core/src/lib.rs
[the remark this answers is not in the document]
zhassan-aws: I think we can just say "This API will be made private in future releases."

celinval: https://github.com/model-checking/kani/issues/3561

zhassan-aws (approved):

PR #3559, opened 2024-10-01
Remove obsolete linker options

zhassan-aws:
These were made obsolete a while back.


tests/cargo-kani/mir-linker/src/lib.rs
celinval, line 1: I actually think this is a good test. Can we just remove mir-linker but keep the test?

kani-driver/src/args/mod.rs
```diff
@@ -40,6 +40,7 @@ where
         .unwrap()
 }
 
+#[allow(dead_code)]
```
celinval, line 43: Why?

celinval (approved):
Awesome!! Thanks for cleaning up this.

tests/cargo-kani/mir-linker/src/lib.rs
zhassan-aws: Added back

kani-driver/src/args/mod.rs
zhassan-aws: This function is no longer used, so clippy complains. I thought the function is good to keep to use if we decide to obsolete any options in the future, so that we use the same message wording.

PR #3580, opened 2024-10-08
Exclude Charon from workspace

zhassan-aws:
@celinval pointed out that `cargo update` will attempt to update the `charon` submodule as well.

Exclude the `charon` submodule from the workspace so that it's not considered when running `cargo update`, `cargo deny`, `cargo clippy`, and `cargo fmt`. The PR also reverts some of the configuration changes made in #3514.


celinval (changes requested):
Can you please wait for the release before merging this? Thanks!

celinval (approved):
Thanks!

PR #3583, opened 2024-10-08
Add fn that checks pointers point to same allocation

celinval:
This changes adds a new function that returns whether two pointers points to the same allocated object. Since we cannot reason about addresses to objects that are dead or haven't been allocated, we panic if the user provides dangling pointers.

## Call-out:

- Creating this as draft for now for early feedback

### DIFF
--- a/library/kani_core/src/mem.rs
+++ b/library/kani_core/src/mem.rs
@@ -154,11 +154,32 @@ macro_rules! kani_mem {
             <T as Pointee>::Metadata: PtrProperties<T>,
         {
             let (thin_ptr, metadata) = ptr.to_raw_parts();
-            // Need to assert `is_initialized` because non-determinism is used under the hood, so it
-            // does not make sense to use it inside assumption context.
+            // Need to assert `is_initialized` because non-determinism is used under the hood, so it
+            // does not make sense to use it inside assumption context.
             is_inbounds(&metadata, thin_ptr)
                 && assert_is_initialized(ptr)
                 && unsafe { has_valid_value(ptr) }
+        }
+
+        /// Check if two pointers points to the same allocated object, and that both pointers
+        /// are in bounds of that object.
+        ///
+        /// Note that a pointer is still in-bounds if it points to 1-byte past the allocation.
+        #[crate::kani::unstable_feature(
+            feature = "mem-predicates",
+            issue = 2690,
+            reason = "experimental memory predicate API"
+        )]
+        #[allow(clippy::not_unsafe_ptr_arg_deref)]
+        pub fn same_allocation<T>(ptr1: *const T, ptr2: *const T) -> bool {
+            let obj1 = crate::mem::pointer_object(ptr1);
+            (obj1 == crate::mem::pointer_object(ptr2)) && {
+                super::assert(
+                    unsafe { is_allocated(obj1 as *const (), 0) },
+                    "Kani does not support reasoning about pointer to unallocated memory",
+                );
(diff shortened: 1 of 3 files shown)


celinval:
@tautschnig @zhassan-aws what do you think about this?

library/kani_core/src/mem.rs
```diff
@@ -154,13 +154,34 @@ macro_rules! kani_mem {
...
+            feature = "mem-predicates",
+            issue = 2690,
+            reason = "experimental memory predicate API"
+        )]
+        #[allow(clippy::not_unsafe_ptr_arg_deref)]
+        pub fn same_allocation<T>(ptr1: *const T, ptr2: *const T) -> bool {
+            let obj1 = crate::mem::pointer_object(ptr1);
+            (obj1 == crate::mem::pointer_object(ptr2)) && {
+                super::assert(
+                    unsafe { is_allocated(obj1 as *const (), 0) },
```
zhassan-aws, line 178: Should this be `ptr1` and not `obj1`?

zhassan-aws (commented):
Makes sense!

library/kani_core/src/mem.rs
celinval: No, I think this should actually be obj1. If I'm not wrong, CBMC allocation always starts at offset 0 of the allocated object. So this will succeed even if ptr1 is out-of-bounds. @tautschnig is that correct?

tautschnig: This will work, but it only works because of how CBMC's memory model works today. This implies that we are also restricting ourselves to the CBMC back-end at this point. I don't think you should expect for this to work with Aeneas. Maybe we want to wrap this in a new function that is sort of an inverse to `pointer_object`, taking a `usize` and returning a pointer? Something like `pointer_from_object` perhaps?

celinval: Are you saying that the function signature is CBMC specific or the implementation?

We can easily swap the implementation later once we have a second backend. I think what we really want is something that reason about ptr provenance, but CBMC can't do that today.

celinval: gotcha

celinval:
@tautschnig @zhassan-aws ?

library/kani_core/src/mem.rs
```diff
@@ -338,16 +334,30 @@ macro_rules! kani_mem {
...
 
+        pub(super) mod cbmc {
+            use super::*;
+            /// CBMC specific implementation of [super::same_allocation].
+            pub fn same_allocation<T>(ptr1: *const T, ptr2: *const T) -> bool {
+                let obj1 = crate::kani::mem::pointer_object(ptr1);
+                (obj1 == crate::kani::mem::pointer_object(ptr2)) && {
+                    crate::kani::assert(
+                        unsafe {
+                            is_allocated(ptr1 as *const (), 0) || is_allocated(ptr2 as *const (), 0)
```
zhassan-aws, line 345: I'm now confused whether we settled on `ptr1` or `obj1`.

zhassan-aws (approved):
Nice tests!

library/kani_core/src/mem.rs
celinval: I prefer object since the assertion would still succeed even if both pointers are out of bounds. But I got the impression that we didn't want to do that. So I changed to assert that at least one of the pointers is valid.

zhassan-aws: Got it

Issue #3582, opened 2024-10-08
Spurious failure in pointer offset


zhassan-aws:
I tried this code:

```rust
#[kani::proof]
fn check_foo() {
    let x: u8 = 0;
    let ptr: *const u8 = &x as *const u8;
    let count = isize::MAX - 800;
    let _ = ptr.wrapping_offset(count);
}
```

using the following command line invocation:

```
kani lib.rs
```

with Kani version: fe17fdc31f5

I expected to see this happen: Verification successful.

Instead, this happened: Verification failed:
```bash
RESULTS:
Check 1: std::ptr::const_ptr::<impl *const u8>::wrapping_offset.arithmetic_overflow.1
         - Status: SUCCESS
         - Description: "arith_offset: attempt to compute number in bytes which would overflow"
         - Location: ../../../.rustup/toolchains/nightly-2024-10-03-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/const_ptr.rs:481:18 in function std::ptr::const_ptr::<impl *const u8>::wrapping_offset

Check 2: std::ptr::const_ptr::<impl *const u8>::wrapping_offset.arithmetic_overflow.2
         - Status: FAILURE
         - Description: "attempt to compute offset which would overflow"
         - Location: ../../../.rustup/toolchains/nightly-2024-10-03-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/const_ptr.rs:481:18 in function std::ptr::const_ptr::<impl *const u8>::wrapping_offset

Check 3: check_foo.assertion.1
         - Status: SUCCESS
         - Description: "attempt to subtract with overflow"
         - Location: lib.rs:5:17 in function check_foo


SUMMARY:
 ** 1 of 3 failed
Failed Checks: attempt to compute offset which would overflow
 File: "/home/ubuntu/.rustup/toolchains/nightly-2024-10-03-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/const_ptr.rs", line 481, in std::ptr::const_ptr::<impl *const u8>::wrapping_offset

VERIFICATION:- FAILED
```

zhassan-aws:
I get a similar failure if using `offset` as well, which should fail due to the pointer going out of bounds, but **not** during the computation of the number of bytes (since `size_of::<u8>` is 1):
```rust
#[kani::proof]
fn check_foo() {
    let x: u8 = 0;
[...]

PR #3584, opened 2024-10-09
Fix the computation of the number of bytes of a pointer offset

zhassan-aws:
This PR fixes the logic that computes the number of bytes of a pointer offset. The logic was incorrectly using the size of the pointer as opposed to the size of the pointee.

This fixes the soundness issue in #3582 (but not the spurious failures).


celinval (commented):
Thanks! Why are we still seeing spurious failures?

tautschnig:
Are we perhaps making the same mistake in `codegen_offset`?

zhassan-aws:
The usage of the pointer instead of the pointee was causing a soundness/spurious issue in the _multiplication_ operation used for computing the number of bytes. Those issues are fixed by this PR. But the remaining spurious failures are in the addition operation (adding the pointer value to the number of bytes). I'm not clear on their cause yet.

celinval (approved):

PR #3589, opened 2024-10-09
Remove the overflow checks for wrapping_offset

### LINKED ISSUE #3582: Spurious failure in pointer offset

zhassan-aws:
This PR removes the overflow checks done for the `arith_offset` intrinsic, which is used in implementation of `ptr::wrapping_offset`. According to the documentation, this operation is always safe:

https://doc.rust-lang.org/std/primitive.pointer.html#method.wrapping_offset
https://doc.rust-lang.org/std/intrinsics/fn.arith_offset.html

See https://github.com/model-checking/kani/issues/3582#issuecomment-2403453345 for the context.

Resolves #3582 


tests/expected/offset-overflow/main.rs
```diff
@@ -8,13 +8,12 @@ use std::intrinsics::offset;
...
 #[kani::proof]
 fn test_offset_overflow() {
-    let a: [i32; 3] = [1, 2, 3];
-    let ptr: *const i32 = a.as_ptr();
+    let s: &str = "123";
+    let ptr: *const u8 = s.as_ptr();
 
-    // a value that when multiplied by the size of i32 (i.e. 4 bytes)
-    // would overflow `isize`
-    let count: isize = isize::MAX / 4 + 1;
```
celinval, line 16: What's wrong with this?

celinval (approved):
Thanks

tests/expected/offset-overflow/main.rs
zhassan-aws: Nothing. I changed this test in #3584 to have an overflow in the multiplication operation, but then realized that it was meant to check overflow in the addition operation, so I reverted the change.

PR #3594, opened 2024-10-11
Add a gen_proof_for_contract macro

celinval:
This is just an experiment to get early feedback. This currently does not support methods.

Resolves #ISSUE-NUMBER


celinval:
On a second thought, I don't think we need to add this to Kani library.

Issue #3467, opened 2024-08-29
ICE Resolving `proof_for_contract` target


carolynzech:
I tried this code:
```rust
pub(crate) const unsafe fn align_offset<T: Sized>(p: *const T, a: usize) -> usize {
    ...
    #[safety::requires(m.is_power_of_two())]
    #[safety::requires(x < m)]
    const unsafe fn mod_inv(x: usize, m: usize) -> usize { ... }

    #[cfg(kani)]
    #[kani::proof_for_contract(mod_inv)]
    fn check_mod_inv() {
        let x = kani::any::<usize>();
        let m = kani::any::<usize>();
        unsafe { mod_inv(x, m) };
    }
   ...
}
```

For the actual code (to run this locally), add the contract and proof above [here](https://github.com/model-checking/verify-rust-std/blob/f16e2972d62565598aaadca5188db3005fee3495/library/core/src/ptr/mod.rs#L1887). 

using the following command line invocation:

```
kani verify-std -Z unstable-options ./library --target-dir /tmp/verify-rust-std -Z function-contracts -Z mem-predicates --harness check_mod_inv
```

with Kani version: 0.5.4

I expected to see this happen: verification success

Instead, this happened: 

```
error: internal compiler error: Kani unexpectedly panicked at panicked at kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs:46:14:
                                called `Option::unwrap()` on a `None` value.

thread 'rustc' panicked at kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs:46:14:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: core::option::Option<T>::unwrap
             at /rustc/60d146580c10036ce89e019422c6bc2fd9729b65/library/core/src/option.rs:967:21
   5: kani_compiler::codegen_cprover_gotoc::codegen::contract::<impl kani_compiler::codegen_cprover_gotoc::context::goto_ctx::GotocCtx>::handle_check_contract
             at /Users/cmzech/kani/kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs:35:22
   6: kani_compiler::codegen_cprover_gotoc::compiler_interface::GotocCodegenBackend::codegen_items::{{closure}}::{{closure}}
[...]

PR #3609, opened 2024-10-17
Emit an error when proof_for_contract function is not found

zhassan-aws:
Currently, Kani panics if the function specified in the `proof_for_contract` attribute is not found (e.g. because the function is not reachable) (see #3467). This PR adds an error message pointing out the issue.

Towards #3467 


celinval (approved):
I wonder if this should be done as part of `kani::middle::check_reachable_items` which is where we run our validations. The main advantage is that we emit all possible configuration errors.

I won't block this PR though since it's a clear improvement to the existing flow. Thanks!

zhassan-aws:
> I wonder if this should be done as part of `kani::middle::check_reachable_items` which is where we run our validations

That makes sense. I'll check if it's easy to do that.

zhassan-aws:
I moved the check to `check_reachable_items`.

celinval (approved):
Thanks!!!

Issue #3585, opened 2024-10-09
Tracking Issue: Address follow-up comments on #3514 (LLBC Backend)


zhassan-aws:
This is a tracking issue for addressing the comments on #3514. 

- [x] Use "Lean" instead of "Aeneas" for any user-visible feature
- [x] Upgrade Charon to a more recent commit
- [ ] Once we start calling Aeneas in the Kani driver, the driver should automatically delete the .llbc files
- [ ] Clean-up the handling of names
- [ ] Use "public functions" mode as opposed to "harness" mode
- [ ] Minimize code duplication between goto compiler interface and llbc compiler interface
- [ ] Generate generic code (using `CrateItem`) instead of monomorphized code (that uses `Instance`)

ssyram:
Hi, I appreciate your work very much but am personally curious about what is the intended difference between your work and Charon? I think you are both converting / exporting Rust codes to LLBC?

zhassan-aws:
Hi @ssyram. We wanted to add the ability to export LLBC in Kani to allow interaction with the CBMC backend, e.g. through function contracts (which Kani supports).

ssyram:
Thanks for this! From my understanding, the codegen of LLBC is parallel to the GOTO format -- the codegen either generates GOTO to be put into CBMC or LLBC. Am I correct? And where will the interaction happen and how is it supposed to happen?

zhassan-aws:
Correct. The plan was for the interaction to be via function contracts where the CBMC backend is used for lower-level functions (e.g with unsafe) and LLBC is used higher up.

PR #3630, opened 2024-10-22
[Lean] Rename user-facing options from Aeneas to Lean

zhassan-aws:
This is a follow-up on #3514. As @celinval suggested ([here](https://github.com/model-checking/kani/pull/3514#discussion_r1790949304)), renaming all user-facing options from "Aeneas" to Lean.

Inside the Kani compiler which is only concerned with producing LLBC (and doesn't know about Lean/Aeneas), I'm renaming the relevant feature from `aeneas` to `llbc`, and the backend is now called the LLBC backend as opposed to the Aeneas backend.

Towards #3585 

### DIFF
--- a/kani-compiler/src/args.rs
+++ b/kani-compiler/src/args.rs
@@ -7,15 +7,15 @@ use tracing_subscriber::filter::Directive;
 #[derive(Debug, Default, Display, Clone, Copy, AsRefStr, EnumString, VariantNames, PartialEq, Eq)]
 #[strum(serialize_all = "snake_case")]
 pub enum BackendOption {
-    /// Aeneas (LLBC) backend
-    #[cfg(feature = "aeneas")]
-    Aeneas,
-
     /// CProver (Goto) backend
     #[cfg(feature = "cprover")]
     #[strum(serialize = "cprover")]
     #[default]
     CProver,
+
+    /// LLBC backend (Aeneas's IR)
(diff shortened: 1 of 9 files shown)


kani-compiler/src/args.rs
```diff
@@ -87,7 +87,7 @@ pub struct Arguments {
     /// Option name used to select which backend to use.
     #[clap(long = "backend", default_value_t = BackendOption::CProver)]
     pub backend: BackendOption,
-    /// Print the final LLBC file to stdout. This requires `-Zaeneas`.
+    /// Print the final LLBC file to stdout.
```
celinval, line 90: Why did you remove this comment?

celinval (approved):
Thanks!

kani-compiler/src/args.rs
zhassan-aws: My understanding is that those requires comments should be in the kani-driver help, and not in the Kani compiler one and the kani-driver option already mentions the requirement:

https://github.com/model-checking/kani/pull/3630/files#diff-6f15b74109fec7be3b13bdc2f8c01481d82be82f413c39a7f54d53f1990c6454R280

None of the other Kani compiler options that require an unstable flag (e.g. `--check_coverage`, `--ub-check`) mention the unstable flag requirement in their help message.

PR #3640, opened 2024-10-24
Fix loop contracts transformation when loops in branching

qinheping:
In loop contracts transformation, we want to visit non-loop blocks before visiting loop blocks. So we pop `loop_queue` only when `queue` is empty. We should use `or_else` instead of `or` which always pop the front of `loop_queue`.

### DIFF
--- a/kani-compiler/src/kani_middle/transform/loop_contracts.rs
+++ b/kani-compiler/src/kani_middle/transform/loop_contracts.rs
@@ -114,7 +114,7 @@ impl TransformPass for LoopContractPass {
                     let mut loop_queue: VecDeque<BasicBlockIdx> = VecDeque::new();
                     queue.push_back(0);
 
-                    while let Some(bb_idx) = queue.pop_front().or(loop_queue.pop_front()) {
+                    while let Some(bb_idx) = queue.pop_front().or_else(|| loop_queue.pop_front()) {
                         visited.insert(bb_idx);
 
                         let terminator = new_body.blocks()[bb_idx].terminator.clone();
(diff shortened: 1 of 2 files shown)


tests/expected/loop-contract/multiple_loops.rs
celinval, line 1: why did you have to change this test?

zhassan-aws: I would've preferred to add a separate case than to modify an existing one. But given that there is another test that is very similar to the existing one (https://github.com/model-checking/kani/blob/main/tests/expected/loop-contract/simple_while_loop.rs), it is fine to change this one.

zhassan-aws (approved):

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -114,7 +114,7 @@ impl TransformPass for LoopContractPass {
                     let mut loop_queue: VecDeque<BasicBlockIdx> = VecDeque::new();
                     queue.push_back(0);
 
-                    while let Some(bb_idx) = queue.pop_front().or(loop_queue.pop_front()) {
+                    while let Some(bb_idx) = queue.pop_front().or_else(|| loop_queue.pop_front()) {
```
zhassan-aws, line 117: Is this change necessary?

qinheping: Yes, with `or` the `loop_queue` will also `pop_front` at each iteration, while we want it to `pop_front` only when `queue` is empty.

PR #3641, opened 2024-10-24
Build and include `kani-cov` in the bundle

adpaco-aws:
This PR adds another step to the `cargo bundle` command so that `kani-cov` is built in release mode and then included in the bundle. This will allow users to call `kani-cov` once Kani has been setup.

I have manually tested it works by running:

```
cargo bundle
tar -xzf kani-0.56.0-x86_64-unknown-linux-gnu.tar.gz
ls kani-0.56.0/bin/
```

This showed `kani-cov` along with other binaries like `cbmc` and `kissat`.

### DIFF
--- a/tools/build-kani/src/main.rs
+++ b/tools/build-kani/src/main.rs
@@ -10,7 +10,9 @@
 mod parser;
 mod sysroot;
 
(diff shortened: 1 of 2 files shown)


zhassan-aws (approved):
Thanks

PR #3642, opened 2024-10-24
Call `goto-instrument` with `DFCC` only once

qinheping:
DFCC should be called only once. This PR combined the two DCFF calls into one.

### DIFF
--- a/kani-driver/src/call_goto_instrument.rs
+++ b/kani-driver/src/call_goto_instrument.rs
@@ -37,16 +37,12 @@ impl KaniSession {
             self.goto_sanity_check(output)?;
         }
 
-        if self
+        let is_loop_contracts_enabled = self
             .args
             .common_args
             .unstable_features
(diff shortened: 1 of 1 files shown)


kani-driver/src/call_goto_instrument.rs
```diff
@@ -37,15 +37,15 @@ impl KaniSession {
...
         if self
             .args
             .common_args
             .unstable_features
             .contains(kani_metadata::UnstableFeature::LoopContracts)
         {
-            self.instrument_loop_contracts(harness, output)?;
+            self.instrument_contracts_with_loop_contracts(harness, output)?;
+        } else {
+            self.instrument_contracts(harness, output)?;
```
celinval, line 48: Can you merge these functions and take a boolean on whether to initialize loop contracts?

celinval (approved):
Just a minor suggestion. Not a blocker.

PR #3646, opened 2024-10-25
Move any_slice_from_array to kani_core

qinheping:
Move `any_slice_from_array` to `kani_core` so that we can call them in verify-rust-std (an example harness in https://github.com/model-checking/verify-rust-std/pull/122).


celinval (changes requested):
Is there a reason why this have to be a non-backward compatible change?

qinheping:
@celinval I was trying to make the usage of `any_slice` consistent with `arbitrary_ptr`, and didn't realize the issue of non-backward compatibility. I agree that keeping them backward compatible is more preferable.

celinval (approved):
Thanks!

PR #3649, opened 2024-10-25
Add a timeout option

zhassan-aws:
This PR adds a new option, `--cbmc-timeout` which can be used for setting a timeout for the CBMC run.

The timeout applies on a per-harness basis in the Kani run.

The implementation uses [tokio ](https://tokio.rs/) to perform non-blocking reads from the CBMC process's `stdout` so that the timeout can be checked asynchronously.

**Call-outs:**

There are other names I considered for the option:
- `--harness-timeout`: This makes it clearer that the timeout applies to each individual harness (and not to the overall Kani run). It doesn't indicate that the timeout only applies to the CBMC portion of the flow though.
- `--engine-timeout`: This makes it a bit more general across multiple engine/backends. The notion of an engine is not really part of the Kani jargon currently though, so might not be clear to users.


kani-driver/src/args/mod.rs
```diff
@@ -281,6 +281,10 @@ pub struct VerificationArgs {
     #[arg(long, hide = true)]
     pub print_llbc: bool,
 
+    /// Timeout for CBMC command in seconds
+    #[arg(long)]
+    pub cbmc_timeout: Option<u32>,
```
celinval, line 286: I prefer something more generic like `--harness-timeout`. The timeout implemented in this PR is not CBMC specific, right?

kani-driver/src/args/mod.rs
```diff
@@ -281,6 +281,10 @@ pub struct VerificationArgs {
     #[arg(long, hide = true)]
     pub print_llbc: bool,
 
+    /// Timeout for CBMC command in seconds
```
celinval, line 284: Should we make this unstable for now? Require `-Z unstable-options`

kani-driver/src/call_cbmc.rs
```diff
@@ -45,6 +48,15 @@ pub enum FailedProperties {
     Other,
 }
 
+/// The possible CBMC exit statuses
+#[derive(Clone, Copy, Debug)]
+pub enum CbmcExitStatus {
```
celinval, line 53: Is that really CBM specific?

kani-driver/src/args/mod.rs
celinval: Also, I was wondering if we could accept an optional suffix like the `timeout` command.

celinval (commented):
Thanks!!

kani-driver/src/args/mod.rs
zhassan-aws: Sure. Yes, the implementation is meant to apply to any command.

kani-driver/src/args/mod.rs
zhassan-aws: > Should we make this unstable for now? Require `-Z unstable-options`

Sure.

> Also, I was wondering if we could accept an optional suffix like the `timeout` command.

Yes, that would be convenient.

kani-driver/src/call_cbmc.rs
zhassan-aws: No, I'll update the name.

zhassan-aws:
The changes seems to have slowed down Kani (regressions have been running for over an hour). Turning the PR into a draft till I investigate.

zhassan-aws:
Addressed the performance issue in https://github.com/model-checking/kani/pull/3649/commits/0502cc20ac1296e10f0e7d4ad0a13c527f22f3a6.

kani-driver/src/args/mod.rs
zhassan-aws: > > Also, I was wondering if we could accept an optional suffix like the `timeout` command.
> 
> Yes, that would be convenient.

Actually, can I do this in a follow-up PR? It's not difficult to implement, but will require a number of tests to cover the different scenarios (including negative tests).

zhassan-aws: Never mind; added support for a unit suffix in https://github.com/model-checking/kani/pull/3649/commits/c96ea92a87dec18a485be6ba36dfd6b3bf6ca804

kani-driver/src/call_cbmc.rs
```diff
@@ -87,18 +97,70 @@ impl KaniSession {
...
-            })?;
+            })
+            .await)
+        };
 
+        let verification_results = if res.is_err() {
+            // An error occurs if the timeout was reached
+
+            // Kill the process
+            cbmc_process.kill().await?;
```
tautschnig, line 151: It might be difficult to test this in CI, but could you please locally run a test with `--cbmc-args --external-solver cadical` (or whichever solver you like), and perhaps also with `--cbmc-args --z3` on examples that time out and require killing? I am asking for this as CBMC does some signal handling and I have previously seen issues where we couldn't reliably kill CBMC or its child processes.

zhassan-aws: I tested this scenario with cadical and z3.

For cadical, both the CBMC and cadical processes were terminated right after reaching the timeout.

For z3, the CBMC process was terminated, but the z3 one continued to run, apparently till completion.

I'm not sure if there's anything that can be done from the Kani side to deal with the orphaned z3 process.

tautschnig: Would you mind filing an issue against CBMC with this observation? I agree this isn’t for Kani to fix, but perhaps we will finally manage to settle this problem on the CBMC side for this scenario looks like one that is finally one that is reproducible. Thank you!

zhassan-aws: Never mind, I was able to reproduce this with CBMC alone. The crucial piece is to send SIGKILL to the CBMC process (which is what [`Child::kill`](https://docs.rs/tokio/latest/tokio/process/struct.Child.html#method.kill) does) and not SIGTERM.

I filed https://github.com/diffblue/cbmc/issues/8485

kani-driver/src/args/mod.rs
```diff
@@ -62,6 +63,58 @@ pub fn print_deprecated(verbosity: &CommonArgs, option: &str, alternative: &str)
 // By default we configure CBMC to use 16 bits to represent the object bits in pointers.
 const DEFAULT_OBJECT_BITS: u32 = 16;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
celinval, line 66: You should be able to use strum here instead of implementing FromStr manually: https://docs.rs/strum/latest/strum/derive.EnumString.html

celinval (approved):
Awesome! Thanks

kani-driver/src/args/mod.rs
zhassan-aws: Done in https://github.com/model-checking/kani/pull/3649/commits/482ed1130f4f75c1aa90896092758c2db0250301

Thanks for the suggestion.

celinval (approved):

PR #3660, opened 2024-10-30
Add support for float_to_int_unchecked

### LINKED ISSUE #3629: Add support for `float_to_int_unchecked`
Hi @carolynzech @celinval @feliperodri @zhassan-aws 

When I was testing my harness for `f32::to_int_unchecked`, I encountered the error indicating that **float_to_int_unchecked is not currently supported by Kani**, as shown below. Is it possible to support it?
```
SUMMARY:
 ** 1 of 1277 failed (1276 undetermined)
Failed Checks: float_to_int_unchecked is not currently supported by Kani. Please post your example at https://github.com/model-checking/kani/issues/new/choose
 File: "/Users/yew005/Docs/Academic/CMU/Fall24/practicum/verify-rust-std/library/core/src/convert/num.rs", line 30, in <f32 as convert::num::FloatToInt<i32>>::to_int_unchecked

VERIFICATION:- FAILED
** WARNING: A Rust construct that is not currently supported by Kani was found to be reachable. Check the results for more details.
Verification Time: 5.8839436s

Summary:
Verification failed for - num::verify::checked_to_int_unchecked_f32
Complete - 0 successfully verified harnesses, 1 failures, 1 total.
```

`library/core/src/convert/num.rs` Line 20 to 35:
```rust
20 macro_rules! impl_float_to_int {
21     ($Float:ty => $($Int:ty),+) => {
22         #[unstable(feature = "convert_float_to_int", issue = "67057")]
23         impl private::Sealed for $Float {}
24         $(
25             #[unstable(feature = "convert_float_to_int", issue = "67057")]
26             impl FloatToInt<$Int> for $Float {
27                 #[inline]
28                 unsafe fn to_int_unchecked(self) -> $Int {
29                     // SAFETY: the safety contract must be upheld by the caller.
30                     unsafe { crate::intrinsics::float_to_int_unchecked(self) }
31                 }
32            }
33         )+
34     }
35 }
```

Test harness:
```rust
#[kani::proof_for_contract(f32::to_int_unchecked)]
pub fn checked_to_int_unchecked_f32() {
    let num1: f32 = kani::any::<f32>();

    let result = unsafe { num1.to_int_unchecked::<i32>() };

    assert_eq!(result, num1 as i32);
}
```

Contracts added to `f32::to_int_unchecked` (in `library/core/src/num/f32.rs`):
```rust
/// # Safety
///
/// The value must:
///
/// * Not be `NaN`
/// * Not be infinite
/// * Be representable in the return type `Int`, after truncating off its fractional part
/// ...
#[requires(!self.is_nan() && !self.is_infinite())]
#[requires(self >= Self::MIN && self <= Self::MAX)]
pub unsafe fn to_int_unchecked<Int>(self) -> Int
where
    Self: FloatToInt<Int>,
{ ... }
```

Thank you very much!

_Originally posted by @Yenyun035 in https://github.com/model-checking/verify-rust-std/issues/59#issuecomment-2428184293_

zhassan-aws:
This PR adds support for the [`float_to_int_unchecked`](https://doc.rust-lang.org/std/intrinsics/fn.float_to_int_unchecked.html) intrinsic for `f32` and `f64`.

Towards #3629 

Keeping it as draft till I add more tests.


kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1929,3 +1993,150 @@ fn instance_args(instance: &Instance) -> GenericArgs {
...
+}
+
+fn get_lower_upper_f64_uint(uint_ty: UintTy, mm: &MachineModel) -> (f64, f64) {
+    let lower = -1.0;
+    let upper = match uint_ty {
+        UintTy::U8 => (1u128 << 8) as f64,
+        UintTy::U16 => (1u128 << 16) as f64,
+        UintTy::U32 => (1u128 << 32) as f64,
+        UintTy::U64 => (1u128 << 64) as f64,
+        UintTy::U128 => unimplemented!(),
```
tautschnig, line 2062: Would `(1u128 << 64) as f64 * (1u128 << 64) as f64` work for this case?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1929,3 +1993,150 @@ fn instance_args(instance: &Instance) -> GenericArgs {
...
+fn get_lower_upper_f32_int(int_ty: IntTy, mm: &MachineModel) -> (f32, f32) {
+    let lower = match int_ty {
+        IntTy::I8 => -129.0,
+        IntTy::I16 => -32769.0,
+        IntTy::I32 => -2147483904.0,
+        IntTy::I64 => -9223373136366403584.0,
+        IntTy::I128 => -170141203742878835383357727663135391744.0,
+        IntTy::Isize => match mm.pointer_width {
+            32 => -2147483904.0,
+            64 => -9223373136366403584.0,
```
tautschnig, line 2097: The above comment says how the value can _not_ be computed, but can you perhaps also include a comment that does explain how you found those constants?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1929,3 +1993,150 @@ fn instance_args(instance: &Instance) -> GenericArgs {
...
+    (lower, upper)
+}
+
+fn get_lower_upper_f64_int(int_ty: IntTy, mm: &MachineModel) -> (f64, f64) {
+    let lower = match int_ty {
+        IntTy::I8 => -129.0,
+        IntTy::I16 => -32769.0,
+        IntTy::I32 => -2147483649.0,
+        IntTy::I64 => -9223372036854777856.0,
+        IntTy::I128 => unimplemented!(),
```
tautschnig, line 2122: I suppose this value can be computed?

tautschnig (approved):

tautschnig:
Please don't forget to update `docs/src/rust-feature-support/intrinsics.md`.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
zhassan-aws: I hard coded the value.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
zhassan-aws: I found most of the values using a script, and others using the `next_up()` and `next_down()` methods. I'll try to clean up the script and perhaps commit it somewhere.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
zhassan-aws: Yes, I added it.

zhassan-aws:
> Please don't forget to update `docs/src/rust-feature-support/intrinsics.md`.

Thanks for the reminder. Added.

zhassan-aws:
This PR could use quite a bit of cleanup, especially around hard-coded values and unit tests. In particular, it's probably better to store the hard-coded values using their byte representation as opposed to their decimal one.

celinval (commented):
Should we be worried about the cast using `as` logic?

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1929,3 +1993,286 @@ fn instance_args(instance: &Instance) -> GenericArgs {
...
+fn get_bounds_f32_uint(uint_ty: UintTy, mm: &MachineModel) -> (f32, f32) {
+    let lower: f32 = -1.0;
+    let upper: f32 = match uint_ty {
+        UintTy::U8 => (1u128 << 8) as f32,   // 256.0
+        UintTy::U16 => (1u128 << 16) as f32, // 65536.0
+        UintTy::U32 => (1u128 << 32) as f32, // 4294967296.0
+        UintTy::U64 => (1u128 << 64) as f32, // 18446744073709551616.0
+        UintTy::U128 => f32::INFINITY,       // all f32's fit in u128!
+        UintTy::Usize => match mm.pointer_width {
+            32 => (1u128 << 32) as f32,
```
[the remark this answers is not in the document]
zhassan-aws: That looks neater, but I couldn't get it work because apparently the guard cannot be applied to only one of the patterns: https://stackoverflow.com/questions/42355502/is-it-possible-to-combine-two-patterns-one-with-a-match-guard-in-the-same-matc

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1929,3 +1993,286 @@ fn instance_args(instance: &Instance) -> GenericArgs {
...
+    match float_ty {
+        FloatTy::F32 => {
+            let (lower, upper) = get_bounds_f32_uint(uint_ty, mm);
+            (Expr::float_constant(lower), Expr::float_constant(upper))
+        }
+        FloatTy::F64 => {
+            let (lower, upper) = get_bounds_f64_uint(uint_ty, mm);
+            (Expr::double_constant(lower), Expr::double_constant(upper))
+        }
+        _ => unimplemented!(),
```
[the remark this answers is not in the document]
zhassan-aws: Do you prefer I do this in this PR or in a follow-up one?

celinval:
I'm really sorry, I think you might've misunderstood my comment about the `as`. I think we should be using `as` in the kani-compiler code since this is getting rustc's values as the source of truth.

My comment about the `as` is about Kani's implementation of casting float to integer using `as`. I would expect the same bounds to be used for both, but it sounds like we don't.

kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
```diff
@@ -1929,3 +1993,279 @@ fn instance_args(instance: &Instance) -> GenericArgs {
...
+
+const F32_U_LOWER: [u8; 4] = [0x00, 0x00, 0x80, 0xBF]; // -1.0
+const F32_U8_UPPER: [u8; 4] = [0x00, 0x00, 0x80, 0x43]; // 256.0
+const F32_U16_UPPER: [u8; 4] = [0x00, 0x00, 0x80, 0x47]; // 65536.0
+const F32_U32_UPPER: [u8; 4] = [0x00, 0x00, 0x80, 0x4F]; // 4294967296.0
+const F32_U64_UPPER: [u8; 4] = [0x00, 0x00, 0x80, 0x5F]; // 18446744073709551616.0
+// The largest f32 value fits in a u128, so there is no upper bound
+const F32_U128_UPPER: [u8; 4] = [0x00, 0x00, 0x80, 0x7F]; // inf
+
+const F32_I8_LOWER: [u8; 4] = [0x00, 0x00, 0x01, 0xC3]; // -129.0
```
celinval, line 2033: Sorry for the misunderstanding, but can we use the old logic as before? I think it is fine to define them as constants, but I prefer using the conversion logic, for example: `i8::MIN as f32 - 1.0`

zhassan-aws:
> I'm really sorry, I think you might've misunderstood my comment about the `as`. I think we should be using `as` in the kani-compiler code since this is getting rustc's values as the source of truth.

Not at all. I was planning to make this change before your comment. The reason is that casting a decimal value into a floating-point value can have surprising behavior. For example, [this program](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5ab279cd1c148158b6148339a767e56f):
```rust
    println!("{}", u32::MAX);
    let f1: f32 = u32::MAX as f32;
    let f2: f32 = u32::MAX as f32 + 1.0;
    let f3: f32 = (u32::MAX as u128 + 1) as f32;
    
    println!("{:.32}", f1);
    println!("{:.32}", f2);
    println!("{:.32}", f3);
```
prints:
```
4294967295
4294967296.00000000000000000000000000000000
4294967296.00000000000000000000000000000000
4294967296.00000000000000000000000000000000
```
So casting the decimal value `u32::MAX` (4294967295) to a float results in a different number. This is because the cast operation rounds to the nearest number that can be represented as an `f32`. In other words, when assigning an `f32` a decimal value, the actual value it stores may end up being different (could be smaller or larger). This calculator, which @tautschnig pointed out to me helped me determine the correct bounds.

One way to guarantee a specific value is to use the byte representation of the floating-point number. This is the reason I switched to it.

zhassan-aws:
I made a few final touches:
1. Moved most of the code to a new `float_utils` module to avoid polluting `intrinsics.rs` (Thanks @celinval for the suggestion)
2. I added the code I used to determine the bounds as a unit test. If run with `--nocapture`, it'll print the constants.
3. I added a small optimization that skips the comparison if the bound is +/- infinity. 

@tautschnig @celinval Let me know if you want to take another look. Otherwise, I'll go ahead and merge.

celinval (approved):
It's fine by me. Thanks!

PR #3675, opened 2024-11-03
codegen: Ask the layout if it is uninhabited, not its impl detail

workingjubilee:
Directly question the layout if it is uninhabited. This makes Kani more resilient to changes in the upstream rustc toolchain, as the convenience function which exposes this is precisely to allow compilers to not fixate as much on the innermost details of rustc's codegen process.

Found while reviewing https://github.com/model-checking/kani/pull/3671


celinval (approved):

celinval:
Thanks!

PR #3644, opened 2024-10-24
Fix codegen for rvalue aggregate raw pointer to an adt with slice tail

carolynzech:
Fix the codegen for a raw pointer to a dynamically-sized ADT. Traverse the type until we find the tail, which will either be a trait object or a slice, and generate the pointer accordingly.

Resolves #3638, #3615

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #3687, opened 2024-11-06
Fix issues with how we compute DST size

### LINKED ISSUE #3612: Memory predicates won't work if function with contract is compiled as a dependency
Let's say you have a crate `crate_a` that defines a `copy_val` function that is annotated with contracts:
```rust
///----- Crate with contract
#[requires(kani::mem::can_dereference(ptr))]
pub unsafe fn copy_val<T: Copy>(ptr: *const T) -> T {
}
```
Now we have a crate `crate_b` that uses `crate_a` that we want to verify using Kani:

```rust
pub fn safe_copy<T: Copy>(x: &T) -> T {
    unsafe { copy_val(x) }
}
```

```
cargo kani --only-codegen
```

with Kani version: 0.56.0

I expected to see this happen: Compilation should succeed.

Instead, this happened: Kani will fail to compile `crate_a` due to missing trait bounds.

### LINKED ISSUE #3616: Kani miss overflow when computing `size_of_val`
I tried this code:

```rust
#[kani::proof]
pub fn check_size_of_overflows() {
    let var:  [u64; 4] = kani::any();
    let fat_ptr: *const [u64] = &var as *const _;
    let (thin_ptr, size) = fat_ptr.to_raw_parts();
    let new_size: usize = kani::any();
    let new_ptr:  *const [u64] = ptr::from_raw_parts(thin_ptr, new_size);
    if let Some(expected_size) = new_size.checked_mul(size_of::<u64>()) {
        assert_eq!(unsafe { size_of_val_raw(new_ptr) }, expected_size);
    } else {
        // Expect UB detection
        let _should_ub = unsafe { size_of_val_raw(new_ptr) };
    }
}
```

```
kani slice.rs
```

with Kani version: 0.56.0

I expected to see this happen: UB should be detected

### LINKED ISSUE #3627: Memory predicates size computation is incorrect for ADTs with non-sized tail
I tried this code:

```rust
extern crate kani;

use kani::mem::{can_dereference, can_write};

#[derive(Clone, Copy, kani::Arbitrary)]
struct Wrapper<T: ?Sized> {
    _size: usize,
    _value: T,
}

mod invalid_access {
    use super::*;
    use std::ptr;

    #[kani::proof]
    pub fn check_wrong_size_should_fail() {
        let mut var: Wrapper<[u64; 4]> = kani::any();
        let fat_ptr: *mut Wrapper<[u64]> = &mut var as *mut _;
        let (thin_ptr, size) = fat_ptr.to_raw_parts();
        let new_size: usize = size + 1;
        let new_ptr: *const [u64] = ptr::from_raw_parts(thin_ptr, new_size);
        // This should fail, however today we don't take the sized members into consideration.
        assert!(can_dereference(new_ptr));
    }

}
```

```
kani -Z mem-predicates wrapper.rs
```

with Kani version: 0.56.0

I expected to see this happen: Verification failed

celinval:
This change fixes how we compute size of the object in our mem predicates, and provide user visible methods to try to retrieve the size of the object if known and valid (`checked_size_of_raw` and `checked_align_of_raw`.

Fixes #3612
Fixes #3627

## Call-outs

To simplify this PR, I moved the following changes to their own PRs:
1. #3644
2. #3718

I also removed the fix for the intrinsics `size_of_val` and `align_of_val` from this PR, and I will create a follow up PR once this one is merged.


kani-compiler/src/kani_middle/abi.rs
```diff
@@ -0,0 +1,193 @@
...
+            RigidTy::Adt(adt_def, adt_args) => {
+                // We have to recurse and get the maximum alignment of all sized portions.
+                let fields = adt_def.variants_iter().next().unwrap().fields();
+                let fields_sorted = self.layout.fields.fields_by_offset_order();
+                let last_field_idx = *fields_sorted.last().unwrap();
+                LayoutOf::new(fields[last_field_idx].ty_with_args(adt_args))
+            }
+            RigidTy::Tuple(tys) => {
+                // We have to recurse and get the maximum alignment of all sized portions.
+                let last_ty = tys.last().expect("Expected unsized tail");
```
celinval, line 185: Note to self: Fix this. The order of fields is not guaranteed, and we need to use similar logic as the `Adt` logic to get the last field.

kani-compiler/src/kani_middle/transform/rustc_intrinsics.rs
```diff
@@ -0,0 +1,112 @@
...
+                func.ty(&self.locals).unwrap().kind()
+            {
+                if def.is_intrinsic() {
+                    let instance = Instance::resolve(def, &args).unwrap();
+                    let intrinsic = Intrinsic::from_instance(&instance);
+                    debug!(?intrinsic, "handle_terminator");
+                    let model = match intrinsic {
+                        Intrinsic::SizeOfVal => self.models[&KaniModel::SizeOfVal],
+                        Intrinsic::MinAlignOfVal => self.models[&KaniModel::AlignOfVal],
+                        // The rest is handled in hooks.
```
celinval, line 96:
```suggestion
                        // The rest is handled in codegen.
```

celinval:
FYI, I'm going to create a fix for the std in a separate PR. Also I think I could break down this PR into two if needed.

celinval:
After a second thought, I'm going to try to break this change into the following PRs:
1. Aggregate fix + Abi module
2. Refactoring of of kani_functions
3. Adding the checked APIs and fix of men predicates
4. Fix intrinsics implementation

Note that 1 and 2 are fairly independent, but 3 will depend on 1 and 2, while 4 will depend on them all.

tests/kani/MemPredicates/foreign_type.rs
```diff
@@ -0,0 +1,64 @@
...
+    let ptr = &var as *const _ as *const __CPROVER_size_t;
+    assert!(!can_dereference(ptr));
+    assert!(!can_read_unaligned(ptr));
+}
+
+/// Kani APIs cannot tell if that's safe to write to a foreign type.
+///
+/// However, foreign APIs that have knowledge of the type can still safely set new values.
+#[kani::proof]
+#[kani::should_panic]
```
carolynzech, line 47: I don't understand this test. If we're trying to see if the foreign API can safely set new values, doesn't it defeat the point to make the harness panic? Shouldn't we do the write and see that it succeeds, and then perhaps do the failing read in a separate test?

carolynzech (commented):
I don't understand what the safety check hook is for--as far as I can tell, we never invoke it. How does it fit into the size and alignment fixes you're doing? Did you forget to invoke it somewhere, or am I missing something?
Otherwise, all of these comments are just nits. Beautiful tests!

carolynzech (approved):

tests/kani/MemPredicates/foreign_type.rs
celinval: I can split this into two tests as you suggested. My initial thought was just to check that the value has changed by checking if it is still 0.

library/kani_core/src/models.rs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            /// Compute the size of a slice or object with a slice tail.
+            ///
+            /// Most information are extracted by the Kani compiler at compilation time,
+            /// except for the length of the slice.
+            /// Thus, alignment should be safe to use (power-of-two and smaller than isize::MAX).
+            ///
+            /// Thus, this generate the logic to ensure the size computation does not
+            /// does not overflow and it is smaller than `isize::MAX`.
```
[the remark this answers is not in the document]
celinval:
```suggestion
            /// The slice length may be a symbolic value which is computed at runtime.
            /// All the other inputs are extracted and validated by Kani compiler,
            /// i.e., these are well known concrete values that should be safe to use.
            /// Example, align is a power-of-two and smaller than isize::MAX.
            ///
            /// Thus, this generate the logic to ensure the size computation does not
            /// does not overflow and it is smaller than `isize::MAX`.
```

library/kani_core/src/mem.rs
```diff
@@ -203,92 +211,17 @@ macro_rules! kani_mem {
...
-            }
-
-            fn min_alignment(&self, _: Internal) -> usize {
-                self.align_of()
-            }
-
-            fn dangling(&self, _: Internal) -> *const () {
-                NonNull::<&T>::dangling().as_ptr() as _
+        // Return whether the pointer is aligned
+        #[allow(clippy::manual_is_power_of_two)]
```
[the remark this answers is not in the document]
celinval: trying to make sure CBMC can do constant propagation whenever is possible

PR #3694, opened 2024-11-07
Apply loop contracts only if there exists some usage

qinheping:
In this PR, we detect the usage of loop contracts, and apply loop contracts with goto-instrument only if there exists some usage of loop contracts in the codegen unit.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
@@ -576,25 +576,43 @@ impl GotocHook for LoopInvariantRegister {
 
         gcx.has_loop_contracts = true;
 
-        // Add `free(0)` to make sure the body of `free` won't be dropped to
-        // satisfy the requirement of DFCC.
-        Stmt::block(
-            vec![
-                BuiltinFn::Free
-                    .call(vec![Expr::pointer_constant(0, Type::void_pointer())], loc)
-                    .as_stmt(loc),
-                unwrap_or_return_codegen_unimplemented_stmt!(
-                    gcx,
-                    gcx.codegen_place_stable(assign_to, loc)
-                )
-                .goto_expr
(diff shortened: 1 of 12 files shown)


celinval (commented):
Do we have any tests that check what is Kani's behavior when a loop contract is reachable but user did not enable it? If not, can you please add one.

qinheping:
@celinval I added two tests in this PR: one for harness without loop but loop contracts is enabled. One for harness with loop contracts, but loop contracts is not enabled. In both harness, Kani performs as no loop contracts present---unwind all loops.

We should apply loop contracts only when there exist some usages of loop contracts **and** the loop contracts is enabled by users.

I also fixed a bug in `hook.rs` of not correctly handling `loop_contracts_register_function` when loop contracts is not enabled.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -567,25 +567,52 @@ impl GotocHook for LoopInvariantRegister {
...
-                    .call(vec![Expr::pointer_constant(0, Type::void_pointer())], loc)
-                    .as_stmt(loc),
-                Stmt::goto(bb_label(target.unwrap()), loc).with_loop_contracts(
-                    func_exp.call(fargs).cast_to(Type::CInteger(CIntType::Bool)),
-                ),
-            ],
-            loc,
-        )
+
+        gcx.has_loop_contracts = true;
```
celinval, line 577: Should this be inside the `if` block?

celinval (approved):
Thanks

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
qinheping: No, the hook is applied when there are calls to `register_loop_contracts` in MIR, which means there are loop contracts in the harness. The if block is about whether the loop contracts feature of Kani is enabled.

PR #3695, opened 2024-11-08
Remove symtab json support

celinval:
I believe the plan was to keep this in case of a regression for about a year. The time has long gone, so I think it's time to get rid of it.


zhassan-aws (approved):
Thanks!

PR #3710, opened 2024-11-11
Upgrade toolchain to nightly-2024-11-11

qinheping:
Changes required due to
- https://github.com/rust-lang/rust/pull/132675 (coverage: Restrict empty-span expansion to only cover { and })


kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs
```diff
@@ -286,4 +292,11 @@ pub mod rustc_smir {
             CovTerm::Zero
         }
     }
+
+    pub fn span_file_name(tcx: TyCtxt<'_>, span: Span) -> Symbol {
```
carolynzech, line 296: Instead of copying this over from rustc, could you try converting `body_span` into a StableMIR span and then call `get_filename`? I think we should only paste a private function from rustc into Kani as a last resort.

qinheping: Updated in the latest commit.

carolynzech (approved):

PR #3726, opened 2024-11-18
Enable contracts for const generic functions

### LINKED ISSUE #3667: Cannot annotate a const generic function with a contract
This issue was reported in https://github.com/model-checking/verify-rust-std/discussions/143.

I tried this code:

```rust
#[kani::requires(true)]
const fn bar<T>(_v: T)
where T: Sized
{
}
```

```
kani s.rs -Zfunction-contracts
```

with Kani version: 1fe80f9c5b2

```
$ kani s.rs -Zfunction-contracts
Kani Rust Verifier 0.56.0 (standalone)
error[E0493]: destructor of `T` cannot be evaluated at compile-time
 --> s.rs:2:17
  |
1 | #[kani::requires(true)]
  |                       - value is dropped here
2 | const fn bar<T>(_v: T)
  |                 ^^ the destructor for this type cannot be evaluated in constant functions

For more information about this error, try `rustc --explain E0493`.
error: /home/ubuntu/git/kani/target/kani/bin/kani-compiler exited with status exit status: 101
```

qinheping:
This PR adds dummy assignments of the input variables to local variables in contracts-replace-closure to avoid may-drop checks in const generic functions.

Resolves #3667 


library/kani_macros/src/sysroot/contracts/check.rs
```diff
@@ -175,14 +175,14 @@ impl<'a> ContractConditionsHandler<'a> {
...
     /// This is used so Kani doesn't think that modifying a local argument value is a side effect.
-    fn arg_redefinitions(&self) -> TokenStream2 {
+    pub fn arg_redefinitions(&self) -> TokenStream2 {
         let mut result = TokenStream2::new();
         for (mutability, ident) in self.arg_bindings() {
             if mutability == MutBinding::Mut {
                 result.extend(quote!(let mut #ident = #ident;))
             } else {
                 // This would make some replace some temporary variables from error messages.
-                //result.extend(quote!(let #ident = #ident; ))
```
zhassan-aws, line 185: Remove comment.

zhassan-aws (approved):
Nice, thanks!

library/kani_macros/src/sysroot/contracts/replace.rs
```diff
@@ -70,6 +70,7 @@ impl<'a> ContractConditionsHandler<'a> {
     /// `use_nondet_result` will only be true if this is the first time we are
     /// generating a replace function.
     fn expand_replace_body(&self, before: &[Stmt], after: &[Stmt]) -> TokenStream {
+        let redefs = self.arg_redefinitions(false);
```
celinval, line 73: Can we do this redefinition just once when we first create the replace body instead of doing it for every contract clause?

library/kani_macros/src/sysroot/contracts/replace.rs
```diff
@@ -70,6 +70,7 @@ impl<'a> ContractConditionsHandler<'a> {
     /// `use_nondet_result` will only be true if this is the first time we are
     /// generating a replace function.
     fn expand_replace_body(&self, before: &[Stmt], after: &[Stmt]) -> TokenStream {
+        let redefs = &self.redefs;
```
celinval, line 73: Instead of doing it here, you can create the redefinition inside the initialize function here: https://github.com/model-checking/kani/blob/579151659252603b894e0af8ba55878cd4cd1947/library/kani_macros/src/sysroot/contracts/replace.rs#L22

PR #3729, opened 2024-11-19
List Subcommand Improvements

### LINKED ISSUE #3632: List subcommand accepts redundant arguments
I tried using the following command line invocation:

```
kani list -h
```

with Kani version: 0.56.0

I expected to see this happen: Just a few options to show up

Instead, this happened: Options such as `--unwind`, `--cbmc-args`, `--concrete-playback`, and so on show up as possible arguments.

### LINKED ISSUE #3633: List command using `--format=json` should print the location of the generated json file.
Requested feature: Print the path to the generated `json` file after running `kani list --format=json` unless `--quiet` is provided.
Use case: Improve the user experience when requesting a json file.

Currently, the command will print the regular cargo build logs, and quietly exit after that. I was expecting Kani to either print the json to the standard output or print the location of the file.

carolynzech:
Improvements to the list subcommand.

- Print the location of the JSON file for `--format=json` (#3633)
- Use `CommonArgs` instead of `VerificationArgs` (#3632)
- Add a Markdown option. The pretty table is nice for smaller packages, but as we progress with the standard library verification effort, it's too massive to be easily readable. In this case, it's better to print a markdown table and then just read it in a markdown editor. (Once this PR is merged, the [kani list workflow step](https://github.com/model-checking/verify-rust-std/pull/146) will use the markdown option).

Resolves #3632
Resolves #3633


zhassan-aws (commented):
Looks good. Can we add a test (or update an existing one) to check that the json file is printed in the output?

Similarly for the markdown table?

kani-driver/src/list/output.rs
```diff
@@ -176,5 +118,42 @@ pub fn json(
 
     serde_json::to_writer_pretty(writer, &json_obj)?;
 
+    println!("Wrote list results to {}", std::fs::canonicalize(JSON_FILENAME)?.display());
+
     Ok(())
 }
+
+/// Output the results of the list subcommand.
+pub fn output_list_results(
```
celinval, line 127: Nit: Can you please move this function to the top? This is the entry point of this module, so it's easier to read the code from it.

celinval (approved):
Please add tests as @zhassan-aws mentioned.

celinval (approved):
Thanks

PR #3734, opened 2024-11-25
Fix size and alignment computation for intrinsics

### LINKED ISSUE #3616: Kani miss overflow when computing `size_of_val`

celinval:
Implement overflow checks for `size_of_val`, and panic for foreign types.

Resolves #3616 


tests/expected/intrinsics/size_of_dst.rs
```diff
@@ -0,0 +1,53 @@
...
+use std::ptr;
+
+#[derive(Clone, Copy, kani::Arbitrary)]
+struct Wrapper<T: ?Sized> {
+    _size: usize,
+    _value: T,
+}
+
+#[kani::proof]
+pub fn check_size_of_adt_overflows() {
```
carolynzech, line 19: Should we have tests for `align_of_val_raw` too?

carolynzech (approved):

tests/expected/intrinsics/size_of_dst.rs
celinval: `align_of_val_raw` cannot overflow. It is actually safe today, but there's a question whether it will stay safe depending on how (see [this Zulip thread for more details](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/.E2.9C.94.20Safety.20of.20.60size_of_val_raw.60.20and.20.60align_of_val_raw.60)). But I can add a few sanity checks too.

We could force alignment to be non-multiple of 2 by generating invalid vtable data for now. That itself may eventually become immediate UB, but it doesn't seem to be decided yet: https://github.com/rust-lang/unsafe-code-guidelines/issues/516.

carolynzech: Huh, okay. Maybe just add a comment about how `align_of_val_raw` cannot overflow today, so we don't have tests? (Nothing too elaborate, just something so that a future reader realizes that this wasn't just an oversight on our part, but perhaps something that would need to change depending on whether it becomes unsafe in the future).
If the sanity checks aren't too much trouble, go ahead and add them, but I'd be fine with just a comment explaining the lack of tests.

celinval: Actually, that is indeed UB. So we can't really test the invalid alignment case.

Issue #2546, opened 2023-06-20
Quantifiers


JustusAdam:

d0nutptr:
Hello! I was curious if there was a sense of when this functionality might be available?

I was trying to create a function contract to express the hashing property:

```
kani::ensures(result != hash(kani::any_where(|x| *x != input)))
```

but I believe I need quantifiers for this to actually work. Thank you!

Issue #836, opened 2022-02-17
Support for quantifiers in asserts and assumes


nchong-at-aws:
Requested feature: Allow quantifiers in Kani asserts and assumes
Use case: CBMC supports bounded quantification under constant lower and upper bounds [0]. It would be great to have this in Kani too.

[0] https://github.com/diffblue/cbmc/blob/0a69a64e4481473d62496f9975730d24f194884a/doc/cprover-manual/contracts-quantifiers.md

Link to relevant documentation (Rust reference, Nomicon, RFC): N/A
Is this a breaking change? No

Possible syntax, inspired by [Prusti](https://viperproject.github.io/prusti-dev/user-guide/syntax.html):
```rust
fn main() {
    let xs: [i32; 4] = kani::any();
    kani::assume(
        kani::forall(|i: usize| (0 <= i && i < 4) ==> (0 <= xs[i] && xs[i] < 1024)));
}
```

d0nutptr:
Hello! I was curious if there was a sense of when this functionality might be available?

I was trying to create a function contract to express the hashing property:

```
kani::ensures(result != hash(kani::any_where(|x| *x != input)))
```

but I believe I need quantifiers for this to actually work. Thank you!

tautschnig:
Was resolved with #3993.

PR #3737, opened 2024-11-25
Add support for quantifiers

feliperodri:
We still need to address a few issues:

- [x] Avoid code duplication;
- [ ] Ensure CBMC doesn't ignore the quantifier expressions due to side effects (we should probably use lambda expressions);
- [ ] Only enable quantifiers with unstable flag;
- [ ] Add more regression tests to test UI;
- [ ] Extend [RFC 0010-quantifiers](https://github.com/model-checking/kani/blob/main/rfc/src/rfcs/0010-quantifiers.md) to include explanation on how Kani is building the quantifier expressions;

Resolves https://github.com/model-checking/kani/issues/2546 and https://github.com/model-checking/kani/issues/836.


gmorenz:
I've rebased this branch on master, with a few necessary updates, here: https://github.com/gmorenz/kani/tree/quantifiers

PR #3742, opened 2024-11-27
Add a Kani function that checks if the range of a float is valid for conversion to int

### LINKED ISSUE #3711: Add support for in_range
Requested feature: in_range to check if a float is in the range of the targeting integer type, e.g.
```rust
kani::in_range(IntType, floatType, float) or something similar
Example: f32::to_int_unchecked contract
#[requires(self.is_finite() && kani::in_range(Int, Self, self))]
pub unsafe fn to_int_unchecked<Int>(self) -> Int where Self: FloatToInt<Int> {...}
```

[Full Discussion](https://github.com/model-checking/verify-rust-std/pull/134):
> > Thank you for sharing this. Is it possible to have Kani support this? E.g. expose an `in_range(float, floatType, IntType)` API that we can directly call. We saw that the Kani internally has [codegen_in_range_expr](https://github.com/model-checking/kani/blob/main/kani-compiler/src/codegen_cprover_gotoc/utils/float_utils.rs#L42) which we think useful.
> > Thank you! @zhassan-aws
> 
> Yes, I think it would be possible. This would likely need to be done through providing a trait and its implementation for different float types so that it can be used with the generic `Int`. Can you file a feature request in the Kani repo?

zhassan-aws:
This introduces a new generic Kani function, `kani::float::float_to_int_in_range`, that can be used for checking if a floating-point value satisfies the range requirement of the `to_int_unchecked` methods, e.g https://doc.rust-lang.org/std/primitive.f32.html#method.to_int_unchecked and https://doc.rust-lang.org/std/primitive.f64.html#method.to_int_unchecked.

Resolves #3711 


library/kani_core/src/float.rs
```diff
@@ -0,0 +1,33 @@
...
+        /// assert!(!fits_in_i8);
+        ///
+        /// let f: f64 = 1e6;
+        /// let fits_in_u32 = kani::float::float_to_int_in_range::<f64, u32>(f);
+        /// // fits in `u32` because the value after truncation (`1e6`) is smaller than `u32::MAX`
+        /// assert!(fits_in_u32);
+        /// ```
+        #[kanitool::fn_marker = "FloatToIntInRangeHook"]
+        #[inline(never)]
+        pub fn float_to_int_in_range<Float, Int>(value: Float) -> bool {
```
celinval, line 29: Please mark this function as unstable.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -315,6 +316,54 @@ impl GotocHook for IsAllocated {
...
+
+        let is_in_range = utils::codegen_in_range_expr(
+            &float,
+            float_ty,
+            integral_ty,
+            gcx.symbol_table.machine_model(),
+        )
+        .cast_to(Type::CInteger(CIntType::Bool));
+
+        let pe = unwrap_or_return_codegen_unimplemented_stmt!(
```
celinval, line 357: we really need to get rid of this unwrap_or_return_codegen... stuff. oh well

library/kani_core/src/float.rs
celinval, line 29: If we add a bound here:
```
where
    Float: FloatToInt<Int>,
```
Then we don't need to check if the concrete types are valid floats / integers, right?

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -315,6 +316,54 @@ impl GotocHook for IsAllocated {
...
+        target: Option<BasicBlockIdx>,
+        span: Span,
+    ) -> Stmt {
+        assert_eq!(fargs.len(), 1);
+        let float = fargs.remove(0);
+        let target = target.unwrap();
+        let loc = gcx.codegen_span_stable(span);
+
+        let generic_args = instance.args().0;
+        let RigidTy::Float(float_ty) = generic_args[0].expect_ty().kind().rigid().unwrap().clone()
```
celinval, line 343:
```suggestion
        let RigidTy::Float(float_ty) = generic_args[0].expect_ty().kind().rigid().unwrap()
```

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -315,6 +316,54 @@ impl GotocHook for IsAllocated {
...
+        else {
+            unreachable!()
+        };
+        let integral_ty = generic_args[1].expect_ty().kind().rigid().unwrap().clone();
```
celinval, line 347:
```suggestion
        let integral_ty = generic_args[1].expect_ty().kind().rigid().unwrap();
```

kani-compiler/src/kani_middle/transform/kani_intrinsics.rs
```diff
@@ -606,6 +618,24 @@ impl IntrinsicGeneratorPass {
             Place::from(RETURN_LOCAL),
         );
     }
+
+    fn check_float_to_int_in_range_valid_types(tcx: TyCtxt, instance: Instance) {
```
celinval, line 622: See my comment on float.rs, I don't think we need this.

celinval (commented):
Thanks @zhassan-aws

library/kani_core/src/float.rs
zhassan-aws: Good idea. I'll do that

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
zhassan-aws: This is what I had initially, but this results in an error:
```
error[E0716]: temporary value dropped while borrowed
   --> kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs:343:40
    |
343 |         let RigidTy::Float(float_ty) = generic_args[0].expect_ty().kind().rigid().unwrap()
    |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use
...
346 |         };
    |          - temporary value is freed at the end of this statement
...
351 |             *float_ty,
    |             --------- borrow later used here
```
It works if I add a deref:
```suggestion
        let RigidTy::Float(float_ty) = *generic_args[0].expect_ty().kind().rigid().unwrap()
```
but I think this is equivalent to the `clone()` version.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
zhassan-aws: Same here.

library/kani_core/src/float.rs
zhassan-aws: None of the current unstable features is suitable for this. I'll need to add a new one.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
celinval: Yes, you are correct that you need to dereference it.

library/kani_core/src/float.rs
zhassan-aws: Done.

zhassan-aws: Done.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
zhassan-aws: Done.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
zhassan-aws: I kept this one the same.

kani-compiler/src/kani_middle/transform/kani_intrinsics.rs
zhassan-aws: Removed.

library/kani_core/src/float.rs
```diff
@@ -3,10 +3,12 @@
 
 //! This module contains functions useful for float-related checks
 
+#[allow(clippy::crate_in_macro_def)]
 #[macro_export]
 macro_rules! generate_float {
-    ($core:path) => {
+    ($core:tt) => {
```
celinval, line 9: I don't think we really need this. Using `core::` works in both contexts.

celinval (approved):
Thanks!

PR #3755, opened 2024-12-05
Add out of bounds check for `offset` intrinsics

### LINKED ISSUE #1233: `offset` intrinsic doesn't check that pointers stay within bounds
The [documentation for the `offset` method](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset) mentions in the safety conditions that:

> Both the starting and resulting pointer must be either in bounds or one byte past the end of the same [allocated object](https://doc.rust-lang.org/std/ptr/index.html#allocated-object).

At present, the standard library prevents us from adding this check because otherwise we run into spurious failures (i.e., offsets that go out-of-bounds,  are frequently used).

celinval:
Note that the `offset` intrinsics is lowered to `Rvalue::Offset`. Thus, we have to replace the entire statement.

This PR does not address `ptr_offset_from` and `ptr_offset_from_unsigned`. I'll push those changes to a separate PR.

Resolves #1233 


celinval:
I realized that I ended up not using the unsupported hook. I'm happy to remove it from this PR if needed.

library/kani_core/src/models.rs
```diff
@@ -42,6 +44,71 @@ macro_rules! generate_models {
...
+            /// An offset model that checks UB.
+            #[kanitool::fn_marker = "OffsetModel"]
+            pub fn offset<T, P: Ptr<T>, O: ToISize>(ptr: P, offset: O) -> P {
+                let offset = offset.to_isize();
+                let t_size = core::mem::size_of::<T>() as isize;
+                if offset == 0 || t_size == 0 {
+                    // It's always safe to perform an offset of length 0.
+                    ptr
+                } else {
+                    let (byte_offset, overflow) = offset.overflowing_mul(t_size as isize);
```
carolynzech, line 57:
```suggestion
                    let (byte_offset, overflow) = offset.overflowing_mul(t_size);
```
You already cast to an `isize` when creating `t_size`, so I don't think this cast should be necessary.

library/kani_core/src/models.rs
```diff
@@ -42,6 +44,71 @@ macro_rules! generate_models {
...
+                    kani::safety_check(!overflow, "Offset in bytes overflow isize");
```
carolynzech, line 58:
```suggestion
                    kani::safety_check(!overflow, "Offset in bytes overflows isize");
```

library/kani_core/src/mem.rs
```diff
@@ -149,7 +149,14 @@ macro_rules! kani_mem {
         )]
         #[allow(clippy::not_unsafe_ptr_arg_deref)]
         pub fn same_allocation<T: ?Sized>(ptr1: *const T, ptr2: *const T) -> bool {
-            cbmc::same_allocation(ptr1, ptr2)
+            same_allocation_internal(ptr1, ptr2)
+        }
+
+        #[allow(clippy::not_unsafe_ptr_arg_deref)]
+        pub(super) fn same_allocation_internal<T: ?Sized>(ptr1: *const T, ptr2: *const T) -> bool {
```
carolynzech, line 156: Nit: I think it would be cleaner to delete this function and just do this:

```rust
#[allow(clippy::not_unsafe_ptr_arg_deref)]
pub fn same_allocation<T: ?Sized>(ptr1: *const T, ptr2: *const T) -> bool {
  let addr1 = ptr1 as *const ();
  let addr2 = ptr2 as *const ();
  cbmc::same_allocation(addr1, addr2)
}
```
and call `same_allocation` inside of our offset model. This isn't a huge deal--it just seems simpler to keep one function, unless I'm missing something here.

library/kani_core/src/models.rs
```diff
@@ -42,6 +44,71 @@ macro_rules! generate_models {
...
+                    let orig_ptr = ptr.to_const_ptr();
+                    let new_ptr = orig_ptr.wrapping_byte_offset(byte_offset);
+                    kani::safety_check(
+                        kani::mem::same_allocation_internal(orig_ptr, new_ptr),
+                        "Offset result and original pointer should point to the same allocation",
```
carolynzech, line 63:
```suggestion
                        "Offset result and original pointer must point to the same allocation",
```

tests/expected/offset-bounds-check/out_of_bounds_ub_check.expected
```diff
@@ -0,0 +1,4 @@
+Failed Checks: Offset result and original pointer should point to the same allocation
```
carolynzech, line 1:
```suggestion
Failed Checks: Offset result and original pointer must point to the same allocation
```

tests/expected/offset-bounds-check/start_from_oob.expected
```diff
@@ -0,0 +1,4 @@
+Failed Checks: Offset result and original pointer should point to the same allocation
```
carolynzech, line 1:
```suggestion
Failed Checks: Offset result and original pointer must point to the same allocation
```

tests/expected/offset-bytes-overflow/expected
```diff
@@ -1,2 +1,4 @@
-FAILURE\
-offset: attempt to compute number in bytes which would overflow
\ No newline at end of file
+Failed Checks: Offset in bytes overflow isize
```
carolynzech, line 1:
```suggestion
Failed Checks: Offset in bytes overflows isize
```

tests/expected/offset-from-bytes-overflow/expected
```diff
@@ -1,2 +1,4 @@
-FAILURE\
-attempt to compute number in bytes which would overflow
+Failed Checks: Offset in bytes overflow isize
```
carolynzech, line 1:
```suggestion
Failed Checks: Offset in bytes overflows isize
```

tests/expected/offset-invalid-args/invalid_args.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//! Check Kani output if the arguments provided to the offset intrisic are incorrect.
```
carolynzech, line 3:
```suggestion
//! Check Kani output if the arguments provided to the offset intrinsic are incorrect.
```

tests/expected/offset-overflow/expected
```diff
@@ -1,2 +1,4 @@
-FAILURE\
-attempt to compute offset which would overflow
+Failed Checks: Offset result and original pointer should point to the same allocation
```
carolynzech, line 1:
```suggestion
Failed Checks: Offset result and original pointer must point to the same allocation
```

tests/expected/offset-invalid-args/invalid_args.expected
```diff
@@ -0,0 +1 @@
+Cannot offset non-pointer type
```
carolynzech, line 1: Where does this error message come from? I would have expected to see the message from `validate_raw_ptr`.
Also, can you test that `validate_offset` catches failures too?

tests/expected/offset-wraps-around/main.rs
```diff
@@ -1,35 +1,59 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-// Check that a high offset causes a "wrapping around" behavior in CBMC.
+//! This test checks that Kani correctly computes the value of a wrapping_offset in cases where
+//! the add operation wraps.
+//!
+//! Note that CBMC offset logic will wrap around the object bits, not the entire space address when
```
carolynzech, line 7:
```suggestion
//! Note that CBMC offset logic will wrap around the object bits, not the entire address space, when
```

tests/expected/offset-wraps-around/main.rs
```diff
@@ -1,35 +1,59 @@
...
-        // default behavior.
-        let v_wrap: *const u128 = v_0.add(high_offset.try_into().unwrap());
-        let wrapped_offset = v_wrap.offset_from(v_0);
-        // Both offsets should be the same, but because of the "wrapping around"
-        // behavior in CBMC, `wrapped_offset` does not equal `high_offset`
+        let v_wrap: *const u128 = v_0.wrapping_add(high_offset);
+        // This should trigger UB!!
+        let wrapped_offset = unsafe { v_wrap.offset_from(v_0) };
+        // Without UB detection, the offsets are the same, but CBMC pointer arithmetic
+        // would "wrapping around" making this incorrect
```
carolynzech, line 23:
```suggestion
        // would "wrap around" making this incorrect
```

tests/expected/offset-wraps-around/main.rs
```diff
@@ -1,35 +1,59 @@
...
         // https://github.com/model-checking/kani/issues/1150
         assert!(high_offset == wrapped_offset.try_into().unwrap());
     }
 }
+
+/// This harness is similar to the `original_harness`, but we replace the `offset_from` with
+/// a subtraction on the pointer addresses.
+#[kani::proof]
```
carolynzech, line 31: Can you update the expected file to account for these new harnesses? We should check that verification succeeds.

carolynzech (commented):
Nice! I don't have as many comments as it initially appears, a good chunk of them are just propagating error message suggestions through the tests.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -205,10 +207,15 @@ impl GotocHook for Check {
 
         let (msg, reach_stmt) = gcx.codegen_reachability_check(msg, span);
 
+        // Since `cond` might have side effects, assign it to a temporary
+        // variable so that it's evaluated once, then assert and assume it
+        // TODO: I don't think `cond` can have side effects, this is MIR, it's going to be temps
+        let (tmp, decl) = gcx.decl_temp_variable(cond.typ().clone(), Some(cond), caller_loc);
```
zhassan-aws, line 213: Why this change?

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
```diff
@@ -38,7 +38,7 @@ pub enum PropertyClass {
     /// Overflow panics that can be generated by Intrisics.
     /// NOTE: Not all uses of this are found by rust-analyzer because of the use of macros. Try grep instead.
     ///
-    /// SPECIAL BEHAVIOR: None TODO: Why should this exist?
+    /// SPECIAL BEHAVIOR: Same as SafetyCheck.
```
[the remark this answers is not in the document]
celinval: Yeah, I think we should get rid of them. Just not in this PR. :smile:

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -747,3 +755,25 @@ impl GotocHooks {
         }
     }
 }
+
+fn check_hook(
```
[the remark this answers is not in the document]
celinval: yeah, I agree. I had initially moved two other hook checks to us it, but the are slightly different. I can remove it.

library/kani_core/src/mem.rs
celinval: The reason why I created a second one is because the first one is tagged with `#[unstable]` attribute. Using that one would require us to either remove the attribute or users would have to pass `-Z mem-predicates`.

tests/expected/offset-bounds-check/start_from_oob.expected
celinval: If the first is out of bounds, then the pointers do not point to the same allocation. Let me know if you have a different message in mind before I update all the tests.

zhassan-aws (commented):
Thanks!

tests/expected/offset-invalid-args/invalid_args.expected
celinval: So, the Rust compiler actually crashes before codegen. This is a panic message from the compiler.

I added the validation and tests just in case they eventually decide to handle this case differently, but today, the validations are redundant.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
celinval: That was a mistake. Let me revert it.

tests/kani/ValidValues/unaligned.rs
celinval, line 1: btw, I had this test in my repo for a while, and I accidentally bundled up in this PR. I hope that's OK. :smile:

kani-compiler/src/codegen_cprover_gotoc/codegen/assert.rs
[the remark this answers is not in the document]
celinval: I adjusted the message

library/kani_core/src/mem.rs
carolynzech: Ah okay, works for me then.

tests/expected/offset-wraps-around/main.rs
celinval: Done!

carolynzech (approved):

PR #3757, opened 2024-12-05
Add UB checks for ptr_offset_from* intrinsics

### LINKED ISSUE #3756: Kani does not detect UB for `ptr_offset_from` and `ptr_offset_from_unsigned`
```rust
#[kani::proof]
fn check_offset_from() {
    let val = 10u128;
    let ptr: *const u128 = &val;
    let ptr_oob: *const u128 = ptr.wrapping_add(10);
    // SAFETY: This is not safe!
    let _offset = unsafe { ptr_oob.offset_from(ptr) };
}
```

```
kani check_ub.rs
```

with Kani version: 0.56.0

I expected to see this happen: Verification fails due to UB detection

celinval:
Add a new model for `ptr_offset_from` and `ptr_offset_from_unsigned`
intrinsics that check allocation and address order.

Resolves #3756


carolynzech:
I did some digging into the failing perf tests. The problem seems to be that these models don't properly handle iteration over vectors of vectors, e.g. `Vec<Vec<T>>`.
Minimal reproducer (based on `perf/vec/vec/src/main.rs`:
```rust
#[kani::proof]
#[kani::unwind(5)]
#[kani::solver(minisat)]
fn main() {
    let v1: Vec<Vec<i32>> = vec![vec![1], vec![2]];
    v1.into_iter();
}
```
This code fails with:
```
Check 87: kani::rustc_intrinsics::ptr_offset_from::<std::vec::Vec<i32>>.safety_check.3
- Status: FAILURE
- Description: "Expected the distance between the pointers, in bytes, to be a multiple of the size of `T`"
- Location: ../../../../library/kani/src/lib.rs:54:1 in function kani::rustc_intrinsics::ptr_offset_from::<std::vec::Vec<i32>>
```

celinval:
> I did some digging into the failing perf tests. The problem seems to be that these models don't properly handle iteration over vectors of vectors, e.g. `Vec<Vec<T>>`. Minimal reproducer (based on `perf/vec/vec/src/main.rs`:
> 
> ```rust
> #[kani::proof]
> #[kani::unwind(5)]
> #[kani::solver(minisat)]
> fn main() {
>     let v1: Vec<Vec<i32>> = vec![vec![1], vec![2]];
>     v1.into_iter();
> }
> ```
> 
> This code fails with:
> 
> ```
> Check 87: kani::rustc_intrinsics::ptr_offset_from::<std::vec::Vec<i32>>.safety_check.3
> - Status: FAILURE
> - Description: "Expected the distance between the pointers, in bytes, to be a multiple of the size of `T`"
> - Location: ../../../../library/kani/src/lib.rs:54:1 in function kani::rustc_intrinsics::ptr_offset_from::<std::vec::Vec<i32>>
> ```

I'm investigating this issue. It seems that the failure occurs inside `size_hint` of `vec::IntoIter` which is called by the drop implementation.

celinval:
My bad... I confused size and alignment. Only the second is a power of two. :blush:

library/kani_core/src/models.rs
```diff
@@ -45,6 +45,57 @@ macro_rules! generate_models {
                 }
             }
 
+            /// Implements core::intrinsics::ptr_offfset_from with safety checks in place.
```
carolynzech, line 48:
```suggestion
            /// Implements core::intrinsics::ptr_offset_from with safety checks in place.
```

tests/expected/offset-bounds-check/offset_from.expected
```diff
@@ -0,0 +1,6 @@
+Failed Checks: assertion failed: 0 < pointee_size && pointee_size <= isize::MAX as usize
```
carolynzech, line 1: Nit: it would be nice to format this file like [tests/expected/offset-bounds-check/sub_ptr.expected](https://github.com/model-checking/kani/pull/3757/files#diff-4d7c9aa431be214c52ccf1342d03c33506899ecafb529b0849d40430ce58c41c), so that we can see the checks that should fail for each harness.

tests/expected/offset-from-zst/main.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright Kani Contributors
```
carolynzech, line 1: The ZST test `check_offset_from_unit_panic` in tests/expected/offset-bounds-check/offset_from.rs makes this test redundant.

carolynzech (approved):

tests/expected/offset-from-zst/main.rs
celinval: good catch... thanks

PR #3802, opened 2024-12-30
Add an option to generate contracts as assertions

### LINKED ISSUE #3326: Add an option to turn `requires` clauses into assertions

carolynzech:
Add an unstable option `-Zcontracts-as-assertions` that generates contract clauses as assertions. This works by adding a new contract mode that outputs the original body of the function with its contract clauses asserted. During the `FunctionWithContractPass`, if this option was passed, we use this closure instead of the original body of the function.

I'm omitting a description of why this is useful from the PR description--there's one in the module documentation, and I want that to be a self-contained explanation (i.e., if something is unclear, I want to fix it there).

Resolves #3326

### DIFF
--- a/kani_metadata/src/unstable.rs
+++ b/kani_metadata/src/unstable.rs
@@ -85,6 +85,8 @@ pub enum UnstableFeature {
     FunctionContracts,
     /// Enable loop contracts [RFC 12](https://model-checking.github.io/kani/rfc/rfcs/0012-loop-contracts.html)
     LoopContracts,
+    /// Enable function contracts to be interpreted as assertions. (Requires function-contracts option).
+    ContractsAsAssertions,
     /// Memory predicate APIs.
     MemPredicates,
     /// Automatically check that no invalid value is produced which is considered UB in Rust.
--- a/kani_metadata/src/unstable.rs
+++ b/kani_metadata/src/unstable.rs
@@ -85,7 +85,7 @@ pub enum UnstableFeature {
     FunctionContracts,
     /// Enable loop contracts [RFC 12](https://model-checking.github.io/kani/rfc/rfcs/0012-loop-contracts.html)
     LoopContracts,
-    /// Enable function contracts to be interpreted as assertions. (Requires function-contracts option).
+    /// Interpret function contracts as assertions. (Requires function-contracts option).
     ContractsAsAssertions,
     /// Memory predicate APIs.
     MemPredicates,
--- a/library/kani_macros/src/sysroot/contracts/assert.rs
+++ b/library/kani_macros/src/sysroot/contracts/assert.rs
@@ -1,0 +1,90 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Logic used for generating the code that generates contract preconditions and postconditions as assertions.
+
+use proc_macro2::{Ident, Span, TokenStream as TokenStream2};
+use quote::quote;
+use std::mem;
+use syn::{Stmt, parse_quote};
+
+use super::{
+    ContractConditionsData, ContractConditionsHandler, ContractMode, INTERNAL_RESULT_IDENT,
+    helpers::*,
+    shared::{build_ensures, split_for_remembers},
+};
+
+impl<'a> ContractConditionsHandler<'a> {
+    /// Generate a token stream that represents the assert closure.
+    ///
+    /// See [`Self::make_assert_body`] for the most interesting parts of this
+    /// function.
+    pub fn assert_closure(&self) -> TokenStream2 {
+        let assert_ident = Ident::new(&self.assert_name, Span::call_site());
+        let sig = &self.annotated_fn.sig;
+        let output = &sig.output;
+        let body_stmts = self.initial_assert_stmts();
+        let body = self.make_assert_body(body_stmts);
+
+        quote!(
+            #[kanitool::is_contract_generated(assert)]
+            #[allow(dead_code, unused_variables, unused_mut)]
+            let mut #assert_ident = || #output #body;
+        )
+    }
+
+    /// Expand the assert closure body.
+    pub fn expand_assert(&self, closure: &mut Stmt) {
+        let body = closure_body(closure);
+        *body = syn::parse2(self.make_assert_body(mem::take(&mut body.block.stmts))).unwrap();
+    }
+
+    /// Initialize the list of statements for the assert closure body.
+    fn initial_assert_stmts(&self) -> Vec<syn::Stmt> {
+        let return_type = return_type_to_type(&self.annotated_fn.sig.output);
+        let stmts = &self.annotated_fn.block.stmts;
+        let result = Ident::new(INTERNAL_RESULT_IDENT, Span::call_site());
+        parse_quote! {
+            let #result : #return_type = {#(#stmts)*};
+            #result
+        }
+    }
+
+    /// Create the body of an assert closure.
+    ///
+    /// Wraps the conditions from this attribute around `self.body`.
+    fn make_assert_body(&self, mut body_stmts: Vec<Stmt>) -> TokenStream2 {
+        let Self { attr_copy, .. } = self;
+        match &self.condition_type {
+            ContractConditionsData::Requires { attr } => {
+                quote!({
+                    kani::assert(#attr, stringify!(#attr_copy));
+                    #(#body_stmts)*
+                })
+            }
+            ContractConditionsData::Ensures { attr } => {
+                let (remembers, ensures_clause) = build_ensures(attr);
+
+                let exec_postconditions = quote!(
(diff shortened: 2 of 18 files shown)


library/kani_macros/src/sysroot/contracts/mod.rs
```diff
@@ -226,6 +248,23 @@
...
 //!         }
+//!         kani::internal::ASSERT => {
+//!             #[kanitool::is_contract_generated(assert)]
+//!             #[allow(dead_code, unused_variables, unused_mut)]
+//!             let mut __kani_assert_div =
+//!                 || -> u32
+//!                     {
+//!                         kani::assert(divisor != 0, "divisor != 0");
+//!                         let mut body_wrapper = || -> u32 { dividend / divisor };
+//!                         let result_kani_internal: u32 = body_wrapper();
```
carolynzech, line 259: Should I be calling `arg_redefinitions` when generating the `body_wrapper` closure body? I didn't think so, but I wasn't 100% sure.

tests/expected/function-contract/as-assertions/precedence.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright Kani Contributors
```
zhassan-aws, line 1: Nice test!

tests/expected/function-contract/as-assertions/precedence.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Test that proof_for_contract takes precedence over the assert mode, i.e.
+// that the target of the proof for contract still has its preconditions assumed.
+#[kani::proof_for_contract(add_one)]
+fn proof_for_contract_takes_precedence() {
+    let mut i = kani::any();
+    // if add_one's precondition was asserted, verification would fail,
+    // but since it's assumed, we get a vacuously successful proof instead.
+    kani::assume(i == 2);
```
zhassan-aws, line 36: This assumption should be removed since it's contradictory with `add_one`'s precondition.

library/kani_macros/src/sysroot/contracts/assert.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    /// Expand the assert closure body.
+    pub fn expand_assert(&self, closure: &mut Stmt) {
+        let body = closure_body(closure);
+        *body = syn::parse2(self.make_assert_body(mem::take(&mut body.block.stmts))).unwrap();
+    }
+
+    /// Initialize the list of statements for the assert closure body.
+    /// Construct a closure that wraps the body of the function, then invoke it and return the result.
+    fn initial_assert_stmts(&self) -> Vec<syn::Stmt> {
```
zhassan-aws, line 44:
```suggestion
    fn initial_assert_stmts(&self) -> Vec<Stmt> {
```

zhassan-aws (commented):
Nice, thanks!

Only a few comments.

kani_metadata/src/unstable.rs
```diff
@@ -85,6 +85,8 @@ pub enum UnstableFeature {
     FunctionContracts,
     /// Enable loop contracts [RFC 12](https://model-checking.github.io/kani/rfc/rfcs/0012-loop-contracts.html)
     LoopContracts,
+    /// Interpret function contracts as assertions. (Requires function-contracts option).
+    ContractsAsAssertions,
```
celinval, line 89: Do we need a `-Z contracts-as-assertions`? I think this feature should be under the existing function contracts unstable flags. If you want to implement this behavior as an opt-in, you can add a new argument that requires the function contracts unstable flag.

carolynzech: So are you saying just delete this option and make it the default under `-Z function-contracts` option, with no ability to turn it off? That's fine with me, just making sure before I make changes.

tests/expected/function-contract/as-assertions/assert-postconditions.rs
```diff
@@ -0,0 +1,30 @@
...
+    add_two(add_three_ptr);
+}
+
+#[kani::ensures(|result| old(*add_two_ptr + 1) == *add_two_ptr)] // incorrect -- should be old(*add_two_ptr + 1)
+fn add_two(add_two_ptr: &mut u32) {
+    *add_two_ptr += 1;
+    add_one(add_two_ptr)
+}
+
+#[kani::ensures(|result| old(*add_one_ptr + 1) == *add_one_ptr)] // correct -- assertion should always succeed
```
carolynzech, line 20:
```suggestion
#[kani::ensures(|_| old(*add_one_ptr + 1) == *add_one_ptr)] // correct -- assertion should always succeed
```

tests/expected/function-contract/as-assertions/assert-preconditions.rs
```diff
@@ -0,0 +1,55 @@
...
+#[kani::requires(*ptr < 100)]
+#[kani::ensures(|result| old(*ptr + 3) == *ptr)]
+#[kani::modifies(ptr)]
+fn add_three(ptr: &mut u32) {
+    *ptr += 1;
+    add_two(ptr);
+}
+
+#[kani::requires(*ptr < 101)]
+#[kani::ensures(|result| old(*ptr + 2) == *ptr)]
```
carolynzech, line 16:
```suggestion
#[kani::ensures(|_| old(*ptr + 2) == *ptr)]
```

tests/expected/function-contract/as-assertions/precedence.rs
carolynzech: That was the intention--the idea is that if we were asserting the precondition, verification would fail, since we would assert add_one's precondition and find the contradiction. However, since we have this assumption, we get a vacuously successful proof instead.
That being said, if the test is confusing, we shouldn't have it. I had another precedence-like test in assert-preconditions.rs, so I moved it here and wrote a longer comment explaining the logic. LMK if that's clearer.

zhassan-aws: Wouldn't verification still pass (non-vacuously) without this assumption?

carolynzech: Yes, but I just figured we may as well stress the point that it gets assumed by having a vacuous test case. It's not a necessary condition, I just thought it made the point clearer (which I guess it didn't :))

kani_metadata/src/unstable.rs
celinval: I'm saying that if you want an ability to turn it on and off, it should be a regular argument that requires `-Z function-contracts`. I tend to think that this should be the default if `-Z function-contracts` is enabled. We could add something like `--assert-contracts` or `-no-contract-assert`. @zhassan-aws @feliperodri, any thoughts?

zhassan-aws: I agree. Having it on by default (unless we're aware of cases it won't work in) with an option to turn it off makes sense.

tests/expected/function-contract/as-assertions/precedence.rs
zhassan-aws: My thinking was that if the pre-condition isn't assumed, then the harness will fail because of an overflow in the post-condition. So, just checking that the harness succeeds is sufficient to ensure that `proof_for_contract` is taking precedence.

kani-compiler/src/args.rs
```diff
@@ -36,6 +36,9 @@ pub enum ReachabilityType {
 /// with. Usually stored in and accessible via [`crate::kani_queries::QueryDb`].
 #[derive(Debug, Default, Clone, clap::Parser)]
 pub struct Arguments {
+    /// Option used to enable asserting function contracts.
```
carolynzech, line 39:
```suggestion
    /// Option used to disable asserting function contracts.
```

kani-driver/src/args/mod.rs
```diff
@@ -309,6 +309,10 @@ pub struct VerificationArgs {
     )]
     pub synthesize_loop_contracts: bool,
 
+    /// Do not assert the function contracts of dependencies. Requires -Z function-contracts.
+    #[arg(long)]
```
carolynzech, line 313:
```suggestion
    #[arg(long, hide_short_help = true)]
```

kani_metadata/src/unstable.rs
carolynzech: Done (the option to disable is `--no-assert-contracts`).

zhassan-aws (approved):
Thanks!

library/kani_macros/src/sysroot/contracts/mod.rs
celinval: No, you don't. The `arg_redefinitions` is related to the frame condition checking.

PR #3805, opened 2025-01-01
fix: Allow multiple `stub_verified`s on a single harness

### LINKED ISSUE #3804: Cannot use multiple `#[stub_verified(..)]` attributes to a single harness
The docs says

https://github.com/model-checking/kani/blob/7913e3c05d25237cf73652c2ae0051d5918a5307/library/kani_macros/src/lib.rs#L368-L375

but actually, we can't because

https://github.com/model-checking/kani/blob/7913e3c05d25237cf73652c2ae0051d5918a5307/kani-compiler/src/kani_middle/attributes.rs#L373-L375

I'd like to fix this but I wonder which way it was originally intended for.

ShoyuVanilla:
It seems that we can actually use multple `stub_verified`s on a single harness internally, but the checking logic prevents such usages.
This PR removes such restrection.

Resolves #3804

### DIFF
--- a/kani-compiler/src/kani_middle/attributes.rs
+++ b/kani-compiler/src/kani_middle/attributes.rs
@@ -371,7 +371,7 @@ impl<'tcx> KaniAttributes<'tcx> {
                     attrs.iter().for_each(|attr| self.check_proof_attribute(kind, attr))
                 }
                 KaniAttributeKind::StubVerified => {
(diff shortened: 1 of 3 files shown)


ShoyuVanilla:
Oh, I wrote this PR in my mobile device without actual checks and it doesn't compiles 😅 I'll fix this few hours later

ShoyuVanilla:
Closing this as we have #3808

PR #3806, opened 2025-01-02
fix: clippy

ShoyuVanilla:

### DIFF
--- a/cprover_bindings/src/irep/goto_binary_serde.rs
+++ b/cprover_bindings/src/irep/goto_binary_serde.rs
@@ -936,9 +936,9 @@ where
     }
 }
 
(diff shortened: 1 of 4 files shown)


celinval:
Thank you for the contribution, @ShoyuVanilla. Quick question, how did you bump into these issues in the first place? I'm wondering if we need to update our CI to check these crates as well.

ShoyuVanilla:
I use rust-analyzer with config `rust-analyzer.check.command = "clippy"` and got clippy warnings inside my text editor(nvim). I have no idea why CI couldn't catch those warnings 🤔

celinval (approved):
Thanks!

PR #3808, opened 2025-01-02
Allow multiple  annotations, but check for duplicate targets.

### LINKED ISSUE #3804: Cannot use multiple `#[stub_verified(..)]` attributes to a single harness

remi-delmas-3000:
Resolves #3804.

Enables multiple `stub_verified(TARGET)` annotations on a harness, but still detect and report duplicate targets.

Adds positive and negative tests.


zhassan-aws:
Is this a duplicate of #3805?

ShoyuVanilla:
> Is this a duplicate of https://github.com/model-checking/kani/pull/3805?

I think that this one supersedes mine(#3805)😅

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -568,7 +568,21 @@ impl<'tcx> KaniAttributes<'tcx> {
...
+                dcx.struct_span_err(
+                    span,
+                    format!("Multiple occurrences of `stub_verified({})`.", name),
+                )
+                .with_span_note(
+                    self.tcx.def_span(def_id),
+                    format!("Use a single `stub_verified({})` annotation.", name),
+                )
+                .emit();
+            } else {
```
celinval, line 583: Can this be a warning instead? They will be redundant, but they can be safely ignored.

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -371,7 +371,7 @@ impl<'tcx> KaniAttributes<'tcx> {
                     attrs.iter().for_each(|attr| self.check_proof_attribute(kind, attr))
                 }
                 KaniAttributeKind::StubVerified => {
-                    expect_single(self.tcx, kind, &attrs);
+                    // Actual validation happens when the annotation are handled
```
celinval, line 374: Ideally, this should be similar to `parse_stubs` so we can emit as many errors as possible, and sooner rather than later. Can you please add a test to see if that is the case today. Something like:
```rust
#[kani::proof]
#[kani::stub_verified(dummy)]
fn stub_does_not_exist() { }

#[kani::proof]
#[kani::stub_verified(dummy, dummy2)]
fn stub_invalid_args() { }
```

That said, for this PR, I think a TODO comment should suffice, since you are not introducing this limitation. Thanks!

remi-delmas-3000: We already have tests for these cases in `missing_contracts_for_check.rs` and `missing_contracts_for_replace.rs`

kani-compiler/src/kani_middle/attributes.rs
remi-delmas-3000: Changed it to a warning, and moved the checks earlier in the process.

celinval (approved):
Thanks @remi-delmas-3000

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -587,10 +601,21 @@ impl<'tcx> KaniAttributes<'tcx> {
                     .emit();
                 return;
             }
+        }
+    }
+
+    /// Adds the verified stub names to the `harness.verfied_stubs`.
```
feliperodri, line 607:
```suggestion
    /// Adds the verified stub names to the `harness.verified_stubs`.
```

feliperodri (approved):

feliperodri:
@remi-delmas-3000 it seems CI is only failing due to formatting changes. Could you rebase it and merge it?

kani-compiler/src/kani_middle/attributes.rs
```diff
@@ -587,10 +601,21 @@ impl<'tcx> KaniAttributes<'tcx> {
...
+    /// This method must be called after `check_stub_verified`, to ensure that
+    /// the target names are known and have contracts, and there are no
+    /// duplicate target names.
+    fn handle_stub_verified(&self, harness: &mut HarnessAttributes) {
+        for (name, _, _) in self.interpret_stub_verified_attribute() {
             harness.verified_stubs.push(name.to_string())
         }
     }
 
+
```
celinval, line 618:
```suggestion
```

PR #3818, opened 2025-01-09
Fix hanging command in `std-analysis.sh`

carolynzech:
This command hangs on my machine (ARM64 Mac). Adding quotes around the `tr -d [:alpha:]_, ... ` part fixes the issue.

### DIFF
--- a/scripts/std-analysis.sh
+++ b/scripts/std-analysis.sh
@@ -104,10 +104,8 @@ for f in $results/*overall.csv; do
     fname=$(basename $f)
     crate=${fname%_scan_overall.csv}
     echo -n "$crate," >> $summary
-    temp_file=$(mktemp)
-    tr -d '[:alpha:]_,;' < "$f" > $temp_file
-    tr -s '\n' ',' < $temp_file >> "$summary"
-    rm $temp_file
+    tr -d '[:alpha:]_,;' < $f | tr -s '\n' ',' \
+        >> $summary
     echo "" >> $summary
 done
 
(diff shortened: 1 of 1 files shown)


scripts/std-analysis.sh
```diff
@@ -104,8 +104,10 @@ for f in $results/*overall.csv; do
     fname=$(basename $f)
     crate=${fname%_scan_overall.csv}
     echo -n "$crate," >> $summary
-    tr -d [:alpha:]_,; < $f | tr -s '\n' ',' \
-        >> $summary
+    temp_file=$(mktemp)
+    tr -d '[:alpha:]_,;' < "$f" > $temp_file
```
zhassan-aws, line 108: Is the temp file necessary? It should be possible to pipe the output without using an intermediate file.

carolynzech: It should be, but it hangs for me consistently. Could just be a me issue, but I thought it may be worth fixing if other people try to run this and run into the same problem.
If we think it's just something wonky with my machine, I'm fine to close and just keep this as a local fix.

carolynzech: (I ran into this problem trying to test my [metrics workflow](https://github.com/model-checking/verify-rust-std/pull/235) locally).

zhassan-aws: I reproduced the hang on my machine as well (Ubuntu). Looks like adding quotes is sufficient to resolve it. Can you check?

zhassan-aws:
```suggestion
    tr -d '[:alpha:]_,;' < $f | tr -s '\n' ',' \
        >> $summary
```

carolynzech: Yep, thanks Zyad!

zhassan-aws (approved):
Thanks!

PR #3837, opened 2025-01-16
Add a regression test for `no_std` feature

carolynzech:
Add a regression test to ensure that we can run Kani in crates with the `no_std` feature enabled.

### DIFF
--- a/tests/cargo-kani/no-std/src/lib.rs
+++ b/tests/cargo-kani/no-std/src/lib.rs
@@ -1,0 +1,25 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// Ensure that "extern crate kani" is sufficient to run Kani when no_std is enabled.
+
+#![no_std]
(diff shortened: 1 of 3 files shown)


tests/cargo-kani/no-std/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+    let x = kani::any_where(|n| *n < u8::MAX / 2);
+    let y = kani::any_where(|n| *n < u8::MAX / 2);
+    add(x, y);
+}
+
+use kani::cover;
+
+#[kani::proof]
+fn verify_point() {
+    cover!(true)
```
zhassan-aws, line 24: What is the purpose of this harness?

zhassan-aws (approved):

tests/cargo-kani/no-std/src/lib.rs
carolynzech: To check that we can import a Kani function using a `use` statement in this context

PR #3841, opened 2025-01-17
Enable GitHub Linux/Arm runners in CI

tautschnig:
GitHub made Linux Arm 64-bit runners available:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/. Run regression tests on those and build release packages.


tautschnig:
Waiting for GitHub to fix permissions - see https://github.com/orgs/community/discussions/148648#discussioncomment-11858717.

tautschnig:
> Thanks! I wonder if we should add x86 as a suffix to make the distinction clearer (e.g. `BuildBundle-Linux-x86`). I don't feel strongly, though.

Makes a lot of sense, now done!

Issue #3571, opened 2024-10-04
UB checks should fail verification for harnesses annotated with `#[should_panic]`


celinval:
I believe we need to create a new UB check property category, and audit the compiler to use the new category for UB instrumentation. The `#[should_panic]` implementation should take this new category into consideration. Note that this is already the case for some UB's checks.

The motivation is to allow users to write safety harness that succeed as long as no UB is detected.

Here is a small example of how Kani inconsistencyt:

```rust
#[kani::proof]
#[kani::should_panic]
pub fn rust_ub_fails() {
    let ptr = 0 as *const u32;
    let _invalid_ref = unsafe { &*ptr };
}

#[kani::proof]
#[kani::should_panic]
pub fn rust_ub_should_fail() {
    let ptr = 10 as *const u32;
    let _invalid_read = unsafe { *ptr };
}

```

Running this with kani:
```
$ kani --output-format terse ub_checks.rs
Kani Rust Verifier 0.55.0 (standalone)
Checking harness rust_ub_should_fail...

VERIFICATION RESULT:
 ** 1 of 2 failed
Failed Checks: misaligned pointer dereference: address must be a multiple of its type's alignment
 File: "ub_checks.rs", line 12, in rust_ub_should_fail

VERIFICATION:- SUCCESSFUL (encountered one or more panics as expected)
Verification Time: 0.020638932s

Checking harness rust_ub_fails...

VERIFICATION RESULT:
 ** 1 of 2 failed
Failed Checks: dereference failure: pointer invalid
 File: "ub_checks.rs", line 5, in rust_ub_fails

VERIFICATION:- FAILED (encountered failures other than panics, which were unexpected)
Verification Time: 0.018083837s

Summary:
Verification failed for - rust_ub_fails
Complete - 1 successfully verified harnesses, 1 failures, 2 total.

Issue #2997, opened 2024-02-06
Users should be able to create contracts suitable for verification but not stubbing


celinval:
Requested feature: Enable contract verification for functions that Kani cannot use as stub today.
Use case: Verification of functions that return types that do not implement `Arbitrary`.
Link to relevant documentation (Rust reference, Nomicon, RFC):

Test case:
```rust
impl<T> NonNull<T> {
    pub unsafe fn as_ref<'a>(&self) -> &'a T { /* body */ }
```

## Context

Kani today automatically synthesize the contract into a verification wrapper and a stub. The stub, however, rely on the fact that the return value, and the target for the modifies clauses, to implement `Arbitrary` trait. Thus, writing a simple contract for the example above would fail to compile.

## Possible solution

As discussed with @JustusAdam, one possible solution is to create a new intrinsic for Kani, which tries to instantiate `kani::any()` for the return / modifies type. If the instantiation fails, we can trigger a compilation error. This will only fails if a harness tries to instantiate the generated stub.

JustusAdam:
I think you should include in the description that we would want to emit a warning though or perhaps fail at compile time from a contract-stubbed harness if we encounter the intrinsic.

celinval:
Unfortunately, the intrinsic will have to be marked as public with hidden documentation.

PR #3843, opened 2025-01-20
Upgrade toolchain to nightly-2025-01-22

tautschnig:
Changes required due to
- rust-lang/rust#135344: Less unsafe in `dangling`/`without_provenance`
- rust-lang/rust#135661: Stabilize `float_next_up_down`
- rust-lang/rust#135709: Temporarily bring back `Rvalue::Len`

Resolves: #3840

### DIFF
--- a/tests/expected/function-contract/valid_ptr.rs
+++ b/tests/expected/function-contract/valid_ptr.rs
@@ -28,7 +28,6 @@ mod pre_condition {
     }
 
     #[kani::proof_for_contract(read_ptr)]
-    #[kani::should_panic]
     fn harness_invalid_ptr() {
         let ptr = std::ptr::NonNull::<i32>::dangling().as_ptr();
         assert_eq!(unsafe { read_ptr(ptr) }, -20);
--- a/kani-compiler/src/kani_middle/points_to/points_to_analysis.rs
+++ b/kani-compiler/src/kani_middle/points_to/points_to_analysis.rs
@@ -575,7 +575,7 @@ impl<'tcx> PointsToAnalysis<'_, 'tcx> {
                 // The same story from BinOp applies here, too. Need to track those things.
                 self.successors_for_operand(state, operand)
             }
-            Rvalue::NullaryOp(..) | Rvalue::Discriminant(..) => {
+            Rvalue::NullaryOp(..) | Rvalue::Discriminant(..) | Rvalue::Len(_) => {
                 // All of those should yield a constant.
                 HashSet::new()
             }
--- a/kani-compiler/src/main.rs
+++ b/kani-compiler/src/main.rs
@@ -16,7 +16,6 @@
 #![feature(f128)]
 #![feature(f16)]
 #![feature(non_exhaustive_omitted_patterns_lint)]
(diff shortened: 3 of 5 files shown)


tests/expected/function-contract/valid_ptr.rs
```diff
@@ -28,7 +28,6 @@ mod pre_condition {
     }
 
     #[kani::proof_for_contract(read_ptr)]
-    #[kani::should_panic]
```
celinval, line 31: can you please add a comment explaining what's going on?

celinval (approved):

tests/expected/function-contract/valid_ptr.rs
tautschnig: I’m not even sure which part of the harness we previously expected to panic?

carolynzech: If you run this harness with the v0.58 release, you get:

```rust
SUMMARY:
 ** 1 of 126 failed (4 unreachable)
 File: "/Users/runner/work/kani/kani/library/kani/src/lib.rs", line 54, in kani::mem::is_inbounds::<i32>

VERIFICATION:- SUCCESSFUL (encountered one or more panics as expected)
```

This check counts as a panic because we encode it as an `assert`: https://github.com/model-checking/kani/blob/3138a96c0d0744e290cb9f7a255ed039bbac2518/library/kani_core/src/mem.rs#L213-L216

and if `#[should_panic]` is enabled, any failing assertions count as "expected" panics: https://github.com/model-checking/kani/blob/3138a96c0d0744e290cb9f7a255ed039bbac2518/kani-driver/src/call_cbmc.rs#L467-L470

With this toolchain change, we now have the two additional failed checks, which don't count as panics that we expect, hence why the `#[should_panic]` attribute isn't enough to make verification succeed. Presumably something about the changes made to the `dangling` function on the rustc end triggers the dereference failure on CBMC's end.

I believe this is related to #3571; I would think we would want this unallocated memory check to fail verification no matter what, instead of being counted as an expected panic.

tests/expected/function-contract/valid_ptr.rs
```diff
@@ -28,7 +28,6 @@ mod pre_condition {
...
     fn harness_invalid_ptr() {
         let ptr = std::ptr::NonNull::<i32>::dangling().as_ptr();
         assert_eq!(unsafe { read_ptr(ptr) }, -20);
```
carolynzech, line 33: Would you mind enabling the `post_condition` mod below? #2997 is closed.

tautschnig: Done in #3847 is this is very much unrelated to the toolchain upgrade.

tests/expected/function-contract/valid_ptr.rs
tautschnig: Thanks a lot for the clarification. I'd argue that the previous `should_panic` was just wrong and we are now (with the additional lines in the `expected` outcome) getting the desired behaviour. I'll leave the auditing of the unallocated-memory-check to #3571, and will thus resolve this conversation.

PR #3849, opened 2025-01-23
Add reference for loop contracts

qinheping:
This PR adds reference of loop contracts:
Rendered version: https://github.com/qinheping/kani/blob/docs/loop-contracts/docs/src/reference/experimental/loop-contracts.md 


docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+    x = x - 1;
+    assert!(x >= 1); // check that loop invariant is inductive.
+    assume!(false) // kill this proof path.
+}
+// proof path 2:
+//   loop invariant is satisfied and loop guard is violated.
+assert!(x == 1);
+```
+That is, we assume that we are in an arbitrary iteration after checking that the loop invariant holds for the base case. With the inductive hypothesis (`assume!(x >= 1);`),
+we will either enter the loop (proof path 1) and leave the loop (proof path 2). We prove the two paths separately by killing the path 1 with `assume!(false);`.
```
zhassan-aws, line 91:
```suggestion
we will either enter the loop (proof path 1) or leave the loop (proof path 2). We prove the two paths separately by killing path 1 with `assume!(false);`.
```
?

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+Note that all assertions after `assume!(false)` will be ignored as `false => p` can be deduced as `true` for any `p`.
+
+In the proof path 1, we prove properties inside the loop and at last check that the loop invariant is inductive.
```
zhassan-aws, line 94:
```suggestion
In proof path 1, we prove properties inside the loop and at last check that the loop invariant is inductive.
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+
+In the proof path 2, we prove properties after leaving the loop. As we leave the loop only when the loop guard is violated, the post condition of the loop can be expressed as
```
zhassan-aws, line 96:
```suggestion
In proof path 2, we prove properties after leaving the loop. As we leave the loop only when the loop guard is violated, the post condition of the loop can be expressed as
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+`!guard && inv`, which is `x <= 1 && x >= 1` in the example. The postcondition implies `x == 1`---the property we want to prove at the end of `simple_loop_with_loop_contracts`.
```
zhassan-aws, line 97:
```suggestion
`!guard && inv`, which is `x <= 1 && x >= 1` in the example. The postcondition implies `x == 1`—the property we want to prove at the end of `simple_loop_with_loop_contracts`.
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+## Limitations
+
+Loop contracts comes with the following limitations.
+
+1. Besides `while` loops, there are three other kinds of loops that we don't support loop contracts for: [`loop` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#infinite-loops)
```
zhassan-aws, line 104:
```suggestion
1. Only `while` loops are supported. The other three kinds of loops are not supported: [`loop` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#infinite-loops)
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+   , [`while let` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#predicate-pattern-loops), and [`for` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#iterator-loops).
+2. Kani infer *loop modifies* with alias analysis. Loop modifies are those variables we assume to be arbitrary in the inductive hypothesis, and should cover all memory locations that are written to during the execution of the loops. Proof will fail if the inferred loop modifies miss some targets written in the loops.
```
zhassan-aws, line 106:
```suggestion
2. Kani infers *loop modifies* with alias analysis. Loop modifies are those variables we assume to be arbitrary in the inductive hypothesis, and should cover all memory locations that are written to during the execution of the loops. A proof will fail if the inferred loop modifies misses some targets written in the loops.
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+   We observes this happen when some fields of structs are modified by some other functions called in the loops.
```
zhassan-aws, line 107:
```suggestion
   We observed this happens when some fields of structs are modified by some other functions called in the loops.
```

zhassan-aws: Is there an issue that captures this scenario? If so, can we add a link to it here?

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+3. Kani doesn't check if a loop will always terminate in proofs with loop contracts. So it could be that some properties are proved successfully with Kani but actually are unreachable due to the non-termination of some loops.
```
zhassan-aws, line 108: Will they be reported as [unreachable](https://model-checking.github.io/kani/verification-results.html#check-results) in this case? If so, we should point that out.

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+4. We don't check if loop invariants are side-effect free. A loop invariant with side-effect could lead to unsound proof result. Be sure that the specified loop contracts are side-effect free.
```
zhassan-aws, line 109:
```suggestion
4. We don't check if loop invariants are side-effect free. A loop invariant with a side effect could lead to an unsound proof result. Make sure that the specified loop contracts are side-effect free.
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+
+ Now let's run the proof with loop contracts through kani:
+ ``` bash
+kani simple_loop_with_loop_contracts.rs  -Z loop-contracts
+ ```
+
+
+## Loop contracts for `while` loops
+
+> **Syntax**
```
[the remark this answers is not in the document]
qinheping: It is the same style as Rust reference https://doc.rust-lang.org/reference/expressions/loop-expr.html#predicate-loops.

docs/src/reference/experimental/loop-contracts.md
qinheping: No, they won't be reported as unreachable.

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,109 @@
...
+    };
+
+    assert!(x == 1);
+}
+```
+
+Here, the loop invariant `#[kani::loop_invariant(x >= 1)]` specifies that the condition `x >= 1` must hold true at the start of each iteration before the loop guard is
+ checked. Once Kani verifies that the loop invariant is inductive, it will use the invariant to abstract the loop and avoid unwinding. 
+
+ Now let's run the proof with loop contracts through kani:
```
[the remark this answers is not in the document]
qinheping: Yes, great idea!

qinheping:
@zhassan-aws , thank you for the feedbacks! I addressed the comments above.

docs/src/reference/experimental/loop-contracts.md
zhassan-aws: Do you know why they aren't reported as unreachable?

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -29,6 +29,9 @@ Kani needs to unwind the loop for `u64::MAX` iterations, which is computationall
 With loop contracts, we can specify the loop’s behavior using invariants. For example:
 
 ``` Rust
```
zhassan-aws, line 32: Are these necessary to include here?

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,154 @@
...
+
+Consider the following example:
+
+``` Rust
+fn simple_loop() {
+    let mut x: u64 = kani::any_where(|i| *i >= 1);
+
+    while x > 1 {
+        x = x - 1;
+    };
```
zhassan-aws, line 20:
```suggestion
    }
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,154 @@
...
+
+    assert!(x == 1);
+}
+```
+
+In this program, the loop repeatedly decrements `x` until it equals `1`. Because we haven't specify an upper bound for `x`, to verify this function,
```
zhassan-aws, line 26:
```suggestion
In this program, the loop repeatedly decrements `x` until it equals `1`. Because we haven't specified an upper bound for `x`, to verify this function,
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -0,0 +1,154 @@
...
+#![feature(stmt_expr_attributes)]
+#![feature(proc_macro_hygiene)]
+
+fn simple_loop_with_loop_contracts() {
+    let mut x: u64 = kani::any_where(|i| *i >= 1);
+
+    #[kani::loop_invariant(x >= 1)]
+    while x > 1 {
+        x = x - 1;
+    };
```
zhassan-aws, line 41:
```suggestion
    }
```

docs/src/reference/experimental/loop-contracts.md
qinheping: Yes. We execute only one non-deterministic iteration of the loop when applying loop contracts. So won't know if there will be infinite number of iteration to be executed. In CBMC, we have decreases clauses to check if there are some quantities decreasing along the execution of the loop. Loops will terminate if such quantities exist. However, decreases clauses are not implemented in Kani yet.

zhassan-aws (approved):

docs/src/reference/experimental/loop-contracts.md
zhassan-aws: Got it, thanks.

docs/src/reference/experimental/loop-contracts.md
qinheping: It is necessary for use attribute on `stmt_expr` like loops.

carolynzech:
@qinheping Is this blocked on anything? If not, can you resolve that last comment and merge?

PR #3851, opened 2025-01-24
Move documentation of kani_core modules to right places

### LINKED ISSUE #3815: Missing items in the documention for the `mem` module
There’s a small issue in the documentation for [`mem::can_dereference`](https://model-checking.github.io/kani/crates/doc/kani/mem/fn.can_dereference.html) and similar functions. It refers to items in `kani`'s `mem` module using `crate::mem`, but these items aren’t actually present in the documentation for the `mem` module of `kani`. Instead, they’re found in the documentation for the `mem` module of `kani_core`, where functions like `can_dereference` are defined.

qinheping:
Resolves #3815 

[Documentations](https://github.com/model-checking/kani/blob/main/library/kani_core/src/mem.rs#L3) of `kani_core` modules doesn't show up correctly on https://model-checking.github.io/kani/crates/doc/kani/mem/index.html. This PR move the comments to the right place so that they will show up correctly.

### DIFF
--- a/library/kani_core/src/float.rs
+++ b/library/kani_core/src/float.rs
@@ -1,6 +1,8 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+// This module contains functions useful for float-related checks.
+
 #[allow(clippy::crate_in_macro_def)]
 #[macro_export]
 macro_rules! generate_float {
--- a/library/kani_core/src/float.rs
+++ b/library/kani_core/src/float.rs
@@ -1,8 +1,6 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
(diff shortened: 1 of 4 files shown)


carolynzech (changes requested):
Instead of deleting the documentation from each of the files, can you keep a sentence or two about what they do, and then refer to the new location of the docs for full details? I agree that our full docs should be visible to our users, but for internal development, it's still useful to have some comments at the top of the file explaining what it does. For example, in `kani_core::mem`, can you add a sentence back like "This module contains functions useful for checking unsafe memory accesses. For full documentation, see `lib.rs`"?

(Also, were you able to test this at all to demonstrate that these moves fix the problem?)

qinheping:
@carolynzech Thank you! I tested on my end by compile the rustdoc, and checked if the descriptions of the modules are all correct.

PR #3852, opened 2025-01-24
Remove the openssl-devel package from dependencies

zhassan-aws:
I confirmed the `openssl-devel` package is not needed to build Kani using fresh AL2 instances (both `aarch64` and `x86_64`).

### DIFF
--- a/scripts/setup/al2/install_deps.sh
+++ b/scripts/setup/al2/install_deps.sh
@@ -13,7 +13,6 @@ DEPS=(
   cmake3
   gcc10-c++
   git
-  openssl-devel
   wget
 )
 


carolynzech (approved):
Funny, I wonder why it was there in the first place then? Did we used to need it but don't anymore?

zhassan-aws:
I tried to dig in this history to figure out why it was added, but couldn't see an obvious reason. It might be because CBMC installs it as part of its build process:
https://github.com/diffblue/cbmc/blob/develop/.github/workflows/build-and-test-Linux.yaml#L24

@adpaco-aws, do you know?

adpaco-aws:
I can't remember, and the only reason I can think of is that the dependency may have been missing from a certain kind of AL2 instances. But there's no reason to keep if it's been tested.

feliperodri (approved):

PR #3853, opened 2025-01-24
Fix validity checks for `char`

### LINKED ISSUE #3241: Valid value checks does not check `char` surrogate values

celinval:
Ensure value validity checks for `char` in Kani matches the Rust documentation:

> A char value must not be a surrogate (i.e., must not be in the range
  0xD800..=0xDFFF) and must be equal to or less than char::MAX.

The existing code was relying on the compiler ABI information, which can only express one continuous value range for value validity. But `char` is a special case, and the Rust compiler understands the gap in valid `char`. This change makes Kani compiler aware of this gap too.

Resolves #3241

### DIFF
--- a/kani-compiler/src/kani_middle/transform/check_values.rs
+++ b/kani-compiler/src/kani_middle/transform/check_values.rs
@@ -32,7 +32,7 @@ use stable_mir::mir::{
     Statement, StatementKind, Terminator, TerminatorKind,
 };
 use stable_mir::target::{MachineInfo, MachineSize};
-use stable_mir::ty::{AdtKind, IndexedVal, MirConst, RigidTy, Ty, TyKind, UintTy};
+use stable_mir::ty::{AdtKind, IndexedVal, MirConst, RigidTy, Span, Ty, TyKind, UintTy};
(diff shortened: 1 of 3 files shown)


zhassan-aws (approved):
Awesome! This has been a longstanding issue. Thanks!

PR #3858, opened 2025-01-27
Stub linker to avoid missing symbols errors

### LINKED ISSUE #3817: kani fails to compile a crate but cargo can build  it
I would like to use kani, but it fails to build one of my dependencies (`needletail`), even though `cargo build` runs fine.
Step to reproduce:
```bash
git clone https://github.com/onecodex/needletail
cd needletail
cargo build  # OK
```
However, kani fails to build:
```bash
cargo kani
# Kani Rust Verifier 0.57.0 (cargo plugin)
# [...]
# error: linking with `cc` failed: exit status: 1
#  = note: rust-lld: error: version script assignment of 'global' to symbol 'bz_internal_error' failed: symbol not defined
          collect2: error: ld returned 1 exit
# error: could not compile `needletail` (lib) due to 1 previous error
# error: Failed to execute cargo (exit status: 101). Found 1 compilation errors.
```

Note: when I remove `bzip2` from the dependencies of `needletail`, kani works fine, but I don't need to do it for `cargo build`.

Is there a difference in linking between `cargo build` and `cargo kani`?

celinval:
Kani compiler does not generate real object files when compiling a crate and its dependencies for verification. Instead, the compiler generates goto binaries. This can cause linking errors. Instead, stub it with echo.

We should consider creating a stub linker instead, but echo will do it for now.

Note that this change only applies to target artifacts, not those compiled for the host, such as build dependencies and build scripts.

Resolves #3817 

### DIFF
--- a/kani-driver/src/call_single_file.rs
+++ b/kani-driver/src/call_single_file.rs
@@ -181,6 +181,8 @@ impl KaniSession {
                 "-Z",
                 "mir-enable-passes=-RemoveStorageMarkers",
                 "--check-cfg=cfg(kani)",
+                // Do not invoke the linker since the compiler will not generate real object files
+                "-Clinker=echo",
             ]
             .map(OsString::from),
         );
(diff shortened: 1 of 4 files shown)


zhassan-aws (approved):
Thanks!

PR #3873, opened 2025-02-04
remove flag float-overflow-check

### LINKED ISSUE #3620: Disable floating-point overflow checks
Is there a way to specifically disable the floating-point overflow checks?
I only want to ignore the following errors (and nothing else):
```sh
Failed Checks: NaN on multiplication
Failed Checks: NaN on addition
Failed Checks: arithmetic overflow on floating-point addition
Failed Checks: arithmetic overflow on floating-point multiplication
```
If it's not possible, is there a way to implement custom `add`/`mul` to make kani happy? Any tips.

rajath-mk:
Resolves #3620

Removing default flag float overflow check, which reports overflow for arithmetic operations over floating-point numbers that result in +/-Inf. This doesn't panic in Rust and it shouldn't be consider a verification failure by default.

Users can enable them using --cbmc-args as and when required.

### DIFF
--- a/tests/ui/cbmc_checks/float-overflow/check_message_overflow.rs
+++ b/tests/ui/cbmc_checks/float-overflow/check_message_overflow.rs
@@ -1,7 +1,6 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 //
-// Check we don't print temporary variables as part of CBMC messages.
 // This test verifies that Kani does not report floating-point overflow by default
 // for operations that result in +/-Infinity.
 extern crate kani;


zhassan-aws:
Thanks! Can you add a test that ensures no overflow is reported?

rajath-mk:
Added this test with the `// kani-flags: --enable-unstable --cbmc-args --float-overflow-check` flag: 
```
//
// Check we don't print temporary variables as part of CBMC messages.
// kani-flags: --enable-unstable --cbmc-args --float-overflow-check
extern crate kani;

use kani::any;

// Use the result so rustc doesn't optimize them away.
fn dummy(result: f32) -> f32 {
    result
}

#[kani::proof]
fn main() {
    dummy(any::<f32>() + any::<f32>());
    dummy(any::<f32>() - any::<f32>());
    dummy(any::<f32>() * any::<f32>());
    dummy(any::<f32>() / any::<f32>()); // This is not emitting CBMC check.
    dummy(any::<f32>() % any::<f32>()); // This is not emitting CBMC check.
    dummy(-any::<f32>()); // This is not emitting CBMC check.
}
```

tests/ui/cbmc_checks/float-overflow/check_message_overflow.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+//
+// Check we don't print temporary variables as part of CBMC messages.
```
zhassan-aws, line 4: Can you update the description to mention what this test checks?

tests/ui/cbmc_checks/float-overflow/check_message_overflow.rs
```diff
@@ -0,0 +1,23 @@
...
+extern crate kani;
+
+use kani::any;
```
zhassan-aws, line 7: This doesn't seem to be needed.

tests/ui/cbmc_checks/float-overflow/check_message_overflow.rs
```diff
@@ -0,0 +1,23 @@
...
+
+// Use the result so rustc doesn't optimize them away.
+fn dummy(result: f32) -> f32 {
+    result
+}
+
+#[kani::proof]
+fn main() {
+    let a = kani::any_where(|&x: &f32| x.is_finite() && x.abs() < 1e20);
```
zhassan-aws, line 16: Is the `x.abs() < 1e20` needed?

zhassan-aws: Also, I think it should be sufficient to just assume non NaN, i.e.
```suggestion
    let a = kani::any_where(|&x: &f32| !x.is_nan());
```
Can you check if this works?

rajath-mk: I added the `x.abs() < 1e20` assuming it may cause problems with multiplication. But that wasn't the case, and `let a = kani::any_where(|&x: &f32| x.is_finite());` works just fine.

 `let a = kani::any_where(|&x: &f32| !x.is_nan());` on the other hand is failing since the individual numbers themselves are not NaN but the result is overflowing. --> Getting dereferencing errors.

tests/ui/cbmc_checks/float-overflow/check_message_overflow.rs
rajath-mk: Removed.

tests/ui/cbmc_checks/float-overflow/check_message_overflow.rs
zhassan-aws: > on the other hand is failing since the individual numbers themselves are not NaN but the result is overflowing. --> Getting dereferencing errors.

Can you post the failure? Since we're disabling CBMC's float overflow checks, I wouldn't expect it to fail.

zhassan-aws: For example, this program runs without issues: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c8191caef9b78b217aa6654a01483f20

zhassan-aws: Oh, but perhaps multiplying `f32::INFINITY` by 0.0 results in a NaN and CBMC flags this because of the NaN checks?

rajath-mk: Yes. It is being multiplied by 0, and also due to `f32::INFINITY` - `f32::INFINITY`

zhassan-aws: Confirmed offline that CBMC's NaN checks fail if we only restrict the values to `!is_nan()`. Perhaps we should also consider disabling CBMC's NaN checks since produce a NaN does not seem to be UB.

zhassan-aws: Can you file a tracking issue to consider disabling CBMC's NaN checks as well?

rajath-mk: Sure. #3875

tests/ui/cbmc_checks/float-overflow/check_message_overflow.rs
zhassan-aws, line 4: Can you remove this comment?

zhassan-aws (approved):

Issue #3567, opened 2024-10-03
Modify its contract annotation to allow users to specify safety contract separate from correctness / panic freedom. Safety contract should only be accepted in unsafe funtions.


celinval:

zhassan-aws:
Since panic conditions are part of a function specification and documentation (see [example](https://doc.rust-lang.org/std/string/struct.String.html#panics)), I think we might need to create a new attribute for describing a function's panic condition (e.g. `kani::panics_if`. With this attribute, Kani should:
1. Verify that the function *must* panic under the specified condition
2. Verify that there is no UB and the post-condition holds under the non-panicking condition

For example, for `unwrap`, the contract would be along the lines of:
```rust
#[kani::requires(true)] // the function is safe
#[kani::panics_if(self.is_none())]
#[kani::ensures(|result| if let Some(x) = self { result == x })]
```

BusyBeaver-42:
I need to express something similar, but I believe this use case is not covered by `panics_if`. The idea I want to express is `panics_or`, meaning: "The function either panics, or the program is in a consistent state when it returns."

## Example: Sorting a slice

The function [`bidirectional_merge`](https://github.com/rust-lang/rust/blob/master/library/core/src/slice/sort/shared/smallsort.rs#L765) (in the [`smallsort` module](https://github.com/rust-lang/rust/blob/master/library/core/src/slice/sort/shared/smallsort.rs) of the Rust standard library) might be called with a user-provided comparison function when sorting a slice. Since the slice sorting methods are safe, they cannot rely on the correctness of the comparison function to guarantee safety. In particular, the safety of `bidirectional_merge` cannot depend on the correctness of `is_less`.

If `is_less` is incorrect, `bidirectional_merge` might lead to an inconsistent state and will panic only if such a state is reached. The issue is that whether an inconsistent state is reached depends on the program's execution path. I do not see how this can be expressed with `panics_if`, as it would need to reference local variables that are neither arguments nor outputs of `bidirectional_merge`.

This limitation arises because the behavior of `bidirectional_merge` cannot always be determined solely from its inputs. For example, if `is_less` simply returns a random boolean, then regardless of any pre-conditions, any execution path is possible. As a result, given the same observable input state, `bidirectional_merge` may or may not panic.

gmorenz:
I experimented with this on top of creusot a few years ago with quite a bit of success (though I eventually abandoned the project due to unrelated pain of handling creusot's WIP research prototype state).

[...]

PR #3893, opened 2025-02-17
RFC: Attribute to distinguish safety preconditions from panic freedom

tautschnig:
This is to discuss the proposed solution for #3567 (Separate safety contract from correctness / panic freedom).

### DIFF
--- a/rfc/src/SUMMARY.md
+++ b/rfc/src/SUMMARY.md
@@ -19,3 +19,4 @@
 - [0011-source-coverage](rfcs/0011-source-coverage.md)
 - [0012-loop-contracts](rfcs/0012-loop-contracts.md)
 - [0013-list](rfcs/0013-list.md)
+- [0013-may-panic-if-attr](rfcs/0014-may-panic-if-attr.md)
--- a/rfc/src/SUMMARY.md
+++ b/rfc/src/SUMMARY.md
@@ -19,4 +19,4 @@
 - [0011-source-coverage](rfcs/0011-source-coverage.md)
 - [0012-loop-contracts](rfcs/0012-loop-contracts.md)
 - [0013-list](rfcs/0013-list.md)
-- [0013-may-panic-if-attr](rfcs/0014-may-panic-if-attr.md)
+- [0014-may-panic-if-attr](rfcs/0014-may-panic-if-attr.md)
--- a/rfc/src/rfcs/0014-may-panic-if-attr.md
+++ b/rfc/src/rfcs/0014-may-panic-if-attr.md
@@ -36,8 +36,9 @@ change Kani's verification behavior as follows:
 1. Kani will report successful verification when all properties hold and no
    panic can occur. (This behavior is unchanged.)
 2. Kani will also report successful verification when all properties hold, no
-   panic occurs when the condition given with `may_panic_if` holds, yet some
-   panic occurs when the condition does not hold.
+   panic occurs when the negation of the condition given with `may_panic_if`
+   holds, yet some panic occurs when the negation of the condition does not
+   hold.
 3. Else Kani reports verification failure. (This behavior is unchanged.)
 
 The following example describes what the overall contract for `unwrap` would
--- a/rfc/src/rfcs/0014-may-panic-if-attr.md
+++ b/rfc/src/rfcs/0014-may-panic-if-attr.md
@@ -1,6 +1,6 @@
 - **Feature Name:** Attribute to distinguish safety preconditions from panic freedom (`may-panic-if-attribute`)
 - **Feature Request Issue:** [#3567](https://github.com/model-checking/kani/issues/3567)
-- **RFC PR:** *Link to original PR*
+- **RFC PR:** [#3893](https://github.com/model-checking/kani/pull/3893)
 - **Status:** Under Review
 - **Version:** 0
 - **Proof-of-concept:** Not yet
@@ -36,7 +36,7 @@ change Kani's verification behavior as follows:
 1. Kani will report successful verification when all properties hold and no
    panic can occur. (This behavior is unchanged.)
 2. Kani will also report successful verification when all properties hold, no
-   panic occurs when the condition given with `may_pani_if` holds, yet some
+   panic occurs when the condition given with `may_panic_if` holds, yet some
    panic occurs when the condition does not hold.
 3. Else Kani reports verification failure. (This behavior is unchanged.)
 
@@ -56,7 +56,7 @@ Initial implementation suggestion: we will run Kani twice for any such harness
 (unless the condition is trivially `true` or `false`), once while assuming the
 condition (and then checking that no properties other than reachability checks
 fail); if that run succeeded we remove the assumptions and, similarly to
-`should_fail`, check that the only failing properties are panics and not safety
+`should_panic`, check that the only failing properties are panics and not safety
 checks.
 
 ## Rationale and alternatives
--- a/rfc/src/rfcs/0014-may-panic-if-attr.md
+++ b/rfc/src/rfcs/0014-may-panic-if-attr.md
@@ -37,8 +37,7 @@ change Kani's verification behavior as follows:
    panic can occur. (This behavior is unchanged.)
 2. Kani will also report successful verification when all properties hold, no
    panic occurs when the negation of the condition given with `may_panic_if`
-   holds, yet some panic occurs when the negation of the condition does not
-   hold.
+   holds, yet some panic occurs when the condition holds.
 3. Else Kani reports verification failure. (This behavior is unchanged.)
 
 The following example describes what the overall contract for `unwrap` would
--- a/rfc/src/rfcs/0014-may-panic-if-attr.md
+++ b/rfc/src/rfcs/0014-may-panic-if-attr.md
@@ -1,0 +1,75 @@
+- **Feature Name:** Attribute to distinguish safety preconditions from panic freedom (`may-panic-if-attribute`)
+- **Feature Request Issue:** [#3567](https://github.com/model-checking/kani/issues/3567)
+- **RFC PR:** *Link to original PR*
+- **Status:** Under Review
+- **Version:** 0
+- **Proof-of-concept:** Not yet
+
+-------------------
+
+## Summary
+
+Kani users want to prove absence of undefined behavior ("safety") while
+distinguishing it from panic freedom.
+
+## User Impact
+
+With the `requires` clauses of function contracts we have enabled modular safety
+verification, permitting users to prove the absence of undefined behavior when
+preconditions are met.
+In some cases, however, users may want to go further and
+1. prove the absence of unexpected panics in presence of expected panics
+   (the presence of the latter can already be demonstrated with the
+   `should_panic` attribute);
+2. formally describe the conditions under which a panic is possible;
+3. prove total correctness by precisely describing the conditions under which a
+   panic occur, upon which the post-conditions are no longer guaranteed.
+
+## User Experience
+
+Users will be able to add an attribute
+`#[kani::may_panic_if(<Boolean expression>)]`
+to any function that has a contract (i.e., at least one of `ensures` or
+`requires`) clause.
+When such an attribute is present, users will add `-Z may-panic-if-attribute` to
+change Kani's verification behavior as follows:
+1. Kani will report successful verification when all properties hold and no
+   panic can occur. (This behavior is unchanged.)
+2. Kani will also report successful verification when all properties hold, no
+   panic occurs when the condition given with `may_pani_if` holds, yet some
+   panic occurs when the condition does not hold.
+3. Else Kani reports verification failure. (This behavior is unchanged.)
+
+The following example describes what the overall contract for `unwrap` would
+thus look like:
+```rust
+#[kani::requires(true)] // the function is safe
+#[kani::may_panic_if(self.is_none())]
+#[kani::ensures(|result| Some(result) == self)]
+```
+
+## Software Design
+
+**We recommend that you leave the Software Design section empty for the first version of your RFC**.
+
+Initial implementation suggestion: we will run Kani twice for any such harness
+(unless the condition is trivially `true` or `false`), once while assuming the
+condition (and then checking that no properties other than reachability checks
+fail); if that run succeeded we remove the assumptions and, similarly to
+`should_fail`, check that the only failing properties are panics and not safety
+checks.
+
+## Rationale and alternatives
+
+The linked issue contains suggestions for alternative means to describe panic
+conditions, most notably `panic_if`. This was ruled out as it may at times not
+be possible to exactly describe the conditions under which a panic _must_ occur
+with the syntactic elements in scope at the point of stating the condition.
+
+## Open questions
+
+- Should we permit multiple occurrences of `may_panic_if`?
+
+## Out of scope / Future Improvements
+
+n/a


rfc/src/SUMMARY.md
```diff
@@ -19,3 +19,4 @@
 - [0011-source-coverage](rfcs/0011-source-coverage.md)
 - [0012-loop-contracts](rfcs/0012-loop-contracts.md)
 - [0013-list](rfcs/0013-list.md)
+- [0013-may-panic-if-attr](rfcs/0014-may-panic-if-attr.md)
```
carolynzech, line 22:
```suggestion
- [0014-may-panic-if-attr](rfcs/0014-may-panic-if-attr.md)
```

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
+- **Feature Name:** Attribute to distinguish safety preconditions from panic freedom (`may-panic-if-attribute`)
+- **Feature Request Issue:** [#3567](https://github.com/model-checking/kani/issues/3567)
+- **RFC PR:** *Link to original PR*
```
carolynzech, line 3:
```suggestion
- **RFC PR:** [#3893](https://github.com/model-checking/kani/pull/3893)
```

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+`#[kani::may_panic_if(<Boolean expression>)]`
+to any function that has a contract (i.e., at least one of `ensures` or
+`requires`) clause.
+When such an attribute is present, users will add `-Z may-panic-if-attribute` to
+change Kani's verification behavior as follows:
+1. Kani will report successful verification when all properties hold and no
+   panic can occur. (This behavior is unchanged.)
+2. Kani will also report successful verification when all properties hold, no
+   panic occurs when the condition given with `may_pani_if` holds, yet some
+   panic occurs when the condition does not hold.
```
carolynzech, line 40: I'm confused about 2. -- I would have thought `may_panic_if(cond)` is equivalent to "if `cond` holds, then panics are acceptable/expected." But this seems to say that if `cond` holds, panics are _not_ allowed.

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+3. Else Kani reports verification failure. (This behavior is unchanged.)
+
+The following example describes what the overall contract for `unwrap` would
+thus look like:
+```rust
+#[kani::requires(true)] // the function is safe
+#[kani::may_panic_if(self.is_none())]
```
carolynzech, line 47: Coming back to my confusion re: item 2.:

> Kani will also report successful verification when... no panic occurs when the condition given with `may_panic_if` holds

If the condition `self.is_none()` holds, then a panic does happen, so wouldn't verification fail here?

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+Users will be able to add an attribute
+`#[kani::may_panic_if(<Boolean expression>)]`
+to any function that has a contract (i.e., at least one of `ensures` or
+`requires`) clause.
+When such an attribute is present, users will add `-Z may-panic-if-attribute` to
+change Kani's verification behavior as follows:
+1. Kani will report successful verification when all properties hold and no
+   panic can occur. (This behavior is unchanged.)
+2. Kani will also report successful verification when all properties hold, no
+   panic occurs when the condition given with `may_pani_if` holds, yet some
```
carolynzech, line 39:
```suggestion
   panic occurs when the condition given with `may_panic_if` holds, yet some
```

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+#[kani::requires(true)] // the function is safe
+#[kani::may_panic_if(self.is_none())]
+#[kani::ensures(|result| Some(result) == self)]
+```
+
+## Software Design
+
+**We recommend that you leave the Software Design section empty for the first version of your RFC**.
+
+Initial implementation suggestion: we will run Kani twice for any such harness
```
carolynzech, line 55: This seems sound to me, which is obviously our first priority -- I would love if we could come up with something that didn't require running twice, but I don't currently have any ideas.

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+fail); if that run succeeded we remove the assumptions and, similarly to
+`should_fail`, check that the only failing properties are panics and not safety
+checks.
+
+## Rationale and alternatives
+
+The linked issue contains suggestions for alternative means to describe panic
+conditions, most notably `panic_if`. This was ruled out as it may at times not
+be possible to exactly describe the conditions under which a panic _must_ occur
+with the syntactic elements in scope at the point of stating the condition.
```
carolynzech, line 67: Why not `panic_or`? https://github.com/model-checking/kani/issues/3567#issuecomment-2569213652 says:

> The idea I want to express is panics_or, meaning: "The function either panics, or the program is in a consistent state when it returns."

which seems to be to be more consistent with a `may_panic_iff` attribute, rather than `may_panic_if`. Put differently, my intuitive understanding of what a `may_panic_if` attribute _should_ mean from its name alone would be "If `cond` holds, then a panic is allowed." Notably, that doesn't imply the "only if" direction, i.e., there may be other conditions under which a panic is permissible. My understanding of #3567 is that we want this stronger "if and only if guarantee," i.e., we want to be able to say "these are the only conditions under which a panic is permissible."

I think(?) that this "if and only if" behavior is actually what you mean, since your initial implementation suggestion would only work if the specified condition was a complete specification of the circumstances under which panics are allowed. So I suspect this comment is more about bikeshedding than suggesting different implementation behavior, but let me know if that's not the case.

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Open questions
+
+- Should we permit multiple occurrences of `may_panic_if`?
```
carolynzech, line 71: I think it depends whether we implement this as a `may_panic_if`, with a unidirectional implication, or a `may_panic_iff`, with a bidirectional implication. For example, given:

```rust
fn foo(x: u8) {
  if x == 1 {
     panic!("oh no 1!");
  } else if x == 2 {
     panic!("oh no 2!");
  } else {
     println!("all is well!");
  }
}
```
I would think we should allow:
```rust
#[kani::may_panic_if(x == 1)]
#[kani::may_panic_if(x == 2)]
```
which is equivalent to:
```rust
#[kani::may_panic_if(x == 1 || x == 2)]
```
So it's similar to how we allow multiple `#[kani::requires]`, but or-ing the conditions instead of and-ing them.

My opinion is that we should change the attribute to `may_panic_iff` or `panic_or` to enforce that the condition is a complete specification of when the function will panic. In that case, we would only allow the one attribute, since it has to be complete by definition. I can't think of a scenario in which partial specifications are useful, and I worry that allowing them would cause confusion.

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Software Design
+
+**We recommend that you leave the Software Design section empty for the first version of your RFC**.
+
+Initial implementation suggestion: we will run Kani twice for any such harness
+(unless the condition is trivially `true` or `false`), once while assuming the
+condition (and then checking that no properties other than reachability checks
+fail); if that run succeeded we remove the assumptions and, similarly to
+`should_fail`, check that the only failing properties are panics and not safety
```
carolynzech, line 59:
```suggestion
`should_panic`, check that the only failing properties are panics and not safety
```

carolynzech (commented):
Thanks for writing this up!

rfc/src/rfcs/0014-may-panic-if-attr.md
tautschnig: No wonder you were confused - I had forgotten to say "negation of" twice. Fixed.

rfc/src/rfcs/0014-may-panic-if-attr.md
tautschnig: Apologies again, that was me getting the logic wrong in item 2.

rfc/src/rfcs/0014-may-panic-if-attr.md
tautschnig: Doesn't the very same comment that you referred to stipulate that there are situation where we cannot precisely describe the conditions under which a panic can occur (and neither can we precisely describe the situations where a panic cannot occur)? (Such a situation sounds a bit scary to me, but that's what I read into that comment.)

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,76 @@
...
+to any function that has a contract (i.e., at least one of `ensures` or
+`requires`) clause.
+When such an attribute is present, users will add `-Z may-panic-if-attribute` to
+change Kani's verification behavior as follows:
+1. Kani will report successful verification when all properties hold and no
+   panic can occur. (This behavior is unchanged.)
+2. Kani will also report successful verification when all properties hold, no
+   panic occurs when the negation of the condition given with `may_panic_if`
+   holds, yet some panic occurs when the negation of the condition does not
+   hold.
```
carolynzech, line 41:
```suggestion
   holds, yet some panic occurs when the condition holds.
```

rfc/src/rfcs/0014-may-panic-if-attr.md
carolynzech: Yeah, you're right, my reading of this comment was wrong. Upon re-reading, I think he perhaps wanted something like `#[panics_or(cond)]`, which means "either the function panics, or this postcondition holds." I believe that `#[may_panic_if(true)]` and `#[ensures(cond)]` would replicate this behavior.

rfc/src/rfcs/0014-may-panic-if-attr.md
carolynzech: Given the clarification in the thread above about complete specifications being infeasible, I now do think we should allow multiple attributes--we have the precedent with `#[requires]`, and I don't see a real downside to letting people break it up for readability.

carolynzech (approved):

zhassan-aws:
Rendered version: https://github.com/tautschnig/kani/blob/rfc-may-panic/rfc/src/rfcs/0014-may-panic-if-attr.md

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+condition (and then checking that no properties other than reachability checks
+fail); if that run succeeded we remove the assumptions and, similarly to
+`should_panic`, check that the only failing properties are panics and not safety
+checks.
+
+## Rationale and alternatives
+
+The linked issue contains suggestions for alternative means to describe panic
+conditions, most notably `panic_if`. This was ruled out as it may at times not
+be possible to exactly describe the conditions under which a panic _must_ occur
```
zhassan-aws, line 66: Do you have an example?

carolynzech: https://github.com/model-checking/kani/issues/3567#issuecomment-2569213652

zhassan-aws: This sounds to me like a corner case. The more common case is that a function's documentation includes a section of the panic condition (e.g. https://doc.rust-lang.org/std/string/struct.String.html#panics), and thus is clearly defined in terms of the function's input parameters.

carolynzech: Okay, so your argument is that it's better to introduce a stronger `#[panics_if]` that provides a stronger guarantee that it *must* panic, with the tradeoff that it's not applicable to cases like the linked issue?
(Also, there isn't fundamentally a reason we can't just have both, as long as we document the differences well).

zhassan-aws: > #[may_panic], where #[may_panic] just says "if it panics, succeed; otherwise, succeed iff the postcondition holds.

Which would be the same as the `#[should_panic]` that harnesses can be annotated with?

> `(#[panics_if(cond)] is equivalent to #[may_panic(cond)] #[ensures(!cond)])`

I don't see how the two are equivalent since `cond` is essentially a pre-condition, so its not clear to me how adding `#[ensures(!cond)]` would provide the stronger guarantee that `#[panics_if]` provides.

carolynzech: > Which would be the same as the #[should_panic] that harnesses can be annotated with?

Not quite, e.g.:
```rust


#[kani::proof_for_contract(]
#[kani::should_panic]
fn prove_foo() {
    foo();
}
``` 
fails with:
```
VERIFICATION:- FAILED (encountered no panics, but at least one was expected)
```

whereas with `#[may_panic]`, it would succeed. (If I could go back and bikeshed, I would have named `#[should_panic]` `#[must_panic]` instead 😄 )

> I don't see how the two are equivalent since cond is essentially a pre-condition, so its not clear to me how adding #[ensures(!cond)] would provide the stronger guarantee that #[panics_if] provides.

We only check the postcondition if the function does *not* panic, so if we reach the point where we are enforcing the postcondition, `cond` must not hold. (If it does hold, then we should have panicked).
Taking your example from https://github.com/model-checking/kani/issues/3567#issuecomment-2542597602, 
```rust
#[kani::requires(true)] // the function is safe
#[kani::panics_if(self.is_none())]
#[kani::ensures(|result| if let Some(x) = self { result == x })]
```
would become
```rust
#[kani::requires(true)] // the function is safe
#[kani::may_panic_if(self.is_none())]
// `panic_if(cond)` semantics imply that we check the postcondition iff !cond,
// so if are checking the `#[ensures]`, `cond = self.is_none()` cannot hold
#[kani::ensures(|| !self.is_none())]
#[kani::ensures(|result| if let Some(x) = self { result == x })]
```

zhassan-aws: Perhaps it needs to be `#[ensures(old(!cond))]` instead in case the function mutates `self`?

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+3. prove total correctness by precisely describing the conditions under which a
+   panic occur, upon which the post-conditions are no longer guaranteed.
+
+## User Experience
+
+Users will be able to add an attribute
+`#[kani::may_panic_if(<Boolean expression>)]`
+to any function that has a contract (i.e., at least one of `ensures` or
+`requires`) clause.
+When such an attribute is present, users will add `-Z may-panic-if-attribute` to
```
celinval, line 34: I'm wondering if this should be part of the function contracts RFC since it hasn't been stabilized yet.

Users would also be able to reuse `-Z function-contracts`?

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+-------------------
+
+## Summary
+
+Kani users want to prove absence of undefined behavior ("safety") while
+distinguishing it from panic freedom.
+
+## User Impact
+
+With the `requires` clauses of function contracts we have enabled modular safety
```
celinval, line 17: I think this RFC contradicts the function contracts one. We have never restricted `requires` to safety conditions. If this is the way we want to go, the contracts RFC should be updated and we should consider at least warning users that add `requires` to safe functions.

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+preconditions are met.
+In some cases, however, users may want to go further and
+1. prove the absence of unexpected panics in presence of expected panics
+   (the presence of the latter can already be demonstrated with the
+   `should_panic` attribute);
+2. formally describe the conditions under which a panic is possible;
+3. prove total correctness by precisely describing the conditions under which a
+   panic occur, upon which the post-conditions are no longer guaranteed.
+
+## User Experience
```
celinval, line 28: How will this affect stubbing with contracts?

celinval (commented):
Thanks @tautschnig

rfc/src/rfcs/0014-may-panic-if-attr.md
carolynzech: +1 to this idea.

rfc/src/rfcs/0014-may-panic-if-attr.md
carolynzech: I don't know that it's a contradiction--we have enabled modular safety verification, that's just not _all_ that we've enabled (i.e., you can write preconditions for safe functions too).
It so happens that we've found preconditions most useful for unsafe functions, but if people want to use it differently (perhaps as a form of documentation on a safe function), I have no issue with that, and I don't think anything in this RFC contradicts that?

rfc/src/rfcs/0014-may-panic-if-attr.md
```diff
@@ -0,0 +1,75 @@
...
+2. Kani will also report successful verification when all properties hold, no
+   panic occurs when the negation of the condition given with `may_panic_if`
+   holds, yet some panic occurs when the condition holds.
+3. Else Kani reports verification failure. (This behavior is unchanged.)
+
+The following example describes what the overall contract for `unwrap` would
+thus look like:
+```rust
+#[kani::requires(true)] // the function is safe
+#[kani::may_panic_if(self.is_none())]
```
remi-delmas-3000, line 47: Naming this "may" makes it seem like panic may also not happen.

rfc/src/rfcs/0014-may-panic-if-attr.md
remi-delmas-3000: should, may, are both indicating the behaviour is optional. I find this misleading

rfc/src/rfcs/0014-may-panic-if-attr.md
celinval: I'm rather confused now. What exactly are you proposing to be the semantic of `requires` and `may_panic`? Do you expect the `may_panic` to be a subset of the `requires` condition or do you expect that the correctness of the function to be described as a conjuction of both?

celinval: Note that in the only example of this PR implies that the correctness contract is a conjunction of `requires` and `may_panic`. It has the following attributes:

```
#[kani::requires(true)] // the function is safe
#[kani::may_panic_if(self.is_none())]
```
This is literally implying that `requires` should always be `true` for safe functions. I.e., `requires` specifies only the safety contract, and it contradicts all the examples from the contracts RFC.

carolynzech: I am quite confused now as well, and I'm not optimistic that we're going to sort this out in Github discussion comments. I would recommend that perhaps @tautschnig schedules an optional meeting to discuss this RFC further--I think we'll make a lot more progress in a discussion than in Github comments. (Alternatively, we can block out more time at the Tuesday meeting for this RFC).

tautschnig:
Just a status note as this RFC has been open without apparent action from my side for quite a while: let's gather data from our standard library verification work to see how many panic-induced verification failures we see, and what we would then deem the most appropriate attribute to rule those out.

PR #3895, opened 2025-02-18
Document demonic non-determinism

tautschnig:
This PR adds developer-facing documentation to clarify what we mean by demonic non-determinism, and why we use it.

### DIFF
--- a/docs/src/demonic-nondet.md
+++ b/docs/src/demonic-nondet.md
@@ -1,0 +1,80 @@
+# Demonic non-determinism in CBMC and Kani
+
+CBMC and Kani use demonic non-determinism to more efficiently prove certain
+classes of program properties.
+This document describes what we mean by "demonic non-determinism," how and why
+CBMC and Kani use it, and what challenges arise as a consequence thereof.
+
+## What is demonic non-determinism?
+
+When we seek to prove that a property holds for all possible values of some
+object then a counterexample demonstrates the existence of at least one such
+value that violates the property.
+Other properties range across all objects or expressions of a particular kind,
+e.g., we want to prove that for all dereference expressions the underlying
+pointer is valid.
+For a counterexample to this property it suffices to find one pointer that
+cannot be safely dereferenced.
+We can use this fact to non-deterministically choose to only track a single
+pointer across an execution via expressions of the form
+`global_pointer = if nondet { new_pointer_to_track } else { global_pointer };`.
+We can now use `global_pointer` in describing properties such that the property
+fails if `new_pointer_to_track` is invalid and `global_pointer` equals
+`new_pointer_to_track`.
+Note that this is an "if" statement, not "if, and only if:" there may also be
+some other invalid pointer `other_pointer` that is invalid at the time of
+dereference, and the back-end solver may choose to resolve the non-determinism
+such that the property fails because `global_pointer` equals `other_pointer`.
+We call such use of non-determinism _demonic non-determinism_, because the
+SAT or SMT-solver will resolve the non-deterministic choice such that a
+counterexample is reported _if_ there exists a model such that the overall
+property fails.
+
+## Why use demonic non-determinism?
+
+For the above example, one way to avoid the non-determinism is to use an array
+or map to track all pointers rather than non-deterministically choosing to track
+a single one.
+The encoding thereof, however, may be much more expensive.  See
+[cbmc#6506](https://github.com/diffblue/cbmc/pull/6506) for the performance
+penalty for such a change from non-determinism to arrays.
+
+## What are disadvantages of demonic non-determinism?
+
+The consequence of this choice of encoding is that any given model provided by
+the back-end SAT/SMT solver can only encode a violation of one property of the
+kind that the demonic non-determinism is used for, even if multiple properties
+can be violated along a single execution path.
+When we now look at assumptions the problem is that these cannot accomplish what
+we’d perhaps expect to get out of them: the `assume` statement with a condition of
+"single global variable does not have the same value as the pointer to be
+dereferenced" does not guarantee that the subsequent dereference is safe with
+regard to objects having gone out of scope: the solver can choose to pick
+non-deterministic values such that the address of the object having gone out of
+scope has never been assigned.
+
+As such, demonic non-determinism poses a problem in context of modular
+verification, where we need to assume that the given preconditions hold.
+Such assumed preconditions would not be preceded by an assertion, and we would
+effectively assume that none of the pointers involved in such preconditions were
+being tracked by the solver.
+
+## Where do CBMC and Kani use demonic non-determinism?
+
+One way we use this today is tracking objects that go out of scope.
+We use a single, global variable (initialized to the null pointer) to store the
+address of objects going out of scope.
+Upon a pointer dereference we then check whether the pointer being dereferenced
+has the same (address) value as the value held in that single global variable.
+If that is true then we know a dead object (one that has gone out of scope) is
+being accessed, amounting to undefined behavior, and we fail that property check.
+
+This setup can obviously only track a single object (the one that has gone out
+of scope most recently).
+By adding non-determinism when assigning the address to that single global
+variable (`global-ptr = nondet ? global-ptr : &object-leaving-scope;`) we leave
+the choice of which object to track to the solver.
+This is sound even in presence of multiple objects (and, possibly, multiple
+invalid pointers), because the solver will
+return at least one counterexample if there is any undefined behavior, and can
+indeed be forced to return all such counterexamples.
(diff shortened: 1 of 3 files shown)


docs/src/demonic-nondet.md
```diff
@@ -0,0 +1,80 @@
...
+fails if `new_pointer_to_track` is invalid and `global_pointer` equals
+`new_pointer_to_track`.
+Note that this is an "if" statement, not "if, and only if:" there may also be
+some other invalid pointer `other_pointer` that is invalid at the time of
+dereference, and the back-end solver may choose to resolve the non-determinism
+such that the property fails because `global_pointer` equals `other_pointer`.
+We call such use of non-determinism _demonic non-determinism_, because the
+SAT or SMT-solver will resolve the non-deterministic choice such that a
+counterexample is reported _if_ there exists a model such that the overall
+property fails.
```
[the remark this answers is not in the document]
carolynzech: I think for those with less background in non-determinism, I'd like a final sentence like:

> Thus, choosing non-deterministically is not the same as choosing randomly--by choosing non-determistically, soundness is ensured, since the solver will always resolve the non-deterministic choice such that it finds violations of the property if they exist.

I think my primary confusion was initially conflating nondeterminism with randomness, so thinking that only tracking one object at a time would be unsound. This is somewhat redundant with what you have--I just want to hammer this point home, since I suspect this will be the most confusing/unintuitive point for people.

docs/src/demonic-nondet.md
```diff
@@ -0,0 +1,80 @@
...
+The encoding thereof, however, may be much more expensive.  See
+[cbmc#6506](https://github.com/diffblue/cbmc/pull/6506) for the performance
+penalty for such a change from non-determinism to arrays.
+
+## What are disadvantages of demonic non-determinism?
+
+The consequence of this choice of encoding is that any given model provided by
+the back-end SAT/SMT solver can only encode a violation of one property of the
+kind that the demonic non-determinism is used for, even if multiple properties
+can be violated along a single execution path.
```
carolynzech, line 47: I don't understand quite what you mean here--could you give an example?

docs/src/demonic-nondet.md
```diff
@@ -0,0 +1,80 @@
...
+When we now look at assumptions the problem is that these cannot accomplish what
+we’d perhaps expect to get out of them: the `assume` statement with a condition of
+"single global variable does not have the same value as the pointer to be
+dereferenced" does not guarantee that the subsequent dereference is safe with
+regard to objects having gone out of scope: the solver can choose to pick
+non-deterministic values such that the address of the object having gone out of
+scope has never been assigned.
```
carolynzech, line 54: I don't get this part either--I think a short example would really help.

docs/src/demonic-nondet.md
```diff
@@ -0,0 +1,80 @@
...
+
+This setup can obviously only track a single object (the one that has gone out
+of scope most recently).
+By adding non-determinism when assigning the address to that single global
+variable (`global-ptr = nondet ? global-ptr : &object-leaving-scope;`) we leave
+the choice of which object to track to the solver.
+This is sound even in presence of multiple objects (and, possibly, multiple
+invalid pointers), because the solver will
+return at least one counterexample if there is any undefined behavior, and can
+indeed be forced to return all such counterexamples.
```
carolynzech, line 80: I think this sentence about soundness would better go in the "What is demonic nondeterminism" section (or you could address soundness twice), but I wouldn't leave it until the last sentence.

carolynzech (commented):
Thank you for writing this up!

docs/src/demonic-nondet.md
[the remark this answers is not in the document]
remi-delmas-3000: > is it an accurate mental model to think of this as a sort of breadth-first search, where we explore all possible assignments to `global_pointer` at every point of the program and succeed iff the property holds in all cases?

The `global_pointer` is indeed a symbolic pointer that ranges over all possible pointers in the program, controlled by the `nondet` choices used to update the variable.

These choices are hidden inputs for the program, so if you have `n` choices, the final formula encodes 2^n executions, one for each valuation of these variables. The breadth-first search analogy is not the best one  though. We don't control in which order the SAT solver explores these decisions.

```C
void foo() {
  // allocate objects
  int *p1 = malloc(...);
  int *p2 = malloc(...);

  // free objects
  free(p1);
  free(p2);

  // use dead pointers
  *p1 = 1;
  *p2 = 2;
}
```

The instrumented program looks like this:

```C
// demonic monitor
static void* dead_ptr = (void *)0;

void foo() {

  // allocate objects
  int *p1 = malloc(...);
  int *p2 = malloc(...);

  // free objects
  // demonic choice variable
  bool b1 = nondet_bool();
  dead_ptr = b1  ? p1 : dead_ptr;
  free(p1);


  // demonic choice variable
  bool b2 = nondet_bool();
  dead_ptr = b2  ? p2 : dead_ptr;
  free(p2);

  // use dead pointers
  assert(!__CPROVER_same_object(p1, dead_ptr));
  *p1 = 1;
  assert(!__CPROVER_same_object(p2, dead_ptr));
  *p2 = 2;
}
```

So our program is controlled by b1, b2 we have 4 possible ways of picking values for these variables:

1. 0, 0 = track null 
2. 0, 1 = track null then switch to p2
3. 1, 0 = track null then switch to p1
4. 1, 1 = track null then switch to p1 then switch to p2

The  SAT solver analyzes all of them, but there is no notion of depth/breadth first.

The idea is that any property of the form `forall x \in {x1, ,,,, xn}, P(x)` can be checked by introducing a nondet variable `x` that ranges over `{x1, ,,,, xn}`, instead of explicitly expanding the forall into `P(x1) and ... and P(xn)`. This makes the encoding more compact and defers the reasoning to SAT/SMT solving time.

docs/src/demonic-nondet.md
remi-delmas-3000: Looking at the example above, to find the safety violation for both p1 and p2 we need to invoke the SAT solver twice, because any choice of b1 and b2 only violates at most one assertion along the same trace

If on the other hand we had this program:

```C
void foo() { 
   // allocate objects 
   int *p1 = malloc(...); 
   int *p2 = malloc(...); 
   
   // free/use
  free(p1);
   *p1 = 1;
   
   // free/use
  free(p2);
   *p2 = 2;
 }
```

```C
void foo() { 
   // allocate objects 
   int *p1 = malloc(...); 
   int *p2 = malloc(...); 
   
   // free/use
   bool b1 = nondet_bool(); 
   dead_ptr = b1 ? p1 : dead_ptr;
  free(p1);
   assert(!__CPROVER_same_object(p1, dead_ptr));
   *p1 = 1;
   
   // free/use
   bool b2 = nondet_bool(); 
   dead_ptr = b2 ? p1 : dead_ptr;
  free(p2);
   assert(!__CPROVER_same_object(p2, dead_ptr));
   *p2 = 1;
 }
 ```
 
 By picking `b1 = 1` and `b2 = 1` we can violate both assertions along a single trace.

File: docs/src/demonic-nondet.md (hunk 1 above)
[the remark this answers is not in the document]
Comment (carolynzech) [reply docs/src/demonic-nondet.md:31]: This makes a ton of sense, thank you @remi-delmas-3000!
(I would love if this text made it into the document--I think a longer example like this is really useful).

File: docs/src/demonic-nondet.md (hunk 2 above)
Comment (carolynzech) [reply docs/src/demonic-nondet.md:47]: Ah okay, so this is more of a statement about not being _guaranteed_ to catch every violation in a single trace--it depends on the structure of your program. If the program looks like the one in your initial example (free, free, use, use) we'll only find one violation. But if the structure is (free, use, free, use) like in this example, we can find both. So we'll find _at least one_ violation of the property if it exists.

File: docs/src/demonic-nondet.md (hunk 3 above)
Comment (remi-delmas-3000) [reply docs/src/demonic-nondet.md:54]: Extending the example above, imagine we encapsulate the condition  `!__CPROVER_same_object(p, dead_ptr)` into a predicate to hide implementation detail and offer an intrisinc to check for dead pointers:
```
extern void *dead_ptr;
bool __pointer_not_dead(void *ptr) {
  return !__CPROVER_same_object(ptr, dead_ptr);
}
```

Due to the demonic behavior of `dead_ptr`, this predicate will only work as expected under an assertion context (where it  implicitly becomes quantified over all nondet choices that define `dead_ptr`).

If you try to use it in an assume:

```C
int *ptr = malloc(sizeof(int));
bool alive = ptr != NULL;
if (alive && nondet_bool()) {
  free(ptr);
  alive = false;
}
// here `ptr` is either alive or dead depending on malloc failure and nondet choice

// assume that `ptr` is not dead, i.e. that we are on a path where `do_free` is false
assume(__pointer_not_dead(ptr));
assume(__pointer_not_dead(ptr)); // passes, unsound
assert(alive); // still falsifiable !
*ptr = 0;
```

This assume does not really force us to be on the path `alive == true` all it does is force the demonic monitor to not track the deallocation of `ptr`, and introduces unsoundness (because once we force the monitor to not look at an object, we miss all violations related to that object).

PR #3897, opened 2025-02-21
Remove isize overflow check for zst offsets

### LINKED ISSUE #3896: `offset` safety check is overly restrictive for ZSTs
I tried this code:

```rust
#[kani::proof]
fn main() {
    let mut x = ();
    let ptr: *mut () = &mut x as *mut ();
    let count: usize = (isize::MAX as usize) + 1;
    let res = unsafe { ptr.add(count) };
}

```

```
cargo kani
```

with Kani version: 0.59

I expected to see this happen: verification succeeds because for ZSTs, the offset can overflow isize.

Instead, this happened: Kani failed with this safety check: https://github.com/model-checking/kani/blob/006e5daa9c48ff5e055de342780dd5172213a979/library/kani_core/src/models.rs#L165

carolynzech:
#3755 introduced additional safety checks for the `offset` intrinsic, including a check for whether `count` overflows `isize`. However, such overflow is allowed for ZSTs. This PR changes the model to check for ZSTs before computing the offset to avoid triggering an overflow failure for ZSTs.

I also moved an existing test called `offset-overflow` into another test, since #3755 changed the failure for that test to be about an out of bounds allocation, not an isize overflow.

Resolves #3896


zhassan-aws (approved):
Thanks

PR #3915, opened 2025-03-03
scanner: Fix loop stats in overall function stats summary

tautschnig:
We already have loop statistics ever since d2141e4c5f59683, they were just not merged into the overall per-function summary (which had an unset `has_loop` column).

### DIFF
--- a/tools/scanner/src/analysis.rs
+++ b/tools/scanner/src/analysis.rs
@@ -44,7 +44,6 @@ impl FnStats {
             is_unsafe: None,
             has_unsafe_ops: None,
             has_unsupported_input: None,
-            // TODO: Implement this.
             has_loop: None,
         }
     }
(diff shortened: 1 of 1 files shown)


tools/scanner/src/analysis.rs
```diff
@@ -191,7 +190,10 @@ impl OverallStats {
                 if !kind.is_fn() {
                     return None;
                 };
-                Some(FnLoops::new(item.name()).collect(&item.body()))
+                let fn_props = FnLoops::new(item.name()).collect(&item.body());
+                self.fn_stats.get_mut(&item).unwrap().has_loop =
+                    Some(fn_props.has_iterators() || fn_props.has_loops());
```
carolynzech, line 195: So `props.has_loop` indicates whether it has an iterator or a loop? Can we change the name of the boolean to reflect that? I would expect the line to read `self.fn_stats.get_mut(&item).unwrap().has_loop = Some(fn_props.has_loops());` if we're talking just about loops, or `self.fn_stats.get_mut(&item).unwrap().has_loop_or_iterator = Some(fn_props.has_iterators() || fn_props.has_loops());`

tautschnig: Good call, renamed to `has_loop_or_iterator`.

carolynzech (approved):

Issue #3832, opened 2025-01-14
Tracking Issue: Make Harnesses Optional


carolynzech:
Requested feature: Make proof harnesses optional, when possible
Use case: There are plenty of "boilerplate harnesses" that look something like this:

```rust
fn foo(x: u8, y: u8) { ... }

#[kani::proof]
fn foo_proof() {
  let x = kani::any();
  let y = kani::any();
  foo(x, y);
}
```
There's nothing surprising in this harness--the user just creates arbitrary variables for each argument of the function they're verifying, and then calls the function. Ideally, the user shouldn't have to write this harness at all; Kani should detect that each of the arguments implements `Arbitrary`, then attempt to verify it automatically.

## Tasks

(We define "standard harness" as a `#[kani::proof]` harness and a "contract harness" as a `#[kani::proof_for_contract]` harness).
I'll break these tasks down into subtasks as I start working, but a road map for now:

- [ ] Leverage the work from https://github.com/model-checking/kani-auto-harness to autogenerate standard harnesses for types that _trivially_ implement `Arbitrary`. (Write the harnesses to a temporary file and run them automatically).
- [ ] Detect which functions have contracts and autogenerate contract harnesses for them.
- [ ] Extend the harness autogeneration to support more complex use cases (c.f [known issues](https://github.com/model-checking/kani-auto-harness/issues))--namely, pointer and loop support.
- [ ] Transition away from writing harnesses to a file, then running them: we shouldn't have to write literal harness Rust code at all. We should just decide which functions we want to verify and verify them, without needing to find some function with a `#[kani::proof]` annotation as an entrypoint for verification.

## Open Questions
  - UI: Support Miri-esque error messages where we say which function caused the problem, rather than which harness, ideally pointing to the problematic line(s) of code.
  - Performance: If we try to verify everything that we can verify, it may take quite a long time. Should we try to be more discerning about what we verify, e.g., introducing some kind of ordering (e.g., look for unsafe functions first, then safe abstractions?) Should this feature have an automatic harness timeout, or perhaps an attribute you can put on a function to opt out of automatic verification?
  - Reasonable defaults: For types that do not trivially implement `Arbitrary` (e.g., pointers), we need to decide:
    - Do reasonable default values exist?
    - If so, what are they, and how do we communicate our default assumptions clearly so that users don't have a false sense of security if their code verifies?

[...]

PR #3922, opened 2025-03-05
Autoharness Misc. Improvements

carolynzech:
Follow up to https://github.com/model-checking/kani/pull/3874 implementing the features requested during review. Hopefully reviewing commit-by-commit makes the size manageable (most of the changes come from the additional metadata logging) but happy to split this into multiple PRs if not.

The list below maps each commit to the review comment on https://github.com/model-checking/kani/pull/3874 that inspired it.

- 86859257546de3ed1da1493ae772b31aa75e2eb5: https://github.com/model-checking/kani/pull/3874#discussion_r1945282348
- f060d66e01973b57e281953c935b89332bb870b9: https://github.com/model-checking/kani/pull/3874#discussion_r1950007068
- 7a7f6548efb707fda468dbfc43772157e3631e9b: https://github.com/model-checking/kani/pull/3874#discussion_r1956657233
- 4f3add81ac0a6e856d1979853eb9b1dd2ddc3b15: https://github.com/model-checking/kani/pull/3874#discussion_r1950071073
- 212b0ffe1e8f3ecab07a3b091357b99eef121414: https://github.com/model-checking/kani/pull/3874#discussion_r1950034157
- f105a9c796cfa88bc81bdda7b5c1127c0a183a0e: https://github.com/model-checking/kani/pull/3874#pullrequestreview-2599556038 and 
- 6a8dc61ef682d93a3db46b29cdf9c0522631741c: https://github.com/model-checking/kani/pull/3874#pullrequestreview-2599556038
- 6f946bb7937f8b2c40b9ce5d4a89b62ac151d019 and 6997afb86730c66216b99ded74d71fdca4916858: https://github.com/model-checking/kani/pull/3874#issuecomment-2640076433


carolynzech:
cc @zhassan-aws and @tautschnig since you reviewed the last PR. (That doesn't mean you need to review this one too, but I thought I'd at least bring it to your attention in case you have opinions about how I addressed your comments).

kani-driver/src/harness_runner.rs
```diff
@@ -269,10 +265,8 @@ impl KaniSession {
             }
         }
 
-        // We currently omit a summary if there was just 1 harness
-        if failing > 0 {
-            println!("Manual Harness Summary:");
-        }
+        println!("Manual Harness Summary:");
```
carolynzech, line 268: Note that this changes the behavior of "normal" Kani, i.e., Kani run without autoharness. The reason is that autoharness runs both manual and automatic harnesses, so printing the message about the number of harnesses we verified is confusing without a header above it specifying that they're _manual_ harnesses. I figured it's not a big deal to just print this line every time.

kani-compiler/src/args.rs
```diff
@@ -91,10 +91,14 @@ pub struct Arguments {
...
     #[clap(long)]
     pub print_llbc: bool,
-    /// If we are running the autoharness subcommand, the functions to autoharness
-    #[arg(long = "autoharness-include-function", num_args(1))]
+    /// If we are running the autoharness subcommand, the functions to include
+    #[arg(
+        long = "autoharness-include-function",
+        num_args(1),
+        conflicts_with = "autoharness_excluded_functions"
+    )]
```
tautschnig, line 99: Why do those have to conflict? I would have assumed the following selection algorithm:
1. Determine the set of all functions.
2. Intersect the set computed in the prior step with `autoharness-include-function`
3. Remove from the set computed in the prior step those matching `autoharness-exclude-function`

kani-compiler/src/args.rs
```diff
@@ -91,10 +91,14 @@ pub struct Arguments {
     /// Print the final LLBC file to stdout.
     #[clap(long)]
     pub print_llbc: bool,
-    /// If we are running the autoharness subcommand, the functions to autoharness
-    #[arg(long = "autoharness-include-function", num_args(1))]
+    /// If we are running the autoharness subcommand, the functions to include
+    #[arg(
+        long = "autoharness-include-function",
+        num_args(1),
+        conflicts_with = "autoharness_excluded_functions"
```
tautschnig, line 98: Does `conflicts_with` require the name of the Rust variable or the name of the parameter (which would be `autoharness-exclude-function`)?

tautschnig (approved):
Love the tables that are being generated!

kani-compiler/src/args.rs
carolynzech: I thought allowing both would be ambiguous. What if I allowed both and implemented this selection algorithm instead?

1. Determine the set of all functions.
2. Remove from the set computed in the prior step those matching autoharness-exclude-function
3. Add back any functions specified in autoharness-include-function

If you allow both, one of them needs to be able to override the other. Your algorithm assumes that exclude takes precedence, while mine assumes that include does. I'd rather just avoid the ambiguity altogether by only allowing one. Rather than specifying a broad include filter and then using exclude to prune it further, I'd prefer to force users to make their include more precise.

kani-compiler/src/args.rs
carolynzech: The variable, see https://github.com/search?q=repo%3Amodel-checking%2Fkani+conflicts_with+%3D+&type=code for examples (e.g. it's `conflicts_with(json_output)` instead of `conflicts_with(json-output)`).

PR #3945, opened 2025-03-18
Support function implementations of known built-ins

tautschnig:
Kani should produce valid symbol tables in GOTO binaries when Rust code provides an implementation for a built-in known to Kani. Without this fix, goto-cc fails the following invariant:
```
Invariant check failed
File: ../src/ansi-c/goto-conversion/goto_convert_functions.cpp:164 function: convert_function
Condition: parameter identifier should not be empty
Reason: !p.empty()
```


carolynzech (approved):
Thanks!

PR #3948, opened 2025-03-18
Autoharness: metadata improvements and enable standard library application

carolynzech:
Update the autoharness subcommand to:

- Print the types of the arguments that don't implement Arbitrary
- Print a table of the functions we chose to verify after codegen finishes, and move the printing of the functions we skipped to be after codegen as well, instead of after verification. This lets us iterate faster to debug because we can pass `--only-codegen` and see pretty-printed data instead of waiting for verification to finish.
- Add a `--std` flag which, when provided, runs autoharness against the provided path to the standard library.

Towards #3832


tautschnig (approved):
Thank you! With your idea of adding in `--std`, would it be possible to similarly add a `--list` argument that will do what the `kani list` command does, after having auto-generated harnesses?

PR #3951, opened 2025-03-20
Add support for loop-contract historic values

### LINKED ISSUE #3697: `loop_old` in loop contracts
Requested feature: `loop_old` similar to `old` that can refers to historic values in loop contracts
Use case:
```
#[kani::loop_invariants(loop_old(x) == y)]
```
where `loop_old(x)` refers to the value of `x` upon the entrance of the loop.
Link to relevant documentation (Rust reference, Nomicon, RFC): #3167 

Test case:
```rust
let mut x: u8 = kani::any_where(|v| *v < 10);
let mut y: u8 = kani::any();
let mut i = 0;
#[kani::loop_invariants(i == 0 || loop_old(x) == y)]
while i < 5{
  if i == 0{
    y = x
  }
  x += 1;
  i += 1;
}
```

thanhnguyen-aws:
Add support for `old` and `prev` to refers to historic values in loop contracts where
- `old(expr)` refers to the value of the expr before the loop.
- `prev(expr)` refers to the value of the expr at the previous iteration.

Resolves #3697


tests/expected/loop-contract/loop_with_previous.expected
```diff
@@ -0,0 +1 @@
+VERIFICATION:- SUCCESSFUL
```
tautschnig, line 1: This file has the wrong name and is thereby causing tests to fail - should be "loop_with_prev.expected"

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -9,7 +9,135 @@ use proc_macro_error2::abort_call_site;
...
+        }
+    }
+
+    fn generate_var_name(&mut self) -> proc_macro2::Ident {
+        let var_name = format_ident!("{}_{}", self.var_prefix, self.counter);
+        self.counter += 1;
+        var_name
+    }
+
+    fn should_replace(&self, expr_path: &syn::ExprPath) -> bool {
```
qinheping, line 43: Could you add some comments to explain when `should_replace` return true?

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -9,7 +9,135 @@ use proc_macro_error2::abort_call_site;
...
+                    let new_var = self.generate_var_name();
+                    self.replacements.push((expr.clone(), new_var.clone()));
+                    *expr = syn::parse_quote!(#new_var);
+                }
+            }
+        }
+    }
+}
+
+fn transform_function_calls(
```
qinheping, line 80: Could you add more comments to the newly added functions?

remi-delmas-3000:
Could we use `on_entry(...)` instead of `old(...)` ? The doc itself says "old(...) denotes the value on loop entry"

tests/expected/loop-contract/loop_with_old.expected
```diff
@@ -1 +1,6 @@
+Check 10: loop_with_old.loop_invariant_base.1
+	 - Status: SUCCESS
+	 - Description: "Check invariant before entry for loop loop_with_old.0"
+	 - Location: src/loop_with_old.rs:16:5 in function loop_with_old
```
carolynzech, line 4: You should remove details that will be brittle between runs (like the check number, or the .1). We also don't typically include location. You should have backslashes between the lines to enforce that the lines are consecutive--with this writing, something like this would pass:

```
loop_with_old.loop_invariant_base.1
	 - Status: FAILURE

random_other_check
     - Status: SUCCESS
```
Without the backslashes, it'll just check for each line in isolation, so there's nothing tying the success to the particular property we care about.

I'd recommend looking at other expected files for examples.

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -9,8 +9,68 @@ use proc_macro_error2::abort_call_site;
...
+Then the loop_invartiant is transformed
+
+*/
+
+/*
+Transform the loop to support prev(expr) : the value of expr at the end of the previous iteration
+If there is a prev(expr) in the loop_invariant:
+1. Firstly, Add an assertion of the loop_quard: the loop must run at least once, otherwise prev(expr) is undefined.
+1. For each prev(expr) in the loop variant, replace it with a newly generated "memory" variable prev_k
+2. Add the declaration of i before the loop: let mut prev_k = expr
```
qinheping, line 36: `of i` or `of prev_k`?

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -9,8 +9,68 @@ use proc_macro_error2::abort_call_site;
...
+For example:
+#[kani::loop_invariant(prev(x+y) = x + y -1 && ...)]
+while(loop_guard)
+{
+    loop_body
+}
+
+is transformed into
+
+assert!(loop_guard);
```
qinheping, line 53: The loop guard can fail for some inputs which means that the loop won't be executed at all.

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -9,8 +9,68 @@ use proc_macro_error2::abort_call_site;
...
+let mut prev_1 = x + y;
+let mut loop_body_closure = || {
+    loop_body_replaced //replace breaks/continues in loop_body with returns
+};
+if loop_body_closure(){
```
qinheping, line 58: When will `loop_body_closure` return false?

tests/expected/loop-contract/loop_with_old_and_prev.rs
```diff
@@ -0,0 +1,22 @@
...
+
+//! Check the use of both "old" and "prev" in loop invariant
+
+#![feature(stmt_expr_attributes)]
+#![feature(proc_macro_hygiene)]
+
+#[kani::proof]
+pub fn loop_with_old_and_prev() {
+    let mut i = 100;
+    #[kani::loop_invariant((i >= 2) && (i <= 100) && (i % 2 == 0) && (on_entry(i) == 100) && (prev(i) == i + 2))]
```
qinheping, line 14: I like the name `prev` and `on_entry`, but they seems to be too common and might be ambiguous with users-defined functions.

qinheping:
I am a bit confused about the semantics of `prev`. How do we check the loop invariant for the first case where there is no `prev` valuation?

library/kani_macros/src/sysroot/loop_contracts/mod.rs
thanhnguyen-aws: We only assert the loop_guard when there is "prev" in the loop_invariant (see line 320 to 360).  prev(expr) is only defined when the loop can execute at least once, so I assume that when prev is in the loop_invariant, the invariant check will fail if the loop_guard is not satisfied at the beginning.

library/kani_macros/src/sysroot/loop_contracts/mod.rs
thanhnguyen-aws: It returns false if we hit a break statement in the original loop body (which is turned into "return false" in the loop_body_closure), see line 37. When it returns false, the loop only has 1 iteration, so we assert the invariant and skip the loop.

library/kani_macros/src/sysroot/loop_contracts/mod.rs
thanhnguyen-aws: I edited.

tests/expected/loop-contract/loop_with_old_and_prev.rs
thanhnguyen-aws: Any suggestions?

qinheping (approved):
LGTM. Thank you!

PR #3971, opened 2025-04-01
Fix `autoharness` termination test & print metadata in alphabetical order

carolynzech:
https://github.com/model-checking/kani/pull/3968 [failed in the merge queue](https://github.com/model-checking/kani/actions/runs/14180398658/job/39724995535) because the `cargo_autoharness_termination` test assumed that the recursion test wouldn't time out, but it did. This doesn't happen often, since this test has worked up until now, but in the interest of having robust tests, this PR splits the termination test into two:

- `autoharness_timeout`, which tests that the default timeout of 60s kicks in
- `autoharness_unwind`, which tests that the default unwind bound of 20 kicks in

The unwind test specifies a timeout of 5m to ensure that the unwind bound takes effect before the timeout.

While I was fixing this, I also fixed some inconsistencies with the printed metadata:
- We use "Chosen Function" in one table and "Selected Function" in another--changed it to "Selected Function" in both to be consistent
- The functions aren't printed alphabetically in 2/3 tables--fixed so that they're always sorted


tautschnig:
In some previous PR I had already run into the issue of timeouts occurring before the unwinding limit was reached, at which point I increases the default timeout to 60 seconds instead of 30 seconds. Can we instead adjust the timeout for just this test (via command-line options) to 5 minutes?

carolynzech:
> In some previous PR I had already run into the issue of timeouts occurring before the unwinding limit was reached, at which point I increases the default timeout to 60 seconds instead of 30 seconds. Can we instead adjust the timeout for just this test (via command-line options) to 5 minutes?

I thought about doing this, but I thought that was a long time to let a single test run. I suppose it's okay to make one exception though.

carolynzech:
I've split the termination test into two tests, one to test the unwind bound and one for the timeout. The unwind test sets the timeout to 5m, but at least testing locally it never gets close to that amount, so I think we're good.

tautschnig (approved):

PR #3972, opened 2025-04-01
Use Kani's Cargo when calling `cargo metadata`

### LINKED ISSUE #3967: Kani doesn't work with Rust versions below 1.77
```sh
cargo +1.76 new demo
cd demo
cargo +1.76 kani
```

This results in the error
```
error: expected a version like "1.32"
error: Failed to execute cargo (exit status: 101). Found 0 compilation errors.
```

#3593 uses the package ID from `cargo metadata` as the argument to `-p`, but before the [package ID spec](https://doc.rust-lang.org/nightly/cargo/CHANGELOG.html#added-10) was stabilized in 1.77, the package ID couldn't be used in this way.

Feel free to close if supporting older Rust versions is not desired

clubby789:
When Kani is installed normally, use the Cargo binary from the linked Rust toolchain when calling `cargo metadata`.
If a `rust-toolchain.toml` is present, or the toolchain has been manually overridden with `cargo +1.xx kani` an older version of `cargo metadata` can be used, which causes incompatible package IDs to be passed.

Resolves #3967

Tested by installing and verifying that `cargo +1.76 kani` no longer produces an error.

### DIFF
--- a/kani-driver/src/call_cargo.rs
+++ b/kani-driver/src/call_cargo.rs
@@ -4,7 +4,9 @@
 use crate::args::VerificationArgs;
 use crate::call_single_file::{LibConfig, to_rustc_arg};
 use crate::project::Artifact;
-use crate::session::{KaniSession, lib_folder, lib_no_core_folder, setup_cargo_command};
+use crate::session::{
+    KaniSession, bundled_cargo_path, lib_folder, lib_no_core_folder, setup_cargo_command,
+};
 use crate::util;
 use anyhow::{Context, Result, bail};
 use cargo_metadata::diagnostic::{Diagnostic, DiagnosticLevel};
@@ -241,6 +243,9 @@ crate-type = ["lib"]
 
     pub fn cargo_metadata(&self, build_target: &str) -> Result<Metadata> {
         let mut cmd = MetadataCommand::new();
+        if let Some(cargo_path) = bundled_cargo_path()? {
+            cmd.cargo_path(cargo_path);
+        }
 
         // restrict metadata command to host platform. References:
         // https://github.com/rust-lang/rust-analyzer/issues/6908
--- a/kani-driver/src/session.rs
+++ b/kani-driver/src/session.rs
@@ -442,3 +442,11 @@ pub fn setup_cargo_command() -> Result<Command> {
 
     Ok(cmd)
 }
+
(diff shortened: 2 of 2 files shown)


zjp-CN:
Since kani relies on pinned toolchain for analyzing Rust code through rustc, is it a good practice to run `cargo +1.76 kani` when you're using incompatible toolchain? 
Why not just run `cargo +nightly-xxxx-xx-xx kani` where `nightly-xxxx-xx-xx` is required by kani?

clubby789:
This was motivated by running Kani in an existing repo with an old `rust-toolchain.toml` and getting a very obscure error that took several hours to track down - `+1.76` is just for ease of testing.
I also experimented with trying to translate old-format IDs to new-format, but given that the format is unspecified I'm not confident in being able to write a reliable solution there

zjp-CN:
I think `cargo +nightly-xxxx-xx-xx kani` is more reliable, because
* `+nightly-xxxx-xx-xx` is prior to rust-toolchain.toml by rustup
* you'll need `+nightly-xxxx-xx-xx` anyways to make kani really work to do analysis

clubby789:
Unless I'm mistaken, Kani always uses its linked nightly-xxxx toolchain (unless built in a dev environment) - this fixes the inconsistency here, which was a bit of a beginner papercut when trying to use the tool initially

carolynzech:
Thanks for the PR! This `cargo +1.76 kani` invocation pattern is not something we'd like to support--for a given version of Kani, we only support it working with its `rust-toolchain.toml` version. I worry that this PR will lead users into thinking that Kani is accepting their 1.76 override, which IMO is more confusing than just failing loudly when they try to override the toolchain. One thing we could do is try to detect if someone is overriding the toolchain version and error with a message about that, since I agree that the current `cargo metadata` related error message is quite obscure.

I'm curious about this comment:

> This was motivated by running Kani in an existing repo with an old rust-toolchain.toml and getting a very obscure error that took several hours to track down - +1.76 is just for ease of testing.

Can you post a follow up in #3967 elaborating more on what happened here? You should be able to invoke Kani in a repo that has a different `rust-toolchain.toml` file. If there are papercuts with using the tool, we would very much like to know about them (and fix them).

I'm going to close this PR, but please do follow up in #3967 if you have the bandwidth--if something took hours of debugging, we would like to improve on that.

PR #3974, opened 2025-04-01
Clarify `is_inbounds` docs

carolynzech:
See https://github.com/model-checking/kani/pull/3956#discussion_r2019734480

### DIFF
--- a/library/kani_core/src/mem.rs
+++ b/library/kani_core/src/mem.rs
@@ -169,13 +169,12 @@ macro_rules! kani_mem {
         /// Checks that `ptr` points to an allocation that can hold data of size calculated from `T`.
         ///
         /// This function always returns `true` for ZSTs, since every pointer to a ZST is valid.
-        /// For non-ZSTs, this function will:
-        ///  - Panic if `ptr` does not point to allocated memory,
-        ///  - Return `false` if the size of the val pointed to exceeds `isize::MAX`,
-        ///  - Return `false` if the pointer is null.
-        ///
-        /// If none of the above conditions hold, it will return `true` if and only if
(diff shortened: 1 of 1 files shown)


library/kani_core/src/mem.rs
```diff
@@ -168,15 +168,14 @@ macro_rules! kani_mem {
...
-        /// - The computed size overflows.
-        /// - The computed size exceeds `isize::MAX`.
-        /// - The pointer is null (except for zero-sized types).
-        /// - The pointer references unallocated memory.
-        ///
-        /// This function aligns with Rust's memory safety requirements, which restrict valid allocations
-        /// to sizes no larger than `isize::MAX`.
+        /// This function always returns `true` for ZSTs, since every pointer to a ZST is valid.
+        /// For non-ZSTs, this function will:
+        ///  - Panic if `ptr` does not point to allocated memory,
```
carolynzech, line 173:
```suggestion
        ///  - Fails verification if `ptr` does not point to allocated memory,
```

Not sure what the better wording here is--it gets codegened to an `UnsupportedHook`, which gets codegened as an `assert(false)` so that verification fails. This isn't *really* the same thing as a panic, but I'm not sure if we want to be talking about verification in the doc comments for this, since it's somewhat orthogonal to the question of what the function returns.

zhassan-aws: I would spell that out:
```rust
        /// For non-ZSTs, this function will panic if `ptr` does not point to allocated memory
        /// Currently, Kani does not support reasoning about pointers to unallocated memory,
        /// so if `ptr` does not point to allocated memory, verification will fail
```
or something along those lines.

zhassan-aws (approved):

PR #3979, opened 2025-04-02
Fix the bug: Loop contracts are not composable with function contracts 

### LINKED ISSUE #3910: Loop contracts are not composable with function contracts
I tried this code:

```rust

#[kani::requires(true)]
fn simple_loop_with_loop_contracts() {
    let mut x: u64 = kani::any_where(|i| *i >= 1);

    #[kani::loop_invariant(x >= 1)]
    while x > 1 {
        x = x - 1;
    }

    assert!(x == 1);
}


#[kani::proof_for_contract(simple_loop_with_loop_contracts)]
fn foo() {
    simple_loop_with_loop_contracts()
}
```

```
cargo kani -Z function-contracts -Z loop contracts
```

with Kani version: 0.59

I expected to see this happen: verification succeeds

Instead, this happened: the loop invariant is ignored, so the loop unwinds forever.

I find it confusing that we call it a "loop contract" and yet we can't use a "proof for contract" for it.

The same unwinding behavior happens if I change the harness to be a `#[kani::proof]`--it appears we can't prove invariants at all in the presence of other contracts.

thanhnguyen-aws:
This PR fixes the bug of loop contracts are not composable with function contracts.
Previously, using loop-contract inside a function with contract may result in unwinding the loop instead of contracting it.

Resolves #3910


carolynzech (changes requested):
Thanks for fixing this! Can you also update our loop contracts documentation in our book to talk about this change, and more generally, explain when to use `#[kani::proof]` vs. `#[kani::proof_for_contract]`? If I just have a loop contract without a function contract, can I use `#[kani::proof_for_contract]`?

(I haven't reviewed the code, I'm leaving that to @qinheping).

thanhnguyen-aws:
> Thanks for fixing this! Can you also update our loop contracts documentation in our book to talk about this change, and more generally, explain when to use `#[kani::proof]` vs. `#[kani::proof_for_contract]`? If I just have a loop contract without a function contract, can I use `#[kani::proof_for_contract]`?
> 
> (I haven't reviewed the code, I'm leaving that to @qinheping).

Using `#[kani::proof]` or `#[kani::proof_for_contract]` is irrelevant to the loop or its contract, so we just use it as normal. I added the tests just to make sure there is no bug in both cases.

carolynzech:
> Using #[kani::proof] or #[kani::proof_for_contract] is irrelevant to the loop or its contract, so we just use it as normal. I added the tests just to make sure there is no bug in both cases.

Thanks for the test, but please also update the book documentation so that our users know how to use this feature. I find it confusing that the annotation is irrelevant--why would we have both be acceptable if they do the same thing? Can you do some more thinking about the UX here before we merge this? We want to make sure that our users are set up to use this correctly.

thanhnguyen-aws:
> > Using #[kani::proof] or #[kani::proof_for_contract] is irrelevant to the loop or its contract, so we just use it as normal. I added the tests just to make sure there is no bug in both cases.
> 
> Thanks for the test, but please also update the book documentation so that our users know how to use this feature. I find it confusing that the annotation is irrelevant--why would we have both be acceptable if they do the same thing? Can you do some more thinking about the UX here before we merge this? We want to make sure that our users are set up to use this correctly.

Let me explain it more clearly. This PR does not add any new feature, it just fixing the bug that when a function has contract, then the loop inside, the loop will be unwinded instead of contracted even if there is a loop invariant and kani flag -Z loop-contracts is enabled. There is no change in the definitions of #[kani::proof] or #[kani::proof_for_contract] whether there are loops/loop-invariants inside the function or not, and the two harnesses are different. As you can see the verification results (in two expected files) are different, #[kani::proof_for_contract] succeed because it assumes the requires clause, while #[kani::proof] failed because it asserts (not assumes) the requires clause. Therefore, I think that the users need no additional knowledge in this case.

carolynzech:
> As you can see the verification results (in two expected files) are different, #[kani::proof_for_contract] succeed because it assumes the requires clause, while #[kani::proof] failed because it asserts (not assumes) the requires clause. 

That makes sense.

> There is no change in the definitions of #[kani::proof] or #[kani::proof_for_contract] whether there are loops/loop-invariants inside the function or not, and the two harnesses are different. 

I'm not sure that's true. This PR changes the behavior of Kani such that you can use `#[kani::proof_for_contract]` on code with loop invariants. This would change the meaning of `#[kani::proof_for_contract]`--can we use it to prove loop invariants now, whereas before we could just use `#[kani::proof]`? Can I use `#[kani::proof_for_contract]` if we *just* have a loop contract, and not function contracts? If not, then it's confusing that we call it "loop contract" and we can't use a `#[kani::proof_for_contract]` to prove it, and we should make that clear in our documentation.

> Therefore, I think that the users need no additional knowledge in this case.

I think our documentation does need some kind of example about composing loop contracts with function contracts, since there's nothing in our book to suggest that this is even possible (all of our loop contract examples use `#[kani::proof]`). 

What I really want is documentation that tells users which kind of harness they should use for each case:
- If my function just has a loop contract, should I use `#[kani::proof]` or `#[kani::proof_for_contract]`?
- If my function has both a loop contract and a function contract, should I use `#[kani::proof]` or `#[kani::proof_for_contract]`?
- How do the answers to the above questions change if I pass `--no-assert-contracts`?

thanhnguyen-aws:
> This PR changes the behavior of Kani such that you can use `#[kani::proof_for_contract]` on code with loop invariants. This would change the meaning of `#[kani::proof_for_contract]`--can we use it to prove loop invariants now, whereas before we could just use `#[kani::proof]`? Can I use `#[kani::proof_for_contract]` if we _just_ have a loop contract, and not function contracts? 

This PR does not change the behavior of Kani for `#[kani::proof_for_contract(funcname)]`: even with loop invariants, Kani will throw an error if the function does not have contract. I think what is confusing here is the name `#[kani::proof_for_contract (function name)]`: contract here means function contract, not loop contract. We do not have a harness attribute to enable checking loop-contract, the only way is using -Z loop-contract flag. 

> If not, then it's confusing that we call it "loop contract" and we can't use a `#[kani::proof_for_contract]` to prove it, and we should make that clear in our documentation.

I strongly agree that the easiest way to get rid of the confusion is removing the name "loop contract" and consistently calling it "loop-invariant". If everyone else agree, we can create a separate PR  to change the flag name and all related tests.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -103,40 +103,13 @@ impl TransformPass for LoopContractPass {
...
-                        }
-                    }
-                    (contain_loop_contracts, new_body.into())
+                    self.transform_body_with_loop(tcx, body)
                 }
             }
+            RigidTy::Closure(_, _) => self.transform_body_with_loop(tcx, body),
             _ => {
+                println!("Instance defid: {:?}", instance.def.name());
+                println!("Instance ty: {:?}", instance.ty().kind().rigid().unwrap());
```
carolynzech, line 112:
```suggestion
```

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -310,7 +318,7 @@ impl LoopContractPass {
                 for stmt in &new_body.blocks()[bb_idx].statements {
                     if let StatementKind::Assign(place, rvalue) = &stmt.kind {
                         match rvalue {
-                            Rvalue::Ref(_,_,rplace) => {
+                            Rvalue::Ref(_,_,rplace) | Rvalue::CopyForDeref(rplace) => {
```
carolynzech, line 321: Why this change?

tests/expected/loop-contract/function_call_with_loop.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// kani-flags: -Z loop-contracts -Z function-contracts
+
+//! Calling a function that contains loops
```
carolynzech, line 6: Please update this text to explain what the expected output of this test is and why. (Something along the lines of: "Test that using a `#[kani::proof]` harness fails because the function's precondition gets asserted.")

tests/expected/loop-contract/contract_proof_function_with_loop.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// kani-flags: -Z loop-contracts -Z function-contracts
+
+//! Proving contract for function with loop
```
carolynzech, line 6:
```suggestion
//! Test that we can prove a function contract and a loop contract in tandem.
```

carolynzech (commented):
Can you also add this test?

```rust
// -Z loop-contracts -Z function-contracts --no-assert-contracts

#[kani::requires(i>=2)]
#[kani::ensures(|ret| *ret == 2)]
pub fn has_loop(mut i: u16) -> u16 {
    #[kani::loop_invariant(i>=2)]
    while i > 2 {
        i = i - 1
    }
    i
}

#[kani::proof]
fn contract_proof() {
    let i: u16 = kani::any();
    let j = has_loop(i);
}
```

This proof should let you just prove the loop invariant while ignoring the contract.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
thanhnguyen-aws: The function contract attributes create some closures whose body are the same as that of the function, those places that are assigned Rvalue::CopyForDeref(rplace) are used in those closures to replace the user variables (rplace) in user code, so they need to be added in to the list of supported vars.

thanhnguyen-aws:
> Can you also add this test?
> 
> ```rust
> // -Z loop-contracts -Z function-contracts --no-assert-contracts
> 
> #![feature(proc_macro_hygiene)]
> #![feature(stmt_expr_attributes)]
> 
> #[kani::requires(i>=2)]
> #[kani::ensures(|ret| *ret == 2)]
> pub fn has_loop(mut i: u16) -> u16 {
>     #[kani::loop_invariant(i>=2)]
>     while i > 2 {
>         i = i - 1
>     }
>     i
> }
> 
> #[kani::proof]
> fn contract_proof() {
>     let i: u16 = kani::any();
>     let j = has_loop(i);
> }
> ```
> 
> This proof should let you just prove the loop invariant while ignoring the contract.

I added it

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -194,6 +165,41 @@ impl LoopContractPass {
         ))
     }
 
+    fn transform_body_with_loop(&mut self, tcx: TyCtxt, body: Body) -> (bool, Body) {
```
qinheping, line 168: Could you add some comments explaining what does this transformation do?

qinheping (approved):
LGTM, thank you!

carolynzech (approved):
Thanks!

PR #3984, opened 2025-04-03
Fix cargo invocations to only use `pkg_args` where appropriate

carolynzech:
## Summary

When we invoke `cargo rustc` in `kani-driver`, we have some options in a `pkg_args` variable, and some in `kani_compiler_flags`. This PR removes a couple of uses of `pkg_args` that should have really been `kani_compiler_flags`, and provides documentation about the difference.

## Explanation
Hopefully the documentation in the code is sufficient to understand the difference (please suggest changes if it's not!), but here's a longer explanation:

`cargo kani` invokes [cargo rustc](https://doc.rust-lang.org/cargo/commands/cargo-rustc.html), described as follows:

> cargo rustc [options] [-- args]
> The specified target for the current package (or package specified by -p if provided) will be compiled along with all of its  dependencies. The specified args will all be passed to the final compiler invocation, not any of the dependencies.

Our `pkg_args` variable is what we provide for `-- args`, i.e., the arguments that we want to provide to kani-compiler when it compiles the package under verification, but *not its dependencies.* 

The docs then say:

> To pass flags to all compiler processes spawned by Cargo, use the RUSTFLAGS [environment variable](https://doc.rust-lang.org/cargo/reference/environment-variables.html)

We use the `RUSTFLAGS` environment variable to provide the `kani_compiler_flags` that should be passed when we invoke `kani-compiler` on the package to verify *and its dependencies.*

So we should use `kani_compiler_flags` when the dependencies of the target package should receive the flag, and `pkg_args` when it shouldn't. I concluded that the only argument that it makes sense to provide in `pkg_args` is `--reachability`, because when `--reachability` isn't provided it defaults to `None`, which is the behavior we want. (Otherwise, we'd run Kani harnesses that we find in dependencies, or if autoharness is running, generate automatic harnesses for functions in dependencies, neither of which we want). Dependencies can get all of the other compiler arguments, since they don't do anything with them when `--reachability=None` anyway.

## Commit by Commit

- `--no-assert-contracts` is already provided in `kani_compiler_flags`, and never should have been in `pkg_args` in the first place.
- Having `--backend=llbc` as a `pkg_arg` means that it doesn't get provided to dependencies, so that when we run Kani's compiler on dependencies, we'd actually enter the cprover compiler interface. Move it to compiler args so that it gets passed to the target crate and its dependencies.
- Autoharness should also use `kani_compiler_flags`, but that involves a larger refactor than I want to do this close to a release, so added a TODO for now.
- Add documentation explaining the difference between `pkg_args` and `kani_compiler_flags`.

### DIFF
--- a/kani-driver/src/call_cargo.rs
+++ b/kani-driver/src/call_cargo.rs
@@ -182,7 +182,18 @@ crate-type = ["lib"]
             cargo_args.push("-v".into());
         }
 
-        // Arguments that will only be passed to the target package.
+        // Arguments that will only be passed to the target crate (the crate under verification)
+        // and not its dependencies, c.f. https://doc.rust-lang.org/cargo/commands/cargo-rustc.html.
+        // The difference between pkg_args and rustc_args is that rustc_args are also provided when
+        // we invoke rustc on the target crate's dependencies.
+        // We do not provide the `--reachability` argument to dependencies so that it has the default value `None`
+        // (c.f. kani-compiler::args::ReachabilityType) and we skip codegen for the dependency.
+        // This is the desired behavior because we only want to construct `CodegenUnits` for the target crate;
+        // i.e., if some dependency has harnesses, we don't want to run them.
+
+        // If you are adding a new `kani-compiler` argument, you likely want to put it `kani_compiler_flags()` instead,
+        // unless there a reason it shouldn't be passed to dependencies.
+        // (Note that at the time of writing, passing the other compiler args to dependencies is a no-op, since `--reachability=None` skips codegen anyway.)
         self.pkg_args.push(self.reachability_arg());
 
         let mut found_target = false;
--- a/kani-driver/src/call_cargo.rs
+++ b/kani-driver/src/call_cargo.rs
@@ -187,9 +187,6 @@ crate-type = ["lib"]
         if let Some(backend_arg) = self.backend_arg() {
             self.pkg_args.push(backend_arg);
         }
-        if self.args.no_assert_contracts {
-            self.pkg_args.push("--no-assert-contracts".into());
-        }
 
         let mut found_target = false;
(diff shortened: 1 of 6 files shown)


zhassan-aws (approved):
Thanks for sorting this out @carolynzech!

kani-driver/src/call_cargo.rs
```diff
@@ -182,14 +182,19 @@ crate-type = ["lib"]
...
 
-        // Arguments that will only be passed to the target package.
+        // Arguments that will only be passed to the target crate (the crate under verification)
+        // and not its dependencies, c.f. https://doc.rust-lang.org/cargo/commands/cargo-rustc.html.
+        // The difference between pkg_args and rustc_args is that rustc_args are also provided when
+        // we invoke rustc on the target crate's dependencies.
+        // We do not provide the `--reachability` argument to dependencies so that it has the default value `None`
+        // (c.f. kani-compiler::args::ReachabilityType) and we skip codegen for the dependency.
+        // This is the desired behavior because we only want to construct `CodegenUnits` for the target crate;
+        // i.e., if some dependency has harnesses, we don't want to run them.
```
carolynzech, line 192:
```suggestion
        // Arguments that will only be passed to the target package (the package under verification)
        // and not its dependencies, c.f. https://doc.rust-lang.org/cargo/commands/cargo-rustc.html.
        // The difference between pkg_args and rustc_args is that rustc_args are also provided when
        // we invoke rustc on the target package's dependencies.
        // We do not provide the `--reachability` argument to dependencies so that it has the default value `None`
        // (c.f. kani-compiler::args::ReachabilityType) and we skip codegen for the dependency.
        // This is the desired behavior because we only want to construct `CodegenUnits` for the target package;
        // i.e., if some dependency has harnesses, we don't want to run them.
```

PR #4000, opened 2025-04-07
Introduce BoundedArbitrary trait and macro for bounded proofs

sgpthomas:
This PR introduces the trait `BoundedArbitrary` which allows for creation of symbolic values for types up to some bound. This is useful for doing bounded proofs on types like strings and vecs. I wrote a more detailed overview of what this PR provides as a new section in the book.

The implementation of BoundedArbitrary for the container types that I provide are pretty basic and experimental at the moment. Any pointers here would be useful, and we will probably need to iterate on their implementations based on actual performance testing.

I added a test case that should fail verification using the `cover!` macro, however, I'm not sure that I used this correctly. Let me know if I should make this an expect test, or something else.

I slightly restructured how we deal with the the `std/core` usage in `kani_core::kani_lib`. Previously we passed in either `std` or `core` to `kani_core::generate_arbitrary` and `generate_arbitrary` renamed the import with `use $core as core_path`. However, I couldn't use that same strategy for `kani_core::generate_bounded_arbitrary` because the import names would clash. So I lifted the `use $core as core_path` out of the `kani_core::generate_*` macros.


tests/kani/BoundedArbitrary/vec.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: --default-unwind 17
+//
+// Check that users can implement Arbitrary to a simple data struct with Vec<>.
```
zhassan-aws, line 5: Test description is not accurate

tests/kani/BoundedArbitrary/vec.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: --default-unwind 17
```
zhassan-aws, line 3: You can annotate individual harnesses with `#[kani::unwind(N)]`. For example, `check_reverse_is_its_own_inverse` only needs an unwinding of 5?

tests/kani/BoundedArbitrary/derive.rs
```diff
@@ -0,0 +1,44 @@
...
+        #[bounded]
+        x: Vec<T>,
+        y: bool,
+    },
+}
+
+#[kani::proof]
+fn check_my_vec() {
+    let my_vec: BoundedAny<MyVector<bool>, 4> = kani::any();
+    assert!(my_vec.vector.len() <= 4)
```
zhassan-aws, line 33: A stronger property is to ensure all vector lengths are possible, which you can do using:
```rust
kani::cover!(my_vec.vector.len() == 0);
kani::cover!(my_vec.vector.len() == 1);
kani::cover!(my_vec.vector.len() == 2);
kani::cover!(my_vec.vector.len() == 3);
kani::cover!(my_vec.vector.len() == 4);
```

tests/kani/BoundedArbitrary/derive.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright Kani Contributors
```
zhassan-aws, line 1: As you mentioned in the PR description, it's better to turn these tests into expected tests to ensure Kani's output includes the expected results. See https://github.com/model-checking/kani/blob/main/tests/expected/cover/cover-unreachable/expected for an example.

zhassan-aws (commented):
A few comments on the tests.

zhassan-aws:
This is a link to the rendered docs for the reference: https://github.com/sgpthomas/kani/blob/bounded_arbitrary/docs/src/reference/experimental/bounded_arbitrary.md

docs/src/reference/experimental/bounded_arbitrary.md
```diff
@@ -0,0 +1,80 @@
...
+}
+```
+
+We can use `BoundedAny` to write a proof harness:
+
+```rust
+#[kani::proof]
+fn check_reverse_is_its_own_inverse() {
+    // We use BoundedAny to construct a vector that has at most length 16
+    let input: BoundedAny<Vec<bool>, 16> = kani::any();
```
zhassan-aws, line 25: Is the `BoundedAny` wrapper necessary? A more intuitive API would be:
```rust
let input: Vec<bool> = Vec<bool>::bounded_any::<4>();
```
or something along those lines.

docs/src/reference/experimental/bounded_arbitrary.md
```diff
@@ -0,0 +1,80 @@
...
+
+    let double_reversed = reverse_vector(reverse_vector(input.clone().into_inner()));
+
+    // we assert that every value in the input is the same as the value in the
+    // doubly reversed list
+    for i in 0..input.len() {
+        assert!(input[i] == double_reversed[i])
```
zhassan-aws, line 32: Nit: use `assert_eq!`

docs/src/reference/experimental/bounded_arbitrary.md
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+```
+
+Then, with `kani --harness check --unwind 17` we can prove that, for vectors up to size 16, our reverse function is indeed it's own inverse.
```
zhassan-aws, line 37: You can add `#[kani::unwind(17)]` to avoid needing the command line. You can also omit the `--harness`. Probably don't need to mention the command at all.

zhassan-aws, line 37:
```suggestion
Then, with `kani --harness check --unwind 17` we can prove that, for vectors up to size 16, our reverse function is indeed its own inverse.
```

docs/src/reference/experimental/bounded_arbitrary.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Proof Incompleteness
+
+It's very important to note, that this is not a complete proof that this function is correct. To drive this point home, consider this bad implementation of `reverse_vector`:
```
zhassan-aws, line 41:
```suggestion
It's very important to note, that this is **not** a complete proof that this function is correct. To drive this point home, consider this bad implementation of `reverse_vector`:
```

docs/src/reference/experimental/bounded_arbitrary.md
```diff
@@ -0,0 +1,80 @@
...
+    // doubly reversed list
+    for i in 0..input.len() {
+        assert!(input[i] == double_reversed[i])
+    }
+}
+```
+
+Then, with `kani --harness check --unwind 17` we can prove that, for vectors up to size 16, our reverse function is indeed it's own inverse.
+
+## Proof Incompleteness
```
zhassan-aws, line 39: Useful section!

docs/src/reference/experimental/bounded_arbitrary.md
```diff
@@ -0,0 +1,80 @@
...
+    vector: Vec<T>,
+    capacity: usize
+}
+```
+
+You must specify which fields should be bounded using the `#[bounded]` attribute. All other fields must derive `Arbitrary`.
+
+### Limitations
+
+Currently you can only specify a single bound for the entire type, and all bounded fields use the same bound. If different bounds would be useful, let us know and we can probably lift this restriction.
```
zhassan-aws, line 80:
```suggestion
Currently you can only specify a single bound for the entire type, and all bounded fields use the same bound. If different bounds would be useful, let us know through [filing an issue](https://github.com/model-checking/kani/issues/new/choose) and we can probably lift this restriction.
```

zhassan-aws (commented):
Some comments on the documentation

docs/src/reference/experimental/bounded_arbitrary.md
sgpthomas: That's also supported, I'll change the docs to use that.

carolynzech (commented):
Can you write tests for `Option`, `Result`, `HashMap`, etc. as well?

library/kani_macros/src/derive_bounded.rs
```diff
@@ -0,0 +1,155 @@
...
+/// example:
+///
+/// ```rust
+/// #[derive(BoundedArbitrary)]
+/// struct MyVec<T> {
+///     #[bounded]
+///     vec: Vec<T>,
+///     cap: usize
+/// }
+/// ```
```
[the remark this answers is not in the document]
sgpthomas: The derive adds a bound for `T: Arbitrary`. This is pretty standard for derive macros. It's maybe non-intuitive that it's `T: Arbitrary` instead of `T: BoundedArbitrary`, but the idea is that we are bounding the size of the container, not the elements in the container. Does that reasoning seem right?

zhassan-aws (commented):
Another round of comments

library/kani/src/bounded_arbitrary.rs
```diff
@@ -0,0 +1,66 @@
...
+        bytes.utf8_chunks().for_each(|chunk| string.push_str(chunk.valid()));
+
+        kani::assume(string.len() <= N);
+        string
+    }
+}
+
+impl BoundedArbitrary for std::ffi::OsString {
+    fn bounded_any<const N: usize>() -> Self {
+        let bounded_string = String::bounded_any::<N>();
```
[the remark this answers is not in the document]
sgpthomas: Hm OsStrings are platform specific. We might need an implementation per platform, so I wonder if it's best to just elide this implementation for now

library/kani/src/bounded_arbitrary.rs
```diff
@@ -0,0 +1,66 @@
...
+        vec
+    }
+}
+
+impl BoundedArbitrary for String {
+    fn bounded_any<const N: usize>() -> Self {
+        let bytes: [u8; N] = kani::any();
+
+        let mut string = String::new();
+        bytes.utf8_chunks().for_each(|chunk| string.push_str(chunk.valid()));
```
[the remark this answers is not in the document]
sgpthomas: wouldn't this just return a string that is a single char?

library/kani/src/bounded_arbitrary.rs
```diff
@@ -0,0 +1,66 @@
...
+impl<K, V> BoundedArbitrary
+    for std::collections::HashMap<K, V, std::hash::BuildHasherDefault<std::hash::DefaultHasher>>
+where
+    K: Arbitrary + std::cmp::Eq + std::hash::Hash,
+    V: Arbitrary,
+{
+    fn bounded_any<const N: usize>() -> Self {
+        let mut hash_map = std::collections::HashMap::default();
+        for _ in 0..N {
+            hash_map.insert(K::any(), V::any());
```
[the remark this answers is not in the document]
sgpthomas:
```
fn check_hashmap() {
    // A larger bound causes this to take a long time
    const BOUND: usize = 1;
    let hash_map: std::collections::HashMap<u8, bool, _> = kani::bounded_any::<_, BOUND>();
    for i in 0..=BOUND {
        kani::cover!(hash_map.len() == i);
    }
}
```
This check passes. I see how we can get hashmaps smaller than N because keys can overlap. However, I'm not sure why the check for length 0 passes. that seems wrong. any idea why?

library/kani/src/bounded_arbitrary.rs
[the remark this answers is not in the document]
zhassan-aws: No, `utf8_chunks` returns an iterator over valid UTF8 ranges. Since it's possible that the entire byte array is a valid UTF8 encoding, the very first chunk could span the entire array.

Here's an example:
```rust
#[kani::proof]
#[kani::unwind(5)]
fn main() {
    let bytes: [u8; 4] = kani::any();

    let s: String = if let Some(s) = bytes.utf8_chunks().next() {
        s.valid().into()
    } else {
        String::new()
    }; 

    kani::cover!(s == String::from(""));
    kani::cover!(s == String::from("a"));
    kani::cover!(s == String::from("ab"));
    kani::cover!(s == String::from("abc"));
    kani::cover!(s == String::from("abcd"));
}
```
Kani covers all properties:
```
SUMMARY:
 ** 0 of 340 failed (5 unreachable)

 ** 5 of 5 cover properties satisfied


VERIFICATION:- SUCCESSFUL
Verification Time: 6.4006224s
```

library/kani/src/bounded_arbitrary.rs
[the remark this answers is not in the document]
zhassan-aws: I agree it's better to drop it for now

library/kani/src/bounded_arbitrary.rs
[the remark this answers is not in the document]
zhassan-aws: Good point. Are you sure it passed? Note that a failed cover property does not cause verification to fail (it'll just be reported as uncovered).

library/kani/src/bounded_arbitrary.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl<T: Arbitrary> BoundedArbitrary for Vec<T> {
+    fn bounded_any<const N: usize>() -> Self {
+        let real_length = kani::any_where(|&size| size <= N);
+        let array: [T; N] = kani::any();
+        let mut vec = Vec::from(array);
+        vec.truncate(real_length);
+        vec
+    }
+}
```
carolynzech, line 16: Wait, what is the difference between this and `any_vec`? https://github.com/model-checking/kani/blob/59b25fd511363f19286a058079b74633f4e4c90e/library/kani/src/vec.rs#L6

I haven't thought too deeply about this, so it's possible there's a good reason to have both, but we should at the very least document the difference.

library/kani/src/bounded_arbitrary.rs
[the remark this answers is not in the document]
sgpthomas: Seems to pass the cover property:
```
Checking harness check_hashset...

VERIFICATION RESULT:
 ** 0 of 2161 failed (58 unreachable)

 ** 1 of 1 cover properties satisfied
```

library/kani/src/bounded_arbitrary.rs
sgpthomas: I missed this. I think probably we should just have the bounded version. So that we have an explicit API for bounded types, and so that it's easier to report this information to the user at some point in the future.

zhassan-aws: Yes, we should deprecate `any_vec`, since `BoundedArbitrary` provides a more general and uniform interface for creating bounded values for many data types.

library/kani/src/bounded_arbitrary.rs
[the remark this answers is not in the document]
zhassan-aws: That's likely because it's a single `cover` statement for multiple lengths, so if it is covered for one (in this case for `i == 1`), it will be marked as covered. If you separate them into two cover statements:
```rust
kani::cover!(hash_map.len() == 0);
kani::cover!(hash_map.len() == 1);
```
I expect the first to be uncovered.

library/kani_macros/src/derive_bounded.rs
```diff
@@ -0,0 +1,155 @@
...
+    let where_clauses = generics.type_params().map(|param| quote!(#param : #kani_path::Arbitrary));
+    if !params.is_empty() {
+        (quote!(<#(#params),*>), quote!(where #(#where_clauses),*))
+    } else {
+        Default::default()
+    }
+}
+
+/// Generates a symbolic value based on whether the field has the `#[bounded]`
+/// attribute. If the field is bounded, generate `kani::any()` otherwise generate
```
zhassan-aws, line 142:
```suggestion
/// attribute. If the field is not bounded, generate `kani::any()` otherwise generate
```
?

tests/expected/bounded-arbitrary/hash/hash.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: Please add a brief description of what this test is doing.

tests/expected/bounded-arbitrary/option/option.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
```
zhassan-aws, line 3: Same here.

tests/expected/bounded-arbitrary/reverse_vec/vec.expected
```diff
@@ -0,0 +1,10 @@
+Checking harness check_reverse_is_its_own_inverse_should_fail...\
+
+Failed Checks: unwinding assertion loop 0 \
```
zhassan-aws, line 3: Which harness has a failed unwinding assertion? We should increase the unwind value for it and make sure the check we're expecting to fail is the one failing.

tests/expected/derive-bounded-arbitrary/enum.expected
```diff
@@ -0,0 +1,3 @@
+Checking harness check_enum...
+
+Complete - 1 successfully verified harnesses, 0 failures, 1 total.
```
zhassan-aws, line 3: This doesn't check much since the harness only uses cover properties.

tests/expected/derive-bounded-arbitrary/enum.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+#[kani::proof]
+#[kani::unwind(6)]
+fn check_enum() {
+    let any_enum: Enum<bool> = kani::bounded_any::<_, 4>();
+    match any_enum {
+        Enum::A(s) => {
+            for i in 0..=4 {
+                kani::cover!(s.len() == i);
```
zhassan-aws, line 25: It's better to unroll this loop (i.e. have explicit statements for each length like in other tests) to make sure all cover properties are satisfied.

tests/expected/derive-bounded-arbitrary/struct.expected
```diff
@@ -0,0 +1,3 @@
+Checking harness check_my_vec...
+
+Complete - 1 successfully verified harnesses, 0 failures, 1 total.
```
zhassan-aws, line 3: Same comment here.

zhassan-aws (commented):
Looks great! Mostly minor comments. Please make sure to add a description to all the tests.

zhassan-aws (commented):
Also, can we make sure this doesn't break anything in https://github.com/model-checking/verify-rust-std/?

CC @tautschnig @carolynzech

library/kani_core/src/bounded_arbitrary.rs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            fn bounded_any<const N: usize>() -> Self {
+                let opt: Option<()> = any();
+                opt.map(|_| T::bounded_any::<N>())
+            }
+        }
+
+        impl<T, E> BoundedArbitrary for Result<T, E>
+        where
+            T: BoundedArbitrary,
```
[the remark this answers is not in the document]
sgpthomas: Unfortunately I don't think that there is a way to do this :/ Because it would result in conflicting trait implementations. Potentially we could use trait [specialization](https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md), but I think we should leave that to a future PR.

docs/src/SUMMARY.md
```diff
@@ -24,6 +24,7 @@
     - [Contracts](./reference/experimental/contracts.md)
     - [Loop Contracts](./reference/experimental/loop-contracts.md)
     - [Concrete Playback](./reference/experimental/concrete-playback.md)
+    - [Bounded Non-deterministic variables](./reference/experimental/bounded_arbitrary.md)
```
zhassan-aws, line 27: Why is this under the `Experimental features` section? These are features that currently require a `-Z` option.

zhassan-aws (approved):
Thanks!

PR #4001, opened 2025-04-09
Support `trait_upcasting`

### LINKED ISSUE #358: Update vtable implementation for initial Rust trait upcasting PR
We'll need to update our implementation as soon as this PR lands in Rust (which will be soon):
https://github.com/rust-lang/rust/pull/86461

### LINKED ISSUE #3998: Trait upcasting doesn't work properly with multiple supertraits
I tried this code (extracted and minimized from a real example):

```rust
trait SubTrait: SuperTrait2 + SuperTrait1 {}

impl<T: SuperTrait1 + SuperTrait2> SubTrait for T {}

trait SuperTrait1 {
    fn trigger(&self, _old: &()) {}
}

trait SuperTrait2 {
    #[allow(unused)]
    fn func(&self) {}
}

#[derive(Clone, Copy, Default)]
struct Struct;

impl SuperTrait1 for Struct {}
impl SuperTrait2 for Struct {}

#[kani::proof]
fn main() {
    let val: &dyn SubTrait = &Struct;
    (val as &dyn SuperTrait1).trigger(&());
}
```

```
cargo kani
```

with Kani version: 0.61.0 (034b0569561ebe841bf73155360a8f830e1512c5)

Instead, this happened: I get a strange assertion failure:

```
RESULTS:
Check 1: demo::main.missing_definition.1
	 - Status: FAILURE
	 - Description: "assertion"
	 - Location: Unknown file in function demo2::main
```
If we replace `fn trigger` with
```rs
    fn trigger(&self) {
        kani::assert(false, "should fail");
    }
```

We also unexpectedly do *not* get any kind of failure.

clubby789:
This adds support for the recently stabilized `trait_upcasting` feature. Currently, when we cast between non-matching dyn traits, codegen incorrectly reuses the old vtable, leading to confusing issues (#3998).

This reproduces the upcasting logic from upstream rustc to load the vptr when performing the coercion.

Resolves #358
Resolves #3998

This was tested using the motivating example from the linked issue. I will add further tests to make sure this feature is robust.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
@@ -26,7 +26,10 @@ use stable_mir::mir::mono::Instance;
 use stable_mir::mir::{
     AggregateKind, BinOp, CastKind, NullOp, Operand, Place, PointerCoercion, Rvalue, UnOp,
 };
-use stable_mir::ty::{ClosureKind, IntTy, RigidTy, Size, Ty, TyConst, TyKind, UintTy, VariantIdx};
+use stable_mir::ty::{
+    Binder, ClosureKind, ExistentialPredicate, IntTy, RigidTy, Size, Ty, TyConst, TyKind, UintTy,
+    VariantIdx,
+};
 use std::collections::BTreeMap;
 use tracing::{debug, trace, warn};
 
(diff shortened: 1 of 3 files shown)


zhassan-aws (commented):
Awesome, thanks for your contribution! Can you add a reference to the upstream implementation in the code? Otherwise, this is good to go.

zhassan-aws (approved):

PR #4002, opened 2025-04-09
Add an option to skip codegen

zhassan-aws:
This PR adds a new unstable option, `--no-codegen`, that can be used for running Kani, just to check if the code compiles, i.e. similar to running `cargo check`. This is useful as a lightweight check for compilation errors including code under `#[cfg(kani)]`.

This is implemented through passing rustc's `-Zno-codegen` option, which is what `cargo check` passes to `rustc`.

### DIFF
--- a/kani-driver/src/args/mod.rs
+++ b/kani-driver/src/args/mod.rs
@@ -242,6 +242,11 @@ pub struct VerificationArgs {
     #[arg(long, hide_short_help = true)]
     pub only_codegen: bool,
 
+    /// Run Kani without codegen. Useful for checking compilation errors (similar to `cargo check`).
+    /// This feature is unstable and requires `-Z unstable-options` to be used
+    #[arg(long, hide_short_help = true)]
+    pub no_codegen: bool,
+
     /// Specify the value used for loop unwinding in CBMC
     #[arg(long)]
     pub default_unwind: Option<u32>,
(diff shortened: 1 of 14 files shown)


kani-driver/src/project.rs
```diff
@@ -177,6 +177,13 @@ impl Artifact {
 /// be collected from the project.
 pub fn cargo_project(session: &mut KaniSession, keep_going: bool) -> Result<Project> {
     let outputs = session.cargo_build(keep_going)?;
+    if session.args.no_codegen {
+        info_operation(
+            "info",
+            "Skipping codegen. Rerun without `--no-codegen` to perform codegen.",
```
carolynzech, line 183: Nit:
```suggestion
            "Compilation succeeded up until codegen. Skipping codegen because of `--no-codegen` option. Rerun without `--no-codegen` to perform codegen.",
```

The point being to give some indicator of success.

kani-driver/src/args/mod.rs
```diff
@@ -242,6 +242,11 @@ pub struct VerificationArgs {
     #[arg(long, hide_short_help = true)]
     pub only_codegen: bool,
 
+    /// Run Kani without codegen. Useful for checking compilation errors (similar to `cargo check`).
```
carolynzech, line 245:
```suggestion
    /// Run Kani without codegen. Useful for quick feedback on whether the code would compile successfully (similar to `cargo check`).
```

carolynzech (approved):
Two documentation nits, take them or leave them. The changes in phrasing are meant to make clearer that this is more useful than the default in the success case.

PR #4023, opened 2025-04-15
Clarify installation instructions

### LINKED ISSUE #4008: installation instructions are highly unusual
typically, `cargo install` builds things from source, so if I see something advertising it as a method to install things not from source, it simply looks like a mistake.

it seems like `kani-verifier` actually contains a small installer script, and that's what is built from source.

it then uses rustup to perform the actual installation?

the wording of "this package contains proxy binaries" is also confusing, as it makes it sound like the actual binaries themselves were uploaded to crates.io.

the entire process feels needlessly obfuscated.

zhassan-aws:
1. Remove the reference to Python
2. Clarify what the installation steps do

Resolves #4008 

### DIFF
--- a/docs/src/install-guide.md
+++ b/docs/src/install-guide.md
@@ -14,7 +14,6 @@ GitHub CI workflows, see [GitHub CI Action](./install-github-ci.md).
 
 The following must already be installed:
 
-* **Python version 3.7 or newer** and the package installer `pip`.
 * Rust 1.58 or newer installed via `rustup`.
 
 ## Installing the latest version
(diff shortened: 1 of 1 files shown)


docs/src/install-guide.md
```diff
@@ -14,19 +14,20 @@ GitHub CI workflows, see [GitHub CI Action](./install-github-ci.md).
...
 
 ## Installing the latest version
 
 To install the latest version of Kani, run:
 
 ```bash
-cargo install --locked kani-verifier
-cargo kani setup
+cargo install --locked kani-verifier # build Kani's installer
+cargo kani setup # run the installer to download and install the prebuilt binaries as well as supporting libraries and data
```
tautschnig, line 25: May I suggest to break this pre-formatted code block into two and have the comments as plain text instead? Especially the second comment is very long, and pre-formatting will require that it be rendered as a single line.

tautschnig (approved):
Thank you!

docs/src/install-guide.md
zhassan-aws: Done. Let me know if this looks better.

tautschnig (approved):

PR #4025, opened 2025-04-17
Autoharness: Update Filtering Options

carolynzech:
## Summary
- fda814bcdfadec64b570c9362693f6ea92bd009e: Make the autoharness filter flags work on the standard library by moving them to `kani_compiler_flags`, which ensures they're passed to all Kani compiler invocations.
- c0430ec8aecb827ac29d81916081c91dc95665e6: Print the crate name in our output tables.
- 093fc6b57b56e8bac9f5fb3baa8a49b8c586f7b2: **Breaking Change** to rename `--include-function` and `--exclude-function` to mention `pattern`s instead, which makes it clearer that they talk about substrings of the total paths (e.g., modules). Also implement the suggestion from https://github.com/model-checking/kani/pull/3922#discussion_r1982170163 so that the flags are no longer mutually exclusive.
- 9e35fcab81ef0dcf8c0f01441b8b1a753948bdd8: Let the above flags filter on crate names as well.
- 66444e40d7d12ab7ad1b118cf814f17206be9ee4: Warn if an exclude flag makes an include flag moot.

## Detail
Some more context on why f933799c54b09210cb267963ff1dc431c7a9385a allows for both flags to be passed now: I realized as part of https://github.com/model-checking/kani/pull/3984 how when we call `cargo rustc` for a `cargo kani` invocation, we don't pass `--reachability` to dependencies to avoid running harnesses in them. The problem is that we can't do the same for our cargo command to build the standard library, since that uses `cargo build`, which does not have the same ability to pass flags only to the final compiler invocation and not the dependencies. So we end up passing `--reachability=AllFns` to the dependencies of the standard library as well and generate automatic harnesses for them. If we can pass both filter flags, we can run commands like `kani autoharness --std --include-pattern core --exclude-pattern miniz_oxide`, which will include functions from the `core` crate while excluding functions in the `miniz_oxide` that just happen to have the word "core" in them.


carolynzech:
https://github.com/model-checking/kani/pull/4024 includes the changes required to include the crate name as part of `kani autoharness --list`; since that's really more of a list subcommand change than an autoharness change, I thought it was better in its own PR.

carolynzech:
Leaving this PR in draft until I add a test for using `--include` and `--exclude` together, but I've done some local testing on the standard library and it's looking good.

kani-compiler/src/kani_middle/codegen_units.rs
```diff
@@ -394,12 +394,34 @@ fn automatic_harness_partition(
...
+            }
+            // If only --include-pattern was provided, filter out the function if included_patterns does not contain its name.
+            (false, true) => {
+                if !filter_contains(&name, &args.autoharness_included_patterns) {
+                    return Some(AutoHarnessSkipReason::UserFilter);
+                }
+            }
+            // If both are specified, filter out the function if included_patterns does not contain its name.
+            // Then, filter out any functions that excluded_patterns does match.
+            // This order is important, since it preserves the semantics described in kani_driver::autoharness_args where include takes precedence over exclude.
```
tautschnig, line 420: Doesn't the above mean that `exclude` takes precedence over `include`?

kani-driver/src/args/autoharness_args.rs
```diff
@@ -9,27 +9,28 @@ use crate::args::{ValidateArgs, VerificationArgs, validate_std_path};
...
+    pub include_pattern: Vec<String>,
+
+    /// Only create automatic harnesses for functions that do not match the given pattern.
+    /// This option can be provided multiple times, which will verify functions that do not match any of the patterns.
+    /// Kani considers a function to match the pattern if its fully qualified path contains PATTERN as a substring.
+
+    /// This option takes precedence over `--include-pattern`, i.e., Kani will first select all functions that match `--include-pattern`,
+    /// then exclude those that match `--exclude-pattern.`
+    /// Example: `--include-pattern foo --exclude-pattern foo::bar` creates automatic harnesses for all functions whose paths contain "foo" without "foo::bar".
+    /// Example: `--include-pattern foo::bar --exclude-pattern foo` makes the `--include-pattern` a no-op, since the exclude pattern is a superset of the include pattern.
```
tautschnig, line 30: Could you please add a new test that uses a combination of include and exclude?

tautschnig (approved):

kani-compiler/src/kani_middle/codegen_units.rs
carolynzech: Yes this was a misstatement, oops.

PR #4028, opened 2025-04-17
CI Formatting Fixes

carolynzech:
The majority of the changes in this PR were made automatically by running the command specified in 5314a5d772cf4e8f152334dcbaa081336b631e02, so the actual diff to review is much smaller than it initially appears.

1. If `--check` was provided to`kani-fmt.sh`, it would run `cargo fmt --check` but not `rustfmt --check`. This meant that our regression would catch formatting errors from `cargo fmt` but not `rustfmt` (it would just format the problematic files, which isn't useful in CI--we want it to fail instead). So add the `--check` flag to the `rustfmt` calls as well.
2. Run clippy with `kani_sysroot` feature enabled as well to catch more formatting errors.


carolynzech:
[Here](https://github.com/carolynzech/kani/actions/runs/14523887467/job/40750985301) are the warnings produced by 6cfa1f4d3834db0ef52ef16c08b4312aed7abd79, which the following commit automatically fixes.

zhassan-aws (approved):
Nice! Thanks for figuring this one out!

PR #4032, opened 2025-04-18
Make expected test more lenient

zhassan-aws:
During some experimentation with the stable branch, it seems that in some cases, the command that gets executed is `cargo kani`, i.e. without an explicit `+<toolchain>` or `+<stable-version>` argument. Make the expected test more lenient to avoid failures for those cases.

### DIFF
--- a/tests/cargo-ui/verbose-cmds/expected
+++ b/tests/cargo-ui/verbose-cmds/expected
@@ -1,5 +1,5 @@
 CARGO_ENCODED_RUSTFLAGS=
-cargo +
+cargo
 Running: `goto-cc
 Running: `goto-instrument
 Checking harness dummy_harness...


carolynzech (approved):
Is the command that gets executed in our control? I'd think it'd be better to change the command that gets executed to always include the +stable-version explicitly.

zhassan-aws:
I'm not sure. @rajath-mk do you know?

carolynzech:
Ok, thanks. I approved the change because I don't feel too strongly about it, so don't burn too much time if it's proving annoying for some reason.

rajath-mk:
Okay so there seems to be no issue with using the toolchain argument. And not removing it does seem to keep the code changes minimal. I would say we can close this PR in the interest of that. Thanks for pointing this out @carolynzech !

zhassan-aws:
Closing as discussed above.

PR #4037, opened 2025-04-21
Analyze unsafe code reachability

carolynzech:
Continuation of #3546.

From @celinval in #3546:

> Add call graph analysis to scanner in order to find the distance between functions in a crate and unsafe functions.
> 
> For that, we build the crate call graph and collect the unsafe functions. After that, do reverse BFS traversal from the unsafe functions and store the distance to other functions. The result is stored in a new csv file.

### DIFF
--- a/tests/script-based-pre/tool-scanner/scanner-test.expected
+++ b/tests/script-based-pre/tool-scanner/scanner-test.expected
@@ -1,6 +1,6 @@
 5 test_scan_fn_loops.csv
-19 test_scan_functions.csv
+20 test_scan_functions.csv
 5 test_scan_input_tys.csv
 16 test_scan_overall.csv
 3 test_scan_recursion.csv
-5 test_scan_unsafe_ops.csv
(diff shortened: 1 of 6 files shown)


carolynzech:
> What does this distance metric tell us?
More generally: what is the goal of the new analysis? And what else can we maybe do with it? 

On my local machine, I added a filter to the queue initialization [here](https://github.com/model-checking/kani/pull/4037/files#diff-67f1bc424bfea598aa230c955b47a37de6470151fb08da11cfe4de64562977acR28-R29) to filter out unsafe functions and safe abstractions as roots of the tree. By starting only from safe functions and performing the analysis, we can compute how many safe functions with transitive unsafe dependencies there are.

celinval:
Just for some context, initially I implemented this analysis to find out the percentage of safe functions that may lead to an unsafe operation. The distance is not as relevant, but it's something we get for free.

tautschnig (approved):

PR #4041, opened 2025-04-23
Scanner: log crate-level visibility of functions

tautschnig:
This will help determine whether a failed property appears in a public or private function.

### DIFF
--- a/tools/scanner/src/analysis.rs
+++ b/tools/scanner/src/analysis.rs
@@ -243,8 +243,9 @@ impl OverallStats {
                 if !kind.is_fn() {
                     return None;
                 };
-                let is_public = tcx.visibility(item.def_id()).is_public()
-                    || tcx.visibility(item.def_id()).is_visible_locally();
+                let int_def_id = rustc_internal::internal(*tcx, item.def_id());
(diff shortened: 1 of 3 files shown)


tools/scanner/src/analysis.rs
```diff
@@ -228,6 +232,30 @@ impl OverallStats {
...
+        let (public_fns, private_fns) = all_items
+            .into_iter()
+            .filter_map(|item| {
+                let kind = item.ty().kind();
+                if !kind.is_fn() {
+                    return None;
+                };
+                let int_def_id = rustc_internal::internal(*tcx, item.def_id());
+                let is_public =
+                    effective_visibilities.is_directly_public(int_def_id.expect_local());
```
carolynzech, line 249: If I'm understanding this enum's documentation correctly, I think you'd also want to include `reexported`. But I think you could just do `tcx.visibility(def_id).is_public()`?

tautschnig: Thank you!! I was unaware that I can get this from `TyCtxt` directly.

carolynzech (approved):

Issue #2631, opened 2023-07-27
ICE: Intrinsic `simd_add` crashes when using float vector


celinval:
I tried this code:

```rust
// Copyright Kani Contributors
// SPDX-License-Identifier: Apache-2.0 OR MIT

//! Ensure we can handle float vectors
#![allow(non_camel_case_types)]
#![feature(repr_simd, platform_intrinsics)]

extern "platform-intrinsic" {
    fn simd_add<T>(x: T, y: T) -> T;
    fn simd_eq<T, U>(x: T, y: T) -> U;
}

#[repr(simd)]
#[derive(Clone, PartialEq, kani::Arbitrary)]
pub struct f32x4(f32, f32);

impl f32x4 {
    fn as_array(&self) -> &[f32; 2] {
        unsafe { &*(self as *const f32x4 as *const [f32; 2]) }
    }
}

#[kani::proof]
fn check_sum() {
    let a = f32x4(0.0, 0.0);
    let b = kani::any::<f32x4>();
    let sum = unsafe { simd_add(a.clone(), b) };
    assert_eq!(sum.as_array(), a.as_array());
}

```

using the following command line invocation:

```
kani simd_float.rs
```

with Kani version: kani 0.33.0 (dev)

I expected to see this happen: Harness should verify correctly

Instead, this happened: Kani crashes

<details>

```
thread 'rustc' panicked at 'BinaryOperation Expression does not typecheck OverflowPlus Expr { value: Index { array: Expr { value: Symbol { identifier: "_RNvCs4OCrPoGFD1p_20simd_float_ops_fixme9check_sum::1::var_4" }, typ: Vector { typ: Float, size: 2 }, location: None, size_of_annotation: None }, index: Expr { value: IntConstant(0), typ: CInteger(SSizeT), location: None, size_of_annotation: None } }, typ: Float, location: None, size_of_annotation: None } Expr { value: Index { array: Expr { value: Symbol { identifier: "_RNvCs4OCrPoGFD1p_20simd_float_ops_fixme9check_sum::1::var_6" }, typ: Vector { typ: Float, size: 2 }, location: None, size_of_annotation: None }, index: Expr { value: IntConstant(0), typ: CInteger(SSizeT), location: None, size_of_annotation: None } }, typ: Float, location: None, size_of_annotation: None }', cprover_bindings/src/goto_program/expr.rs:1029:9
stack backtrace:
   0: rust_begin_unwind
[...]

PR #4049, opened 2025-04-28
Overflow operators can also be used with vectors

tautschnig:
We already supported the same for plus and minus, but need to equally do so for overflow-plus, overflow-minus.

Resolves: #2631


zhassan-aws (approved):
Thanks!

PR #4051, opened 2025-04-30
Contracts/Stubs for multiple inherent impls: fix checking the generic args path for equality

carolynzech:
#3829 fixed an issue where we couldn't verify contracts or stubs for types with multiple inherent impls. However, the logic to  check for path equality incorrectly assumed that there would only be one element in the path before the generic argument, when in fact the path can be arbitrarily long if there are modules involved. Change the logic to just look at the last two elements, since we can expect those will be the generic argument and the method name.

Before this PR, the new test would fail with this error:

```
error: Failed to resolve checking function NegativeNumber::<i32>::unchecked_mul because the generic arguments ::<i32> are invalid. The available implementations are: 
       num::negative::NegativeNumber::<i32>::unchecked_mul
       num::negative::NegativeNumber::<i16>::unchecked_mul
  --> /Users/cmzech/kani/tests/kani/FunctionContracts/multiple_inherent_impls.rs:57:5
   |
57 |     #[kani::proof_for_contract(NegativeNumber::<i32>::unchecked_mul)]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```

because we'd try to compare `::negative::NegativeNumber::<i32>::unchecked_mul` to `::<i32>::unchecked_mul` and find they weren't equal.

### DIFF
--- a/kani-compiler/src/kani_middle/resolve.rs
+++ b/kani-compiler/src/kani_middle/resolve.rs
@@ -715,6 +715,17 @@ fn is_item_name_with_generic_args(
     name: &str,
 ) -> bool {
     let item_path = tcx.def_path_str(item);
-    let all_but_base_type = item_path.find("::").map_or("", |idx| &item_path[idx..]);
-    all_but_base_type == format!("{generic_args}::{name}")
+    let parts: Vec<&str> = item_path.split("::").collect();
+
+    if parts.len() < 2 {
+        return false;
+    }
+
+    let actual_last_two =
+        format!("{}{}{}{}", "::", parts[parts.len() - 2], "::", parts[parts.len() - 1]);
+
+    let last_two = format!("{}{}{}", generic_args, "::", name);
+
+    // The last two components of the item_path should be the same as ::{generic_args}::{name}
+    last_two == actual_last_two
 }
--- a/kani-compiler/src/kani_middle/resolve.rs
+++ b/kani-compiler/src/kani_middle/resolve.rs
@@ -708,6 +708,8 @@ fn is_item_name(tcx: TyCtxt, item: DefId, name: &str) -> bool {
     last == name
 }
 
+/// Use this when we don't just care about the item name matching (c.f. is_item_name),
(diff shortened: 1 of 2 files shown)


kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -715,6 +715,17 @@ fn is_item_name_with_generic_args(
     name: &str,
 ) -> bool {
     let item_path = tcx.def_path_str(item);
-    let all_but_base_type = item_path.find("::").map_or("", |idx| &item_path[idx..]);
-    all_but_base_type == format!("{generic_args}::{name}")
+    let parts: Vec<&str> = item_path.split("::").collect();
```
tautschnig, line 718: Does this kind of splitting still work when you have something like `convert::num::<impl convert::From<num::nonzero::NonZero<i8>> for num::nonzero::NonZero<i128>>::from`? I know we have functions in the library that have such a fully-qualified path, but I don't know whether at this point inside Kani we would get to see this string.

tautschnig (approved):

kani-compiler/src/kani_middle/resolve.rs
carolynzech: Hm, one way to find out! I can write some unit tests.

tautschnig: I'd like to suggest the following path forward:
1. Look into Rust's implementation of `def_path_str` to see how they manage to produce information when `def_path` does not seem to provide the same.
2. Let's create an issue logging the additional problem that you identified above.
3. Let's merge this PR as-is, unless item 1 yields a better option.

carolynzech: 1. I took a look. It uses quite a bit of internals, I would prefer to just create the string and split it then go mucking about in internals too much; I think the odds of a bug/code complexity are too high.
2. Done, see https://github.com/model-checking/kani/issues/4057
3. Merging.

PR #4074, opened 2025-05-14
Clippy/Stylistic Fixes

carolynzech:
Various style nits I stumbled across while looking at our codegen code more closely.

### DIFF
--- a/.github/workflows/format-check.yml
+++ b/.github/workflows/format-check.yml
@@ -48,6 +48,7 @@ jobs:
       - name: 'Run Clippy'
         run: |
           cargo clippy --workspace -- -D warnings
+          cargo clippy --workspace --tests -- -D warnings
           RUSTFLAGS="--cfg=kani_sysroot" cargo clippy --workspace -- -D warnings
 
       - name: 'Print Clippy Statistics'
--- a/.github/workflows/format-check.yml
+++ b/.github/workflows/format-check.yml
@@ -47,7 +47,6 @@ jobs:
 
       - name: 'Run Clippy'
         run: |
-          cargo clippy --workspace -- -D warnings
           cargo clippy --workspace --tests -- -D warnings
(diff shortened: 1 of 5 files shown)


.github/workflows/format-check.yml
```diff
@@ -48,6 +48,7 @@ jobs:
       - name: 'Run Clippy'
         run: |
           cargo clippy --workspace -- -D warnings
+          cargo clippy --workspace --tests -- -D warnings
```
zhassan-aws, line 51: Does this subsume the previous command?

cprover_bindings/src/goto_program/typ.rs
```diff
@@ -1509,14 +1446,6 @@ mod type_tests {
         assert_eq!(type_def.type_name().unwrap().to_string(), format!("tag-{NAME}"));
     }
 
-    #[test]
```
zhassan-aws, line 1512: Why was the test removed?

carolynzech: Because it calls the function that was removed.

.github/workflows/format-check.yml
carolynzech: I didn't think so when I read [the documentation](https://doc.rust-lang.org/cargo/commands/cargo-check.html#target-selection), but I just played around with it locally and yes, you're correct.

.github/workflows/format-check.yml
```diff
@@ -48,6 +48,7 @@ jobs:
       - name: 'Run Clippy'
         run: |
           cargo clippy --workspace -- -D warnings
```
carolynzech, line 50:
```suggestion
```

cprover_bindings/src/goto_program/typ.rs
zhassan-aws: I see. The function was probably removed because it's public and not used anywhere (except from this test, which makes the test useless).

zhassan-aws (approved):

PR #4088, opened 2025-05-19
Fix the error that Kani panics when there is no external parameter in quantifier's closure.

thanhnguyen-aws:
Previously, if we run Kani on the following harness:
```Rust
#[kani::proof]
fn test() {
    let quan = kani::forall!(|j in (0, 100)| j < 100);
    assert!(quan);
}
```
it will panic: 
```
thread 'rustc' panicked at kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs:827:43:
index out of bounds: the len is 2 but the index is 2
```

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
@@ -568,6 +568,7 @@ impl GotocCtx<'_> {
     ///
     /// N.B. public only because instrinsics use this directly, too.
     pub(crate) fn codegen_funcall_args(&mut self, fn_abi: &FnAbi, args: &[Operand]) -> Vec<Expr> {
+        //println!("funcall_args: {:?}",args);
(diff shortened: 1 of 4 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/statement.rs
```diff
@@ -564,17 +565,54 @@ impl GotocCtx<'_> {
...
 
+    /// Generate Goto-C for each argument to a function call.
+    ///
+    /// N.B. public only because instrinsics use this directly, too.
+    pub(crate) fn codegen_funcall_args_for_quantifiers(
+        &mut self,
+        fn_abi: &FnAbi,
+        args: &[Operand],
+    ) -> Vec<Expr> {
+        //println!("funcall_args: {:?}",args);
```
tautschnig, line 576: I'd suggest removing all the commented-out `println` instructions.

tautschnig (approved):

PR #4090, opened 2025-05-19
Use our toolchain when invoking `cargo metadata`

### LINKED ISSUE #3997: Kani's list subcommand fails on s2n-quic
Steps to reproduce:
1. `git clone https://github.com/aws/s2n-quic`
2. `cd s2n-quic/quic/s2n-quic-core`
3. `cargo kani list -Z list`

with Kani version: 034b0569561

I expected to see this happen: Kani prints the list of harnesses

Instead, this happened:
```bash
$ cargo kani list -Z list
Kani Rust Verifier 0.61.0 (cargo plugin)
error: unexpected prerelease field, expected a version like "1.32"
error: Failed to execute cargo (exit status: 101). Found 0 compilation errors.
```

carolynzech:
## Summary
Change our invocation of `cargo metadata` to use Kani's toolchain.

## Explanation
Kani uses the [cargo_metadata](https://crates.io/crates/cargo_metadata) API to invoke `cargo metadata`. We were neglecting to override the Rust toolchain version with the one from our `rust-toolchain.toml` file, so the command would use the toolchain from the target crate. The issue is that Rust 1.77 [stabilized the package id format](https://doc.rust-lang.org/nightly/cargo/CHANGELOG.html#cargo-177-2024-03-21), so Kani, which is past 1.77, would expect the new version, and therefore can do this:
https://github.com/model-checking/kani/blob/5d76510f8881153292124176d02300e6c13c2ffa/kani-driver/src/call_cargo.rs#L207

and know that the string representation of the package ID will work with `-p`. Pre-stabilization, Cargo made no such guarantees, so fetching the metadata with a pre 1.77 toolchain would return an id that's not compatible with the `-p` flag.

Resolves #3997

### DIFF
--- a/kani-driver/src/call_cargo.rs
+++ b/kani-driver/src/call_cargo.rs
@@ -247,6 +247,9 @@ crate-type = ["lib"]
     pub fn cargo_metadata(&self, build_target: &str) -> Result<Metadata> {
         let mut cmd = MetadataCommand::new();
 
+        // Use Kani's toolchain when running `cargo metadata`
+        cmd.cargo_path(env!("CARGO"));
+
         // restrict metadata command to host platform. References:
         // https://github.com/rust-lang/rust-analyzer/issues/6908
         // https://github.com/rust-lang/rust-analyzer/pull/6912
--- a/kani-driver/src/call_cargo.rs
+++ b/kani-driver/src/call_cargo.rs
@@ -4,7 +4,9 @@
 use crate::args::VerificationArgs;
 use crate::call_single_file::{LibConfig, to_rustc_arg};
 use crate::project::Artifact;
-use crate::session::{KaniSession, lib_folder, lib_no_core_folder, setup_cargo_command};
+use crate::session::{
+    KaniSession, get_cargo_path, lib_folder, lib_no_core_folder, setup_cargo_command,
+};
 use crate::util;
 use anyhow::{Context, Result, bail};
 use cargo_metadata::diagnostic::{Diagnostic, DiagnosticLevel};
@@ -248,7 +250,8 @@ crate-type = ["lib"]
         let mut cmd = MetadataCommand::new();
 
         // Use Kani's toolchain when running `cargo metadata`
-        cmd.cargo_path(env!("CARGO"));
+        let cargo_path = get_cargo_path().unwrap();
+        cmd.cargo_path(cargo_path);
 
         // restrict metadata command to host platform. References:
         // https://github.com/rust-lang/rust-analyzer/issues/6908
--- a/kani-driver/src/session.rs
+++ b/kani-driver/src/session.rs
@@ -439,3 +439,17 @@ pub fn setup_cargo_command() -> Result<Command> {
 
     Ok(cmd)
 }
+
+// Get the cargo path corresponding to the toolchain version in rust-toolchain.toml.
+// If kani is being run in developer mode, then we use the compile-time toolchain, i.e. the one used during cargo build-dev.
+// For release versions of Kani, we use a version of cargo that's in the toolchain that's been symlinked during `cargo-kani` setup.
+pub fn get_cargo_path() -> Result<PathBuf> {
+    let install_type = InstallType::new()?;
+
+    let cargo_path = match install_type {
+        InstallType::DevRepo(_) => env!("CARGO").into(),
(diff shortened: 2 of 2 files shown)


carolynzech:
Thanks to @weihanglo for suggesting the `CARGO` environment variable!

kani-driver/src/call_cargo.rs
```diff
@@ -247,6 +247,9 @@ crate-type = ["lib"]
     pub fn cargo_metadata(&self, build_target: &str) -> Result<Metadata> {
         let mut cmd = MetadataCommand::new();
 
+        // Use Kani's toolchain when running `cargo metadata`
+        cmd.cargo_path(env!("CARGO"));
```
weihanglo, line 251: I didnt know it is using `cargo_metadata` crate. this might not help because it is already using that fallback: https://docs.rs/cargo_metadata/0.19.2/cargo_metadata/struct.MetadataCommand.html#method.cargo_path

weihanglo: Also, it is wrong using `env!` here. It expands at compile time. `std::env::var` is preferred.

https://github.com/rust-lang/rust-clippy/blob/66697e84b19c117b308eb1c02edcd6e6f299e30d/src/main.rs#L110-L110

carolynzech: > I didnt know it is using cargo_metadata crate. this might not help because it is already using that fallback: https://docs.rs/cargo_metadata/0.19.2/cargo_metadata/struct.MetadataCommand.html#method.cargo_path

This is a good point, but this fix works for me on https://github.com/model-checking/kani/issues/3997 locally. Before this PR:

```
cmzech@80a9971b5e20 s2n-quic-core % cargo kani                                  
error: unexpected prerelease field, expected a version like "1.32"
error: Failed to execute cargo (exit status: 101). Found 0 compilation errors.
```

with the change in this PR, Kani compiles and starts running the proofs as normal.

I don't know why that's happening, I'll look into it more tomorrow. My hunch is that this `cargo_path` override points it to *Kani's toolchain cargo*, whereas the CARGO env variable that `cargo_metadata` finds without the override points to `s2n-quic`'s toolchain version.

> Also, it might be wrong using env! here. It expands at compile time. std::env::var is preferred.
See how clippy does it https://github.com/rust-lang/rust-clippy/blob/66697e84b19c117b308eb1c02edcd6e6f299e30d/src/main.rs#L110-L110

Thanks! That's good to know. Kani uses `env!` for Rust environment variables in some other places, e.g. [here](https://github.com/model-checking/kani/blob/5d76510f8881153292124176d02300e6c13c2ffa/kani-driver/src/version.rs#L10)

Is `std::env::var` always preferred over `env!` If so, I'll change all the other uses too.

weihanglo: > Is std::env::var always preferred over env! If so, I'll change all the other uses too.

It is not a always/never thing. You need to know what the tool needs.

See the official doc to know more on environment variable: https://doc.rust-lang.org/cargo/reference/environment-variables.html

To me, I was just surprised that a compile time CARGO is used. That means there is no runtime behavior around CARGO. The value of it is determined and may contain your local user name (not user's unless they compile kani from source)

kani-driver/src/session.rs
```diff
@@ -439,3 +439,17 @@ pub fn setup_cargo_command() -> Result<Command> {
...
 }
+
+// Get the cargo path corresponding to the toolchain version in rust-toolchain.toml.
+// If kani is being run in developer mode, then we use the compile-time toolchain, i.e. the one used during cargo build-dev.
+// For release versions of Kani, we use a version of cargo that's in the toolchain that's been symlinked during `cargo-kani` setup.
+pub fn get_cargo_path() -> Result<PathBuf> {
+    let install_type = InstallType::new()?;
+
+    let cargo_path = match install_type {
+        InstallType::DevRepo(_) => env!("CARGO").into(),
```
zhassan-aws, line 450: Do you know where the `CARGO` environment variable is set?

zhassan-aws (approved):

kani-driver/src/session.rs
carolynzech: It's set when the crate is compiled, see https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates.

So we do it for development builds because we expect someone to build from source, then run.

PR #4108, opened 2025-05-27
Deprecate Options

carolynzech:
> Please ensure your PR description includes the following:
> 4. If you had to perform any manual test, please describe them.
>
> **Make sure you remove this list from the final PR description.**

Resolves #ISSUE-NUMBER


kani-driver/src/args/autoharness_args.rs
```diff
@@ -31,7 +33,7 @@ pub struct CommonAutoharnessArgs {
     #[arg(long = "exclude-pattern", num_args(1), value_name = "PATTERN")]
     pub exclude_pattern: Vec<String>,
 
-    /// Run the `list` subcommand after generating the automatic harnesses. Requires -Z list. Note that this option implies --only-codegen.
+    /// Run the `list` subcommand after generating the automatic harnesses. Note that this option implies --only-codegen.
```
zhassan-aws, line 34: Unrelated to this PR, but what does "implies --only-codegen" mean? Does Kani generate GOTO binaries?

kani-driver/src/args/mod.rs
```diff
@@ -56,7 +56,31 @@ pub fn print_deprecated(verbosity: &CommonArgs, option: &str, alternative: &str)
...
+    }
+}
+
+/// First step in two-phase stabilization.
+/// When an unstable option is first stabilized, print this warning that `-Z unstable-options` has no effect.
+/// This warning should last for one release only; in the next Kani release, remove it.
+pub fn print_stabilized_option_warning(verbosity: &CommonArgs, option: &str) {
+    if !verbosity.quiet {
+        warning(&format!(
+            "The `{option}` option has been stabilized, so -Z {} has no effect for it. \
```
zhassan-aws: Can the message be along the lines of the rustc one for stabilization? This is what rustc emits for stabilized features:
```
warning: the feature `X` has been stable since 1.26.0 and no longer requires an attribute to enable
```

kani-driver/src/args/mod.rs
```diff
@@ -56,7 +56,31 @@ pub fn print_deprecated(verbosity: &CommonArgs, option: &str, alternative: &str)
...
+            Remove it unless you are passing another unstable option.",
```
zhassan-aws:
```suggestion
            Remove it unless it is needed for another unstable option.",
```

kani-driver/src/args/autoharness_args.rs
carolynzech: Yes. It means that we generate GOTO binaries and don't run verification.

The initial implementation of this subcommand introduced a compiler flag to the [Arguments struct](https://github.com/model-checking/kani/blob/35e549b67efae1c4549cc6e65a9cafb3eb0c846c/kani-compiler/src/args.rs#L41) that informed the compiler that we were running the list subcommand. The compiler used this information to 1) conditionally generate the extra list metadata and 2) return after generating the metadata (before codegen). Celina requested that we instead make the compiler unaware of the subcommand, since then a subsequent run of Kani after `kani list` could reuse the build artifacts from the list run. (It also simplified the implementation, which is another benefit). Thus, the metadata for list actually gets produced on every run of Kani; the subcommand just reads that data and displays the result. See https://github.com/model-checking/kani/pull/3523#pullrequestreview-2316489499 for context.

Feel free to suggest a better phrasing of this comment in this PR.

zhassan-aws: Sounds good.

kani-driver/src/args/mod.rs
```diff
@@ -623,242 +645,223 @@ impl ValidateArgs for CargoKaniArgs {
...
-
-        self.common_args.check_unstable(
-            self.no_restrict_vtable,
-            "--no-restrict-vtable",
-            UnstableFeature::RestrictVtable,
-        )?;
+        // Check for any deprecated/obsolete options, or providing an unstable flag that has since been stabilized.
+        let deprecated_stabilized_obsolete = || -> Result<(), Error> {
+            self.checks.print_deprecated(&self.common_args);
+            self.check_unnecessary_unstable_option(self.jobs.is_some(), "jobs");
```
carolynzech, line 820: This line is the only addition to this function; the rest of the diff is just me grouping the checks into closures that describe what they do. Apologies that it makes the PR harder to read--I was having trouble convincing myself that our checks were comprehensive when they were intermingled.

I also removed the inline `--c-lib` error validation, since we already called `check_unstable` for `c-lib`. Hopefully the new closures in this function help us avoid such duplication and make it easier for people to do deprecation/stabilization work.

kani-driver/src/args/mod.rs
carolynzech: Done. Example output:

```
cmzech@80a9971b5e20 playground % cargo kani --default-checks --undefined-function-checks --jobs --output-format=terse -Z unstable-options
warning: The `--default-checks` option has been deprecated since 0.63.0 and will be removed soon. Consider omitting the argument, since this is already the default behavior instead
warning: The `--undefined-function-checks` option has been deprecated since 0.63.0 and will be removed soon. Consider omitting the argument, since this is already the default behavior instead
warning: The `--jobs` option has been stable since 0.63.0 and no longer requires -Z unstable-options to enable. Remove it unless it is needed for another unstable option.
```

kani_metadata/src/unstable.rs
```diff
@@ -115,6 +115,20 @@ impl UnstableFeature {
...
+    /// Serialize this feature into a format ideal for error messages.
+    pub fn as_argument_string(&self) -> String {
+        self.as_argument().join(" ")
+    }
+
+    /// If this unstable feature has been stabilized, return the version it was stabilized in.
+    /// Use this function to produce warnings that the unstable flag is no longer necessary.
+    pub fn stabilization_version(&self) -> Option<String> {
+        match self {
+            UnstableFeature::List => Some("0.63.0".to_string()),
```
carolynzech, line 128: Note to self: this PR assumes that it will be in the 0.63.0 release; change to 0.64.0 if it doesn't make it.

zhassan-aws (approved):
Thanks!

PR #3401, opened 2024-07-31
Add missing padding needded for alignment

zhassan-aws:
This PR adds extra padding for enums when needed for alignment. The missing alignment was exposed in #2857 where the layout size from MIR for `der::error::ErrorKind` was 48 bytes, and from `codegen_enum` was only 41 bytes. The difference turned out to be caused by an 8-byte alignment that the compiler uses for this enum.

Resolves #2857 
Resolves #3318 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #4112, opened 2025-05-28
Fix the bug: Static union values can panic Kani

### LINKED ISSUE #4097: Static `union` values can panic kani
Kani compilation can panic on static `union` type declarations.

I ran this code:

```rust
#![allow(dead_code)]

static FOO: Data = Data { a: [0; 3] };

union Data {
    a: [u8; 3],
    b: u16,
}

#[kani::proof]
fn main() {
    let _x = &FOO;
}
```

```
kani test.rs
```

with Kani version:

main (commit: 0b68df0b8ed585b4bae9d6c07db64875137f1ae7)

I expected to see this happen: Kani does not crash.

Instead, this happened: Kani crashes:

```
thread 'rustc' panicked at cprover_bindings/src/goto_program/expr.rs:958:9:
assertion `left == right` failed
  left: 32
 right: 24
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
   4: cprover_bindings::goto_program::expr::Expr::transmute_to
   5: kani_compiler::codegen_cprover_gotoc::codegen::operand::<impl kani_compiler::codegen_cprover_gotoc::context::goto_ctx::GotocCtx>::codegen_alloc_in_memory
   6: kani_compiler::codegen_cprover_gotoc::codegen::static_var::<impl kani_compiler::codegen_cprover_gotoc::context::goto_ctx::GotocCtx>::codegen_static
   7: std::thread::local::LocalKey<T>::with
   8: kani_compiler::codegen_cprover_gotoc::compiler_interface::GotocCodegenBackend::codegen_items
   9: <kani_compiler::codegen_cprover_gotoc::compiler_interface::GotocCodegenBackend as rustc_codegen_ssa::traits::backend::CodegenBackend>::codegen_crate::{{closure}}
  10: rustc_smir::rustc_internal::init
  11: scoped_tls::ScopedKey<T>::set
  12: rustc_smir::rustc_internal::run
  13: <kani_compiler::codegen_cprover_gotoc::compiler_interface::GotocCodegenBackend as rustc_codegen_ssa::traits::backend::CodegenBackend>::codegen_crate
  14: <rustc_interface::queries::Linker>::codegen_and_build_linker
  15: rustc_interface::passes::create_and_enter_global_ctxt::<core::option::Option<rustc_interface::queries::Linker>, rustc_driver_impl::run_compiler::{closure#0}::{closure#2}>
  16: rustc_interface::interface::run_compiler::<(), rustc_driver_impl::run_compiler::{closure#0}>::{closure#1}

[Kani] current codegen item: codegen_static: FOO
[Kani] current codegen location: Loc { file: "test.rs", function: None, start_line: 3, start_col: Some(1), end_line: 3, end_col: Some(17), pragmas: [] }
```

thanhnguyen-aws:
This PR fixes the bug: Static union values can panic Kani.

The reason behind the bug is Kani's translation of Union type is not consistent with its layout defined by Rust. Specifically, the size of the Union is not just the max of its fields but that maximum should be round up to a multiple of its alignment.
See:
https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/reference/type-layout.html#:~:text=The%20layout%20of%20a%20type,also%20part%20of%20type%20layout

In this PR, I add another variant for DataComponent: DataComponent::UnionField, with a padded_type of a Union (a struct which contains a padding and the real field type) to fix the translated layout and support reasoning about transmute function for union type. 

Resolves #4097


tests/expected/static_union/static_union2.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#![allow(dead_code)]
```
carolynzech, line 4: Can you just combine this with the other test? They seem quite similar.

carolynzech (commented):
I'm wondering if we can do something similar to https://github.com/model-checking/kani/pull/3401 instead of adding a padding field?

In particular, we already have logic for `codegen_alignment_padding`, so I'd prefer not to introduce a second way of reasoning about alignment and padding if we can avoid it.

carolynzech (commented):
I'm not following why we need a separate `DatatypeComponent` field for unions. Can't we represent the `Type` as a vector of `DatatypeComponent::Field` and `DatatypeComponent::Padding` like we do for structs and enums? So in the example of this test, we'd end up with a `Type` constructed from a vector of `DataTypeComponent`s where the first item is the `[u8; 3]` field, the second element is 1 byte of padding, the third element is the `u16` field, and the fourth element is 2 bytes of padding.

Happy to take this discussion offline if that's easier. I would prefer to use the same padding infrastructure that we have for structs and enums if possible. Perhaps it isn't possible, but if it's not I'm not seeing why.

thanhnguyen-aws:
> I'm not following why we need a separate `DatatypeComponent` field for unions. Can't we represent the `Type` as a vector of `DatatypeComponent::Field` and `DatatypeComponent::Padding` like we do for structs and enums? So in the example of this test, we'd end up with a `Type` constructed from a vector of `DataTypeComponent`s where the first item is the `[u8; 3]` field, the second element is 1 byte of padding, the third element is the `u16` field, and the fourth element is 2 bytes of padding.
> 
> Happy to take this discussion offline if that's easier. I would prefer to use the same padding infrastructure that we have for structs and enums if possible. Perhaps it isn't possible, but if it's not I'm not seeing why.

It is difficult to come up with a short answer for it, but after trying several ways of implementation, I feel that this is the most convenient way, for the following observations:
1. There are two things that we need to deal with: projection and transmute. For projection, we need to convert the union to the type of the field (with the layout of that type). But for transmute, we need to wrap the type in a struct with some padding, and use the exact layout of the union. 
2. It is not convenient to use Vec of Padding and Field like the case of struct, because in union, when we want to deal with transmute or calculate the length, for each Field we need the information of its following Padding, so Vec<DataComponent>.iter() is not convenient. 
3. If we wrap a Field and a Padding into a struct, and use it as the type of the union field, it is not convenient to implement projection because we need to deal with union separately (not consistent with other types). Many related functions: lookup_field_type, lookup_field do not take the parent type as argument, so modifying them just for union does not look pretty.  In contrast, for enum, although it is translated into union, we can do the wrapping because all if its variants are translated into struct types and we can add the padding inside it directly.
4. The wrapped struct (the padded_typ of UnionField) is only used in to_irep, so keeping the typ field to be the original mir type makes dealing with union consistent with other types everywhere else.

Issue #4007, opened 2025-04-10
Kani crashes on https://github.com/DataDog/libdatadog


danielsn:
I tried this code:
```sh
gh repo clone https://github.com/DataDog/libdatadog
cd libdatadog/datadog-crashtracker
cargo kani autoharness -Z autoharness
```

with Kani version: 0.61.

I expected to see this happen: Not to crash

Instead, this happened: 
```
error: internal compiler error: Kani unexpectedly panicked at panicked at cprover_bindings/src/goto_program/expr.rs:879:9:
                                Error in struct_expr; value type does not match field type.
                                    StructTag("tag-gimli::common::Encoding")
                                    [Field { name: "address_size", typ: Unsignedbv { width: 8 } }, Field { name: "format", typ: StructTag("tag-_109125825021383352072398879471058531001") }, Field { name: "version", typ: Unsignedbv { width: 16 } }]
                                    [Expr { value: Symbol { identifier: "_ZN5gimli4read7aranges30ArangeHeader$LT$R$C$Offset$GT$5parse17hde230b08470d762dE::1::var_36::val" }, typ: Unsignedbv { width: 8 }, location: None, size_of_annotation: None }, Expr { value: Symbol { identifier: "_ZN5gimli4read7aranges30ArangeHeader$LT$R$C$Offset$GT$5parse17hde230b08470d762dE::1::var_4::format" }, typ: StructTag("tag-_207664766485444868521932709622141405546"), location: None, size_of_annotation: None }, Expr { value: Symbol { identifier: "_ZN5gimli4read7aranges30ArangeHeader$LT$R$C$Offset$GT$5parse17hde230b08470d762dE::1::var_21::val" }, typ: Unsignedbv { width: 16 }, location: None, size_of_annotation: None }].


thread 'rustc' panicked at cprover_bindings/src/goto_program/expr.rs:879:9:
Error in struct_expr; value type does not match field type.
	StructTag("tag-gimli::common::Encoding")
	[Field { name: "address_size", typ: Unsignedbv { width: 8 } }, Field { name: "format", typ: StructTag("tag-_109125825021383352072398879471058531001") }, Field { name: "version", typ: Unsignedbv { width: 16 } }]
[...]

PR #4114, opened 2025-05-28
`ty_mangled_name`: only use non-mangled name if `-Zcffi` is enabled.

carolynzech:
See https://github.com/model-checking/kani/issues/4007#issuecomment-2914059181. Ideally, we'd have a better solution for C structs more generally, but this solution is enough to fix this particular crash.

Before this fix, the new diamond dependency test harness crashed with:

```
stderr:
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.00s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.00s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.00s
   Compiling main v0.1.0 (/Users/cmzech/kani/tests/cargo-kani/dependency-test/diamond-dependency/main)

thread 'rustc' panicked at cprover_bindings/src/goto_program/expr.rs:887:9:
Error in struct_expr; value type does not match field type.
        StructTag("tag-dependency3::ReprCStruct")
        [Field { name: "field", typ: StructTag("tag-_14480980443891294070469871430240215039") }]
        [Expr { value: Symbol { identifier: "_RNvCslFpuY9hqlfk_11dependency113create_struct::1::var_1" }, typ: StructTag("tag-_12679625102465156734981384663764492039"), location: None, size_of_annotation: None }]

[Kani] current codegen item: codegen_function: dependency1::create_struct
_RNvCslFpuY9hqlfk_11dependency113create_struct
[Kani] current codegen location: Loc { file: "/Users/cmzech/kani/tests/cargo-kani/dependency-test/diamond-dependency/dependency1/src/lib.rs", function: None, start_line: 15, start_col: Some(1), end_line: 15, end_col: Some(51), pragmas: [] }
error: could not compile `main` (lib)
```

Resolves https://github.com/model-checking/kani/issues/4007

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
@@ -4,6 +4,7 @@ use crate::codegen_cprover_gotoc::GotocCtx;
 use cbmc::goto_program::{DatatypeComponent, Expr, Location, Parameter, Symbol, SymbolTable, Type};
 use cbmc::utils::aggr_tag;
 use cbmc::{InternString, InternedString};
+use kani_metadata::UnstableFeature;
 use rustc_abi::{
     BackendRepr::SimdVector, FieldIdx, FieldsShape, Float, Integer, LayoutData, Primitive, Size,
     TagEncoding, TyAndLayout, VariantIdx, Variants,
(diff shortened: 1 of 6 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -487,9 +488,20 @@ impl<'tcx> GotocCtx<'tcx> {
         // This is not a restriction because C can only access non-generic types anyway.
         // TODO: Skipping name mangling is likely insufficient if a dependent crate has two versions of
         // linked C libraries
-        // https://github.com/model-checking/kani/issues/450
+        // https://github.com/model-checking/kani/issues/450.
+        // Note that we only mangle #[repr(C)] types if the unstable c-ffi feature is enabled; otherwise,
```
zhassan-aws: Should this be "we **don't** mangle #[repr(C)] types if the unstable c-ffi feature is enabled"?

tests/cargo-kani/dependency-test/diamond-dependency/main/src/lib.rs
```diff
@@ -12,3 +12,11 @@ fn harness() {
     assert!(dependency1::delegate_use_struct() == 3);
     assert!(dependency2::delegate_use_struct() == 1);
 }
+
+// Test that Kani can codegen repr(C) structs from two different versions of the same crate,
+// c.f. https://github.com/model-checking/kani/issues/4007
+#[kani::proof]
+fn reprc_harness() {
```
zhassan-aws, line 19: Does this harness fail without this change?

carolynzech: The harness crashes without the change, see the PR description.

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
carolynzech: 🤦‍♀️ good catch

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -487,9 +488,20 @@ impl<'tcx> GotocCtx<'tcx> {
...
+        // we assume that this struct is a #[repr(C)] in Rust code and mangle it,
```
carolynzech:
```suggestion
        // Note that #[repr(C)] types are unmangled only if the unstable c-ffi feature is enabled; otherwise,
        // we assume that this struct is a #[repr(C)] in Rust code and mangle it,
```

PR #4127, opened 2025-06-03
Set target features depending on the target architecture

zhassan-aws:
Updates the `target_config` method to set the target features depending on the target architecture. 

Resolves #4119 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #4129, opened 2025-06-04
Refactor simd_bitmask to reduce the number of iterations

zhassan-aws:
This PR is co-authored with @tautschnig.

Background: In https://github.com/model-checking/kani/pull/4127 we enabled certain target features that are platform dependent (e.g. `sse` and `sse2` for x86_64 and `neon` for aarch64) which resulted in using the `simd_bitmask` intrinsic more frequently. Kani's current model of that intrinsic (https://github.com/model-checking/kani/blob/5f14b735b74f3ae3f9f1c64ce5656e1e735d42ea/library/kani/src/models/mod.rs#L72) uses a loop that iterates `LANE` times, which requires harnesses that touch this code (e.g. that use the `HashSet` data structure) to have a large unwinding value, which results in a significant slowdown.

This PR refactors this function through rewriting it using a nested loop and manually unwinding the inner loop (which operates over the bits in a byte) so that a large unwinding value is not needed.

### DIFF
--- a/library/kani/src/models/mod.rs
+++ b/library/kani/src/models/mod.rs
@@ -74,17 +74,79 @@ mod intrinsics {
         T: MaskElement,
     {
         let mut mask_array = [0; mask_len(LANES)];
-        for lane in (0..input.len()).rev() {
-            let byte = lane / 8;
-            let mask = &mut mask_array[byte];
-            let shift_mask = *mask << 1;
-            *mask = if input[lane] == T::TRUE {
-                shift_mask | 0x1
-            } else {
-                assert_eq!(input[lane], T::FALSE, "Masks values should either be 0 or -1");
-                shift_mask
-            };
+        let mask_len = mask_array.len();
+
+        // Process 8 lanes at a time when possible
+        for byte in 0..mask_len {
+            // Calculate the starting lane for this byte
+            let start_lane = byte * 8;
+            // Calculate how many bits to process (handle the last byte which might be partial)
+            let bits_to_process = (LANES - start_lane).min(8);
+
+            // disable formatting for this code block to make it easier to edit/read
+            #[rustfmt::skip]
+            let byte_mask =
+                if bits_to_process > 0 && input[start_lane + 0] == T::TRUE { 1 << 0 } else { 0 } |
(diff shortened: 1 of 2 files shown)


library/kani/src/models/mod.rs
```diff
@@ -74,17 +74,79 @@ mod intrinsics {
...
-            let byte = lane / 8;
-            let mask = &mut mask_array[byte];
-            let shift_mask = *mask << 1;
-            *mask = if input[lane] == T::TRUE {
-                shift_mask | 0x1
-            } else {
-                assert_eq!(input[lane], T::FALSE, "Masks values should either be 0 or -1");
-                shift_mask
-            };
+        let mask_len = mask_array.len();
```
tautschnig: Is this re-definition perhaps more confusing than providing a value-add? I'd suggest to just get rid of it.

tests/kani/SIMD/simd_bitmask_equiv.rs
```diff
@@ -0,0 +1,170 @@
...
+            shift_mask | 0x1
+        } else {
+            assert_eq!(input[lane], T::FALSE, "Masks values should either be 0 or -1");
+            shift_mask
+        };
+    }
+    mask_array
+}
+
+pub fn simd_bitmask_impl2<T, const LANES: usize>(input: &[T; LANES]) -> [u8; mask_len(LANES)]
```
tautschnig: Can we not invoke the implementation from library/kani/src/models/mod.rs?

library/kani/src/models/mod.rs
zhassan-aws: The second commit got rid of it.

tests/kani/SIMD/simd_bitmask_equiv.rs
zhassan-aws: Not directly since it's not a public function :(

tautschnig: But ... can't we use `simd_bitmask`?

zhassan-aws: We can call the intrinsic directly. Let me check.

zhassan-aws: Done

PR #4136, opened 2025-06-05
Store the git log in a file

### LINKED ISSUE #4128: Automatic toolchain upgrade fails with argument-list-too-long
https://github.com/model-checking/kani/actions/runs/15432529867/job/43432847335 failed as the git_log is exported as an environment variable, yet that log appears too long for other commands to make use of.

We should try to communicate it via a file.

zhassan-aws:
Switch from using an environment variable for storing the git log to using a file.

Resolves #4128 


zhassan-aws:
This shouldn't be necessary. See https://github.com/model-checking/kani/issues/4128#issuecomment-2946411954. Closing.

PR #4138, opened 2025-06-05
Flamegraph

### LINKED ISSUE #4075: Add a debug option to measure fine-grained compiler performance
**Proposed change:** For a given code generation pass, we'd get the call tree annotated with the time spent in each function. For instance, if `codegen_items` calls `codegen_function`, which calls `codegen_statement` `n` times, etc., we'd get the time we spent in each of those calls, down to the end of the call tree. I'm imagining we'd visualize it as a tree because the depth of the call stack is also important, but I'm flexible on the format; the main point is getting data on how long different *parts* of codegen take, not just codegen as a whole.

**Motivation:** From some rudimentary benchmarking with `with_timer`:
https://github.com/model-checking/kani/blob/2227614c99bf75a08ea14b23ad5ad42bfbac117a/kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs#L831-L833

we've determined that most of Kani's compiler runtime comes from code generation. But we need more fine-grained data to know *why* code generation is slow. Currently, it's unclear whether:

- Most functions are fast enough, but there are specific outliers we can blame, e.g., if codegening a function with fat pointers or dynamic dispatch is much slower than codegening functions without, or 
- Most functions are fast enough in isolation, but we have a deep call tree so their cumulative time is slow, suggesting that we need a larger-scale refactoring to produce speedups, or at least small-scale optimizations of multiple functions.

Having this data would help us make that determination.

AlexanderPortland:
Adds the ability for `cargo kani` to optionally output the info needed for generating flamegraphs of the compiler & driver's performance. When enabled with the `FLAMEGRAPH` environment variable, it will generate an output file for the driver and one for each crate in the workspace as they are compiled. These output files can then be rendered as flamegraphs to see where Kani is spending most of its execution time.

This should help us pinpoint which pieces of the compilation process are most to blame for our compilation times and focus future efforts to improve them.

Resolves #4075


docs/src/profiling.md
```diff
@@ -0,0 +1,25 @@
...
+
+To profile Kani's performance at a fine-grained level, we use a tool called [`samply`](https://github.com/mstange/samply) that allows the compiler & driver to periodically record the current stack trace, allowing us to construct flamegraphs of where they are spending most of their time.
+
+## Install samply
+First, install `samply` using [the instructions](https://github.com/mstange/samply?tab=readme-ov-file#installation) from their repo. The easier methods include installing a prebuilt binary or installing from crates.io.
+
+
+## Running Kani for profiling output
+1. First, build Kani from source with `cargo build-dev --profile profiling` to ensure you are getting all release mode optimizations without stripping useful debug info.
+2. Then, you can profile the Kani compiler on a crate of your choice by running `FLAMEGRAPH=[OPTION] cargo kani` within the crate.
```
carolynzech:
```suggestion
2. Then, you can profile the Kani compiler on a crate of your choice by [exporting Kani to your local PATH](build-from-source.md#adding-kani-to-your-path) and running `FLAMEGRAPH=[OPTION] cargo kani` within the crate.
```

docs/src/profiling.md
```diff
@@ -0,0 +1,25 @@
...
+
+The `FLAMEGRAPH` environment variable can be set to `driver` (to profile the complete `kani-driver` execution), `compiler` (to profile each time the `kani-compiler` is called) or `all` (to profile both).
```
carolynzech:
```suggestion
The `FLAMEGRAPH` environment variable can be set to `driver` (to profile the complete `kani-driver` execution), `compiler` (to profile each time the `kani-compiler` is called) or `both` (to profile both).
```
Nit: I prefer `both` since "all" implies > 2

docs/src/profiling.md
```diff
@@ -0,0 +1,25 @@
...
+
+We have to instrument the driver and compiler separately because samply's instrumentation usually cannot handle detecting the subprocess the driver uses to call the compiler.
+
+> Note: Specifically when profiling the compiler, ensure you are running `cargo clean` immediately before `cargo kani`, or parts of the workspace may not be recompiled by the Kani compiler.
+
+
+## Displaying profiling output
+This will create a new `flamegraph` directory in the crate which will contain a single `driver.json.gz` output file and one `compiler-{crate_name}.json.gz` file for each crate in the workspace. Run `samply load flamegraphs/XXX.json.gz` on any of these to open a local server that will display the file's flamegraph.
```
carolynzech:
```suggestion
This will create a new `flamegraphs` directory in the crate which will contain a single `driver.json.gz` output file and one `compiler-{crate_name}.json.gz` file for each crate in the workspace. Run `samply load flamegraphs/XXX.json.gz` on any of these to open a local server that will display the file's flamegraph.
```

kani-driver/src/session.rs
```diff
@@ -418,18 +418,46 @@ fn init_logger(args: &VerificationArgs) {
...
+            let flamegraph_dir = "flamegraphs";
+
+            if let Some(profiler_out_path) = profiling_out_path
+                && instrument_compiler
+            {
+                let _ = std::fs::create_dir(flamegraph_dir); // create temporary flamegraph directory
+
+                let mut cmd = Command::new("samply");
+                cmd.arg("record");
+                cmd.arg("-r").arg("8000"); // adjust the sampling rate (in Hz)
```
carolynzech: ditto re: constant

kani-driver/src/session.rs
```diff
@@ -418,18 +418,46 @@ fn init_logger(args: &VerificationArgs) {
...
         InstallType::DevRepo(_) => {
-            let mut cmd = Command::new("cargo");
-            cmd.arg(self::toolchain_shorthand());
-            cmd
+            // check if we should instrument the compiler for a flamegraph
+            let instrument_compiler = matches!(
+                std::env::var("FLAMEGRAPH"),
+                Ok(ref s) if s == "all" || s == "compiler"
+            );
+            let flamegraph_dir = "flamegraphs";
```
carolynzech: let's make this a constant at the top of the file

scripts/cargo-kani
```diff
@@ -18,4 +18,16 @@ then
...
+
+if [[ ( -n "${FLAMEGRAPH:-}" ) && ( "${FLAMEGRAPH}" == "driver" || "${FLAMEGRAPH}" == "all" ) ]]; then
+    FLAMEGRAPH_DRIVER=1
+fi
+
+if (( FLAMEGRAPH_DRIVER == 0 )); then 
+    exec -a cargo-kani "${KANI_DRIVER}" "$@"
+else 
+    mkdir -p "${FLAMEGRAPH_DIR}"
+    exec -a cargo-kani samply record -r 8000 -o "${FLAMEGRAPH_DIR}/driver.json.gz" --save-only "${KANI_DRIVER}" "$@"
```
carolynzech: let's make the sampling rate a constant at the top too so that it's easier for the reader to tell what it is

docs/src/profiling.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+## Running Kani for profiling output
+1. First, build Kani from source with `cargo build-dev --profile profiling` to ensure you are getting all release mode optimizations without stripping useful debug info.
+2. Then, you can profile the Kani compiler on a crate of your choice by running `FLAMEGRAPH=[OPTION] cargo kani` within the crate.
+
+The `FLAMEGRAPH` environment variable can be set to `driver` (to profile the complete `kani-driver` execution), `compiler` (to profile each time the `kani-compiler` is called) or `all` (to profile both).
+
+We have to instrument the driver and compiler separately because samply's instrumentation usually cannot handle detecting the subprocess the driver uses to call the compiler.
+
```
carolynzech, line 16: Can you add a note about the sampling rate: that there's a default but you can go into the code and adjust it? I'd add comments to the compiler and driver code mentioning that it can be adjusted, but that if you're profiling the whole run you should make sure that the rates in the compiler and driver are the same--i.e., if you go change it one place you should change it in the other

carolynzech (commented):
Awesome work! First round of comments.
As a general note, it'd be great if we could add a unique suffix to the outputted graphs so that new graphs don't overwrite old ones (perhaps the current time, [example](https://github.com/brownsys/paralegal-bench/blob/ff1c2a6ae4e54a78d21a7cd6a6be1f35e119cbe1/griswold/src/run.rs#L133-L142))

Cargo.lock
```diff
@@ -184,6 +199,12 @@ dependencies = [
  "which",
 ]
 
+[[package]]
+name = "bumpalo"
+version = "3.18.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "793db76d6187cd04dff33004d8e6c9cc4e05cd330500379d2394209271b4aeee"
+
```
carolynzech, line 213: I think at least some of these Cargo.lock changes are coming from other work you're doing unrelated to this PR --can you remove the ones that aren't related to chrono?

kani-driver/src/session.rs
```diff
@@ -25,6 +25,9 @@ pub const BUG_REPORT_URL: &str =
 /// the driver logs separately, by using the logger directives to  select the kani-driver crate.
 /// `export KANI_LOG=kani_driver=debug`.
 const LOG_ENV_VAR: &str = "KANI_LOG";
+const FLAMEGRAPH_ENV_VAR: &str = "FLAMEGRAPH";
```
carolynzech, line 29:
```suggestion
// Constants related to the option to create flamegraphs to debug compiler performance. See our mdbook's developer documentation for details.
const FLAMEGRAPH_ENV_VAR: &str = "FLAMEGRAPH";
```

kani-driver/src/session.rs
```diff
@@ -434,21 +437,22 @@ pub fn setup_cargo_command_inner(profiling_out_path: Option<String>) -> Result<C
...
+                std::env::var(FLAMEGRAPH_ENV_VAR),
+                Ok(ref s) if s == "compiler"
             );
-            let flamegraph_dir = "flamegraphs";
 
             if let Some(profiler_out_path) = profiling_out_path
                 && instrument_compiler
             {
-                let _ = std::fs::create_dir(flamegraph_dir); // create temporary flamegraph directory
+                let _ = std::fs::create_dir_all(FLAMEGRAPH_DIR); // create temporary flamegraph directory
```
carolynzech:
```suggestion
                // create temporary flamegraph directory
                let _ = std::fs::create_dir_all(FLAMEGRAPH_DIR);
```

kani-driver/src/session.rs
```diff
@@ -434,21 +437,22 @@ pub fn setup_cargo_command_inner(profiling_out_path: Option<String>) -> Result<C
...
+                let time_postfix = chrono::Utc::now().format("%Y-%m-%dT%H:%M:%S");
 
                 let mut cmd = Command::new("samply");
                 cmd.arg("record");
-                cmd.arg("-r").arg("8000"); // adjust the sampling rate (in Hz)
-                cmd.arg("-o")
-                    .arg(format!("{flamegraph_dir}/compiler-{profiler_out_path}.json.gz",)); // set the output file
+                cmd.arg("-r").arg(FLAMEGRAPH_SAMPLING_RATE); // adjust the sampling rate (in Hz)
```
carolynzech:
```suggestion
                // adjust the sampling rate (in Hz)
                cmd.arg("-r").arg(FLAMEGRAPH_SAMPLING_RATE);
```

carolynzech (commented):
Once the Cargo.lock thing is fixed, I'll approve!
(also, I assume you've tested this again with the new date postfix thing--can you do some sanity check runs locally to make sure everything looks good with consecutive runs?)

Cargo.lock
AlexanderPortland: I thought the same thing, but I double checked and it looks like this may just be an indirect dependency of `chrono` (through chrono > wasm-bindgen > ... > wasm-bindgen-backend > bumpalo)

zhassan-aws: Do we need the chrono's `wasmbind` feature? See default features here:

https://docs.rs/chrono/0.4.41/chrono/#features

AlexanderPortland: Thats a very good idea, I've removed our dependency on the `wasmbind` feature. We still seem to indirectly need this whole `wasm-bindgen` subtree in our Cargo.lock (through a time zone crate we need for chronos `clock` feature to write out the local time), but this is a good change!

kani-driver/src/session.rs
```diff
@@ -444,18 +445,22 @@ pub fn setup_cargo_command_inner(profiling_out_path: Option<String>) -> Result<C
             if let Some(profiler_out_path) = profiling_out_path
                 && instrument_compiler
             {
-                let _ = std::fs::create_dir_all(FLAMEGRAPH_DIR); // create temporary flamegraph directory
-                let time_postfix = chrono::Utc::now().format("%Y-%m-%dT%H:%M:%S");
+                // create temporary flamegraph directory
+                let _ = std::fs::create_dir_all(FLAMEGRAPH_DIR);
```
carolynzech:
```suggestion
                std::fs::create_dir_all(FLAMEGRAPH_DIR)?;
```

PR #4142, opened 2025-06-06
Edit quantifiers' documentation.

### LINKED ISSUE #4133: Quantifiers documentation is wrong
I tried this code from the [quantifiers documentation](https://model-checking.github.io/kani/reference/experimental/quantifiers.html#supported-quantifiers):

```rust
#[kani::proof]
fn test_forall() {
    let v = vec![10; 10];
    kani::assert(kani::forall!(|i in 0..10| v[i] == 10));
}
```

```
cargo kani -Z function-contracts
```

with Kani version: 0.62.0

```
error[E0061]: this function takes 2 arguments but 1 argument was supplied
  --> src/lib.rs:24:5
   |
24 |     kani::assert(kani::forall!(|i in 0..10| v[i] == 10));
   |     ^^^^^^^^^^^^---------------------------------------- argument #2 of type `&'static str` is missing
   |
note: function defined here
  --> /Users/cmzech/kani/library/kani/src/lib.rs:55:1
   |
55 | kani_core::kani_lib!(kani);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: this error originates in the macro `kani_core::kani_intrinsics` which comes from the expansion of the macro `kani_core::kani_lib` (in Nightly builds, run with -Z macro-backtrace for more info)
help: provide the argument
   |
24 |     kani::assert(kani::forall!(|i in 0..10| v[i] == 10), /* &'static str */);
   |                                                        ++++++++++++++++++++

```

I added a message, then got this error:

```rust
   Compiling playground v0.1.0 (/Users/cmzech/playground)
error: no rules expected `0`
  --> src/lib.rs:24:38
   |
24 |     kani::assert(kani::forall!(|i in 0..10| v[i] == 10), "msg");
   |                                      ^ no rules expected this token in macro call
   |
note: while trying to match `(`
  --> /Users/cmzech/kani/library/kani/src/lib.rs:55:1
   |
55 | kani_core::kani_lib!(kani);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: this error originates in the macro `kani_core::kani_intrinsics` which comes from the expansion of the macro `kani_core::kani_lib` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I got this to work:

```rust
#[kani::proof]
fn test_forall() {
    let v = vec![10; 10];
    kani::assert(kani::forall!(|i in (0,10)| v[i] == 10), "msg");
}
```

So we should update our documentation and RFC to match the syntax we actually support.

thanhnguyen-aws:
This PR fixed the examples in the documentation of quantifier feature.

Resolves #4133

### DIFF
--- a/docs/src/reference/experimental/quantifiers.md
+++ b/docs/src/reference/experimental/quantifiers.md
@@ -14,8 +14,8 @@ Kani currently supports the following quantifiers:
 ```rust
 #[kani::proof]
 fn test_forall() {
-    let v = vec![10; 10];
-    kani::assert(kani::forall!(|i in 0..10| v[i] == 10));
+    let v : [u8;10] = [10; 10];
+    assert!(kani::forall!(|i in (0,10)| v[i] == 10) );
 }
 ```
 
@@ -27,10 +27,12 @@ fn test_forall() {
 ```rust
 #[kani::proof]
 fn test_exists() {
-    let v = vec![1, 2, 3, 4, 5];
-    kani::assert(kani::exists!(|i in 0..v.len()| v[i] == 3));
+    let v : [u8;5] = [1, 2, 3, 4, 5];
+    assert!(kani::exists!(|i in (0,v.len())| v[i] == 3));
 }
 ```
+
+When the range is not specified, it is assumed to be the whole `usize` range.
 
 ### Limitations
 
@@ -39,7 +41,6 @@ fn test_exists() {
 The performance of quantifiers can be affected by the depth of call stacks in the quantified expressions. If the call stack is too deep, Kani may not be able to evaluate the quantifier effectively, leading to potential timeouts or running out of memory. Actually, array indexing in Rust leads to a deep call stack, which can cause issues with quantifiers. To mitigate this, consider using *unsafe* pointer dereferencing instead of array indexing when working with quantifiers. For example:
 
 ```rust
-
 #[kani::proof]
 fn vec_assert_forall_harness() {
     let v = vec![10 as u8; 128];


docs/src/reference/experimental/quantifiers.md
```diff
@@ -14,8 +14,8 @@ Kani currently supports the following quantifiers:
 ```rust
 #[kani::proof]
 fn test_forall() {
-    let v = vec![10; 10];
-    kani::assert(kani::forall!(|i in 0..10| v[i] == 10));
+    let v : [u8;10] = [10; 10];
+    assert!(kani::forall!(|i in (0,10)| v[i] == 10) );
```
carolynzech, line 18:
```suggestion
    let v : [u8;10] = [10; 10];
    kani::assert(kani::forall!(|i in (0,10)| v[i] == 10), "array contents did not match expectations");
```
We may as well showcase `kani::assert` IMO

docs/src/reference/experimental/quantifiers.md
```diff
@@ -27,19 +27,20 @@ fn test_forall() {
 ```rust
 #[kani::proof]
 fn test_exists() {
-    let v = vec![1, 2, 3, 4, 5];
-    kani::assert(kani::exists!(|i in 0..v.len()| v[i] == 3));
+    let v : [u8;5] = [1, 2, 3, 4, 5];
+    assert!(kani::exists!(|i in (0,v.len())| v[i] == 3));
```
carolynzech, line 31:
```suggestion
    let v : [u8;5] = [1, 2, 3, 4, 5];
    kani::assert(kani::exists!(|i in (0,v.len())| v[i] == 3), "array contents do not match expectations");
```

docs/src/reference/experimental/quantifiers.md
```diff
@@ -27,19 +27,20 @@ fn test_forall() {
...
 }
 ```
 
+When the range is not specified, it is assumed to be the whole `usize` range.
+
```
carolynzech: Can you include an example of what it looks like to not specify the whole range?

docs/src/reference/experimental/quantifiers.md
zhassan-aws: @carolynzech Is `kani::assert` required to use quantifiers? If not, I would prefer keeping the examples in terms of Rust's `assert`.

carolynzech: No! Regular assert is fine too. I don't feel strongly about this.

docs/src/reference/experimental/quantifiers.md
thanhnguyen-aws: The syntax is supported, but I could not find an example that works. Even the most simple one runs forever: 
```rust
 assert!(kani::exists!(| i | i == 3));
```

carolynzech: Hm, if that's the case I don't think it's worth mentioning?

carolynzech (approved):

PR #4143, opened 2025-06-06
Add loop-contracts support for `for` loop

thanhnguyen-aws:
This PR adds loop-contracts support for `for` loop for 4 types: array, slice, Iter, and Vec.

### DIFF
--- a/docs/src/reference/experimental/loop-contracts.md
+++ b/docs/src/reference/experimental/loop-contracts.md
@@ -93,7 +93,7 @@ Complete - 1 successfully verified harnesses, 0 failures, 1 total.
 ```
 
 
-## Loop contracts for `while` loops
+## Loop contracts
 
 ### Syntax
 > 
@@ -168,10 +168,12 @@ Kani automatically contracts (instead of unwinds) all loops in the functions tha
 
 Loop contracts comes with the following limitations.
 
-1. `while` loops and `loop` loops are supported. The other kinds of loops are not supported: [`while let` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#predicate-pattern-loops), and [`for` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#iterator-loops).
-2. Kani infers *loop modifies* with alias analysis. Loop modifies are those variables we assume to be arbitrary in the inductive hypothesis, and should cover all memory locations that are written to during 
+1. `while` loops and `loop` loops are supported. The [`while let` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#predicate-pattern-loops) are not supported. 
+2. [`for` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#iterator-loops) are supported for some common types: `array`, `slice`, `Iter`, `Vec`. 
+The memory predicates feature must be enabled (`-Z loop-contracts -Z memory-predicates`).
+3. Kani infers *loop modifies* with alias analysis. Loop modifies are those variables we assume to be arbitrary in the inductive hypothesis, and should cover all memory locations that are written to during 
    the execution of the loops. A proof will fail if the inferred loop modifies misses some targets written in the loops.
    We observed this happens when some fields of structs are modified by some other functions called in the loops.
-3. Kani doesn't check if a loop will always terminate in proofs with loop contracts. So it could be that some properties are proved successfully with Kani but actually are unreachable due to the 
+4. Kani doesn't check if a loop will always terminate in proofs with loop contracts. So it could be that some properties are proved successfully with Kani but actually are unreachable due to the 
    non-termination of some loops.
-4. We don't check if loop invariants are side-effect free. A loop invariant with a side effect could lead to an unsound proof result. Make sure that the specified loop contracts are side-effect free.
+5. We don't check if loop invariants are side-effect free. A loop invariant with a side effect could lead to an unsound proof result. Make sure that the specified loop contracts are side-effect free.
--- a/library/kani_core/src/iter.rs
+++ b/library/kani_core/src/iter.rs
@@ -1,19 +1,17 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! This macro generates implementations of the `Arbitrary` trait for various types. The `Arbitrary` trait defines
-//! methods for generating arbitrary (unconstrained) values of the implementing type.
-//! trivial_arbitrary and nonzero_arbitrary are implementations of Arbitrary for types that can be represented
-//! by an unconstrained symbolic value of their size (e.g., `u8`, `u16`, `u32`, etc.).
-//!
-//! TODO: Use this inside kani library so that we dont have to maintain two copies of the same proc macro for arbitrary.
+//! This macro generates implementations of the `KaniIntoIter` trait for various common types that are used in for loop.
+//! We use this trait to overwrite the Rust IntoIter trait to reduce call stacks and avoid complicated loop invariant specifications,
+//! while maintaining the semantic of the loop.
+
 
 #[macro_export]
 #[allow(clippy::crate_in_macro_def)]
 macro_rules! generate_iter {
     () => {
         use core_path::slice::Iter;
-        
+
         pub struct KaniIter<T> {
             pub iptr: *const T,
             pub len: usize,
@@ -52,7 +50,7 @@ macro_rules! generate_iter {
 
         impl<T> KaniIntoIter for Iter<'_, T> {
             type Item = T;
-            fn kani_into_iter(self) -> (*const Self::Item, usize){
+            fn kani_into_iter(self) -> (*const Self::Item, usize) {
                 (self.as_slice().as_ptr(), self.len())
             }
         }
@@ -73,11 +71,10 @@ macro_rules! generate_iter {
 
         impl<T> KaniIntoIter for Vec<T> {
             type Item = T;
-            fn kani_into_iter(self) -> (*const Self::Item, usize){
+            fn kani_into_iter(self) -> (*const Self::Item, usize) {
                 let s = self.iter();
                 (s.as_slice().as_ptr(), s.len())
             }
         }
-
     };
 }
--- a/library/kani_core/src/iter.rs
+++ b/library/kani_core/src/iter.rs
@@ -13,7 +13,7 @@
 macro_rules! generate_iter {
     () => {
         use core_path::slice::Iter;
-
+        
         pub struct KaniIter<T> {
             pub iptr: *const T,
             pub len: usize,
@@ -26,7 +26,7 @@ macro_rules! generate_iter {
             }
             pub fn next(&mut self) -> Option<&T> {
                 if self.current_pos < self.len {
-                    let elem = unsafe { &*self.iptr.offset(self.current_pos as isize) };
+                    let elem = unsafe { &*self.iptr.wrapping_add(self.current_pos) };
                     self.current_pos += 1;
                     Some(elem)
                 } else {
@@ -40,21 +40,44 @@ macro_rules! generate_iter {
             Self: Sized,
         {
             type Item;
-            fn kani_into_iter(self) -> KaniIter<Self::Item>;
+            fn kani_into_iter(self) -> (*const Self::Item, usize);
         }
 
         impl<T, const N: usize> KaniIntoIter for [T; N] {
             type Item = T;
-            fn kani_into_iter(self) -> KaniIter<Self::Item> {
-                KaniIter::new(self.as_ptr(), N)
+            fn kani_into_iter(self) -> (*const Self::Item, usize) {
+                (self.as_ptr(), N)
             }
         }
 
         impl<T> KaniIntoIter for Iter<'_, T> {
             type Item = T;
-            fn kani_into_iter(self) -> KaniIter<Self::Item> {
-                KaniIter::new(self.as_slice().as_ptr(), self.len())
+            fn kani_into_iter(self) -> (*const Self::Item, usize){
+                (self.as_slice().as_ptr(), self.len())
             }
         }
+
+        impl<'a, T> KaniIntoIter for &'a [T] {
+            type Item = T;
+            fn kani_into_iter(self) -> (*const Self::Item, usize) {
+                (self.as_ptr(), self.len())
+            }
+        }
+
+        impl<'a, T> KaniIntoIter for &'a mut [T] {
+            type Item = T;
+            fn kani_into_iter(self) -> (*const Self::Item, usize) {
+                (self.as_ptr(), self.len())
+            }
+        }
+
+        impl<T> KaniIntoIter for Vec<T> {
+            type Item = T;
+            fn kani_into_iter(self) -> (*const Self::Item, usize){
+                let s = self.iter();
+                (s.as_slice().as_ptr(), s.len())
+            }
+        }
+
     };
 }
--- a/library/kani_core/src/iter.rs
+++ b/library/kani_core/src/iter.rs
@@ -5,7 +5,6 @@
 //! We use this trait to overwrite the Rust IntoIter trait to reduce call stacks and avoid complicated loop invariant specifications,
 //! while maintaining the semantic of the loop.
 
-
 #[macro_export]
 #[allow(clippy::crate_in_macro_def)]
 macro_rules! generate_iter {
--- a/library/kani_core/src/iter.rs
+++ b/library/kani_core/src/iter.rs
@@ -1,0 +1,60 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This macro generates implementations of the `Arbitrary` trait for various types. The `Arbitrary` trait defines
+//! methods for generating arbitrary (unconstrained) values of the implementing type.
+//! trivial_arbitrary and nonzero_arbitrary are implementations of Arbitrary for types that can be represented
+//! by an unconstrained symbolic value of their size (e.g., `u8`, `u16`, `u32`, etc.).
+//!
+//! TODO: Use this inside kani library so that we dont have to maintain two copies of the same proc macro for arbitrary.
+
+#[macro_export]
+#[allow(clippy::crate_in_macro_def)]
+macro_rules! generate_iter {
+    () => {
+        use core_path::slice::Iter;
+
+        pub struct KaniIter<T> {
+            pub iptr: *const T,
+            pub len: usize,
+            pub current_pos: usize,
+        }
+
+        impl<T> KaniIter<T> {
+            pub fn new(iptr: *const T, len: usize) -> Self {
+                KaniIter { iptr, len, current_pos: 0 }
+            }
+            pub fn next(&mut self) -> Option<&T> {
+                if self.current_pos < self.len {
+                    let elem = unsafe { &*self.iptr.offset(self.current_pos as isize) };
+                    self.current_pos += 1;
+                    Some(elem)
+                } else {
+                    None
+                }
+            }
+        }
+
+        pub trait KaniIntoIter
+        where
+            Self: Sized,
+        {
+            type Item;
+            fn kani_into_iter(self) -> KaniIter<Self::Item>;
+        }
+
+        impl<T, const N: usize> KaniIntoIter for [T; N] {
+            type Item = T;
+            fn kani_into_iter(self) -> KaniIter<Self::Item> {
+                KaniIter::new(self.as_ptr(), N)
+            }
+        }
+
+        impl<T> KaniIntoIter for Iter<'_, T> {
+            type Item = T;
+            fn kani_into_iter(self) -> KaniIter<Self::Item> {
+                KaniIter::new(self.as_slice().as_ptr(), self.len())
+            }
+        }
+    };
+}
--- a/library/kani_macros/src/sysroot/loop_contracts/mod.rs
+++ b/library/kani_macros/src/sysroot/loop_contracts/mod.rs
@@ -4,14 +4,13 @@
 //! Implementation of the loop contracts code generation.
 //!
 
-use proc_macro::{Span, TokenStream};
+use proc_macro::TokenStream;
 use proc_macro_error2::abort_call_site;
 use quote::{format_ident, quote};
 use syn::spanned::Spanned;
 use syn::token::AndAnd;
 use syn::{
-    BinOp, Block, Expr, ExprBinary, ExprForLoop, ExprLoop, Ident, Pat, Stmt, parse_quote,
-    visit_mut::VisitMut,
+    BinOp, Block, Expr, ExprBinary, ExprForLoop, Ident, Stmt, parse_quote, visit_mut::VisitMut,
 };
 
 /*
@@ -227,25 +226,29 @@ fn transform_break_continue(block: &mut Block) {
     block.stmts.push(return_stmt);
 }
 
-pub fn transform_for_to_loop(for_loop: ExprForLoop) -> (Stmt, Option<Stmt>, Option<Stmt>) {
+pub fn transform_for_to_loop(
+    for_loop: ExprForLoop,
+    loop_id: &String,
+) -> (Stmt, Option<Stmt>, Option<Stmt>, Option<Stmt>, Option<Stmt>) {
     // Extract components from the for loop
     let pat = *for_loop.pat;
     let expr = for_loop.expr;
     let body = for_loop.body;
-    let attrs = for_loop.attrs;
 
     // Create an iterator variable name
-    let itername = "kaniiter_iter".to_owned();
+    let itername = "kaniiter".to_owned();
     let iter_ident = format_ident!("{}", itername);
 
-    let ptrname = "kaniiter_ptr".to_owned();
+    let mut ptrname = "kaniiter_ptr".to_owned();
+    ptrname.push_str(&loop_id);
     let ptr_ident = format_ident!("{}", ptrname);
 
-    let lenname = "kaniiter_len".to_owned();
+    let mut lenname = "kaniiter_len".to_owned();
+    lenname.push_str(&loop_id);
     let len_ident = format_ident!("{}", lenname);
 
     // Create initialization statement for the iterator
-    let init_stmt: Stmt = parse_quote! {
+    let init_ptr_stmt: Stmt = parse_quote! {
         let (#ptr_ident, #len_ident) = kani::KaniIntoIter::kani_into_iter(#expr);
     };
 
@@ -253,33 +256,30 @@ pub fn transform_for_to_loop(for_loop: ExprForLoop) -> (Stmt, Option<Stmt>, Opti
         let mut #iter_ident = 0;
     };
 
+    let init_pat_stmt: Stmt = parse_quote! {
+        let mut #pat = unsafe {*#ptr_ident};
+        //let mut #pat = 0;
+    };
+
     // Create the new loop body with iterator advancement
     let mut new_body_stmts = Vec::new();
 
-    // Add the pattern binding using next()
-    //let nextoptionname = "__nextoption".to_owned();
-    //let nextoption_ident = format_ident!("{}", nextoptionname);
-    let next_stmt: Stmt = parse_quote! {
+    let alloc_assume_stmt: Stmt = parse_quote! {
+        kani::assume(unsafe{kani::mem::is_allocated(#ptr_ident as *const (), #len_ident)});
+    };
+
+    // Increase the iter
+    let increase_iter_stmt: Stmt = parse_quote! {
         #iter_ident = #iter_ident + 1;
     };
 
-    //let check_stmt: Stmt = parse_quote! {
-    //    if #nextoption_ident.is_none() { break; };
-    //};
-
-    let next_ident = match pat {
-        Pat::Ident(ref patident) => patident.ident.clone(),
-        _ => abort_call_site!("Unsupported pattern in for loop"),
+    let pat_assign_stmt: Stmt = parse_quote! {
+        #pat = unsafe {*#ptr_ident.wrapping_add(#iter_ident)};
     };
 
-    let pat_assign: Stmt = parse_quote! {
-        let #pat = unsafe {*#ptr_ident.wrapping_add(#iter_ident)};
-    };
-
-    new_body_stmts.push(pat_assign);
-    new_body_stmts.push(next_stmt);
-    //new_body_stmts.push(check_stmt);
-    //new_body_stmts.push(next_unwrap_stmt);
+    new_body_stmts.push(alloc_assume_stmt.clone());
+    new_body_stmts.push(pat_assign_stmt);
+    new_body_stmts.push(increase_iter_stmt);
 
     // Add the original loop body statements
     new_body_stmts.extend(body.stmts.iter().cloned());
@@ -290,18 +290,21 @@ pub fn transform_for_to_loop(for_loop: ExprForLoop) -> (Stmt, Option<Stmt>, Opti
                 #(#new_body_stmts)*
             }
     };
-    (loop_loop, Some(init_stmt), Some(init_iter_stmt))
+    (loop_loop, Some(init_ptr_stmt), Some(init_iter_stmt), Some(init_pat_stmt), Some(alloc_assume_stmt))
 }
 
 pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
     // parse the stmt of the loop
     let mut loop_stmt: Stmt = syn::parse(item.clone()).unwrap();
     let loop_id = generate_unique_id_from_span(&loop_stmt);
-    let mut initstmt: Option<Stmt> = None;
-    let mut init_iter_stmto: Option<Stmt> = None;
+    let mut init_ptr_stmt: Option<Stmt> = None;
+    let mut init_iter_stmt: Option<Stmt> = None;
+    let mut init_pat_stmt: Option<Stmt> = None;
+    let mut alloc_assume_stmt: Option<Stmt> = None;
     if let Stmt::Expr(ref e, _) = loop_stmt {
         if let Expr::ForLoop(for_loop) = e {
-            (loop_stmt, initstmt, init_iter_stmto) = transform_for_to_loop(for_loop.clone());
+            (loop_stmt, init_ptr_stmt, init_iter_stmt, init_pat_stmt, alloc_assume_stmt) =
+                transform_for_to_loop(for_loop.clone(), &loop_id);
         }
     }
 
@@ -400,7 +403,7 @@ pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
             note = "for now, loop contracts is only supported for while-loops.";
         ),
     }
-    let ret: TokenStream = if initstmt.is_none() {
+    let ret: TokenStream = if init_ptr_stmt.is_none() {
         if has_prev {
             quote!(
             {
@@ -447,15 +450,19 @@ pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
             .into()
         }
     } else {
-        let inititer = initstmt.unwrap();
-        let init_iter_stmt = init_iter_stmto.unwrap();
+        let init_ptr_stmt = init_ptr_stmt.unwrap();
+        let init_iter_stmt = init_iter_stmt.unwrap();
+        let init_pat_stmt = init_pat_stmt.unwrap();
+        let alloc_assume_stmt = alloc_assume_stmt.unwrap();
         if has_prev {
             quote!(
             {
             #(#onentry_decl_stms)*
             #(#prev_decl_stms)*
             #init_iter_stmt
-            #inititer
+            #init_ptr_stmt
+            #alloc_assume_stmt
+            #init_pat_stmt
             let mut #loop_body_closure = ||
             #loop_body;
             let (#loop_body_closure_ret_1, #loop_body_closure_ret_2) = #loop_body_closure ();
@@ -491,7 +498,9 @@ pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
                 true
             }
             #init_iter_stmt
-            #inititer
+            #init_ptr_stmt
+            #alloc_assume_stmt
+            #init_pat_stmt
             #loop_stmt
             })
             .into()
--- a/library/kani_macros/src/sysroot/loop_contracts/mod.rs
+++ b/library/kani_macros/src/sysroot/loop_contracts/mod.rs
@@ -290,7 +290,13 @@ pub fn transform_for_to_loop(
                 #(#new_body_stmts)*
             }
     };
-    (loop_loop, Some(init_ptr_stmt), Some(init_iter_stmt), Some(init_pat_stmt), Some(alloc_assume_stmt))
+    (
+        loop_loop,
+        Some(init_ptr_stmt),
+        Some(init_iter_stmt),
+        Some(init_pat_stmt),
+        Some(alloc_assume_stmt),
+    )
 }
 
 pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
@@ -506,7 +512,6 @@ pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
             .into()
         }
     };
-    println!("{}", ret.to_string());
     ret
 }
 
--- a/library/kani_macros/src/sysroot/loop_contracts/mod.rs
+++ b/library/kani_macros/src/sysroot/loop_contracts/mod.rs
@@ -4,12 +4,15 @@
 //! Implementation of the loop contracts code generation.
 //!
 
-use proc_macro::TokenStream;
+use proc_macro::{Span, TokenStream};
 use proc_macro_error2::abort_call_site;
 use quote::{format_ident, quote};
 use syn::spanned::Spanned;
 use syn::token::AndAnd;
-use syn::{BinOp, Block, Expr, ExprBinary, Ident, Stmt, parse_quote, visit_mut::VisitMut};
+use syn::{
+    BinOp, Block, Expr, ExprBinary, ExprForLoop, ExprLoop, Ident, Pat, Stmt, parse_quote,
+    visit_mut::VisitMut,
+};
 
 /*
     Transform the loop to support on_entry(expr) : the value of expr before entering the loop
@@ -224,14 +227,75 @@ fn transform_break_continue(block: &mut Block) {
     block.stmts.push(return_stmt);
 }
 
+pub fn transform_for_to_loop(for_loop: ExprForLoop) -> (Stmt, Option<Stmt>) {
+    // Extract components from the for loop
+    let pat = *for_loop.pat;
+    let expr = for_loop.expr;
+    let body = for_loop.body;
+    let attrs = for_loop.attrs;
+
+    // Create an iterator variable name
+    let itername = "kaniiter".to_owned();
+    let iter_ident = format_ident!("{}", itername);
+
+    // Create initialization statement for the iterator
+    let init_stmt: Stmt = parse_quote! {
+        let mut #iter_ident = kani::KaniIntoIter::kani_into_iter(#expr);
+    };
+
+    // Create the new loop body with iterator advancement
+    let mut new_body_stmts = Vec::new();
+
+    // Add the pattern binding using next()
+    let nextoptionname = "__nextoption".to_owned();
+    let nextoption_ident = format_ident!("{}", nextoptionname);
+    let next_stmt: Stmt = parse_quote! {
+        let #nextoption_ident = #iter_ident.next();
+    };
+
+    let check_stmt: Stmt = parse_quote! {
+        if #nextoption_ident.is_none() { break; };
+    };
+
+    let next_ident = match pat {
+        Pat::Ident(ref patident) => patident.ident.clone(),
+        _ => abort_call_site!("Unsupported pattern in for loop"),
+    };
+
+    let next_unwrap_stmt: Stmt = parse_quote! {
+        let #pat = #nextoption_ident.unwrap();
+    };
+
+    new_body_stmts.push(next_stmt);
+    new_body_stmts.push(check_stmt);
+    new_body_stmts.push(next_unwrap_stmt);
+
+    // Add the original loop body statements
+    new_body_stmts.extend(body.stmts.iter().cloned());
+
+    // Create the final expression with the iterator initialization
+    let loop_loop: Stmt = parse_quote! {
+            loop{
+                #(#new_body_stmts)*
+            }
+    };
+    (loop_loop, Some(init_stmt))
+}
+
 pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
     // parse the stmt of the loop
     let mut loop_stmt: Stmt = syn::parse(item.clone()).unwrap();
+    let loop_id = generate_unique_id_from_span(&loop_stmt);
+    let mut initstmt: Option<Stmt> = None;
+    if let Stmt::Expr(ref e, _) = loop_stmt {
+        if let Expr::ForLoop(for_loop) = e {
+            (loop_stmt, initstmt) = transform_for_to_loop(for_loop.clone());
+        }
+    }
 
     // name of the loop invariant as closure of the form
     // __kani_loop_invariant_#startline_#startcol_#endline_#endcol
     let mut inv_name: String = "__kani_loop_invariant".to_owned();
-    let loop_id = generate_unique_id_from_span(&loop_stmt);
     inv_name.push_str(&loop_id);
 
     // expr of the loop invariant
@@ -324,52 +388,102 @@ pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
             note = "for now, loop contracts is only supported for while-loops.";
         ),
     }
-
-    if has_prev {
-        quote!(
-        {
-        if (#loop_guard) {
-        #(#onentry_decl_stms)*
-        #(#prev_decl_stms)*
-        let mut #loop_body_closure = ||
-        #loop_body;
-        let (#loop_body_closure_ret_1, #loop_body_closure_ret_2) = #loop_body_closure ();
-        if #loop_body_closure_ret_2.is_some() {
-            return #loop_body_closure_ret_2.unwrap();
-        }
-        if #loop_body_closure_ret_1 {
-        // Dummy function used to force the compiler to capture the environment.
-        // We cannot call closures inside constant functions.
-        // This function gets replaced by `kani::internal::call_closure`.
+    let ret: TokenStream = if initstmt.is_none() {
+        if has_prev {
+            quote!(
+            {
+            if (#loop_guard) {
+            #(#onentry_decl_stms)*
+            #(#prev_decl_stms)*
+            let mut #loop_body_closure = ||
+            #loop_body;
+            let (#loop_body_closure_ret_1, #loop_body_closure_ret_2) = #loop_body_closure ();
+            if #loop_body_closure_ret_2.is_some() {
+                return #loop_body_closure_ret_2.unwrap();
+            }
+            if #loop_body_closure_ret_1 {
+            // Dummy function used to force the compiler to capture the environment.
+            // We cannot call closures inside constant functions.
+            // This function gets replaced by `kani::internal::call_closure`.
+                #[inline(never)]
+                #[kanitool::fn_marker = "kani_register_loop_contract"]
+                const fn #register_ident<F: Fn() -> bool>(_f: &F, _transformed: usize) -> bool {
+                    true
+                }
+                #loop_stmt
+            }
+            else {
+                assert!(#inv_expr);
+            };
+            }
+            })
+            .into()
+        } else {
+            quote!(
+            {
+            #(#onentry_decl_stms)*
+            // Dummy function used to force the compiler to capture the environment.
+            // We cannot call closures inside constant functions.
+            // This function gets replaced by `kani::internal::call_closure`.
             #[inline(never)]
             #[kanitool::fn_marker = "kani_register_loop_contract"]
             const fn #register_ident<F: Fn() -> bool>(_f: &F, _transformed: usize) -> bool {
                 true
             }
             #loop_stmt
+            })
+            .into()
         }
-        else {
-            assert!(#inv_expr);
-        };
-        }
-        })
-        .into()
     } else {
-        quote!(
-        {
-        #(#onentry_decl_stms)*
-        // Dummy function used to force the compiler to capture the environment.
-        // We cannot call closures inside constant functions.
-        // This function gets replaced by `kani::internal::call_closure`.
-        #[inline(never)]
-        #[kanitool::fn_marker = "kani_register_loop_contract"]
-        const fn #register_ident<F: Fn() -> bool>(_f: &F, _transformed: usize) -> bool {
-            true
+        let inititer = initstmt.unwrap();
+        if has_prev {
+            quote!(
+            {
+            #(#onentry_decl_stms)*
+            #(#prev_decl_stms)*
+            #inititer
+            let mut #loop_body_closure = ||
+            #loop_body;
+            let (#loop_body_closure_ret_1, #loop_body_closure_ret_2) = #loop_body_closure ();
+            if #loop_body_closure_ret_2.is_some() {
+                return #loop_body_closure_ret_2.unwrap();
+            }
+            if #loop_body_closure_ret_1 {
+            // Dummy function used to force the compiler to capture the environment.
+            // We cannot call closures inside constant functions.
+            // This function gets replaced by `kani::internal::call_closure`.
+                #[inline(never)]
+                #[kanitool::fn_marker = "kani_register_loop_contract"]
+                const fn #register_ident<F: Fn() -> bool>(_f: &F, _transformed: usize) -> bool {
+                    true
+                }
+                #loop_stmt
+            }
+            else {
+                assert!(#inv_expr);
+            };
+            })
+            .into()
+        } else {
+            quote!(
+            {
+            #(#onentry_decl_stms)*
+            // Dummy function used to force the compiler to capture the environment.
+            // We cannot call closures inside constant functions.
+            // This function gets replaced by `kani::internal::call_closure`.
+            #[inline(never)]
+            #[kanitool::fn_marker = "kani_register_loop_contract"]
+            const fn #register_ident<F: Fn() -> bool>(_f: &F, _transformed: usize) -> bool {
+                true
+            }
+            #inititer
+            #loop_stmt
+            })
+            .into()
         }
-        #loop_stmt
-        })
-        .into()
-    }
+    };
+    println!("{}", ret.to_string());
+    ret
 }
 
 fn generate_unique_id_from_span(stmt: &Stmt) -> String {
--- a/library/kani_macros/src/sysroot/loop_contracts/mod.rs
+++ b/library/kani_macros/src/sysroot/loop_contracts/mod.rs
@@ -227,7 +227,7 @@ fn transform_break_continue(block: &mut Block) {
     block.stmts.push(return_stmt);
 }
 
-pub fn transform_for_to_loop(for_loop: ExprForLoop) -> (Stmt, Option<Stmt>) {
+pub fn transform_for_to_loop(for_loop: ExprForLoop) -> (Stmt, Option<Stmt>, Option<Stmt>) {
     // Extract components from the for loop
     let pat = *for_loop.pat;
     let expr = for_loop.expr;
@@ -235,51 +235,62 @@ pub fn transform_for_to_loop(for_loop: ExprForLoop) -> (Stmt, Option<Stmt>) {
     let attrs = for_loop.attrs;
 
     // Create an iterator variable name
-    let itername = "kaniiter".to_owned();
+    let itername = "kaniiter_iter".to_owned();
     let iter_ident = format_ident!("{}", itername);
+
+    let ptrname = "kaniiter_ptr".to_owned();
+    let ptr_ident = format_ident!("{}", ptrname);
+
+    let lenname = "kaniiter_len".to_owned();
+    let len_ident = format_ident!("{}", lenname);
 
     // Create initialization statement for the iterator
     let init_stmt: Stmt = parse_quote! {
-        let mut #iter_ident = kani::KaniIntoIter::kani_into_iter(#expr);
+        let (#ptr_ident, #len_ident) = kani::KaniIntoIter::kani_into_iter(#expr);
+    };
+
+    let init_iter_stmt: Stmt = parse_quote! {
+        let mut #iter_ident = 0;
     };
 
     // Create the new loop body with iterator advancement
     let mut new_body_stmts = Vec::new();
 
     // Add the pattern binding using next()
-    let nextoptionname = "__nextoption".to_owned();
-    let nextoption_ident = format_ident!("{}", nextoptionname);
+    //let nextoptionname = "__nextoption".to_owned();
+    //let nextoption_ident = format_ident!("{}", nextoptionname);
     let next_stmt: Stmt = parse_quote! {
-        let #nextoption_ident = #iter_ident.next();
+        #iter_ident = #iter_ident + 1;
     };
 
-    let check_stmt: Stmt = parse_quote! {
-        if #nextoption_ident.is_none() { break; };
-    };
+    //let check_stmt: Stmt = parse_quote! {
+    //    if #nextoption_ident.is_none() { break; };
+    //};
 
     let next_ident = match pat {
         Pat::Ident(ref patident) => patident.ident.clone(),
         _ => abort_call_site!("Unsupported pattern in for loop"),
     };
 
-    let next_unwrap_stmt: Stmt = parse_quote! {
-        let #pat = #nextoption_ident.unwrap();
+    let pat_assign: Stmt = parse_quote! {
+        let #pat = unsafe {*#ptr_ident.wrapping_add(#iter_ident)};
     };
 
+    new_body_stmts.push(pat_assign);
     new_body_stmts.push(next_stmt);
-    new_body_stmts.push(check_stmt);
-    new_body_stmts.push(next_unwrap_stmt);
+    //new_body_stmts.push(check_stmt);
+    //new_body_stmts.push(next_unwrap_stmt);
 
     // Add the original loop body statements
     new_body_stmts.extend(body.stmts.iter().cloned());
 
     // Create the final expression with the iterator initialization
     let loop_loop: Stmt = parse_quote! {
-            loop{
+            while (#iter_ident < #len_ident) {
                 #(#new_body_stmts)*
             }
     };
-    (loop_loop, Some(init_stmt))
+    (loop_loop, Some(init_stmt), Some(init_iter_stmt))
 }
 
 pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
@@ -287,9 +298,10 @@ pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut loop_stmt: Stmt = syn::parse(item.clone()).unwrap();
     let loop_id = generate_unique_id_from_span(&loop_stmt);
     let mut initstmt: Option<Stmt> = None;
+    let mut init_iter_stmto: Option<Stmt> = None;
     if let Stmt::Expr(ref e, _) = loop_stmt {
         if let Expr::ForLoop(for_loop) = e {
-            (loop_stmt, initstmt) = transform_for_to_loop(for_loop.clone());
+            (loop_stmt, initstmt, init_iter_stmto) = transform_for_to_loop(for_loop.clone());
         }
     }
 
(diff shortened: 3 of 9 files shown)


library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -230,14 +234,142 @@ fn transform_break_continue(block: &mut Block) {
     block.stmts.push(return_stmt);
 }
 
+fn make_pat_tuple_mutable(pat: &Pat) -> Pat {
+    match pat {
```
carolynzech: Please add some doc comments to this function explaining what it's doing.

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -230,14 +234,142 @@ fn transform_break_continue(block: &mut Block) {
...
+            let mut mut_ident = pat_ident.clone();
+            mut_ident.mutability = Some(Token![mut](Span::call_site()));
+            Pat::Ident(mut_ident)
+        }
+        // If not a tuple, return unchanged
+        _ => pat.clone(),
+    }
+}
+
+pub struct ForLoopExtraStmts {
```
carolynzech, line 295: doc comment

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -230,14 +234,142 @@ fn transform_break_continue(block: &mut Block) {
...
+    init_len_stmt: Stmt,
+    init_index_stmt: Stmt,
+    init_pat_stmt: Stmt,
+    alloc_assume_stmt: Stmt,
+}
+
+pub fn transform_for_to_loop(
+    for_loop: ExprForLoop,
+    loop_id: &str,
+) -> (Stmt, Option<ForLoopExtraStmts>) {
```
carolynzech, line 309: doc comment

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -330,8 +461,72 @@ pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
             note = "for now, loop contracts is only supported for while-loops.";
         ),
     }
-
-    if has_prev {
+    let ret: TokenStream = if let Some(ForLoopExtraStmts {
+        init_iter_stmt,
```
carolynzech, line 579: comment

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,331 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This macro generates implementations of the `KaniIntoIter` trait for various common types that are used in for loop.
+//! We use this trait to overwrite the Rust IntoIter trait to reduce call stacks and avoid complicated loop invariant specifications,
+//! while maintaining the semantic of the loop.
```
carolynzech:
```suggestion
//! This macro generates implementations of the `KaniIntoIter` trait for various common types that are used in `for` loops.
//! We use this trait to overwrite the Rust IntoIter trait to reduce call stacks and avoid complicated loop invariant specifications,
//! while maintaining the semantics of the loop.
```

library/kani/src/iter.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This macro generates implementations of the `KaniIntoIter` trait for various common types that are used in for loop.
+//! We use this trait to overwrite the Rust IntoIter trait to reduce call stacks and avoid complicated loop invariant specifications,
+//! while maintaining the semantic of the loop.
```
carolynzech:
```suggestion
//! This macro generates implementations of the `KaniIntoIter` trait for various common types that are used in `for` loops.
//! We use this trait to overwrite the Rust IntoIter trait to reduce call stacks and avoid complicated loop invariant specifications,
//! while maintaining the semantics of the loop.
```

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -230,14 +234,142 @@ fn transform_break_continue(block: &mut Block) {
...
+}
+
 pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
     // parse the stmt of the loop
     let mut loop_stmt: Stmt = syn::parse(item.clone()).unwrap();
-
+    let loop_id = generate_unique_id_from_span(&loop_stmt);
+    let mut forloopextras: Option<ForLoopExtraStmts> = None;
+    if let Stmt::Expr(Expr::ForLoop(for_loop), _) = &loop_stmt {
+        (loop_stmt, forloopextras) = transform_for_to_loop(for_loop.clone(), &loop_id);
```
carolynzech:
```suggestion
    let mut for_loop_extras: Option<ForLoopExtraStmts> = None;
    if let Stmt::Expr(Expr::ForLoop(for_loop), _) = &loop_stmt {
        (loop_stmt, for_loop_extras) = transform_for_to_loop(for_loop.clone(), &loop_id);
```

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -230,14 +234,142 @@ fn transform_break_continue(block: &mut Block) {
...
+    };
+    let forloopextras = ForLoopExtraStmts {
+        init_iter_stmt,
+        init_len_stmt,
+        init_index_stmt,
+        init_pat_stmt,
+        alloc_assume_stmt,
+    };
+
+    (loop_loop, Some(forloopextras))
```
carolynzech:
```suggestion
    let for_loop_extras = ForLoopExtraStmts {
        init_iter_stmt,
        init_len_stmt,
        init_index_stmt,
        init_pat_stmt,
        alloc_assume_stmt,
    };

    (loop_loop, Some(for_loop_extras))
```

tests/expected/loop-contract/for_loop_with_quantifier.rs
```diff
@@ -0,0 +1,22 @@
...
+#![feature(proc_macro_hygiene)]
+#![feature(stmt_expr_attributes)]
+
+#[kani::proof]
+fn forloop() {
+    let mut sum: u32 = 0;
+    let a: [u8; 10] = kani::any();
+    kani::assume(kani::forall!(|i in (0,10)| a[i] <= 10));
+
+    #[kani::loop_invariant( kaniindex <= 10 && sum <= (kaniindex as u32 * 10) )]
```
rod-chapman: What is "kaniindex" here? Is that supposed to be connected to the value of the loop counter "i"? If so, why not just write "i"?

zhassan-aws:
Can you expand the PR description and/or the comments in `library/kani_macros/src/sysroot/loop_contracts/mod.rs` to explain the idea behind the implementation? It is hard to infer what this is doing just through reading the code.

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -168,7 +168,7 @@ Kani automatically contracts (instead of unwinds) all loops in the functions tha
 
 Loop contracts comes with the following limitations.
 
-1. `while` loops and `loop` loops are supported. The other kinds of loops are not supported: [`while let` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#predicate-pattern-loops), and [`for` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#iterator-loops).
+1. `while` loops, `loop` loops are supported. `for` loops are supported for array, slice, Iter, Vec, Range, StepBy, Chain, Zip, Map, and Enumerate. The other kinds of loops are not supported: [`while let` loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#predicate-pattern-loops).
```
rod-chapman, line 283: You also need to document:
1. The use and meaning of the "kaniindex" name.
2. How iter().enumerate() can be used to get the index number of a for loop so it can be used in an invariant. Please illustrate this with a full example.
3. Please give a full example of a nested loop with invariants - for example, forming the sum of a two-dimemsional array of integers, and proving the upper-bound of the result.
4. Give an example of how to an iterator is applied with a for loop to iterate over a _dynamic_ slice (i.e. a slice whose bounds are not known at compile time which is passed into a function as a formal parameter).

rod-chapman: As you write the examples, add them as new tests cases obviously.

rod-chapman: I would also like to see test cases where functions are declared inside a module with full pre/post-conditions contracts, rather than the "assume/assert test harness" style. We'll be using the former style in Crypto work, so you need to convince me that it really does work that way.

thanhnguyen-aws: Hello, thank you very much for your suggestion.
1. I updated the document (loop-contracts.md file), which includes the kaniindex.
2. You can find the example in this file too. 
3. The nested loop for `for` loop needs one extra component: the loop_assign clause because sometimes CBMC cannot infer it for the inner loop. I am implementing it and expect to finish next week. I will add that example after both PRs are merged.
4. Currently, Kani does not support fully unbounded slices, but it supports slices of arbitrary length. Please take a look at the file for_loop_for_slice, we consider a slice (`s`) unbounded when we set the length of `a` to be `isize::MAX`. However, until array theory is supported (we are working on it), the performance is unscalable w.r.t array size (slice length).
5. You can find examples of function-contracts in tests/expected/function-contracts.

rod-chapman: Iteration and quantification over unbounded slices will be required for crypto work (e.g. where we're computing a function over a potentially very large array of bytes, such as a user-supplied message.)

I do not see the need for a scaling problem. The proofs can be generated in a fashion that is entirely symbolic with respect to the bounds of the slice. This is how it works in tools like Frama-C and SPARK Pro.

rod-chapman: If I have a slice parameter - something like
  fn f(a : &[u8]) -> i32
but I have a pre-condition something like
  a.len <= 100

Why is that a problem?

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -230,18 +232,147 @@ fn transform_break_continue(block: &mut Block) {
...
+    for_loop: ExprForLoop,
+    loop_id: &str,
+) -> (Stmt, Option<ForLoopExtraStmts>) {
+    // Extract components from the for loop
+    let pat = *for_loop.pat;
+    let expr = for_loop.expr;
+    let body = for_loop.body;
+
+    // Create an iterator variable name
+    let indexname = "kaniindex".to_owned();
```
tautschnig: Does this approach still work when you have nested loops?

thanhnguyen-aws: Yes, in the macro, I put an extra `{}` surrounding the loop and those extra statements, so those variables will be dropped after the loop. Just for demonstration, the generated code would look like:
```rust
let kanindex = 0;
#[kani::loop_invariant(kaniindex < 5)
for a in as {
...
   { //this one
     let kaniindex = 0;  
     #[kani::loop_invariant(kaniindex < 6)
     for b in bs { .....}
   } //this one
 
}
```

Like I mentioned in the doc: Note that the two extra variables above are only associated with the `for` loop right below the loop-invariant.

tautschnig: Can you please include a regression test demonstrating this?

thanhnguyen-aws: I added 2 of them

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -105,7 +107,7 @@ Complete - 1 successfully verified harnesses, 0 failures, 1 total.
 An invariant contract `#[kani::loop_invariant(cond)]` accepts a valid Boolean expression `cond` over the variables visible at the same scope as the loop.
```
carolynzech, line 108: Can you expand the syntax section to indicate that it also works for `for` loops now?

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -114,33 +116,81 @@ Mathematical induction is the working principle here. (1) establishes the base c
...
-`!guard && inv`, which is `x <= 1 && x >= 1` in the example. The postcondition implies `x == 1`—the property we want to prove at the end of `simple_loop_with_loop_contracts`.
+`!guard && inv`, which is `x <= 1 && x >= 1` in the example. The postcondition implies `x == 1`—the property we want to prove at the end of `simple_loop_with_loop_invariants`.
+
+## Historic values and extra variables
+
+### Historic values
+
+We support two notations for historic values in loop-invariants:
+1. `on_entry(expr)` : The value of the `expr` before entering the loop.
+2. `prev(expr)` : the value of `expr` in the previous iteration. Note that Kani will assert that the loop has at least one iteration of `prev` is used in loop-invariants.
```
carolynzech:
```suggestion
2. `prev(expr)` : the value of `expr` in the previous iteration. Note that Kani will assert that the loop has at least one iteration if `prev` is used in loop-invariants.
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -114,33 +116,81 @@ Mathematical induction is the working principle here. (1) establishes the base c
...
+    }
+    assert!(i == 2);
+}
+```
+
+### Extra variables in `for` loop
+
+There are two extra variables that can be used in loop-invariants for `for` loops:
+1. `kaniindex` : the position (index) of the current iteration in the iterator.
+2. `kaniiterlen` : the length of the iterator.
```
carolynzech: What is the point of exposing these values? I don't find the example compelling; I feel like if people want the position of the iterator they should just call `enumerate` and write assertions about `i` instead of us exposing a value?
Same goes for the length of the iterator; what's an example of something we might want to express but couldn't without exposing this value?

carolynzech: Also, what happens if someone tries to use `kaniindex` or `kaniiterlen` as local variables in their loop, like `let kaniindex = ...`?

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,387 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This macro generates implementations of the `KaniIntoIter` trait for various common types that are used in `for` loops.
+//! We use this trait to overwrite the Rust IntoIter trait to reduce call stacks and avoid complicated loop invariant specifications,
+//! while maintaining the semantics of the loop.
+
```
carolynzech, line 48: Can you expand on this comment more so people understand why we need to overwrite these traits? What does "complicated" mean here?

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,387 @@
...
+            type Item = &'a T;
+            fn indexing(&self, i: usize) -> Self::Item {
+                unsafe { &*self.ptr.wrapping_add(i) }
+            }
+            fn first(&self) -> Self::Item {
+                unsafe { &*self.ptr }
+            }
+            fn assumption(&self) -> bool {
+                unsafe { mem::is_allocated(self.ptr as *const (), self.len) }
+            }
```
carolynzech: Can you write `// SAFETY` comments above all of the unsafe code to explain why it's safe, c.f. the [safety docs](https://doc.rust-lang.org/book/ch20-01-unsafe-rust.html):

> Whenever we write an unsafe function, it is idiomatic to write a comment starting with SAFETY and explaining what the caller needs to do to call the function safely. Likewise, whenever we perform an unsafe operation, it is idiomatic to write a comment starting with SAFETY to explain how the safety rules are upheld.

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,387 @@
...
+macro_rules! generate_iter {
+    () => {
+        use core::iter::StepBy;
+        use core_path::cmp::min;
+        use core_path::iter::{Chain, Enumerate, Map, Zip};
+        use core_path::mem as stdmem;
+        use core_path::ops::Range;
+        use core_path::slice::Iter;
+
+        pub trait KaniIter
```
carolynzech, line 60: Can you put doc comments throughout this macro describing the purpose of each trait and the method in each trait? I'm having trouble telling if these implementations are correct when I'm not sure of the purpose of each thing.

docs/src/reference/experimental/loop-contracts.md
carolynzech: @thanhnguyen-aws can you add @rod-chapman's example program here as a test (or a fixme test if it doesn't work?)

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
 On the CBMC site, `goto-instrument` will extract the map and instrument the customized
 havocing functions for the modifies targets.
 
+### For-loop rewrite
```
carolynzech, line 217: I just noticed this RFC has a section on `loop-decreases` that seems to imply that decreases will always be required to prove termination, but we've written plenty of loops without it. Can you update that text to reflect the current implementation? Can be in a separate PR if you prefer. Also the status field at the top should be changed to unstable.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
 On the CBMC site, `goto-instrument` will extract the map and instrument the customized
 havocing functions for the modifies targets.
 
+### For-loop rewrite 
+
+When there is loop-invariant for `for-loop`, the `loop_invariant` procedural macros
```
carolynzech: This is a really nice explanation; thank you for writing it!

library/kani_core/src/iter.rs
carolynzech: Coming back after reading your refined RFC text; I'd be happy with 1-2 sentences to the effect of "if we just take the `for` loop as-is we end up havocing a lot of extra stuff and encountering verification failures; see the RFC (rfc link) for more details."

docs/src/reference/experimental/loop-contracts.md
thanhnguyen-aws: 1. We don't want to change user code in order to write a loop invariant. kaniiter and kaniindex are useful variables for loop-invariant, so I think it is helpful to provide them to the users. In case the users use enumerate, the index of the Enumerate Iterator is also different from kaniindex as only kaniindex is updated at the end of the loop body as demonstrated in [nested_forloop_for_enumerate.rs].  
2. All extra variables are put together with a loop inside a block {} (by the proc macro) , so if the user defines variables with similar names, the loop-invariant always takes the ones that is generated by kani. I don't think user may name their variables that way unless they want to "hack" kani.

docs/src/reference/experimental/loop-contracts.md
thanhnguyen-aws: @carolynzech Which program are you referring to?

carolynzech: This one: https://github.com/model-checking/kani/pull/4143#discussion_r2157399761

rfc/src/rfcs/0012-loop-contracts.md
thanhnguyen-aws: `loop-decrease` is not supported yet. As mentioned `Limitation` section of loop-contract.md : "Kani doesn't check if a loop will always terminate in proofs with loop contracts. So it could be that some properties are proved successfully with Kani but actually are unreachable due to the non-termination of some loops." .

tautschnig:
The PR description says "`for` loop for 4 types: array, slice, Iter, Vec, Range, StepBy, Map, Chain, Zip, Enumerate." -- that's 10 items enumerated here, not 4?

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+    data: DATA,
+}
+// and so on ...
+```
+
+In order to contract the loop, the user must provide the loop-invariant for the variable `kani_iter`, which is modified and 
+will be havoc. 
+The loop-invariant should include some propositions such as `kani_iter.inner.alive.start <= kani_iter.inner.alive.end`,
+`kani_iter.inner.data.as_ptr() == a.as_ptr()`, ... until the loop-invariant is successfully verified.
+The obvious tree disadvantages are:
```
tautschnig:
```suggestion
The obvious three disadvantages are:
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+1. All the fields `inner`, `alive`, `start`, `end`, ... are private, so those propositions are not allowed by the Rust compiler.
+2. The users must know about the all implementation designs and related internal types that Rust standard library uses
```
tautschnig:
```suggestion
2. The users must know about all the implementation designs and related internal types that the Rust standard library uses 
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+`kani_iter.inner.data.as_ptr() == a.as_ptr()`, ... until the loop-invariant is successfully verified.
+The obvious tree disadvantages are:
+1. All the fields `inner`, `alive`, `start`, `end`, ... are private, so those propositions are not allowed by the Rust compiler.
+2. The users must know about the all implementation designs and related internal types that Rust standard library uses 
+to implement the `into_iter` function, and everything are different for the types that implements the `IntoIter` trait. 
+Even then, it is so complicated and tedious to write the loop-invariant. 
+3. The long call stack will slow down the performance.
+
+To avoid those disadvantages, we overide the `into_iter` function and `IntoIter` trait by our own version (`kani_into_iter()` and `KaniIntoIter`) 
+with simplified implementation but keeping the same semantic.
```
tautschnig:
```suggestion
with simplified implementation but keeping the same semantics.
```

tautschnig (commented):
Looks good to me!

docs/src/reference/experimental/loop-contracts.md
zhassan-aws: I tend to agree with @carolynzech here. It is customary to modify code for verification purposes, e.g. through adding "ghost" variables that are only used in invariants. I would vote for not exposing those variables.

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -1,13 +1,15 @@
-# Loop Contracts
+# Loop Invariants
```
zhassan-aws: Why the name change?

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -1,13 +1,15 @@
...
 
-Loop contract are used to specify invariants for loops for the sake of extending Kani's *bounded proofs* to *unbounded proofs*.
-A [loop invariant](https://en.wikipedia.org/wiki/Loop_invariant) is an expression that holds upon entering a loop and after every execution of the loop body.
+## Introduction
+
+Loop-invariants are used to specify invariants for loops for the sake of extending Kani's *bounded proofs* to *unbounded proofs*.
```
zhassan-aws:
```suggestion
Loop invariants are used to specify invariants for loops for the sake of extending Kani's *bounded proofs* to *unbounded proofs*.
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -1,13 +1,15 @@
...
+A [loop-invariant](https://en.wikipedia.org/wiki/Loop_invariant) is an expression that holds upon entering a loop and after every execution of the loop body.
 It captures something that does not change about every step of the loop.
 
 It is worth revisiting the discussion about [bounded proof](../../tutorial-loop-unwinding.md#bounded-proof) and
 [loop unwinding](../../tutorial-loop-unwinding.md#loops-unwinding-and-bounds). In short, bounds on the number of times Kani unwinds loops also bound the size of inputs,
 and hence result in a bounded proof.
-Loop contracts are used to abstract out loops as non-loop blocks to avoid loop unwinding, and hence remove the bounds on the inputs.
+Loop-invariants are used to abstract out loops as non-loop blocks to avoid loop unwinding, and hence remove the bounds on the inputs.
```
zhassan-aws: Please replace all occurrences of "Loop-invariants" with "Loop invariants" (i.e. remove the dash).

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -24,15 +26,15 @@ fn simple_loop() {
 ```
 
 In this program, the loop repeatedly decrements `x` until it equals `1`. Because we haven't specified an upper bound for `x`, to verify this function,
-Kani needs to unwind the loop for `u64::MAX` iterations, which is computationally expensive. Loop contracts allow us to abstract the loop behavior, significantly reducing the verification cost.
+Kani needs to unwind the loop for `u64::MAX` iterations, which is computationally expensive. Loop invariants allow us to abstract the loop behavior, significantly reducing the verification cost.
```
zhassan-aws:
```suggestion
Kani needs to unwind the loop for `u64::MAX` iterations, which is intractable. Loop invariants allow us to abstract the loop behavior, significantly reducing the verification cost.
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -44,42 +46,42 @@ fn simple_loop_with_loop_contracts() {
...
-Check 10: simple_loop_with_loop_contracts.loop_invariant_base.1
+Check 10: simple_loop_with_loop_invariants.loop_invariant_base.1
          - Status: SUCCESS
-         - Description: "Check invariant before entry for loop simple_loop_with_loop_contracts.0"
-         - Location: simple_while_loop.rs:15:5 in function simple_loop_with_loop_contracts
+         - Description: "Check invariant before entry for loop simple_loop_with_loop_invariants.0"
+         - Location: simple_while_loop.rs:15:5 in function simple_loop_with_loop_invariants
 
-Check 11: simple_loop_with_loop_contracts.loop_modifies.1
+Check 11: simple_loop_with_loop_invariants.loop_assigns.1
```
zhassan-aws: Was `loop_modifies` renamed to `loop_assigns` in this PR? I'm not clear where this is done.

docs/src/reference/experimental/loop-contracts.md
zhassan-aws: If the renaming is something that was agreed on, it should be done in a separate PR. Adding support for "for" loops already requires so many changes, so we should avoid convoluting it with other unrelated changes.

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -114,33 +116,81 @@ Mathematical induction is the working principle here. (1) establishes the base c
...
 Note that all assertions after `kani::assume(false)` will be ignored as `false => p` can be deduced as `true` for any `p`.
 
-In proof path 1, we prove properties inside the loop and at last check that the loop invariant is inductive.
+In proof path 1, we prove properties inside the loop and at last check that the loop-invariant is inductive.
 
 In proof path 2, we prove properties after leaving the loop. As we leave the loop only when the loop guard is violated, the post condition of the loop can be expressed as
-`!guard && inv`, which is `x <= 1 && x >= 1` in the example. The postcondition implies `x == 1`—the property we want to prove at the end of `simple_loop_with_loop_contracts`.
+`!guard && inv`, which is `x <= 1 && x >= 1` in the example. The postcondition implies `x == 1`—the property we want to prove at the end of `simple_loop_with_loop_invariants`.
+
+## Historic values and extra variables
```
zhassan-aws, line 145: The documentation of historic values is unrelated to this PR, so should be moved to its own PR.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+### For-loop rewrite 
+
+When there is loop-invariant for `for-loop`, the `loop_invariant` procedural macros 
+will rewrite `for-loop` into a `while-loop`, but not exactly the same way as 
+https://doc.rust-lang.org/reference/expressions/loop-expr.html#iterator-loops. 
+Let's use the following example to demonstrate the disadvantages of rewriting 
+using the original Rust `into_iter` function:
+
+```Rust
+let a : [u8,10] = kani::any();
```
zhassan-aws:
```suggestion
let a: [u8; 10] = kani::any();
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+#[kani::loop_invariant(...)]
+for i in a {
+  ... //loop body
+}
+```
+
+is rewritten into:
+
+```Rust
+let a : [u8,10] = kani::any();
```
zhassan-aws:
```suggestion
let a: [u8; 10] = kani::any();
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+let mut kani_iter = a.into_iter();
+#[kani::loop_invariant(...)]
+loop {
+  match kani_iter.next() {
+    Some (i) => {
```
zhassan-aws:
```suggestion
    Some(i) => { 
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+```Rust
+let a : [u8,10] = kani::any();
+let mut kani_iter = a.into_iter();
+#[kani::loop_invariant(...)]
+loop {
+  match kani_iter.next() {
+    Some (i) => { 
+      ...  //loop body
+      }
+    None => {break; }
```
zhassan-aws:
```suggestion
    None => { break; }
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+The returned type of `into_iter` depends on the type of the caller that implements the `IntoIter` trait.
+In this case, it is:
+
+```Rust
+pub struct IntoIter<T, const N: usize> {
+    inner: InnerSized<T, N>,
+}
+```
+
+where the dependences are:
```
zhassan-aws:
```suggestion
where the dependencies are:
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+    DATA: PartialDrop,
+{
+    alive: IndexRange,
+    data: DATA,
+}
+// and so on ...
+```
+
+In order to contract the loop, the user must provide the loop-invariant for the variable `kani_iter`, which is modified and 
+will be havoc.
```
zhassan-aws:
```suggestion
will be havocked. 
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+In order to contract the loop, the user must provide the loop-invariant for the variable `kani_iter`, which is modified and 
+will be havoc. 
+The loop-invariant should include some propositions such as `kani_iter.inner.alive.start <= kani_iter.inner.alive.end`,
+`kani_iter.inner.data.as_ptr() == a.as_ptr()`, ... until the loop-invariant is successfully verified.
+The obvious three disadvantages are:
+1. All the fields `inner`, `alive`, `start`, `end`, ... are private, so those propositions are not allowed by the Rust compiler.
+2. The users must know about all the implementation designs and related internal types that the Rust standard library uses 
+to implement the `into_iter` function, and everything are different for the types that implements the `IntoIter` trait. 
+Even then, it is so complicated and tedious to write the loop-invariant. 
+3. The long call stack will slow down the performance.
```
zhassan-aws, line 282: I'm not sure this is true given that CBMC inlines everything.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+
+To avoid those disadvantages, we overide the `into_iter` function and `IntoIter` trait by our own version (`kani_into_iter()` and `KaniIntoIter`)
```
zhassan-aws:
```suggestion
To avoid those disadvantages, we override the `into_iter` function and `IntoIter` trait by our own version (`kani_into_iter()` and `KaniIntoIter`) 
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+`kani_iter.inner.data.as_ptr() == a.as_ptr()`, ... until the loop-invariant is successfully verified.
+The obvious three disadvantages are:
+1. All the fields `inner`, `alive`, `start`, `end`, ... are private, so those propositions are not allowed by the Rust compiler.
+2. The users must know about all the implementation designs and related internal types that the Rust standard library uses 
+to implement the `into_iter` function, and everything are different for the types that implements the `IntoIter` trait. 
+Even then, it is so complicated and tedious to write the loop-invariant. 
+3. The long call stack will slow down the performance.
+
+To avoid those disadvantages, we overide the `into_iter` function and `IntoIter` trait by our own version (`kani_into_iter()` and `KaniIntoIter`) 
+with simplified implementation but keeping the same semantics.
```
zhassan-aws:
```suggestion
with a simplified implementation but keeping the same semantics.
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+Our rewritten version for the `for-loop` is as follows:
+
+```Rust
+let a : [u8,10] = kani::any();
```
zhassan-aws:
```suggestion
let a: [u8; 10] = kani::any();
```

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+let kani_iter_xxxx = kani::kani_into_iter(a);
```
zhassan-aws, line 290: Why the `_xxxx` suffix?

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+let kaniiterlen = kani_iter_xxxx.len();
```
zhassan-aws: Use snake case for variable names, e.g. `kani_iter_len`.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+let mut i = kani_iter_xxxx.first()
+let mut kaniindex = 0;
```
zhassan-aws: Same here: `kaniindex` -> `kani_index`.

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+#[kani::loop_invariant(...)]
+while (kaniindex < kaniiterlen) {
```
zhassan-aws: There are no parentheses for while/if conditions in Rust

rfc/src/rfcs/0012-loop-contracts.md
```diff
@@ -214,6 +214,104 @@ the name of its `kani::any()` function, and add the map to the loop latch too.
...
+  i = kani_iter_xxxx.indexing(kaniindex);
```
zhassan-aws: What does the `indexing` method do? Where is it defined?

docs/src/reference/experimental/loop-contracts.md
tautschnig: But if we don't expose those variables, how are people going to express those loop invariants? Above, Thanh already explained how the approach with `enumerate()` wouldn't work.

zhassan-aws: I discussed with @thanhnguyen-aws offline and the reason `enumerate` wouldn't work is because of how we implemented it, but it should be possible to fix the implementation to make it work (@thanhnguyen-aws please correct me if I misunderstood).

One concern I raised with @thanhnguyen-aws is that exposing those variables means that we're committing to providing them for all possible iterator types. While an "index" shouldn't be an issue, "len" might not be easy to provide for all iterators (e.g. [the Chars iterator](https://doc.rust-lang.org/std/str/struct.Chars.html)). So if we decide to drop one of those variables in the future, this would break user code.

rfc/src/rfcs/0012-loop-contracts.md
thanhnguyen-aws: _xxxx is a hash code that is generated from the position of the loop. It is a technique to hide proc-macro generated variables from user. We would apply the same technique to kani_index and kani_iter_len if we decide to hide them.

docs/src/reference/experimental/loop-contracts.md
thanhnguyen-aws: It was an accident when I replace all loop_assigns to loop_modifies, I will correct it.

tautschnig:
A few notes from a discussion that was held out-of-band: it would be nice to have a syntax like `kani::index` (instead of `kaniindex` or `kani_index`) to make it entirely obvious that this isn't a user-provided variable. Apart from this, however: let's get this merged and experiment with it to see what the user experience will be like.

rod-chapman:
Please translate and verify the inc_vector() function that I wrote here:
 https://github.com/rod-chapman/cbmc-examples/blob/8a708ed2f47aba237afe049e902686f2206c595a/arrays/ar.c#L315

using Kani. This should be an interesting demo that combines "for" loops with quantified invariants.

thanhnguyen-aws:
> A few notes from a discussion that was held out-of-band: it would be nice to have a syntax like `kani::index` (instead of `kaniindex` or `kani_index`) to make it entirely obvious that this isn't a user-provided variable. Apart from this, however: let's get this merged and experiment with it to see what the user experience will be like.

I changed it.

rod-chapman:
Have you got my inc_vector() test case done? If so, please point me at the sources in the repo...

thanhnguyen-aws:
> Have you got my inc_vector() test case done? If so, please point me at the sources in the repo...

Hello, we are fixing some issue with using quantifiers in loop invariants. I will let you know when we finish. Thank you very much.

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -116,31 +118,79 @@ must be sufficient to establish subsequent assertions. If it is not, the abstrac
...
+        i = i - 2;
+    }
+    assert!(i == 2);
+}
+```
+
+### Extra variables in `for` loop
+
+There are two extra variables that can be used in loop contracts for `for` loops:
+1. `kani::indexx` : the position (index) of the current iteration in the iterator.
```
zhassan-aws:
```suggestion
1. `kani::index` : the position (index) of the current iteration in the iterator.
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -116,31 +118,79 @@ must be sufficient to establish subsequent assertions. If it is not, the abstrac
...
+2. `kani_iter_len` : the length of the iterator.
```
zhassan-aws: Is `kani_iter_len` removed? I no longer see it in any of the tests. If so, remove it from here.

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -116,31 +118,79 @@ must be sufficient to establish subsequent assertions. If it is not, the abstrac
...
+
+Note that the two extra variables above are only associated with the `for` loop right below the loop contract.
```
zhassan-aws:
```suggestion
Note that the two extra variables above are only associated with the `for` loop that immediately follows the loop contract.
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -116,31 +118,79 @@ must be sufficient to establish subsequent assertions. If it is not, the abstrac
...
+    kani::assume(kani::forall!(|i in (0,10)| a[i] <= 20));
+    #[kani::loop_invariant( i == kani::index && i <= kani_iter_len && sum <= (kani::index as u32 * 29) )]
+    for (i, j) in a.iter().enumerate() {
+        sum = sum + (i as u32) + (*j as u32);
+    }
+    assert!(sum <= 290);
+}
+```
+
+## loop contracts inside functions with contracts
```
zhassan-aws:
```suggestion
## Loop contracts inside functions with contracts 
```

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -182,6 +183,342 @@ impl LoopContractPass {
         user_vars
     }
 
+    // Get the list of tuples:
+    // (firstpat, the block_id where firstpat get assigned, the corresponding indexpat, the block_id where indexpat get assigned)
+    fn get_firstpats_and_indexingpats(
```
zhassan-aws: Please add underscores between words: `get_firstpats_and_indexingpats` -> `get_first_pats_and_indexing_pats`.

rfc/src/rfcs/0012-loop-contracts.md
zhassan-aws:
```suggestion
When there is a loop invariant for a `for` loop, the `loop_invariant` procedural macro
```

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,487 @@
...
+        impl<'a, T: Copy> KaniIter for KaniRefIter<'a, T> {
+            type Item = &'a T;
+            fn indexing(&self, i: usize) -> Self::Item {
+                unsafe { &*self.ptr.wrapping_add(i) }
+            }
+            fn first(&self) -> Self::Item {
+                unsafe { &*self.ptr }
+            }
+            fn assumption(&self) -> bool {
+                //SAFETY: this call is safe as Rust compiler will complain if we write a for-loop for initnitialized object
```
zhassan-aws:
```suggestion
                // SAFETY: this call is safe as Rust compiler will complain if we write a for-loop for initialized object
```
And did you mean uninitialized?

I'm unclear what this comment is saying.

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,487 @@
...
+                    }
+                }
+            };
+        }
+
+        generate_impl_range!(i8);
+        generate_impl_range!(i16);
+        generate_impl_range!(i32);
+        generate_impl_range!(i64);
+        generate_impl_range!(isize);
```
zhassan-aws, line 153: What about `i128` and `u128`?

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,487 @@
...
+        use core_path::mem as stdmem;
+        use core_path::ops::Range;
+        use core_path::slice::Iter;
+
+        pub trait KaniIter
+        where
+            Self: Sized,
+        {
+            type Item;
+            fn indexing(&self, i: usize) -> Self::Item;
```
zhassan-aws: Should this be called `get` instead?

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,487 @@
...
+            fn first(&self) -> Self::Item;
```
zhassan-aws, line 66: What happens if there are no elements?

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -254,20 +254,239 @@ fn while_let_rewrite(loopexpr: Stmt) -> Stmt {
...
+    for_loop: ExprForLoop,
+    loop_id: &str,
+) -> (Stmt, Option<ForLoopExtraStmts>) {
+    // Extract components from the for loop
+    let pat = *for_loop.pat;
+    let expr = for_loop.expr;
+    let body = for_loop.body;
+
+    // Create an iterator variable name
+    let mut indexname = "kani_index".to_owned();
```
zhassan-aws:
```suggestion
    let mut index_name = "kani_index".to_owned();
```

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -254,20 +254,239 @@ fn while_let_rewrite(loopexpr: Stmt) -> Stmt {
...
+    indexname.push_str(loop_id);
+    let kani_index = format_ident!("{}", indexname);
+
+    let mut itername = "kaniiter".to_owned();
```
zhassan-aws:
```suggestion
    let mut iter_name = "kaniiter".to_owned();
```

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -254,20 +254,239 @@ fn while_let_rewrite(loopexpr: Stmt) -> Stmt {
...
+    itername.push_str(loop_id);
+    let iter_ident = format_ident!("{}", itername);
+
+    let mut lenname = "kani_iter_len".to_owned();
```
zhassan-aws:
```suggestion
    let mut len_name = "kani_iter_len".to_owned();
```

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -358,8 +577,76 @@ pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+            })
+            .into()
+        } else {
+            quote_spanned!(original_span =>
+            {
+            #init_iter_stmt
+            #init_len_stmt
+            if #kani_iter_len > 0
+            {
+            #init_index_stmt
```
zhassan-aws: The indentation is off

zhassan-aws:
Also, can you add an example where verification fails because the invariant is not strong enough?

library/kani_core/src/iter.rs
thanhnguyen-aws: Yes, I mean uninitialized

library/kani_core/src/iter.rs
thanhnguyen-aws: I added the two

library/kani_core/src/iter.rs
thanhnguyen-aws: The indexing and first functions are only called inside an `if kani_iter_len > 0 { ... }` block generated by the loop_invariant macro.

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
```diff
@@ -0,0 +1,25 @@
...
+//! Check for-loop invariant with quantifier for array increasing.
+//! Having SMT issue that need to be fix (see issue #4282)
+
+#![feature(proc_macro_hygiene)]
+#![feature(stmt_expr_attributes)]
+
+#[kani::proof]
+#[kani::solver(z3)]
+fn array_inc() {
+    let a: [u8; 60] = kani::any();
```
rod-chapman: Nope... the array upper-bound must be 256. The element type must be i32 please. I chose these types for a good reason!

Can you also show me this done as a function which has a pre-condition, so no need for the initial assignment from kani::any() and no assume statement please.

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -116,31 +118,77 @@ must be sufficient to establish subsequent assertions. If it is not, the abstrac
...
+        if i == 1 {
+            break;
+        }
+        i = i - 2;
+    }
+    assert!(i == 2);
+}
+```
+
+### kani::index variable in `for` loop
```
zhassan-aws:
```suggestion
### `kani::index` variable in `for` loop
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -116,31 +118,77 @@ must be sufficient to establish subsequent assertions. If it is not, the abstrac
...
+Example:
+
+```Rust
+#[kani::proof]
+fn forloop() {
+    let mut sum: u32 = 0;
+    let a: [u8; 10] = kani::any();
+    kani::assume(kani::forall!(|i in (0,10)| a[i] <= 20));
+    #[kani::loop_invariant(sum <= (kani::index as u32 * 29) )]
+    for i in a.iter().enumerate() {
```
zhassan-aws: This should be:
```suggestion
    for (i, j) in a.iter().enumerate() {
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -116,31 +118,77 @@ must be sufficient to establish subsequent assertions. If it is not, the abstrac
...
+
+Example:
+
+```Rust
+#[kani::proof]
+fn forloop() {
+    let mut sum: u32 = 0;
+    let a: [u8; 10] = kani::any();
+    kani::assume(kani::forall!(|i in (0,10)| a[i] <= 20));
+    #[kani::loop_invariant(sum <= (kani::index as u32 * 29) )]
```
zhassan-aws: Should 29 be 20?

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -116,31 +118,77 @@ must be sufficient to establish subsequent assertions. If it is not, the abstrac
...
+    for i in a.iter().enumerate() {
+        sum = sum + (i as u32) ;
+    }
+    assert!(sum <= 200);
+}
+```
+
 ## Loop contracts inside functions with contracts 
-Kani supports using loop contracts together with function contracts, as demonstrated in the following example:
+Kani supports using loop contracts together with function-contracts, as demonstrated in the following example:
```
zhassan-aws: Revert this change

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -162,7 +210,7 @@ fn contract_proof() {
 }
 ```
 
-When loop contracts and function contracts are both enabled (by flags `-Z loop-contracts -Z function-contracts`), 
+When loop contracts and function-contracts are both enabled (by flags `-Z loop-contracts -Z function-contracts`),
```
zhassan-aws: And this one

library/kani/src/iter.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This macro generates implementations of the `KaniIntoIter` trait for various common types that are used in `for` loops.
+//! We use this trait to overwrite the Rust IntoIter trait to reduce call stacks and avoid complicated loop invariant specifications,
+//! while maintaining the semantics of the loop.
+use crate::{KaniIntoIter, KaniPtrIter};
+
+impl<T: Copy> KaniIntoIter for Vec<T> {
```
zhassan-aws: Why is `Copy` required?

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -254,20 +254,239 @@ fn while_let_rewrite(loopexpr: Stmt) -> Stmt {
...
+        body
+    }
+}
+```
+Note that
+
+1. expr's type is required to impl KaniIntoIter trait, which is the override impl of Rust's IntoIter (see library/kani_core/src/iter.rs).
+Reason:
+    a) Reduce stack-call
+    b) Avoid types that cannot be havoc effectively (user cannot state the type invariant in the loop invariant due to private fields)
```
zhassan-aws:
```suggestion
    b) Avoid types that cannot be havocked effectively (user cannot state the type invariant in the loop invariant due to private fields)
```

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -254,20 +254,239 @@ fn while_let_rewrite(loopexpr: Stmt) -> Stmt {
...
+        kani_index = kani_index + 1; \\ increase_iter_stmt
+        body
+    }
+}
+```
+Note that
+
+1. expr's type is required to impl KaniIntoIter trait, which is the override impl of Rust's IntoIter (see library/kani_core/src/iter.rs).
+Reason:
+    a) Reduce stack-call
```
zhassan-aws: I suggest removing this one.

library/kani_macros/src/sysroot/loop_contracts/mod.rs
```diff
@@ -358,33 +576,101 @@ pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+            #(#onentry_decl_stms)*
+            #(#prev_decl_stms)*
+            let mut #loop_body_closure = ||
+            #loop_body;
+            let (#loop_body_closure_ret_1, #loop_body_closure_ret_2) = #loop_body_closure ();
+            if #loop_body_closure_ret_2.is_some() {
+                return #loop_body_closure_ret_2.unwrap();
+            }
+            if #loop_body_closure_ret_1 {
+            // Dummy function used to force the compiler to capture the environment.
```
zhassan-aws, line 607: The comment indentation is off

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
```diff
@@ -0,0 +1,25 @@
...
+    kani::assume(kani::forall!(|i in (0,60)| a[i] <= u8::MAX));
```
zhassan-aws: What is the purpose of this assumption?

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
```diff
@@ -0,0 +1,25 @@
...
+    let mut b: [u8; 60] = a.clone();
+    #[kani::loop_invariant(i < 60
+            && kani::forall!(|j in (kani::index, 60)| b[j] == a[j])
```
zhassan-aws: Why is one forall in terms of `kani::index` and the other in terms of `i`?

tests/expected/loop-contract/for_loop_for_chain.rs
```diff
@@ -0,0 +1,23 @@
...
+//! Check for-loop invariant for Chain.
+#![feature(proc_macro_hygiene)]
+#![feature(stmt_expr_attributes)]
+
+use std::cmp::min;
+
+#[kani::proof]
+fn forloop() {
+    let mut sum: u32 = 0;
+    let mut a: [u8; 10] = kani::any();
```
zhassan-aws: Neither `a` nor `b` need `mut`.

tests/expected/loop-contract/for_loop_for_zip.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+// kani-flags: -Z loop-contracts -Z mem-predicates -Z quantifiers
+
+//! Check for-loop invariant for Vec.
```
zhassan-aws: Vec -> Zip

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -182,6 +183,339 @@ impl LoopContractPass {
...
+            {
+                // Get the function signature of the terminator call.
+                let Some(RigidTy::FnDef(fn_def, _)) = terminator_func
+                    .ty(body.locals())
+                    .ok()
+                    .map(|fn_ty| fn_ty.kind().rigid().unwrap().clone())
+                else {
+                    continue;
+                };
+                // Check if the function is `kani::internal::run_contract_fn`.
```
zhassan-aws: The comment looks wrong.

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,528 @@
...
+        impl<I: KaniIter> KaniRevIter<I> {
+            pub fn new(iter: I) -> Self {
+                KaniRevIter { iter }
+            }
+        }
+
+        impl<I: KaniIter> KaniIter for KaniRevIter<I> {
+            type Item = I::Item;
+            fn nth(&self, i: usize) -> Self::Item {
+                //assert!(i < self.n && i < self.iter.len());
```
zhassan-aws: Does this assertion not hold? If so, please remove it.

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,528 @@
...
+                        self
+                    }
+                }
+            };
+        }
+
+        generate_impl_KaniIter_range!(i8);
+        generate_impl_KaniIter_range!(i16);
+        generate_impl_KaniIter_range!(i32);
+        generate_impl_KaniIter_range!(i64);
```
zhassan-aws, line 414: i128

library/kani_core/src/iter.rs
```diff
@@ -0,0 +1,528 @@
...
+        generate_impl_KaniIter_range!(isize);
+        generate_impl_KaniIter_range!(u8);
+        generate_impl_KaniIter_range!(u16);
+        generate_impl_KaniIter_range!(u32);
+        generate_impl_KaniIter_range!(u64);
```
zhassan-aws, line 420: u128

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
```diff
@@ -0,0 +1,25 @@
...
+            && kani::forall!(|j in (0, i as usize)| b[j] == a[j] + 1)
+    )]
+    for i in 0..60 {
+        b[i as usize] = a[i as usize] + 1;
+    }
```
tautschnig: Can we please also have as test the variant that uses `.iter_mut()`?

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
rod-chapman: The type of the array element should be i32 not u8. Then the precondition needs to be that
`a[i] < i32::MAX` so that incrementing each element does not overflow. I specified these types very carefully. Please do not change them!

Secondly, please express the pre-condition as a "requires" contract not an "assume" - I need to see this working with contracts please.

thanhnguyen-aws: @zhassan-aws Just to make sure we will pass the overflow check in the loop.
@rod-chapman Yes, I will fix that.

zhassan-aws: This would make sense if the array element is `i32` as @rod-chapman mentioned.

library/kani/src/iter.rs
thanhnguyen-aws: The struct `KaniPtrIter` requires `T: Copy`, otherwise, Rust will complain about `unsafe { *self.ptr.wrapping_add(i) }` in the `nth` function.

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
thanhnguyen-aws: This PR does not support `iter_mut` yet. I will add that variant when we support `iter_mut` (another PR in the future).

library/kani/src/iter.rs
zhassan-aws: Perhaps the `nth` function should return a reference to the `Item` then? Requiring `Vec` elements to be `Copy` is too restrictive.

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
tautschnig: But maybe now is still a good time to add a fixme test?

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -1,7 +1,9 @@
 # Loop Contracts
 
-Loop contract are used to specify invariants for loops for the sake of extending Kani's *bounded proofs* to *unbounded proofs*.
-A [loop invariant](https://en.wikipedia.org/wiki/Loop_invariant) is an expression that holds upon entering a loop and after every execution of the loop body.
+## Introduction
+
+Loop contracts are used to specify invariants for loops for the sake of extending Kani's *bounded proofs* to *unbounded proofs*.
+A [loop contract](https://en.wikipedia.org/wiki/Loop_invariant) is an expression that holds upon entering a loop and after every execution of the loop body.
```
tautschnig:
```suggestion
Loop contracts are used to specify invariants for loops for the sake of extending Kani's *bounded proofs* to *unbounded proofs*.
A [loop invariant](https://en.wikipedia.org/wiki/Loop_invariant) is an expression that holds upon entering a loop and after every execution of the loop body.
Loop contracts are composed of one or more loop invariants as well as optional `loop_modifies` attributes.
```

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -44,8 +46,8 @@ fn simple_loop_with_loop_contracts() {
 }
 ```
 
-Here, the loop invariant `#[kani::loop_invariant(x >= 1)]` specifies that the condition `x >= 1` must hold true at the start of each iteration before the loop guard is
-checked. Once Kani verifies that the loop invariant is inductive, it will use the invariant to abstract the loop and avoid unwinding. 
+Here, the loop contract `#[kani::loop_invariant(x >= 1)]` specifies that the condition `x >= 1` must hold true at the start of each iteration before the loop guard is
+checked. Once Kani verifies that the loop contract is inductive, it will use the invariant to abstract the loop and avoid unwinding.
```
tautschnig: I believe this change should be reverted, the previous text was IMHO correct.

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -61,7 +63,7 @@ Check 10: simple_loop_with_loop_contracts.loop_invariant_base.1
          - Description: "Check invariant before entry for loop simple_loop_with_loop_contracts.0"
          - Location: simple_while_loop.rs:15:5 in function simple_loop_with_loop_contracts
 
-Check 11: simple_loop_with_loop_contracts.loop_modifies.1
+Check 11: simple_loop_with_loop_contracts.loop_assigns.1
```
tautschnig, line 67: Should this really be changed?

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
thanhnguyen-aws: I added

docs/src/reference/experimental/loop-contracts.md
thanhnguyen-aws: It is the output of CBMC, so `loop_assigns` is what it actually print out.

docs/src/reference/experimental/loop-contracts.md
thanhnguyen-aws: @zhassan-aws suggested that we don't make any change in "loop invariant" - "loop contract" naming in this PR. We will create a separate PR or it, so I reverted all the changes.

zhassan-aws: I think while reverting the changes, you changed some current occurrences of "loop invariant" to "loop contract".

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
```diff
@@ -0,0 +1,54 @@
...
+
+#![feature(proc_macro_hygiene)]
+#![feature(stmt_expr_attributes)]
+
+#[kani::requires(a.len() < 256)]
+#[kani::requires(kani::forall!(|i in (0, a.len())| a[i] < i32::MAX))]
+#[kani::ensures(kani::forall!(|i in (0, a.len())| a[i] = on_entry(a[i]) + 1))]
+fn array_inc(a: &mut [i32]) {
+    let b: [u8; 60] = a.clone();
+    #[kani::loop_invariant(i < 60
```
rod-chapman: Please see comments on commit: https://github.com/model-checking/kani/commit/a77f0cdfd3d151674964f545c34ec8df7f3ef7fb

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
```diff
@@ -27,10 +26,9 @@ fn array_inc(a: &mut [i32]) {
 #[kani::requires(kani::forall!(|i in (0, a.len())| a[i] < i32::MAX))]
 #[kani::ensures(kani::forall!(|i in (0, a.len())| a[i] = on_entry(a[i]) + 1))]
 fn array_inc_iter_mut(a: &mut [i32]) {
-    let b: [u8; 60] = a.clone();
     #[kani::loop_invariant(i < 60
-            && kani::forall!(|j in (kani::index, 60)| b[j] == a[j])
-            && kani::forall!(|j in (0, kani::index)| a[j] == b[j] + 1)
+            && kani::forall!(|j in (kani::index, 60)| a[j] == on_entry(a[j]))
```
rod-chapman: Why does "60" still appear here? Surely, it should be "a.len()"? How can this prove?

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
```diff
@@ -0,0 +1,52 @@
...
+//! Check for-loop invariant with quantifier for array increasing.
+//! Having SMT issue that need to be fix (see issue #4282)
+
+#![feature(proc_macro_hygiene)]
+#![feature(stmt_expr_attributes)]
+
+#[kani::requires(a.len() < 256)]
+#[kani::requires(kani::forall!(|i in (0, a.len())| a[i] < i32::MAX))]
+#[kani::ensures(kani::forall!(|i in (0, a.len())| a[i] = on_entry(a[i]) + 1))]
+fn array_inc(a: &mut [i32]) {
```
tautschnig, line 15: Didn't Rod's original example have a fixed-size array here rather than a slice? With a fixed size we should indeed be able to use autoharness.

docs/src/reference/experimental/loop-contracts.md
```diff
@@ -116,29 +119,75 @@ must be sufficient to establish subsequent assertions. If it is not, the abstrac
...
+//   loop contract is satisfied and loop guard is violated.
 assert!(x == 1);
 ```
-That is, we assume that we are in an arbitrary iteration after checking that the loop invariant holds for the base case. With the inductive hypothesis (`kani::assume(x >= 1);`),
+That is, we assume that we are in an arbitrary iteration after checking that the loop contract holds for the base case. With the inductive hypothesis (`kani::assume(x >= 1);`),
 we will either enter the loop (proof path 1) or leave the loop (proof path 2). We prove the two paths separately by killing path 1 with `kani::assume(false);`.
 Note that all assertions after `kani::assume(false)` will be ignored as `false => p` can be deduced as `true` for any `p`.
 
-In proof path 1, we prove properties inside the loop and at last check that the loop invariant is inductive.
+In proof path 1, we prove properties inside the loop and at last check that the loop contract is inductive.
```
tautschnig: I think these specific changes from `invariant` to `contract` should also be reverted.

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[kani::requires(a.len() < 256)]
+#[kani::requires(kani::forall!(|i in (0, a.len())| a[i] < i32::MAX))]
+#[kani::ensures(kani::forall!(|i in (0, a.len())| a[i] = on_entry(a[i]) + 1))]
+fn array_inc_iter_mut(a: &mut [i32]) {
+    #[kani::loop_invariant(i < 60
+            && kani::forall!(|j in (kani::index, a.len())| a[j] == on_entry(a[j]))
+            && kani::forall!(|j in (0, kani::index)| a[j] == on_entry(a[j]) + 1)
+    )]
+    for x in a.iter_mut() {
```
tautschnig, line 33: Can you explain what it would take to support `iter_mut()` (or maybe why it cannot easily be supported at this point)?

docs/src/reference/experimental/loop-contracts.md
rod-chapman: For me, loops have contracts that come in many forms: modifies or "assigns", invariants, "variant" or "decreases" contracts and so on.

A "loop invariant" is one type of "loop contract" which has a specific meaning and use in verification.

So... say "loop contract" when you mean any one of them. Say "loop invariant" when you specifically mean the thing that really is the invariant of the loop.

This conversation should not be necessary in 2025!

library/kani/src/iter.rs
thanhnguyen-aws: I changed it into Clone.

tests/expected/loop-contract/array_icr_for_loop_fixme.rs
thanhnguyen-aws: I don't think it is very challenging, but I need to test it though-roughly. I just don't want to add more into this PR as we want it to be merged ASAP.

rfc/src/rfcs/0012-loop-contracts.md
thanhnguyen-aws: I am sure that the performance (of verification) is significantly improved when we use this implementation instead of the original Rust one, but I am not sure whether it comes from shortening the call stack or from reducing the number of variables or option checking ...

tautschnig (approved):
We may still need to follow up this work with some refinement, but this should be ready for broader evaluation.

PR #4148, opened 2025-06-09
Optimize goto binary exporting in `cprover_bindings`

AlexanderPortland:
Motivated by the finding that the [`write_goto_binary_file`](https://github.com/model-checking/kani/blob/1b25a1853fab3262380e9baefdfeb929edd53fd1/cprover_bindings/src/irep/goto_binary_serde.rs#L97-L103) function is responsible for around half of our compiler's codegen time, this PR does a collection of low-hanging fruit optimizations for the way we export goto files in the `cprover_bindings` crate.

This includes:
- Switching `IrepNumbering` caches to use the faster [`FxHashMap`](https://nnethercote.github.io/2021/12/08/a-brutally-effective-hash-function-in-rust.html) (as used in the rust compiler)
- Using `.entry().or_insert()` to initialize the irep cache
    - This avoids having to hash separately for a `.get()` and `.insert()` (as the compiler seemed to have not optimized them together)
- Making `IrepId` strings copy-on-write
    - This avoids unnecessary string allocations by allowing us to construct `InternedString`s directly from borrowed string literals when possible

Initial testing (on [`perf/format`](https://github.com/model-checking/kani/blob/main/tests/perf/format/src/lib.rs) & [`perf/vec/vec`](https://github.com/model-checking/kani/blob/main/tests/perf/vec/vec/src/main.rs)) suggests that these changes make our compiler ~13% faster, but I'm also working on comparing performance on the whole std project for the future.

### DIFF
--- a/cprover_bindings/src/irep/goto_binary_serde.rs
+++ b/cprover_bindings/src/irep/goto_binary_serde.rs
@@ -80,6 +80,9 @@
 
 use crate::irep::{Irep, IrepId, Symbol, SymbolTable};
 use crate::{InternString, InternedString};
+#[cfg(not(test))]
+use fxhash::FxHashMap;
+#[cfg(test)]
 use std::collections::HashMap;
 use std::fs::File;
 use std::hash::Hash;
@@ -199,6 +202,23 @@ impl IrepNumberingInv {
     }
 }
 
+#[cfg(not(test))]
+/// A numbering of [InternedString], [IrepId] and [Irep] based on their contents.
+struct IrepNumbering {
+    /// Map from [InternedString] to their unique numbers.
+    string_cache: FxHashMap<InternedString, usize>,
+
+    /// Inverse string cache.
+    inv_string_cache: Vec<NumberedString>,
+
+    /// Map from [IrepKey] to their unique numbers.
+    cache: FxHashMap<IrepKey, usize>,
+
+    /// Inverse cache, allows to get a NumberedIrep from its unique number.
+    inv_cache: IrepNumberingInv,
+}
+
+#[cfg(test)]
 /// A numbering of [InternedString], [IrepId] and [Irep] based on their contents.
 struct IrepNumbering {
     /// Map from [InternedString] to their unique numbers.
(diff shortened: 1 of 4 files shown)


cprover_bindings/src/irep/goto_binary_serde.rs
```diff
@@ -199,6 +202,23 @@ impl IrepNumberingInv {
     }
 }
 
+#[cfg(not(test))]
+/// A numbering of [InternedString], [IrepId] and [Irep] based on their contents.
```
tautschnig, line 206: Why is it that this is only done for the non-test scenario? Asking differently: why can we not use this for testing?

Cargo.toml
```diff
@@ -31,6 +31,8 @@ path = "src/bin/cargo_kani.rs"
 
 [profile.release]
 strip = "debuginfo"
+codegen-units = 1
+lto = "fat"
```
tautschnig: I'd appreciate if this change was done in a separate PR. It has consequences beyond optimising goto-binary export (and, as you say in the PR description, might not have any measurable performance gains).

cprover_bindings/src/irep/goto_binary_serde.rs
AlexanderPortland: For the testing scenario, there's [a module](https://github.com/model-checking/kani/blob/1b25a1853fab3262380e9baefdfeb929edd53fd1/cprover_bindings/src/irep/goto_binary_serde.rs#L922) that collects stats on memory usage w/ the `mem::DynamicUsage` trait. The default `HashMap` implements that trait, but `FxHashMap` doesn't and since neither the trait nor the type are defined by us I couldn't implement it myself. Only using `FxHashMap` for non-test was my work around to keep the memory usage stats (under the assumption memory consumption would be the same between the two and just the hash function would be different), but maybe there's a better way?

Cargo.toml
AlexanderPortland: Yes can do! Do you think it's worthwhile holding off on this change until there's a demonstrated performance gain? Based on my early testing any improvement doesn't appear to be significant, but I initially included it w/ the reasoning that it wouldn't hurt either way.

carolynzech: Since you so nicely put each change in its own commit 😀 I'd recommend cherry-picking this commit over to some separate local branch and rebasing this branch to not include it. For now, let's not make a PR for it. If we do some measurements later that suggest it does help, we can make a PR then. My inclination would be to not make changes unless we can prove they help, especially if they add complexity.

AlexanderPortland: Okay, this change has been removed!

cprover_bindings/src/irep/goto_binary_serde.rs
tautschnig: Ah, thank you for explaining. Could you please include some comment in the source code to state what you just said above?

PR #4151, opened 2025-06-11
Ensure that contract closures are FnOnce

### LINKED ISSUE #3764: Can't write contracts for functions that return mutable references to input arguments
I tried this code:

```rust
#[kani::requires(*val != 0)]
unsafe fn foo(val: &mut u8) -> &mut u8 {
    val
}

#[kani::proof]
fn harness() {
    let mut x: u8 = kani::any();
    unsafe { foo(&mut x) };
}
```

```
cargo kani
```

with Kani version: 0.56

Instead, this happened: 
```rust
error: captured variable cannot escape `FnMut` closure body
  --> src/lib.rs:12:5
   |
11 | unsafe fn foo(val: &mut u8) -> &mut u8 {
   |               ---              - inferred to be a `FnMut` closure
   |               |
   |               variable defined here
12 |     val
   |     ^^^
   |     |
   |     returns a reference to a captured variable which escapes the closure body
   |     variable captured here
   |
   = note: `FnMut` closures only have access to their captured variables while they are executing...
   = note: ...therefore, they cannot allow references to captured variables to escape

error: captured variable cannot escape `FnMut` closure body
  --> src/lib.rs:10:1
   |
10 | #[kani::requires(*val != 0)]
   | ^^^^^^^^^^^^^^^^^^---^^^^^^^
   | |                 |
   | |                 variable captured here
   | returns a reference to a captured variable which escapes the closure body
11 | unsafe fn foo(val: &mut u8) -> &mut u8 {
   |               ---              - inferred to be a `FnMut` closure
   |               |
   |               variable defined here
   |
   = note: `FnMut` closures only have access to their captured variables while they are executing...
   = note: ...therefore, they cannot allow references to captured variables to escape
   = note: this error originates in the attribute macro `kani::requires` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This is a minimal reproducer of a problem encountered trying to write contracts for `NonZero::from_mut_unchecked`, which has this function signature:
```rust
pub unsafe fn from_mut_unchecked(n: &mut T) -> &mut Self
```

vonaka:
Rust believes that Kani's contract closures are `FnMut`. This prevents us from writing contracts for functions that return mutable references to their input arguments (#3764).

To ensure Rust correctly infers these closures as `FnOnce`, they need to be wrapped in a dummy function that explicitly requires an `FnOnce`. This wrapping must be done at the point of closure definition, as doing it later, when calling the function, doesn't seem to have any effect.

Resolves #3764


vonaka:
@carolynzech I added some basic tests. Regarding PR itself, defining `force_fn_once` to later replace it with literarily the same function feels suboptimal. Maybe at the very least, the body of `force_fn_once` should be `unreachable!`, just like `kani_register_contract`. Speaking of which, I wonder if this function is still needed considering this PR

tests/expected/function-contract/mutable-references/ensures_with_two_mut_refs_fail.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Z function-contracts
+
```
carolynzech, line 4: Can you add a comment describing what this tests and why it fails?

tests/expected/function-contract/mutable-references/return_mut_ref.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright Kani Contributors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// kani-flags: -Z function-contracts
+
```
carolynzech, line 4: Add a comment describing what this tests, linking to the Github issue that this closes

carolynzech: Also can you add a test for the Github issue test case specifically? I did it locally and confirmed it works; it's just nice to have the sanity check that we did in fact solve that particular issue.

library/kani_macros/src/sysroot/contracts/bootstrap.rs
```diff
@@ -47,6 +47,18 @@ impl<'a> ContractConditionsHandler<'a> {
             #[kanitool::asserted_with = #assert_name]
             #[kanitool::modifies_wrapper = #modifies_name]
             #vis #sig {
+                // Dummy functions used to force the compiler to annotate Kani's
+                // closures as FnOnce.
+                #[inline(never)]
```
carolynzech: Can you expand this comment to describe why this is important?
(Perhaps link to the issue here again)

tests/expected/function-contract/mutable-references/return_mut_ref.expected
```diff
@@ -0,0 +1 @@
+VERIFICATION:- SUCCESSFUL
```
tautschnig, line 31: I'd suggest also including a particular "assertion ..." line(s).

vonaka:
> Remove the compiler & kani_core changes entirely and just do the transformations entirely in the macro expansion logic. I would think you could just change the macro logic to have an extra () that calls these const functions, so that then the closures are just there. You'd have to try it, though.

Right, so I simply reverted all the changes to the core and the compiler. I assume this is enough? The only problem I see is that if the file Kani tries to verify has its own `kani_force_fn_once`, Kani will fail to compile. But this is also the case for `kani_register_contract`, so at least I'm being consistent with the existing code. In general, I'm not sure how to fix this elegantly (and I'm not sure if it's worth fixing).

tests/expected/function-contract/mutable-references/return_mut_ref.expected
vonaka: I added a few lines, although the test is really elementary and doesn't have any interesting assertion.

library/kani_macros/src/sysroot/contracts/bootstrap.rs
vonaka: Thanks, I've added some comments.

carolynzech (approved):
Thanks! Before merging, can you check if we can indeed get rid of `kani_register_contract`?

vonaka:
> Thanks! Before merging, can you check if we can indeed get rid of `kani_register_contract`?

We can't, we still need this workaround to call closures from const functions.

PR #4154, opened 2025-06-14
Introduce compiler timing script & CI job

### LINKED ISSUE #2442: Measure compiler performance and overall execution time as part of our benchmarks

AlexanderPortland:
We have an existing `benchcomp` script to measure Kani's performance in CI, but it is prone to noise and only focuses on the end-to-end performance of compilation and verification together, leaving a need for more granular measurements.

This script focuses solely on changes in the runtime of the Kani compiler and runs with warm ups, repeats and outlier detection (based on the rust compiler's method in [`rustc-perf`](https://github.com/rust-lang/rustc-perf)) in an attempt to limit noise. The new `compile-timer-short` CI job uses this script on a subset of the `perf` tests (currently excluding `s2n-quic`, `kani-lib/arbitrary` & `misc/display-trait`) to produce a `benchcomp`-like table comparing the compiler performance per-crate before and after a given commit.

This also modifies our auto-labeller to ensure end-to-end benchmarks (like `benchcomp`) and the new compiler-specific ones are only run when the parts of Kani that they profile have changed.

We manually tested the CI job on my personal fork (see [this regressing run](https://github.com/AlexanderPortland/kani/actions/runs/15788016660?pr=6) from a test PR that intentionally slows down the compiler).

Resolves #2442

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -18,7 +18,7 @@ Z-CompilerBenchCI:
   - changed-files:
     # any parts of the compiler change
     - any-glob-to-any-file: ['kani-compiler/**', 'cprover_bindings/**', 'library/**']
-    # the way we call the compiler through cargo changes
-    - any-glob-to-any-file: ['kani-driver/src/call_cargo.rs']
+    # the way we call the compiler changes
+    - any-glob-to-any-file: ['kani-driver/src/call_cargo.rs', 'kani-driver/src/call_single_file.rs']
     # or if our dependencies change
-    - any-glob-to-any-file: ['rust-toolchain.toml', 'Cargo.lock', 'kani-dependencies']
+    - any-glob-to-any-file: ['rust-toolchain.toml', 'Cargo.lock']
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,9 +5,20 @@
 #
 # Note that we enable dot, so "**" matches all files in a folder
 
-Z-BenchCI:
+Z-EndToEndBenchCI:
 - any:
   - changed-files:
-    - any-glob-to-any-file: ['kani-compiler/**', 'kani-driver/src/call-*', 'cprover_bindings/**', 'library/**']
+    - any-glob-to-any-file: ['kani-compiler/**', 'kani-driver/src/call_*', 'cprover_bindings/**', 'library/**']
     - any-glob-to-any-file: ['rust-toolchain.toml', 'Cargo.lock']
     - any-glob-to-any-file: ['kani-dependencies']
+
+Z-CompilerBenchCI:
+- any:
+  # we want to run compiler benchmarks if:
+  - changed-files:
+    # any parts of the compiler change
+    - any-glob-to-any-file: ['kani-compiler/**', 'cprover_bindings/**', 'library/**']
+    # the way we call the compiler through cargo changes
+    - any-glob-to-any-file: ['kani-driver/src/call_cargo.rs']
+    # or if our dependencies change
+    - any-glob-to-any-file: ['rust-toolchain.toml', 'Cargo.lock', 'kani-dependencies']
--- a/.github/workflows/bench-compiler.yml
+++ b/.github/workflows/bench-compiler.yml
@@ -75,4 +75,68 @@ jobs:
           export PATH="${{ github.workspace }}/new/scripts:$PATH"
           cd new/tests/perf && ../../target/release/compile-timer --out-path compile-times-new.json --ignore kani-lib --ignore display_trait --ignore s2n-quic
       - name: Run analysis between the two
-        run: ./new/target/release/compile-analyzer --path-pre old/tests/perf/compile-times-old.json --path-post new/tests/perf/compile-times-new.json --only-markdown >> "$GITHUB_STEP_SUMMARY"
+        run: ./new/target/release/compile-analyzer --path-pre old/tests/perf/compile-times-old.json --path-post new/tests/perf/compile-times-new.json --only-markdown --suite-name short >> "$GITHUB_STEP_SUMMARY"
+
+  compile-timer-long:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Save push event HEAD and HEAD~ to environment variables
+        if: ${{ github.event_name == 'push' }}
+        run: |
+          echo "NEW_REF=${{ github.event.after}}" | tee -a "$GITHUB_ENV"
+          echo "OLD_REF=${{ github.event.before }}" | tee -a "$GITHUB_ENV"
+      - name: Save pull request HEAD and base to environment variables
+        if: ${{ contains(fromJSON('["pull_request", "pull_request_target"]'), github.event_name) }}
+        run: |
+          echo "OLD_REF=${{ github.event.pull_request.base.sha }}" | tee -a "$GITHUB_ENV"
+          echo "NEW_REF=${{ github.event.pull_request.head.sha }}" | tee -a "$GITHUB_ENV"
+      
+      - name: Check out Kani (old variant)
+        uses: actions/checkout@v4
+        with:
+          path: ./old
+          ref: ${{ env.OLD_REF }}
+          fetch-depth: 2
+
+      - name: Check out Kani (new variant)
+        uses: actions/checkout@v4
+        with:
+          path: ./new
+          ref: ${{ env.NEW_REF }}
+          fetch-depth: 1
+
+      - name: Set up Kani Dependencies (old variant)
+        uses: ./old/.github/actions/setup
+        with:
+          os: ubuntu-24.04
+          kani_dir: old
+
+      - name: Set up Kani Dependencies (new variant)
+        uses: ./new/.github/actions/setup
+        with:
+          os: ubuntu-24.04
+          kani_dir: new
+
+      - name: Copy benchmarks from new to old
+        run: rm -rf ./old/tests/perf ; cp -r ./new/tests/perf ./old/tests/
+
+      - name: Build `compile-timer` in old
+        run: cd old/tools/compile-timer && cargo build --release
+      - name: Build `kani` in old
+        run: cd old && cargo build-dev --release
+
+      - name: Build `compile-timer` in new
+        run: cd new/tools/compile-timer && cargo build --release
+      - name: Build `kani` in new
+        run: cd new && cargo build-dev --release
+
+      - name: Run `compile-timer` on old
+        run: |
+          export PATH="${{ github.workspace }}/old/scripts:$PATH"
+          cd old/tests/perf/s2n-quic && ../../../target/release/compile-timer --out-path compile-times-old.json --also-visit quic/s2n-quic-core --also-visit quic/s2n-quic-platform --also-visit common/s2n-codec --skip-current
+      - name: Run `compile-timer` on new
+        run: |
+          export PATH="${{ github.workspace }}/new/scripts:$PATH"
+          cd new/tests/perf/s2n-quic && ../../../target/release/compile-timer --out-path compile-times-new.json --also-visit quic/s2n-quic-core --also-visit quic/s2n-quic-platform --also-visit common/s2n-codec --skip-current
+      - name: Run analysis between the two
+        run: ./new/target/release/compile-analyzer --path-pre old/tests/perf/s2n-quic/compile-times-old.json --path-post new/tests/perf/s2n-quic/compile-times-new.json --only-markdown --suite-name long >> "$GITHUB_STEP_SUMMARY"
--- a/.github/workflows/bench-compiler.yml
+++ b/.github/workflows/bench-compiler.yml
@@ -1,11 +1,7 @@
 # Copyright Kani Contributors
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 #
-# Run performance benchmarks comparing two different Kani versions.
-# This workflow takes much longer than other workflows, so we don't run it by default.
-# This workflow will run when:
-#  - Changes are pushed to 'main'.
-#  - Triggered by another workflow
+# Run performance benchmarks comparing the compiler performance of two different Kani versions.
 name: Kani Compiler Performance Benchmarks
 on:
   push:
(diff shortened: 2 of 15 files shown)


AlexanderPortland:
I also ran this workflow on an old, slightly performance-regressing PR (#3374), and it successfully detects the small regression with [this output](https://github.com/AlexanderPortland/kani/blob/recreate-artems-branch/results.md).

carolynzech:
I haven't reviewed the code yet, but a thought--I think we'll get more stable results out of this if we run on more LoC--the perf tests are a good starting point (and a way to get good feedback), but given that we're measuring compilation times I'd imagine the results would be more stable if we had more to compile.

I'm not opposed to having a short job for quick feedback, though. Are you planning to add a longer one later?

carolynzech:
(I realize that I told you to start on the perf tests, so it's not a criticism 😁 ) I think we should also test on some bigger codebases too.

AlexanderPortland:
@carolynzech I've added an additional `compiler-timing-long` job that runs on all the `s2n-quic` harnesses and takes about ~90s for each full compilation run (compared to the `-short` job which only takes ~3s). This should reduce noise even more with the tradeoff of taking about 50 minutes to run in CI rather than 15.

AlexanderPortland:
[Here's](https://github.com/AlexanderPortland/kani/actions/runs/15766783026) the link to the newest CI run on my branch that has updated significance thresholds and small formatting changes. Should be finally ready for review now!

.github/labeler.yml
```diff
@@ -5,9 +5,20 @@
...
     - any-glob-to-any-file: ['kani-dependencies']
+
+Z-CompilerBenchCI:
+- any:
+  # we want to run compiler benchmarks if:
+  - changed-files:
+    # any parts of the compiler change
+    - any-glob-to-any-file: ['kani-compiler/**', 'cprover_bindings/**', 'library/**']
+    # the way we call the compiler through cargo changes
+    - any-glob-to-any-file: ['kani-driver/src/call_cargo.rs']
```
carolynzech:
```suggestion
    # the way we call the compiler changes
    - any-glob-to-any-file: ['kani-driver/src/call_cargo.rs', 'kani-driver/src/call_single_file.rs']
```

.github/labeler.yml
```diff
@@ -5,9 +5,20 @@
...
+    # or if our dependencies change
+    - any-glob-to-any-file: ['rust-toolchain.toml', 'Cargo.lock', 'kani-dependencies']
```
carolynzech:
```suggestion
    - any-glob-to-any-file: ['rust-toolchain.toml', 'Cargo.lock']
```
`kani-dependencies` is just for our CBMC version, which would only affect verification time (unless a change to the CBMC version also requires some change to our goto file format, but that would show up as a compiler change).

.github/workflows/bench-compiler.yml
```diff
@@ -0,0 +1,142 @@
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+# Run performance benchmarks comparing two different Kani versions.
+# This workflow takes much longer than other workflows, so we don't run it by default.
+# This workflow will run when:
+#  - Changes are pushed to 'main'.
+#  - Triggered by another workflow
```
carolynzech:
```suggestion
# Run performance benchmarks comparing the compiler performance of two different Kani versions.
```

.github/workflows/bench-compiler.yml
```diff
@@ -0,0 +1,142 @@
...
+name: Kani Compiler Performance Benchmarks
+on:
+  push:
+    branches:
+      - 'main'
+  workflow_call:
+
+jobs:
+  compile-timer-short:
+    runs-on: ubuntu-24.04
```
carolynzech, line 14:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/bench-compiler.yml
```diff
@@ -0,0 +1,142 @@
...
+          kani_dir: old
+
+      - name: Set up Kani Dependencies (new variant)
+        uses: ./new/.github/actions/setup
+        with:
+          os: ubuntu-24.04
+          kani_dir: new
+
+      - name: Copy benchmarks from new to old
+        run: rm -rf ./old/tests/perf ; cp -r ./new/tests/perf ./old/tests/
```
carolynzech, line 53: The point being to make sure they're running the same benchmarks, if the PR changes a benchmark? Perhaps add a comment to this effect.

tools/compile-timer/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+By default, the script recursively goes into directories and will use `cargo kani` to profile any Rust projects it encounters (which it determines by looking for a `Cargo.toml`). You can tell it to ignore specific subtrees by passing in the `--ignore [DIR_NAME]` flag.
+
+## Visualizing Compiler Times with `compile-analyzer`
+`compile-timer` itself will have some debug output including each individual run's time and aggregates for each crate and the whole analysis, but `compile-analyzer` is specifically for comparing performance across multiple commits. Once you've run `compile-timer` on both commits, you can run `compile-analyzer --path-pre [FIRST_JSON_FILE] --path-post [SECOND_JSON_FILE]` to see the change in performance going from the first to second commit. By default, `compile-analyzer` will just print to the console, but if you specify the `--only-markdown` option, it's output will be formatted for GitHub flavored markdown (as is useful in CI).
```
carolynzech:
```suggestion
`compile-timer` itself will have some debug output including each individual run time and aggregates for each crate. 
`compile-analyzer` is specifically for comparing performance across multiple commits. 
Once you've run `compile-timer` on both commits, you can run `compile-analyzer --path-pre [FIRST_JSON_FILE] --path-post [SECOND_JSON_FILE]` to see the change in performance going from the first to second commit. 
By default, `compile-analyzer` will just print to the console, but if you specify the `--only-markdown` option, it's output will be formatted for GitHub flavored markdown (as is useful in CI).
```

tools/compile-timer/src/common.rs
```diff
@@ -0,0 +1,66 @@
...
+use serde::{Deserialize, Serialize};
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct AggrResult {
+    pub krate: PathBuf,
+    pub krate_trimmed_path: String,
+    /// the stats for only the 25th-75th percentile of runs on this crate
+    pub iqr_stats: Stats,
```
carolynzech:
```suggestion
    /// the stats for only the 25th-75th percentile of runs on this crate, i.e., the interquartile range
    pub iqr_stats: Stats,
```

tools/compile-timer/src/common.rs
```diff
@@ -0,0 +1,66 @@
...
+        "{:?}",
+        krate
+            .canonicalize()
+            .unwrap()
+            .strip_prefix(std::env::current_dir().unwrap().parent().unwrap())
+            .unwrap()
+    )
+}
+
+#[allow(dead_code)]
```
carolynzech: I would just do a crate-level `#![allow(dead_code)]`, but up to you

tools/compile-timer/src/compile-timer.rs
```diff
@@ -0,0 +1,186 @@
...
+    Duration::from_micros(micros)
+}
+
+fn aggregate_results(path: &Path, results: &[Duration]) -> AggrResult {
+    assert!(results.len() == TIMED_RUNS);
+
+    // sort and calculate the subset of times in the interquartile range
+    let mut sorted = results.to_vec();
+    sorted.sort();
+    let iqr_bounds = (0.25 * results.len() as f64, 0.75 * results.len() as f64);
```
carolynzech: Why cast this to an `f64` if you end up casting back to a `usize` later?

tools/compile-timer/src/compile-analyzer.rs
```diff
@@ -0,0 +1,232 @@
...
+    let post_results = post_ser.into_iter::<AggrResult>().collect::<Vec<_>>();
+
+    let mut results = pre_results
+        .into_iter()
+        .filter_map(Result::ok)
+        .zip(post_results.into_iter().filter_map(Result::ok))
+        .collect::<Vec<_>>();
+    results.sort_by_key(|a| {
+        -(signed_percent_diff(&a.0.iqr_stats.avg, &a.1.iqr_stats.avg).abs() * 1000_f64) as i64
+    });
```
carolynzech: Can you write a comment explaining what this does?

tools/compile-timer/src/compile-analyzer.rs
```diff
@@ -0,0 +1,232 @@
...
+    let total_pre = results.iter().map(|i| i.0.iqr_stats.avg).sum();
+    let total_post = results.iter().map(|i| i.1.iqr_stats.avg).sum();
+    println!(
+        "### *on the whole: {:.2?} → {:.2?} —* {}",
+        total_pre,
+        total_post,
+        diff_string(total_pre, total_post)
+    );
+    // Note that we have to call the fourth column "heterogeneousness" because the color-formatted
+    // diff will cut off if the column isn't wide enough for it, so verbosity is required.
```
carolynzech, line 88: Could we possibly do "percentage difference" to achieve the width? I was confused by "heterogeneousness" when I saw it.
But it's not a huge deal if not.

tools/compile-timer/src/compile-analyzer.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    let avg_std_dev = (pre.full_std_dev() + post.full_std_dev()) / 2;
+    if post_time.abs_diff(pre_time) < fraction_of_duration(avg_std_dev, FRAC_STD_DEV_THRESHOLD) {
+        return Verdict::ProbablyNoise(NoiseExplanation::SmallComparedToStdDevOf(avg_std_dev));
+    }
+
+    if post_time.abs_diff(pre_time) < fraction_of_duration(pre_time, FRAC_ABSOLUTE_THRESHOLD) {
+        return Verdict::ProbablyNoise(NoiseExplanation::SmallPercentageChange);
+    }
```
carolynzech: I think let's check the percentage before standard deviation; that measurement is more intuitive to the average reader than standard deviation, so if it takes effect first that's better.

carolynzech: Wouldn't it make sense to check for noise before returning improvements? So that we also report if a suspected improvement is just noise too.

tools/compile-timer/src/compile-analyzer.rs
```diff
@@ -0,0 +1,232 @@
...
+            "krate {:krate_column_len$} -- [{:.2?} => {:.2?} ({change_dir}{change_amount:5.2}%)] {:?}",
+            pre_res.krate_trimmed_path,
+            pre_time,
+            post_time,
+            verdict_on_change(pre_res, post_res)
+        );
+    }
+}
+
+fn verdict_on_change(pre: &AggrResult, post: &AggrResult) -> Verdict {
```
carolynzech, line 171: I think this could use a doc comment explaining the logic.

tools/compile-timer/src/compile-analyzer.rs
carolynzech: After you do that, can you do a couple more test runs? I'm struggling to think of a situation where we would expect the 5% check to miss something that the standard deviation check catches. If we can think of one, let's add a comment explaining why; otherwise, it's likely simpler to just remove the standard deviation stuff altogether.

carolynzech: I think I was too in the mindset of the end to end benchmarks, where we expect pretty large fluctuations from run to run, and in which case some measurement of that deviation is useful to detect noise. From looking at the logs of your fork runs, it seems like compilation times deviate much less between runs (which makes sense, in hindsight), so I don't know that the extra complexity is worth it. LMK if you disagree; we can always keep it and gate it behind an option or something, but I'm inclined to remove it for now and introduce more complex math in a future PR if needed.

carolynzech (commented):
Thanks! Be sure to also update the PR description with whatever the new runs on your fork look like.

carolynzech:
[Here](https://github.com/AlexanderPortland/kani/actions/runs/15788016660?pr=6) is a run of the workflow on a PR that we expect to cause a small-to-medium regression. I tried to create a change that isn't overly contrived. The results gives me confidence that the workflow is at a good spot with sensitivity--it catches the regression comfortably, without failing on every harness (which might suggest it's overly sensitive).

tools/compile-timer/src/compile-analyzer.rs
AlexanderPortland: Good point, this specific wording was one of the least clear options, but markdown will split multiple words onto multiple lines to shorten the column, so I was looking for a synonym with the longest possible single word. Maybe there's something I can do with HTML syntax to get the same effect without making the wording less clear? I can take a look.

.github/workflows/bench-compiler.yml
AlexanderPortland: Should this change also modify the `setup` action to run with `os: ubuntu-latest`? Based on how we use the `os` flag [to look for a setup script](https://github.com/model-checking/kani/blob/b64e59de669cd77b625cc8c0b9a94f29117a0ff7/.github/actions/setup/action.yml#L32-L34), I don't think that would work. But, if we don't, then when `ubuntu-latest` is no longer `ubuntu-24.04` is it possible we'll start running an incorrect setup script for the OS our runner is using?

tools/compile-timer/src/compile-analyzer.rs
AlexanderPortland: Yes, you're totally right. I can't think of a situation with our current thresholds where 5% would be less than twice the standard deviation (unless there's a ton of noise), so I think the std deviation test is pretty redundant as is. This was more a thing from the initial thresholds where 1.5x the std deviation was the primary detection method and 1% was just a backup for runs with very low noise. I agree we should remove the std deviation check for now and bring something like it back later if needed!

tools/compile-timer/src/compile-timer.rs
AlexanderPortland: I had done this intentionally so that we could multiply the two floats together before converting to an integer, but looking back I think `results.len() / 4` is equivalent to `results.len() as f64 * 0.25`. I'll change to that

Issue #4168, opened 2025-06-19
Bloated MIR remains for `Result` functions


AlexanderPortland:
**Proposed change:**
Maybe some strategy for stubbing all string formatting could be helpful for bloat and not affect our correctness? At the very least finding a way to stub `unwrap_failed()` specifically would help these instances.

**Motivation:**
`Result::unwrap()` and `Result::expect()` seem to have unique call trees going through the [`core::result::unwrap_failed()`](https://github.com/rust-lang/rust/blob/8de4c7234dd9b97c9d76b58671343fdbbc9a433e/library/core/src/result.rs#L1760-L1762) that still cause significant MIR bloat even with panic stubbing. This seems to be due to the way the `panic!()` macro is called, causing it to expand into calls to display and debug format functions as well as the actual call to `core::panic()` that is eventually stubbed.

carolynzech:
The problem appears to more generally happen if `panic` uses `{}` formatting, e.g.:

```rust
let x = 0;
panic!("{x} is zero");
```

The formatting code for `x` gets called _before_ panic. `unwrap_failed` has this problem because its body is:
```rust
panic!("{msg}: {error:?}")
```
i.e., `msg` and `error` get formatted with `{}`.

I suspect that we could fix this if we could find a way to stub out the `Arguments` constructor functions called by the `format_args` [macro](https://doc.rust-lang.org/std/macro.format_args.html).

AlexanderPortland:
It seems that the issue is the internal use of the `core::panic!()` macro in `unwrap_failed()`. When a crate uses `panic!()` directly, the Kani compiler builds it with an `--extern` flag that will overwrite std (and the `panic!()` macro) with [our own version](https://github.com/model-checking/kani/blob/96eb502ffce80c3408c79d18620c84e364be9310/library/std/src/lib.rs#L203) that has no formatting code. However, this doesn't overwrite the macro in dependencies, so the `unwrap_failed()` function will still be compiled with the base version of std and use the real `core::panic!()` macro's complex formatting.

PR #4169, opened 2025-06-19
Stub panics during MIR transformation

### LINKED ISSUE #2835: Bloated MIR due to panic code in standard library
**Proposed change:**
Reconsider the way in which `panic` code is handled.
For example, when an `assert` statement is added, the `panic` call is ignored. This, however, is not the case when there are `panic` calls in standard library functions that are used in a verified program.

**Motivation:**
See the attached example, which is a variant of the `switchInt` example from Kani's tests directory.
The use of iterator pulls in standard library code that calls the `panic` function. It seems that due to that, a lot of code that is irrelevant for verification. This causes the internal go-to representation in CBMC to include a lot of irrelevant information. This may hinder performance.
[main1.rs.zip](https://github.com/model-checking/kani/files/13111362/main1.rs.zip)

AlexanderPortland:
Modifies the `RustcIntrinsicsPass` to stub out panic functions during body transformation at the MIR level. This avoids having to bring in (and do transformation passes on) the large amounts of code they use for string formatting. Our panic hook has been modified to then detect these stubs during codegen and still generate the same Goto code for them.

The results of this change on the MIR bloat of `panic!()` and its callers is shown below:
| test | lines of MIR (before) | lines of MIR (w/ panic stubbing) |
| - | - | - |
| `Option<usize>::unwrap()` | 292 | 217 |
| `Option<usize>::expect()` | 8664 | 225 | 
| `Result<usize, usize>::unwrap()` | 20461 | 20374 |
| `Result<usize, usize>::expect()` | 20463 | 20376 | 
| `panic!("hello!")` | 8498 | 41 |
| `panic!("hello {x}!")` (x being a `&'static str`) | 8502 | 45 |

The lack of significant improvement on the two `Result` functions is elaborated on in #4168 and will take additional consideration to fix.

Resolves #2835

### DIFF
--- a/kani-compiler/src/kani_middle/transform/rustc_intrinsics.rs
+++ b/kani-compiler/src/kani_middle/transform/rustc_intrinsics.rs
@@ -174,11 +174,12 @@ impl MutMirVisitor for ReplaceIntrinsicCallVisitor<'_, '_> {
             && let TyKind::RigidTy(RigidTy::FnDef(def, args)) =
                 func.ty(&self.locals).unwrap().kind()
         {
-            if def.is_intrinsic() {
+            // Get the model we should use to replace this function call, if any.
+            let replacement_model = if def.is_intrinsic() {
                 let instance = Instance::resolve(def, &args).unwrap();
                 let intrinsic = Intrinsic::from_instance(&instance);
                 debug!(?intrinsic, "handle_terminator");
-                let model = match intrinsic {
+                match intrinsic {
                     Intrinsic::SizeOfVal => self.models[&KaniModel::SizeOfVal],
                     Intrinsic::MinAlignOfVal => self.models[&KaniModel::AlignOfVal],
                     Intrinsic::PtrOffsetFrom => self.models[&KaniModel::PtrOffsetFrom],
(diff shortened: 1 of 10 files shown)


kani-compiler/src/kani_middle/transform/rustc_intrinsics.rs
```diff
@@ -159,27 +173,40 @@ impl MutMirVisitor for ReplaceIntrinsicCallVisitor<'_> {
...
+                let span = term.span;
+                let new_func = ConstOperand { span, user_ty: None, const_: literal };
+                *func = Operand::Constant(new_func);
+                self.changed = true;
+            } else if is_panic_function(&self.tcx, def.0) {
+                // if we find a panic function, replace it with our stub
+                let new_instance =
+                    Instance::resolve(self.models[&KaniModel::PanicStub], &args).unwrap();
+
+                let literal = MirConst::try_new_zero_sized(new_instance.ty()).unwrap();
```
zhassan-aws: Can you add a comment that describes what this is doing?

AlexanderPortland: Okay, I just did! I also tried to reorganize the code to make it fit better with the existing logic for replacing function calls, so let me know if you have suggestions

AlexanderPortland:
@zhassan-aws I just added a test that runs `kani` on three of the `panic!()` examples from above and checks to make sure that the emitted MIR properly contains our panic stub.

tests/script-based-pre/mir_stub_panic/expected
```diff
@@ -0,0 +1,52 @@
...
+}
+
+
+--VERIFYING option.rs--
+Manual Harness Summary: \
+Verification failed for - main \
+Complete - 0 successfully verified harnesses, 1 failures, 1 total.
+
+--READING MIR for option.rs--
+fn kani::rustc_intrinsics::panic_stub(_1: &str) -> ! { \
```
zhassan-aws: Due to how the expected tests work (they check if this string is contained in the output), this will succeed if the panic stub appears in _any_ of the 3 tests. We could perhaps add some prefix when executing the cat commend (e.g using `sed`) to make sure each one contains the panic stub.

AlexanderPortland: 🤦‍♂️ Sorry, that's totally my bad. I thought the order of the expected lines also had to be maintained in the output. I'll fix that later today!

zhassan-aws: Cool. The only mechanism to enforce the order is to add a backslash at the end of the line (as you did for the panic stub function). Otherwise, it's just a containment check.

AlexanderPortland: Just added prefixes using `sed` like you suggested!

PR #4174, opened 2025-06-23
Add support for loop assigns in loop contracts

### LINKED ISSUE #3871: Support loop modifies in loop contracts
Requested feature: Allowing users to specify loop modifies instead of inferring them.
Link to relevant documentation (Rust reference, Nomicon, RFC): #3167 

Test case:

```rust

#[kani::proof]
fn simple_while_loop_with_old_harness() {
    let mut x: u8 = kani::any_where(|i| *i >= 2);
    let mut y: u8 = x;

    #[kani::loop_invariant(x >= 2 && y <= old(y)]
    #[kani::loop_modifies(x)]
    while x > 2 {
        x = x - 1;
    }

    assert!(x == 2);
}
```

thanhnguyen-aws:
This PR adds support for loop assigns in loop contracts.

Resolves #3871

### DIFF
--- a/library/kani_macros/src/lib.rs
+++ b/library/kani_macros/src/lib.rs
@@ -454,7 +454,7 @@ mod sysroot {
     mod loop_contracts;
 
     pub use contracts::{ensures, modifies, proof_for_contract, requires, stub_verified};
-    pub use loop_contracts::{loop_modifies, loop_invariant};
+    pub use loop_contracts::{loop_invariant, loop_modifies};
 
     use super::*;
 


library/kani_macros/src/lib.rs
```diff
@@ -440,6 +440,10 @@ pub fn loop_invariant(attr: TokenStream, item: TokenStream) -> TokenStream {
     attr_impl::loop_invariant(attr, item)
 }
 
+#[proc_macro_attribute]
+pub fn loop_assigns(attr: TokenStream, item: TokenStream) -> TokenStream {
+    attr_impl::loop_assigns(attr, item)
+}
```
tautschnig: Can you please add documentation for `loop_assigns` to docs/src/reference/experimental/loop-contracts.md?

carolynzech:
Could you update our tutorial documentation along with this?

carolynzech:
Also, I'm curious why we're calling it `loop_assigns` rather than `loop_modifies`--I know CBMC says `assigns`, but IMO it's better to be consistent with our function contracts `modifies` clause.

thanhnguyen-aws:
> Could you update our tutorial documentation along with this?

I did

thanhnguyen-aws:
> Also, I'm curious why we're calling it `loop_assigns` rather than `loop_modifies`--I know CBMC says `assigns`, but IMO it's better to be consistent with our function contracts `modifies` clause.

It reminds me of the `loop_invariant` or `loop_contract` stuffs that we discussed long time ago. I think we should discuss with more team members to make the decision whether we should separate or unify loop and function terminologies. Then we may create a separate PR for it.

carolynzech:
> It reminds me of the loop_invariant or loop_contract stuffs that we discussed long time ago. I think we should discuss with more team members to make the decision whether we should separate or unify loop and function terminologies. Then we may create a separate PR for it.

Hm, I disagree--the difference there was that the loop invariant vs loop contract disparity was already in our code, and we were discussing which term of those to choose and to try to stick to it. This change hasn't been merged yet, so I don't see a reason to do loop assigns when we've already established modifies for contracts. Seems needlessly confusing.

PR #4177, opened 2025-06-24
Optimize reachability with non-mutating global passes

AlexanderPortland:
## Context
When transforming MIR, we use both `TransformPass`-es that operate over a single body and `GlobalPass`-es that require further context thus and operate over the whole program. Since these global passes could potentially modify the bodies they operate over, we currently re-run reachability after them to ensure that any changes are reflected in our reachability output.

However, our global passes are largely gated behind unstable features (e.g. `-Z uninit-checks` to check for uninitialized memory) or specific user actions (e.g. compiling w/ `RUSTFLAGS="--emit mir"`), so this extra reachability collection is often done unnecessarily. Furthermore, some global passes (like emitting the generated MIR) do not modify the bodies they interact with at all, so re-doing collection is unneeded even when they are enabled.

## Changes
This PR modifies the `GlobalPass` trait to return a boolean representing if the pass modified the MIR in a way that could affect reachability. We then use the boolean flags for each result to redo reachability analysis _only if a `GlobalPass` could have affected its result._

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/compiler_interface.rs
@@ -87,7 +87,7 @@ impl GotocCodegenBackend {
         // disadvantage of not having a precomputed call graph for the global passes to use. The
         // call graph could be used, for example, in resolving function pointer or vtable calls for
         // global passes that need this.
-        let (items, call_graph) = with_timer(
+        let (mut items, call_graph) = with_timer(
             || collect_reachable_items(tcx, &mut transformer, starting_items),
             "codegen reachability analysis",
         );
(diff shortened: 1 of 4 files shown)


AlexanderPortland:
For context, I intentionally implemented this as separate traits rather than having `GlobalPass::transform()` return a boolean as we do for [`TransformPass::transform()`](https://github.com/model-checking/kani/blob/96eb502ffce80c3408c79d18620c84e364be9310/kani-compiler/src/kani_middle/transform/mod.rs#L172) (this seemed more explicit and elegant type-wise), but I'm very open to feedback and can switch if the latter seems preferable!

tautschnig (dismissed):

AlexanderPortland:
I actually just switched this to use the same boolean strategy as `TransformPass`. I love my `RefCell<T>`s, but this made the changes a lot cleaner without the same runtime overhead.

carolynzech (approved):
Thanks!

PR #4179, opened 2025-06-26
Fix the bug that assign clause cannot be inferred for the inner loop of nested loops

### LINKED ISSUE #4159: Loop Contract Issue: Check some_var is assignable
I tried this code:

```rust

fn func(x: usize) {
    let mut j = 0;
    const CHUNK_SIZE: usize = 32;
    #[kani::loop_invariant(j <=x )]
    while j < x {
        let mut i = 0;
        let mut sum = 0_usize;
        #[kani::loop_invariant(i <= CHUNK_SIZE && prev(i) <= CHUNK_SIZE && prev(i) + 1 == i && sum <= CHUNK_SIZE && on_entry(sum) + i >= sum )]
        while i < CHUNK_SIZE {
            sum += kani::any::<bool>() as usize;
            i += 1;
        }

        j += 1;
    }
}

#[kani::proof]
fn harness() {
    let _ = func(10);
}

fn main() {}
```

```
kani src/main.rs -Z -loop-contracts -Z function-contracts 
```

with Kani version: 0.63.0 (branch `main`, commit: `8d18568`)

Instead, this happened: 
```
 ** 3 of 235 failed (3 unreachable)
Failed Checks: Check assigns clause inclusion for loop func.0
 File: "src/main.rs", line 11, in func
Failed Checks: Check that *var_9 is assignable
 File: "src/main.rs", line 13, in func::{closure#1}
Failed Checks: Check that *var_12 is assignable
 File: "src/main.rs", line 14, in func::{closure#1}

VERIFICATION:- FAILED
Verification Time: 0.15695447s
```

The code has nested loops, and the errors are associated with the inner loop. From my understanding, `var_9` represents `sum`, and `var_12` represents `i`. I don't know how to resolve these errors. Perhaps a `loop_modifies` clause is required for this loop invariant?

thanhnguyen-aws:
This PR fixes the bug that assign clause cannot be inferred for the inner loop of nested loops.

General ideas: For local variables of the outer loop, after the loop-contracts transformation, if their StorageLive/StorageDead statements are still inside the outer loop body, CBMC cannot infer the assign clause for the inner loop if those variables are assigned inside the inner loop's body. We have to move their StorageLive to the loop head of the outer-loop and their StorageDead to the block right after where the outer loop terminates. Together with the StorageLive, we have to initiate the variables with the values they are initiated with inside the outer loop's body. Note that we handle three ways how a variable is initiated in MIR: by an assign statement following after a StorageLive one, by a Terminator for function call, and by a group of five statements which perform the `clone()` function.

Resolves #4159

### DIFF
--- a/kani-compiler/src/kani_middle/transform/loop_contracts.rs
+++ b/kani-compiler/src/kani_middle/transform/loop_contracts.rs
@@ -17,7 +17,7 @@ use rustc_span::Symbol;
 use stable_mir::mir::mono::Instance;
 use stable_mir::mir::{
     AggregateKind, BasicBlock, BasicBlockIdx, Body, ConstOperand, Operand, Rvalue, Statement,
-    StatementKind, Terminator, TerminatorKind, VarDebugInfoContents,
+    StatementKind, SwitchTargets, Terminator, TerminatorKind, VarDebugInfoContents,
 };
 use stable_mir::ty::{FnDef, GenericArgKind, MirConst, RigidTy, TyKind, UintTy};
 use std::collections::{HashMap, HashSet, VecDeque};
@@ -30,6 +30,42 @@ pub struct LoopContractPass {
     /// The map from original loop head to the new loop latch.
     /// We use this map to redirect all original loop latches to a new single loop latch.
     new_loop_latches: HashMap<usize, usize>,
+}
+
+//Useful for debugging
+#[allow(dead_code)]
+pub fn print_stable_mir_body(body: &Body) {
+    println!("=== MIR Body Start ===");
+    // Print basic blocks
+    for (idx, block) in body.blocks.iter().enumerate() {
+        println!("bb{idx}: {{");
+        // Print statements
+        for stmt in &block.statements {
+            println!(" {stmt:?}");
+        }
+        // Print terminator
+        println!(" {:?}", block.terminator);
+        println!("}}");
+    }
+    println!("=== MIR Body End ===");
+}
+
+//Useful for debugging
+#[allow(dead_code)]
+pub fn print_mutablebody(body: &MutableBody) {
+    println!("=== MIR Body Start ===");
+    // Print basic blocks
+    for (idx, block) in body.blocks().iter().enumerate() {
+        println!("bb{idx}: {{");
+        // Print statements
+        for stmt in &block.statements {
+            println!(" {stmt:?}");
+        }
+        // Print terminator
+        println!(" {:?}", block.terminator);
+        println!("}}");
+    }
+    println!("=== MIR Body End ===");
 }
 
 impl TransformPass for LoopContractPass {
@@ -206,18 +242,20 @@ impl LoopContractPass {
         }
     }
 
+    //Get all loop-positions: (loop_head_id, loop_latch_id) in the body
     fn get_loop_positions(&self, body: &MutableBody, tcx: TyCtxt) -> Vec<(usize, usize)> {
         let mut loop_pos: Vec<(usize, usize)> = Vec::new();
         for (block_idx, _) in body.blocks().iter().enumerate() {
             if self.is_loop_head(body, tcx, block_idx) {
-                let loop_latch_id = self.get_loop_latch_id(body, block_idx);
+                let loop_latch_id = self.get_last_loop_latch_id(body, block_idx);
                 loop_pos.push((block_idx, loop_latch_id));
             }
         }
         loop_pos
     }
 
-    fn get_closest_loop_head(
+    //Get the associated loop-head of a block_id
+    fn get_associated_loop_head(
         &self,
         block_idx: usize,
         loop_positions: &Vec<(usize, usize)>,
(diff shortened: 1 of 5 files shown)


kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -32,6 +32,42 @@ pub struct LoopContractPass {
...
+        for stmt in &block.statements {
+            println!(" {stmt:?}");
+        }
+        // Print terminator
+        println!(" {:?}", block.terminator);
+        println!("}}");
+    }
+    println!("=== MIR Body End ===");
+}
+
```
tautschnig: I acknowledge those might be useful, but can we please have them in some place that is a better fit for general utilities independent of loop contracts?

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -156,6 +192,326 @@ impl LoopContractPass {
         body.replace_statements(&SourceInstruction::Terminator { bb: bb_idx }, stmts);
     }
 
+    fn get_user_defined_variables(&self, body: &MutableBody) -> Vec<usize> {
```
tautschnig, line 162: Could you please comments to newly introduced functions. Here, I am specifically looking to understand what is meant by "user" defined variables.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -156,6 +192,326 @@ impl LoopContractPass {
...
+                loop_head_map.insert(block_idx, loop_head);
+            }
+        }
+        loop_head_map
+    }
+
+    ///In case of nested loop, if a variable is declared and initiated inside a loop body, and assigned inside an inner-loop,
+    ///then CBMC cannot infer the assign clause for the inner-loop after the loop-contract transformation.
+    //Move all variables initiation using assign inside the loop body to the loop-head
+    fn move_storagelive_assign_to_loophead(
```
tautschnig, line 259: I have now read this in the PR description and also the similar comment here, but what isn't yet clear to me is the following: Why is is correct to move a variable's initialization into the inner loop (imagine a scenario where that variable is used between its original initialization and before the start of the inner loop)? My understanding is that moving StorageLive in the way that you do really turns a variable defined (and possibly used!) in the outer loop into a variable that is restricted to the inner loop.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -169,6 +525,11 @@ impl LoopContractPass {
     /// It is the core of fn transform, and is separated just to avoid code repetition.
     fn transform_body_with_loop(&mut self, tcx: TyCtxt, body: Body) -> (bool, Body) {
         let mut new_body = MutableBody::from(body);
+        let loop_head_map = self.get_associated_loop_head_hashmap(&new_body, tcx);
+        let found_local_list =
+            self.move_storagelive_assign_to_loophead(&mut new_body, &loop_head_map);
+        //print_mutablebody(&new_body);
+        //assert!(1==0);
```
tautschnig: Nit pick: I'd suggest removing these commented-out lines.

tests/expected/loop-contract/nested_loop_local_var_assign.rs
```diff
@@ -0,0 +1,31 @@
...
+#![feature(proc_macro_hygiene)]
+#![feature(stmt_expr_attributes)]
+
+fn func(x: usize) {
+    let mut j = 0;
+    const CHUNK_SIZE: usize = 32;
+    #[kani::loop_invariant(j <=x )]
+    while j < x {
+        let mut i = 0;
+        let mut sum = 0_usize;
```
tautschnig, line 17: To create a concrete example of my question above: would this test still pass if you changed this code to be:
````rust
let mut i = 0;
let mut sum : usize = i;
````

kani-compiler/src/kani_middle/transform/loop_contracts.rs
carolynzech: This should not be in Kani; StableMIR already has APIs to pretty-print bodies (you can just do `dbg!(body)` and that works fine). Kani can also [dump reachable MIR bodies](https://model-checking.github.io/kani/cheat-sheets.html#debug). So we should just delete this code entirely.

carolynzech: See https://github.com/rust-lang/rust/blob/master/compiler/rustc_smir/src/stable_mir/mir/pretty.rs for the StableMir implementation that makes `dbg!` calls with StableMir constructs render well.

kani-compiler/src/kani_middle/transform/loop_contracts.rs
thanhnguyen-aws: I have updated the description to clear out the confusion. We do NOT move the variables' initializations into the inner loop, but to the loop-head of the outer-loop. Specifically, when a variable is initialized inside a loop-body by a pair of commands `StorageLive(x)` followed by `x = ...`, we add exactly the two commands to the loop-head, and remove the `StorageLive(x)`command inside the loop body (we keep the assign command in the loop-body). Doing this way, we do not change the semantic of the code.

tests/expected/loop-contract/nested_loop_local_var_assign.rs
thanhnguyen-aws: Yes, I tested and it passed.

Issue #2235, opened 2023-02-25
hangs on string.repeat() ?


matthiaskrgr:
I tried this code:

```rust
#[kani::proof]  fn main() {
	
	let x = String::new().repeat(1);
	let z = x.chars().nth(1).unwrap();
}
```

using the following command line invocation:

```
RUSTC_WRAPPER="" cargo kani --harness  main --tests
```

with Kani version: 0.22.0

I expected to see this happen: *explanation*
It looks like there was some kind of hang..? kani/cmbc was crunching on that for more than 5 minutes until I cancelled it.
If I remove the `repeat(1)` , it finishes within a second :thinking:

zhassan-aws:
Hi @matthiaskrgr. Kani currently performs poorly on programs involving strings (e.g. see #2001 and #2015). The issue here seems to be a combination of using `repeat` and `chars` because this program that involves `repeat` verifies quickly:
```rust
#[kani::proof]
fn main() {
    let x = String::new().repeat(1);
    assert!(x.is_empty());
}
```
```

SUMMARY:
 ** 0 of 322 failed (9 unreachable)

VERIFICATION:- SUCCESSFUL
Verification Time: 0.68734515s
```
The `chars` method in particular deals with the UTF-8 encoding, which seems difficult to reason about. If it's possible to express properties in term of bytes, Kani may perform better. For instance, this program is verified in ~1 second:
```rust
#[kani::proof]
fn main() {
    let x = String::from("abc").repeat(5);
    let bytes = x.as_bytes();
    assert_eq!(bytes.len(), 15);
}
```
```

SUMMARY:
 ** 0 of 322 failed (3 unreachable)

VERIFICATION:- SUCCESSFUL
Verification Time: 1.2358496s
```

zhassan-aws:
CC @tautschnig. This doesn't seem to get to the SAT solver stage after running for 20 minutes.

zhassan-aws:
This is the CBMC call stack after running for an hour:
```
#270 0x0000556fa14997f3 in irept::compare(irept const&) const ()
#271 0x0000556fa14997f3 in irept::compare(irept const&) const ()
#272 0x0000556fa14997f3 in irept::compare(irept const&) const ()
#273 0x0000556fa14997f3 in irept::compare(irept const&) const ()
#274 0x0000556fa14997f3 in irept::compare(irept const&) const ()
#275 0x0000556fa14997f3 in irept::compare(irept const&) const ()
[...]

PR #4180, opened 2025-06-26
`ptr_offset_from`: Replace arithmetic over pointers by offset arithmetic

tautschnig:
Makes `pointer_offset` crate-available just like `pointer_object` is. This enables its use in `ptr_offset_from`, whereby we create expressions that CBMC's simplifier can more readily deal with.

Resolves: #2235


tests/expected/shadow/unsupported_object_size/test.rs
```diff
@@ -5,12 +5,17 @@
 // This test checks the maximum object size supported by Kani's shadow
 // memory model (currently 64)
```
carolynzech, line 6: Is this comment still accurate?

library/kani_core/src/mem.rs
```diff
@@ -287,14 +287,9 @@ macro_rules! kani_mem {
 
         /// Get the object offset of the given pointer.
         #[doc(hidden)]
-        #[crate::kani::unstable_feature(
-            feature = "ghost-state",
-            issue = 3184,
-            reason = "experimental ghost state/shadow memory API"
-        )]
```
carolynzech, line 294: So we're claiming this is stable now?

tests/expected/shadow/unsupported_object_size/test.rs
```diff
@@ -5,12 +5,17 @@
...
+
 static mut SM: kani::shadow::ShadowMem<bool> = kani::shadow::ShadowMem::new(false);
 
 fn check_max_objects<const N: usize>() {
     let arr: [u8; N] = [0; N];
     let last = &arr[N - 1];
-    assert_eq!(kani::mem::pointer_offset(last as *const u8), N - 1);
+    unsafe {
+        assert_eq!(ptr_offset_from_unsigned(last as *const u8, &arr[0] as *const u8), N - 1);
+    }
```
carolynzech, line 18: Confused why we need to change the test?

tests/expected/shadow/unsupported_object_size/test.rs
tautschnig: I think so - the test still appears to confirm this. (It has a succeeding and an expected-to-fail case.)

tests/expected/shadow/unsupported_object_size/test.rs
tautschnig: Because `kani::mem::pointer_offset` is now `pub(crate)` instead of being conditionally-public (under -Zghost-state).

library/kani_core/src/mem.rs
tautschnig: No, it's actually no longer externally available, consistent with `pointer_object`. `pointer_offset` is entirely unrelated to ghost state or shadow memory, and hence should never have been exposed that way.

carolynzech (approved):

PR #4181, opened 2025-06-26
Cache dependencies for CI jobs

### LINKED ISSUE #4153: Cache Dependencies for CI
As far as I can tell we don't take advantage of GitHub's caching feature for CI. I think even just caching the cargo registry & kani dependencies could help with our build times and, since caches can be shared between workflows, this could save on redundant work happening between them.

Although builds are only a small portion of the time it takes longer running workflows to run (e.g. building is just ~3 min of 32 total [in this run of the regression tests](https://github.com/model-checking/kani/actions/runs/15617713096/job/43994612939)), this could make a significant difference on shorter workflows like clippy, the benchcomp tests, and docs building.

AlexanderPortland:
Adds a simple cache of the `~/.cargo/registry` directory and Kani's `target` directory to reduce compilation times of our dependencies in CI. This also caches the `~/.rustup` directory to avoid having to re-download toolchains and re-compile the standard library every CI run. Caches are shared between workflows that compile Kani in the same way (e.g. between the `benchcomp-tests` & `perf` jobs since they both compile in release mode).

Based on testing on my local fork, this change seems to make a minor, but not insignificant impact on runtimes. Although lost in the noise on longer jobs, `llbc-regression` & `benchcomp-tests` got noticeably faster, as did the compiler benchmarking workflows which have to compile Kani for each commit they are comparing.

**Test runs:**
General Kani CI workflow: [before](https://github.com/AlexanderPortland/kani/actions/runs/15886762745) | [after](https://github.com/AlexanderPortland/kani/actions/runs/15887740017)
Compiler bench workflow: [before](https://github.com/AlexanderPortland/kani/actions/runs/15888031249) | [after](https://github.com/AlexanderPortland/kani/actions/runs/15889545021)

Resolves #4153 

### DIFF
--- a/.github/workflows/bench-compiler.yml
+++ b/.github/workflows/bench-compiler.yml
@@ -37,6 +37,18 @@ jobs:
           ref: ${{ env.NEW_REF }}
           fetch-depth: 1
 
+      - name: Cache Kani build artifacts
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
+          shared-key: "kani-rust-cache-bench"
+          workspaces: |
+            new
+            new/tools/compile-timer
+            old
+            old/tools/compile-timer
+          cache-directories: "~/.rustup"
+
       - name: Set up Kani Dependencies (old variant)
         uses: ./old/.github/actions/setup
         with:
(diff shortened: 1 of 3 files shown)


tautschnig (approved):
I think that's a good start, although we might want to figure out what else we could save time on (by caching).

PR #4183, opened 2025-06-26
Include wget in dependencies

zhassan-aws:
Install `wget` as part of the dependencies as it's needed for downloading kissat (`scripts/setup/install_kissat.sh`).


tautschnig:
I'd suggest that we get rid of `wget` and instead use `curl -L --remote-name` instead. We already require `curl` in `src/setup.rs`.

tautschnig:
I'm sorry to be a pain, but can we please use `curl` everywhere? We have 4 more uses of `wget` per `git grep wget`.

zhassan-aws:
Done. Let me know if I missed any.

tautschnig:
I believe https://github.com/model-checking/kani/blob/79fe8af2e694c5dfc801dd15ba3dc0d1ea582146/scripts/setup/ubuntu/install_cbmc.sh#L25 is still there.

zhassan-aws:
This was removed in 60657dd47639d8b671808f4102de5377907d97ba.

tautschnig (approved):

PR #4184, opened 2025-06-26
Upgrade stable branch to Rust 1.88.0

zhassan-aws:
Advance the stable branch to 0b68df0b8ed585b4bae9d6c07db64875137f1ae7 and use the 1.88.0 channel.

This PR was created through:
1. Running: `git merge 0b68df0b8ed585b4bae9d6c07db64875137f1ae7 -X theirs`
2. Modifying the channel in `rust-toolchain.toml` to `1.88.0`

The rationale for picking 0b68df0b8ed585b4bae9d6c07db64875137f1ae7 is as follows:
- The 1.88.0 release was branched from main on 2025-05-09 (see https://releases.rs/docs/1.88.0/)
- The latest commit on main for which compilation and regressions pass with the 1.88.0 channel is 0b68df0b8ed585b4bae9d6c07db64875137f1ae7 (which uses the 2025-05-20 toolchain) 


.github/workflows/extra_jobs.yml
```diff
@@ -28,6 +28,7 @@ jobs:
     name: Auto Label
     permissions:
       contents: read
+      issues: write
```
tautschnig: That unfortunately is not the right fix: the problem is caused by performance benchmarking labels having been changed (in #4154). As (temporary) workaround we could put back the Z-BenchCI label, even though it might no longer have the desired effect.

.github/labeler.yml
```diff
@@ -8,6 +8,6 @@
 Z-BenchCI:
 - any:
   - changed-files:
-    - any-glob-to-any-file: ['kani-compiler/**', 'kani-driver/src/call-*', 'cprover_bindings/**', 'library/**']
+    - any-glob-to-any-file: ['kani-compiler/**', 'kani-driver/src/call_*', 'cprover_bindings/**', 'library/**']
```
tautschnig, line 11: Perhaps just cherry-pick all of 9d87281d4736288d420d7b7f40cf47a2e2125158 into this branch?

.github/workflows/extra_jobs.yml
tautschnig: Or maybe just cherry-pick all of https://github.com/model-checking/kani/commit/9d87281d4736288d420d7b7f40cf47a2e2125158, as suggested in my other comment.

.github/labeler.yml
zhassan-aws: I did, but it's still failing :(

tautschnig:
I'm actually quite confused by each of the CI failures seen here. The "Auto Label" problem we can easily work around by no longer marking it "Required."

carolynzech:
I think what's happening is that it's trying to run the workflow from the branch we're merging _into_, so the stable branch, where the changes from https://github.com/model-checking/kani/commit/9d87281d4736288d420d7b7f40cf47a2e2125158 don't exist. So I think you'd need to cherry pick that commit into `stable` instead of this PR branch.

See https://github.com/orgs/community/discussions/25746.

carolynzech:
I think taking a step back: I may just recommend that you push directly to stable in this instance, instead of having to cherry pick over there just to get this PR merged; seems messy. Seems ok to do in this one-off instance, since we can see from the CI run here that regressions pass.

zhassan-aws:
That makes sense. Thanks for looking into this, @carolynzech. If I can get an approval, I'll merge the PR, and we can follow-up on any failures afterwards.

zhassan-aws:
Or maybe, I misunderstood what you meant by push directly to stable. Can you clarify?

PR #4197, opened 2025-07-03
Update to Rust edition 2024

tautschnig:
Upstream change https://github.com/rust-lang/rust/pull/143214 (Remove let_chains unstable feature) requires us to either remove the use of `let` in several places or wholesale update to Rust edition 2024. Doing this edition update seems worthwhile to keep accessing more recent features, but required code changes in several places.

This is in preparation of the toolchain upgrade to 2025-07-03, but does not actually do this upgrade for it will require other, unrelated changes.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
@@ -1754,7 +1754,7 @@ impl<'tcx> GotocCtx<'tcx> {
         &self,
         instance: InstanceStable,
         fn_abi: &'a FnAbi,
-    ) -> impl Iterator<Item = (usize, &'a ArgAbi)> {
+    ) -> impl Iterator<Item = (usize, &'a ArgAbi)> + use<'a> {
         let requires_caller_location = self.requires_caller_location(instance);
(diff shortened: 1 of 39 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
```diff
@@ -1754,7 +1754,7 @@ impl<'tcx> GotocCtx<'tcx> {
         &self,
         instance: InstanceStable,
         fn_abi: &'a FnAbi,
-    ) -> impl Iterator<Item = (usize, &'a ArgAbi)> {
+    ) -> impl Iterator<Item = (usize, &'a ArgAbi)> + use<'a> {
```
carolynzech, line 1757: what does this mean?

carolynzech (approved):
Thanks!

kani-compiler/src/codegen_cprover_gotoc/codegen/typ.rs
carolynzech: Never mind, answered my own question: https://doc.rust-lang.org/nightly/reference/types/impl-trait.html#precise-capturing

PR #4201, opened 2025-07-04
Implement support for Cargo.toml's default-members

tautschnig:
Cargo has resolved https://github.com/rust-lang/cargo/issues/8033. Invoking `cargo kani` now correctly mimics cargo's behavior when there is a `default-members` entry in `Cargo.toml`.

### DIFF
--- a/kani-driver/src/call_cargo.rs
+++ b/kani-driver/src/call_cargo.rs
@@ -408,14 +408,14 @@ crate-type = ["lib"]
 
     /// Extract the packages that should be verified.
     ///
-    /// The result is built following these rules:
+    /// The result is built following these rules (mimicking cargo, see
+    /// https://github.com/rust-lang/cargo/blob/master/src/cargo/core/workspace.rs):
     /// - If `--package <pkg>` is given, return the list of packages selected.
     /// - If `--exclude <pkg>` is given, return the list of packages not excluded.
     /// - If `--workspace` is given, return the list of workspace members.
-    /// - If no argument provided, return the root package if there's one or all members.
-    ///   - I.e.: Do whatever cargo does when there's no `default_members`.
-    ///   - This is because `default_members` is not available in cargo metadata.
-    ///     See <https://github.com/rust-lang/cargo/issues/8033>.
+    /// - Else obtain the set of packages from cargo's default_workspace_members (i.e., if
+    ///   `default-members` is specified in Cargo.toml, use that list; else if a root package is
+    ///   specified use that; else use all members).
     ///
     /// In addition, if either `--package <pkg>` or `--exclude <pkg>` is given,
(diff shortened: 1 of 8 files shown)


kani-driver/src/call_cargo.rs
```diff
@@ -455,11 +455,10 @@ crate-type = ["lib"]
...
                 .filter(|pkg| !pkg_ids.contains_key(&pkg.id))
                 .collect()
+        } else if args.cargo.workspace {
+            metadata.workspace_packages()
         } else {
-            match (args.cargo.workspace, metadata.root_package()) {
-                (true, _) | (_, None) => metadata.workspace_packages(),
-                (_, Some(root_pkg)) => vec![root_pkg],
-            }
+            metadata.workspace_default_packages()
```
carolynzech, line 461: Don't we still need to fall back to the root package if this is unavailable?

tautschnig: No need to: `workspace_default_packages` automatically does this, i.e., evaluates to the root package when `default-members` is not set.

carolynzech: Where are you seeing that? My reading of this part of [the documentation](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-default-members-field):

> When unspecified, the [root package](https://doc.rust-lang.org/cargo/reference/workspaces.html#root-package) will be used.

was that Cargo will operate on the root package when `default-members` is unspecified, but that doesn't necessarily mean it will set default-members to the root package to do that.

carolynzech: The passive voice in the docs makes it tricky; I can't tell whether the "will be used" means that it'll be used as the value of `default-members` (your reading) or it means it'll be used as the thing we operate on (my reading, from this sentence):

> The default-members field specifies paths of [members](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields) to operate on...

i.e., I think what we're debating is whether Cargo does this (psuedocode):

```rust
if workspace.default_members.is_none() {
  workspace.default_members = workspace.root_package;
}
operate(workspace.default_members);
```
or this:
```rust
if let Some(members) = workspace.default_members {
  operate(members);
} else {
  operate(workspace.root_package); // workspace.default_members is still unset
}
```

Reading it again, I think you're right. I'd like to feel a bit more certain, though; I'll look at this more closely tomorrow and see if I can get some confirmation in the source code or ask someone.

tautschnig: I ran `cargo metadata | jq ".workspace_default_members"` while playing with variations of the `Cargo.toml` file to see what was going on (I also found documentation of this, including the one in the source code, very confusing).

PR #4204, opened 2025-07-04
Use CBMC's shuffle_vector expression

tautschnig:
Ever since CBMC 5.51.0 it is no longer necessary to lower this expression in Kani.


carolynzech (approved):
Thanks! Can you perhaps link to https://doc.rust-lang.org/std/intrinsics/simd/fn.simd_shuffle.html (or some other similar documentation) somewhere in the code, perhaps in expr.rs?

PR #4210, opened 2025-07-07
[DRAFT] Use arena allocator when exporting goto binaries

AlexanderPortland:
### Motivation
A large portion (~7%) of our codegen time is spent simply deallocating the large `SymbolTable` struct that we generate when building each goto binary for CBMC. This ends up taking a while because the struct contains `Irep` structs which can each recursively contain multiple Vecs of even more `Irep`s, requiring a lot of small allocations to be individually handled over and over again.

### Approach
This PR switches all data structures within Kani's binary generation to use a shared bump allocator (using the Bumpalo crate) that ensures all allocations we use while serializing a given binary remain in a single large region of memory. Once a given binary has been serialized, we can then easily deallocate all the memory it used at once.

As part of this change, we also replace the `LinearMap` used to represent [named subs within an Irep](url) with a proper `hashbrown::HashMap`.

### Issues
Based on manual testing there is currently a small memory leak (I think because the underlying vectors for the `BigInt` & `BigUInt` types we use [here](https://github.com/AlexanderPortland/kani/blob/3d34482d21d292c4b1260bb9616346f2bc3cb94e/cprover_bindings/src/irep/irep_id.rs#L19-L21)) still use the global allocator and thus will not be deallocated with the memory arena. I'm looking into fixing this and improving code comments for the more involved changes, but wanted to make a draft PR to get preliminary performance results first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,9 +207,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.18.1"
+version = "3.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "793db76d6187cd04dff33004d8e6c9cc4e05cd330500379d2394209271b4aeee"
+checksum = "46c5e41b57b8bba42a04676d81cb89e9ee8e859a1a66f80a5a72e1cb76b34d43"
 
 [[package]]
 name = "byteorder"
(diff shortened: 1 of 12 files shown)


AlexanderPortland:
Although very interesting, I'm closing this idea as:
* The current implementation has memory leaks since the external `BigUInt` & `BigInt` types (used in our `IrepId`s) represent their data with a heap-allocated `Vec<u64>` that won't use the new bump allocator. We would have to fork the crate that provides them to parametrize those underlying vectors by an allocator.
* The changes described in #4237 would entirely skip all the allocations involved in creating a new `SymbolTable` struct, rather than just making the deallocation of that struct much faster.

PR #4211, opened 2025-07-08
Split compiler flags to avoid dependency recompilation

### LINKED ISSUE #2230: Split "codegen" flags to reduce recompilation

AlexanderPortland:
We currently pass all Kani options to `rustc` for use by the `kani-compiler`. This means that the Cargo build cache will detect a [compilation fingerprint](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/compiler/fingerprint/index.html) change whenever these options change and have to fully recompile the project (including all dependencies). However, for the majority of options, this is wholly unnecessary, as they only affect our codegen of the target crate once it’s already been lowered to MIR.

This PR now passes most of `kani-compiler`’s options only to the target crate so that they don’t change how dependencies are built and we can avoid their recompilation.

There are two exceptions:
- `--ignore-global-assembly`: this is needed even in dependencies because [the check we do](https://github.com/model-checking/kani/blob/2bdc9ba359ce26f80e6eb9b023b39266f726de8f/kani-compiler/src/kani_middle/mod.rs#L47-L60) to detect global assembly happens even if we are skipping codegen.
- `--check-version`: this ensures we recompile everything when the user upgrades Kani versions. While not strictly necessary, existing comments understandably seem to want this behavior, and upgrades are rare enough I don’t see it being a real issue.

Resolves #2230


kani-driver/src/call_single_file.rs
```diff
@@ -80,6 +80,7 @@ impl KaniSession {
         // rustc ones.
         let mut cmd = Command::new(&self.kani_compiler);
         let kani_compiler_args = to_rustc_arg(kani_args);
+        // println!("compiling w/ args {:?}", kani_compiler_args);
```
tautschnig: Let's make this either a debug output or remove the commented-out line entirely.

tests/script-based-pre/build-cache-dirty/rebuild.sh
```diff
@@ -9,6 +9,8 @@
 # - Add a dependency
 # Note: This should run in the folder where the script is.
 
+set -x
+
```
tautschnig: I suppose this is a debug left-over?

tests/script-based-pre/build-cache-dirty/rebuild.sh
```diff
@@ -47,11 +49,19 @@ mkdir -p ${OUT_DIR}
 cp -r target_lib ${OUT_DIR}
 
 echo "Initial compilation"
-check_kani --no-assertion-reach-checks initial.log
+check_kani "--no-assertion-reach-checks" initial.log
```
tautschnig: Why is this change required?

tautschnig (approved):

kani-driver/src/call_single_file.rs
AlexanderPortland: Ahh shoot, sorry I didn't catch these two debugging leftovers before making the PR. Will be fixed!

tests/script-based-pre/build-cache-dirty/rebuild.sh
AlexanderPortland: Good catch, I don't think it is! I'm not very experienced w/ bash, so I must've added it when adding quotes around the args for lines 59 & 63 (which need them bc they contain spaces) without fully realizing the reasoning didn't apply there.

Issue #4212, opened 2025-07-08
Avoid `rustc` re-lowering when Kani options change?


AlexanderPortland:
This may be a dumb idea as I don't know much about Cargo's build system and incremental compilation strategy, but would it be possible to reuse existing compiler-generated MIR for the target crate even if Kani options change? So long as such options would only affect our codegen from MIR, theoretically it doesn't feel like there's a reason the compiler would have to re-lower everything to MIR just to get the same result anyway. 

Maybe the answer is no because we'd have to pass those options to the `kani-compiler`, changing the way `rustc` is called and invalidating the build cache. However, even if that's an unfixable limitation of Cargo, it feels like there could still be a way to bypass that by writing options to env vars or a file or something.

carolynzech:
> Maybe the answer is no because we'd have to pass those options to the kani-compiler, changing the way rustc is called and invalidating the build cache. 

Yes, exactly. We tell Cargo that rustc is kani-compiler here: https://github.com/model-checking/kani/blob/192911d28486f26a36cee6dc0ce817bccd5d9952/kani-driver/src/call_cargo.rs#L219

Cargo uses fingerprints to tell whether to recompile something. [This table](https://doc.rust-lang.org/nightly/nightly-rustc/cargo/core/compiler/fingerprint/index.html#fingerprints-and-unithashs) says that "cargo rustc extra args" are included in the fingerprint, which means that any changes to the `llvm-args` we pass to Cargo (i.e., kani-compiler's arguments) invalidate the cache. See also our documentation about this: https://github.com/model-checking/kani/blob/192911d28486f26a36cee6dc0ce817bccd5d9952/kani-driver/src/call_single_file.rs#L259-L268

carolynzech:
> However, even if that's an unfixable limitation of Cargo, it feels like there could still be a way to bypass that by writing options to env vars or a file or something.

I think as long as we 1) pass kani-compiler arguments with llvm-args and 2) rely on Cargo's build cache, this is an unfixable limitation. We could always 1) try passing the arguments another way, although I don't think that's really possible unfortunately or 2) implement our own build cache, which is probably not worth the effort

carolynzech:
So it's definitely not a dumb idea--it would be better if we invalidated the cache less often. I think the question is: can we figure out a way to do it that isn't 1) a ton of work and 2) risks getting it wrong?
[...]

PR #4213, opened 2025-07-10
Skip codegen for unneeded harnesses

AlexanderPortland:
This PR moves the logic for filtering out specific harnesses (as specified by the `--harness` & `--exact` Kani arguments) to the compiler, allowing us to skip codegen for harnesses that will end up being blocked by the filter. I’ve also added [a dev-build-only assertion](https://github.com/AlexanderPortland/kani/blob/c6283ac0ec53f70bf6ab78f367b6443c25b5290c/kani-driver/src/metadata.rs#L103-L110) to double check that we’re not doing codegen for any unused harnesses.

Based on testing, something like running `cargo kani --harness estimator` on `s2n-quic-codec` used to take ~11s locally to compile 66 harnesses, but only 5 would end up being analyzed. Now, the same command compiles only the 5 needed harnesses, taking just ~5s.

The downside of this is that the target crate now has to be recompiled if the user runs Kani with differing harness filters but, if the idea in #4212 is possible, that too could be avoided.

### DIFF
--- a/kani-compiler/src/kani_middle/codegen_units.rs
+++ b/kani-compiler/src/kani_middle/codegen_units.rs
@@ -21,7 +21,7 @@ use crate::kani_queries::QueryDb;
 use fxhash::{FxHashMap, FxHashSet};
 use kani_metadata::{
     ArtifactType, AssignsContract, AutoHarnessMetadata, AutoHarnessSkipReason, HarnessKind,
-    HarnessMetadata, KaniMetadata,
+    HarnessMetadata, KaniMetadata, find_proof_harnesses,
 };
 use regex::RegexSet;
 use rustc_hir::def_id::DefId;
@@ -74,7 +74,11 @@ impl CodegenUnits {
         let args = queries.args();
         match args.reachability_analysis {
             ReachabilityType::Harnesses => {
-                let all_harnesses = get_all_manual_harnesses(tcx, base_filename);
+                let all_harnesses = determine_targets(
+                    get_all_manual_harnesses(tcx, base_filename),
+                    &args.harnesses,
+                    args.exact,
+                );
                 // Even if no_stubs is empty we still need to store rustc metadata.
                 let units = group_by_stubs(tcx, &all_harnesses);
                 validate_units(tcx, &units);
@@ -82,7 +86,11 @@ impl CodegenUnits {
                 CodegenUnits { units, harness_info: all_harnesses, crate_info }
             }
             ReachabilityType::AllFns => {
-                let mut all_harnesses = get_all_manual_harnesses(tcx, base_filename);
+                let mut all_harnesses = determine_targets(
+                    get_all_manual_harnesses(tcx, base_filename),
+                    &args.harnesses,
+                    args.exact,
+                );
                 let mut units = group_by_stubs(tcx, &all_harnesses);
                 validate_units(tcx, &units);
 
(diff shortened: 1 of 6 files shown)


kani_metadata/src/harness.rs
```diff
@@ -112,3 +112,47 @@ impl HarnessMetadata {
...
+pub fn find_proof_harnesses<'a, I>(
+    targets: &BTreeSet<&String>,
+    all_harnesses: I,
+    exact_filter: bool,
+) -> Vec<&'a HarnessMetadata>
+where
+    I: IntoIterator,
+    I::Item: Borrow<&'a HarnessMetadata>,
+{
+    // debug!(?targets, "find_proof_harness");
```
AlexanderPortland: I wanted some advice here: since I moved this function to the common `kani_metadata` crate for use in both the compiler and driver, it no longer has access to the tracing dependencies. Would it be better to add that dependency here, or just remove the debug statements?

kani-compiler/src/kani_middle/codegen_units.rs
```diff
@@ -327,6 +335,31 @@ fn get_all_manual_harnesses(
...
 
+/// Filter which harnesses to codegen based on user filters. Shares use of `find_proof_harnesses` with the `determine_targets` function
+/// in `kani-driver/src/metadata.rs` to ensure the filter is consistent and thus codegen is always done for the subset of harnesses we want
+/// to analyze.
+fn determine_targets(
+    harnesses: HashMap<Harness, HarnessMetadata>,
+    harness_filters: &[String],
+    exact_filter: bool,
+) -> HashMap<Harness, HarnessMetadata> {
+    // If there are no filters, nothing to do!
```
carolynzech:
```suggestion
```
think this is self-explanatory :)

kani_metadata/src/harness.rs
carolynzech: I'd add the dependency!

kani-compiler/src/kani_middle/codegen_units.rs
```diff
@@ -327,6 +335,31 @@ fn get_all_manual_harnesses(
         .collect::<HashMap<_, _>>()
 }
 
+/// Filter which harnesses to codegen based on user filters. Shares use of `find_proof_harnesses` with the `determine_targets` function
+/// in `kani-driver/src/metadata.rs` to ensure the filter is consistent and thus codegen is always done for the subset of harnesses we want
+/// to analyze.
+fn determine_targets(
+    harnesses: HashMap<Harness, HarnessMetadata>,
```
carolynzech:
```suggestion
    all_harnesses: HashMap<Harness, HarnessMetadata>,
```

carolynzech: nit b/c the compiler arg for `harness_filters` is `harnesses` but then `harnesses` here refers to something else

AlexanderPortland:
> Thanks! Can you check if we have tests for these flags and add some if not? My goal is to make sure that this doesn't break the existing `--harness` behavior, so I'd be fine with some expected tests in the `ui` folder that demonstrate that we only run proofs for the filtered functions. If you're feeling fancy, you could try to write a script that lists the goto files being generated to ensure that we're only doing codegen for the ones we filter on too, but not necessary.

I think we might already be covered!

For doing unneeded codegen: I'd added a general debug assertion ([here](https://github.com/AlexanderPortland/kani/blob/698d196f5ac5d113ba2a69ea37e50dccaae55850/kani-driver/src/metadata.rs#L99-L109)) that checks to ensure we didn't codegen any harnesses that we won't use for verification. Since our tests are run in debug mode, these should always run and double check that we're doing no extra harness codegen.

For doing too little codegen: I think this would be caught by our existing tests failing (specifically [`exact-harness`](https://github.com/model-checking/kani/tree/main/tests/ui/exact-harness) or [`multiple-harnesses`](https://github.com/model-checking/kani/tree/main/tests/ui/multiple-harnesses)). I totally can add more if that seems helpful, but the existing tests are pretty good.

carolynzech (approved):

Issue #3277, opened 2024-06-19
Add support to Z3


celinval:
Requested feature: Add Z3 as a Kani solver option
Use case: As we investigate adding support to quantifiers, using SMT solvers will increase the quantifier expressiveness.
Link to relevant documentation (Rust reference, Nomicon, RFC):

Note: Z3 should be enable in CBMC with: `--incremental-smt2-solver "z3 --smt2 -in"`

celinval:
I created this commit to enable it (https://github.com/model-checking/kani/compare/main...celinval:kani-dev:issue-3277-z3), but we still need to assess how much support these two solvers have once we upgrade to CBMC 6.0. I tried it with the 5.9 version, but it is pretty broken due to unsupported goto constructs.

PR #4218, opened 2025-07-11
Support bitwuzla, cvc5, z3 as solver attribute values

tautschnig:
Adds support for solver attribute values `bitwuzla`, `cvc5`, and `z3` to make CBMC use the corresponding SMT solver. 

Use Z3 for selected tests when running regression tests with the following rationale: The tests loop_assigns_for_vec.rs and assert-postconditions.rs previously required more than 8 GB of memory on Ubuntu/x86 and solved in 1:15 minutes and 25 seconds, respectively. When using Z3 as back-end solver, they complete in 35 and 2 seconds (2.5 GB and 150 MB), respectively.

Use CVC5 for the test from issue #4226.

We are not testing Bitwuzla support in CI at this point (though it has been confirmed to work locally), because Bitwuzla only releases binaries for some of our CI platforms, and we cannot even build Bitwuzla from source on Ubuntu 22.04 (meson is too old).

Resolves: #3277
Resolves: #4226


tautschnig:
~~I'm inclined to mark this ready-for-review given the recurring problem of (spurious) CI failures as seen in, e.g, #4202. What I'd really like to do, however, is making Z3, CVC5, Bitwuzla first-class solvers in Kani just like, e.g., Kissat is. Unless someone beats me to it I'll try to take care of this tomorrow.~~ Edit: done.

carolynzech (approved):
Thanks! Can you add a Bitwuzla test as a sanity check? It doesn't need to be an improvement over the default in any way, just something to demonstrate that someone could actually specify this attribute and it would work.

tautschnig:
> Thanks! Can you add a Bitwuzla test as a sanity check? It doesn't need to be an improvement over the default in any way, just something to demonstrate that someone could actually specify this attribute and it would work.

Will do, but it will require us building from source on macos-13 runners as https://github.com/bitwuzla/bitwuzla/releases does not include binaries for macOS-x86_64.

tautschnig:
> > Thanks! Can you add a Bitwuzla test as a sanity check? It doesn't need to be an improvement over the default in any way, just something to demonstrate that someone could actually specify this attribute and it would work.
> 
> Will do, but it will require us building from source on macos-13 runners as https://github.com/bitwuzla/bitwuzla/releases does not include binaries for macOS-x86_64.

So I tried, but I am now backtracking on this, see also the (updated) note in the PR description: we cannot build Bitwuzla from source on Ubuntu 22.04, and I don't think this is worth any extra gymnastics.

zhassan-aws:
Can the new attribute values be used in Kani releases?

tautschnig:
> Can the new attribute values be used in Kani releases?

Unlike Kissat, users will still need to install those solvers themselves. I pondered including the solvers in the release bundle, but then decided to defer questions about licensing and the size of the release bundle. So we could certainly change that, but I think some additional care is required.

zhassan-aws:
I see. In this case, I would prefer if we invoke those solvers using `--cbmc-args --z3` in the added regression tests instead of via the solver attribute. IMO, it would be undesirable to provide options that don't work out of the box (not a good UX).

PR #4220, opened 2025-07-11
Strongly type differing compiler args for clarity

AlexanderPortland:
### Context
When working on #4211 & #4213, I was initially very confused by all the machinery in [`call_cargo.rs`](https://github.com/model-checking/kani/blob/main/kani-driver/src/call_cargo.rs) and [`call_single_file.rs`](https://github.com/model-checking/kani/blob/main/kani-driver/src/call_single_file.rs) that we used to pass arguments to the different layers of the compiler. The main source of my confusion was that there were many arguments all destined for different places (e.g. cargo, rustc, or kani-compiler) that all had to be encoded in differing ways (e.g. through the `--llvm-args` flag or the `CARGO_ENCODED_RUSTFLAGS` environment variable) but were mixed together sharing the same `String` type and similar-sounding variable names.

Although misconfiguration would often be caught at runtime with a section of the compiler complaining about unexpected arguments, this confusion silently bit me a few times where argument names were shared.

### Fix
This PR represents an attempt to encode the way arguments target different portions of the compiler into the type system using specific wrapper types for each kind of argument we use. You can convert transparently into these wrapper types, but then can only convert between them or pass them to a cargo command using the proper encoding methods.

Although not perfect (you could still convert into the wrong kind of argument wrapper initially), this does seem to add some slight guardrails on the way we call cargo, especially against smaller changes to the existing patterns like the ones I was making in my PRs. It also may make the code easier to parse.


kani-driver/src/call_cargo.rs
```diff
@@ -37,6 +37,114 @@ pub struct CargoOutputs {
     pub cargo_metadata: Metadata,
 }
 
+pub(crate) mod args {
```
carolynzech: Let's move this to `util.rs`. If multiple files are going to use this logic, better to have it in a common utility file than pick one of the files for it to arbitrarily live in.

kani-driver/src/call_cargo.rs
```diff
@@ -37,6 +37,114 @@ pub struct CargoOutputs {
...
+    #[derive(Clone, PartialEq)]
+    /// Kani-specific arguments passed to rustc and then used by `kani-compiler`.
+    /// This includes everything from the `Arguments` struct in `kani-compiler/src/args.rs`.
+    pub struct KaniArg(String);
+
+    #[derive(Clone, PartialEq)]
+    /// Arguments passed to rustc.
+    /// Basically anything that gets listed when running `rustc --help`.
+    /// This includes options like `--extern` for specify extern crates, `-L` for adding to the lib
+    /// search path or `--emit mir` for emitting MIR.
```
carolynzech:
```suggestion
```

kani-driver/src/call_cargo.rs
```diff
@@ -37,6 +37,114 @@ pub struct CargoOutputs {
...
+    pub struct RustcArg(OsString);
+
+    #[derive(Clone, PartialEq)]
+    /// Arguments passed to Cargo.
+    /// Basically anything that gets listed when running `cargo --help`. This includes options like
+    /// all cargo subcommands, `--config`, `--target`, or any `-Z` unstable Cargo options.
```
carolynzech:
```suggestion
```

kani-driver/src/call_cargo.rs
carolynzech: Also, I suggested deleting the longer explanations of `RustcArgs` and `CargoArgs` above their definitions. I think it'd be more effective to instead have module-level documentation of how these three structs fit together, i.e., a mini architecture explanation. Something like:

> We have kani-specific arguments that the kani compiler uses to direct codegen (the Arguments struct). Before codegen, the Kani compiler invokes rustc to compile src to MIR, so we pass rustc arguments to influence that compilation (c.f. LibConfig). If we're running `cargo kani`, we also want to pass arguments to Cargo, e.g., the location of the manifest path.

(But more formal than that.) 

I find generally documentation is more useful when it focuses on *why* the thing exists instead of providing in-depth examples of usage, which is 1) brittle to source code changes and 2) easy enough for people to search for in the code themselves.

kani-driver/src/call_cargo.rs
```diff
@@ -37,6 +37,114 @@ pub struct CargoOutputs {
...
+    pub trait CommandWrapper {
+        fn pass_cargo_args(&mut self, args: &[CargoArg]) -> &mut Self;
+        fn pass_rustc_args(&mut self, args: &[RustcArg], to: PassTo) -> &mut Self;
+        fn pass_kani_args(&mut self, args: &[KaniArg], to: PassTo) -> &mut Self;
+    }
+
+    impl CommandWrapper for Command {
+        /// Pass general arguments to cargo.
+        fn pass_cargo_args(&mut self, args: &[CargoArg]) -> &mut Self {
+            self.args(args.iter().map(|d| &d.0))
```
carolynzech:
```suggestion
            self.args(args.iter().map(|arg| &arg.0))
```

kani-driver/src/call_cargo.rs
```diff
@@ -37,6 +37,114 @@ pub struct CargoOutputs {
...
+        fn pass_rustc_args(&mut self, args: &[RustcArg], to: PassTo) -> &mut Self {
+            match to {
+                // Since we also want to recursively pass these args to all dependencies,
+                // use an environment variable that gets checked for each dependency.
+                // Use of CARGO_ENCODED_RUSTFLAGS instead of RUSTFLAGS is preferred. See
+                // https://doc.rust-lang.org/cargo/reference/environment-variables.html
+                PassTo::AllDependencies => self.env(
+                    "CARGO_ENCODED_RUSTFLAGS",
+                    args.iter()
+                        .map(|d| d.0.clone())
```
carolynzech:
```suggestion
                        .map(|arg| arg.0.clone())
```

kani-driver/src/call_cargo.rs
```diff
@@ -37,6 +37,114 @@ pub struct CargoOutputs {
...
+                        .collect::<Vec<OsString>>()
+                        .join(OsStr::new("\x1f")),
+                ),
+                // Since we just want to pass to the local crate, just add them as arguments to the command.
+                PassTo::OnlyLocalCrate => self.args(args.iter().map(|d| &d.0)),
```
carolynzech:
```suggestion
                PassTo::OnlyLocalCrate => self.args(args.iter().map(|arg| &arg.0)),
```

kani-driver/src/call_cargo.rs
carolynzech: Maybe I'm missing what `d` means? But otherwise better to pick a meaningful name.

kani-driver/src/call_cargo.rs
```diff
@@ -37,6 +37,114 @@ pub struct CargoOutputs {
...
+    /// We currently pass Kani specific arguments using the `--llvm-args` structure which is the
+    /// hacky mechanism used by other rustc backend to receive arguments unknown to rustc.
+    ///
+    /// Note that Cargo caching mechanism takes the building context into consideration, which
+    /// includes the value of the rust flags. By using `--llvm-args`, we ensure that Cargo takes into
+    /// consideration all arguments that are used to configure Kani compiler. For example, enabling the
+    /// reachability checks will force recompilation if they were disabled in previous build.
+    /// For more details on this caching mechanism, see the
+    /// [fingerprint documentation](https://github.com/rust-lang/cargo/blob/82c3bb79e3a19a5164e33819ef81bfc2c984bc56/src/cargo/core/compiler/fingerprint/mod.rs)
+    pub fn to_rustc_arg(kani_args: &[KaniArg]) -> RustcArg {
```
carolynzech: Can we get rid of this function and change it to be a custom `impl From<&[KaniArg]> for RustcArg`?

carolynzech: or alternatively make this a private function to the module, see my later comments about exposing this

kani-driver/src/call_cargo.rs
```diff
@@ -77,13 +185,12 @@ crate-type = ["lib"]
...
         let lib_path = lib_no_core_folder().unwrap();
         let mut rustc_args = self.kani_rustc_flags(LibConfig::new_no_core(lib_path));
-        rustc_args.push(to_rustc_arg(self.kani_compiler_local_flags()).into());
+        rustc_args.push(to_rustc_arg(&self.kani_compiler_local_flags()));
         // Ignore global assembly, since `compiler_builtins` has some.
-        rustc_args.push(
-            to_rustc_arg(vec!["--ignore-global-asm".to_string(), self.reachability_arg()]).into(),
-        );
+        rustc_args
+            .push(to_rustc_arg(&[KaniArg::from("--ignore-global-asm"), self.reachability_arg()]));
```
carolynzech: I find calling `to_rustc_arg` here odd; shouldn't this be hidden behind the `pass_kani_args` abstraction?
This is another place where module-level documentation would be good; we can make clearer to devs how they're expected to use this API. Rn I think they almost have too much choice--unless I'm missing something, I don't see why they should have both `to_rustc_arg(&[KaniArgs])` and `pass_kani_arg` as options.

kani-driver/src/call_cargo.rs
```diff
@@ -37,6 +37,114 @@ pub struct CargoOutputs {
...
+        .into()
+    }
+
+    pub enum PassTo {
+        /// Only pass arguments for use in the local crate.
+        /// This will just pass them directly as arguments to the command.
+        OnlyLocalCrate,
+        /// Pass arguments for use when compiling all dependencies using the
+        /// `CARGO_ENCODED_RUSTFLAGS` environment variable.
+        AllDependencies,
```
carolynzech:
```suggestion
        AllCrates,
```
Since technically the local crate is not a dependency, so `AllDependencies` suggests it's everything *but* the local crate when really you mean it's passed to every rustc invocation that cargo makes, which includes the local crate

kani-driver/src/call_cargo.rs
```diff
@@ -212,14 +314,13 @@ crate-type = ["lib"]
...
                     .args(vec!["-p", &package.id.to_string()])
                     .args(verification_target.to_args())
-                    .args(&pkg_args)
+                    .arg("--") // Add this delimiter so we start passing args to rustc and not Cargo
+                    .pass_kani_args(&kani_pkg_args, args::PassTo::OnlyLocalCrate)
                     .env("RUSTC", &self.kani_compiler)
-                    // Use CARGO_ENCODED_RUSTFLAGS instead of RUSTFLAGS is preferred. See
-                    // https://doc.rust-lang.org/cargo/reference/environment-variables.html
-                    .env("CARGO_ENCODED_RUSTFLAGS", rustc_args.join(OsStr::new("\x1f")))
+                    .pass_rustc_args(&rustc_args, args::PassTo::AllDependencies)
```
carolynzech: I think this is much clearer; nice!

kani-driver/src/call_cargo.rs
AlexanderPortland: This was largely because I was fearful about whether passing any kind of arguments with PassTo::AllCrates multiple times would work (since it uses an environment variable and presumably may overwrite previous calls). I will do testing on this though and change the API to fit.

AlexanderPortland:
Looking into Carolyn's earlier comment ([here](https://github.com/model-checking/kani/pull/4220#discussion_r2208586063)) I had two realizations that prompted changes:
1. it doesn't really make sense to ever pass `KaniArg`s to `AllCrates`, as the kani compiler will only parse and use them in the local crate anyways -> I've renamed `pass_kani_args()` to `pass_kani_args_to_local()` which now only passes to the local crate.
2. since calling pass_rustc_args w/ `PassTo::AllCrates` uses an environment variable, calling it multiple times will overwrite the impact of all previous calls. -> Since `pass_kani_args()` no longer calls it in this way indirectly, I've just added an assertion that will fail & suggest a fix if we ever make a change that accidentally does this in the future.

kani-driver/src/call_cargo.rs
AlexanderPortland: I had considered this but initially decided against it because the function felt like it was doing a specific encoding rather than a trivial conversion. This comment convinced me and I tried implementing it in this way, but that ran into conflicts with the generic implementations of `Into` for the new arg wrappers. 

Because of this, we would have to change both those implementations and `to_rustc_arg()` to not be generic, forcing us to have ~5 manually monomorphized impls of the former and ~3 of the latter. This felt like it might not be worth it, but if the benefit seems like it would be large, I can definitely do that!

AlexanderPortland:
I re-examined the code with feedback in mind (specifically the use of `pass_kani_args` and `to_rustc_args` feeling redundant) and came up with some changes.
- the redundancy was originally done out of fear that calling `pass_rustc_args` with `PassTo::AllCrates` multiple times would overwrite all but the last call (since we use an environment variable to encode the flags in that case). this turned out to be true, so I added an assertion to detect this and point to a fix.
- because of this, I thought it made sense to remove the `pass_kani_args` method and instead have users use the pub `to_rustc_args` to pass their `KaniArg`s into the same `pass_rustc_args` calls as their `RustcArg`s. This way, you can call `pass_rustc_args(..., PassTo::AllCrates)` only once on each command with all the arguments you'd like to encode into the environment variable. it's also fairly easy to just recreate the behavior of the old `pass_kani_args` in the few cases that it's needed.

carolynzech (approved):
🎉  Thanks! Excited about this

carolynzech:
https://github.com/model-checking/kani/pull/4213 merged first, so now we get this error in the queue:

```
   --> kani-driver/src/call_single_file.rs:167:24
    |
167 |             flags.push(format!("--harness {harness}"));
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `KaniArg`, found `String`
    |
    = note: this error originates in the macro `format` (in Nightly builds, run with -Z macro-backtrace for more info)
help: call `Into::into` on this expression to convert `std::string::String` into `KaniArg`
    |
167 |             flags.push(format!("--harness {harness}").into());
    |                                                      +++++++
```

@AlexanderPortland can you merge main into this branch and update the new code so that it compiles?

Issue #3006, opened 2024-02-07
Support harness partition


celinval:
Requested feature: Users should be able to partition a harness which will be verified as if it was two harnesses.
Use case: One way to mitigate the cost of verifying a harness is to split it one harness into multiple harnesses that contains different assumptions. This is however error prone, potentially impacting the soundness of user proof.

Test case:

```rust
/// Dummy function to be verified
pub fn target_fn(input: i32) {
    let val = if input > 0 {
        do_something_expensive(val)
    } else {
        do_something_else_expensive()
    };
    unsafe { yet_another_fn(val) }
}

/// First proof
#[kani::proof]
pub fn positive_harness() {
    let input = kani::any_where(|i| i > 0i32);
    target_fn(input)
}

/// Second proof
#[kani::proof]
pub fn negative_harness() {
    let input = kani::any_where(|i| i < 0i32); // This should've been i <= 0
    target_fn(input)
}
```

Note that we are missing the case where input is 0. And this can only be caught today by inspecting the proofs. Instead, it would be nice if Kani could allow users to do this as part of one harness. Something like:

```rust
#[kani::proof]
pub fn target_fn_harness() {
    let input: i32 = kani::any();
    kani::partition([i > 0, i <= 0], || target_fn(input));
}
```

The syntax shall be improved... But the idea would be to create a chain of harnesses, where each harness assumes a given condition. And we create a check to make sure that all the assumptions together cover all reachable cases.

celinval:
Yep! That's the plan.

PR #4228, opened 2025-07-15
RFC: Partitioned proofs

AlexanderPortland:
This discusses a prototype implementation of the feature suggested in #3006 that would allow users to soundly partition proofs into harnesses that could be verified separately and in parallel.

[Rendered](https://github.com/AlexanderPortland/kani/blob/harness-partitioning-rfc/rfc/src/rfcs/0014-harness-partition.md)


rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
+- **Feature Name:** Harness Partition (`harness-partition`)
+- **Feature Request Issue:** [#3006](https://github.com/model-checking/kani/issues/3006)
+- **RFC PR:** https://github.com/model-checking/kani/pull/4228
+- **Status:** Under Review
+- **Version:** 0
+- **Proof-of-concept:** [prototype on local branch](https://github.com/model-checking/kani/compare/main...AlexanderPortland:kani:harness-partitioning)
+
+-------------------
+
+## Summary
```
carolynzech, line 10: Formatting nit: Can you put each sentence on a separate line? It makes it easier to comment on a particular part of the text. I think as long as there aren't blank lines between the sentences they should still render as part of the same paragraph.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+
+It can often be useful to subdivide an expensive proof harness so that different parts of the input space are verified separately. Adding the built-in ability to partition a proof harness into different pieces (that each make differing assumptions about their inputs) could reduce the cost of expensive proofs, while allowing us to automatically check that the partitions cover the entire input space and, thus, will not affect soundness.
+
+## User Impact
+
+Imagine that you have a function to verify like the following (based on the example from [#3006](https://github.com/model-checking/kani/issues/3006)). Since there are two tricky to analyze function calls, but only one will ever be called on a given input, you might want to verify all inputs that will take the first branch separately from those that will take the second. This way, each solve would be smaller in isolation, and you'd be able to take advantage of CBMC's internal parallelism by running both proofs at once.
```
carolynzech:
```suggestion
Imagine that you have a function to verify like the following (based on the example from [#3006](https://github.com/model-checking/kani/issues/3006)). Since there are two tricky to analyze function calls, but only one will ever be called on a given input, you might want to verify all inputs that will take the first branch separately from those that will take the second. This way, each solve would be smaller in isolation, and you could use Kani's parallel proof runner to run both proofs at once.
```

carolynzech: bc we want to focus on a Kani user and how they'd run the proof in parallel, and technically Kani users shouldn't have to know about CBMC.

carolynzech: I would put the code example after the sentence ending in "like the following," then make the point about partitioning. You can then refer to the input as a concrete source element, `input`, to tie it more closely to the example.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+
+#[kani::proof]
+pub fn second_branch_harness() {
+    let input = kani::any_where(|i: &i32| *i < 0i32); // This should've been i <= 0
+    assert!(target_fn(input) > 0)
+}
+```
+
+However, this can affect soundess, as there's no guarantee that your partitions will fully span the space of possible inputs. The only way to determine that a set of proofs like the one above are incorrect (as it forgets to account for when `i == 0`) is by manual inspection, which gets infeasible for proofs with complex types or partition rules.
+
```
carolynzech, line 63: I think you can also make the point that it increases maintenance and proof writing burden, since now instead of having one proof to maintain you have partition# of proofs.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+}
+
+#[kani::proof]
+pub fn second_branch_harness() {
+    let input = kani::any_where(|i: &i32| *i < 0i32); // This should've been i <= 0
+    assert!(target_fn(input) > 0)
+}
+```
+
+However, this can affect soundess, as there's no guarantee that your partitions will fully span the space of possible inputs. The only way to determine that a set of proofs like the one above are incorrect (as it forgets to account for when `i == 0`) is by manual inspection, which gets infeasible for proofs with complex types or partition rules.
```
carolynzech: One example of a proof with complex partition rules is [unchecked multiplication in the standard library](https://github.com/model-checking/verify-rust-std/blob/1c4ea17a99b9202f96608473083998b116bb6508/library/core/src/num/mod.rs#L1818-L1836). Could be worth linking to as an example.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+
+It would be helpful if Kani provided this as a built-in feature that could reason about given partition conditions to provide soundess guarantees.
```
carolynzech:
```suggestion
Instead, Kani should provide a feature to specify partition conditions for a given harness, automatically checking that the partitioned harness cover the entire input space.
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Open questions
+
+1. What's the best way to ensure that type errors in the `kani::partitioned_proof` macro are clear, understandable and actionable for users?
+2. Can function names be generated more prettily while still remainging unique? (the current hashes are often opaque for users)
+3. Are there any additional correctness issues introduced by this approach?
+4. Is there a way to simplify the generated `..._full_coverage()` proof harness? The iteration seems to be instrumented by our code which adds additional checks.
+5. Is there a way to promote the fact that verification failed on the `..._full_coverage()` proof harness as an error that could affect soundess? If Kani's users tend to look into any failed checks, this may not be needed, but it may be possible for a single failure to get lost in big projects.
+6. Would it be desirable to add an optional argument that specifies which subset of inputs you're trying to verify with the proof as a whole (if not provided just the whole input space)?
```
carolynzech: Yeah, this is my biggest question--I could see users wanting to create a partition that only covers a subset of the input space by design. For example, for multiplication on 64 bit integers, maybe you do only care about certain ranges of integers and are content to ignore the ones in the middle.

I could imagine a sister `kani::unsafe_partition` that provides the same semantics as partition but without the coverage check.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+}
+```
+
+However, this can affect soundess, as there's no guarantee that your partitions will fully span the space of possible inputs. The only way to determine that a set of proofs like the one above are incorrect (as it forgets to account for when `i == 0`) is by manual inspection, which gets infeasible for proofs with complex types or partition rules.
+
+It would be helpful if Kani provided this as a built-in feature that could reason about given partition conditions to provide soundess guarantees.
+
+## User Experience
+
+The current thought is for users to interact with this using a new `#[kani::partitioned_proof()]` attribute where they provide the conditions by which to partition the input space of their proof. Each condition must be of type `fn(&T) -> bool` (where `T` is the input type to the proof body that implements `kani::Arbitrary`) and is used to filter which values are part of that partition.
```
carolynzech:
```suggestion
Kani will introduce a `#[kani::partitioned_proof(cond)]` attribute, where `cond` specifies how to partition the input space of the proof. Each condition must be of type `fn(&T) -> bool` (where `T` is the input type to the proof body that implements `kani::Arbitrary`) and is used to filter which values are part of that partition.
```
I had trouble understanding the second sentence on a first read, since I'm used to Kani proofs not having any inputs. I would move the example to be after the first sentence, then point out explicitly that this proof takes inputs (unlike regular Kani proofs, which forbid that) and then proceed with the second sentence.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Software Design
+
+A prototype of this design has been implemented locally [here](https://github.com/AlexanderPortland/kani/tree/harness-partitioning).
+
+It introduces the `#[kani::partitioned_proof()]` attribute as variant of `#[kani::proof]` that takes in a set of closures representing the partition bounds. With those bounds, it keeps the main harness function the same, but does not annotate it with the `#kani_attributes` that would typically mark it as an entrypoint for Kani's verification. Instead, for each partition, it will generate a new function definition for that portion of the proof. These functions are just wrappers that constrain an arbitrary value to be within the partition and then call the real harness that will contain the meat of the proof. Their names are derived from a hash of their condition closure for uniqueness (but see [below](#open-questions) for discussion on how to improve this). So, for the running example above, the call to `kani::partitioned_proof` generates:
+
+```rust
+#[kani::proof]
+//                         (hash of |a| *a > 0)
```
carolynzech:
```suggestion
// (hash of |a| *a > 0)
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+pub fn partitioned_harness_7199877664941740246() {
+    let t = kani::any_where(|a: &i32| { *a > 0 });
+    partitioned_harness(t)
+}
+
+#[kani::proof]
+//                         (hash of |a| *a < 0)
```
carolynzech:
```suggestion
// (hash of |a| *a < 0)
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+A prototype of this design has been implemented locally [here](https://github.com/AlexanderPortland/kani/tree/harness-partitioning).
+
+It introduces the `#[kani::partitioned_proof()]` attribute as variant of `#[kani::proof]` that takes in a set of closures representing the partition bounds. With those bounds, it keeps the main harness function the same, but does not annotate it with the `#kani_attributes` that would typically mark it as an entrypoint for Kani's verification. Instead, for each partition, it will generate a new function definition for that portion of the proof. These functions are just wrappers that constrain an arbitrary value to be within the partition and then call the real harness that will contain the meat of the proof. Their names are derived from a hash of their condition closure for uniqueness (but see [below](#open-questions) for discussion on how to improve this). So, for the running example above, the call to `kani::partitioned_proof` generates:
+
+```rust
+#[kani::proof]
+//                         (hash of |a| *a > 0)
+pub fn partitioned_harness_7199877664941740246() {
+    let t = kani::any_where(|a: &i32| { *a > 0 });
+    partitioned_harness(t)
```
carolynzech: But where is `partitioned_harness` defined? Presumably it needs to inject that too?

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+
+#[kani::proof]
+//                         (hash of |a| *a < 0)
+pub fn partitioned_harness_2423413845937420568() {
+    let t = kani::any_where(|a: &i32| { *a < 0 });
+    partitioned_harness(t)
+}
+```
+
+It will then also inject a new proof that has an assert to ensure that all possible values fall into one of the given partitions. In this case, it will fail to be verified (as `t` can be 0), telling the user that their partitions are not complete and could affect proof soundness.
```
carolynzech:
```suggestion
It will then also inject a new proof that has an assert to ensure that all possible values fall into one of the given partitions. In this case, it will fail to be verified (as `t` can be 0), telling the user that their partitions are not complete and could affect proof soundness:
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+    let t: i32 = kani::any();
+    let partitions: [fn(&i32) -> bool; 2] = [|a: &i32| { *a > 0 }, |a: &i32| { *a < 0 }];
+    let partitions_have_full_coverage: bool = partitions.into_iter().any(|condition| condition(&t));
+    assert!(partitions_have_full_coverage)
+}
+```
+
+### Corner cases
+1. This current implementation could run into issues if the main proof harness & partition condition closures both don't specify concrete types (e.g. by using generics), as the compiler may not be able to determine what `T` to use `kani::any()` on. This should be fixed by explicitly enforcing the current implicit assumption that the input type is a concrete type.
+2. This implementation doesn't play nicely if your proof is intentionally only covers a subspace of inputs (e.g. if you're trying to partition a proof that initially used `kani::any_where(...)`). (see the open questions [below](#open-questions) for thoughts on how to fix this)
```
carolynzech: Can you be more concrete about what "doesn't play nicely" means? I assume it'll just fail verification?

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+
+The current thought is for users to interact with this using a new `#[kani::partitioned_proof()]` attribute where they provide the conditions by which to partition the input space of their proof. Each condition must be of type `fn(&T) -> bool` (where `T` is the input type to the proof body that implements `kani::Arbitrary`) and is used to filter which values are part of that partition.
+
+For example, the above would become
+
+```rust
+#[kani::partitioned_proof(|a: &i32| { *a > 0 }, |a: &i32| { *a < 0 })]
+pub fn partitioned_harness(input: i32) {
+    assert!(target_fn(input) > 0)
+}
```
carolynzech: I don't love that there's an implicit assumption here that `input` is a `kani::any()`

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+The `kani::partition([i > 0, i < 0], || target_fn(input))` syntax suggested ([here](https://github.com/model-checking/kani/issues/3006#issue-2123964835)) in the initial issue is clear, but runs into some implementation issues irregardless of whether it's a function call or macro. The goal of this feature is to generate multiple proof harnesses from the single call to `partition`, and this is very difficult to do that from a macro that's already inside a function (as your codegen is limited to within that function's scope).
+
+The design described above is simpler as it is an attribute macro and, thus, has access to the program's global scope, allowing it to keep the proof body the same and simply generate additional function definition wrappers for each partitioned proof.
+
+This was also initially implemented as an addition to the existing `#[kani::proof]` attribute, where users would specify partition conditions with `kani::proof(partitions = [|a| a % 2 == 9, ...])`. However, since these kinds of proofs require that the annotated function take an input, while `#[kani::proof]` functions typically don't, it seemed more consistent to introduce a separate attribute macro.
```
carolynzech: I would rephrase this as "We could also specify it as an addition" and change the past tense ("seemed" -> "seems")--RFCs should be more centered on UX/interface than implementation history.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+
+### 2. Checking for overlapping partitions
+As a corellary to checking that partitions span the space of all possible inputs, I had initially considered giving a warning if partitions are overlapping as overlap could indicate ill-defined bounds. However, I decided against this, as overlapping partitions may be useful in certain cases where the user wishes to purposefully oversimplify their partition conditions in a way that will cause overlap but make them easier to express.
```
carolynzech:
```suggestion
As a corollary to checking that partitions span the space of all possible inputs, I had initially considered giving a warning if partitions are overlapping as overlap could indicate ill-defined bounds. However, I decided against this, as overlapping partitions may be useful in certain cases where the user wishes to purposefully oversimplify their partition conditions in a way that will cause overlap but make them easier to express.
```

carolynzech: Can you give an example of such a case?

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Open questions
+
+1. What's the best way to ensure that type errors in the `kani::partitioned_proof` macro are clear, understandable and actionable for users?
+2. Can function names be generated more prettily while still remainging unique? (the current hashes are often opaque for users)
```
carolynzech: So it matters because in the verification output they're going to see "Checking harness ... hash"? I would add a section to the user experience section about what the output looks like and how someone debugs a failure. In particular, if I partition my proof in half and one of the harnesses fails, how do I know which partition is the problem?

carolynzech: One way to address this could be to inject an `assert!` with a static string of the partition closure. You could also just call them `harness_name_partition_i`, where `i` is the index of the partition in the list in the `partitioned_proof` attribute.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+    let input = kani::any_where(|i: &i32| *i < 0i32); // This should've been i <= 0
+    assert!(target_fn(input) > 0)
+}
+```
+
+However, this can affect soundess, as there's no guarantee that your partitions will fully span the space of possible inputs. The only way to determine that a set of proofs like the one above are incorrect (as it forgets to account for when `i == 0`) is by manual inspection, which gets infeasible for proofs with complex types or partition rules.
+
+It would be helpful if Kani provided this as a built-in feature that could reason about given partition conditions to provide soundess guarantees.
+
+## User Experience
```
carolynzech, line 66: Add a sentence about the unstable flag that users need to pass to opt into this feature, something like `-Z partition-proof`

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+2. This implementation doesn't play nicely if your proof is intentionally only covers a subspace of inputs (e.g. if you're trying to partition a proof that initially used `kani::any_where(...)`). (see the open questions [below](#open-questions) for thoughts on how to fix this)
+
+## Rationale and alternatives
+
+### 1. Alternative APIs
+The `kani::partition([i > 0, i < 0], || target_fn(input))` syntax suggested ([here](https://github.com/model-checking/kani/issues/3006#issue-2123964835)) in the initial issue is clear, but runs into some implementation issues irregardless of whether it's a function call or macro. The goal of this feature is to generate multiple proof harnesses from the single call to `partition`, and this is very difficult to do that from a macro that's already inside a function (as your codegen is limited to within that function's scope).
```
carolynzech: My preference would be to go with the approach outlined in the issue, for two reasons:

1. It preserves the idiom that harnesses don't take inputs, meaning that if users want nondet inputs they need to create them explicitly with `kani::any`, and:
2. It more closely ties the partitioned values to the invocation of the target function, since the partitioned values are declared immediately before invoking the target function in a closure.

> The goal of this feature is to generate multiple proof harnesses from the single call to `partition`, and this is very difficult to do that from a macro that's already inside a function (as your codegen is limited to within that function's scope).

You can get around this problem by generating the extra harnesses in the compiler instead of in the macro expansion phase ([like autoharness does](https://github.com/model-checking/kani/blob/d49a9ff8f7df7aa347c7fd570971097cf89f69e2/kani-compiler/src/kani_middle/transform/automatic.rs#L300)). Then you can replace the partition call in the original harness with the coverage check.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+
+The design described above is simpler as it is an attribute macro and, thus, has access to the program's global scope, allowing it to keep the proof body the same and simply generate additional function definition wrappers for each partitioned proof.
+
+This was also initially implemented as an addition to the existing `#[kani::proof]` attribute, where users would specify partition conditions with `kani::proof(partitions = [|a| a % 2 == 9, ...])`. However, since these kinds of proofs require that the annotated function take an input, while `#[kani::proof]` functions typically don't, it seemed more consistent to introduce a separate attribute macro.
+
+### 2. Checking for overlapping partitions
+As a corellary to checking that partitions span the space of all possible inputs, I had initially considered giving a warning if partitions are overlapping as overlap could indicate ill-defined bounds. However, I decided against this, as overlapping partitions may be useful in certain cases where the user wishes to purposefully oversimplify their partition conditions in a way that will cause overlap but make them easier to express.
+
+## Open questions
```
carolynzech, line 199: Would we support partitioning on multiple nondet values, e.g. `target_fn(a, b)`?

carolynzech (commented):
Thanks! First round of comments.

rfc/src/rfcs/0014-harness-partition.md
tautschnig: More than once in this document: s/soundess/soundness/

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+Imagine that you have a function to verify like the following (based on the example from [#3006](https://github.com/model-checking/kani/issues/3006)). Since there are two tricky to analyze function calls, but only one will ever be called on a given input, you might want to verify all inputs that will take the first branch separately from those that will take the second. This way, each solve would be smaller in isolation, and you'd be able to take advantage of CBMC's internal parallelism by running both proofs at once.
+
+```rust
+pub fn target_fn(input: i32) -> isize {
+    if input > 0 {
+        hard_to_analyze_fn_1(input)
+    } else {
+        hard_to_analyze_fn_2(input)
+    }
+}
```
tautschnig: I must admit that I found this example a bit unconvincing, yet the standard library (look at `unchecked_mul` in `library/core/src/num/mod.rs`) does have ones did convince me of the overall idea. The reason I found this particular example unconvincing is that it seemed like the prime example of doing modular verification: you'd add pre- and post-conditions to `hard_to_analyze_fn_{1,2}`, verify them in isolation, and then verify `target_fn` with just their contracts.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+For example, the above would become
+
+```rust
+#[kani::partitioned_proof(|a: &i32| { *a > 0 }, |a: &i32| { *a < 0 })]
+pub fn partitioned_harness(input: i32) {
+    assert!(target_fn(input) > 0)
+}
+```
+
+And Kani would automatically handle checking the partition conditions for soundess and generating the partitioned proofs. Overlaps would be allowed as they don't affect soundess and may be useful in certain cases (see [below](#2-checking-for-overlapping-partitions) for more discussion).
```
tautschnig: Just curious: how do you actually accomplish the soundness check?

tautschnig (approved):
The various comments need to be driven to a conclusion, but I'm certainly supportive of the overall idea.

rfc/src/rfcs/0014-harness-partition.md
sgpthomas: I also found this example unconvincing. This is maybe my lack of experience with this code base, but it seems to me that this could be something that Kani does automatically? Because the partition is just exactly defined as the condition and the negation of the condition. I think a more complicated example, with a partition that isn't easily inferrable, would make this more convincing. Or saying that this lays the groundwork to make this kind of analysis automatic in the future.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,138 @@
...
+    let input = kani::any_where(|i: &i32| *i > 0i32);
+    assert!(target_fn(input) > 0)
+}
+
+#[kani::proof]
+pub fn second_branch_harness() {
+    let input = kani::any_where(|i: &i32| *i < 0i32); // This should've been i <= 0
+    assert!(target_fn(input) > 0)
+}
+```
```
sgpthomas, line 54: would we need a third harness that checks that the input partitions cover the input space? Even if that is something that is totally internal to the compiler, it would be nice to see what the statement of the proposition would be.

rfc/src/rfcs/0014-harness-partition.md
sgpthomas: Grammar: "if your proof is intentionally only" -> "if your proof intentionally covers only a subspace ..."

rfc/src/rfcs/0014-harness-partition.md
sgpthomas: I agree with Carolyn here, partitioning logically feels like an action that you are performing on the symbolic input, rather than an action on the proof harness. I think that the fact that under the hood you are generating multiple harnesses should be something that the user never needs to think about.

rfc/src/rfcs/0014-harness-partition.md
sgpthomas: Yeah I would also be interested in an example here

AlexanderPortland:
Carolyn and I just had a talk about the syntax thoughts discussed [above](https://github.com/model-checking/kani/pull/4228#discussion_r2210721204) and echoed by Sammy [here](https://github.com/model-checking/kani/pull/4228#discussion_r2211650258), and we started to converge on an idea that seems compelling. (Just dumping our thoughts here now until I have time to edit the RFC at a later date)

The initial syntax of `kani::partition([i > 0, i < 0], || target_fn(input))` has some great benefits, mainly a) tying the partition conditions directly to the part of the proof being partitioned (rather than associating them with the whole harness) and b) maintaining the very logical idiom that harnesses don't take any arguments.

However, this syntax also brings up some questions, such as how a condition like `i > 0` knows that it's operating over the `input` variable without relying solely on them sharing an ident, and how to ensure that `input` is filtered to be within the bounds of the partition after it has already been captured by the `|| target_fn(input)` closure.

To fix this, we discussed a new API where `partition` would be a function Kani users can call with the following signature. The compiler would then detect this call within a proof, and generate the multiple partition harnesses, with each `kani::assume`-ing for a different one of the partition conditions.

```rust
pub fn partition<T: Arbitrary, R, const N: usize>(
    variable: T,
    conditions: [fn(&T) -> bool; N],
    and_run: fn(T) -> R,
) -> R {
    // This inner code will likely be a compiler intrinsic, 
    // but is just included to get a sense of the logical shape for now
    let partition_num = /* DEPENDS ON PARTITION */;
    kani::assume(conditions[partition_num](&variable));
    and_run(variable)
}
```

Which could be called along the lines of 

```rust
#[kani::proof]
pub fn partitioned_harness2() {
    let input = kani::any();
    let r = kani::partition(input, [|a| *a > 10, ...], |input| target_fn(input));
}
```

Some key notes:
- for now, both the `conditions` and `and_run` variables are function pointers, which forces them to not capture any variables. Allowing capturing of non-deterministic values (e.g. those generated with `kani::any()`) feels like it could lead to correctness issues, but this may be an overapproximation of a fix.
- for now, the `T: Arbitrary` bound ensures partitioning can only be done on types that can be generated from `kani::any()`--that seems like the most clear set of use cases. However, it may eventually be useful to relax that restriction in some cases where a non-`Arbitrary` value is generated from an `Arbitrary` one, and i makes more sense for the partition to be defined on the latter.
- we'd likely want to prevent multiple calls to `kani::partition` within the same proof, as this could lead to exponential growth in the number of generated harnesses

(Will add these to the RFC's open questions when I have time)

rfc/src/rfcs/0014-harness-partition.md
AlexanderPortland: Do you think it makes sense to make users still provide a filter that they wish to cover, even in the `unsafe_partition` case? For most use cases I can imagine (including the 64-bit int multiplication), it seems like you'd still have a well-defined range of inputs you want to verify rather than entirely avoiding a coverage check. Although the latter definitely provides the user more freedom.

carolynzech: > For most use cases I can imagine (including the 64-bit int multiplication), it seems like you'd still have a well-defined range of inputs you want to verify rather than entirely avoiding a coverage check

Yeah, I think this is right. For the scope of this RFC, though, I would leave it as an open question; I don't think we should implement any kind of unsafe partition until the safe one is merged as an unstable feature. Once it is, we can open a tracking issue for open questions that links back to the RFC.

AlexanderPortland:
When working on a second draft of this RFC, I encountered a potential issue with the above syntax. Consider the following proof harness:

```rust
#[kani::proof]
pub fn strange_partitioned_harness() {
    let input: i32 = kani::any();

    let f: fn(&i32) -> bool = if input > 10 {
        |a: &i32| *a < 0
    } else {
        |a: &i32| *a < -2
    };

    kani::partition(input, [f, |a| *a >= 0], |i| target_fn(i));
}
```

Since one of the partition conditions depends on the non-deterministic partition variable, it seems almost non-sensical, as partition conditions should just be pre-defined filters to split that variable, rather than a function determined by runtime values.

To fix this, I tried forcing the conditions array to be compile-time evaluable by making it a `const` generic (with the signature below), but then the compiler complained saying "the type of const parameters must not depend on other generic parameters".
```rust
pub fn partition<T: kani::Arbitrary, R, const N: usize, const C: [fn(&T) -> bool; N]>(
    variable: T,
    conditions: C,
    and_run: fn(T) -> R,
) -> R
```
Is there a better way to ensure that the value of the function pointers we get can't be impacted by non-det values?

rfc/src/rfcs/0014-harness-partition.md
AlexanderPortland: The thought (and implementation in the previous prototype) is that you would generate a new proof harness that declares `input` as a `kani::any::<i32>()` and asserts that at least one of the partition conditions evaluates to true. This, conceptually, should ensure that any `i32` value will fall into at least one partition.

AlexanderPortland:
Second draft of this RFC! Note that in discussion w/ @carolynzech and @camelid, the issue raised in my comment above should be fixed in this newly proposed API if we also check the validity of partition conditions.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+}
+
+#[kani::proof]
+pub fn second_branch_harness() {
+    let input = kani::any_where(|i: &i32| *i < 0i32); // ERROR: This should've been i <= 0
+    assert!(target_fn(input) > 0)
+}
+```
+
+However, this strategy:
```
carolynzech:
```suggestion
One could also write preconditions for `very_complex_fn_{1_2}` that restrict `input` accordingly, write `proof_for_contract` harnesses for those, and then use `stub_verified` on the `target_fn` harness. 

However, either of these strategies:
```

carolynzech: @tautschnig mentioned in a previous review that we can use preconditions for this, so nice to acknowledge that's true while still not solving the problem of having to specify the partition conditions in multiple places.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+- **can affect soundness**--there's no guarantee that your partitions will fully span the space of possible inputs.
+The only way to determine that a set of proofs like the one above are incorrect (as it forgets to verify the value of 0 for `input`) is by manual inspection.
+This gets infeasible for proofs with complex partition rules like those found in the [proofs for the standard library's unchecked multiplication](https://github.com/model-checking/verify-rust-std/blob/1c4ea17a99b9202f96608473083998b116bb6508/library/core/src/num/mod.rs#L1818-L1836).
+- **increases user burden**--instead of having to write and maintain a single proof, the user now has to handle a proof for each partition.
+
+Instead, Kani should provide a feature to automatically partition a proof based on certain conditions and ensure that the newly partitioned harnesses are equivalent to a single-harness proof.
+
+## User Experience
+
+The current thought is for users to set a `-Z partition-proof` flag which will allow them to use a new `kani::partition` function.
```
carolynzech:
```suggestion
Kani will introduce a new `kani::partition` function, gated behind an unstable feature flag.
```
Let's leave the feature flag name out of the RFC and put it in the header once we're merging the initial implementation. (I realize that I told you to mention the name, so that's on me--I realized that [we've historically done the header approach](https://model-checking.github.io/kani/rfc/rfcs/0009-function-contracts.html) and I think it's better to do that and avoid implementation details in the user experience section).

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+When calling this function, they would provide the type of the partition variable to generate, as well as a set of partition conditions and a closure with the actual code to run on the partitioned variable (we'll call this the *partition closure*).
```
carolynzech:
```suggestion
When calling this function, the user would provide the type of the partition variable, as well as a set of partition conditions and a closure with the actual code to run on the partitioned variable (we'll call this the *partition closure*).
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+```
+Checking partition #1 (*input > 0) of harness partitioned_proof...
+CBMC 6.7.1 (cbmc-6.7.1)
+/* MORE CBMC OUTPUT */
+RESULTS:
+Check 1: partitioned_proof_partition_1.assertion.1
+	 - Status: SUCCESS
+	 - Description: "assertion failed: target_fn(input) > 0"
+	 - Location: src/pre_partition.rs:8:5 in function partitioned_proof_partition_1
+     - Partition: #1 - *input > 0
```
carolynzech: I don't understand this part--does this line get injected for every check? Feels redundant since the top line says what the partition condition is; feels like you can delete.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+Checking coverage of partitions for harness partitioned_proof...
+CBMC 6.7.1 (cbmc-6.7.1)
+/* MORE CBMC OUTPUT */
+RESULTS:
+Check 1: partitioned_proof_coverage.assertion.1
+	 - Status: FAILURE
+	 - Description: "partitions for partitioned_proof do not cover all possible values of the i32 type."
+	 - Location: src/pre_partition.rs:11:5 in function partitioned_proof_coverage
+```
+
```
carolynzech, line 113:
```suggestion

Note that the coverage check fails, since we're missing the `i = 0` case.
```

carolynzech: It'd be great if we could tell the user which section of the input space they're missing, but I'm not sure if there's a way to wrangle the CBMC output to get that information... @tautschnig any thoughts? (At the very least, let's add it as an open question).

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+1. one new proof harness for each partition
+
+In the above example, this would be along the lines of:
+
+```rust
+#[kani::proof]
+/// (auto-generated partition of `input` for condition `|a| *a > 0`)
+pub fn partitioned_proof_partition_1() {
+    /* OTHER CODE BEFORE THE CALL TO `kani::partition` IF THERE WAS ANY */
+    let partition_variable = kani::any();
```
carolynzech:
```suggestion
    let partition_variable = kani::any::<i32>();
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+}
+```
+
+2. an additional proof harness to check that the partitions fully cover the domain of the variable
+
+This would be along the lines of:
+```rust
+pub fn partitioned_proof_coverage() {
+    /* OTHER CODE BEFORE THE CALL TO `kani::partition` IF THERE WAS ANY */
+    let partition_variable = kani::any();
```
carolynzech:
```suggestion
    let partition_variable = kani::any::<i32>();
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+```rust
+#[kani::proof]
+pub fn strange_partitioned_proof() {
+    let f: fn(&i32) -> bool = if kani::any::<i32>() > 10 {
+        |a: &i32| *a < 0
+    } else {
+        |a: &i32| *a < -2
+    };
+
+    kani::partition([f, |a: &i32| *a < 0], ...);
```
carolynzech: You're missing the generic argument on `kani::partition` here (`kani::partition::<T>(...)`)

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+    let partition_variable = kani::any();
+
+    // checks that any possible value of partition_variable will fall into at least one partition.
+    let is_in_partitions = [|a: &i32| *a > 0, |a: &i32| *a < 0].into_iter().any(|cond| cond(&partition_variable));
+    kani::assert(is_in_partitions, "partition conditions for partitioned_proof do not cover all possible values of a i32.");
+}
+```
+
+### Corner cases
+Although the fact that the partition conditions are function pointers prevents them from capturing any local variables, their value can still be affected by non-deterministic values.
```
carolynzech:
```suggestion
Under the type signature of `kani::partition`, the user can generate the partition conditions conditionally, e.g.:
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+        |a: &i32| *a < 0
+    } else {
+        |a: &i32| *a < -2
+    };
+
+    kani::partition([f, |a: &i32| *a < 0], ...);
+}
+```
+
+The behavior of a partition in this case seems ill-defined, so any call to `kani::partition` with conditions that are not constant closures would be disallowed.
```
carolynzech:
```suggestion
The behavior of a partition in this case seems ill-defined, so Kani will detect cases where the partition conditions are not constant closures and emit a compile error.
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Rationale and alternatives
+
+### 1. Allowing overlapping partition conditions
+As a corollary to checking that partitions span the space of all possible inputs, I had initially considered giving a warning if partitions are overlapping as this could indicate ill-defined bounds.
```
carolynzech, line 194: I feel like a warning could be interesting future work but I agree it's not necessary for the feature to be useful

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+        |a: &i32| *a < -2
+    };
+
+    kani::partition([f, |a: &i32| *a < 0], ...);
+}
+```
+
+The behavior of a partition in this case seems ill-defined, so any call to `kani::partition` with conditions that are not constant closures would be disallowed.
+
+## Rationale and alternatives
```
carolynzech, line 186: I would expand this section a bit with some rationale for your design, namely:

- Why do we have one round of CBMC output per partitioned harness? (One could imagine a design where you combine the results back together in postprocessing so you can just display "Checking harness..."). I think the splitting is 1) easier to implement for a first draft and 2) easier to debug but has the con of being longer.
- Why is `kani::partition` a function, rather than a proof attribute or a different design?

You've done a lot of thinking to settle on this design, so it'd be good to reflect that in the writing by (briefly) mentioning other points in the design space and why you decided they weren't as good.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+### 1. Allowing overlapping partition conditions
+As a corollary to checking that partitions span the space of all possible inputs, I had initially considered giving a warning if partitions are overlapping as this could indicate ill-defined bounds.
+
+However, this kind of overlap would not affect the correctness of the partition, as every possible value of the partition variable would still be checked, just potentially more than once. Admittedly, it's fairly difficult to come up with a realistic use case in which an overlap is helpful for users, so whether this should be allowed is up for debate.
+
+## Open questions
+
+1. How can we ensure that type errors in the `kani::partition` function call are clear, understandable and actionable for users?
+2. How can we test that there are no correctness issues introduced by our partition implementation?
+3. Should verification fully panic if a partitioned proof is lacking full coverage or just have that one assertion fail verification?
```
carolynzech, line 203: Not sure I understand what you mean here--Kani certainly shouldn't ICE but I presume that's not your point? If at least one property fails, verification will fail.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
...
+4. What's the best way to provide support for partitioning `BoundedArbitrary` types?
+Potentially through a separate function `partition_bounded` which would be generic over `T: BoundedArbitrary` and a `const N: usize` which would be used for the call to `kani::bounded_any()`.
+5. How should users be able to control our coverage checks? Does it make sense to provide another function that makes the user specify a smaller domain that we can still prove coverage for (or skips those checks entirely)?
```
carolynzech:
```suggestion
5. How should users be able to control our coverage checks?
Users may only want to test a subset of the input space (e.g., only testing multiplication for certain integer ranges), but the current design mandates that they cover the whole input space.
We could potentially introduce a separate API, e.g. `unsafe_partition`, that allows this, or some other mechanism to skip the coverage check.
```

carolynzech (approved):
Thanks! This is looking great.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,199 @@
+- **Feature Name:** Partitioned Proofs (`partition-proof`)
+- **Feature Request Issue:** [#3006](https://github.com/model-checking/kani/issues/3006)
+- **RFC PR:** https://github.com/model-checking/kani/pull/4228
+- **Status:** Under Review
+- **Version:** 1
```
AlexanderPortland: Accidentally incremented this on re-write. Change it back

rfc/src/rfcs/0014-harness-partition.md
AlexanderPortland: My question was more along the lines of should we not even give a verification output and have a pretty internal kani error saying something like "your partition bounds are incomplete, so any output we could give you for this proof would be incomplete -> go fix that". Maybe a well-named property failing is already enough though

rfc/src/rfcs/0014-harness-partition.md
camelid: Potentially worth also allowing partition conditions that are paths to free function items (i.e. not fn ptrs in local variables).

rfc/src/rfcs/0014-harness-partition.md
AlexanderPortland: True! In the real implementation I think the compiler would be able to infer it based on the type of the conditions, but definitely better to be clear here.

rfc/src/rfcs/0014-harness-partition.md
zhassan-aws: We could run the coverage harness with concrete playback to generate a test.

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,204 @@
...
+- **RFC PR:** https://github.com/model-checking/kani/pull/4228
+- **Status:** Under Review
+- **Version:** 0
+- **Proof-of-concept:** None yet, the previous prototype uses an out of date API.
+
+-------------------
+
+## Summary
+
+It can often be useful to subdivide an expensive proof harness so that different parts of the input space are verified separately.
```
zhassan-aws, line 12: Worth pointing out that this is also known as proof by cases or case split

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,204 @@
...
+
+#[kani::proof]
+pub fn proof_harness() {
+    let input = kani::any();
+    assert!(target_fn(input) > 0)
+}
+```
+
+Since there are two tricky to analyze function calls, but only one will ever be called on a given input, you might want to verify all values of `input` where `input > 0` that will take the first branch separately from those that will take the second.
+This way, each solve will only have to reason about two of the three complex function calls, and you could use Kani's parallel proof runner to run both proofs at once.
```
zhassan-aws:
```suggestion
This way, each verification run will only have to reason about two of the three complex function calls, and you could use Kani's parallel proof runner to run both proofs at once.
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,204 @@
...
+    let input = kani::any_where(|i: &i32| *i < 0i32); // ERROR: This should've been i <= 0
+    assert!(target_fn(input) > 0)
+}
+```
+
+One could also write preconditions for `very_complex_fn_{1_2}` that restrict `input` accordingly, write `proof_for_contract` harnesses for those, and then use `stub_verified` on the `target_fn` harness. 
+
+However, either of these strategies:
+- **can affect soundness**--there's no guarantee that your partitions will fully span the space of possible inputs.
+The only way to determine that a set of proofs like the one above are incorrect (as it forgets to verify the value of 0 for `input`) is by manual inspection.
```
zhassan-aws:
```suggestion
The only way to determine that a set of proofs like the one above are incomplete (as it forgets to verify the value of 0 for `input`) is by manual inspection.
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,204 @@
...
+
+Kani will introduce a new `kani::partition` function, gated behind an unstable feature flag.
+When calling this function, the user would provide the type of the partition variable, as well as a set of partition conditions and a closure with the actual code to run on the partitioned variable (we'll call this the *partition closure*).
+
+For example, the above would become
+
+```rust
+#[kani::proof]
+pub fn partitioned_proof() {
+    kani::partition::<i32>([|a| *a > 0, |a| *a < 0], |input: i32| target_fn(input));
```
zhassan-aws: Should the assertion be added  make the harness complete?
```rust
    assert!(target_fn(input) > 0)
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,204 @@
...
+
+```rust
+pub fn partition<T: Arbitrary, R, const N: usize>(
+    conditions: [fn(&T) -> bool; N],
+    and_run: fn(T) -> R, // the "partition closure"
+) -> R
+```
+
+This signature makes a few key design decisions:
+- *It takes in an array of `conditions` with the specific length `const N`*, ensuring the number of harnesses we have to generate is knowable at compile time when we're doing partition generation.
```
zhassan-aws:
```suggestion
- *It takes in an array of `conditions` with the specific length `const N`*, ensuring the number of harnesses we have to generate is known at compile time when we're doing partition generation.
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,204 @@
...
+- *Partition conditions & the partition closure are represented by function pointers*.
+This allows the user to provide closures of the same signature, but only if those closures do not capture any state.
+Allowing conditions closures to capture runtime state like local variables would be unwise as partitions are conceptually static divisions of a proof which should not depend on runtime values.
+- *The function returns the same type `R` as the underlying partition closure*, with the thought being that value could then be used in later computation without affecting the mechanics of a partition.
+- *The partition variable is bound by `T: Arbitrary`*.
+This would allow us to construct a new non-deterministic value with `kani::any()` that can then be passed to the partition closure.
+
+When the `kani-compiler` encounters a call to this function, it will transparently replace the entire `#[kani::proof()]` that called it with a set of new generated functions that, as a whole, will represent the partitioned proof. This new partitioned proof would contain:
```
zhassan-aws:
```suggestion
When the `kani-compiler` encounters a call to this function, it will transparently replace the entire `#[kani::proof]` that called it with a set of new generated functions that, as a whole, will represent the partitioned proof. This new partitioned proof would contain:
```

rfc/src/rfcs/0014-harness-partition.md
```diff
@@ -0,0 +1,204 @@
...
+
+In the above example, this would be along the lines of:
+
+```rust
+#[kani::proof]
+/// (auto-generated partition of `input` for condition `|a| *a > 0`)
+pub fn partitioned_proof_partition_1() {
+    /* OTHER CODE BEFORE THE CALL TO `kani::partition` IF THERE WAS ANY */
+    let partition_variable = kani::any::<i32>();
+    kani::assume((|a: &i32| *a > 0)(&partition_variable));
```
zhassan-aws, line 148: Neat!

rfc/src/rfcs/0014-harness-partition.md
zhassan-aws: Also, don't we need to specify the other generics?
```suggestion
    kani::partition::<i32, _, _>([|a| *a > 0, |a| *a < 0], |input: i32| target_fn(input));
```

zhassan-aws (commented):
Nice, thanks! Some minor comments.

rfc/src/rfcs/0014-harness-partition.md
carolynzech: Nice idea! @AlexanderPortland let's add this to the RFC as an open question and put this coverage check as a potential solution... don't think this is necessary for a first implementation but a nice-to-have for sure.

rfc/src/rfcs/0014-harness-partition.md
carolynzech: I think compiler error is always better than verification error but not sure atm how you'd implement it on the compiler side. So not a huge deal.

rfc/src/rfcs/0014-harness-partition.md
AlexanderPortland: True! My thought had been that the code was slightly stylized (plus the generics could be inferred by the compiler based on the input type of the closure anyway) so there was no need to include the other two, but I'll add them for clarity!

PR #4232, opened 2025-07-22
Fix bug: Kani unwinds loops with contract in generic function (with -Z loop-contracts)

thanhnguyen-aws:
This PR fixes the bug that even with -Z loop-contracts, Kani still unwind the loop if the loop is inside a generic function.

### DIFF
--- a/kani-compiler/src/kani_middle/transform/loop_contracts.rs
+++ b/kani-compiler/src/kani_middle/transform/loop_contracts.rs
@@ -661,7 +661,16 @@ impl LoopContractPass {
                 let TyKind::RigidTy(RigidTy::Closure(_, genarg)) = arg_ty.kind() else {
                     return false;
                 };
-                let GenericArgKind::Type(arg_ty) = genarg.0[2] else { return false };
+                let mut fnptrpos = 0;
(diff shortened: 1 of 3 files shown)


kani-compiler/src/kani_middle/transform/loop_contracts.rs
```diff
@@ -661,7 +661,16 @@ impl LoopContractPass {
...
+                let mut fnptrpos = 0;
+                for (i, arg) in genarg.0.iter().enumerate() {
+                    if let GenericArgKind::Type(arg_ty) = arg
+                        && let TyKind::RigidTy(RigidTy::FnPtr(_)) = arg_ty.kind()
+                    {
+                        fnptrpos = i;
+                        break;
+                    }
+                }
+                let GenericArgKind::Type(arg_ty) = genarg.0[fnptrpos + 1] else { return false };
```
carolynzech, line 676: I don't understand how this logic fixes the problem; can you write comments about what it's doing?

thanhnguyen-aws: I added some comments

tautschnig (approved):

PR #4233, opened 2025-07-23
Ensuring that MIR constants are marked as static consts

### LINKED ISSUE #3905: Function contract fail
I tried this code:

```rust
#[ensures(|result| result.as_usize().is_power_of_two())]
pub const fn of<T>() -> Self {
    // This can't actually panic since type alignment is always a power of two.
    const { Alignment::new(mem::align_of::<T>()).unwrap() }
}

#[kani::proof_for_contract(Alignment::of)]
pub fn check_of_i32() {
    let _ = Alignment::of::<i32>();
}
```
and 

```rust
#[kani::proof]
pub fn check_of_i32() {
    let a = Alignment::of::<i32>();
    assert!(a.as_usize().is_power_of_two());
}
```

```
kani src/main.rs  -Z function-contracts
```

with Kani version: 0.59.0

I expected to see this happen:  The same results for both harnesses.

Instead, this happened: The kani::proof_for_contract failed but the kani::proof succeed

### LINKED ISSUE #4029: Spurious failure in tests/kani/FunctionContracts/modify_slice_elem.rs
This is a tracking issue for resolving the spurious failure with `tests/kani/FunctionContracts/modify_slice_elem.rs` which started failing with the 2025-04-05 toolchain upgrade. This was changed to a fixme test in #4018.

More details on the cause for spurious failure is in #4009.

vonaka:
This is an attempt to resolve #3905

Currently, we rely on the `mutability` of the `Allocation` when marking constants. However, my understanding is that for various internal reasons the compiler may use mutable allocations even for constants. Instead, we should probably track whether a variable is constant ourselves.

Currently, some tests are failing. Specifically, `tests/expected/function-contract/valid_ptr.rs` fails because, while checking `pre_condition::harness_invalid_ptr`, the output doesn't include

> Failed Checks: assertion failed: unsafe { read_ptr(ptr) } == -20

Instead, the assertion is `UNREACHABLE`, which honestly makes perfect sense to me (since the precondition doesn't hold for any input, we'll never get a chance to actually verify this assertion).

`tests/kani-fixme/FunctionContracts/modify_slice_elem_fixme.rs` also fails, which [hopefully indicates](https://github.com/model-checking/kani/issues/3905#issuecomment-3074072252) that this also resolves #4029 (though the relation between the issue and the PR is not obvious to me).

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/operand.rs
@@ -131,7 +131,7 @@ impl<'tcx> GotocCtx<'tcx> {
         // First try to generate the constant without allocating memory.
         let expr = self.try_codegen_constant(alloc, ty, loc).unwrap_or_else(|| {
             debug!("codegen_allocation try_fail");
-            let mem_var = self.codegen_const_allocation(alloc, None, loc);
+            let mem_var = self.codegen_const_allocation(alloc, None, loc, true);
             mem_var
                 .cast_to(Type::unsigned_int(8).to_pointer())
                 .cast_to(self.codegen_ty_stable(ty).to_pointer())
@@ -282,7 +282,7 @@ impl<'tcx> GotocCtx<'tcx> {
                     let GlobalAlloc::Memory(data) = GlobalAlloc::from(alloc_id) else {
                         unreachable!()
                     };
-                    let mem_var = self.codegen_const_allocation(&data, None, loc);
+                    let mem_var = self.codegen_const_allocation(&data, None, loc, false);
 
                     // Extract identifier for static variable.
                     // codegen_allocation_auto_imm_name returns the *address* of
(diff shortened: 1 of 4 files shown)


zhassan-aws (approved):
Thanks!

tests/expected/function-contract/valid_ptr.expected
```diff
@@ -4,9 +4,7 @@ Failed Checks: |result| kani::mem::can_dereference(result.0)
 VERIFICATION:- FAILED
 
 Checking harness pre_condition::harness_invalid_ptr...
-Failed Checks: assertion failed: unsafe { read_ptr(ptr) } == -20
```
vonaka, line 7: I believe this shouldn't fail but should be Unreachable. Just like in this example:

```rust
#[kani::requires(false)]
fn foo() -> u8 {
    42
}

#[kani::proof_for_contract(foo)]
pub fn check_foo() {
    let v = foo();
    assert_eq!(v, 0);
}
```

>  check_foo.assertion.1
> 	 - Status: UNREACHABLE
> 	 - Description: "assertion failed: v == 0"

tests/expected/function-contract/valid_ptr.expected
```diff
@@ -4,9 +4,7 @@ Failed Checks: |result| kani::mem::can_dereference(result.0)
...
 Failed Checks: Kani does not support reasoning about pointer to unallocated memory
-Failed Checks: dereference failure: pointer outside object bounds
```
vonaka, line 9: I believe this should succeed, just like in `post_condition::harness`

tautschnig:
Just for my own sanity: Confirming this is the right approach despite what I said in https://github.com/model-checking/kani/issues/4009#issuecomment-2812579323: CBMC's `is_nondet_initializable_static` will honour `CConstant` markers via `is_constant_or_has_constant_components`.

PR #4234, opened 2025-07-24
Autoharness: add support for references

tautschnig:
Creates a fresh local variable for each parameter the type of which is a reference to a type that we have (or can derive `Arbitrary` support for).


tests/script-based-pre/cargo_autoharness_filter/filter.expected
```diff
@@ -1,7 +1,11 @@
-Kani generated automatic harnesses for 43 function(s):
+Kani generated automatic harnesses for 45 function(s):
 +--------------------------+----------------------------------------------+
 | Crate                    | Selected Function                            |
 +=========================================================================+
+| cargo_autoharness_filter | no_harness::unsupported_no_arg_name          |
+|--------------------------+----------------------------------------------|
+| cargo_autoharness_filter | no_harness::unsupported_ref                  |
```
carolynzech: Can we change the names of these? It's confusing to call them unsupported when they are in fact now supported.

carolynzech: Actually, you should change the name of `unsupported_ref` and then change the argument type of `unsupported_no_arg_name` so that we can still test the printing of the `_` argument name.

kani-compiler/src/kani_middle/transform/automatic.rs
```diff
@@ -332,30 +352,21 @@ impl TransformPass for AutomaticHarnessPass {
...
+        let arg_locals = fn_to_verify_body
+            .arg_locals()
+            .iter()
+            .map(|local_decl| {
+                call_kani_any_for_ty(
+                    self.kani_any,
+                    &mut harness_body,
+                    local_decl.ty,
+                    local_decl.mutability,
+                    &mut source,
```
carolynzech, line 366: This is much nicer!

carolynzech (approved):
Can you also update our mdbook documentation to reflect this change?

tests/script-based-pre/cargo_autoharness_filter/filter.expected
tautschnig: Done!

PR #4239, opened 2025-07-28
Add unstable option assume-no-panic

tautschnig:
Compute verification results under the assumption that no panic occurs.


carolynzech (commented):
Can you provide motivation in the PR description? I'm mainly interested in what you think the use cases are for this feature.

tautschnig:
> Can you provide motivation in the PR description? I'm mainly interested in what you think the use cases are for this feature.

Good point :-) Done!

carolynzech:
> LGTM. I'm a bit unsure about the option name. I can't think of a better name though.

@zhassan-aws, @tautschnig and I spoke about this offline and I was thinking `prove-safety-only`, since `assume-no-panic` implies (to me) that we only mean `panic!`.

We also spoke offline about how I'd like to see more tests about how this impacts safety verification and user-defined assertions--looking at all callsites of `codegen_assert_assume`, there's ones like https://github.com/model-checking/kani/blob/313f9b18776f89183e9508a564ccf14290217b36/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs#L68-L76

that give me pause, since it seems like then we'll be assuming a safety condition holds? I'll admit I'm not entirely clear on why we call `codegen_assert` in some cases and `codegen_assert_assume` in others, so the fact that this PR only needs to change one is also confusing to me.

tautschnig:
> We also spoke offline about how I'd like to see more tests about how this impacts safety verification and user-defined assertions--looking at all callsites of `codegen_assert_assume`, there's ones like [...] that give me pause, since it seems like then we'll be assuming a safety condition holds? I'll admit I'm not entirely clear on why we call `codegen_assert` in some cases and `codegen_assert_assume` in others, so the fact that this PR only needs to change one is also confusing to me.

The key here are the changes in #3860: the `PropertyClass` will tell us whether we are dealing with a panic or possible undefined behavior. The example you listed above is a `SafetyCheck`, so `--prove-safety-only` will retain the assertion.

PR #4241, opened 2025-07-28
Upgrade Rust toolchain to 2025-07-15

tautschnig:
Relevant upstream PR:
- https://github.com/rust-lang/rust/pull/143848 (Rename `stable_mir` and `rustc_smir`)

The update was done via
```
git grep -l 'stable_mir::' | xargs sed -i 's/stable_mir::/rustc_public::/'
git grep -l 'rustc_smir::' | xargs sed -i 's/rustc_smir::/rustc_public_bridge::/'
```
followed by manual touch-ups in `kani-compiler/src/main.rs` and `kani-compiler/src/codegen_cprover_gotoc/codegen/function.rs` and then running `scripts/kani-fmt.sh`.

Resolves: #4229


carolynzech (commented):
I think we should also replace the StableMIR string inside our comments and documentation as well, so that later viewers of Kani don't need historical context to know what that refers to.

tautschnig:
> I think we should also replace the StableMIR string inside our comments and documentation as well, so that later viewers of Kani don't need historical context to know what that refers to.

Isn't there a difference between the general idea of a stable MIR representation and the specific crates (now called `rustc_public`) that implement it? I am not sure that upstream PR has generally done away with the name if the overall idea.

zhassan-aws (approved):

carolynzech:
> Isn't there a difference between the general idea of a stable MIR representation and the specific crates (now called rustc_public) that implement it? I am not sure that upstream PR has generally done away with the name if the overall idea.

My concern is more that future developers (or readers of our documentation) won't understand what all the references to StableMIR mean anymore. I think we typically use StableMIR as a proper noun, i.e., to refer to the crate. So you're right that the general idea holds, but I don't think that's how we're using it currently.

tautschnig:
> My concern is more that future developers (or readers of our documentation) won't understand what all the references to StableMIR mean anymore. I think we typically use StableMIR as a proper noun, i.e., to refer to the crate. So you're right that the general idea holds, but I don't think that's how we're using it currently.

Thanks, I have adjusted comments and docs as I deemed it appropriate.

PR #4257, opened 2025-07-31
Add heuristic to order harness codegen

### LINKED ISSUE #4248: Add heuristic for ordering harness codegen
**Proposed change:**
It might make sense to add a simple heuristic to determine the order in which we codegen harnesses.

**Motivation:**
When working on #4236, I noticed that the main compiler thread would sometimes get unlucky and try to codegen a very large harness last, meaning it would be done with all its work but have to wait on a worker thread to export a large goto file. Although the effect isn't all that large, it could be avoided if we did codegen for large harnesses first, so that the thread pool has a longer amount of time to write them to disk off the critical path of compilation.

The plot below (of `kani-compiler` threads' CPU usage over time) shows an example of this situation.

Image

AlexanderPortland:
## Motivation
When compiling with more than a single thread (as in #4236), the order in which we codegen harnesses can have a non-negligable impact on performance. Specifically, if we handle harness that will generate a lot of code near the end of compilation, the main compiler thread can get stuck waiting for worker threads to export that code, slowing down overall compilation. If we could ensure that more difficult to export harnesses are handled first, the thread pool will have more time to write them to disk off the critical path of compilation.

## Approach

This PR switches reachability analysis for all harnesses to happen **before codegen**. It can then use the information we gather from reachability to reorder codegen based on _the total number of items reachable from each harness_ (`MostReachableItems`). 

This heuristic was chosen because it is:
- **simple** -- only requiring a single field access for each harness, and taking only ~10µs to fully sort a typical `CodegenUnit` of 66 harnesses (acceptable given it took ~20ms to codegen the fastest harnesses in that same `CodegenUnit`)
- seemingly **effective** -- on the `s2n-codec` benchmark it only ever places a slower-to-export harness after a faster one if their actual export times were within 10ms (or 7%). Since the goal is just to roughly order harnesses based on order of magnitude, this is a very good result 🎉

### Results
This change made a small, but not insignificant **2.9% reduction in the end to end compile time of the standard library** (@ commit [`177d0fd`](https://github.com/model-checking/verify-rust-std/commit/177d0fd25fbf3cb82c7b8408140158af1e408ede)).

And below is a recreation of the graph of the compiler's per-thread CPU usage from #4248 now that codegen is ordered.
Screenshot 2025-07-31 at 12 16 24 PM

_While functionally complete on its own, this change will not really make a performance impact until #4236 is merged._

Resolves #4248


tautschnig:
Let's get #4236 merged first so that this PR can get concrete data points in its PR description.

tautschnig:
#4236 has now been merged, so this one should be ready to be acted upon.

kani-compiler/src/kani_middle/codegen_order.rs
```diff
@@ -0,0 +1,65 @@
...
+where
+    I: Iterator<Item = Vec<(HarnessWithReachable<'a>, BodyTransformation)>>,
+{
+    /// Apply an codegen ordering heuristic to an iterator over codegen units.
+    fn apply_ordering_heuristic<H: CodegenHeuristic>(self) -> impl Iterator<Item = I::Item> {
+        // Reorder harnesses within each codegen unit according to `T`.
+        self.map(|mut harnesses| {
+            reorder_harnesses::<H>(&mut harnesses);
+            harnesses
+        })
```
AlexanderPortland:
```suggestion
        }).collect::<Vec<_>>().into_iter()
```

I could not tell you why tbh, but testing is showing that forcing all of the ordering to happen up front with a collect rather than lazily with a map makes this change go from slightly worse to significantly better.

zhassan-aws (commented):
Thanks! Is it possible to add a test that ensures the order is what we expect it to be?

kani-compiler/src/kani_middle/codegen_order.rs
AlexanderPortland: I need to look into why this is, as collecting seems to sneakily guzzle memory (peak goes from normal ~2 GB to ~6 GB), I think bc it requires one transformer per codegen unit to be kept in memory? If this change is doable without that drawback, it may be worthwhile but, if not, it should totally be dropped

AlexanderPortland: The cause of the memory overhead is that, now that we are doing reachability in a single pass first, we're inadvertently **storing the reachable call tree for every harness** while we figure out which order to codegen them in. Definitely not ideal as this takes up a ton of space :(. I'll look into how to fix this and if other heuristics could help

feliperodri:
Pushed a new revision that gets this PR back to a compiling, reviewable state and resolves the peak-memory regression discussed in the earlier comments.

The branch no longer compiled. The last merge from `main` left `codegen_crate` half-refactored (`ordered_harnesses` was undefined while a leftover per-unit loop still referenced an out-of-scope `unit`). This was the root cause behind the failing CI jobs (`Kani CI`, `Check Std Verification`, `Release Bundle`, `Kani Extra`, and the format check). It now builds cleanly on the current toolchain.

### Bounded-memory rework

The previous approach ran reachability for every harness up front and **retained** each harness's full result (reachable set, call graph, and its own `BodyTransformation`) so it could sort before codegen. On the standard library this tripled peak memory (~2 GB → ~6 GB), because all of a codegen unit's harnesses were held in memory at once.

The new approach:

- Orders harnesses **within each codegen unit**, preserving `main`'s single shared-transformer-per-unit model.
- Retains **only the heuristic rating** — a single `usize` per harness — not the reachable set or call graph.
- Runs the rating pass with the unit's shared transformer, which **warms its body cache**; codegen then recomputes reachability but hits that cache, so only the (cheap) graph traversal is repeated, not body transformation.

Net effect: **peak memory matches codegening the unit in its original order**, and `codegen_items` is now byte-for-byte identical to the pre-PR version — per-harness codegen semantics are unchanged, only the order within a unit differs.

The ordering logic now lives in `kani_middle::codegen_order`: a `CodegenHeuristic` trait, the `MostReachableItems` heuristic, and an `order_harnesses` helper.

### Scope note

`MostReachableItems` orders harnesses **within** each codegen unit; units themselves are still processed in their original order. This matches the original intent, but flagging it in case global ordering was expected.

feliperodri (approved):

Issue #2247, opened 2023-02-28
Stabilize parallel proof runner


zhassan-aws:
Requested feature: `cargo kani` has an unstable `-j` option to verify harnesses in parallel that was introduced in #1726. This option is currently unstable. This is a tracking issue for stabilizing it.
Use case: Improve verification performance
Link to relevant documentation (Rust reference, Nomicon, RFC):

carolynzech:
Note: see https://model-checking.github.io/kani/rfc/intro.html for stabilization instructions.

Issue #4026, opened 2025-04-17
Document & stabilize the list subcommand


carolynzech:
Document the experimental list subcommand feature in our book and stabilize the feature.

carolynzech:
#4258 stabilizes it officially; still needs documentation.

PR #4258, opened 2025-07-31
[Breaking Changes] Remove unstable list feature and default memory checks

carolynzech:
Follow up to https://github.com/model-checking/kani/pull/4108

- Removes the positive default memory check options (the ones not prefixed with `--no`)
- Removes the unstable list option
- Also stops warning about `--unstable-options` for jobs

The rationale being that https://github.com/model-checking/kani/pull/4108 was in v0.63, so v0.66 should have been enough time for affected users to make changes.

Resolves https://github.com/model-checking/kani/issues/2247
Towards https://github.com/model-checking/kani/issues/4026

### DIFF
--- a/kani-driver/src/args/mod.rs
+++ b/kani-driver/src/args/mod.rs
@@ -416,7 +416,8 @@ impl VerificationArgs {
         }
     }
 
-    /// Given the string representation of an option, warn if it's enabled UnstableFeature::UnstableOptions is enabled.
+    /// Given the string representation of an option, warn if it's enabled while
+    /// UnstableFeature::UnstableOptions is also enabled.
     /// This is for cases where the option was previously unstable but has since been stabilized.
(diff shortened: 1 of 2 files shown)


kani-driver/src/args/mod.rs
```diff
@@ -416,12 +416,25 @@ impl VerificationArgs {
         }
     }
 
-    /// Given an argument, warn if UnstableFeature::UnstableOptions is enabled.
+    /// Given the string representation of an option, warn if it's enabled UnstableFeature::UnstableOptions is enabled.
```
tautschnig, line 419: Did you mean to remove "UnstableFeature::UnstableOptions is enabled" ? As is, this sentence doesn't parse for me.

tautschnig:
```suggestion
    /// Given the string representation of an option, warn if it's enabled while
    /// UnstableFeature::UnstableOptions is also enabled.
```

PR #4269, opened 2025-08-04
Lazily evaluate debug info

AlexanderPortland:
Internally, Kani uses the `call_with_panic_debug_info()` function to call closures with a string that explains what they're trying to do, allowing better error messages if they internally panics. However, this function is currently called for every harness we codegen and has to pretty format the harness' function instance--something that calls into rustc internals and takes a non-negligible amount of time.

We should instead accept a closure that can generate debug info, but only when (and if) it's needed. 

### Results
Based on initial flamegraphing, the formatting passed to `call_with_panic_debug_info()` went from taking 2.3% of codegen on `s2n-codec` to not being called a single time (since there are no compiler errors in a typical execution).

This change made a solid **4.6% reduction in the end to end compile time of the standard library** (at std commit [177d0fd](https://github.com/model-checking/verify-rust-std/commit/177d0fd25fbf3cb82c7b8408140158af1e408ede); assuming #4257, #4259 & #4268 are already merged into Kani).

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/utils/debug.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/utils/debug.rs
@@ -16,7 +16,8 @@ use tracing::debug;
 // Use a thread-local global variable to track the current codegen item for debugging.
 // If Kani panics during codegen, we can grab this item to include the problematic
 // codegen item in the panic trace.
-thread_local!(static CURRENT_CODEGEN_ITEM: RefCell<(Option<String>, Option<Location>)> = const { RefCell::new((None, None)) });
+type FindCurrentItemFn = Box<dyn Fn() -> String>;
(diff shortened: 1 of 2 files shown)


kani-compiler/src/codegen_cprover_gotoc/utils/debug.rs
```diff
@@ -36,9 +37,9 @@ static DEFAULT_HOOK: LazyLock<Box<dyn Fn(&panic::PanicHookInfo<'_>) + Sync + Sen
 
             // Print the current function if available
             CURRENT_CODEGEN_ITEM.with(|cell| {
-                let t = cell.borrow().clone();
-                if let Some(current_item) = t.0 {
-                    eprintln!("[Kani] current codegen item: {current_item}");
+                let t = cell.borrow();
+                if let Some(current_item) = &t.0 {
+                    eprintln!("[Kani] current codegen item: {}", current_item());
```
AlexanderPortland: Might be more descriptive as `get_current_item()`

PR #4291, opened 2025-08-12
Update README

carolynzech:
Update our README to make it clearer that Kani is useful for *both* safety and correctness, not mostly safety with correctness on the side.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,10 @@
 
 The Kani Rust Verifier is a bit-precise model checker for Rust.
 
-Kani is particularly useful for verifying unsafe code blocks in Rust, where the "[unsafe superpowers](https://doc.rust-lang.org/stable/book/ch19-01-unsafe-rust.html#unsafe-superpowers)" are unchecked by the compiler.
-___
-Kani verifies:
- * Memory safety (e.g., null pointer dereferences)
- * User-specified assertions (i.e., `assert!(...)`)
- * The absence of panics (e.g., `unwrap()` on `None` values)
- * The absence of some types of unexpected behavior (e.g., arithmetic overflows)
+Kani is useful for checking both safety and correctness of Rust code.
+- *Safety*: Kani automatically checks for many kinds of [undefined behavior](https://model-checking.github.io/kani/undefined-behaviour.html).
+This makes it particularly useful for verifying unsafe code blocks in Rust, where the "[unsafe superpowers](https://doc.rust-lang.org/stable/book/ch19-01-unsafe-rust.html#unsafe-superpowers)" are unchecked by the compiler.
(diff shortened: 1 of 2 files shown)


carolynzech:
cc @zhassan-aws since we discussed this change offline.

README.md
```diff
@@ -28,15 +25,12 @@ See [the installation guide](https://model-checking.github.io/kani/install-guide
...
-use my_crate::{function_under_test, meets_specification, precondition};
+use my_crate::{function_under_test, meets_specification};
 
 #[kani::proof]
 fn check_my_property() {
    // Create a nondeterministic input
-   let input = kani::any();
-
-   // Constrain it according to the function's precondition
-   kani::assume(precondition(input));
```
carolynzech, line 39: I deleted this since people should really use contracts for preconditions at this point, and I think it's an unnecessary complication for a README example. I'd like to keep this example as simple as possible and direct people to the tutorial for anything more complex.

I also added a type hint to `input`; I think it makes it a bit clearer that `kani::any` produces a nondet value of a *concrete type*, not some magic "nondet type" or something.

README.md
```diff
@@ -4,13 +4,10 @@
...
-___
-Kani verifies:
- * Memory safety (e.g., null pointer dereferences)
- * User-specified assertions (i.e., `assert!(...)`)
- * The absence of panics (e.g., `unwrap()` on `None` values)
- * The absence of some types of unexpected behavior (e.g., arithmetic overflows)
+Kani is useful for checking both safety and correctness of Rust code.
+- *Safety*: Kani automatically checks for many kinds of [undefined behavior](https://model-checking.github.io/kani/undefined-behaviour.html).
+This makes it particularly useful for verifying unsafe code blocks in Rust, where the "[unsafe superpowers](https://doc.rust-lang.org/stable/book/ch19-01-unsafe-rust.html#unsafe-superpowers)" are unchecked by the compiler.
+- *Correctness*: Kani automatically checks for certain behaviors that are likely incorrect (namely, panics and arithmetic overflows), although these checks can be disabled if desired. Kani also supports custom correctness properties, either in the form of assertions (`assert!(...)`) or [function contracts](https://model-checking.github.io/kani/reference/experimental/contracts.html).
```
zhassan-aws: I'm not sure I like the "behaviors that are likely incorrect" part, especially mentioning panics in there. I also don't think we need to mention that these checks can be disabled here (too much detail for the README). How about:
```suggestion
- *Correctness*: Kani automatically checks panics (e.g. `unwrap()` on `None`), arithmetic overflows, and custom correctness properties, either in the form of assertions (`assert!(...)`) or [function contracts](https://model-checking.github.io/kani/reference/experimental/contracts.html).
```
?

Issue #3999, opened 2025-04-07
Docs mention the removed function `kani::expect_fail`


sgpthomas:
The testing documentation (https://model-checking.github.io/kani/regression-testing.html?highlight=expect#testing-stages) tell you to use `kani::expect_fail` but this has been deleted and using `kani::cover` is preferred. We should update the documentation to reflect this change.

Issue #3568, opened 2024-10-03
Document Arbitrary trait


celinval:
Requested feature: Provide documentation of the Arbitrary trait (https://model-checking.github.io/kani/crates/doc/kani/trait.Arbitrary.html)
Use case: Users trying to understand how to use and implement the `Arbitrary` trait.
Link to relevant documentation (Rust reference, Nomicon, RFC):

Issue #3041, opened 2024-02-22
Document how to debug slowness of the proof


Kixunil:
The [documentation says](https://model-checking.github.io/kani/tool-comparison.html):

> Model checking allows you to prove non-trivial properties about programs, and check those proofs in roughly the same amount of time as a traditional test suite would take to run. The downside is many types of properties can quickly become "too large" to practically model-check, and so writing "proof harnesses" (very similar to property tests and fuzzer harnesses) requires some skill to understand why the solver is not terminating and fix the structure of the problem you're giving it so that it does. This process basically boils down to "debugging" the proof.

However I was not able to find any documentation on how to do such debugging. I've encountered some slow tests and would love to learn how to make them faster.

Kixunil:
Thanks! I will look into it. I suspect loops are what made my private experiment slow. There's another [public code we have that implements `u256`](https://github.com/rust-bitcoin/rust-bitcoin/blob/42e8f537e620013a53a7c6b924e6ed857ecba1f3/bitcoin/src/pow.rs) which obviously has to contain multiplication/modulo/division. I guess we'll just accept the slowness for now.

> Remove sources of nondeterminism from the program

Is there any way to do this that doesn't decrease the power of the proof? The linked issue suggests that it's better to just exhaustively test some values (bools?).

adpaco-aws:
> Thanks! I will look into it. I suspect loops are what made my private experiment slow. There's another [public code we have that implements `u256`](https://github.com/rust-bitcoin/rust-bitcoin/blob/42e8f537e620013a53a7c6b924e6ed857ecba1f3/bitcoin/src/pow.rs) which obviously has to contain multiplication/modulo/division. I guess we'll just accept the slowness for now.

Thanks for sharing your code! Please feel free to participate in issues related to slow operations or open new ones if they don't match your experience.

> Is there any way to do this that doesn't decrease the power of the proof?

I can't think of any, assuming that "power" refers to the "generality" of the proof. But it's OK to decompose one proof into multiple harnesses (where one input gets fixed to a different value each time) to make the individual harnesses more efficient. Of course, the more general proof would be preferred in most cases, but that might not be feasible in some of them.

Issue #2539, opened 2023-06-16
Documentation doesn't discuss solver tradeoffs and ways to speed up verification


adpaco-aws:
Our FAQ section could discuss what solvers are more appropriate to use in each case. Our team has some insight into this which could be valuable for many other users. In addition, there are certain options that we routinely use for speeding up harnesses like `--no-memory-safety-checks`. This is often appropriate if there is no usage of unsafe code, so we should mention it.

PR #4296, opened 2025-08-12
Kani Book Documentation Improvements

carolynzech:
Best reviewed commit by commit.

Resolves https://github.com/model-checking/kani/issues/4026
Resolves https://github.com/model-checking/kani/issues/3999
Resolves https://github.com/model-checking/kani/issues/3568
Resolves https://github.com/model-checking/kani/issues/2539
Resolves https://github.com/model-checking/kani/issues/3041

### DIFF
--- a/docs/src/debugging-slow-proofs.md
+++ b/docs/src/debugging-slow-proofs.md
@@ -1,0 +1,107 @@
+# Debugging Slow Proofs
+
+Kani uses SAT/SMT solvers to verify code, which can sometimes result in slow or non-terminating proofs. This chapter outlines common causes of slowness and strategies to debug and improve proof performance.
+
+## Common Causes of Slow Proofs
+
+### Complex/Large Non-deterministic Types
+Some types are inherently more expensive to represent symbolically, e.g. strings, which have complex validation rules for UTF-8 encoding,
+or large bounded collections, like a vector with a large size.
+
+### Large Value Operations
+Mathematical operations on large values can be expensive, e.g., multiplication/division/modulo, especially with larger types (e.g., `u64`).
+
+### Unbounded Loops
+If Kani cannot determine a loop bound, it will unwind forever, c.f. [the loop unwinding tutorial](./tutorial-loop-unwinding.md).
+
+## Debugging Strategies
+
+These are some strategies to debug slow proofs, ordered roughly in terms of in the order you should try them:
+
+### Use Different Solvers
+
+Kani supports multiple SAT/SMT solvers that may perform differently on your specific problem. Try out different solvers with the `#[kani::solver]` [attribute](./reference/attributes.md) or `--solver` option.
+
+### Remove Sources of Nondeterminism
+
+Start by replacing `kani::any()` calls with concrete values to isolate the problem:
+
+```rust
+#[kani::proof]
+fn slow_proof() {
+    // Instead of this:
+    // let x: u64 = kani::any();
+    // let y: u64 = kani::any();
+    
+    // Try this:
+    let x: u64 = 42;
+    let y: u64 = 100;
+    
+    let result = complex_function(x, y);
+    assert!(result > 0);
+}
+```
+
+If the proof becomes fast with concrete values, the issue is likely with the symbolic representation of your inputs. In that case, see you can [partition the proof](#partition-the-input-space) to cover different ranges of possible values, or restrict the proof to a smaller range of values if that is acceptable for your use case.
+
+### Reduce Collection Sizes
+
+Similarly, if smaller values are acceptable for your proof, use those instead:
+
+```rust
+#[kani::proof]
+fn test_with_small_collection() {
+    // Instead of a large Vec
+    // let vec: Vec<u8> = kani::bounded_any::<_, 100>();
+    
+    // Start with a small size
+    let vec: Vec<u8> = kani::bounded_any::<_, 2>();
+    
+    process_collection(&vec);
+}
+```
+
+### Partition the Input Space
+
+Break down complex proofs by partitioning the input space:
+
+```rust
+// Instead of one slow proof with large inputs
+#[kani::proof]
+fn test_multiplication_slow() {
+    let x: u64 = kani::any();
+    let y: u64 = kani::any();
+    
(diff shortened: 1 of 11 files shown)


docs/src/reference/list.md
```diff
@@ -0,0 +1,52 @@
...
++==================================================+
+|       | my_crate | example::verify::check_modify |
+|-------+----------+-------------------------------|
+|       | my_crate | example::verify::check_new    |
+|-------+----------+-------------------------------|
+| Total |          | 2                             |
++-------+----------+-------------------------------+
+```
+
+The "Contracts" table overviews which functions have contract attributes (`#[requires]`, `#[ensures]`, or `modifies`), and which harnesses exist for those functions.
```
tautschnig:
```suggestion
The "Contracts" table shows functions that have contract attributes (`#[requires]`, `#[ensures]`, or `modifies`), and which harnesses exist for those functions.
```

docs/src/regression-testing.md
```diff
@@ -68,6 +68,8 @@ We've extended
 Rust compiler testing framework) to work with these suites. That way, we take
 advantage of all `compiletest` features (e.g., parallel execution).
 
+A good rule of thumb is to put a test in the `kani` or `cargo-kani` folders if all we care about is codegen and verification succeeding. This is usually the case if we're fixing a code pattern that previously crashed, so a successful verification result is a sufficient test. Otherwise, if the output refers to UI, e.g., a compiler error with a highlighted span, put it in the `ui` or `cargo-ui` suites. If the output is for a CBMC-based property that we expect Kani to check (e.g., ensuring that a particular assertion fails), use the `expected` folder.
```
tautschnig:
```suggestion
A good rule of thumb is to put a test in the `kani` or `cargo-kani` folders if all we care about is codegen and verification succeeding. This is usually the case if we're fixing a code pattern that previously crashed, so a successful verification result is a sufficient test. Otherwise, if the output refers to UI, e.g., a compiler error with a highlighted span, put it in the `ui` or `cargo-ui` suites. If the output is for a CBMC-based property that we expect Kani to check (e.g., ensuring that a particular assertion passes or fails), use the `expected` folder. 
```

docs/src/reference/attributes.md
```diff
@@ -239,3 +240,14 @@ default one.
 **Replaces the function/method with name <original> with the function/method with name <replacement> during compilation**
 
 Check the [*Stubbing* section](../reference/stubbing.md) for more information about stubbing.
+
+## Contract Attributes
+
+There are numerous attributes for function and loop contracts. At present, these are:
+
+- Proof harness for contracts: `#[kani::proof_for_contract]`
```
tautschnig:
```suggestion
- Proof harness for contracts: `#[kani::proof_for_contract(name-of-function)]`
```

docs/src/reference/arbitrary.md
```diff
@@ -0,0 +1,120 @@
...
+#[kani::proof]
+fn verify_with_constraints() {
+    let x: u32 = kani::any();
+    kani::assume(x < 1000);  // Constrain x to be less than 1000
+    kani::assume(x % 2 == 0); // Constrain x to be even
+    
+    // Now x represents all even numbers from 0 to 998
+    my_function(x);
+}
+```
```
tautschnig:
```suggestion
Use `any_where` or `kani::assume()` to add constraints to arbitrary values:

```rust
#[kani::proof]
fn verify_with_constraints() {
    let x: u32 = kani::any_where(|t| *t < 1000); // Constrain x to be less than 1000
    kani::assume(x % 2 == 0); // Constrain x to be even
    
    // Now x represents all even numbers from 0 to 998
    my_function(x);
}
```
```

File: docs/src/reference/arbitrary.md [hunk 5]
```diff
@@ -0,0 +1,120 @@
...
+```rust
+impl Arbitrary for NonZeroU8 {
+    fn any() -> Self {
+        let val = u8::any();
+        kani::assume(val != 0);
+        unsafe { NonZeroU8::new_unchecked(val) }
+    }
+}
+```
+
```
tautschnig, line 113:
```suggestion

An alternative means to add value constraints is provided by the [Invariant trait](https://model-checking.github.io/kani/crates/doc/kani/invariant/trait.Invariant.html).

```

docs/src/debugging-slow-proofs.md
```diff
@@ -0,0 +1,107 @@
...
+
+// Insert harnesses for other ranges of `x` and `y`
+```
+
+See this [tracking issue](https://github.com/model-checking/kani/issues/3006) for adding support for such partitioning automatically.
+
+### Limit Loop Iterations
+
+First, try the `#[kani::unwind]` attribute or the `--unwind` option to limit [loop unwinding](./tutorial-loop-unwinding.md). If the proof fails because the unwind value is too low, but raising it causing the proof to be too slow, try specifying a [loop contract](./reference/experimental/loop-contracts.md) instead.
+
```
tautschnig:
```suggestion
```

docs/src/debugging-slow-proofs.md
```diff
@@ -0,0 +1,107 @@
...
+### Large Value Operations
+Mathematical operations on large values can be expensive, e.g., multiplication/division/modulo, especially with larger types (e.g., `u64`).
+
+### Unbounded Loops
+If Kani cannot determine a loop bound, it will unwind forever, c.f. [the loop unwinding tutorial](./tutorial-loop-unwinding.md).
+
+## Debugging Strategies
+
+These are some strategies to debug slow proofs, ordered roughly in terms of in the order you should try them:
```
tautschnig, line 19:
```suggestion
These are some strategies to debug slow proofs, ordered roughly in terms of in the order you should try them:

### Limit Loop Iterations

First, identify whether (unbounded) loop unwinding may be the root cause. Try the `#[kani::unwind]` attribute or the `--unwind` option to limit [loop unwinding](./tutorial-loop-unwinding.md). If the proof fails because the unwind value is too low, but raising it causing the proof to be too slow, try specifying a [loop contract](./reference/experimental/loop-contracts.md) instead.
```

Issue #4303, opened 2025-08-18
Kani fails to compile libcore at `nightly-2025-08-07`


zjp-CN:
Kani CI in verify-rust-std suddenly failed with no commit change of kani or verify-rust-std.

[This is a bug in kani](https://rust-lang.zulipchat.com/#narrow/channel/182449-t-compiler.2Fhelp/topic/Nightly.20compiler.20breaks.20at.20the.20same.20version/near/534886458). 

Commits for record:
* https://github.com/model-checking/verify-rust-std/commit/f66ba419780326bcce3bb51bbd99e2d26533d6d1
* https://github.com/model-checking/kani/tree/9deeec1bec16c4b1f1173d5d652f99d330a4650f
* cc https://github.com/model-checking/verify-rust-std/issues/465

---

We can bump toolchain to see if libcore can compile, or fix kani at the rustc version.

NOTE: the current toolchain in kani is nightly-2025-08-16.

tautschnig:
It seems that contracts and `const` traits interact in some way. Removing the contract on `get_unchecked_mut` makes compilation work. Note that the change from `~const` to `[const]` does not actually make a difference, i.e., this fails to compile even when flipping back to `~const`.

@zjp-CN in [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/182449-t-compiler.2Fhelp/topic/.E2.9C.94.20Nightly.20compiler.20breaks.20at.20the.20same.20version/near/534886458) you said: "Interestingly, at this toolchain version, ~const Trait just compiles!" -- can you elaborate on that?

For me, no past version of Kani compiles this code, which is consistent with our [metrics job suddenly failing](https://github.com/model-checking/verify-rust-std/actions/runs/17014456701) -- despite neither changing the library source code nor the Kani version used for that particular job. This suggests a change in a dependency. The differences that I can spot between the most recent working version and a failing one are:
* Working: syn v2.0.105, proc-macro2 v1.0.97
* Failing: syn v2.0.106, proc-macro2 v1.0.101
I am yet to figure out how to force use of the older version (or which dependency actually drags in those newer versions).

zjp-CN:
> Note that the change from `~const` to `[const]` does not actually make a difference, i.e., this fails to compile even when flipping back to `~const`.

Yeah. I noticed that too. I was confused by it. Now I think the error may result from syn.
  

> Removing the contract on `get_unchecked_mut` makes compilation work.
> This suggests a change in a dependency.
[...]

PR #4305, opened 2025-08-18
Upgrade syn dependency to the latest version

tautschnig:
Fixes compiling the standard library, and avoids pinning to a particular minor version.

Resolves: #4303

### DIFF
--- a/kani-compiler/Cargo.toml
+++ b/kani-compiler/Cargo.toml
@@ -23,7 +23,7 @@ serde = { version = "1", optional = true }
 serde_json = "1"
 strum = "0.27.1"
 strum_macros = "0.27.1"
-syn = { version = "2.0.72", features = ["parsing", "extra-traits"] }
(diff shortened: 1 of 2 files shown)


tautschnig:
> Which version of `syn` do we want to upgrade to? AFAIK, the change in this PR wouldn't affect the version used since "2.0.72" essentially means >= 2.0.72 (see [Version requirement syntax](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax)). For example, this mornings automatic cargo update upgraded the `syn` version: https://github.com/model-checking/kani/pull/4302/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eL2063.

Thank you, I had indeed misunderstood the version syntax. Closing as #4303 has indeed been fixed by  #4302.

PR #4313, opened 2025-08-21
Add a unified codegen cache

AlexanderPortland:
I’ve noticed that Kani will often repeatedly codegen elements that are similar or exactly the same (sometimes as many as ~500 times) while compiling a crate. The two biggest offenders being codegen-ing `rustc_public::Ty`s into `cbmc::Type`s and turning `rustc_public::Span`s into `cbmc::Location`s.

This PR introduces an extensible system for caching portions of Kani’s codegen, using it to speed up both cases mentioned above.

### Results

Testing on the standard library shows this cache maintains a **>98% cache hit rate**, reducing **end-to-end compile times by 20%**, all while only taking up only ~8MB of memory at peak size (<1% of Kani’s total average use).

### Other Ideas

I also tried to cache how we codegen `Rvalue`s and `FnAbi`s, but both failed. The former was far too dependent on the current state of the program, and I think the latter is already cached by the compiler’s query system, so my cache only made it slower when we had to actually query the compiler.

That being said, I’m sure there are more opportunities for caching in our codegen, so I designed the system from the start to try to make it easy to extend the cache with more kinds of elements.


tautschnig:
Generally looks good to me, but could you please got the extra mile and add a regression test that will measure the caching efficacy (producing basically the numbers that you put in the PR description, but perhaps for some different code base - maybe even Kani itself?). Just like you added end-to-end compile-time measurements it would be good to have caching numbers so that we know when some other, seemingly unrelated, change completely breaks caching.

AlexanderPortland:
Hey @tautschnig, would this be best as a CI job (similar to the compiler timing one) or a regression test with a certain cache hit rate hard coded to detect any regression past that? Today's my last day so just trying to plan out what's feasible in the time frame--CI jobs can be very slow to test in my experience. _[for my own memory he said a CI job would be best if possible haha]_

kani-compiler/src/codegen_cprover_gotoc/codegen/cache/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+//!
+//! To interact with the cache elsewhere in Kani, just call [cache_entry] and pass in a key type from where
+//! the cache is defined with the [implement_cache!] macro below. This will return a cache entry for the value type,
+//! which you can then call [or_insert_with](CacheEntry::or_insert_with) on to get the cached value, or insert one if
+//! none was found.
+//!
+//! If you wish to add another element to the cache, add another row to the [implement_cache!] macro call below,
+//! and this will allow you to use [cache_entry] with that new key type.
+
+use fxhash::FxHashMap;
```
tautschnig: Cf. #4341: this should be replaced by `rustc_data_structures::fx`.

tautschnig:
> Hey @tautschnig, would this be best as a CI job (similar to the compiler timing one) or a regression test with a certain cache hit rate hard coded to detect any regression past that? Today's my last day so just trying to plan out what's feasible in the time frame--CI jobs can be very slow to test in my experience. _[for my own memory he said a CI job would be best if possible haha]_

@AlexanderPortland Please let us know whether you think you could still find the time to do this. Thanks!

AlexanderPortland:
Hey @tautschnig! Sorry for the delay, classwork is picking up and I just got sick lol. I did the fxhash change and I'll likely have a free day to sit down and do the CI job too as long as you're okay to wait another few weeks (but I can't make any promises). If it's time critical to have this land, feel free to merge as is!

feliperodri (approved):

PR #4314, opened 2025-08-22
Combo of small performance changes

AlexanderPortland:
This PR is a pair of small performance changes. It used to be 4 (and thus would've been a real combo), but two involved caching so I've moved them under #4313 instead. These changes include:
1. passing information commonly used by our `GotocHook`s' `hook_applies` method as arguments, so that each hook doesn't have to recompute it on it's own
2. a small one-line change to compute a fn instance's name once instead of twice.

These are both very small, but together with the other two changes now moved to #4313, made a ~6% decrease in the compile time of the standard library.


thanhnguyen-aws (approved):

AlexanderPortland:
Hi @thanhnguyen-aws, thanks for the approval! Do you mind merging this too? I got removed from the github model-checking org and no longer can.

PR #4321, opened 2025-08-27
[WIP] Overwrite panic macros directly in libstd

bjorn3:
This is meant as an alternative to https://github.com/rust-lang/rust/pull/145734 that also works for non-builtin macros. It currently doesn't work yet as `kani::panic` can't be referenced from libstd as it doesn't depend on kani. Presumably the definition of this function can be moved into the panic overwrite file. I'm not familiar enough with how everything is set up to actually do this though. If someone with more Kani experience can pick this PR up, that would be appreciated. If you don't like the approach I've taken (eg it is considered too hacky), feel free to close.


zhassan-aws:
Nice, thanks @bjorn3! We'll test this out.

bjorn3:
Why did you assign me @feliperodri? Is the approach in this PR something you and the rest of the kani developers are fine with?

PR #4357, opened 2025-09-12
Do not run PR/issue-creating workflows in forks

tautschnig:
Avoid spamming forks with spurious pull requests that will only ever be merged in the primary repository.

### DIFF
--- a/.github/workflows/toolchain-upgrade.yml
+++ b/.github/workflows/toolchain-upgrade.yml
@@ -16,6 +16,10 @@ permissions:
 
 jobs:
   create-toolchain-pr:
+    # This workflow is restricted to the main repository (model-checking/kani) to prevent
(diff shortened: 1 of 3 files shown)


.github/workflows/toolchain-upgrade.yml
```diff
@@ -68,7 +72,7 @@ jobs:
             })
 
       - name: Create Issue
-        if: ${{ env.next_step == 'create_issue' && github.repository_owner == 'model-checking' }}
+        if: ${{ env.next_step == 'create_issue' }}
```
feliperodri, line 75: why do we need to remove the repository owner here?

tautschnig: It's now being checked for the entire workflow instead of just this one step.

PR #4427, opened 2025-10-25
fix: Make kani attribute nameres work with generic args having `::`

ShoyuVanilla:
I found this while working on https://github.com/model-checking/verify-rust-std/issues/382.

When you try to add `#[proof_for_contract(Rc::<MaybeUninit<T>, A>::assume_init)]` to a proof for this function https://doc.rust-lang.org/1.90.0/src/alloc/rc.rs.html#1211, kani fails to resolve `Rc::<MaybeUninit<T>, A>::assume_init`, as it compares the *last two parts* of the path string,

 https://github.com/model-checking/kani/blob/e0316ff45f7ccf724b70e28af80f719b784cfcc2/kani-compiler/src/kani_middle/resolve.rs#L783-L797

But rustc's `TyCtxt::def_path_str` emits the path `rc::Rc::<core::mem::MaybeUninit<T>, A>::assmue_init` even if there's `use core::mem::MaybeUninit` in scope. So the *last two parts* of it is `::MaybeUninit<T>, A>::assmue_init`, instead of being `::<core::mem::MaybeUninit<T>, A>::assmue_init`, and the resolution fails.


kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -125,21 +125,25 @@ fn resolve_path<'tcx>(
...
+                DefKind::Struct | DefKind::Enum | DefKind::Union => {
+                    resolve_in_type_def(tcx, base, &base_path_args, &name)
+                }
+                DefKind::Trait => resolve_in_trait_def(tcx, base, &name),
+                kind => {
+                    debug!(?base, ?kind, "resolve_path: unexpected item");
+                    Err(ResolveError::UnexpectedType { tcx, item: base, expected: "module" })
+                }
+            };
+            base.map(|base| (base, segment.arguments))
```
ShoyuVanilla, line 144: This is to fix argument passings when it is prefixed with module paths like `foo::bar::Baz::<..>`

kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -793,7 +821,7 @@ fn last_two_items_of_path_match(item_path: &str, generic_args: &str, name: &str)
     let last_two = format!("{}{}{}", generic_args, "::", name);
 
     // The last two components of the item_path should be the same as ::{generic_args}::{name}
-    last_two == actual_last_two
+    last_two.chars().eq(actual_last_two.chars().filter(|c| !c.is_whitespace()))
```
ShoyuVanilla, line 824: We strip whitespaces from the generic args here:

https://github.com/model-checking/kani/blob/e0316ff45f7ccf724b70e28af80f719b784cfcc2/kani-compiler/src/kani_middle/resolve.rs#L589-L591

So multiple args like `<T,  A>` become `<T,A>`. This is necessary for matching those things.

ShoyuVanilla:
Could this be reviewed or should I follow some extra policies I might be missing? This is needed to write [contracts for `Rc::<_>::assume_init`](https://model-checking.github.io/verify-rust-std/challenges/0026-rc.html#success-criteria) 😅

PR #2633, opened 2023-07-28
Add support to array-based SIMD

celinval:
### Description of changes: 

Originally, repr(simd) supported only multi-field form. An array based version was later added and it's likely to become the only supported way (https://github.com/rust-lang/compiler-team/issues/621).

The array-based version is currently used in the standard library, and it is used to implement `portable-simd`. This change adds support to instantiating and using the array-based version.

### Resolved issues:

Resolves #2590 
Resolves #1926

### Related RFC:

Optional #ISSUE-NUMBER.

### Call-outs:

We should still add more tests to check if all simd operations that Kani supports works with the array-based SIMD. I'm planning on doing that in a follow up PR.

### Testing:

* How is this change tested? New tests

* Is this a refactor change?

### Checklist
- [ ] Each commit message has a non-empty body, explaining why the change was made
- [ ] Methods or procedures are documented
- [ ] Regression or unit tests are included, or existing tests cover the modified code
- [ ] My PR is restricted to a single feature or bugfix

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

PR #4467, opened 2025-11-11
Fix SIMD projection mismatch for array-based SIMD types

### LINKED ISSUE #2264: projection mismatch
I tried this code:

```rust
<code>
```

```
<command>
```

with Kani version:

I expected to see this happen: *explanation*

tautschnig:
Fixes the projection mismatch error that occurs when using the `simd_as` intrinsic with array-based SIMD types. The issue manifested when executing code with SIMD types that use array representations (e.g., `#[repr(simd)] struct V<T>([T; 2]))`, causing projection mismatches and "unsupported constructs" warnings.

Resolves: #2264


zhassan-aws:
@tautschnig I fixed the expected case, but it seems the PR needs more work.

feliperodri:
Closing this PR, as it doesn't resolve #2264 and its code change is a no-op against current main. Summary of investigation (verified against main @ nightly-2026-03-02):
  - The projection-mismatch symptom is already fixed by #2633 ("Add support to array-based SIMD"). `codegen_simd_field` reinterpret-casts the SIMD vector to the inner array type, so the `.0/index` projection no longer mismatches, this landed before this PR's base and is in main.
  - The `check_expr_typ_mismatch` change is dead code. Reverting it changes zero test outcomes: the two transmute-based tests pass with or without it, and the `simd_as` test fails either way. Its `RigidTy::Array(..)` arm is also broader than intended. It would suppress genuine projection mismatches for any array where sizes match and one operand is a vector, which risks masking real bugs.
  - The issue's actual scenario still fails. #2264's reproducer uses `simd_as`, and `simd_as` has never been implemented in Kani (it falls through to `Intrinsic::Unimplemented` → "`simd_as` is not currently supported"). This PR doesn't add it, so the simd-as-array-based test can't pass which is the standing CI failure here.

What #2264 actually needs: an implementation of the `simd_as` intrinsic (a per-lane numeric cast), not a change to projection-mismatch detection. I'll leave #2264 open and re-scope it to that. If the two transmute-based array-SIMD tests are worth keeping, they'd be a welcome standalone regression PR for #2633. The analysis here helped pin down that the remaining gap is the missing intrinsic rather than the projection handling.

Issue #942, opened 2022-03-15
Design Machine Readable Output - RFC


jaisnan:
The Machine Readable Output Design needs to be approved by the Team once before merging - this can be done with a simple RFC doc.

PR #4472, opened 2025-11-13
Add `--export-json` for structured verification results

### LINKED ISSUE #598: Organize the output from Kani to make it more user friendly
An RMC run includes multiple stages (codegen, symtab2gb, cbmc, etc.) each of which produces its own output and warning/error messages. It's often daunting to find the relevant pieces of information in the output, for example the results of specific assertions (#477), or whether there were warnings in codegen that may impact soundness (#316), etc. The proposal is to reorganize the output from RMC as follows:

1. The output printed in the terminal (stdout) should be minimal; it should only include an indicator of the progress, e.g. which stage the run is in, the number of codegen'd crates out of the total, how many CBMC checks are solved so far, etc., and a summary at the end that includes the number of warnings/errors from the different stages, and how many checks passed or failed. This would be similar in spirit to the output from `cargo test` where by default no output is emitted for passing tests. The user should be able to increase the output verbosity via a switch.

2. (Optional) A log file that includes all the detailed output from all stages.

3. A report file (e.g. `rmc-report.txt`) that includes the verification results organized by category, where categories include user-specified assertions, arithmetic overflow checks, bound checks, etc. Each section should list the specific checks that were done, and the result of each.

### LINKED ISSUE #1045: kani-verifier tracking issue

### LINKED ISSUE #1194: Cleanup Kani's output

### LINKED ISSUE #1219: compiletest --help panics
```
cargo run -p compiletest -- --help
```

with Kani version: e31d1337093

I expected to see this happen: Just the help output

Instead, this happened: The help output followed by a panic:
```
$ cargo run -p compiletest -- --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/compiletest --help`
Usage: target/debug/compiletest [OPTIONS] [TESTNAME...]

Options:
        --compile-lib-path PATH
                        path to host shared libraries
        --run-lib-path PATH
                        path to target shared libraries
        --rustc-path PATH
                        path to rustc to use for compiling

<snip>

        --force-rerun   rerun tests even if the inputs are unchanged
    -h, --help          show this message
        --edition EDITION
                        default Rust edition


thread 'main' panicked at 'explicit panic', tools/compiletest/src/main.rs:158:9
```

### LINKED ISSUE #2562: RFC: Alternative check format
Requested feature: Kani checks are generally reported as:
```
Check <index>: <property name>
     - Status: <status>
     - Description: <description>
     - Location: <location> in <function>
```
There are a few problems with this output:
 1. `<index>` doesn't have a concrete purpose. These indexes are assigned while printing checks but they aren't used for anything else.
 2. `<property name>` are just the property names used by CBMC. The general format for property names is `<function>.<class>.<counter>` where `<function>` is the function name (note: this information gets repeated in the "Location" line), `<class>` is one of the property classes defined in CBMC, and `<counter>` is another index assigned in CBMC.
 3. The checks define the information they print on each check (i.e., the `<status>` is preceded by "Status", `<description>` is preceded by "Description:", etc.). This clutters the output and is unnecessary once the user is familiar with it. Definitions should come from the documentation instead.

Because of that, the requested feature is to use an alternative format that resembles the `rustc` output, containing not only the information being currently displayed, but also includes better diagnostics and potentially hints to resolve issues.

Use case: General Kani UI.

Link to relevant documentation (Rust reference, Nomicon, RFC): [Errors and Lints in `rustc` development](https://rustc-dev-guide.rust-lang.org/diagnostics.html)

Test case:

```rust
<code>
```

### LINKED ISSUE #2572: RFC: Other tool versions
Requested feature: Kani uses a bunch of other tools (engines, solvers, etc.) to do its job. At present, the only version being reported is CBMC's (and maybe some SAT solver through CBMC's output?), but it'd nicer to have versions for all the artifacts to be used printed after collecting harness metadata. What I'm considering now is something like this:

```
Versions:
 - engine: cbmc 5.89.0
 - solvers: minisat X.Y.Z, kissat A.B.C
 - viewer: ...
```

This would allow users to check all versions (except for Kani's version, unless we added it here) in a single place. However, we could also hide this information by default and enable it through `--verbose`.
Use case: Geared towards multi-harness verification in particular.

### LINKED ISSUE #2574: RFC: Output - Reduce reported checks/properties
Requested feature: Kani's output quickly gets cluttered due to the high number of checks being reported, and assumes that all properties have the same value. We need a proposal to change this. The goal is to:
 1. Reduce the number of checks presented to the user as much as possible.
 2. Keep checks that users care about.
It isn't entirely clear how to do (2), but one can filter in checks in the user's code or checks whose status is deemed relevant despite not being in the user code.

Use case: Standard verification reports.

### LINKED ISSUE #2636: Create user-friendly output for coverage
Requested feature: Enhance the output printed for coverage results, so that it's user-friendly. Currently, this is basically a CSV format:
```
<file>, <line>, <status>
```
But it's not the easiest to interpret.
Use case: Coverage results.
Link to relevant documentation (Rust reference, Nomicon, RFC): #2612 

Test case:

```rust
<code>
```

### LINKED ISSUE #3357: Remove terse output requirement for parallel jobs
Requested feature: Remove terse output requirement for parallel jobs 
Use case: Full output while still having parallel capability

Currently, a terse output is required for running harnesses in parallel. Removing that barrier would allow for a full output while still allowing for parallel running of harnesses.

yimingyinqwqq:
Add opt-in JSON export (--export-json <file>) to emit structured verification results (metadata, per-harness outcomes, CBMC stats).
Improves Kani by enabling reliable machine-readable output for external tools and applications.

Context: Current output is human-readable only, which blocks robust automation and integrations.

Manual tests:
	•	Run cargo kani --export-json out.json
	•	With multiple harnesses: counts in summary match executed harnesses.

Resolves #2572


tautschnig:
Thank you for your contribution! I notice that you include an RFC document in your pull request, but I think the expectation is that an RFC (a request-for-comment!) is submitted as a PR of its own to solicit discussion _before_ coming up with a major new feature or change. You don't need to close this PR, but I think it would be useful to separately have a discussion on this topic as the specific JSON structure is a one-way-door for future changes to it would risk breaking other tools.

yimingyinqwqq:
Thank you for your suggestion! We discussed with Rajath about this and user can opt in using the flag so hopefully nothing will be broken.

zhassan-aws:
Thanks for adding this feature. This has been longstanding!

I would recommend putting this feature under an unstable flag (see https://model-checking.github.io/kani/rfc/rfcs/0006-unstable-api.html) for some time until we ensure the json schema won't change.

kani-driver/src/args/mod.rs
```diff
@@ -242,6 +242,9 @@ pub struct VerificationArgs {
     #[arg(long)]
     pub default_unwind: Option<u32>,
 
+    #[arg(long = "export-json")]
```
zhassan-aws: Please add a doc comment so that it gets displayed in the help message.

zhassan-aws: Just do:
```suggestion
    #[arg(long)]
```
as underscores are automatically mapped to hyphens.

rfc/src/rfcs/0015-json-handler.md
```diff
@@ -0,0 +1,288 @@
...
+
+The validator supports a `--field-path` option for validating specific sections of the JSON (e.g., the `cbmc` block).
+
+### Test Suite Organization
+
+We add a new test suite under `tests/json-handler/` with four test scenarios that cover different aspects of JSON export:
+
+The `basic-export/` test verifies that the JSON export flag works and produces valid JSON with the expected top-level structure. The `schema-validation/` test is more comprehensive—it runs a verification with multiple harnesses and validates the entire JSON structure against the schema template. This directory also houses the canonical `kani_json_schema.json` file, making it easy to find and update.
+
+The `multiple-harnesses/` test specifically checks that results from multiple harnesses are correctly aggregated and that the `harness_id` correlation works across different blocks. Finally, `failed-verification/` tests that error information is correctly captured and that optional error fields are populated on failure.
```
zhassan-aws:
```suggestion
The `multiple-harnesses/` test specifically checks that results from multiple harnesses are correctly aggregated and that the `harness_id` correlation works across different blocks. Finally, `failed-verification/` tests check that error information is correctly captured and that optional error fields are populated on failure.
```

rfc/src/rfcs/0015-json-handler.md
```diff
@@ -0,0 +1,288 @@
...
+5. We serialize this struct to JSON format
+6. The JSON is written to the specified file, with any I/O errors reported to the user
+
+The schema structure organizes information by category (metadata, project info, harness details, results, CBMC stats) with `harness_id` serving as the correlation key across blocks. This organization makes it easy for external tools to either process everything or filter by specific harnesses.
+
+### Corner Cases
+
+Several edge cases need consideration:
+
+**CBMC output parsing**: We extract CBMC statistics using regex patterns. If CBMC's output format changes in future versions, parsing may fail. To handle this gracefully, we can omit CBMC statistics in the json schema.
```
zhassan-aws: It's likely more robust to use CBMC's `--json-ui` option (https://diffblue.github.io/cbmc/man/cbmc.html).

scripts/validate_json_export.py
```diff
@@ -0,0 +1,223 @@
...
+            return False
+
+    # Navigate to the field in both data and schema
+    parts = field_path.split(".")
+    current_data = data
+    current_schema = schema
+
+    for part in parts:
+        if part not in current_data:
+            print(f"ERROR: Field path '{field_path}' not found in data")
```
zhassan-aws: It would be useful to include `part` in the error message.

scripts/validate_json_export.py
```diff
@@ -0,0 +1,223 @@
...
+            return False
+        current_data = current_data[part]
+
+        if part not in current_schema:
+            print(f"ERROR: Field path '{field_path}' not found in schema template")
```
zhassan-aws: Same here.

zhassan-aws:
Would it make sense to use a standard JSON schema format such as https://json-schema.org/ along with a standard validator, e.g. https://python-jsonschema.readthedocs.io/en/stable/validate/?

yimingyinqwqq:
> Thanks for adding this feature. This has been longstanding!
> 
> I would recommend putting this feature under an unstable flag (see https://model-checking.github.io/kani/rfc/rfcs/0006-unstable-api.html) for some time until we ensure the json schema won't change.

Thank you for reviewing this and for the suggestion.
We will move this feature under an unstable flag and update the PR.

feliperodri:
@yimingyinqwqq could you solve the CI failures before we give a final review to merge this PR?

feliperodri:
### Two comments I'm deliberately not acting on in this PR

Both are legitimate, but both are decisions for RFC #4727 rather than fixes:

**Type validation / adopting a real JSON Schema.** Correct that the current validator does no scalar
type checking, so a numeric count could be a string and validation would still pass. But "ideally
with JSON Schema" is precisely the open question in RFC 0016: shipping a schema document implies a
`schemars` dependency that isn't in the workspace today, which is a real dependency decision rather
than something to add quietly. Anything homegrown I write here gets thrown away when that's settled,
so I'd rather leave the gap documented than build a second half-validator. Happy to do it either way
if reviewers prefer type checks in the interim.

**Exporting coverage results.** Also correct that the export only records whether coverage was
enabled and serializes none of the per-harness `coverage_results`. RFC 0016 lists "coverage results —
include here, or leave with `kani-cov`?" as an open question and defers aggregate coverage to
`kani-cov` and RFC 0011. Settling it inside an implementation PR would pre-empt the RFC, so the fix
here is to stop claiming it: the `Resolves #2636` line is going away with the PR description rewrite,
along with the other issue-closing claims that don't hold — #1219 (compiletest `--help` panic) and
#3357 (non-terse parallel output) aren't touched by this change either, and the non-terse parallel
restriction is still in place. The description will say `Tracks #942` and nothing more.

### Three comments that don't hold

Flagging these so they don't get re-raised:

- `JsonHandler::export` using `?` on `serde_json::to_string_pretty` compiles fine — serde_json
  provides `impl From<serde_json::Error> for std::io::Error`. The suggested
  `io::Error::new(io::ErrorKind::Other, e)` would also be a step back, since `ErrorKind::Other` is
  discouraged for this and `io::Error::other` exists.
- `create_harness_metadata_json` and `create_verification_result_json` do not move owned `String`s
  out of a borrow: `json!` serializes through `to_value(&expr)`. Adding `&` is a no-op.
- The `basic-export` cleanup trap was already added in 5d23989c, which the suggestion notes as
  outdated.

The branch builds clean and passes `cargo clippy --workspace --tests -- -D warnings`, both clippy
passes, `cargo test -p kani-driver`, and the `json-handler` (4), `ui` (144) and `coverage` (20)
suites locally.

feliperodri (approved):

PR #4496, opened 2025-12-10
docs: Correct `default-unwind` Cargo.toml examples

### LINKED ISSUE #4495: Docs are misleading about default-unwind configuration
I tried this code:

```toml
[package.metadata.kani.flags]
default-unwind = 1
```

```
cargo kani
```

with Kani version: 0.66.0

I expected to see this happen:

Kani runs with the default unwind set to 1.

Instead, this happened:

I get the message:
`error: Unknown key type default-unwind`

And nothing else happens. This message comes from here:

https://github.com/model-checking/kani/blob/main/kani-driver/src/args_toml.rs#L151-L183

I found that this works correctly if I instead do:

```toml
[package.metadata.kani.flags]
default-unwind = "1"
```

I expected it to work because of the workspace example given here:
https://model-checking.github.io/kani/tutorial-loop-unwinding.html

...and the package example given here:
https://model-checking.github.io/kani/usage.html#configuration-in-cargotoml

...which both use `1` instead of `"1"`.

hashcatHitman:
The examples for configuring `default-unwind` inside Cargo.toml files use unquoted integers. This isn't how they are actually expected to be written, and following these examples leads to kani failing to run.

This corrects the examples in the documentation to align with the expected configuration format.

Resolves #4495

### DIFF
--- a/docs/src/tutorial-loop-unwinding.md
+++ b/docs/src/tutorial-loop-unwinding.md
@@ -112,7 +112,7 @@ You can do this by putting this into your `Cargo.toml` file:
 
 ```toml
 [workspace.metadata.kani.flags]
-default-unwind = 1
+default-unwind = "1"
 ```
 
 ## Bounded proof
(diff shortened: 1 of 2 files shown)


zhassan-aws:
Thanks!

PR #4528, opened 2026-01-26
Add progress indicator and log file output for concise terminal output

tautschnig:
When in an interactive terminal, `kani main.rs --log-file log.txt` will now print:
```
$ kani main.rs --log-file log.txt
Kani Rust Verifier 0.66.0 (standalone)
  [00:00:00] [########################################] 2/2 (0s)
Verification complete | Succeeded: 2 | Failed: 0 | Timed out: 0                                                             Manual Harness Summary:
```
where `########################################` will be updated up until the run is complete. This enables users to more quickly grasp how much work Kani has already completed, which may be particularly useful when running thousands of harnesses with `autoharness`.

Co-authored-by: Kani autonomous agent


feliperodri:
Two observations, neither blocking:

  1. The log file was appended across runs; I made it truncate. Two runs of the same command left 32 then 64 lines with no separator, so someone grepping for `VERIFICATION:- FAILED` could match a stale line from an earlier run and read it as the current result. I went with truncate-on-first-write because `--output-into-files` already writes its per-harness files with `File::create`, so Kani had established that an output-file flag holds the current run. Happy to switch to append-with-a-run-header if you'd rather; it's a one-line change to the open mode, and the behavior is pinned by tests either way.

  2. The test asserted only on stdout, so it would have passed with an empty log file. It also wrote verification.log into the repository root, untracked and not gitignored, so running the suite dirtied the working tree. I replaced it with a script-based-pre test that puts the log in a `mktemp -d`, greps it for the per-harness sections and the verification result, then runs a second time and asserts the harness count didn't double. I checked it fails when reverted to append mode (the re-run count goes 2 → 4).

Two things still open that I deliberately left alone:

  - Multi-line harness output only gets Thread N:  on its first line, so continuation lines are unlabeled in the log. This is pre-existing in main's stdout path, not something this PR introduces, but --log-file makes it more visible since the log is the thing you read afterwards.
  - The progress bar itself isn't covered by any test, since it needs a TTY. Neither the old test nor the new one exercises it.

feliperodri (approved):

PR #4535, opened 2026-02-10
env: enable THP via env

nmanthey:
Solvers and reasoners benefit from better memory management, as many of the memory accesses are spread across a huge memory set, and hard to predict. Using huge pages typically improves the overall performance. Therefore, this change enables Kani to make use of huge pages to reduce the memory management overhead.

At the moment, this change is only effective is the used runtime environment supports it. If the system uses glibc 2.35, the support for the environment variable should be available. Other older versions might have that feature backported.

For other systems, huge pages might be available via system settings. To be able to change these settings, you might need higher privileges.

### Testing Done

We have seen 10% runtime improvement in parallel verification mode, while using a bit more memory.

I tested this change by running some verification on the sudo-rs package and measured runtime.

The summary of that experiment is:

| Metric | With THP | Baseline (no THP) | Difference |
|--------|----------|-------------------|------------|
| Wall clock time | 5:08.42 | 5:45.77 | -37s (10.8% faster) |
| User time | 2422.80s | 2484.30s | -61.5s |
| System time | 86.30s | 148.73s | -62.4s (42% less) |
| CPU utilization | 813% | 761% | +52% |
| Max RSS | 23.8 GB | 22.1 GB | +1.7 GB |
| Minor page faults | 10.5M | 80.4M | -70M (87% fewer) |

The THP modification shows clear benefits:
- **10.8% faster** overall
- **42% less system time** (kernel overhead)
- **87% fewer minor page faults** - this is the key improvement from THP

Command used
```
/usr/bin/time -v cargo kani autoharness -Z autoharness -Z unstable-options --verbose --jobs=32 --output-format=terse --harness-timeout=60
```

### DIFF
--- a/kani-driver/src/main.rs
+++ b/kani-driver/src/main.rs
@@ -46,9 +46,9 @@ mod version;
 fn main() -> ExitCode {
     // Set GLIBC_TUNABLES for child processes (CBMC, etc.) if not already set
     // This enables THP for subprocesses even if the parent didn't have it set at startup
-    if env::var("GLIBC_TUNABLES").is_err() {
+    if std::env::var("GLIBC_TUNABLES").is_err() {
         // SAFETY: Called at start of main before any threads are spawned
-        unsafe { env::set_var("GLIBC_TUNABLES", "glibc.malloc.hugetlb=1") };
+        unsafe { std::env::set_var("GLIBC_TUNABLES", "glibc.malloc.hugetlb=1") };
     }
 
     let invocation_type = determine_invocation_type(Vec::from_iter(std::env::args_os()));


kani-driver/src/main.rs
```diff
@@ -44,6 +44,13 @@ mod version;
 /// The driver can be invoked via `cargo kani` and `kani` commands, which determines what kind of
 /// project should be verified.
 fn main() -> ExitCode {
+    // Set GLIBC_TUNABLES for child processes (CBMC, etc.) if not already set
+    // This enables THP for subprocesses even if the parent didn't have it set at startup
+    if std::env::var("GLIBC_TUNABLES").is_err() {
+        // SAFETY: Called at start of main before any threads are spawned
+        unsafe { std::env::set_var("GLIBC_TUNABLES", "glibc.malloc.hugetlb=1") };
```
tautschnig:
```suggestion
    if env::var("GLIBC_TUNABLES").is_err() {
        // SAFETY: Called at start of main before any threads are spawned
        unsafe { env::set_var("GLIBC_TUNABLES", "glibc.malloc.hugetlb=1") };
```
for consistency.

tautschnig (approved):

PR #4571, opened 2026-04-05
Detect `stub_verified`/`Arbitrary` recursion at compile time

feliperodri:
## Problem

When a type's `kani::Arbitrary` implementation reaches a function targeted by `#[kani::stub_verified]`, verification never terminates — CBMC unwinds until it exhausts memory, after several minutes, with an out-of-memory message that never names the cause.

The cause is the contract replacement's **own return-value havoc**. `initial_replace_stmts` emits `kani::any_modifies::<Ret>()`, which `AnyModifiesPass` rewrites to `kani::any::<Ret>()`. So if `Ret`'s `Arbitrary` impl reaches the stubbed function, the replacement re-enters itself:

```
Wrapper::normalize → kani_register_contract → normalize::{closure#1}   (REPLACE)
  → kani::any::<Wrapper> → <Wrapper as Arbitrary>::any → Wrapper::new
  → Wrapper::normalize                                                  ← closes
```

The cycle is **self-contained inside the replacement**. It reproduces even when the harness passes a concrete value and never calls `kani::any()` itself — so a caller-side guard cannot break it. This recursion has no fixpoint, so no unwind bound makes it converge.

## Solution

Detect the cycle at compile time rather than letting verification hang. For each `ContractMode::Replace` target, `FunctionWithContractPass` resolves `kani::any::<Ret>` and walks the monomorphized call graph for a path back to the target. If one exists, it emits an error naming the path:

```
error: `Wrapper::normalize` is used as a verified stub, but generating an arbitrary
       value of its return type `Wrapper` calls `Wrapper::normalize` again
  = note: the contract replacement havocs its return value with `kani::any::<Wrapper>()`,
          so this forms an unbounded recursion:
              -> <Wrapper as kani::Arbitrary>::any
              -> Wrapper::new
              -> Wrapper::normalize
  = help: derive `Arbitrary` for `Wrapper` instead of implementing it manually, or
          avoid calling `Wrapper::normalize` from the `Arbitrary` implementation
```

This follows the existing `check_mutual_recursion` precedent for the analogous unsupported-recursion case.

### Instance-precise, not `DefId`-based

Both the per-target dedup and the call-graph comparison are keyed on the monomorphized `Instance`, not the `DefId`. A generic `stub_verified` target can be instantiated at several return types, each with its own potential cycle, and a return type's `Arbitrary` impl may call a *different* monomorphization of the same generic function without re-entering the instance under check. Comparing by `DefId` alone would both:

- skip a later cyclic instantiation when an earlier acyclic one was checked first (missed detection → the hang returns), and
- flag a call to a sibling instantiation whose chain terminates (false positive → a working proof is rejected).

### Why not the runtime guard

An earlier revision of this PR tracked `kani::any()` nesting depth (`ARBITRARY_NESTING_DEPTH`) and ran the original body when depth > 0. **That approach is not viable and has been removed:**

- **It guards the wrong thing.** The replace closure itself calls `kani::any()`, so the guard fires *inside* the stub. `stub_verified` would silently drop its abstraction for calls having nothing to do with `Arbitrary`.
- **The soundness argument was inverted.** It keys on dynamic nesting depth, not "is this an `Arbitrary` impl." Any contract-replaced call under a `kani::any()` frame loses its stub; whether that's sound depends on the caller, so it is not the blanket win the earlier description claimed.
- **It conflicts with CBMC's DFCC assigns checking.** Writes to global mutable state inside a contract-checked scope need to be in every `modifies` clause. `REENTRY` survives only because `--nondet-static-exclude` suppresses *havocking* (not assigns-tracking) and it is written outside the `--enforce-contract` frame; a counter inside `kani::any()` is inside that frame.

### Limitation

Detection follows only statically resolvable calls, so a cycle routed through a function pointer or trait object is not reported. This is deliberate: a missed detection reproduces prior behavior, whereas a false positive would reject a working proof. Documented in the code and RFC.

This PR makes the failure diagnosable; it does not make the pattern verify. Doing that would require the return havoc to stop routing through user `Arbitrary`, which changes the soundness model of every `stub_verified` return value and warrants separate design discussion.

## Testing

- `tests/expected/function-contract/stub_verified_arbitrary_cycle.{rs,expected}` — regression test for the new diagnostic.
- `tests/expected/function-contract/stub_verified_arbitrary_cycle_generic.{rs,expected}` — a generic target where only the second instantiation is cyclic; guards against `DefId`-keyed dedup skipping it.
- `tests/kani/FunctionContracts/stub_verified_arbitrary_other_instantiation.rs` — a call to a *sibling* instantiation whose chain terminates; guards against a `DefId`-based false positive. Must verify.
- `tests/kani/FunctionContracts/stub_verified_safe_arbitrary.rs` — derived `Arbitrary` (no cycle) still verifies with `stub_verified`.

Full runs: 592/592 `kani`, `expected` and `ui` suites green (the only failure, `expected/shadow/unsupported_num_objects`, is pre-existing and unrelated — it fails identically on a clean tree). Both CI clippy invocations and `kani-fmt --check` are clean.

---

### DIFF
--- a/docs/src/reference/experimental/contracts.md
+++ b/docs/src/reference/experimental/contracts.md
@@ -62,5 +62,13 @@ fn check_foo() {
 ```
 By leveraging the stubbing feature, we can replace the (expensive) `gcd` call with a *verified abstraction* of its behavior, greatly reducing verification time for `foo`.
 
+> **Note:** A verified stub havocs its return value with `kani::any()`. If the
+> `kani::Arbitrary` implementation for that return type calls the stubbed
+> function, the stub re-enters itself and the recursion never terminates. Kani
+> reports this as a compile-time error naming the call path. To fix it, derive
+> `Arbitrary` (which generates field-by-field values without calling user
+> functions), or avoid calling the stubbed function from the `Arbitrary`
+> implementation.
+
 There is far more to learn about contracts.
 We highly recommend reading our [blog post about contracts](https://model-checking.github.io/kani-verifier-blog/2024/01/29/function-contracts.html) (from which this `gcd` example is taken). We also recommend looking at the `contracts` module in our [documentation](../../crates/index.md).
--- a/kani-compiler/src/kani_middle/transform/contracts.rs
+++ b/kani-compiler/src/kani_middle/transform/contracts.rs
@@ -16,7 +16,9 @@ use rustc_public::mir::{
     Body, ConstOperand, Operand, Rvalue, Terminator, TerminatorKind, VarDebugInfoContents,
 };
 use rustc_public::rustc_internal;
-use rustc_public::ty::{ClosureDef, FnDef, MirConst, RigidTy, TyKind, TypeAndMut, UintTy};
+use rustc_public::ty::{
+    ClosureDef, FnDef, GenericArgKind, GenericArgs, MirConst, RigidTy, TyKind, TypeAndMut, UintTy,
+};
 use rustc_span::Symbol;
 use std::collections::HashSet;
 use std::fmt::Debug;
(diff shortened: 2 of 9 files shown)


Ari4ka (approved):

feliperodri:
I attempted to implement a runtime fix (nesting depth counter) but it conflicts with CBMC's DFCC assigns checking: writes to global mutable state inside contract-checked scopes trigger assigns violations. CBMC provides no mechanism to exempt infrastructure writes from DFCC tracking. So back to the drawing board...

Ari4ka (approved):

tautschnig (approved):

PR #4573, opened 2026-04-08
fix saturating float-to-int cast

Rexicon226:
The issue is that Kani computes int_max_as_float by casting u64::MAX to f64. But u64::MAX (2^64 - 1) isn't exactly representable in f64, so it rounds *up* to 2^64. The old code used a greater-than operation, which would mean a float value equal to exactly 2^64 fails this check, falls through the raw truncation path, and produces a wrong result instead of saturating to u64::MAX.

Instead we use greater-than-or-equal so that the boundary value 2^64 correctly saturates. This is safe for types where the max IS exactly representable.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
@@ -1198,7 +1198,7 @@ impl GotocCtx<'_, '_> {
         // Note: We convert the integer bounds to float for comparison.
         // For very large integer types (i128, u128), the float conversion may lose precision,
         // but this is acceptable because:
-        // 1. Any float value that's truly larger than MAX_INT will still compare as > MAX_INT
+        // 1. Any float value that's truly larger than MAX_INT will still compare as >= MAX_INT
         // 2. Any float value that's truly smaller than MIN_INT will still compare as < MIN_INT
         let int_min_as_float = self.int_bounds_as_float(dst_ty, false, src_ty);
         let int_max_as_float = self.int_bounds_as_float(dst_ty, true, src_ty);
@@ -1208,11 +1208,11 @@ impl GotocCtx<'_, '_> {
 
         // Check for special cases:
         // 1. isnan(src) -> result is 0
-        // 2. src > int_max -> result is MAX
+        // 2. src >= int_max -> result is MAX
         // 3. src < int_min -> result is MIN (or 0 for unsigned)
         // 4. Otherwise -> truncate toward zero
         let is_nan = src_expr.clone().is_nan();
-        let above_max = src_expr.clone().gt(int_max_as_float.clone());
+        let above_max = src_expr.clone().ge(int_max_as_float.clone());
         let below_min = src_expr.clone().lt(int_min_as_float);
 
         // The truncated value (normal cast behavior for values in range)
--- a/tests/kani/Cast/float_to_int_saturating.rs
+++ b/tests/kani/Cast/float_to_int_saturating.rs
@@ -1,6 +1,8 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+#![feature(f128)]
+
 //! Regression test for float-to-int saturating cast (GitHub issue #4536).
 //! Since Rust 1.45, `as` performs saturating casts from float to int.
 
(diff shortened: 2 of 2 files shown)


kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
```diff
@@ -1212,7 +1212,7 @@ impl GotocCtx<'_, '_> {
         // 3. src < int_min -> result is MIN (or 0 for unsigned)
         // 4. Otherwise -> truncate toward zero
         let is_nan = src_expr.clone().is_nan();
-        let above_max = src_expr.clone().gt(int_max_as_float.clone());
+        let above_max = src_expr.clone().ge(int_max_as_float.clone());
```
feliperodri, line 1215: Thank you for this fix! There is also another potential gap, `u128` and `i128` have the same problem.

The `int_bounds_as_float` function converts `u128::MAX` and `i128::MAX` to `f64` the same way:

```rust name=rvalue.rs url=https://github.com/model-checking/kani/blob/e43314b17ee625db430f7102f989e4b2ab02f6e5/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs#L1278-L1279
UintTy::U128 => u128::MAX as f64,
```

`u128::MAX` similarly rounds up to a float that is not `u128::MAX`, causing the same kind of off-by-one at the boundary. The fix in this PR (`ge` for max) should cover this case too — but it's worth adding a test for `f64 as u128` at the boundary value to confirm.

Could you also add tests for `u128` and `i64`/`i128` boundaries?

Consider adding:
```rust
#[kani::proof]
fn check_f64_to_u128_saturation() {
    // u128::MAX rounds up in f64, similar to u64::MAX
    let x: f64 = u128::MAX as f64; // = 2^128 in f64
    let y: u128 = x as u128;
    kani::assert(y == u128::MAX, "matches");
}
```

One last thing. The comment at line 1200–1202 is partially misleading

```rust name=rvalue.rs url=https://github.com/model-checking/kani/blob/e43314b17ee625db430f7102f989e4b2ab02f6e5/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs#L1198-L1204
// Note: We convert the integer bounds to float for comparison.
// For very large integer types (i128, u128), the float conversion may lose precision,
// but this is acceptable because:
// 1. Any float value that's truly larger than MAX_INT will still compare as > MAX_INT
// 2. Any float value that's truly smaller than MIN_INT will still compare as < MIN_INT
```

Point 1 is now slightly inaccurate, since it's precisely the case where a float equals `MAX_INT_as_float` (not *strictly* greater) that this PR fixes. The comment should be updated to reflect that `>=` is used intentionally to handle this rounding.

feliperodri (changes requested):

kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
Rexicon226: Good points, will address this!

Rexicon226: How does this look? I'm happy to refactor the comments more, just tried to be minimal for now.

tests/kani/Cast/float_to_int_saturating.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+#![feature(f128)]
```
feliperodri, line 4: why do you need this?

feliperodri (approved):

tests/kani/Cast/float_to_int_saturating.rs
Rexicon226: It gave me a warning that `f128` is an unstable :thinking: 
There are other tests that also add the feature gate.

feliperodri: Could you address all CI failures?

Rexicon226: I believe I've fixed it :)

Rexicon226:
The remaining failures don't seem related to this PR?

Rexicon226:
@feliperodri sorry to bother, but any progress?

Issue #3316, opened 2024-07-02
Function Contracts: Mutual recursion function contract wrapper for replace code stub


pi314mm:
Function contracts are not being properly expanded when the recursive call to the function occurs outside of the function body

Consider the following code:

```rust
#[kani::recursion]
#[kani::ensures(|r| if n==0 {*r == true} else {true})]
fn even(n: u32) -> bool {
    if n == 0 {
        true
    } else {
        odd(n - 1)
    }
}

#[kani::recursion]
fn odd(n: u32) -> bool {
    if n == 0 {
        false
    } else {
        even(n - 1)
    }
}

#[kani::proof_for_contract(even)]
fn harness(){
    let x = kani::any();
    even(x);
}
```

Within the verification of `even`, we would expect it to call `odd` which refers to the code stub recursively marked version of `even` because you are re-entering the `even` function during the verification of it.

Instead, we get the following expansion:
```rust
#[kanitool::recursion]
#[kanitool::checked_with = "even_recursion_wrapper_882c72"]
#[kanitool::replaced_with = "even_replace_882c72"]
#[kanitool::inner_check = "even_wrapper_882c72"]
fn even(n: u32) -> bool { { if n == 0 { true } else { odd(n - 1) } } }
#[allow(dead_code, unused_variables, unused_mut)]
#[kanitool::is_contract_generated(recursion_wrapper)]
fn even_recursion_wrapper_882c72(arg0: u32) -> bool {
    static mut REENTRY: bool = false;
    if unsafe { REENTRY } {
            even_replace_882c72(arg0)
        } else {
           unsafe { REENTRY = true };
           let result_kani_internal = even_check_882c72(arg0);
           unsafe { REENTRY = false };
           result_kani_internal
       }
}
#[allow(dead_code, unused_variables, unused_mut)]
#[kanitool::recursion]
#[kanitool::is_contract_generated(check)]
fn even_check_882c72(n: u32) -> bool {
    let result_kani_internal: bool = even_wrapper_882c72(n);
    kani::assert(kani::internal::apply_closure(|r|
                if n == 0 { *r == true } else { true },
            &result_kani_internal),
        "|r| if n==0 { *r == true } else { true }");
    result_kani_internal
}
#[allow(dead_code, unused_variables, unused_mut)]
#[kanitool::recursion]
#[kanitool::is_contract_generated(replace)]
fn even_replace_882c72(n: u32) -> bool {
    let result_kani_internal: bool = kani::any_modifies();
[...]

Issue #3273, opened 2024-06-18
Function Contracts: Better error messages


pi314mm:
`tests/expected/function-contract/history/ui/` currently contains several failing test cases revolving around the `old` history expressions. These error messages need to be enhanced to provide clarity to the user.

Additionally, the macro rule of `assert_spanned_err` needs to be adapted to have better error messages upon usage and proper documentation.

PR #4580, opened 2026-04-19
Reject mutual recursion with `#[kani::recursion]`

feliperodri:
Emit a compilation error when a function with `#[kani::recursion]` is involved in mutual recursion. The per-function REENTRY mechanism only handles direct recursion soundly since mutual recursion silently produces unsound verification results.

Contributes to #3316, #3273.

## Problem

The `#[kani::recursion]` attribute uses a per-function REENTRY flag to detect when a function calls itself, replacing the recursive call with the function's contract. For mutual recursion (`f` calls `g`, `g` calls `f`), the REENTRY flag for `g` is never set when verifying `f`'s contract, so `g`'s body executes fully instead of being replaced by its contract. Verification succeeds but the result is meaningless — a silent soundness gap.

## Solution

`check_mutual_recursion()` in the contract transform pass scans the MIR body of functions in `RecursiveCheck` mode for calls to other functions that also have `#[kani::recursion]` and a contract. For each such callee, it checks if the callee's body calls back to the original function. If so, it emits a `span_err` (not a warning) because the result would be unsound.

This is an error rather than a warning because soundness violations should not be silently ignorable.

## Limitations

- Only detects one level of indirection (f→g→f), not deeper chains (f→g→h→f).
- Reports only the first mutual-recursive callee per function.

## Example error

```
error: `#[kani::recursion]` is used on `mutual_a`, which calls `mutual_b`
       that calls back to `mutual_a`. Mutual recursion is not supported by
       contract verification and may produce unsound results. Only direct
       recursion (a function calling itself) is handled correctly.
  --> tests/expected/function-contract/mutual_recursion_unsound.rs:12:30
```

## Testing

- `mutual_recursion_unsound.rs` — two mutually recursive functions (`mutual_a` ↔ `mutual_b`) with contracts and `#[kani::recursion]`, verifying the error is emitted.

---

### DIFF
--- a/kani-compiler/src/kani_middle/transform/contracts.rs
+++ b/kani-compiler/src/kani_middle/transform/contracts.rs
@@ -302,6 +302,9 @@ impl TransformPass for FunctionWithContractPass {
         match instance.ty().kind().rigid().unwrap() {
             RigidTy::FnDef(def, args) => {
                 if let Some(mode) = self.contract_mode(tcx, *def) {
+                    if mode == ContractMode::RecursiveCheck {
+                        check_mutual_recursion(tcx, *def, &body);
+                    }
                     self.mark_unused(tcx, *def, &body, mode);
                     let new_body = self.set_mode(tcx, body, mode);
                     (true, new_body)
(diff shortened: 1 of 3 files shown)


rajathmCMU:
Should this be a warn or an error?

rajath-mk (approved):

PR #4583, opened 2026-04-22
Support pointer arithmetic in quantifier predicates

feliperodri:
Lower wrapping pointer arithmetic intrinsics (`wrapping_add`, `wrapping_byte_offset`) directly to CBMC pointer `Plus` expressions in the pure expression inliner. These intrinsics have no GOTO body to inline, so they need special handling.

## Problem

Quantifier predicates that dereference pointers with offsets — e.g., `*ptr.wrapping_byte_offset(i as isize)` — failed because the `inline_as_pure_expr` inliner couldn't resolve pointer arithmetic intrinsics. These functions have no body in the symbol table (they're compiler intrinsics), so the inliner returned the original expression unchanged, leaving unresolved function calls that CBMC rejected.

## Solution

In `inline_call_as_pure_expr` (`goto_ctx.rs`), before attempting to look up the function body, check if the function name matches a known wrapping pointer arithmetic intrinsic (`wrapping_add`, `wrapping_byte_offset`). If so, directly emit `ptr.plus(offset)` — the CBMC expression for pointer arithmetic. A type guard (`arguments[0].typ().is_pointer()`) prevents misapplication to non-pointer functions.

Non-wrapping variants (`offset`, `add`, `arith_offset`) are intentionally excluded because they trigger CBMC bounds checks inside quantifier bodies, which fail in the symbolic evaluation context.

## Example

```rust
kani::forall!(|i in (0, len)| unsafe { *ptr.wrapping_byte_offset(i as isize) == 0 })
```

## Changes

- `goto_ctx.rs`: Recognize wrapping pointer arithmetic intrinsics and lower to `ptr.plus(offset)`
- `rfc/src/rfcs/0010-quantifiers.md`: Document pointer arithmetic intrinsic lowering in the Detailed Design section
- `tests/kani/Quantifiers/pointer_arithmetic.rs`: 4 harnesses covering `wrapping_byte_offset` (forall + exists) and `wrapping_add` (u32 + u8)

---


kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
```diff
@@ -512,6 +512,23 @@ impl GotocCtx<'_, '_> {
...
+        // (e.g., matching on DefId or a dedicated KaniIntrinsic enum) for
+        // robustness against mangling variations.
+        let fn_name = fn_id.to_string();
+        let is_ptr_fn = fn_name.contains("::ptr::") || fn_name.contains("const_ptr");
+        let is_ptr_arith = is_ptr_fn
+            && (fn_name.contains("wrapping_add") || fn_name.contains("wrapping_byte_offset"));
+        if is_ptr_arith && arguments.len() >= 2 && arguments[0].typ().is_pointer() {
+            let ptr = self.inline_as_pure_expr(&arguments[0], visited);
+            let offset = self.inline_as_pure_expr(&arguments[1], visited);
+            return ptr.plus(offset);
```
adpaco: IIUC the `Plus` in CBMC scales by the pointee size, so this is correct for `wrapping_add`. But is it correct for `wrapping_byte_offset`? Maybe the pointer and offset have already been lowered to bytes?

rfc/src/rfcs/0010-quantifiers.md
```diff
@@ -234,6 +234,15 @@ To solve this, Kani generates **pure expression trees** for quantifier bodies:
    are inlined by looking up the function body in the symbol table, resolving its
    return expression, and substituting parameters — producing a pure expression.
 
+5. **Pointer arithmetic intrinsic lowering**: Wrapping pointer arithmetic functions
+   (`wrapping_byte_offset`, `wrapping_add`) are compiler intrinsics with no GOTO body
```
adpaco: What about `wrapping_sub`? I guess it fails because it's using the original inliner?

adpaco (approved):

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
feliperodri: You were right. CBMC's pointer `Plus` scales the offset by the pointee size, so the original code incorrectly scaled the byte count by `sizeof::<T>()` when intercepting `wrapping_byte_offset`. The existing test only happened to pass because it used `[u8; 8]` (pointee size = 1). Rust's `wrapping_byte_offset` is implemented as `self.cast::<u8>().wrapping_offset(count).with_metadata_of(self)`, i.e., it casts to `*u8` first so the subsequent offset is in bytes. Fixed by mirroring that in the lowering: byte-offset variants now cast the pointer to `*u8` before `Plus` so the scaling is by 1 byte. Added `check_wrapping_byte_offset_u32` to catch this regression, it exercises a non-`u8` pointee and would fail with the old code.

rfc/src/rfcs/0010-quantifiers.md
feliperodri: Yes. `wrapping_sub` was falling through to the generic inliner, which can't process intrinsics without a GOTO body. Added it alongside `wrapping_byte_sub` and `wrapping_offset`. Sub variants just negate the offset before applying `Plus`. Full list now covered:
 - Element-offset: `wrapping_add`, `wrapping_sub`, `wrapping_offset`;
 - Byte-offset: `wrapping_byte_offset`, `wrapping_byte_add`, `wrapping_byte_sub`;

Added `check_wrapping_sub_forall` and `check_wrapping_byte_sub_forall` tests, and updated the RFC to document all supported variants and the scaling rules.

kani-compiler/src/codegen_cprover_gotoc/context/goto_ctx.rs
```diff
@@ -515,6 +515,52 @@ impl GotocCtx<'_, '_> {
...
+            && arguments[0].typ().is_pointer()
+        {
+            let ptr = self.inline_as_pure_expr(&arguments[0], visited);
+            let offset = self.inline_as_pure_expr(&arguments[1], visited);
+            let offset = if is_sub { offset.neg() } else { offset };
+            // For byte-offset variants, cast the pointer to `*u8` so CBMC's Plus
+            // scales the offset by 1 byte (matching the Rust semantics that
+            // first casts `self.cast::<u8>()`).
+            let base = if is_byte { ptr.cast_to(Type::unsigned_int(8).to_pointer()) } else { ptr };
+            return base.plus(offset);
```
adpaco: You need to cast back to the original pointer type after the offset:
```suggestion
            let base = if is_byte { ptr.cast_to(Type::unsigned_int(8).to_pointer()) } else { ptr };
            let result = base.plus(offset);
            return if is_byte { result.cast_to(arguments[0].typ().clone()) } else { result };
```

It seems to be working now for some reason (maybe CBMC is doing it automatically?), but it might produce the wrong results in other cases.

rajath-mk (approved):

PR #4587, opened 2026-04-22
Support stubbing trait method implementations

### LINKED ISSUE #1997: Resolve trait methods during stubbing

feliperodri:
Enable stubbing of trait method implementations using fully-qualified syntax (`<Type as Trait>::method`), including generic traits. Previously, Kani's stub resolution only searched inherent `impl` blocks and trait method implementations were silently ignored.

Resolves #1997.

## Problem

Users could not stub trait method implementations. Writing `#[kani::stub(<MyType as MyTrait>::method, mock_method)]` failed because the resolution algorithm did not search through trait implementations for the concrete type. This was a significant limitation since many Rust APIs are trait-based.

### Resolution (`resolve.rs`)

`resolve_in_trait_impl` now correctly resolves trait method implementations by:
1. Extracting generic arguments from the trait segment (e.g., `<u32>` from `Convert<u32>`)
2. Building the full `GenericArgs` with the Self type + trait generic parameters
3. Resolving the `Instance` for the concrete type's implementation

This handles both simple traits (`<Vec<u8> as MyTrait>::method`) and generic traits (`<MyType as Convert<u32>>::convert`).

### Documentation (`stubbing.md`)

Updated the user-facing docs to document trait method stubbing with examples covering:
- Basic trait method stubbing syntax
- Generic trait methods
- Supported patterns (dyn dispatch, supertraits, overridden defaults)
- Known limitation: non-overridden default methods

## Testing

**5 passing tests:**
- `stub_trait_method.rs` — basic `<Type as Trait>::method` stubbing
- `stub_generic_trait_method.rs` — generic trait `<Type as Trait<T>>::method`
- `stub_trait_dyn_dispatch.rs` — dynamic dispatch through `&dyn Trait` and `Box<dyn Trait>`
- `stub_trait_supertrait.rs` — methods defined in supertraits
- `stub_trait_default_overridden.rs` — default trait methods that are overridden in the impl

**1 fixme test:**
- `fixme_stub_trait_default_method.rs` — non-overridden default methods cause a type mismatch during stub validation (the default body uses `Self` as a placeholder type)

---


kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -411,16 +437,49 @@ fn resolve_in_any_trait<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty, name: &str) -> Option<F
...
+                    return Err(ResolveError::UnsupportedPath {
+                        kind: "const generic arguments in trait stubs",
+                    });
+                }
+                syn::GenericArgument::Lifetime(_) => {
+                    // Lifetimes are erased — safe to skip.
+                }
+                _ => {
+                    return Err(ResolveError::UnsupportedPath {
+                        kind: "associated type/const constraints in trait stubs",
```
adpaco, line 476: I don't think we're covering the `Const` and `_`, so it'd be nice to add a test for them.

kani-compiler/src/kani_middle/resolve.rs
```diff
@@ -411,16 +437,49 @@ fn resolve_in_any_trait<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty, name: &str) -> Option<F
...
     // that implements that method for `ty`.
     let trait_fn_fn_def = stable_fn_def(tcx, trait_fn_id).unwrap();
-    let desired_generic_args = GenericArgs(vec![GenericArgKind::Type(ty)]);
+
+    // Build generic args: Self type + any trait generic parameters (e.g., T in Convert<T>).
+    let mut args = vec![GenericArgKind::Type(ty)];
+    if let Some(syn_args) = trait_args {
+        let Some(module) = current_module else {
+            return Err(ResolveError::UnsupportedPath {
+                kind: "generic trait arguments require a known module context",
```
adpaco: This message doesn't mention stubbing so the user might not understand what's happening.

adpaco (approved):
Thank you!

rajath-mk (approved):

Issue #4079, opened 2025-05-16
Kani always rebuilds when running verify-std


dawidl022:
I was following [the instructions to run Kani on the verify-rust-std project](https://model-checking.github.io/verify-rust-std/tools/kani.html#using-kani-to-verify-the-rust-standard-library), and everytime I execute the command the entire project gets rebuilt, even when there are no changes to the code. Ideally, if no code changes were made, only the model checker should run.

Seems to be the same behaviour as described in #895, but that should have been fixed by #2232, so I'm not sure if this is specific to just verify-std or a more general issue.

zhassan-aws:
Thanks for creating this issue @dawidl022. Indeed, this seems to be specific to verify-rust-std. On a normal package, compilation is skipped if the code hasn't changed:
```bash
$ cargo clean
     Removed 31 files, 2.7MiB total
$ cargo kani --verbose |& grep kani-compiler
[...]

PR #4600, opened 2026-05-20
Add 'kani verify-artifacts' subcommand

lovesegfault:
`cargo kani` and `kani` both insist on owning the compile phase. There is no way to verify artifacts from an earlier `--only-codegen` run, a build system that drives `kani-compiler` directly, or a CI job that splits compilation and verification across machines: every invocation re-runs cargo or rustc, and a build system with its own per-crate cache ends up re-implementing kani-driver's verify half — the goto-cc/goto-instrument/cbmc pipeline, harness scheduling, verdict rendering — to avoid throwing the cache away. #4079 and #4212 are the same friction from the build side.

Adds `kani verify-artifacts <DIR>...`, a standalone subcommand that scans the given directories for `*.kani-metadata.json`, resolves each harness's `goto_file` by basename against the directory the metadata was found in (the recorded path is the producing machine's `OUT_DIR`, wrong after relocation), and runs `verify_project()` over the result. Mirrors `verify-std`: same `Project` construction shape, gated behind `-Z unstable-options`. The `script-based-pre` test round-trips `cargo kani --only-codegen` → `kani verify-artifacts` against a fixture with one passing and one deliberately-failing harness, so a non-zero exit proves the verdict actually reached cbmc rather than silently no-op'd.

Two known limitations, documented in the args help: artifacts must come from the same kani version that verifies them (kani-driver does not version-check the metadata schema; happy to add a stamp as a follow-up if there's appetite), and the artifact directory must be writable since `Project::try_new()` writes the linked `.out` next to each `.symtab.out`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,7 @@ exclude = [
   "tests/script-based-pre/kani_lib_dep",
   "tests/script-based-pre/no_codegen",
   "tests/script-based-pre/no_codegen_error",
+  "tests/script-based-pre/verify_artifacts_cmd",
 ]
 
 [workspace.lints.clippy]
--- a/kani-driver/src/args/mod.rs
+++ b/kani-driver/src/args/mod.rs
@@ -8,6 +8,7 @@ pub mod common;
 pub mod list_args;
 pub mod playback_args;
 pub mod std_args;
+pub mod verify_artifacts_args;
 
 use self::common::*;
 use crate::args::cargo::CargoTargetArgs;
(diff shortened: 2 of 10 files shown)


feliperodri:
@lovesegfault Thank you for your contribution! The thing that is missing the most for me is documentation. Usually for new features, we have an RFC. However, we could have at least a proper documentation for users on how to use this, short-comes, why is valuable... For instance:
1. **CLI/UX clarity**
   - Help text makes unstable status and limitations very explicit.
2. **Failure semantics**
   - Non-zero exit on failing harnesses is stable and tested (sounds covered).
3. **Path resolution robustness**
   - Basename-based `goto_file` resolution: confirm behavior on collisions.
4. **Safety/compat**
   - Error messaging when metadata is mismatched or missing files.
5. **Future-proofing**
   - Optional follow-up issue for metadata schema/version stamp.

Could you add some before I can do a review? Also, @tautschnig what do you think?

feliperodri:
ping @lovesegfault to try to include this in the next release 😉

Issue #692, opened 2021-12-11
Add support to stack unwind panic strategy


celinval:
**Requested feature:** We would like to be able to verify the logic that is triggered while unwinding a panic statement.
**Use case:** Programs sometimes rely on the unwind logic to ensure that there is not resource leak or persistent data inconsistency. Thus, it is important to users to be able to verify that the program will correctly handle its resource if a panic occurs.
**Link to relevant documentation (Rust reference, Nomicon, RFC):** N/A
**Is this a breaking change?**

**Test case:**

```rust
// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0 OR MIT
//
// We currently do not support stack unwinding panic strategy. Once we do, this testcase should
// fail during the verification with both the panic and the assertion failing.
// To run manually, execute:
// ```
// RUSTFLAGS="--C panic=unwind --crate-type lib" rmc unwind_fixme.rs  --function create
// ```
//
// compile-flags: --C panic=unwind --crate-type lib
// rmc-flags: --function create
// rmc-verify-fail

pub struct DummyResource {
    pub data: Option<String>,
}

impl Drop for DummyResource {
    fn drop(&mut self) {
        assert!(self.data.is_some(), "This should fail");
    }
}

#[no_mangle]
pub fn create(empty: bool) -> DummyResource {
    let mut dummy = DummyResource { data: None };
    if empty {
        unimplemented!("This is not supported yet");
    }
    dummy.data = Some(String::from("data"));
    dummy
}

```

With stack unwinding, the assertion inside the DummyResource::drop() should fail. However, with the abort strategy, the same assertion should hold.

tedinski:
* Related: #267

PR #4601, opened 2026-05-20
Set kani-compiler's required rustc flags unconditionally

lovesegfault:
kani-compiler requires a specific set of rustc flags for its MIR analysis and goto-c codegen to be correct: `-Cpanic=abort`, `-Coverflow-checks=on`, `-Csymbol-mangling-version=v0`, `-Zalways-encode-mir`, `-Zpanic_abort_tests=yes`, `-Zmir-enable-passes=-RemoveStorageMarkers`, `--cfg=kani`, `--check-cfg=cfg(kani)`. Today they are derived inside kani-driver (`base_rustc_flags()` / `kani_rustc_flags()`, both private) and threaded through cargo's RUSTFLAGS. Anyone invoking `kani-compiler` outside that path — a build system that compiles with `RUSTC=kani-compiler` directly, or a contributor running `RUSTC=kani-compiler cargo build` to debug a codegen issue — has to re-derive the flag set from kani-driver source, and the failure modes range from a hard error (missing `-Cpanic=abort`, see #692) to a silently-wrong result. Missing `--cfg=kani` is the worst case: any `#[cfg(kani)]`-gated harness module just doesn't compile and the metadata reports zero harnesses, a vacuous pass with nothing verified.

Append the flags in `kani_compiler::run()`, after `is_kani_compiler()` has confirmed kani mode and stripped the marker arg. For `cargo kani` this is idempotent — scalar `-C`/`-Z` flags are last-flag-wins and `--cfg`/`--check-cfg` are additive. For any other caller the compiler now enforces the correct values regardless of what they passed.

Deliberately not included: `-Clinker=echo` (would clobber a real linker a build system needs for rlib output), `-Zcrate-attr=feature(register_tool)` / `-Zcrate-attr=register_tool(kanitool)` (rustc errors on a duplicate registration, and `cargo kani` already passes them — appending would break every `cargo kani` invocation; omitting them is a loud `unrecognized tool kanitool`, not silent unsoundness), and the conditional and diagnostic-format flags that encode session intent.

The `script-based-pre` test invokes `kani-compiler` directly with only install paths, `--kani-compiler`, `-Zcrate-attr`, and a reachability flag — none of the eight defaults — and asserts the `#[cfg(kani)]`-gated harness still appears in the metadata. Verified by reverting the change and re-running: the same invocation then errors immediately on the missing `panic=abort`.


kani-compiler/src/kani_compiler.rs
```diff
@@ -32,9 +32,52 @@ use rustc_public::rustc_internal;
...
+const KANI_REQUIRED_RUSTC_ARGS: &[&str] = &[
+    "-Cpanic=abort",
+    "-Coverflow-checks=on",
+    "-Csymbol-mangling-version=v0",
+    "-Zalways-encode-mir",
+    "-Zpanic_abort_tests=yes",
+    "-Zmir-enable-passes=-RemoveStorageMarkers",
+    "--cfg=kani",
+    "--check-cfg=cfg(kani)",
+];
```
feliperodri, line 106: Can we confirm that rustc accepts these in this single-token `--flag=value` form on this path?

The main guarantee of this PR depends on `cfg(kani)` definitely being enabled. If rustc expects these as separate argv tokens (`--cfg`, `kani` and `--check-cfg`, `cfg(kani)`), then we may not actually be enforcing the invariant we want here.

If that form is definitely accepted, a test that proves this exact encoding works would make me more confident. Otherwise, I think these should be passed as split tokens instead.

tests/script-based-pre/compiler_defaults/compiler_defaults.sh
```diff
@@ -0,0 +1,59 @@
...
+KANI_COMPILER="${KANI_HOME}/bin/kani-compiler"
+KANI_LIB="${KANI_HOME}/lib"
+
+# Minimal invocation. Deliberately omits every flag KANI_REQUIRED_RUSTC_ARGS
+# now defaults: -Cpanic=abort, -Coverflow-checks=on, -Csymbol-mangling-version,
+# -Zalways-encode-mir, -Zpanic_abort_tests, -Zmir-enable-passes, --cfg=kani,
+# --check-cfg=cfg(kani). What remains is what every caller still has to pass:
+# install paths, the routing marker, -Zunstable-options (gates `--extern
+# noprelude:`), -Zcrate-attr (errors on duplicate so it stays caller-supplied),
+# and the reachability intent.
```
feliperodri: Oh, nice regression test it proves the defaults are added for a minimal direct invocation.

One gap I still see is that the PR rationale depends on `kani-compiler` enforcing these values unconditionally, including when the caller passes conflicting values. This test currently exercises the “missing flags” case, but not the “conflicting flags” case.

Could we add a case that passes an explicit conflict (for example `-Coverflow-checks=off`) and verify Kani still behaves with the required value? That would directly test the intended “last-flag-wins” behavior this change relies on.

kani-compiler/src/kani_compiler.rs
lovesegfault: ah, nice catch, i added coverage for this in the compiler_defaults test. the fixture now carries a deliberately undeclared cfg as a negative control :)

FWIW this is the same encoding kani-driver has always passed (`base_rustc_flags()`), and rustc's `getopts` CLI treats `--flag=value` and `--flag value` identically

tests/script-based-pre/compiler_defaults/compiler_defaults.sh
lovesegfault: i added the conflict case, a second invocation now passes `-Cpanic=unwind -Coverflow-checks=off -Csymbol-mangling-version=legacy` and must behave identically to the minimal one :)

feliperodri (approved):

feliperodri:
@lovesegfault could you take a look at Copilot comments and address the conflicts so we can merge the PR?

feliperodri:
Note for direct callers: because `--check-cfg=cfg(kani)` is appended unconditionally, a caller that passed no `--check-cfg` now gets cfg checking enabled, so its own undeclared cfgs begin emitting `unexpected_cfgs` warnings. Warnings only, never a build failure. cargo kani is unaffected since kani-driver already passes this flag.

PR #4612, opened 2026-06-23
Emit explicit `contract::` symbols for functions with contracts

tautschnig:
CBMC's DFCC contract instrumentation (`--enforce-contract` / `--replace-call-with-contract`) expects a dedicated `contract::<fn>` symbol to exist for any function carrying a contract. For C inputs this symbol is created by the ANSI-C typechecker, which also moves the contract clauses (e.g. `c_spec_assigns`) onto it.

Kani builds goto-programs directly and never runs that typechecker, so the symbol was absent. CBMC used to paper over this by deriving an empty contract symbol from the function on the fly, but that fallback was removed for soundness in diffblue/cbmc#8739, turning the missing symbol into a hard error and breaking every `#[kani::proof_for_contract]` harness.

Synthesize the companion `contract::<name>` symbol during symbol-table serialization. It shares the function's contract-bearing type, has no body, and is marked as a property, reconstructing exactly the symbol CBMC's typechecker (and its old fallback) produced.

This is a necessary (though not necessarily sufficient) change to enable upgrading to newer CBMC versions.


feliperodri (approved):
We can just create a ticket to investigate the performance improvement mentioned in the Copilot comment. LGTM 🚢 it.

PR #4619, opened 2026-07-02
Update CBMC dependency to 6.10.0

### LINKED ISSUE #4605: CBMC upgrade to 6.9.0 failed
Updating CBMC from 6.8.0 to 6.9.0 failed.
The failed automated run [can be found here.](https://github.com/model-checking/kani/actions/runs/26738496646)

### LINKED ISSUE #4617: CBMC upgrade to 6.10.0 failed
Updating CBMC from 6.8.0 to 6.10.0 failed.
The failed automated run [can be found here.](https://github.com/model-checking/kani/actions/runs/28352125207)

tautschnig:
Upgrading to 6.9.0 requires changes as described below. Upgrading from 6.9.0 to 6.10.0 does not require any further changes.

Changes in CBMC's set of static objects require retuning shadow-memory object-count test: The `unsupported_num_objects` test depends on CBMC's internal object-ID numbering: it creates N objects and expects the (N+k)-th object's ID to cross Kani's 1024-object shadow-memory limit. Newer CBMC allocates one fewer auxiliary object, so the previous N values (1019 pass / 1020 fail) no longer straddle the limit -- the fail harness verified successfully instead of failing.

Bump the thresholds by one (1020 pass / 1021 fail) to restore the boundary, update the explanatory comment, and note that these counts are sensitive to CBMC's object numbering and may need adjusting on future CBMC upgrades.

CBMC 6.10 makes the `slow` harness's Display/`to_string` formatting path far more memory-heavy: the original `"foo"`/unwind(6) version peaks at ~26 GB with Kani's default Cadical solver and deterministically OOM-kills the 16 GB GitHub-hosted `perf` runner ("runner has received a shutdown signal"). No solver is both under 16 GB and fast enough (minisat2/kissat stay ~8 GB but do not finish in >15 min), so the harness itself must shrink.

Use an empty payload (formatted result `"A."`) and unwind(3) -- the smallest bound that still fully unrolls the formatting loops. This keeps peak memory to ~11 GB while still exercising the trait/formatting path the test guards. Both harnesses verify successfully.

Resolves: #4605
Resolves: #4617


tautschnig:
We need to investigate why CBMC's brew releases are no longer fully working.

feliperodri (approved):

PR #4642, opened 2026-07-14
Bound each regression test's wall time to avoid unattributable hangs

tautschnig:
kani-regression.sh ran compiletest with no per-test timeout, so a test whose verification fails to terminate (e.g. under a new toolchain) stalled the whole run until the CI job was force-killed hours later. This is especially harmful for the nightly toolchain-upgrade automation: a clean regression failure lets it open a tracking issue, whereas an unbounded hang produces a ~3h job timeout with no signal.

Pass a per-test --timeout (default 900s, override via KANI_REGRESSION_TEST_TIMEOUT) to the compiletest invocations, mirroring kani-perf.sh. The cap is a guardrail against hangs, well above the slowest legitimate test.

### DIFF
--- a/scripts/kani-regression.sh
+++ b/scripts/kani-regression.sh
@@ -71,13 +71,29 @@ echo "-----------------------------"
 cargo build -p kani-cov
 
 # Extract testing suite information and run compiletest
+# Bound each test's wall time so a hanging test (e.g. one whose verification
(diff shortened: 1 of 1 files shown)


feliperodri:
CI failure: the expected test `tests/expected/shadow/unsupported_num_objects/test.rs` now times.

feliperodri (approved):

PR #4655, opened 2026-07-18
Upgrade Rust toolchain to nightly-2026-02-06

tautschnig:
rust-lang/rust stopped encoding *optimized* MIR for ADT constructors (`instance_mir` uses `mir_for_ctfe` for them). As a result, `rustc_public::Instance::body()` returns `None` for constructors imported from other crates (its `has_body` check keys off `is_mir_available`, i.e. optimized MIR), so codegenning a constructor used as a function value (e.g. `.map(Some)`) panicked on `instance.body().unwrap()`.

Reconstruct the missing constructor body the same way `rustc_public`'s `BodyBuilder` would (fetch `instance_mir`, monomorphize, convert to stable MIR) when `body()` is `None` and the instance is a constructor; any other bodyless instance still panics.

### DIFF
--- a/kani-compiler/src/kani_middle/transform/mod.rs
+++ b/kani-compiler/src/kani_middle/transform/mod.rs
@@ -29,14 +29,41 @@ use crate::kani_middle::transform::stubs::{ExternFnStubPass, FnStubPass};
 use crate::kani_queries::QueryDb;
 use automatic::{AutomaticArbitraryPass, AutomaticHarnessPass};
 use dump_mir_pass::DumpMirPass;
-use rustc_middle::ty::TyCtxt;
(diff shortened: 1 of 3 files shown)


feliperodri (approved):
Tried to re-run the failed job but got:
```
Run actions/checkout@v6
Error: Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.
```

PR #4660, opened 2026-07-20
Update Charon submodule to v0.1.69

tautschnig:
Advance the Charon pin from v0.1.68 (63f7749d) to v0.1.69 (485bd36e), the next step of the incremental effort to catch the pinned Charon up with upstream (the endgame is dropping scripts/charon-patch.diff entirely).

Charon v0.1.69 makes `Vector::elem_count` the way to obtain the number of non-empty slots (`len` is removed); adapt the tuple-field-projection translation in the LLBC backend accordingly.

Refresh scripts/charon-patch.diff so its hunks apply against the v0.1.69 sources (context-line drift only; the patch content is unchanged).


feliperodri (approved):
🚢 it.

PR #4669, opened 2026-07-24
Update Charon submodule to v0.1.71

tautschnig:
Advance the Charon pin from v0.1.69 (485bd36e) to v0.1.71 (de910497), continuing the incremental effort to catch the pinned Charon up with upstream (the endgame is dropping scripts/charon-patch.diff entirely).

Charon v0.1.70/71 make error levels explicit (`span_err` takes a `Level`) and add `ItemMeta::lang_item` tracking (`None` for the items Kani translates, which carry no `#[lang]` attribute).

Refresh scripts/charon-patch.diff so its hunks apply against the v0.1.71 sources (context-line drift only; the patch content is unchanged).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -288,7 +288,7 @@ checksum = "f079e83a288787bcd14a6aea84cee5c87a67c5a3e660c30f557a3d24761b3527"
 
 [[package]]
 name = "charon"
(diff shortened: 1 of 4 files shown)


feliperodri (approved):
🚢 it

Issue #1150, opened 2022-05-03
High offsets cause wrapping-around behavior in CBMC


adpaco-aws:
In an example introduced in #1131 , we reproduced an issue where a high offset causes a "wrapping around" behavior in CBMC.
There is more information on the test, but basically we should either enforce stricter checks than Rust (because of CBMC object bits limitations) or add these checks to CBMC itself.

celinval:
I think we will need to stop using CBMC's pointer offset operation in the case of `wrapping_offset`. Instead, we need to convert the pointer into a `usize`, perform a `wrapping_add` considering the object size, then convert back to a pointer.

celinval:
BTW, here is a small example that shows the problem:
```rust
/// This harness shows the issue with the current implementation of wrapping offset.
///
/// Invoking `wrapping_byte_offset` should return a pointer that is different from the original
/// pointer if the offset value is not 0.
#[kani::proof]
fn fixme_incorrect_wrapping_offset() {
    let ptr: *const u8 = &0u8;
    let offset = kani::any_where(|v: &isize| *v != 0);
    let new_ptr = ptr.wrapping_byte_offset(offset);
    assert_ne!(ptr, new_ptr, "Expected new_ptr to be different than ptr");
}

```

tautschnig:
Status update after a detailed investigation of this issue (all findings reproduced against current `main` + CBMC 6.10.0; details in the linked PRs).

**Root cause, characterized:** CBMC's propositional encoding packs pointers into `object|offset` bits, so pointer arithmetic wrapped offsets at 2^(64 − object_bits) instead of 2^64 (with Kani's default `--object-bits 16`: at 2^48). Constant-folded arithmetic, however, is computed at full width by CBMC's simplifier — so semantically identical harnesses gave different verdicts for constant vs. symbolic offsets. Beyond the spurious counterexample shown above, this was a soundness hole: `p.offset(k)` on an 8-byte allocation with symbolic `k == 1 << 48` **verified successfully** — the wrapped result aliases `p`, so the same-allocation UB check passed spuriously.

**The fixes and how they compose:**

[...]

PR #4671, opened 2026-07-24
Fix OffsetModel missing UB for offsets wrapping CBMC's pointer encoding

tautschnig:
CBMC encodes pointers as an (object, offset) pair where the offset field has pointer_width - object_bits bits. Pointer arithmetic therefore wraps offsets at 2^(64 - object_bits) rather than 2^64 (https://github.com/model-checking/kani/issues/1150). For the OffsetModel this meant that a symbolic byte offset that is a non-zero multiple of 2^(64 - object_bits) (2^48 with Kani's default of 16 object bits) made `wrapping_byte_offset` alias the original pointer, so the subsequent same-allocation safety check passed spuriously and genuine out-of-bounds `offset` calls verified successfully - including "proving" the false fact that the result equals the base pointer.

Detect the wrap by additionally requiring that the address of the result is consistent with full-width integer arithmetic on the address of the original pointer; any truncation in the pointer addition falsifies this equation. Two alternative formulations do not work:

* Computing the new pointer via integer arithmetic (addr().wrapping_add_signed(off) followed by a cast back, as previously suggested in a FIXME here) is sound, but makes the solver case-split the integer address over every object; harnesses using e.g. `sort()` then run out of memory. These are the "unexpected failures" mentioned in the removed comment (reproduced on: expected/cover/cover-pass, expected/string-repeat, expected/loop-backedge, expected/shadow/unsupported_num_objects).

* Comparing POINTER_OFFSET(new) against full-width POINTER_OFFSET(orig) + offset is defeated by CBMC's simplifier, which rewrites POINTER_OFFSET(p + k) to POINTER_OFFSET(p) + k in full-width arithmetic, making the equation trivially true.

The `addr()` transmute is opaque to that simplification, keeps all operations in the pointer domain, and adds negligible solver cost: the full expected (459) and kani (593) suites pass unchanged.

The underlying encoding wrap is fixed on the CBMC side in https://github.com/diffblue/cbmc/pull/9134 (out-of-range results are directed to the invalid object). The equation here also states Rust's address-arithmetic contract for `offset` and produces the same verdicts with and without that CBMC fix, so it is kept as a semantic guard rather than a version-specific workaround.

Partially addresses #1150: this fixes the missed-UB soundness hole in the checked `offset`/`arith_offset` intrinsic model. Not addressed is the `wrapping_offset` family, which Kani codegens to plain CBMC pointer addition: under CBMC versions without the encoding fix its result aliases the base pointer for wrap-multiple offsets (tests/kani/PointerOffset/fixme_wrap_nonzero_offset.rs), and even with the CBMC fix the result address is indeterminate rather than the exactly-wrapped address that Rust defines. Exact modeling would require integer arithmetic plus an integer-to-pointer cast, which makes the solver case-split over all objects (see above).

### DIFF
--- a/library/kani_core/src/models.rs
+++ b/library/kani_core/src/models.rs
@@ -124,16 +124,39 @@ macro_rules! generate_models {
                 let (byte_offset, overflow) = offset.overflowing_mul(t_size);
                 kani::safety_check(!overflow, "Offset in bytes overflows isize");
                 let orig_ptr = ptr.to_const_ptr();
-                // NOTE: For CBMC, using the pointer addition can have unexpected behavior
-                // when the offset is higher than the object bits since it will wrap around.
-                // See for more details: https://github.com/model-checking/kani/issues/1150
+                // The equation below asserts Rust's semantic contract for `offset`:
+                // the address of the result must equal the address of the original
+                // pointer plus the byte offset, in full-width (2^64) integer
+                // arithmetic.
                 //
-                // However, when I tried implementing this using usize operation, we got some
-                // unexpected failures that still require further debugging.
-                // let new_ptr = orig_ptr.addr().wrapping_add_signed(byte_offset) as *const T;
+                // Besides documenting that contract, it guards against CBMC's
+                // pointer encoding wrapping offsets at
+                // 2^(pointer_width - object_bits) rather than 2^pointer_width: for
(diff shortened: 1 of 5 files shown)


feliperodri (approved):
I'd implement the Copilot comments before merging it.

PR #4672, opened 2026-07-26
Add support for `volatile_copy_memory`, `volatile_copy_nonoverlapping_memory` and `volatile_set_memory`

ivmat:
Implements three of the intrinsics still unchecked on the tracking issue #1163:
`volatile_copy_memory`, `volatile_copy_nonoverlapping_memory` and `volatile_set_memory`.

## Why the existing placeholders could not simply be un-gated

The two copy intrinsics already had arms, but behind `unstable_codegen!`:

```rust
Intrinsic::VolatileCopyMemory => unstable_codegen!(codegen_intrinsic_copy!(Memmove)),
Intrinsic::VolatileCopyNonOverlappingMemory => {
    unstable_codegen!(codegen_intrinsic_copy!(Memcpy))
}
```

`unstable_codegen!` is declared as `($($tt:tt)*)` and never expands its tokens — it always emits a
`codegen_unimplemented_expr`. The bodies above were therefore never type-checked, and they have since
rotted: `codegen_intrinsic_copy!` no longer exists anywhere in `kani-compiler`. Removing the gate alone
does not compile, so both arms are rewritten rather than un-gated.

`volatile_set_memory` had no `Intrinsic` variant at all and fell through to the unsupported catch-all.

## Argument order

The Rust intrinsics take the destination first:

```rust
pub unsafe fn volatile_copy_memory<T>(dst: *mut T, src: *const T, count: usize);
pub unsafe fn volatile_copy_nonoverlapping_memory<T>(dst: *mut T, src: *const T, count: usize);
```

whereas `codegen_copy` consumes source first (`let src = fargs.remove(0); let dst = fargs.remove(0);`)
and indexes `farg_types[0]` / `farg_types[1]` for the `src` and `dst` alignment checks respectively.
Both `fargs` and `farg_types` are therefore swapped before delegating. Without swapping *both*, the two
alignment assertions are silently applied to the wrong pointers — a discrepancy that a test using two
equally-aligned buffers would not detect.

`volatile_set_memory(dst, val, count)` matches `write_bytes(dst, val, count)`, so no swap is needed there.

## Soundness of reusing the non-volatile codegen

Volatility constrains what the *optimizer* may do — it must not elide, duplicate or reorder the access.
It is not a memory-safety property: the UB conditions for these three intrinsics are exactly those of
`copy`, `copy_nonoverlapping` and `write_bytes`. Kani's codegen performs no optimization that volatility
is meant to suppress, so there is nothing additional to model. This mirrors the existing treatment of
`volatile_load` / `volatile_store`, which already delegate to their non-volatile counterparts.

## Points-to analysis

Adding the `VolatileSetMemory` variant required one further line. `points_to_analysis.rs` matches
exhaustively over `Intrinsic` with a final `unimplemented!()` wildcard; `WriteBytes` is protected from it
by being listed in `is_identity_aliasing_intrinsic`. Before this change `volatile_set_memory` reached the
analysis as `Intrinsic::Unimplemented { .. }`, which is explicitly handled as a no-op. Introducing the new
variant would have routed it to the panicking wildcard instead, so it is listed alongside `WriteBytes` in
the same way.

The other two visitors that match on `Intrinsic`
(`check_uninit/ptr_uninit/uninit_visitor.rs`, `check_uninit/delayed_ub/initial_target_visitor.rs`) have
non-panicking fallbacks, so behaviour there is unchanged.

## Tests

- `tests/kani/Intrinsics/Volatile/copy.rs` — both copy intrinsics move the expected bytes, including an
  overlapping case for `volatile_copy_memory` (memmove semantics) to distinguish it from the
  non-overlapping variant.
- `tests/kani/Intrinsics/Volatile/set.rs` — `volatile_set_memory` fills the destination, full and partial.
- `tests/expected/intrinsics/volatile_copy/overlapping/` — `volatile_copy_nonoverlapping_memory` on
  overlapping ranges fails, matching the existing
  `tests/expected/intrinsics/copy-nonoverlapping/copy-overlapping/` test for the non-volatile intrinsic.
- `tests/expected/intrinsics/volatile_copy/unaligned/` — a misaligned pointer fails the alignment check.

Test layout follows #1347, which added `volatile_load` support.

## Related

Towards #1163. The remaining unchecked volatile entries — `unaligned_volatile_load` and
`unaligned_volatile_store` — are deliberately left out of this PR: the gated body for
`unaligned_volatile_load` is a plain dereference, which does not model the unaligned access itself, so
those two need a separate decision about how unaligned accesses should be represented.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0
and MIT licenses.

### DIFF
--- a/docs/src/rust-feature-support/intrinsics.md
+++ b/docs/src/rust-feature-support/intrinsics.md
@@ -232,10 +232,10 @@ unchecked_sub | Yes | |
 unlikely | Yes | |
 unreachable | Yes | |
 variant_count | Yes | |
-volatile_copy_memory | No | See [Notes - Concurrency](#concurrency) |
-volatile_copy_nonoverlapping_memory | No | See [Notes - Concurrency](#concurrency) |
+volatile_copy_memory | Partial | See [Notes - Concurrency](#concurrency) |
+volatile_copy_nonoverlapping_memory | Partial | See [Notes - Concurrency](#concurrency) |
 volatile_load | Partial | See [Notes - Concurrency](#concurrency) |
-volatile_set_memory | No | See [Notes - Concurrency](#concurrency) |
+volatile_set_memory | Partial | See [Notes - Concurrency](#concurrency) |
 volatile_store | Partial | See [Notes - Concurrency](#concurrency) |
 wrapping_add | Yes | |
 wrapping_mul | Yes | |
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
@@ -576,11 +576,45 @@ impl GotocCtx<'_, '_> {
             Intrinsic::UncheckedDiv => codegen_op_with_div_overflow_check!(div),
             Intrinsic::UncheckedRem => codegen_op_with_div_overflow_check!(rem),
             Intrinsic::Unlikely => self.codegen_expr_to_place_stable(place, fargs.remove(0), loc),
-            Intrinsic::VolatileCopyMemory => unstable_codegen!(codegen_intrinsic_copy!(Memmove)),
+            // Volatility is an I/O-observability property (it tells the
+            // optimizer not to elide or reorder the access) and not a
+            // memory-safety property: the UB conditions for
+            // `volatile_copy_memory`/`volatile_copy_nonoverlapping_memory`
+            // are exactly those of the non-volatile `copy`/
+            // `copy_nonoverlapping` intrinsics (see `codegen_copy`'s doc
+            // comment). Since Kani's codegen never performs the kind of
+            // optimization that volatility is meant to suppress, there is
+            // nothing extra to model, and it is sound to reuse
+            // `codegen_copy` as-is.
+            //
+            // Both intrinsics are declared as `(dst, src, count)`, but
+            // `codegen_copy` expects `fargs`/`farg_types` ordered as
+            // `[src, dst, ..]` (it does `fargs.remove(0)` for `src` followed
+            // by `fargs.remove(0)` for `dst`). So we swap the first two
+            // entries of `fargs`, and build a `farg_types` slice with the
+            // first two entries swapped as well, before delegating.
+            Intrinsic::VolatileCopyMemory => {
+                fargs.swap(0, 1);
+                let swapped_types = [farg_types[1], farg_types[0]];
+                self.codegen_copy(intrinsic_str, false, fargs, &swapped_types, Some(place), loc)
+            }
             Intrinsic::VolatileCopyNonOverlappingMemory => {
-                unstable_codegen!(codegen_intrinsic_copy!(Memcpy))
+                fargs.swap(0, 1);
+                let swapped_types = [farg_types[1], farg_types[0]];
+                self.codegen_copy(intrinsic_str, true, fargs, &swapped_types, Some(place), loc)
             }
             Intrinsic::VolatileLoad => self.codegen_volatile_load(fargs, farg_types, place, loc),
+            // `volatile_set_memory(dst, val, count)` has the same argument
+            // order as `write_bytes(dst, val, count)` (no `dst`/`src` swap
+            // is needed, unlike the two copy intrinsics above). As with the
+            // copy intrinsics, volatility here is purely an
+            // optimizer-observability concern, so the UB conditions match
+            // `write_bytes` exactly and `codegen_write_bytes` can be reused
+            // unmodified.
+            Intrinsic::VolatileSetMemory => {
+                assert!(self.place_ty_stable(place).kind().is_unit());
+                self.codegen_write_bytes(fargs, farg_types, loc)
+            }
             Intrinsic::VolatileStore => {
                 assert!(self.place_ty_stable(place).kind().is_unit());
                 self.codegen_volatile_store(fargs, farg_types, loc)
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/intrinsic.rs
@@ -576,16 +576,21 @@ impl GotocCtx<'_, '_> {
             Intrinsic::UncheckedDiv => codegen_op_with_div_overflow_check!(div),
             Intrinsic::UncheckedRem => codegen_op_with_div_overflow_check!(rem),
             Intrinsic::Unlikely => self.codegen_expr_to_place_stable(place, fargs.remove(0), loc),
-            // Volatility is an I/O-observability property (it tells the
-            // optimizer not to elide or reorder the access) and not a
-            // memory-safety property: the UB conditions for
-            // `volatile_copy_memory`/`volatile_copy_nonoverlapping_memory`
-            // are exactly those of the non-volatile `copy`/
-            // `copy_nonoverlapping` intrinsics (see `codegen_copy`'s doc
-            // comment). Since Kani's codegen never performs the kind of
-            // optimization that volatility is meant to suppress, there is
-            // nothing extra to model, and it is sound to reuse
-            // `codegen_copy` as-is.
+            // These two document their safety requirements as "consistent
+            // with `copy`" / "consistent with `copy_nonoverlapping`", so
+            // reusing `codegen_copy` checks the conditions their own docs
+            // state. Volatility itself constrains the optimizer (the access
+            // must not be elided or reordered), which Kani has no need to
+            // model since it performs no such optimization.
+            //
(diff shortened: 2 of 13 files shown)


ivmat:
Pushed two follow-up commits after a review pass, and corrected one claim in the description above.

**The comment about volatility overclaimed.** It said the UB conditions are *exactly* those of the non-volatile counterparts. That is not right as a generalisation: the volatile family's docs additionally permit pointers outside any Rust allocation (the MMIO carve-out documented on `ptr::read_volatile`), which `copy`/`copy_nonoverlapping`/`write_bytes` do not. The modelling is still sound, but for a narrower reason, and the comments now say so: these intrinsics document their safety requirements as *consistent with* `copy_nonoverlapping` / `write_bytes`, so reusing that codegen checks **at least** the documented UB conditions. It is conservative rather than exact — a legal MMIO access may be rejected by `--pointer-check`, but real UB is never missed.

**`volatile_set_memory` was missing from the memory-initialization visitor.** The compiler matches on `Intrinsic` in three places — codegen, points-to, and `check_uninit`'s visitor — and only the first two had been updated, so under `-Z uninit-checks` the intrinsic produced a spurious "does not support reasoning about memory initialization" failure. It now shares the `write_bytes` arm: same `(dst, val, count)` shape, same initialization effect.

Also added `tests/expected/intrinsics/volatile_set/out-of-bounds`, mirroring the existing `write_bytes` test — the happy-path test alone did not pin that reusing `codegen_write_bytes` carries its safety checks across.

Full regression run locally on CBMC 6.10.0: no failures introduced (the only failures are three pre-existing `Quantifiers` tests that fail identically on unmodified `main` here, for want of z3).

ivmat:
Three test fixes from a review pass, all the same defect class.

**The `dst` pointers in the copy tests were derived from a shared borrow.** All three harnesses formed
`dst` by casting a pointer derived from `arr.as_ptr()` to `*mut i32` and then wrote through it, which is
undefined behavior under Rust's aliasing rules regardless of what the intrinsic does with it. Both
pointers are now derived from `arr.as_mut_ptr()`.

For the two `tests/expected` harnesses this was not just a technicality, and it is the sharper reason to
care: each is meant to fail for exactly one stated reason — the overlap precondition in
`volatile_copy/overlapping`, the alignment precondition in `volatile_copy/unaligned` — and a second,
unintended source of UB means the harness can pass its `expected` file while proving less than it
claims. Checked under Miri on the non-volatile counterparts (`ptr::copy` /
`ptr::copy_nonoverlapping`, whose aliasing requirements are the ones these intrinsics' docs defer to):
before, `volatile_copy/overlapping` reported a `SharedReadOnly` write violation rather than the overlap
it is named for; after, it reports `copy_nonoverlapping called on overlapping ranges`, the unaligned
harness reports only the misalignment, and the `tests/kani` overlap proof runs clean with its three
assertions intact.

**One of the three was not reported.** The automated review found the instances in
`tests/kani/Intrinsics/Volatile/copy.rs` and `tests/expected/intrinsics/volatile_copy/overlapping`; the
one in `tests/expected/intrinsics/volatile_copy/unaligned` came out of grepping the diff for the whole
pattern afterwards. Worth stating plainly, since it was the same mistake made three times in one PR.

Also re-verified, since missing it caused two earlier defects here: all three places that match on
`Intrinsic` — codegen, points-to, and `check_uninit`'s visitor — cover all three intrinsics, and the two
pre-existing copy arms in points-to and the uninit visitor already handle the reversed
`(dst, src, count)` order correctly.

Regression re-run locally on **CBMC 6.10.0** (the version this tree's `kani-dependencies` declares),
against a baseline build of `de332bbb9`, the commit this branch is based on: **no failures introduced**
— the failing set is identical on the baseline and on this branch, test for test. `kani` suite 593
passed / 3 failed here vs 591 / 3 on the baseline; `expected` suite 461 / 2 vs 458 / 2. All five
failures carry `#[kani::solver(z3)]` and fail because z3 is not installed on this machine, not because
of anything in this PR.

One correction to my earlier comment while I am here: it said "the only failures are three pre-existing
`Quantifiers` tests ... for want of z3". The cause was right but the set was incomplete — there are
**five** such tests, the two others being
`expected/function-contract/as-assertions/assert-postconditions.rs` and
`expected/loop-contract/loop_assigns_for_vec.rs`. Same cause, wider than stated.

ivmat:
One practical note for whoever reviews or lands this, from exercising the new codegen end-to-end.

I ran the five now-supported intrinsics against real harnesses — the ones in `verify-rust-std`'s challenge-2 set, which have sat `#[cfg(not(kani))]`-gated precisely because Kani refused them. Four verify in about a second each. The fifth, the self-overlapping `volatile_copy_memory` case, ran **40 minutes and 12.3 GB without converging.**

**That wall is the harness, not this PR's codegen.** The harness was written as:

```rust
let shift: usize = kani::any();
kani::assume(shift >= 1 && shift < N);
...
unsafe { volatile_copy_memory(dst_ptr, src_ptr, N - shift) };
for i in 0..(N - shift) {
    assert_eq!(buf[i + shift], original[i]);
}
```

A symbolic `shift` combined with `for i in 0..(N - shift)` puts a symbolic trip count in the formula. That is the same self-overlap wall plain `copy` hits, and `verify-rust-std` already contains the resolution — its `check_copy_overlapping_shift_no_ub` uses a fixed representative `SHIFT` plus a symbolic *index* the solver may choose freely, which is equivalent in coverage without the symbolic-bound loop:

```rust
const SHIFT: usize = 3;
...
let i: usize = kani::any();
kani::assume(i < N - SHIFT);
assert_eq!(buf[i + SHIFT], original[i]);
```

Same harness, same property, same build: **no result in 40 minutes → 0.133 s.**

Worth flagging because the failure mode is misleading. This defect was invisible for months — Kani's "not currently supported" verdict fires before codegen, so the harness could never run far enough to blow up. Anyone who lands this PR and un-gates the obvious `volatile_copy_memory` harness will hit a 40-minute hang and could reasonably conclude the new codegen is at fault. It isn't.

**How this was measured, and its limits:** on a build of `d4df833c8` — the Kani commit `verify-rust-std` currently pins — with this PR's commits and #4673's cherry-picked on top, CBMC 6.8.0. I used a graft rather than this branch directly because this branch is on `nightly-2026-02-06` while `verify-rust-std` and its pinned Kani are on `nightly-2025-11-25`, so the branch as-is cannot verify that tree. So this is evidence about the composed pipeline (new codegen ⊕ CBMC) on a patched, non-official Kani — not a claim about official Kani, and not a claim that the fixed-shift harness is as strong as the symbolic one. It proves the property for a representative shift distance, not universally over distances.

ivmat:
The `perf-benchcomp` failure here looks like gate noise rather than a regression from this PR. Sharing the evidence in case it's useful.

**The emitted program is unchanged.** benchcomp reports `number_vccs` and `number_program_steps` alongside the wall-clock metrics, and both are deterministic functions of the compiled program. They are identical between `kani_old` and `kani_new` for all 130 benchmarks, in all three runs on this branch. For the benchmark that tripped the gate (`slice::tests::vectored_copy_fuzz_test`) they are `(3085, 93093)` on both sides every time. In the same run its `symex_runtime` went 22.6156 → 22.2413 s (1.7% *faster*) while `solver_runtime` rose 55%.

That matches the diff: the compiler-side changes are five hunks, each inside a match arm keyed on `VolatileCopyMemory` / `VolatileCopyNonOverlappingMemory` / `VolatileSetMemory`. `codegen_copy` and `codegen_write_bytes` are called, not modified, and `scripts/` is untouched, so CBMC and its flags are identical on both sides.

**The same gate fires on unrelated PRs whose emitted programs are also unchanged.** In the same week:

- #4679 (`autoharness-generics`) — `vectored_copy_fuzz_test` 100.486 → 260.364 s (**2.591×**), with 0 of 130 benchmarks showing any change in `vccs`/`steps`
- #4678 (`autoharness-raw-pointers`) — `u32_u16_differential` 14.008 → 23.847 s (**1.702×**), likewise 0 of 130

Both ratios are larger than the 1.5517× here.

**On this branch specifically**, the job also failed on 2026-07-29, on a *different* benchmark (`u32_u16_differential`, 1.918×), on a head that differs from the current one only in files under `tests/` — and that same head passed this job on 2026-07-26.

The one thing that would settle it is a re-run: the failing ratio is `max(new)/min(old)` across six measurements of identical work, and paired with either of the other two base draws it would have been 1.145 or 1.195.

I'm not proposing a code change, since I can't find a path from this diff to that benchmark — happy to be corrected if I've missed one. Whether the threshold itself is worth revisiting is entirely your call.

feliperodri (approved):

feliperodri:
Thanks for the contribution @ivmat!

ivmat:
thank you @feliperodri  for checking the PR !

Issue #1597, opened 2022-08-26
Ambiguous name error with `assert!` macro


tedinski:
The current version of the `futures` library builds fine. But many crates seem to have `compat` features to enable cross-compatibility with the `0.1` version of the `futures` library. This version does something odd that creates build failures.

```
$ cargo new futures-old
     Created binary (application) `futures-old` package
$ cd futures-old/
$ cargo add futures@0.1
    Updating crates.io index
      Adding futures v0.1 to dependencies.
             Features:
             + use_std
$ cargo kani
   Compiling futures v0.1.31
error[E0659]: `assert` is ambiguous
   --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-0.1.31/src/future/select_all.rs:41:5
    |
41  |     assert!(ret.inner.len() > 0);
    |     ^^^^^^ ambiguous name
    |
    = note: ambiguous because of a conflict between a name from a glob import and an outer scope during import or macro resolution
note: `assert` could refer to the macro imported here
   --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-0.1.31/src/future/select_all.rs:5:5
    |
5   | use std::prelude::v1::*;
    |     ^^^^^^^^^^^^^^^^^^^
    = help: consider adding an explicit import of `assert` to disambiguate
    = help: or use `self::assert` to refer to this macro unambiguously
note: `assert` could also refer to the macro imported here
   --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-0.1.31/src/lib.rs:163:1
    |
163 | #[macro_use]
    | ^^^^^^^^^^^^
```

PR #4687, opened 2026-07-28
Add --no-assert-overrides escape hatch for macro-override conflicts

tautschnig:
### Description

Adds a `--no-assert-overrides` flag that skips injecting Kani's macro overrides (`assert!`, `panic!`, ...) into the crate under verification.

Kani injects `#[macro_use] extern crate std as _kani_std_macros;` so that its overrides shadow the prelude macros. rustc's restricted-shadowing rule makes a *glob import* of the same macro name ambiguous against the macro_use prelude — E0659 with kind `GlobVsOuter`, which is a hard error with no lint-level escape. Crates that glob-import a module re-exporting `core::assert` therefore cannot compile under Kani. #4666 already skips the injection for external dependencies, but the same pattern in the *crate under verification* still fails: verifying libc itself (whose `prelude!` macro re-exports `core::assert` into an internal prelude that its modules glob-import) fails with ``error[E0659]: `assert` is ambiguous`` (this is #1597 / the primary-crate remainder of #4665).

With the flag, verification remains sound — assertion failures are still detected, just reported as generic panics without the original condition/message, matching how external and `no_std` dependencies already behave since #4666. The flag is an escape hatch rather than a default change; the error-message-preserving behavior is untouched for everyone else.

Found while evaluating `kani autoharness` on the top-100 crates.io crates, where libc 0.2.189 fails to build. With this flag (plus #4682 for its `no_std`-ness), libc compiles cleanly under Kani.

### Testing

New script-based test `cargo_no_assert_overrides` reproducing the libc pattern (macro-generated prelude re-exporting `core::assert`, glob-imported by a module calling `assert!`): asserts E0659 without the flag, successful compilation with it, and that an intentionally failing assertion still fails verification (soundness of the degraded mode).

`kani-driver` unit tests and formatting pass; manually verified libc 0.2.189 compiles under `cargo kani --no-assert-overrides`.

### DIFF
--- a/tests/script-based-pre/cargo_no_assert_overrides/no-assert-overrides.sh
+++ b/tests/script-based-pre/cargo_no_assert_overrides/no-assert-overrides.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+echo "[without flag]"
+cargo kani 2>&1 | grep -m1 'E0659'
+echo "[with flag]"
+cargo kani --no-assert-overrides
--- a/tests/script-based-pre/cargo_no_assert_overrides/no-assert-overrides.sh
+++ b/tests/script-based-pre/cargo_no_assert_overrides/no-assert-overrides.sh
@@ -2,7 +2,31 @@
 # Copyright Kani Contributors
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
+set -euo pipefail
+
 echo "[without flag]"
-cargo kani 2>&1 | grep -m1 'E0659'
+# Kani's macro-override injection makes `assert!` ambiguous, so `cargo kani`
+# must fail with rustc's E0659 (GlobVsOuter). Capture the status explicitly:
+# the previous `cargo kani ... | grep E0659` pipeline reported grep's status,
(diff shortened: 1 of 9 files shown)


tests/script-based-pre/cargo_no_assert_overrides/no-assert-overrides.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# Copyright Kani Contributors
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+echo "[without flag]"
+cargo kani 2>&1 | grep -m1 'E0659'
+echo "[with flag]"
+cargo kani --no-assert-overrides
```
feliperodri, line 32: Wouldn't this produce false positives? This only checks that `E0659` appears somewhere in output, not that the *first run actually failed for the intended ambiguity*. If output format shifts, or another E0659 appears, the test can pass for the wrong reason, right? Also, there’s no explicit exit-status assertion for the first command. So maybe we could:
1. non-zero exit from `cargo kani` without the flag, and
2. output contains the specific ambiguity text (`assert is ambiguous` / `GlobVsOuter` marker if stable).
3. Use `set -euo pipefail` for script reliability.

feliperodri (approved):

PR #4688, opened 2026-07-28
Do not fail on crates that forbid(unstable_features)

tautschnig:
### Description

Kani injects `feature(register_tool)` via `-Zcrate-attr` into every crate it compiles, so crates that `forbid(unstable_features)` fail to build under Kani with "error: use of an unstable feature" — through no fault of their own. rustls is a prominent example: it applies the forbid unless benchmarking/coverage cfgs are set, so `cargo kani` on rustls fails immediately.

Downgrade the lint with `--force-warn unstable_features`, which overrides even a `forbid` in the crate source. The injected feature gate is a Kani implementation detail, not something the crate author can or should fix, so downgrading unconditionally is the right semantics (the warning remains visible).

Found while evaluating `kani autoharness` on the top-100 crates.io crates; verified rustls-0.23.42 (previously failing) now compiles under `cargo kani --only-codegen`.

### Testing

New regression test `tests/kani/ForbidUnstableFeatures/forbid_unstable_features.rs`: a crate with `#![forbid(unstable_features)]` and a proof harness now verifies successfully. `kani-driver` unit tests and formatting pass.

### DIFF
--- a/kani-driver/src/call_single_file.rs
+++ b/kani-driver/src/call_single_file.rs
@@ -259,6 +259,12 @@ pub fn base_rustc_flags(lib_config: LibConfig) -> Vec<RustcArg> {
         "crate-attr=feature(register_tool)",
         "-Z",
         "crate-attr=register_tool(kanitool)",
+        // Kani injects unstable features (`register_tool` above) into every crate it compiles,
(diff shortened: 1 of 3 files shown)


feliperodri (approved):
We should update the test case `ui/ice-size-overflow/large_array.rs`

PR #4695, opened 2026-07-29
Upgrade Rust toolchain to nightly-2026-02-15

tautschnig:
Two rustc changes needed handling:

1. `#[rustc_diagnostic_item]` may no longer be applied to statics. Kani tagged the memory-initialization shadow-memory static (`MEM_INIT_STATE`) with one so the compiler could locate it for function-contract `assigns` clauses. Tag it with Kani's own `kanitool::fn_marker` attribute instead (which is allowed on statics), and locate it in `codegen_contract` by scanning the kani (or core) crate's statics for that marker, using the static's mangled name to look it up in the goto symbol table.

2. Some uninhabited enums are now given a `Variants::Empty` layout that reaches `codegen_get_discriminant`, which previously hit an `unreachable!`. Reading the discriminant of an uninhabited (variant-less) enum is dead code; mirror the rustc SSA backend (which returns a poison value) by returning a nondet value of the target type.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/contract.rs
@@ -136,26 +136,36 @@ impl GotocCtx<'_, '_> {
             .typ
             .clone();
 
-        let shadow_memory_assign = self
-            .tcx
-            .all_diagnostic_items(())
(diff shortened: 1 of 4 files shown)


feliperodri (approved):
🚢 it

PR #4697, opened 2026-07-29
Implement Arbitrary for Rc<T> and Arc<T>

tautschnig:
### Description

`Box<T>` implements `Arbitrary`, but `Rc<T>` and `Arc<T>` did not, so functions taking reference-counted arguments could not be verified against nondeterministic inputs — and were skipped by `kani autoharness` with "Missing Arbitrary implementation" (smart-pointer receivers are among the largest skip classes in the top-100 crates.io evaluation). This PR adds the analogous implementations.

Note that unlike slice/container arguments (#4691/#4693), these need no bound and no opt-in flag: a smart pointer to `T` covers exactly the values of `T`, so the generated values retain Kani's usual full-coverage guarantee.

A follow-up will extend autoharness to smart pointers around types that only *can-derive* `Arbitrary` (compiler-synthesized implementations); that requires compiler-side models that depend on `alloc` and hence some optional-model plumbing for the `no_core` flow.

### Testing

New test `tests/kani/Arbitrary/rc_arc.rs` with cover checks proving extreme values, specific values, and nested smart pointers (`Rc<Arc<u8>>`) are all generated (all SATISFIED). The `Arbitrary` suite and `kani` library unit/doc tests pass; verified via autoharness that `Rc<T>`/`Arc<T>`-taking functions are now selected and verified.

Towards #3832

### DIFF
--- a/library/kani/src/arbitrary.rs
+++ b/library/kani/src/arbitrary.rs
@@ -15,6 +15,24 @@ where
     }
 }
 
+impl<T> Arbitrary for std::rc::Rc<T>
+where
+    T: Arbitrary,
+{
+    fn any() -> Self {
+        std::rc::Rc::new(T::any())
+    }
+}
+
+impl<T> Arbitrary for std::sync::Arc<T>
+where
+    T: Arbitrary,
+{
+    fn any() -> Self {
+        std::sync::Arc::new(T::any())
+    }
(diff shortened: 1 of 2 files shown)


feliperodri (approved):
🚢 it!

PR #4700, opened 2026-07-29
Upgrade Rust toolchain to nightly-2026-02-18

tautschnig:
Advance from nightly-2026-02-15 to nightly-2026-02-18. One source change is required (first needed at nightly-2026-02-16); 02-17 and 02-18 then pass with no further changes.

nightly-2026-02-16 changed the Rust allocation shims (`__rust_alloc`, `__rust_alloc_zeroed`, `__rust_dealloc`, `__rust_realloc`) to take their alignment argument as `core::ptr::Alignment` instead of `usize`.

`Alignment` is `repr(transparent)` over a `repr(usize)` enum (ABI-identical to `usize`), but Kani codegens it as a struct, whose goto type does not match the `size_t` parameters of the C definitions in `kani_lib.c`. At link time goto-cc reported a parameter type mismatch on `__rust_alloc::align`, leaving the allocator bodies unlinked. The allocator then havocked and could "return null", making the out-of-memory path (`handle_alloc_error` -> `__rust_alloc_error_handler`, an unsupported foreign function, plus `caller_location`) reachable and failing verification for every harness that allocates.

Represent `core::ptr::Alignment` as `size_t` in foreign-function signatures so the extern declarations match `kani_lib.c`, and, since the argument is codegen'd as a struct, reinterpret it with `transmute_to` (rather than `cast_to`, which only supports scalars) when passing it to the C shim. The allocator bodies link again, so allocation is modeled as succeeding and the OOM path is unreachable.

## Additional change: hook `core::str::slice_error_fail` as a panic function

Beyond the toolchain compatibility fixes above, **this PR also changes how Kani codegens `core::str::slice_error_fail`**, the diverging function that produces the panic for out-of-bounds/non-boundary string slicing:

- **Why:** nightly-2026-02-16 rewrote this function's panic-message formatting to use `floor_char_boundary`/`ceil_char_boundary`, whose loops are bounded by the index *value* rather than UTF-8's 4-byte width. Kani codegens this (never-taken-in-verification-success) formatting code into every string-slicing call site, and CBMC then unwinds those loops at each of them. On perf/smol_str this exploded verification from 73s/2.3GB to out-of-memory on a 30GB machine — the cause of the perf CI job consistently OOM-killing its 16GB runner.
- **What:** `slice_error_fail` is added to Kani's panic hook, alongside the `panic`/`panic_fmt`/`begin_panic` lang items. It is not a lang item, so it is matched by its symbol name (precedent: the `exchange_malloc` hook). Kani already models every panic as a verification failure (`assert false`) and never renders panic messages at runtime, so slicing failures are still detected exactly as before — the only effect is that the dead message-formatting code is no longer analyzed.
- **Effect:** perf/smol_str verifies in 32s/0.8GB, faster than before the toolchain change. Any harness slicing strings gets the same improvement.

## Additional change: speed up the shadow/unsupported_num_objects test

The nightly's new MIR for `Vec::push` also makes CBMC's symbolic execution of this test several times slower, threatening the per-test timeout on slow CI runners. The test's object-creation loop now allocates directly via `std::alloc::alloc` (leaked, never deallocated, no container bookkeeping to execute symbolically). The object-count thresholds move to 1021/1022, still tightly bracketing Kani's 1024-object shadow-memory limit, and the test runs several times faster than the unmodified version.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/foreign_function.rs
@@ -109,7 +109,20 @@ impl GotocCtx<'_, '_> {
             .unwrap()
             .iter()
             .zip(args)
-            .map(|(param, arg)| arg.cast_to(param.typ().clone()))
+            .map(|(param, arg)| {
+                let param_typ = param.typ().clone();
+                // `cast_to` only handles scalar conversions. When an aggregate is
+                // involved — e.g. `core::ptr::Alignment` (a struct) passed where the C
+                // shim in `kani_lib.c` declares `size_t` — reinterpret the bits with
+                // `transmute_to` instead (the two are ABI-identical).
+                if arg.typ().is_scalar() && param_typ.is_scalar() {
+                    arg.cast_to(param_typ)
+                } else if *arg.typ() == param_typ {
+                    arg
+                } else {
+                    arg.transmute_to(param_typ, &self.symbol_table)
+                }
+            })
             .collect::<Vec<_>>();
         let call_expr = fn_expr.call(expected_args);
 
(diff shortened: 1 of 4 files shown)


feliperodri (approved):

feliperodri:
@tautschnig another performance hick up...

PR #4710, opened 2026-08-03
Do not assert dependency contracts for calls made by contract clauses

tautschnig:
Stacked on #4709 (first commit; review only the last commit here).

Contracts of dependencies are asserted by default (#3802) as an aid for detecting API misuse in *user code*. Calls made while evaluating *contract clauses*, however, are specification-level plumbing: clause expressions compute a predicate over pre-/post-states, and the functions they call are best executed with their exact semantics. Re-asserting dependency contracts inside every clause evaluation multiplies verification cost on contract-dense code (clauses in model-checking/verify-rust-std routinely call contracted functions such as `NonNull::as_ptr`, paying the assert-closure overhead per clause instance) without checking any user code.

This PR extends the clause-context dispatch introduced in #4709 from check modes to assert mode: calls to a contracted dependency occurring during clause evaluation execute the original body (mode 0). The body remains fully inlined and UB-checked; only the requires/ensures assertions of the callee's contract are skipped in this context — a genuine UB-causing misuse inside a clause is still caught by the UB checks themselves.

The regression test covers both halves: a clause calling a contracted function with precondition-violating (but well-defined) arguments verifies successfully, while the same misuse in user code still fails on the asserted precondition.

Measurements on verify-rust-std (Kani pin 152c6a8c + CBMC 6.10.0, contracts asserted): semantics-driven change with modest performance impact on sampled harnesses after the fixes in verify-rust-std#622/#623/#624 landed (`check_to_bytes` and `non_null_check_read` within noise; `num::nonzero::verify::check_mul_i128_edge_pos` improved from 29.4s to 8.1s measured against the pre-fix baseline, attribution shared with verify-rust-std#624). The full `expected/function-contract` suite passes with `--force-rerun` (112/112).


feliperodri (approved):
LGTM, but orrectness of assert-mode suppression in regular proofs depends on the `__CPROVER_`-prefixed depth counter never being havocked (so it stays 0 without an explicit reset). This holds in the current configuration and is covered by the regression test, but if a future verification config ever havocked such statics, regular proofs could under-assert dependency contracts. This is documented in `kani_core`, so I left it as-is rather than adding a redundant reset to every `#[kani::proof]`.

PR #4711, opened 2026-08-03
Fix ICE on non-literal `cover`/`assert`/`check` message expressions

ivmat:
## Problem

`kani-compiler`'s codegen hooks for `kani::cover`, `kani::assert`, `kani::check` and the internal
safety-check/unsupported-check hooks all call `gcx.extract_const_message(&msg).unwrap()` to recover the
message string.

`extract_const_message` returns `None` whenever the message operand does not codegen down to a
string-literal constant — for example when it is a function parameter — so the `.unwrap()` produces an
**internal compiler error** rather than a normal diagnostic.

Reproducer (`kani::cover` case):

```rust
#[kani::proof]
fn main() {
    let msg = "not a literal";
    kani::cover!(true, msg);
}
```

This still reproduces on `main`: all six `extract_const_message(&msg).unwrap()` call sites are present
in `kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs`.

## Fix

Add `extract_msg_or_err`, which mirrors the existing `utils::span_err` + `abort_if_errors` pattern
already used by neighbouring intrinsic codegen in the same file, and use it at all six call sites
(`Cover`, `Assert`, `UnsupportedCheck`, `SafetyCheck`, `SafetyCheckNoAssume`, `Check`).

Each now emits a spanned ``` `<construct>` message must be a string literal ``` error and aborts
compilation cleanly instead of panicking.

## Tests

UI regression tests modelled on the existing `tests/ui/ice-size-overflow` test (a prior
"ICE → clean error" regression test):

- `tests/ui/cover-non-literal-message/`
- `tests/ui/assert-non-literal-message/`

`kani::check` is `pub(crate)` in `library/kani_core/src/lib.rs` with no public re-export, so user code
cannot invoke it directly (it fails earlier with `E0425`). The remaining hooks
(`safety_check`, `safety_check_no_assume`, `unsupported_check`) are compiler-generated and not
reachable from user code either, so neither is given a UI test.

## Testing performed

- `cargo build -p kani-compiler` and `cargo check -p kani-compiler` on this branch — pass.
- Full local regression via compiletest (`--force-rerun`, all suites + unit tests): **1472 passed,
  3 failed — all 3 reproduce identically on a clean `main` checkout (`2b7972b7c`)**, so they are
  pre-existing environment issues (one requires z3, which is not installed here), not caused by this
  change.
- `cargo clippy` on the pinned nightly — clean.
- `scripts/kani-fmt.sh --check` and the copyright/format checks — clean.
- The two new UI tests fail as expected when the fix is reverted (negative control).

Happy to adjust the diagnostic wording or the test placement if you'd prefer something different.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
@@ -79,7 +79,12 @@ fn extract_msg_or_err(gcx: &GotocCtx, msg_expr: &Expr, span: Span, construct: &s
     gcx.extract_const_message(msg_expr).unwrap_or_else(|| {
         utils::span_err(gcx.tcx, span, format!("`{construct}` message must be a string literal"));
         gcx.tcx.dcx().abort_if_errors();
-        unreachable!("Rustc should have aborted already")
+        // `span_err` above emits a hard error, which `abort_if_errors` is guaranteed to
+        // observe and turn into a fatal error, unwinding before we get here.
+        unreachable!(
+            "rustc should have aborted after the `{construct}` message error emitted above; \
+             reaching this point means that error was never counted by the diagnostic context"
+        )
     })
 }
 
--- a/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
@@ -79,7 +79,7 @@ fn extract_msg_or_err(gcx: &GotocCtx, msg_expr: &Expr, span: Span, construct: &s
     gcx.extract_const_message(msg_expr).unwrap_or_else(|| {
         utils::span_err(gcx.tcx, span, format!("`{construct}` message must be a string literal"));
         gcx.tcx.dcx().abort_if_errors();
-        String::new()
+        unreachable!("Rustc should have aborted already")
     })
 }
 
--- a/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
@@ -62,6 +62,27 @@ struct Cover;
 
 const UNEXPECTED_CALL: &str = "Hooks from kani library handled as a map";
 
+/// Extracts a string message from a constant `Expr` produced by one of Kani's
+/// hooks (e.g. `kani::cover`, `kani::assert`, `kani::check`, ...).
+///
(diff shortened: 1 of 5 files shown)


tests/ui/cover-non-literal-message/main.rs
```diff
@@ -0,0 +1,21 @@
...
+//! function parameter rather than a literal expression at the call site).
+
+fn cover_with_msg(cond: bool, msg: &'static str) {
+    kani::cover(cond, msg);
+}
+
+#[kani::proof]
+fn main() {
+    cover_with_msg(true, "not actually a literal at the call site");
+}
```
feliperodri, line 25: These new UI tests are likely too easy to inline away. Both `cover-non-literal-message` and `assert-non-literal-message` use tiny helper functions. Copilot noted these helpers may be inlined, which could turn the call back into a literal at the call site and stop exercising the intended non-literal-message path. That would weaken the regression test and make it flaky across optimization settings/toolchain changes. A stronger test would prevent inlining explicitly or use a setup that preserves the non-literal expression form.

kani-compiler/src/codegen_cprover_gotoc/overrides/hooks.rs
```diff
@@ -62,6 +62,27 @@ struct Cover;
...
+/// spanned, user-facing error and abort compilation.
+///
+/// `construct` should be the user-facing name of the Kani construct being
+/// codegen'd (e.g. `"cover"`, `"assert"`), and is only used for the error
+/// message.
+fn extract_msg_or_err(gcx: &GotocCtx, msg_expr: &Expr, span: Span, construct: &str) -> String {
+    gcx.extract_const_message(msg_expr).unwrap_or_else(|| {
+        utils::span_err(gcx.tcx, span, format!("`{construct}` message must be a string literal"));
+        gcx.tcx.dcx().abort_if_errors();
+        unreachable!("Rustc should have aborted already")
```
feliperodri: This message could be way more descriptive to tell the user when it sees this message *why* rustc should have aborted already.

feliperodri (changes requested):

feliperodri:
@ivmat could you also tackle the two suppressed Copliot comments? Almost there.

ivmat:
> @ivmat could you also tackle the two suppressed Copliot comments? Almost there.

@feliperodri i think i covered everyting now. i also could not rerrun all same CI tests locally but i dont think there should be any issues

feliperodri (approved):

PR #4719, opened 2026-08-05
Fail verification when the solver backend drops quantifiers

tautschnig:
### Description

CBMC's SAT-based backends only support quantifiers with constant bounds (`boolbv_quantifier.cpp`'s eager instantiation). A `kani::forall!`/`kani::exists!` with a symbolic bound reaches the backend's quantifier post-processing, which has no handling: the expression is replaced with unconstrained values, and CBMC reports only a low-visibility `warning: ignoring forall` among its status messages — which Kani currently swallows entirely.

The consequences are severe for usability:
- a `kani::assume` containing such a quantifier is silently **not enforced**: the harness may report `VERIFICATION:- SUCCESSFUL` while covering none of the intended property (found while building quantified element-validity assumptions for autoharness slice generation, where the vacuous assume was only caught by a non-tautological probe);
- a `kani::assert` containing one may fail spuriously.

This PR detects CBMC's ignoring-quantifier messages in the output parser, counts them on `VerificationResult`, and renders a prominent warning after the result (on both successful and failed outcomes), explaining the effect and suggesting `#[kani::solver(z3)]`, which handles these quantifiers (verified: the same harnesses verify correctly, fast, under z3).

A proper fix (quantifier instantiation in CBMC's SAT backend) is in progress upstream; this warning closes the silent-degradation window until it lands.

### Testing

New expected test pins the dangerous case: a harness that succeeds only because its assertion does not depend on the (unenforced) quantified assumption, with the warning attached. Verified no-warning behavior for constant-bound quantifiers (eagerly instantiated) and under `--solver z3`. All quantifier expected tests and kani-driver unit tests pass.


feliperodri:
Behavior change (experimental `-Z quantifiers`): Previously, when the SAT solver backend dropped a quantifier with non-constant bounds, Kani printed a warning but still reported SUCCESSFUL. Because a `kani::assume` containing such a quantifier is silently not enforced, that "success" could be vacuous (an unsound false negative). Kani now reports `VERIFICATION:- FAILED` in this case and directs users to an SMT backend that supports quantifiers (e.g., `#[kani::solver(z3)]`). Proofs that relied on the old warn-and-pass behavior will now fail — **correctly**.

@tautschnig do you agree to merge it with this new behavior?

feliperodri (approved):

tautschnig:
> @tautschnig do you agree to merge it with this new behavior?

Yes, I think that's the right behaviour.

Issue #4729, opened 2026-08-12
`--fail-fast` discards results for harnesses that already completed


feliperodri:
When `--fail-fast` stops a run, the results of harnesses that already finished are discarded. Kani reports only the failing harness, so its own summary contradicts the output printed just above it.

### Reproduce

```rust
// ff.rs
#[kani::proof]
fn aaa_passes() {
    let x: u8 = kani::any();
    kani::assume(x < 10);
    assert!(x < 20);
}

#[kani::proof]
fn bbb_fails() {
    let x: u8 = kani::any();
    assert!(x < 20);
}

#[kani::proof]
fn ccc_passes() {
    let x: u8 = kani::any();
    kani::assume(x < 5);
    assert!(x < 20);
}
```

```
$ kani ff.rs --fail-fast
Checking harness ccc_passes...
VERIFICATION:- SUCCESSFUL
Checking harness bbb_fails...
VERIFICATION:- FAILED
Verification failed for - bbb_fails
Complete - 0 successfully verified harnesses, 1 failures, 1 total.
```

`ccc_passes` ran and was reported `SUCCESSFUL` two lines earlier, but the summary says **0 successfully verified harnesses** and **1 total**.

### Expected

The summary should account for the harnesses that actually ran: 1 successful, 1 failed, 2 executed out of 3 selected. Whether the remaining unrun harnesses are reported as skipped is a separate question; the ones that completed should not vanish.

### Root cause

In `kani-driver/src/harness_runner.rs`, `check_all_harnesses` collects with `collect::<Result<Vec<_>>>()`, which short-circuits on the first `Err` and drops the `Ok` results already produced. The fail-fast branch then rebuilds a vector containing only the failing harness:

```rust
if err.is::<FailFastHarnessInfo>() {
    let failed = err.downcast::<FailFastHarnessInfo>().unwrap();
    Ok(vec![HarnessResult {
        harness: sorted_harnesses[failed.index_to_failing_harness],
        result: failed.result,
    }])
}
```

That vector is what feeds `print_final_summary`, which is why the counts are wrong.

### Impact

Two places, and the second is new:

1. **The rendered summary** is self-contradicting, as above. A user who runs with `--fail-fast` cannot tell from the summary how much of the suite got through before the failure.
[...]

Issue #4731, opened 2026-08-12
`--export-json`: a failed, empty, or partial run can serialize as a clean pass


ivmat:
`--export-json` is a `-Z`-gated, currently-undocumented output. Using it as a CI pass/fail
oracle in a standard-library verification effort, we hit several cases where the exported file
does not faithfully represent the run: a run that **failed, was skipped, matched nothing, or
wasn't fully measured can be read as a clean, complete pass** by a consumer that trusts the file.

These are **reporting-fidelity** issues in an unstable output — none changes a verification
verdict, a property status, or Kani's exit code. Grouped below by how each is best fixed.

### Driver bugs — pre-existing (predate #4472), surfaced but not caused by `--export-json`; tracked as two small standalone PRs

**A. A filter that matches zero harnesses runs to a `completed / 0 failed` export.**
`determine_targets` only bails on a missing filter under `--exact` (`metadata.rs:112`); the
generic "no harnesses matched" bail lives in `print_final_summary` (`harness_runner.rs:302`),
which runs *after* the export is written — so a filter typo yields `Ok([])`, a clean-looking
export, and exit 0. Same area, independent of `--export-json`: `print_final_summary`
early-returns under `--quiet` (`harness_runner.rs:255`) *before* that bail, so
`kani --quiet --harness <typo>` skips the "no harnesses matched" error and exits 0.
Fix in `determine_targets`, before codegen and before any export.

**B. `--fail-fast` drops the harnesses it skipped (#4729).**
Root cause: `collect::<Result<Vec<_>>>()` (`harness_runner.rs:105`) short-circuits on the first
`Err` and discards already-completed `Ok` results. The lost harnesses are then absent from the
per-harness details, and the plain-text summary and SARIF under-report too — not only the JSON.
Fix the collect (#4729), not the reporter.

### Deferred to RFC #4727

- **Stale export is never invalidated.** A run that reuses an existing `--export-json` path and
  dies before the write leaves the previous complete-looking file in place. The completeness
  mechanism (atomic temp-file + rename, "file exists ⇒ complete") is a schema/design decision
  specified in RFC #4727, not a writer patch.
- **No leaf-value / semantic validation of the exported file.** Pending the RFC's open question
  on shipping a JSON Schema (`schemars`) vs. a homegrown validator.
- **`cbmc_stats` parsing is brittle** (exact `strip_prefix`/`strip_suffix`; a malformed duplicate
[...]

PR #4732, opened 2026-08-12
Fix `--export-json` reporting a failed, skipped, or zero-match run as clean

ivmat:
Follow-up to #4472. While using `--export-json` as a CI pass/fail oracle I hit several cases where the exported file did not faithfully represent the run — a run that failed, was skipped, matched nothing, or wasn't fully measured could be read as a clean, complete pass by a consumer that trusts the file (which is the point of the file). This addresses them:

- **Stale exports.** `verify_project` now writes a `run_state:"incomplete"` marker to the target *before* verification, promoted to `"complete"` only when every selected harness produced a result — so a run that dies before finishing can't leave a previous run's clean file to be read as this run's result.
- **Zero / unmatched harnesses.** A filter that matches nothing used to write a `completed / 0-failed` document (the "no harnesses matched" error is only raised after the export). It's now marked `run_state:"no_harnesses_selected"`, with a `harness_selection` block recording the requested filters, matched count, and any unmatched filters.
- **`--fail-fast`.** Harnesses skipped after the first failure kept their `harness_metadata` entry but had no `error_details`/`property_details` entry, so a consumer correlating the arrays could read absence as success. Every selected harness now gets an explicit entry, and `run_state` is `"partial"` when some selected harnesses didn't run.
- **Attribution.** Per-harness details join on the unique `mangled_name` rather than `pretty_name`, which two crates in one workspace can share.
- **`cbmc_stats`.** Parsing no longer clobbers a valid value with `null` on a later malformed line, tolerates harmless CBMC wording changes, rejects a duration reported in the wrong unit rather than mis-scaling it, and widens counts to `u64`.
- **Validator.** `scripts/validate_json_export.py` previously checked structure only, so `failed:-1` or `successful:"yes"` passed. It now checks leaf value types, count reconciliation, and the `run_state` enum and its invariants, with a `validator-negative` test suite; `run_state`/`harness_selection` are added to the schema template so their presence is required.

Rationale and the per-case scenarios are in #4731. One case from that report — a partial, non-timeout CBMC exit — is left as follow-up pending confirmation, and is not addressed here.

**Testing:** `cargo test -p kani-driver` (incl. new `call_cbmc`/validator unit tests), `clippy -D warnings`, and `fmt` are clean; the `validator-negative` suite rejects the malformed fixtures. The compiletest `tests/json-handler/*` integration tests (which require CBMC) were not run in my environment, so the new validator leaf-type checks have been exercised against synthetic and reference fixtures but not yet a live CBMC-generated export.

### DIFF
--- a/kani-driver/src/call_cbmc.rs
+++ b/kani-driver/src/call_cbmc.rs
@@ -37,10 +37,14 @@ pub struct CbmcInfo {
 #[derive(Debug, Clone, Default)]
 pub struct CbmcStats {
     pub runtime_symex_s: Option<f64>,
-    pub size_program_expression: Option<u32>,
-    pub slicing_removed_assignments: Option<u32>,
-    pub vccs_generated: Option<u32>,
-    pub vccs_remaining: Option<u32>,
+    // `u64`, not `u32`: these are unbounded counts scraped from CBMC's own text output (program
+    // expression size, VCCs, sliced assignments), and a sufficiently large real run overflowing
+    // `u32` used to collapse silently to `null` via `parse::<u32>().ok()` -- indistinguishable
+    // from "not measured".
+    pub size_program_expression: Option<u64>,
+    pub slicing_removed_assignments: Option<u64>,
+    pub vccs_generated: Option<u64>,
+    pub vccs_remaining: Option<u64>,
     pub runtime_postprocess_equation_s: Option<f64>,
     pub runtime_convert_ssa_s: Option<f64>,
     pub runtime_post_process_s: Option<f64>,
@@ -107,6 +111,12 @@ fn merge_cbmc_stats(items: &[ParserItem]) -> Option<CbmcStats> {
 /// Record the statistic a single CBMC status message carries, if it carries one. Later messages win,
 /// matching CBMC's own behaviour of reporting a running figure more than once.
 /// Returns whether this message was recognized.
+///
+/// Every field assignment below goes through [`record_stat`], which only overwrites a field when
+/// parsing succeeds. Without that, a later message that merely *resembles* a recognized label but
+/// fails to parse (a wording tweak, an unexpected unit, a truncated line) would silently erase an
+/// already-recorded valid measurement by assigning it `None` -- indistinguishable from "never
+/// measured" to a consumer of the export.
 fn record_cbmc_stat(message: &str, stats: &mut CbmcStats) -> bool {
     // "Generated 1 VCC(s), 1 remaining after simplification"
     if let Some(counts) = message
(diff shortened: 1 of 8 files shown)


feliperodri (changes requested):
This PR contradicts its own RFC #4727... We shouldn't merge it until RFC 0015 lands with my requested changes. It's a one-way door and it's currently being decided in follow-up PRs rather than the RFC itself. I added all my concerns to https://github.com/model-checking/kani/issues/4731.
  
The PR adds `run_state ∈ {incomplete, no_harnesses_selected, partial, complete}` and a `harness_selection` block to the #4472 schema. #4727 specifies a different vocabulary for the same information: `schema_version`, `outcome.kind`, `outcome.verdict`, `failure_kind`... and a different mechanism for the stale-file problem: `temp-file-plus-rename` so that "file exists implies complete," plus the up-front delete. That's the paragraph I called the best part of the RFC... This PR implements an in-band "incomplete" marker instead, written non-atomically.

The real bugs are in the wrong layer. Three of the four behavioral fixes patch the JSON writer instead of the thing that's broken:
1. Zero-match should fail in determine_targets, before codegen, before any export. And there's a bigger bug hiding behind it: `print_final_summary` early-returns on quiet before reaching the bail, so reading the code, `kani --quiet --harness` typo skips the "no harnesses matched" error entirely and exits 0. No test covers that combination. That affects every user and every CI that uses the exit code, not just JSON consumers and fixing it in `determine_targets` makes issue item 2 disappear for free. (I did not build Kani to execute this; it's a code-path reading of harness_runner.rs:265 and metadata.rs:112.)
2. `--fail-fast` — root cause is `collect::<Result<Vec<_>>>()` short-circuiting and discarding already-completed Ok results (harness_runner.rs:105), see #4729. This PR doesn't fix it; it labels the lost harnesses `has_errors: true`, `error_type: "not_reported"` asserting an error for a harness that may have passed. That trades one false statement for another, and the text summary and SARIF still under-report. Fix #4729 and item 3 is moot.
4. `harness_id` — the PR moves the join predicate to `mangled_name` but explicitly keeps emitting `harness_id: pretty_name`. The consumer-facing ambiguity it was filed to fix is still there.

**ALSO... maintenance...** 98 added comment lines to 125 added Rust lines. The comments narrate reasoning in first person in the tree. "I should note this was partly self-inflicted", "the honest, disjunctive truth", multi-paragraph justifications above three-line functions (likely AI slop). Some of that is already in main from #4472's rescue commits. It will rot the moment 0015 reshapes the schema, and it's the kind of prose a future contributor won't dare touch.

ivmat:
closing this PR, we go with plan from #4731. you are right,  this PR is incomplete, sloppy and not aligned with atomic write design.

PR #4743, opened 2026-08-18
Fail a zero-match harness filter before codegen and export

ivmat:
A `--harness` filter that matches no harness only fails in the end-of-run summary. `--quiet` skips that summary, so the run exits 0. With `--export-json`, the run also writes a file that reports a clean, empty run.

This PR moves the zero-match check into `determine_targets`, before codegen and before any export. The error is built by one shared helper, used by both the pre-codegen check and the summary-stage guard, and its wording is unified with the `--exact` message. The summary check stays as a last guard for paths that skip harness filtering.

Testing: a new script-based test covers the `--quiet` exit code, the absent export file, and the several-filters message. The test fails on `main` and passes with this change. `cargo test -p kani-driver`, `rustfmt`, and `clippy` are clean.

First of the two driver fixes agreed in #4731.

### DIFF
--- a/kani-driver/src/metadata.rs
+++ b/kani-driver/src/metadata.rs
@@ -1,7 +1,7 @@
 // Copyright Kani Contributors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-use anyhow::{Result, bail};
+use anyhow::Result;
 use std::path::Path;
 
 use kani_metadata::{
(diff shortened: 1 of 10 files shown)


feliperodri (approved):

kani-driver/src/metadata.rs
```diff
@@ -127,6 +127,16 @@ impl KaniSession {
             );
         }
 
+        // A `--harness` filter that matches nothing must fail here
+        if !harness_filters.is_empty() && compiler_filtered_harnesses.is_empty() {
+            match self.args.harnesses.as_slice() {
```
feliperodri: This block is identical to the one in `harness_runner.rs` (~347-351). Consider extracting a shared helper so the two zero-match messages cannot drift.

kani-driver/src/metadata.rs
```diff
@@ -127,6 +127,16 @@ impl KaniSession {
...
+                [harness] => bail!("no harnesses matched the harness filter: `{harness}`"),
```
feliperodri: This wording differs from the `--exact` zero-match message just above (`Failed to match the following harness(es)`). Two different messages for the same "filter matched nothing" situation; worth unifying.

kani-driver/src/metadata.rs
```diff
@@ -127,6 +127,16 @@ impl KaniSession {
             );
         }
 
+        // A `--harness` filter that matches nothing must fail here
+        if !harness_filters.is_empty() && compiler_filtered_harnesses.is_empty() {
```
feliperodri, line 124: `determine_targets` runs on every `verify_project`, so this makes the summary-stage check in `harness_runner.rs` unreachable for the filter case. A short comment there noting it now only guards non-filtered paths would help.

kani-driver/src/metadata.rs
ivmat: done

kani-driver/src/metadata.rs
ivmat: done

PR #4744, opened 2026-08-18
Keep completed results when --fail-fast aborts a run

### LINKED ISSUE #4729: `--fail-fast` discards results for harnesses that already completed

ivmat:
`check_all_harnesses` collected results with `collect::<Result<Vec<_>>>()`, which short-circuits on the first error. A `--fail-fast` abort was such an error, so every result that had already completed was dropped. The summary then contradicted the per-harness output above it, and `--export-json` under-reported the run the same way.

Completed results now accumulate in a shared vector as harnesses finish. The abort signal carries no payload; the failing harness records its result like any other. Results are re-sorted into harness order after the parallel loop, since completion order is nondeterministic.

The parallel fail-fast UI test pinned the old behavior: "1 failures, 1 total" for ten failing harnesses under `--jobs 4`. With completed results retained, those counts depend on thread scheduling, so the test becomes a script-based test asserting the stable properties: the run aborts early (fewer than ten run) and every counted harness is a failure. The sequential UI test is unchanged: it aborts on its first harness, so its pinned summary stays correct. A new sequential script-based test proves retention deterministically ("1 successfully verified, 1 failures, 2 total"); it fails on `main`.

Testing: `cargo test -p kani-driver`, `rustfmt`, and `clippy` are clean; the two new script-based tests and the existing `stop_at_single_fail` UI test pass.

Resolves #4729

### DIFF
--- a/kani-driver/src/harness_runner.rs
+++ b/kani-driver/src/harness_runner.rs
@@ -38,6 +38,52 @@ pub(crate) struct HarnessResult<'pr> {
     pub result: VerificationResult,
 }
 
+/// The outcome of one unit of work run by [`run_until_abort`].
+struct Completed<T> {
+    payload: T,
+    /// Whether this outcome should stop units that have not started yet.
+    aborts: bool,
+}
+
+/// Run `run_one` over `items` in parallel, keeping the payloads of units that completed even when
+/// one of them aborts the run, and returning them in `items` order together with whether an abort
+/// happened.
+///
+/// The abort is latched rather than raised through the `Err` channel, which stays reserved for
+/// genuine errors. `try_for_each` surfaces only one `Err`, so signalling an abort that way would
+/// let it displace a real failure raised by a sibling unit still in flight.
+fn run_until_abort<I: Sync, T: Send>(
+    items: &[I],
+    run_one: impl Fn(&I) -> Result<Completed<T>> + Sync,
+) -> Result<(Vec<T>, bool)> {
+    let completed: Mutex<Vec<(usize, T)>> = Mutex::new(Vec::new());
+    let aborted = AtomicBool::new(false);
+
+    items.par_iter().enumerate().try_for_each(|(idx, item)| -> Result<()> {
+        // A unit that has not started returns without running once an abort is latched. Units
+        // already in flight finish and record their payloads.
+        if aborted.load(Ordering::Relaxed) {
+            return Ok(());
+        }
+
+        let Completed { payload, aborts } = run_one(item)?;
+        completed.lock().unwrap().push((idx, payload));
+
+        if aborts {
+            aborted.store(true, Ordering::Relaxed);
+        }
+        Ok(())
+    })?;
+
+    // Completion order under parallelism is nondeterministic; restore `items` order.
+    let mut collected = completed.into_inner().unwrap();
+    collected.sort_by_key(|(idx, _)| *idx);
+    let payloads = collected.into_iter().map(|(_, payload)| payload).collect();
+
+    // Read after the parallel region has joined, which orders every store before this load.
+    Ok((payloads, aborted.load(Ordering::Relaxed)))
+}
+
 impl<'pr> HarnessRunner<'_, 'pr> {
     /// Given a [`HarnessRunner`] (to abstract over how these harnesses were generated), this runs
     /// the proof-checking process for each harness in `harnesses`.
@@ -78,23 +124,8 @@ impl<'pr> HarnessRunner<'_, 'pr> {
             builder.build()?
         };
 
-        // Completed results accumulate here so a `--fail-fast` abort keeps them.
-        let completed: Mutex<Vec<(usize, HarnessResult<'pr>)>> = Mutex::new(Vec::new());
-
-        // `--fail-fast` is signalled by this latch rather than through the `Err` channel, which
-        // stays reserved for genuine errors. `try_for_each` surfaces only one `Err`, so sharing
-        // the channel would let an abort mask a real failure raised by a sibling harness.
-        let fail_fast = AtomicBool::new(false);
-
-        let run_result = pool.install(|| -> Result<()> {
-            sorted_harnesses.par_iter().enumerate().try_for_each(|(idx, harness)| -> Result<()> {
-                // Once a harness has failed under `--fail-fast`, the remaining harnesses return
-                // without starting. Harnesses already in flight still finish and record their
-                // results.
-                if fail_fast.load(Ordering::Relaxed) {
-                    return Ok(());
-                }
-
+        let run_result = pool.install(|| {
+            run_until_abort(&sorted_harnesses, |harness| {
                 let goto_file =
                     self.project.get_harness_artifact(harness, ArtifactType::Goto).unwrap();
 
@@ -114,31 +145,18 @@ impl<'pr> HarnessRunner<'_, 'pr> {
                     progress_indicator.update_with_result(succeeded, timed_out);
                 }
 
-                let triggered =
+                let aborts =
                     self.sess.args.fail_fast && result.status == VerificationStatus::Failure;
 
-                completed.lock().unwrap().push((idx, HarnessResult { harness, result }));
-
-                if triggered {
-                    fail_fast.store(true, Ordering::Relaxed);
-                }
-                Ok(())
+                Ok(Completed { payload: HarnessResult { harness, result }, aborts })
             })
         });
 
         // Finish progress indicator
         progress_indicator.finish();
 
-        // The `Err` channel now carries genuine errors only, so any error propagates.
-        run_result?;
-
-        // Read after the parallel region has joined, which orders every store before this load.
-        let fail_fast = fail_fast.load(Ordering::Relaxed);
-
-        // Completion order under parallelism is nondeterministic; restore harness order.
-        let mut collected = completed.into_inner().unwrap();
-        collected.sort_by_key(|(idx, _)| *idx);
-        let results: Vec<HarnessResult<'pr>> = collected.into_iter().map(|(_, r)| r).collect();
+        // The `Err` channel carries genuine errors only, so any error propagates.
+        let (results, fail_fast) = run_result?;
 
         if let Some(handler) = json_handler {
             let status_label = if fail_fast { "completed_with_fail_fast" } else { "completed" };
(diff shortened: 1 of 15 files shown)


kani-driver/src/harness_runner.rs
```diff
@@ -90,75 +90,62 @@ impl<'pr> HarnessRunner<'_, 'pr> {
...
+                }
+
+                let fail_fast_triggered =
+                    self.sess.args.fail_fast && result.status == VerificationStatus::Failure;
+
+                completed.lock().unwrap().push((idx, HarnessResult { harness, result }));
+
+                // Ask rayon to stop scheduling further harnesses (best effort); harnesses
+                // already in flight still complete and record their results.
+                if fail_fast_triggered { Err(Error::new(FailFastAbort)) } else { Ok(()) }
```
feliperodri: This is the one I'd like changed. `FailFastAbort` shares rayon's `Err` channel with real errors (`instrument_model`, `check_harness`, playback). If one harness trips fail-fast while another returns a real error, `try_for_each` surfaces one of them arbitrarily — and if the abort wins, the real error is swallowed and we print a normal-looking summary with exit 1.

Pre-existing in shape, but your restructure makes the clean fix nearly free: use an `AtomicBool` latch for fail-fast, check it at the top of the closure and `return Ok(())`, and keep `Err` for genuine errors only. You lose rayon's producer-level short-circuit, but a skipped harness then costs one atomic load instead of a CBMC run, and error precedence stops being luck.

kani-driver/src/harness_runner.rs
```diff
@@ -90,75 +90,62 @@ impl<'pr> HarnessRunner<'_, 'pr> {
...
-                }
-            }
+        // A genuine error (not the fail-fast signal) must still propagate.
+        let fail_fast = match run_result {
+            Ok(()) => false,
+            Err(err) if err.is::<FailFastAbort>() => true,
+            Err(err) => return Err(err),
+        };
+
+        // Completion order under parallelism is nondeterministic; restore harness order.
```
feliperodri: The order is fixed here, but the *set* is still schedule-dependent, and that's a user-visible contract change: `--export-json` is no longer reproducible for `--fail-fast --jobs N`, so CI consumers diffing exports will see churn run to run.

I think that's the right behavior (it's the honest one), but it needs to be written down somewhere. `--fail-fast` has zero mentions in `docs/` and a one-line help string (`args/mod.rs:261`). Please add a sentence there: in-flight harnesses finish and get counted, so counts vary under `--jobs`.

tests/script-based-pre/fail_fast_keeps_completed/keeps_completed.sh
```diff
@@ -0,0 +1,82 @@
...
+    exit 1
+fi
+
+# The export must retain it as well: counters and per-harness results.
+if [[ ! -f "${EXPORT_FILE}" ]]; then
+    echo "FAIL: --export-json produced no file on a --fail-fast run"
+    exit 1
+fi
+
+python3 - "${EXPORT_FILE}" << 'EOF'
```
feliperodri: Wrong suite for this half. Every other `--export-json` test lives in `tests/json-handler/`, so nobody auditing export coverage will find this. Also, unlike its siblings there, it never runs `scripts/validate_json_export.py` — so nothing checks that the `completed_with_fail_fast` document still validates structurally.

Suggest splitting: keep the summary assertions here, move the export assertions to `tests/json-handler/fail-fast/` and run the validator.

tests/script-based-pre/fail_fast_parallel/early_abort.sh
```diff
@@ -0,0 +1,59 @@
...
+fi
+
+# At least the failing harness itself is counted.
+if [[ "${TOTAL}" -lt 1 ]]; then
+    echo "FAIL: expected at least one counted harness, got ${TOTAL}"
+    exit 1
+fi
+
+# The run must abort early: strictly fewer than all ten harnesses.
+if [[ "${TOTAL}" -ge 10 ]]; then
```
feliperodri, line 51: Flagging, not blocking: this turns a pinned test into a timing test. Realistically safe — 4 threads means ~4 chunks and `full()` is checked before each item, so `TOTAL` lands near 4 — but the bound is doing no work if scheduling ever changes.

The `VERDICTS == TOTAL` check below is the load-bearing assertion and holds under any schedule. Fine to keep the `< 10` guard, just don't count it as the regression check.

feliperodri (changes requested):
Diagnosis and fix are right, and this is the direction I suggested in #4729 — the accumulator + re-sort is exactly it. Nice catch on the sort in particular: the old `collect::<Result<Vec>>` on an indexed rayon iterator *was* order-preserving, so without it you'd have silently changed export ordering for every parallel run, not just fail-fast ones. I also checked the determinism argument for the new sequential test and it holds (`sort_harnesses_by_loc` reverses on start line, `jobs()` defaults to 1 thread), and the existing sequential UI test really is unaffected.

Blocking on the first inline comment; the rest are cheap.

One more that I can't anchor inline because the file isn't in the diff: `schema_utils.rs:362-365` says `verification_results.results` "is in completion order", which is why entries need `harness_id`. After this PR the order is deterministic. The conclusion still stands (sorted-by-loc != harness-metadata order) but the stated reason is now wrong — please reword.

Regression suites are still running, so I haven't seen an end-to-end green.

ivmat:
@feliperodri done, all items covered and retested. 

here's a summary per review item:

1. Err channel (blocking): the abort is now an AtomicBool latch in check_all_harnesses — the
   closure loads it first and returns Ok(()) if set; a fail-fast failure stores it and returns
   Ok(()); Err is reserved for instrument_model, synthesize_loop_contracts and check_harness,
   so run_result? propagates a genuine error unconditionally. FailFastAbort and its
   Error/Display impls are gone. The cost is as the review predicted: rayon no longer
   short-circuits at the producer, so remaining harnesses are visited and each pays one relaxed
   atomic load instead of a CBMC invocation; no result entry is pushed for them, so summary and
   export still count only harnesses that ran. The latch sits behind a small generic seam
   (run_until_abort) purely so the abort/error precedence is unit-tested without CBMC in the
   loop — reverting the latch makes those tests fail.

2. Schedule-dependent counts: the help string on fail_fast now states that in-flight harnesses
   finish and are counted, so counts vary under --jobs N. docs/ has no fail-fast section at all
   today (grep -rn fail-fast docs/ is empty), so none was created in this PR; if a prose
   paragraph is wanted, docs/src/usage.md under "Common command line flags" is the natural home
   and it can be added here rather than in a follow-up.

3. Test split: tests/script-based-pre/fail_fast_keeps_completed/ now asserts only the rendered
   summary. The export half moved to tests/json-handler/fail-fast/ — same two-harness fixture
   under --fail-fast --export-json, then scripts/validate_json_export.py, then the counter and
   per-harness assertions (executed 2 / successful 1 / failed 1, z_passes Success, a_fails
   Failure). It follows the multiple-harnesses layout, set -eu plus pipefail with only the kani
   invocation exempted (as failed-verification/test.sh does), so a failed validation fails the
   test — the aborted run's document now gets an enforced structural check it did not have.

4. The < 10 bound: comments in early_abort.sh relabelled — it is a smoke check that the abort
   happened at all; VERDICTS == TOTAL is the regression assertion, reproduces #4729's symptom,
   and holds under any schedule. Both checks unchanged.

5. schema_utils.rs comment: reworded — the array is built in harness-metadata order while
   verification_results.results is in sort_harnesses_by_loc (source-location) order, so
   position still does not correlate the two and harness_id remains what makes an entry
   attributable.

Also folded, from the earlier Copilot comments (never answered in-thread): the export case is
the json-handler test above; the summary-vs-verdicts comparison is the VERDICTS == TOTAL check.

One change beyond the threads: the ordering step was extracted into in_input_order and
unit-tested on a shuffled input — the sort was previously witnessed by no test, since every
runner test uses a one-thread pool where completion order is already input order.

Local regression at the pushed commit: unit tests green (87 in kani-driver) and 15 compiletest
suites with 0 failures, including json-handler.

feliperodri (approved):
Thanks for sticking with this, @ivmat — really nice work. The latch refactor came out clean, error-vs-abort precedence is now airtight, and the test coverage is excellent (that concurrent-abort-vs-genuine-error case in particular). Verified everything locally and CI is green. LGTM! 🚀

PR #4749, opened 2026-08-21
Fix compiler crash on slice-modifies verified stubs (#4748)

### LINKED ISSUE #4748: Verified function contract stubs fail compilation on slice references
Function contract stubs fail compilation when using slice references as function arguments. This is true for `&mut [u8]` as well as `&[u8]`.

> _This is particularly relevant for types which implement `std::io::Read` and `std::io::Write` in order to reduce loop unwinding complexity. It would be nice to replace complex write implementation by their stubs so that other methods which call them can be verified without blowing up proving time. A concrete example of this can be found [here](https://github.com/Trantorian1/http/blob/13aac40fe314b48319ad1c71fc65619e3e5e7da0/http_primitives/src/mem/stream/mod.rs#L223). Without stubbing, an [end-to-end test using a custom std::io::Write implementor](https://github.com/Trantorian1/http/blob/13aac40fe314b48319ad1c71fc65619e3e5e7da0/http_primitives/src/mem/buffer/write.rs#L109) ends up taking 1h40 and 72GB of RAM to validate, even on relatively small buffer sizes._

The following code provides a minimal reproducible example:

```rust
//! src/lib.rs

// Bogus implementation which will end up using the verified stub during validation
pub struct StackBuffer<'data> {
    buf: &'data mut [u8],
}

impl<'data> StackBuffer<'data> {
    // `is_clear` is replaced here by `stub_verified`
    pub fn new(buf: &'data mut [u8]) -> Self {
        assert!(util::is_clear(buf));
        Self { buf }
    }

    // The function we want to validate with the stub.
    pub fn len(&self) -> usize {
        self.buf.len()
    }
}

mod util {
    #[cfg_attr(kani, kani::modifies(buf))]
    #[cfg_attr(kani, kani::ensures(|res| *res == buf.iter().all(|n| *n == 0)))]
    pub fn is_clear(buf: &[u8]) -> bool {
        for i in 0..buf.len() {
            if buf[i] != 0 {
                return false;
            }
        }
        true
    }
}

#[cfg(all(test, kani))]
mod validate {
    use super::*;

    const MAX_SIZE: usize = 16;
    const _: () = assert!(MAX_SIZE > 0);

    #[kani::proof]
    #[kani::unwind(17)]
    fn is_clear_harness() {
        let data: [u8; MAX_SIZE] = kani::any();
        let oracle = data;

        let size = kani::any();
        kani::assume(size < MAX_SIZE);

        assert_eq!(
            util::is_clear(&data[..size]),
            oracle[..size].iter().all(|n| *n == 0)
        );
    }

    #[kani::proof_for_contract(util::is_clear)]
    #[kani::unwind(17)]
    fn is_clear_contract() {
        let data: [u8; MAX_SIZE] = std::array::from_fn(|i| i as u8);

        let size = kani::any();
        kani::assume(size < MAX_SIZE);

        util::is_clear(&data[..size]);
    }

    #[kani::proof]
    #[kani::unwind(17)]
    #[kani::stub_verified(util::is_clear)] // Removing this line fixes the compilation error
    fn buffer_len() {
        let mut data = [0; MAX_SIZE];

        let size = kani::any();
        kani::assume(size < MAX_SIZE);

        let buffer = StackBuffer::new(&mut data[..size]);

        assert_eq!(buffer.len(), size);
    }
}
```

```
kani src/lib.rs --tests --harness validate::buffer_len -Z function-contracts -Z stubbing
```

with Kani version: kani 0.67.0, commit [`2bf550f`](https://github.com/model-checking/kani/tree/2bf550f91c76e6ef439f742cb7963f26d2925678).

I expected to see this happen:

`is_clear` should have been replaced by its verified stub in `validate::buffer_len`.

Instead, this happened:

```bash
thread 'rustc' (162965) panicked at kani-compiler/src/kani_middle/reachability.rs:420:69:
called `Result::unwrap()` on an `Err` value: Error("Failed to resolve `FnDef(DefId { id: 402, name: \"std::iter::IntoIterator::
into_iter\" })` with `GenericArgs([Type(Ty { id: 125, kind: RigidTy(Ref(Region { kind: ReErased }, Ty { id: 126, kind: RigidTy(
Slice(Ty { id: 31, kind: RigidTy(Slice(Ty { id: 22, kind: RigidTy(Uint(U8)) })) })) }, Mut)) })])`")
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: <kani_compiler::kani_middle::reachability::MonoItemsCollector>::reachable_items
   4: kani_compiler::kani_middle::reachability::collect_reachable_items
   5: <kani_compiler::codegen_cprover_gotoc::compiler_interface::GotocCodegenBackend>::codegen_items
   6: <scoped_tls::ScopedKey<core::cell::Cell<*const ()>>>::set::<rustc_public::compiler_interface::run<rustc_public::rustc_int
ernal::run<<kani_compiler::codegen_cprover_gotoc::compiler_interface::GotocCodegenBackend as rustc_codegen_ssa::traits::backend
::CodegenBackend>::codegen_crate::{closure#0}, alloc::boxed::Box<dyn core::any::Any>>::{closure#0}, alloc::boxed::Box<dyn core:
:any::Any>>::{closure#0}, core::result::Result<alloc::boxed::Box<dyn core::any::Any>, rustc_public::error::Error>>
   7: rustc_public::rustc_internal::run::<<kani_compiler::codegen_cprover_gotoc::compiler_interface::GotocCodegenBackend as rus
tc_codegen_ssa::traits::backend::CodegenBackend>::codegen_crate::{closure#0}, alloc::boxed::Box<dyn core::any::Any>>
   8: <kani_compiler::codegen_cprover_gotoc::compiler_interface::GotocCodegenBackend as rustc_codegen_ssa::traits::backend::Cod
egenBackend>::codegen_crate
   9: <rustc_interface::queries::Linker>::codegen_and_build_linker
  10: rustc_interface::interface::run_compiler::<(), rustc_driver_impl::run_compiler::{closure#0}>::{closure#1}

[Kani] no current codegen item.
[Kani] no current codegen location.
```

feliperodri:
## Description

`#[kani::modifies(x)]` on a whole-slice argument (e.g. `x: &mut [u8]`) crashed the compiler when the contract was reused as a verified stub via `#[kani::stub_verified]`. Reachability collection panicked with:

```
Failed to resolve `FnDef(... "std::iter::IntoIterator::into_iter")`
with `GenericArgs([Ref(ReErased, Slice(Slice(Uint(U8))), Mut)])`
```

i.e. a doubled slice `&mut [[u8]]`.

## Context

`AnyModifiesPass::replace_any_modifies` rewrites the `write_any` marker into `write_any_slim` / `write_any_slice` / `write_any_str` based on the pointee type. `write_any_slice<T>(slice: *mut [T])` is generic over the **element** type `T`, but the pass resolved it with the marker's `instance_args`, which hold the **pointee** type `[T]`. That produced `write_any_slice::<[u8]>(*mut [[u8]])`; its `fill_with(T::any)` body then iterated `&mut [[u8]]`, and `into_iter` failed to resolve because `[u8]` is neither `Sized` nor `Arbitrary`.

This path is only reached in replace mode (`stub_verified`), which is why `proof_for_contract` harnesses were unaffected and the existing `modifies_fat_pointer` tests (all check-mode) did not catch it — removing the `#[kani::stub_verified]` line from the reproducer avoids the crash.

The fix resolves `write_any_slice` with the slice's element type instead of the pointee type.

## Resolved issues

Resolves #4748

## Testing

- Added `tests/expected/function-contract/modifies_fat_pointer/slice_replace.rs`, which exercises both `proof_for_contract` (check) and `stub_verified` (replace) on a `&mut [u8]` slice-modifies contract — the replace harness reproduces the crash before this change.
- Verified locally that the reproducer from the issue and the new test now complete codegen without panicking.


tautschnig (approved):

Trantorian1:
Hey @feliperodri, thanks a lot for the fix! Can confirm, all works good on my end as well now :)

PR #4756, opened 2026-08-25
Upgrade Rust toolchain to nightly-2026-05-01

feliperodri:
### Description

Bumps `rust-toolchain.toml` from `nightly-2026-04-01` to `nightly-2026-05-01`.

#### What changed upstream, and what it required here

**1. `NonNull<T>` now holds a pattern type.** This is the root cause of most of the breakage:

```rust
// nightly-2026-04-01
pub struct NonNull<T: PointeeSized> { pointer: *const T }
// nightly-2026-05-01
pub struct NonNull<T: PointeeSized> { pointer: crate::pattern_type!(*const T is !null) }
```

Kani codegens a pattern type as a struct with a single tuple-like field holding the type it constrains, so the raw pointer behind `NonNull` moved one level deeper. Five places assumed a fixed nesting depth, and because `DynMetadata` holds a `NonNull<VTable>` this reached vtable/metadata codegen too:

| File | Change |
| --- | --- |
| `kani_middle/coercion.rs` | New `RigidTy::Pat` arm in `CoerceUnsizedIterator::next`, so the unsize-coercion path steps through the pattern type (field `"0"`) and the base case reaches the real pointer. |
| `codegen/rvalue.rs` | `codegen_struct_unsized_coercion` accepts pattern types via a new `is_struct_like` (ADT *or* `Pat`, both codegen'd as assignable structs). Both `DynMetadata` sites use the new recursive helpers instead of hardcoded `.member("pointer")` / one-level `lookup_components()[0]`. |
| `utils/utils.rs` | New `codegen_ptr_in_wrappers` (wrap a pointer in N single-field struct layers) and `codegen_ptr_out_of_wrappers` (peel them off). |
| `codegen/typ.rs` | `receiver_data_path` walks through pattern types; `codegen_trait_receiver`'s rebuild handles a `Pat` path element. |
| `codegen/place.rs` | `codegen_field` projects into a pattern type (reads field `"0"`) instead of `unreachable!`. Rust itself forbids this projection, but Kani synthesizes one in `raw_ptr_arg`, so it must be supported. |
| `check_uninit/.../uninit_visitor.rs` | `raw_ptr_arg` derives its projection chain from the actual field types via a new `single_field_ty`, looping until it reaches a raw pointer, instead of pushing one hardcoded `Field(0, *const T)`. |

All of these are written to be **depth-agnostic** rather than adding one more hop. `raw_ptr_arg` had already broken once on exactly this assumption (when the allocation shims switched from `*mut u8` to `NonNull<u8>` in nightly-2026-03-21), so hardcoding the new depth would only defer the next break.

**2. `rustc_public` lost the `ty::Alias` stable conversion.** The alias kind moved into `AliasTy` and the conversion was left as `todo!()`, so the stable `Ty::kind()` now panics on an un-normalized alias. `kani-compiler` normalizes before inspecting types and is unaffected; the `scanner` tool does not, so it ICE'd. `tools/scanner/src/{lib,analysis,call_graph}.rs` now thread `TyCtxt` down to `BodyVisitor`/`Node::try_new` so the union check uses `rustc_internal::internal(tcx, ty).is_union()`. The internal representation reports an alias as not-a-union, which is exactly what the stable `TyKind::Alias` did before it became unconvertible, so behaviour is preserved rather than just the crash avoided. **Worth reporting upstream.**

**3. Assorted API changes:**

| Change | Site |
| --- | --- |
| `rustc_middle::mir::mono::CodegenUnitNameBuilder` moved to `rustc_middle::mono` | `utils/names.rs` |
| `type_of(..).instantiate*()` and `predicates_of(..).instantiate(..)` now return an unnormalized wrapper; need `.skip_normalization()` | `codegen_units.rs`, `intrinsics.rs`, `stubbing/mod.rs`, `codegen/typ.rs` |
| `tcx.has_attrs_with_path` removed; use `get_attrs_by_path(..).next().is_some()` | `check_uninit/ptr_uninit/mod.rs` |
| `test::run_tests_console` now takes a `TestList` tagged with an ordering | `tools/compiletest/src/main.rs` |

On the compiletest change: the list must be tagged `Unsorted`. compiletest collects tests by walking directories, so it is not name-ordered, and the harness binary-searches a list tagged `Sorted`.

#### Note on a previously-suspected regression

An earlier pass on this upgrade flagged `tests/cargo-kani/mir-linker` as a genuine non-termination regression (`memchr_naive` unwinding 3400+ times against a 4.4s baseline on 04-01). **It is not a regression.** I reduced it to a standalone reproducer with the same harness and pinned dependency versions: `semver 1.0.22` hangs, `semver 1.0.28` verifies in seconds. That test's manifest requires only `semver = "1.0"`, and `tests/cargo-kani/.gitignore` ignores `Cargo.lock`, so no lockfile is tracked — the 1.0.22 pin was a stale local artifact. A fresh checkout resolves the newest `1.0.x` and the test passes; after refreshing the local lockfile it reports `VERIFICATION:- SUCCESSFUL` with 40 unwinding lines. No code change was needed, and the full `cargo-kani` suite is 71/71.

There is also an optional follow-up cleanup, deliberately left out to keep this diff focused: `PatternKind::NotNull` became convertible in 05-01, which makes the internal-representation workaround in `check_uninit/../check_values.rs` (`is_pattern_type`, added for nightly-2026-03-01) redundant. It is still correct.

### Testing

Local, macOS aarch64, CBMC 6.10.0 (`cbmc-6.9.0-214-g45436eea34`):

| Suite | Result |
| --- | --- |
| `kani` | **607 passed**, 0 failed, 23 ignored |
| `cargo-kani` | **71 passed**, 0 failed (previously 70/71, see above) |
| `script-based-pre` | **66 passed**, 0 failed, 1 ignored (includes `tool-scanner`, which covers the `ty::Alias` fix, and `verify_std_cmd`) |
| `std-checks` | **5 passed**, 0 failed (this suite was never run on the earlier pass) |
| `cargo-ui` | 30 passed, 0 failed |
| `coverage` | 20 passed, 0 failed |
| `expected` (targeted: `dynamic-trait`, `uninit`) | 28 passed, 0 failed |
| `kani` (targeted: `DynTrait`, `Uninit`) | 35 passed, 0 failed |

Other gates, all clean:

- `cargo build-dev`
- `cargo build-dev -- --features cprover --features llbc`
- `cargo clippy --workspace --tests`
- Unit tests: `cprover_bindings`, `kani-compiler`, `kani-driver`, `kani_metadata`

I ran targeted subsets rather than the full remaining suites (`expected`, `ui`, `firecracker`, `prusti`, `smack`, `kani-fixme`, `cargo-coverage`, `kani-docs`) locally, since CI covers those. Two caveats on local runs: this CBMC build has no `cadical`, so `ui/solver-{attribute,option}/cadical` fail locally on output text only, and `expected/shadow/slices/slice_split` is very slow under the MiniSat fallback. Both are expected to be clean on CI.

### DIFF
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/place.rs
@@ -296,9 +296,17 @@ impl GotocCtx<'_, '_> {
                         )
                     }
                     TyKind::RigidTy(RigidTy::Pat(..)) => {
-                        // See https://github.com/rust-lang/types-team/issues/126
-                        // for what is currently supported.
-                        unreachable!("projection inside a pattern is not supported, only transmute")
+                        // Rust itself does not allow projecting inside a pattern type, only
+                        // transmuting it: see https://github.com/rust-lang/types-team/issues/126
+                        // for what is currently supported. Kani does synthesize such a projection to
+                        // see through the pattern type that `NonNull` wraps its address in, and
+                        // codegens a pattern type as a struct with a single tuple-like field
+                        // holding the constrained type, so read that field.
+                        assert_eq!(
+                            field_idx, 0,
+                            "a pattern type only has field 0, but found {field_idx}"
+                        );
+                        Ok(parent_expr.member(Self::tuple_fld_name(field_idx), &self.symbol_table))
                     }
                 }
             }
--- a/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
+++ b/kani-compiler/src/codegen_cprover_gotoc/codegen/rvalue.rs
@@ -702,13 +702,11 @@ impl GotocCtx<'_, '_> {
                                     slice_fat_ptr(typ, data_cast, meta, &self.symbol_table)
                                 }
                                 RigidTy::Dynamic(..) => {
-                                    let vtable_expr = meta
-                                        .member("_vtable_ptr", &self.symbol_table)
-                                        .member("pointer", &self.symbol_table)
-                                        .cast_to(
-                                            typ.lookup_field_type("vtable", &self.symbol_table)
-                                                .unwrap(),
-                                        );
+                                    let vtable_expr = self.codegen_ptr_out_of_wrappers(
+                                        meta,
+                                        typ.lookup_field_type("vtable", &self.symbol_table)
+                                            .unwrap(),
+                                    );
                                     dynamic_fat_ptr(typ, data_cast, vtable_expr, &self.symbol_table)
                                 }
                                 _ => {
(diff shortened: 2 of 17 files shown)


rajath-mk (approved):
LGTM

Issue #4759, opened 2026-08-25
vtable_size_align_drop no longer checks the drop slot holds the right type's drop glue


feliperodri:
### Context

`tests/kani/DynTrait/vtable_size_align_drop.rs` reads a vtable's raw layout through the pointer-metadata API and asserts each of the three header slots:

```rust
assert!(drop_from_vtable(vtable_ptr) == drop_in_place::<Sheep> as *mut ());
assert!(size_from_vtable(vtable_ptr) == size_of::<i32>());
assert!(align_from_vtable(vtable_ptr) == size_of::<i32>());
```

### Problem

As of `nightly-2026-06-01` (#4757) the drop-glue shim that rustc places in the vtable's first slot is `core::ptr::drop_glue::<T>`, not `core::ptr::drop_in_place::<T>`. Traced with Kani's own debug logging on `codegen_vtable_drop_in_place`:

```
drop_instance=Instance { kind: Shim, def: "_RINvNtCs..._4core3ptr9drop_glueNtCs..._5probe5SheepEBD_", .. }
```

Kani puts the right thing in the slot — `Instance::resolve_drop_in_place(T)`'s mangled symbol — but that is now a *different* function from the user-callable `core::ptr::drop_in_place::<T>` that the test compares against, and `drop_glue` is not nameable from Rust source. Confirmed with cover checks that the vtable slot and `drop_in_place::<T>` are provably distinct for both of the test's types.

Note the vtable *layout* is unchanged: `COMMON_VTABLE_ENTRIES` is still `[MetadataDropInPlace, MetadataSize, MetadataAlign]` with the drop pointer at index 0. Only the identity of the function in that slot changed.

This was verified to be a genuine `nightly-2026-06-01` regression rather than a pre-existing fragility: rebuilding the parent commit against `nightly-2026-05-01` passes.

### What #4757 did

Since the test's stated purpose is the `size` and `align` fields (it is named for them, and those use the public `DynMetadata::size_of()`/`align_of()` API), #4757 replaced the two exact-identity assertions with a non-null check on the slot, with a comment explaining why:

```rust
assert!(!drop_from_vtable(vtable_ptr).is_null());
```

### Why this is worth tracking

The weaker assertion no longer checks that the slot holds the drop glue *for the right type* — a codegen bug that put another type's drop glue (or any other function pointer) in the slot would now pass. Restoring that coverage needs a way to name the drop-glue shim, or to compare against it indirectly. Options:

[...]

Issue #4758, opened 2026-08-25
Autoharness: generate values for pattern-type fields, restoring layout-niche coverage for user-defined ranged types


feliperodri:
### Context

#4716 added an assumption of a type's layout niche after each nondeterministically generated value, so that autoharness stops producing language-level-invalid values for ranged scalar types (e.g. a date type packing a validated ordinal) and stops reporting the resulting false alarms.

The toolchain upgrade to `nightly-2026-06-01` (#4757) removed `rustc_layout_scalar_valid_range_start`/`_end` in favour of pattern types — the same move `core::num::niche_types` made upstream. A ranged scalar newtype is now written as:

```rust
#[repr(transparent)]
pub struct Month(std::pat::pattern_type!(u8 is 1..=12));
```

### Problem

A pattern type is not an ADT and has no `Arbitrary` implementation, so `can_derive_arbitrary` cannot synthesize a struct that has one as a field. Every locally-defined ranged newtype is therefore **skipped** rather than harnessed:

```
| niche_probe | days_left_in_year | Missing Arbitrary implementation for argument(s) s: Schedule |
| niche_probe | signed_niche      | Missing Arbitrary implementation for argument(s) p: PosI8   |
```

So the niche assumption, which is the thing #4716 added, no longer applies to the user-defined types it was aimed at. It still works end to end for `std::time::Duration`, whose `Nanoseconds` field carries the same kind of range and which does get harnessed — that is the motivating real-world case from the #3832 triage, so the feature is not dead, just much narrower than intended.

`tests/script-based-pre/autoharness_niche` pins both halves (the working `Duration`/`NonZeroU8` cases and the skipped local ones) so this limitation is asserted by a test rather than silently regressing further.

### Proposed fix

Teach autoharness to generate values for pattern-type fields. The pieces already exist:

- `scalar_niche` (added in #4716, `kani-compiler/src/kani_middle/mod.rs`) already computes the `(bits, start, end)` valid range from the *layout*, and a pattern type's layout carries exactly that restricted range.
- `call_kani_any_for_ty` already knows how to generate a base integer and how to emit `kani::assume` for a niche.

So generating a `pattern_type!(u8 is 1..=12)` is: generate the base integer with `kani::any`, assume `scalar_niche`'s range, and transmute. Concretely that means:

1. `can_derive_arbitrary` accepting `RigidTy::Pat` whose base type is generatable.
[...]

PR #4760, opened 2026-08-25
Upgrade Rust toolchain to nightly-2026-06-01

feliperodri:
### Description

Bumps `rust-toolchain.toml` from `nightly-2026-05-01` to `nightly-2026-06-01`. Four upstream changes drive most of this.

#### 1. `Retag` moved onto `Rvalue::Use`

`StatementKind::Retag` and `RetagKind` are gone; `Rvalue::Use` now carries a `WithRetag` flag instead:

```rust
Use(Operand, WithRetag),
```

Kani never modelled retags (they are Stacked-Borrows/Miri only), so the statement arms are dropped, the flag is ignored when reading a `Use`, and synthesized `Use`s pass `WithRetag::No`. `internal_mir`'s `RetagKind` conversion becomes a `WithRetag` one.

#### 2. `Variants::Multiple` stores a `VariantLayout`, not a `LayoutData`

The new type carries only per-field offsets — no `FieldsShape` (so no field order) and no alignment. Variant layouts now come from `Layout::for_variant`, which is what `rustc_codegen_ssa` does and which restores the field order. A new `variant_layout` helper is used by **both** the type side (`codegen_enum_cases`) and the value side (`codegen_aggregate`), so the goto struct's components and the operands initializing them cannot disagree.

That change bit back: `for_variant` reports the *enum's* align for a variant (`align: parent.align`), which over-pads every variant and inflates the enum. `check_vtable_size` caught it as an ICE on `tests/cargo-kani/iss2857` (48 vs 55 bytes). So `codegen_struct_fields`/`codegen_alignment_padding` now take the align explicitly, and a variant's own align is computed as the maximum of its fields' aligns — which is what the per-variant `LayoutData` used to carry.

#### 3. `rustc_layout_scalar_valid_range_start`/`_end` were removed

Replaced by pattern types, the same move `core::num::niche_types` made. The tests that define ranged scalar newtypes are converted to `std::pat::pattern_type!`.

This narrows the layout-niche feature added in #4716: a pattern type is not an ADT and has no `Arbitrary` implementation, so `can_derive_arbitrary` cannot synthesize a struct that has one as a field, and locally-defined ranged types are now skipped rather than harnessed. The assumption still works end to end through `std::time::Duration` (the motivating case from the #3832 triage), and `tests/script-based-pre/autoharness_niche` pins both halves so the reduced reach is asserted by a test rather than silent. Tracked in #4758 with a concrete fix sketch.

#### 4. New `Rvalue::Reborrow`

User-definable reborrowing of ADTs via `CoerceShared`. It is documented as a bitwise copy today, but the same docs anticipate it changing memory layout, so codegen reports it as an unsupported construct rather than silently modelling it as a copy. The points-to analysis treats it as pointing wherever its place does.

#### Assorted API changes

| Change | Site |
| --- | --- |
| `CodegenBackend` moved `CrateInfo` from `codegen_crate` to `join_codegen` | both backends' `compiler_interface.rs` |
| `rustc_data_structures::stable_hasher` renamed to `stable_hash`; `HashStable`/`hash_stable` to `StableHash`/`stable_hash` | `reachability.rs` |
| The `normalize` callback of `ptr_metadata_ty{,_or_tail}` now takes `Unnormalized` | `typ.rs` |
| More `FieldDef::ty` and `instantiate*` sites need `.skip_normalization()` | `typ.rs`, `intrinsics.rs` |
| `TagEncoding::Niche`'s `niche_variants` is the lang `RangeInclusive` (public `start`/`last` fields, not methods) | `rvalue.rs`, `statement.rs` |
| New `useless_borrows_in_formatting` clippy lint | 11 sites across the workspace |

#### Test change worth calling out

`vtable_size_align_drop` no longer asserts the exact identity of the vtable's drop pointer. Traced with Kani's debug logging: the drop-glue shim is now `core::ptr::drop_glue::<T>` rather than `core::ptr::drop_in_place::<T>`, and `drop_glue` is not nameable from source. It checks the slot is populated instead; the size and align fields the test is named for are unchanged. This weakens the test — it no longer checks the slot holds the drop glue *for the right type* — so it is tracked in #4759 with options for restoring that coverage.

Both of these failures were confirmed to be genuine `nightly-2026-06-01` regressions rather than pre-existing fragility, by rebuilding the parent commit against `nightly-2026-05-01`, where both pass.

#### `drop_in_place` now takes a reference to reach the drop glue

Related to the same rename: `core::ptr::drop_in_place` used to *be* the drop lang item, with its body replaced wholesale by the compiler. It is now a wrapper that calls the new `drop_glue` lang item through `&mut *to_drop`, and creating that reference asserts the pointee is aligned and valid — so dropping through a raw pointer carries a precondition it did not before.

That surfaced in `expected/loop-contract/loop_assigns_for_vec.rs`, whose `loop_modifies` havocs the word holding `v.len()` while the invariant only constrains `i`; the drop after the loop then sees an unconstrained length. Strengthening the invariant with `v.len() == i * 3 + 3` is rejected (`Rust intrinsic assumption failed`), so expressing what the drop needs looks like a loop-contracts limitation rather than something this PR can fix. The test is renamed `loop_assigns_for_vec_fixme.rs` with the analysis recorded in the file, tracked in #4761. Verified that `Vec`'s internal layout is unchanged, so the test's hard-coded `.add(2)` offset for `len` is still correct.

Two contract tests that deliberately pin a *total check count* needed their numbers updated: `capture_load_checks_elided` 36 → 34 and `capture_load_user_deref_pass` 60 → 58 (two checks in `read_answer`'s own body became unreachable). In both cases the assertion the test actually exists for still holds — no user-facing `pointer_dereference` group came back in the first, and the pinned `{closure#3}::{closure#0}.pointer_dereference.1 - Status: SUCCESS` is still present in the second — so the counts were updated with a note in each test explaining that the total is plumbing-sensitive and what it still guards against.

### Testing

Local, macOS aarch64, CBMC 6.10.0 (`cbmc-6.9.0-214-g45436eea34`), all re-run after rebasing onto current `main` (which now includes #4717, whose new `Rvalue::Use` construction site also needed the `WithRetag` argument):

| Suite | Result |
| --- | --- |
| `kani` | **607 passed**, 0 failed, 23 ignored |
| `cargo-kani` | **71 passed**, 0 failed |
| `script-based-pre` | **68 passed**, 0 failed, 1 ignored |
| `std-checks` | 5 passed, 0 failed |
| `cargo-ui` | 30 passed, 0 failed |
| `coverage` | 20 passed, 0 failed |
| `ui` | 151 passed, 2 failed — both the `cadical` tests, see below |
| `expected` | **474 passed**, 0 failed, 16 fixme/ignored |
| `prusti` / `smack` / `kani-docs` / `json-handler` / `cargo-coverage` / `firecracker` | 8 / 40 / 13 / 5 / 2 / 0 passed, 0 failed |

Other gates, all clean:

- `cargo build-dev`
- `cargo build-dev -- --features cprover --features llbc`
- `cargo clippy --workspace --tests` and `RUSTFLAGS="--cfg=kani_sysroot" cargo clippy --workspace`
- `RUSTFLAGS="-D warnings" cargo build --no-default-features --features cprover`
- Unit tests: `cprover_bindings`, `kani-compiler`, `kani-driver`, `kani_metadata`, `kani --features concrete_playback`, `kani_macros`

Environment caveat: this CBMC build has no `cadical`, so `ui/solver-{attribute,option}/cadical` fail locally on output text only (`"The specified solver, 'cadical', is not available"`), and `expected/shadow/slices/slice_split` is very slow under the MiniSat fallback. Both are expected to be clean on CI.

Towards #4758, #4759, #4761.


feliperodri:
The next upgrade, `nightly-2026-07-01`, is up as a draft in #4764 stacked on this one. It is much smaller (no verification behaviour changed and no test needed adjusting) and should be reviewed once this lands.

rajath-mk (approved):

